PR #397, opened 2016-09-21
Warn when a shrinkwrap exists in the cwd

### LINKED ISSUE #396: Warn about npm shrinkwraps
**Do you want to request a _feature_ or report a _bug_?** Feature

Yarn won't be able to respect npm shrinkwraps because of how lossy they are and hoisted dependencies.

However, users migrating from npm to Yarn who currently use npm shrinkwraps will get different versions of packages because Yarn isn't respecting their shrinkwrap. This causes things to break with Yarn that didn't with npm because the version we install may have a breaking change (even by accident).

When running install in a directory with an npm shrinkwrap we should have a warning that explains we aren't going to respect it and maybe link to a page that explains how to deal with it.

sebmck:
Fixes #396.

### DIFF
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -51,7 +51,7 @@ let messages = {
   invalidHostedGitFragment: 'Invalid hosted git fragment $0.',
   couldntFindManifestIn: "Couldn't find manifest in $0.",
   invalidFragment: 'Invalid fragment $0.',
-  shrinkwrapWarning: 'npm-shrinkwrap.json found. This will not be updated or respected. See TODO for more information.',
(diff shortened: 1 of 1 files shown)


src/cli/commands/install.js
```diff
@@ -199,6 +199,11 @@ export class Install {
     // remove integrity hash to make this operation atomic
     await fs.unlink(match.loc);
 
+    // warn if we have a shrinkwrap
+    if (await fs.exists(path.join(this.config.cwd, 'npm-shrinkwrap.json'))) {
+      this.reporter.error(this.reporter.lang('shrinkwrapWarning'));
```
Turbo87: shouldn't this be `reporter.warn(...)` instead? I'd expect `error` only when the issue is not recoverable.

shahmirn:
This also relates to https://github.com/yarnpkg/yarn/issues/396

PR #415, opened 2016-09-22
Only enable emoji for OS X

sebmck:


src/cli/index.js
```diff
@@ -94,7 +94,7 @@ if (commander.json) {
   Reporter = JSONReporter;
 }
 let reporter = new Reporter({
-  emoji: true,
+  emoji: process.stdout.isTTY && process.platform === 'darwin',
```
elyobo: :+1: for adding a flag to allow people to override it, will add PR (as I can't see any that do this yet).

PR #501, opened 2016-10-05
Add error when `version` is used without `--new-version`.

### LINKED ISSUE #472: yarn version [version]
**Do you want to request a _feature_ or report a _bug_?** Feature

When you enter `yarn version` it opens up a prompt to enter a new version. This is hard for scripts to deal with.

The following should create a new version without prompting the user:

``` sh
yarn version [1.0.0,major,minor,patch,etc]
```

See: https://docs.npmjs.com/cli/version

cpojer:
**Summary**
Fixes #472.

I'm always put off by APIs like `npm version <number>` so I think `--new-version` makes a lot of sense. However, we should at least error in the case someone tries to use an npm command in yarn and hint at the proper argument.

**Test plan**
- `yarn version 1.2.3` throws.
- `yarn version` or `yarn version --new-version 1.2.3` work.

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -12,12 +12,13 @@ let invariant = require('invariant');
 let semver = require('semver');
 let path = require('path');
 
+const NEW_VERSION_FLAG = '--new-version [version]';
 function isValidNewVersion(oldVersion: string, newVersion: string, looseSemver: boolean): boolean {
   return !!(semver.valid(newVersion, looseSemver) || semver.inc(oldVersion, newVersion, looseSemver));
 }
 
 export function setFlags(commander: Object) {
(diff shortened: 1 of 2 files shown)


adamreisnz:
I agree, I don't see why because someone is "put off" by API's that just work we now need to type 14 additional characters just to release a new version.

These kind of things are preventing me from fully embracing yarn as an npm replacement. Why fix something that wasn't broken? Half the internet is used to typing `npm version patch`, so why include this redundant `--new-version` flag?

blakek:
Would a PR that makes the `--new-version` flag optional would be positively received?
Thanks!

adamreisnz:
That'd be great 👍

PR #516, opened 2016-10-05
Add `generate-lock-entry` command

### LINKED ISSUE #198: tool to generate kpm.lock file entries
Sometimes for testing or as a way around issues we need to generate a lock file instead of editing it manually.

API could look like this:

```
kpm generate-lock-entry --name yeoman-welcome@^1.0.0 --resolved  https://example.com/yeoman-welcome-1.0.1.tgz
```

Output:

```
yeoman-welcome@^1.0.0:
  name yeoman-welcome
  version "1.0.1"
  resolved https://example.com/yeoman-welcome-1.0.1.tgz#f6cf198fd4fba8a771672c26cdfb8a64795c84ec
  dependencies:
    chalk "^1.0.0"
```

### LINKED ISSUE #342: ability for offline e2e installs on CI
We need to test frameworks (React Native, Reason) with e2e tests on CI that don't require network connectivity.

---

Example:
1.React Native has offline-mirror of dependencies and a kpm.lock file
2.On CI I want to install a React Native app
3.Generate a package.json

```
{
  "name": "packager-test",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start"
  },
  "dependencies": {
    "react-native": "1000.0.0"
  }
}
```

4.Generate kpm.lock (???)
5.Run kpm install from a folder with package.json and kpm.lock

---

Before 0.13.0 I could append React Native's kpm.lock with a dummy

```
react-native@1000.0.0:
  version "1000.0."
  resolved react-native-1000.0.0.tgz#$REACT_NATIVE_HASH
```

and it would get me a working React Native App lockfile.
But it does not work anymore because it kpm clears out all dependencies that are not listed in lock file as dependencies of react-native@1000.0.0.

sebmck:
Fixes #198, fixes #342.

### DIFF
--- a/src/cli/commands/generate-lock-entry.js
+++ b/src/cli/commands/generate-lock-entry.js
@@ -1,0 +1,58 @@
+/* @flow */
+
+import type {Reporter} from '../../reporters/index.js';
+import type Config from '../../config.js';
+import {MessageError} from '../../errors.js';
+import {implodeEntry} from '../../lockfile/wrapper.js';
+import stringify from '../../lockfile/stringify.js';
+import * as fs from '../../util/fs.js';
+
+export function hasWrapper(): boolean {
+  return false;
+}
+
+export async function run(
+  config: Config,
+  reporter: Reporter,
+  flags: Object,
+  args: Array<string>,
+): Promise<void> {
+  let manifest;
+  if (flags.useManifest) {
+    manifest = await fs.readJson(flags.useManifest);
+  } else {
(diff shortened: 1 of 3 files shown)


rarkins:
@kittens I don't think this works (ever worked?) as intended and should ideally be removed or fixed (see #2340)

expelledboy:
There are quite afew usecases for this with yarn workspaces. Ideally fix 👍

PR #562, opened 2016-10-10
Add team command

sebmck:


src/util/request-manager.js
```diff
@@ -287,6 +287,11 @@ export default class RequestManager {
 
         successHosts[parts.hostname] = true;
 
+        if (body && typeof body.error === 'string') {
```
EECOLOR: This is possibly problematic. In some cases the response is a 40X status with a body that contains an error property. An example is a 404 that has a json body containing error with 'Not Found' (if I recall correctly the registry sometimes returns this).

In the lines below you can see:
```
if (res.statusCode === 400 || res.statusCode === 404) {
  body = false;
}
```
The behavior here is that body is set to false on these status codes, but if the response body happened to contain an error property, the behavior is changed.

BYK: @EECOLOR mind submitting a PR to change it?>

EECOLOR: I wouldn't mind if knew what the behavior needs to be. Without extra knowledge it's impossible to change and ensure other parts will not break. If the code was created in order to reject on a 404, and I changed it into the other 404 behavior (the one that resolves with `false`) it would break.

The commit doesn't help much for an outside like me because I don't know what motivated the addition of this line. My guess is that there was an API that returned 200 with an error in the body, but we would need input from the author of this code.

PR #594, opened 2016-10-11
allow proxy to be specified via yarn/npm config

### LINKED ISSUE #575: Yarn does not observe proxy settings in ~/.npmrc (or ~/.yarnrc)
Both?

Yarn does not observe proxy settings in `~/.npmrc`, resulting in `error connect ENETUNREACH` errors in environments that require the usage of a proxy to reach the internet.

If the user has a proxy configured for `npm`, we can probably assume that they want to use the same proxy for Yarn too. Example of an `.npmrc` specifying a proxy:

```
proxy=http://fwdproxy:8080/
https-proxy=http://fwdproxy:8080
```

At the moment, it looks like [we do parse the `.npmrc` file](https://github.com/yarnpkg/yarn/blob/d3e3433ce288bce15961bdb2784ebbe2becac0e0/src/registries/npm-registry.js#L109), but we're not using the proxy specified within.

This blocks using Yarn on devservers at Facebook, as outbound network access is firewalled and we need to use a proxy to access the web from them.

yarn 0.14
CentOS 7

sebmck:
Fixes #575.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -159,6 +159,8 @@ export default class Config {
 
     this.requestManager.setOptions({
       userAgent: String(this.getOption('user-agent')),
+      httpProxy: String(this.getOption('proxy') || ''),
+      httpsProxy: String(this.getOption('https-proxy') || ''),
     });
   }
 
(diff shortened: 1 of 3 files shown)


jgogstad:
Would you care to add support for `no-proxy` as well? Npm supports it, but only through environment variables at the moment. It's on the roadmap though: https://github.com/npm/npm/issues/7168

PR #608, opened 2016-10-11
fix typo: self-udpate => self-update

azu:
**Summary**

fix a typo.


cpojer:
classic!

torifat:
Ha ha ha, This is why I added a `udpate -> update` with my `update` pull request 😄

PR #622, opened 2016-10-11
chore(package,yarn): remove eslint-plugin-flow-vars

mrtnbroder:
eslint-plugin-flow-vars is deprecated and has been merged into eslint-plugin-flowtype.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "eslint-config-fbjs": "^1.0.0",
     "eslint-config-kittens": "^2.0.1",
     "eslint-plugin-babel": "^3.3.0",
-    "eslint-plugin-flow-vars": "^0.4.0",
     "eslint-plugin-flowtype": "^2.15.0",
     "eslint-plugin-no-async-without-await": "^1.0.0",
     "eslint-plugin-yarn-internal": "file:scripts/eslint-rules",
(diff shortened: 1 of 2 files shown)


kittens:
Can you double check that we aren't still referencing this from inside of the eslint config in `package.json`?

mrtnbroder:
It's still a dependency of your personal eslint-config @kittens, https://github.com/kittens/eslint-config-kittens/blob/master/package.json#L19

Other than that, I'm not sure what you mean by "inside of the eslint config"?

pluma:
@mrtnbroder I think Seb means this: https://github.com/yarnpkg/yarn/blob/master/package.json#L99

I.e. make sure the config doesn't rely on flow-vars.

hzoo:
Yeah its good - https://github.com/facebook/jest/blob/master/packages/eslint-config-fb-strict/package.json, https://github.com/facebook/fbjs/pull/181/files

We made the same change in https://github.com/babel/eslint-config-babel/blob/master/package.json which should be the same config as eslint-config-kittens

cpojer:
would you mind rebasing this?

mrtnbroder:
Done, @cpojer !

kittens:
Thank you @mrtnbroder!

PR #625, opened 2016-10-11
Changed Tilde to $HOME

DavidBrear:
Using ZSH when the file would add `export $PATH="~/.yarn/bin:$PATH" to my .zshrc file, the quotes were causing the tilde to not be evaluated to $HOME. This diff changes the tilde to be `$HOME` so even in quotes it is evaluated.

Verified this change caused my `which yarn` to be exposed.


kittens:
Thanks @DavidBrear! Much appreciated.

PR #658, opened 2016-10-11
Ignore request-cache

AlicanC:
**Summary**

Ignored `/__tests__/fixtures/request-cache/**/*`.

**Test plan**

None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /.nyc_output
 /coverage
 /__tests__/fixtures/**/_*
+/__tests__/fixtures/request-cache/**/*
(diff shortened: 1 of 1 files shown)


kittens:
What's the motivation for this change? We require the request cache to speed up CI and not have a reliance on the network.

AlicanC:
It felt like you have pushed it as a mistake since they can regenerate just like `node_modules`.

kittens:
It's intentional, otherwise we're reliant on the network. They'll only be regenerated if new requests are made that aren't in the cache. Thank you for the PR!

PR #676, opened 2016-10-11
Fix self-update

### LINKED ISSUE #612: self-update throws error
bug

```bash
yarn self-update v0.15.0
error OAuth2 authentication requires a token or key & secret to be set
    at __dirname.authenticate (/Users/danieltschinder/.yarn/node_modules/github/lib/index.js:334:23)
    at Object.<anonymous> (/Users/danieltschinder/.yarn/lib-legacy/cli/commands/self-update.js:36:12)
    at next (native)
    at step (/Users/danieltschinder/.yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /Users/danieltschinder/.yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
    at new Promise (/Users/danieltschinder/.yarn/node_modules/core-js/library/modules/es6.promise.js:191:7)
    at Object.<anonymous> (/Users/danieltschinder/.yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12)
    at Object.run (/Users/danieltschinder/.yarn/lib-legacy/cli/commands/self-update.js:105:17)
    at run (/Users/danieltschinder/.yarn/lib-legacy/cli/index.js:249:18)
    at /Users/danieltschinder/.yarn/lib-legacy/cli/index.js:357:12
info Visit http://yarnpkg.com/en/docs/cli/self-update for documentation about this command.
```

```bash
❯ yarn self-update
```

Update

node 4.6.0
yarn 0.15.0
macOS 10.12.0

danez:
This fixes self-update by doing:

* enable the tests
* use npm run build in the test instad of make build
* invalidated roadrunner cache, so that the symlink to `current` gets reevaluated
* do not do the request to github for the assets as they are already included in the release response
* pick the correct asset from the asset list by doing a name-check (content-type is sometimes set to octet-stream for the tar file, so we can't use that)

The roadrunner change is necessary as the tests need to invalidate the path cache so that the next test can reevaluate the symlinks. Same for the actual self-update command which also changes the symlink. This depends on kittens/roadrunner#1

Fixes #612 

Docs: https://github.com/yarnpkg/website/pull/173

### DIFF
--- a/__tests__/commands/self-update.js
+++ b/__tests__/commands/self-update.js
@@ -53,10 +53,10 @@ function run(checks: (reporter: reporters.Reporter, config: Config) => Promise<v
       await config.init({cwd});
       roadrunner.reset(CACHE_FILENAME);
       await checks(reporter, config);
-      roadrunner.reset(CACHE_FILENAME);
     } catch (err) {
       throw new Error(`${err} \nConsole output:\n ${out}`);
     } finally {
+      roadrunner.reset(CACHE_FILENAME);
       await fs.unlink(updatesFolder);
     }
   });
@@ -64,13 +64,14 @@ function run(checks: (reporter: reporters.Reporter, config: Config) => Promise<v
 
 it('Self-update should download a release and symlink it as "current"', (): Promise<void> => {
   return run(async (reporter, config) => {
-    await selfUpdate(config, reporter, {}, ['v0.14.1']);
-    expect(await fs.exists(path.resolve(updatesFolder, 'current')));
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.14.1')));
+    await selfUpdate(config, reporter, {version: () => '0.0.1'}, []);
+
+    expect(await fs.exists(path.resolve(updatesFolder, 'current'))).toBe(true);
     const packageJson = await fs.readJson(path.resolve(updatesFolder, 'current', 'package.json'));
-    expect(packageJson.version === '0.14.1');
+
+    expect(await fs.exists(path.resolve(updatesFolder, packageJson.version))).toBe(true);
     const version = await child.exec('node bin/yarn.js -V');
-    expect(version[0].trim(), '0.14.1');
+    expect(version[0].trim()).toBe(packageJson.version);
   });
 });
 
@@ -78,34 +79,36 @@ it('Self-update should download a release and symlink it as "current"', (): Prom
 it('Self-update should work from self-updated location', (): Promise<void> => {
   return run(async (reporter, config) => {
     // mock an existing self-update
-    await child.exec('npm run build');
-    await fs.copy(path.resolve(updatesFolder, '..'), path.resolve(updatesFolder, 'v0.99.0'));
-    await fs.symlink(path.resolve(updatesFolder, 'v0.99.0'), path.resolve(updatesFolder, 'current'));
+    await child.exec('yarn run build');
+    await fs.copy(path.resolve(updatesFolder, '..'), path.resolve(updatesFolder, '0.2.0'));
+    await fs.symlink(path.resolve(updatesFolder, '0.2.0'), path.resolve(updatesFolder, 'current'));
     let packageJson = await fs.readJson(path.resolve(updatesFolder, 'current', 'package.json'));
-    packageJson.version = '0.99.0';
+    packageJson.version = '0.2.0';
     await fs.writeFile(path.resolve(updatesFolder, 'current', 'package.json'),
       JSON.stringify(packageJson, null, 4));
     let version = await child.exec('node bin/yarn.js -V');
-    expect(version[0].trim(), '0.99.0');
+    expect(version[0].trim()).toBe('0.2.0');
 
     // mock a to_clean folder
-    packageJson.version = '0.98.0';
-    fs.mkdirp(path.resolve(updatesFolder, 'v0.98.0'));
-    await fs.symlink(path.resolve(updatesFolder, 'v0.98.0'), path.resolve(updatesFolder, 'to_clean'));
-    await fs.writeFile(path.resolve(updatesFolder, 'v0.98.0', 'package.json'),
+    packageJson.version = '0.1.0';
+    fs.mkdirp(path.resolve(updatesFolder, '0.1.0'));
+    await fs.symlink(path.resolve(updatesFolder, '0.1.0'), path.resolve(updatesFolder, 'to_clean'));
+    await fs.writeFile(path.resolve(updatesFolder, '0.1.0', 'package.json'),
       JSON.stringify(packageJson, null, 4));
 
-    await child.exec('node bin/yarn.js self-update v0.15.0');
+    await child.exec('node bin/yarn.js self-update');
 
     // new version is current
     version = await child.exec('node bin/yarn.js -V');
-    expect(version[0].trim(), '0.15.0');
+    expect(await fs.exists(path.resolve(updatesFolder, 'current'))).toBe(true);
+    packageJson = await fs.readJson(path.resolve(updatesFolder, 'current', 'package.json'));
+    expect(version[0].trim()).toBe(packageJson.version);
 
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.98.0'))).toBe(false);
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.99.0'))).toBe(true);
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.15.0'))).toBe(true);
+    expect(await fs.exists(path.resolve(updatesFolder, '0.1.0'))).toBe(false);
+    expect(await fs.exists(path.resolve(updatesFolder, '0.2.0'))).toBe(true);
+    expect(await fs.exists(path.resolve(updatesFolder, packageJson.version))).toBe(true);
 
     packageJson = await fs.readJson(path.resolve(updatesFolder, 'to_clean', 'package.json'));
-    expect(packageJson.version).toBe('0.99.0');
+    expect(packageJson.version).toBe('0.2.0');
   });
 });
--- a/__tests__/commands/self-update.js
+++ b/__tests__/commands/self-update.js
@@ -78,7 +78,7 @@ xit('Self-update should download a release and symlink it as "current"', (): Pro
 });
 
 
-it('Self-update should work from self-updated location', (): Promise<void> => {
+xit('Self-update should work from self-updated location', (): Promise<void> => {
   return run(async (reporter, config) => {
     // mock an existing self-update
     await child.exec('npm run build');
--- a/__tests__/commands/self-update.js
+++ b/__tests__/commands/self-update.js
@@ -81,9 +81,10 @@ it('Self-update should download a release and symlink it as "current"', (): Prom
 it('Self-update should work from self-updated location', (): Promise<void> => {
   return run(async (reporter, config) => {
     // mock an existing self-update
-    await child.exec('yarn run build');
-    await fs.copy(path.resolve(updatesFolder, '..'), path.resolve(updatesFolder, '0.2.0'));
-    await fs.symlink(path.resolve(updatesFolder, '0.2.0'), path.resolve(updatesFolder, 'current'));
+    await child.exec('npm run build');
+    const versionFolder = path.resolve(updatesFolder, '0.2.0');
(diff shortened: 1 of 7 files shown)


danez:
This uses now a patched version of roadrunner see kittens/roadrunner#1 and resets roadrunner correctly
The only bad thing is that I add the roadrunner cache file to constants, but I can't use it in the bin as it is not yet sure which version of the constants to load: modern/legacy.

bestander:
Nice!
Should we wait for roadrunner fix to be propagated to upstream?

danez:
Yes I think so @kittens, can you have a look at kittens/roadrunner#1 ?

src/constants.js
```diff
@@ -38,6 +38,7 @@ export function getModuleCacheDirectory(): string {
 export const MODULE_CACHE_DIRECTORY = getModuleCacheDirectory();
 export const LINK_REGISTRY_DIRECTORY = `${MODULE_CACHE_DIRECTORY}/.link`;
 export const GLOBAL_MODULE_DIRECTORY = `${MODULE_CACHE_DIRECTORY}/.global`;
+export const CACHE_FILENAME = path.join(userHome, '.yarn', '.roadrunner.json');
```
danez: I would like to use this variable in `bin/yarn.js`. But there i cannot simple import this file right? currently this is duplicate here and in `bin/yarn.js`

__tests__/commands/self-update.js
```diff
@@ -59,32 +62,31 @@ function run(checks: (reporter: reporters.Reporter, config: Config) => Promise<v
...
   });
 });
 
 
-xit('Self-update should work from self-updated location', (): Promise<void> => {
+it('Self-update should work from self-updated location', (): Promise<void> => {
   return run(async (reporter, config) => {
     // mock an existing self-update
-    await child.exec('make build');
+    await child.exec('npm run build');
```
danez: It is probably safe to assume that npm is available when running tests?

__tests__/commands/self-update.js
```diff
@@ -59,32 +62,31 @@ function run(checks: (reporter: reporters.Reporter, config: Config) => Promise<v
...
+    await selfUpdate(config, reporter, {}, ['v0.14.1']);
     expect(await fs.exists(path.resolve(updatesFolder, 'current')));
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.11.0')));
+    expect(await fs.exists(path.resolve(updatesFolder, 'v0.14.1')));
     const packageJson = await fs.readJson(path.resolve(updatesFolder, 'current', 'package.json'));
-    expect(packageJson.version === '0.11.0');
+    expect(packageJson.version === '0.14.1');
     const version = await child.exec('node bin/yarn.js -V');
-    expect(version[0].trim(), `0.11.0`);
+    expect(version[0].trim(), '0.14.1');
```
danez: Had to update the version as <0.14 was using the name `kpm` in the release

__tests__/commands/self-update.js
```diff
@@ -93,15 +95,15 @@ xit('Self-update should work from self-updated location', (): Promise<void> => {
...
 
     // new version is current
     version = await child.exec('node bin/yarn.js -V');
-    expect(version[0].trim(), `0.11.0`);
+    expect(version[0].trim(), '0.15.0');
 
     expect(await fs.exists(path.resolve(updatesFolder, 'v0.98.0'))).toBe(false);
     expect(await fs.exists(path.resolve(updatesFolder, 'v0.99.0'))).toBe(true);
-    expect(await fs.exists(path.resolve(updatesFolder, 'v0.11.0'))).toBe(true);
+    expect(await fs.exists(path.resolve(updatesFolder, 'v0.15.0'))).toBe(true);
```
danez: Here is use 0.15 to check it picks the correct asset.

__tests__/commands/self-update.js
bestander: Can we do "kpm run build"? :)
For now, yeah, we can assume it is safe

src/constants.js
bestander: .yarn is prone to change afaik.
There should be a constant with the folder name

__tests__/commands/self-update.js
bestander: Oh yeah, kpm has been engraved in my brain now :)

src/constants.js
danez: I can extract that to an other constant, but then still it wouldn't be usable in `bin/yarn.js` I think.

Daniel15:
I think we can remove all the Github API stuff from this. Instead of using the Github API, you can hit https://yarnpkg.com/latest-version to get the version number. Then you can simply hit https://yarnpkg.com/latest.tar.gz to get the latest tarball. Much simpler 😄  . There's also `https://yarnpkg.com/downloads/{version}/{file}` for specific version.

bestander:
Good point, @Daniel15, dropping Github wrapper from dependencies is good for size and startup performance.
Limiting self-update to `latest` is a limitation but I guess reasonable.

bestander:
A good RFC discussion is how to enable self-update to be automatic.

Daniel15:
> A good RFC discussion is how to enable self-update to be automatic.

For what it's worth, updates are handled along with the rest of the operating system if you use the Linux package (Ubuntu/Debian/CentOS), that's one major advantage of using them and one of the main reasons we built them :)

bestander:
Well, I mean the other 99.5 percent of users who don't use Linux :)

Daniel15:
@bestander  Chocolatey handles updates too I believe, so Windows users are covered too once our Chocolatey package is approved. The Windows dev userbase is larger than the Mac OS dev userbase 😛 . The Windows installer probably needs a proper auto-update too, that downloads the latest MSI and installs it.

bestander:
You don't expect every version of windows to install yarn with chocolatey?
But bikeshedding aside, I think we need to decide whether we want to keep this feature here or ditch it. If we keep it then let's offer users whether to enable it by default.

Some history: I added it here when we found that maintaining packages via yummy and Chef was too much hassle and it seemed like a reasonable idea to have a lightweight mechanism to get everyone on the same version within the org.
In the end we ended up checking in yarn.js instead of relying on Chef or building our own Chef  yarn self-updates.

As for the community project, I think we get quite a lot of benefits if everyone/most is getting updated to the latest stable whether it is done via package managers or via self-update.

bestander:
Also I am not a huge fan of this implementation, a bit too hacky

danez:
For macOS probably adding yarn to brew might probably also make it easy to install and update there.

How do we want to proceed with this PR? Should I do all the mentioned changes (not github etc.) and later you can still decide to remove/simplify it?

bestander:
@danez, let's do as @Daniel15 suggests and use the website API to get latest version.
Are you up to it?
We can bikeshed about whether we need this feature or enable it later.

Daniel15:
> Should I do all the mentioned changes (not github etc.) and later you can still decide to remove/simplify it?

Yeah, that sounds good to me! Sorry I went a little off-topic in the comments. Removing all the Github stuff should simplify it quite a bit :)

ithinkihaveacat:
@Daniel15 I don't know the implications of this but right now https://yarnpkg.com/latest-version does not list the latest version:

```sh
$ yarn -V
0.15.1
$ curl -s https://yarnpkg.com/latest-version
0.15.0
```

bestander:
Michael, could you file a bug?

danez:
This is beacuse there was no release created on github, only the tag in git

ithinkihaveacat:
@bestander Sure, reported as https://github.com/yarnpkg/yarn/issues/989.

Daniel15:
0.15.1 is not a "real" release, it was just a patch to the tarball and npm
package. All the other installation methods are still at 0.15.0

package.json
```diff
@@ -33,7 +32,7 @@
     "request": "^2.75.0",
     "request-capture-har": "^1.1.4",
     "rimraf": "^2.5.0",
-    "roadrunner": "^1.0.6",
+    "roadrunner": "danez/roadrunner#patch-1",
```
Daniel15: Should we wait for a release of roadrunner before merging this?

yarn.lock
```diff
@@ -2237,14 +2223,6 @@ getpass@^0.1.1:
   dependencies:
     assert-plus "^1.0.0"
 
-github@2.5.1:
```
Daniel15: Nice 👍 🎉

danez:
The test are still failing, because of 0.15.1.

The problem is that https://yarnpkg.com/latest-version says to 0.15.0, but when downloading latest.tar.gz the package.json in this archive says 0.15.1. Im gonna try instead of relying on the package.json instead see where the current symlink points.

Daniel15:
It's tricky because 0.15.1 is not a "real" release - It was just a fixed re-release of the npm package and tarball, but everything else (Windows installer, Linux packages) are still on 0.15.0

GantMan:
looking forward to this 👍

bin/yarn.js
```diff
@@ -5,14 +5,12 @@
 'use strict';
 
 // init roadrunner
-var userHome = require('user-home');
 var mkdirp = require('mkdirp');
-var path = require('path');
-var CACHE_FILENAME = path.join(userHome, '.yarn', '.roadrunner.json');
-mkdirp.sync(path.dirname(CACHE_FILENAME));
+var constants = require('../lib-legacy/constants');
```
danez: I now also require constants here from lib-legacy, so that I can use the folder constants. I think that shouldn't be a significant problem.

yarn.lock
```diff
@@ -4449,9 +4415,9 @@ ripemd160@^1.0.0:
   version "1.0.1"
   resolved "https://registry.yarnpkg.com/ripemd160/-/ripemd160-1.0.1.tgz#93a4bbd4942bc574b69a8fa57c71de10ecca7d6e"
 
-roadrunner@^1.0.6:
+roadrunner@danez/roadrunner#patch-1:
   version "1.0.6"
-  resolved "https://registry.yarnpkg.com/roadrunner/-/roadrunner-1.0.6.tgz#e466a15f9844289413d09943ad263574623ec1c9"
+  resolved "https://codeload.github.com/danez/roadrunner/tar.gz/2ead9501e3681fa7584b79e6e6546fb046a83f54"
```
bestander: Let's push the change upstream before merging this.

danez:
While waiting for roadrunner I noticed that it is possible to download a certain version of yarn see https://github.com/yarnpkg/yarn/issues/751#issuecomment-253058942 
Should I add back the possibility to specify the version?

Daniel15:
> Should I add back the possibility to specify the version?

Let's keep it simple - It's totally fine to only support the most recent version 😄

thejameskyle:
@danez you have a broken test: 

```
 FAIL  __tests__/commands/self-update.js (10.597s)
  ● Self-update should work from self-updated location

    Error: Command failed: yarn run build
    /bin/sh: 1: yarn: not found
     
    Console output:
      
       
      at __tests__/commands/self-update.js:57:13
      at throw (native)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:30:20

  ✓ Self-update should download a release and symlink it as "current" (10365ms)
  ✕ Self-update should work from self-updated location (73ms)
```

danez:
CircleCI is not updating the dependencies :(

bestander:
I'll restart with cache clean

danez:
Seems there is no step on circleCI to install dependencies? At least I can't see one.

bestander:
Yeah, odd, looks like cleaning all cache lost eslint and webpack.

bestander:
Created this one https://github.com/yarnpkg/yarn/pull/1157

bestander:
@danez, could you please rebase it?
This should work now

danez:
Seems circleCI is to slow to run the tests :/ The test is anyway really ugly, and involves building yarn, so not sure if we should keep it.

bestander:
Hm, I would assume that download gets stuck.
Building yarn should be quite fast though.
This is an important test because it checks that self-update did not degrade because of the symlinks and yarn inside yarn situation.

What do you think of keeping this test disabled for a while and merge the rest?
Great work anyway!

danez:
Okay I disabled it now, even after 3 minutes it was running into a timeout.

As all the other downloads seem to be quiet fast it might also be the FS which is slow, and we copy a lot of stuff in this test.

bestander:
Thanks, @danez!

PR #690, opened 2016-10-11
Handle installation for 'fish' shell

hasit:
**Summary**

This PR adds support for handling installation of `yarn` on [fish shell](https://fishshell.com).

fish is a popular shell. Me and many I know use it regularly for its awesome auto-completion features and convenience.

The current problem is that the `yarn` install script only checks for bash and zsh shells. This was causing `~/.yarn/bin` to not be added to the `$PATH` in fish.

With this commit, the `yarn` install script first checks if the user's shell is fish. If so, it adds the following line to `~/.config/fish/config.fish`.

```fish
set -xg PATH ~/.yarn/bin $PATH 
``` 

See [Startup (Where's .bashrc?)](https://fishshell.com/docs/current/tutorial.html#tut_startup) section in official fish documentation for more information.

**Test plan**

When `yarn/scripts/install_latest.sh` is run, following happens:

```bash
bash-3.2$ ./install-latest.sh
Installing Yarn!
> Downloading tarball...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    92  100    92    0     0    369      0 --:--:-- --:--:-- --:--:--   368
100   595    0   595    0     0    697      0 --:--:-- --:--:-- --:--:--   697
100 3202k  100 3202k    0     0   566k      0  0:00:05  0:00:05 --:--:--  741k
> Extracting to ~/.yarn...
> Adding to $PATH...
> We've added the following to your /Users/hasitmistry/.config/fish/config.fish
> If this isn't the profile of your current shell then please add the following to your correct profile:

export PATH="$HOME/.yarn/bin:$PATH"

> Successfully installed! Please open another terminal where the `yarn` command will now be available.
```

After successful installation, my `~/.config/fish/config.fish` looks like this:

```fish
# Go setup
set -xg GOPATH $HOME/go
set -xg GOROOT /usr/local/go
set -xg PATH $GOPATH/bin $PATH
set -xg GO15VENDOREXPERIMENT 1

# Editor
set -xg EDITOR atom

set -xg PATH ~/.yarn/bin $PATH
```

### DIFF
--- a/scripts/install-latest.sh
+++ b/scripts/install-latest.sh
@@ -21,7 +21,6 @@ yarn_link() {
   printf "$cyan> Adding to \$PATH...$reset\n"
   YARN_PROFILE="$(yarn_detect_profile)"
   SOURCE_STR="\nexport PATH=\"\$HOME/.yarn/bin:\$PATH\"\n"
-  FISH_SOURCE_STR="fish -c 'set -U fish_user_paths \$fish_user_paths ~/.yarn/bin'"
 
   if [ -z "${YARN_PROFILE-}" ] ; then
     printf "$red> Profile not found. Tried ${YARN_PROFILE} (as defined in \$PROFILE), ~/.bashrc, ~/.bash_profile, ~/.zshrc, and ~/.profile.\n"
@@ -33,7 +32,7 @@ yarn_link() {
   else
     if ! grep -q 'yarn' "$YARN_PROFILE"; then
       if [[ $YARN_PROFILE == *"fish"* ]]; then
-        eval $FISH_SOURCE_STR
+        command fish -c 'set -U fish_user_paths $fish_user_paths ~/.yarn/bin'
       else
         command printf "$SOURCE_STR" >> "$YARN_PROFILE"
       fi
--- a/scripts/install-latest.sh
+++ b/scripts/install-latest.sh
@@ -21,7 +21,7 @@ yarn_link() {
   printf "$cyan> Adding to \$PATH...$reset\n"
   YARN_PROFILE="$(yarn_detect_profile)"
   SOURCE_STR="\nexport PATH=\"\$HOME/.yarn/bin:\$PATH\"\n"
-  FISH_SOURCE_STR="\nset -xg PATH ~/.yarn/bin \$PATH"
+  FISH_SOURCE_STR="fish -c 'set -U fish_user_paths \$fish_user_paths ~/.yarn/bin'"
 
   if [ -z "${YARN_PROFILE-}" ] ; then
     printf "$red> Profile not found. Tried ${YARN_PROFILE} (as defined in \$PROFILE), ~/.bashrc, ~/.bash_profile, ~/.zshrc, and ~/.profile.\n"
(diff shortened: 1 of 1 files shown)


hildjj:
Manipulating `fish_user_paths` seems like a more modern approach.  

```
set -U fish_user_paths ~/.yarn/bin (~/.yarn/bin/yarn global bin) $fish_user_paths
```

See https://fishshell.com/docs/current/tutorial.html#tut_path

hasit:
@hildjj 

Yes and no.

While it is supported, I have come across many problems with the `set -U ...` approach. See [fish-shell#527](https://github.com/fish-shell/fish-shell/issues/527).

`set -xg ...` approach still remains as the safe way to set `$PATH` persistently.

maxnordlund:
Trying to use the PATH as an universal variable isn't supported by fish. However, setting the fish_user_paths works just fine. Note that this should only be run _once_ on installation since universal variables are saved across sessions. This is also why you shouldn't have setup in config.fish if you can avoid it. I use fish_user_paths, and have for a few years now, and haven't had any problems, but I guess your milage may vary.

hasit:
@maxnordlund @hildjj makes sense. I will update my commit.

hasit:
As @ridiculousfish [suggests](https://github.com/fish-shell/fish-shell/issues/527#issuecomment-13811988), the `set -U fish_user_paths $fish_user_paths ~/.yarn/bin` approach is a good one for adding `yarn`'s path globally and persistently. So, @hildjj and @maxnordlund, I agree with you points.

The problem I see here is that the install script is run in bash. If we run the command `set -U ...`, bash would not recognize it. We will have to set it in `~/config/fish/config.fish` file.

Thoughts?

maxnordlund:
Maybe `fish -c 'set -U ...'` would do the trick. I.e. run it inside fish.

hildjj:
Note that adding `yarn global bin` to the path is either waiting on #728 or some sed work.

hasit:
@maxnordlund, @hildjj check the new commit. Let me know what you think.

maxnordlund:
I like it, but see my comment above.

scripts/install-latest.sh
```diff
@@ -31,7 +32,11 @@ yarn_link() {
     command printf "${SOURCE_STR}"
   else
     if ! grep -q 'yarn' "$YARN_PROFILE"; then
-      command printf "$SOURCE_STR" >> "$YARN_PROFILE"
+      if [[ $YARN_PROFILE == *"fish"* ]]; then
+        eval $FISH_SOURCE_STR
```
hasit: Makes sense. Saves the space for storing a command string as well.

hasit:
@maxnordlund, done and works. Thank you 👍

maxnordlund:
Kein problem, just happy to help 😸

maxnordlund:
I made #766 for fixing the escape codes, after that the `yarn global bin` should work.

hasit:
@maxnordlund looks good. We need to hear from the maintainers on this one. Once your PR gets through, I can add the (~/.yarn/bin/yarn global bin) part too.

cpojer:
thank you! I think this is reasonable to support.

cc @kittens

hasit:
@cpojer Yay! Thank you.

PR #669, opened 2016-10-11
Remove unnecessary deps

chicoxyzzy:
React presets and plugins for Babel and ESLint are unnecessary cause React is not used in the project

PR #701, opened 2016-10-11
Add back eslint-plugin-react since it is a peerDep of eslint-config-fbjs

hzoo:
**Summary**

Ref https://github.com/yarnpkg/yarn/pull/669#issuecomment-253009368

```
warning Unmet peer dependency "eslint-plugin-react@^5.2.2".
```

npm run test will fail on eslint because there is a missing plugin dependency in https://github.com/facebook/fbjs/blob/446badb82d7774a9d23df70b51cb04abb97e11bc/packages/eslint-config-fbjs/package.json#L24.

```
Oops! Something went wrong! :(

ESLint couldn't find the plugin "eslint-plugin-react". This can happen for a couple different reasons:
```

**Test plan**

`npm t`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "debug": "^2.2.0",
     "defaults": "^1.0.3",
     "diff": "^2.2.1",
+    "eslint-plugin-react": "5.2.2",
     "github": "2.5.1",
     "ini": "^1.3.4",
     "invariant": "^2.2.0",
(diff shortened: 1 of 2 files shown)


addyosmani:
cc @Daniel15 Change LGTM but would like a sanity check :)

Daniel15:
:shipit:

hzoo:
I am confused on the other pkg since I don't see anything in the config that removes flow types (it was the react preset) and the tests are all still passing?

PR #702, opened 2016-10-11
CONTRIBUTING: add setup instructions [skip ci]

hzoo:
Adding setup instructions similar to what we have in babel https://github.com/babel/babel/blob/master/CONTRIBUTING.md#setup

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 
 ## Setup
 
-> First check that `yarn` is installed globally with `npm install yarn --global`
+> Install yarn on your system: https://yarnpkg.com/en/docs/install
 
 ```sh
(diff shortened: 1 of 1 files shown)


Daniel15:
> First check that `yarn` is installed globally with `npm install yarn --global`

It seems pretty confusing to have Yarn installed globally when working on Yarn itself :)

Additionally the npm module is just supposed to be a fallback - Ideally people should be using the Windows installer or Linux packages (or Homebrew once we have that) to install a release build of Yarn

hzoo:
Ah ok wasn't sure

hzoo:
Ok I'l just link to the install docs?

zetoke:
>`npm install yarn --global`

yarnpkg, not yarn

hzoo:
it's actually `yarn` now (either way I removed that and used the link instead)

zetoke:
@hzoo Oh, I see. Thanks for information!

PR #703, opened 2016-10-11
Revert "Remove unnecessary deps"

Daniel15:
Reverts yarnpkg/yarn#669

Seems like something still requires this.


Daniel15:
@hzoo got to it first: #701

PR #707, opened 2016-10-11
fix base-fetcher: use proper util to stringify, clean up imports

vladikoff:
**Summary**

Clean up and util function utilization

**Test plan**

Change is already tested by existing base-fetcher coverage


cpojer:
thank you!

PR #716, opened 2016-10-11
use first non-flag argument as command

### LINKED ISSUE #635: Adding unused option before `add` command yields error

A bug. 

Adding unused option before `add` command yields error

```bash
$ yarn --foo add gulp-asdasdasdasdasdasdasdasdasd
yarn install v0.15.1
error `install` has been replaced with `add` to add new dependencies. Run "yarn add gulp-asdasdasdasdasdasdasdasdasd" instead.
info Visit http://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

```bash
$ yarn --foo add gulp-asdasdasdasdasdasdasdasdasd
yarn install v0.15.1
warning `--foo` is an unknown option for `add` and will be ignored.
info Visit http://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

```bash
- node.js: v6.5.0
- yarn: v0.15.1
```

edge:
**Summary**

yarn does not parse the arguments correctly if a flag is passed before the first command. The patch makes the cli use the first non-flag argument instead of unconditionally using the first argument.

Also partially fixes #635.

**Test plan**

Previously, yarn would default to `install` if a flag was passed first. In the case
`yarn --foo add colors`, the cli would call the `install` command with a nonzero arguments array, triggering a deprecation error. Now, it correctly passes `--foo` as a flag to the `add` command.

Because `--foo` is not a valid flag, this should trigger a warning/error. I added a message to the reporter for when this is implemented.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -89,8 +89,8 @@ if (commandName === 'help') {
   }
 }
 
-// if no args or command name looks like a flag then default to `install`
-if (!commandName || commandName[0] === '-') {
+// if no args then default to `install`
+if (!commandName) {
   if (commandName) {
     args.unshift(commandName);
   }
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -59,9 +59,20 @@ commander.option(
 );
 commander.allowUnknownOption();
 
-// get command name
-let commandName: string = args.shift() || '';
+// get command name, defaulting to `install`
+let commandName: string = 'install';
+
+for (let i = 0; i < args.length; i++) {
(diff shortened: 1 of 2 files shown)


kittens:
I don't think we should support this since it's ambiguous and makes it impossible for `yarn --mutex etc` to work since we'll take `etc` to be the command rather than the flags value.

kittens:
I'm going to close this for the aforementioned reason. The ambiguity and potential risks outweigh the slight convenience gained from this. Thanks for the pull request @edge!

PR #718, opened 2016-10-11
Don't remove compiled files in `yarn clean`

bouk:
**Summary**

Some packages (like 'timepicker') use minified files as the main entrypoint. This makes them incompatible with `yarn clean`.

I want to use a custom `.yarnclean` while still being able to use this package, and it seems there is no way to do this without modifying yarn itself.


cpojer:
I think we should make it so that `.yarnclean` has a default entry like `# extend yarn-default` on top. If it's present, it uses yarn's default clean mechanism, if not, it only uses files from the `.yarnclean` file. Would you mind changing your PR to that and writing a test as well?

PR #725, opened 2016-10-11
Hostname check inside isPossibleOfflineError()

pi0:
**Summary**
`hostname` is not always available on network `error` objects when checking that object inside `lib/util.js:isPossibleOfflineError()` function this will always `return false;` so **INSTALL BREAKS ON ANY TYPE OF ERROR** and doesn't attempts to retry and recover from that network error.

**UPDATE** PR Re-based  and used a cleaner solution for this problem :)


pi0:
@kittens would you please check this out ? :)

kittens:
Thanks! Sorry, trying to get around to these as quickly as I can.

PR #729, opened 2016-10-11
Set the correct filename for the npm package

koddsson:
This might fix the build?

### DIFF
--- a/scripts/build-dist.sh
+++ b/scripts/build-dist.sh
@@ -6,7 +6,7 @@ npm run build
 npm pack
 rm -rf dist
 mkdir dist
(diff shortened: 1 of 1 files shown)


scripts/build-dist.sh
```diff
@@ -6,7 +6,7 @@ npm run build
 npm pack
 rm -rf dist
 mkdir dist
-mv yarnpkg-*.tgz dist/pack.tgz
+mv yarn-*.tgz dist/pack.tgz
```
Daniel15: I don't think we'll change the package name again so this is probably okay.

PR #734, opened 2016-10-11
Fix Discord server ID in badge

jhgg:
**Summary**
This PR fixes the server ID in the discord badge. The server ID in it was pointing another server that was not yarnpkg's. Merging this PR also means going into server settings & enabling the Widget so that it functions properly.

_general

**Test plan**
N/A


cpojer:
thank you! Also enabled the widget.

Issue #631, opened 2016-10-11
SSL handshake failure


iamdanfox:
When running `yarn` at work, I hit the following error:

```
yarn install v0.15.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
error write EPROTO 140735253005056:error:140790E5:SSL routines:ssl23_write:ssl handshake failure:../deps/openssl/openssl/ssl/s23_lib.c:177:

    at exports._errnoException (util.js:1036:11)
    at WriteWrap.afterWrite (net.js:794:14)
info Visit http://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

If it's at all relevant, we have an internal npm mirror (defined in the `.npmrc` file).  One other guess is that this is failing due to internal TLS-decryption.

```
registry = https://<someinternalurl>/artifactory/api/npm/all-npm/
```

yarn -V: 0.15.1
Mac 10.10.5 (14F1912)
node v6.7.0

Daniel15:
Do you have a legitimate SSL cert on your internal server, or is it a self-signed cert?

Can you `wget` a file from that server?

iamstarkov:
@Daniel15 if its self-signed cert, what should i do?

iamstarkov:
@Daniel15 if its self-signed cert, what should i do?

BagelTheKid:
I'm receiving the same issue when trying to install any package. 
I'm behind a corporate proxy that has a self-signed cert. I've set strict-ssl to false in my npm settings, I'm unsure if that will affect Yarn's config.

Daniel15:
@iamstarkov and @BagelTheKid - One option is to set the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable to `0`. This disables TLS checking for **all** calls performed from the Node.js process.  On Mac OS and Linux, you can run Yarn like this:
```
NODE_TLS_REJECT_UNAUTHORIZED=0 yarn install
```


However, for a corporate proxy, I'd strongly suggest for your sysadmin to use a real certificate. They could use an internal CA and have its root cert preinstalled on computers that are on the corporate network (this is what we do for internal servers at Facebook).

kittens:
We've merged support for `cafile` and `strict-ssl` so this issue should be resolved.

Issue #606, opened 2016-10-11
Yarn doesn't read from .npmrc Files for custom registry settings


raybooysen:
With an .npmrc file like below:

```
registry=http://local-npm-install:8080
```

It seems yarn will not respect this and try the normal NPM upstream.

tyrsius:
We use custom registries for a lot of private modules, this is must-have functionality.

maniator:
Cannot use yarn until this issue is resolved.

nkabardin:
I wonder, so Facebook doesn't have internal registry for private stuff? Or it's just not yet implemented for open source?

jkrems:
Even worse - it uses the yarn registry config when asked to fetch from npm:

```diff
 class NpmRegistry extends (_baseRegistry || _load_baseRegistry()).default {
   /* ... */

   request(pathname) {
     // ...
-    const registry = (0, (_misc || _load_misc()).removeSuffix)(String(this.registries.yarn.getOption('registry')), '/');
+    const registry = (0, (_misc || _load_misc()).removeSuffix)(String(this.registries.npm.getOption('registry')), '/');
```

thejameskyle:
Hey everyone, please use the upvote button on GitHub so we can keep this issue from turning into 1000 comments

jkrems:
Workaround - put the following into a `.yarnrc` file in the project root:

```
registry "https://my-registry.corp"
```

tyrsius:
@jkrems What is the syntax for scoped registries?

```
@scope:registry "http://place.com"
```

throws an `unknown token` error

jkrems:
@tyrsius It's a custom file format (`src/lockfile/parse.js`). I reverse engineered just enough to make my example above work. ;)

andypham:
Would be awesome if yarn can address this issue as I work with internal registry daily.

alexgorbatchev:
possible list of supported `.yarnrc` defaults is here https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/registries/yarn-registry.js#L16

alexgorbatchev:
Anyone figured out yet how to configure `ca` values?

tomccabe:
Complete blocker to using this tool. Changed the registry in yarn config (because "drop in replacement for npm"!! 🤔), `yarn login`'d, no dice.

ljharb:
[...]

PR #736, opened 2016-10-11
Add support for config file option 'cafile'

chlunde:
**Summary**

With a private registry using a company CA issued certificate, yarn fails because nodejs does not use the OS SSL/TLS trust store (at least on Linux).  The same issue affects registries using self signed certificates and might also apply for the public registries in companies using SSL/TLS intercepting proxies.

This is change adds support for the `cafile`-option, which should point to a file path containing one or more PEM files concatenated together, like in `npm`.

I've chosen `cafile` and not the `ca` option, because it is simpler to point to the operating system TLS trust store, which often will already have the correct CA certificates for corporate computers.

This does not add support for disabling strict-ssl, because that will open for MITM attacks.

References #606, #631

**Test plan**

***Integration test***
I've added a test which sets up a web server with a self signed certificate and a trust store containing a normal CA and the self signed certificate, and verifies that a connection set up using RequestManager can perform a GET-request from that server.

***Manual test***
With default config an no cafile set, I get following stack trace:
```
~/yarn/bin/yarn install
yarn install v0.15.1
[1/4] Resolving packages...
[2/4] Fetching packages...
error unable to verify the first certificate
    at Error (native)
    at TLSSocket.<anonymous> (_tls_wrap.js:1017:38)
    at emitNone (events.js:67:13)
    at TLSSocket.emit (events.js:166:7)
    at TLSSocket._init.ssl.onclienthello.ssl.oncertcb.TLSSocket._finishInit (_tls_wrap.js:582:8)
    at TLSWrap.ssl.onclienthello.ssl.oncertcb.ssl.onnewsession.ssl.onhandshakedone (_tls_wrap.js:424:38)
```

With `cafile=/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in `.npmrc` or `cafile /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in `.yarnrc` installation succeeds


cpojer:
@chlunde Thanks! Do you mind adding an integration test for this feature?

sindresorhus:
Wouldn't it be better to make Yarn use the OS TLS trust store instead of adding an option?

donaldpipowitch:
I'm not sure if this the same error, but... on our custom registry we need to set I also get `error unable to verify the first certificate` on macOS.`.npmrc` supports a config `strict-ssl=false` in this case.

chlunde:
@sindresorhus Having nodejs, not Yarn, use the OS trust store by default would be a good thing in the long run.  There's some information related to that issue here: https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/

But I think a change like this would be required anyway if someone uses a self-signed certificate or cannot add the registry CA to the default trust store for some other reason.

sindresorhus:
There might not be any way around this, true. Just want to recommend being careful about adding options. It's usually better if it *just works*. Now users must know to set the `cafile` option instead of it just relying on the OS trust store like other apps. And you don't want Yarn to end up with [almost 100 options](https://docs.npmjs.com/misc/config), like npm.

src/util/request-manager.js
```diff
@@ -118,6 +123,12 @@ export default class RequestManager {
     if (opts.httpsProxy != null) {
       this.httpsProxy = opts.httpsProxy;
     }
+
+    if (opts.cafile != null && opts.cafile != '') {
```
kittens: This is awesome! Would you mind adding a short comment here explaining what this does and how it works?

chlunde:
@kittens @cpojer Thanks for the comments, I've updated the PR with integration tests.  I added support for comments between the certificates (like used by https://curl.haxx.se/docs/caextract.html), and no longer throw an exception if the file is missing, because yarn config set cafile "correctpath" was not working when the current cafile was missing.

Issues:
* tests broken since https://github.com/yarnpkg/yarn/pull/880#issuecomment-253274383
* is it OK to hardcode port 3443 for the tests?
* is there a better way to prevent caching the test file?  I added nocache as a query parameter and look for that in the fixture request handler
* importing http-proxy fails, is this an issue with my code or jest-resolve?

```
import httpProxy from 'http-proxy';
console.dir(httpProxy);
```
fails with
```
    TypeError: Cannot set property 'createProxy' of undefined

      at Object.<anonymous> (node_modules/http-proxy/lib/http-proxy.js:28:30)
      at Object.<anonymous> (node_modules/http-proxy/index.js:13:18)
      at _load_httpProxy (__tests__/fetchers.js:27:1636)
      at Object.<anonymous> (__tests__/fetchers.js:29:55)
```

chlunde:
I've updated the PR to address my own comments, except for the "nocache" query parameter in the mock which I've kept.  The branch `cafile-old` in my repo contains the original integration test.

PR #748, opened 2016-10-11
shorthands execute commands

lopno:
**Summary**
Allow support for shorthands to actually execute commands.
`yarn i` will now execute `install` instead of giving an error.
This is the same behavior as npm where `npm install` and `npm i` are equivalent.
I really like that feature of npm and I would like to have it in yarn.

**Test plan**
screenshot from 2016-10-11 23-32-29

### DIFF
--- a/src/cli/aliases.js
+++ b/src/cli/aliases.js
@@ -1,20 +1,6 @@
 /* @flow */
 
 export default {
-  // shorthands
-  'run-script': 'run',
-  c: 'config',
-  i: 'install',
-  list: 'ls',
-  ln: 'link',
-  rb: 'rebuild',
-  runScript: 'run',
-  t: 'test',
-  tst: 'test',
-  un: 'remove',
-  up: 'update',
-  v: 'info',
-
   // affordances
   'add-user': 'login',
   'dist-tag': 'tag',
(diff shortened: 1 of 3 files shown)


btecu:
But why? Isn't `yarn` the same as `npm install`?

ljharb:
-1, shorthands are a horrible idea in npm, and make googling for things and finding documentation harder.

Explicit errors are a supremely better way to teach newcomers.

edge:
I think shorthands are still useful. They might not be the best thing to use in a beginner tutorial, but I do save a lot of time day to day using `npm i -D` instead of `npm install --save-dependencies`.

Sinewyk:
@edge you still didn't explain why you need a shorthand when simply `yarn` is equivalent to `npm i(nstall)`.
Why would you want "shorthand" versus just "no command".

edge:
I was talking about shorthands in general, and defaulting is not the general case. Not typing a command inherently gives you fewer choices i.e. less power.

cpojer:
Thank you for the pull request and the discussion. We made a conscious effort not to support them in yarn and `yarn` itself is a shorthand for `yarn install` so `yarn i` seems completely superfluous.

PR #752, opened 2016-10-11
Flag to ignore platform compatibility

mdarse:
**Summary**

Hi everyone,

I had an issue with some packages with _not quite correct_ platform properties in `package.json`, blocking completely an install process that was working well with _npm_.

So the output was:
```
❯ yarn
yarn install v0.15.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
error babel-plugin-react-require@2.1.0: The platform "darwin" is incompatible with this module.
error babel-plugin-react-require@2.1.0: The CPU architecture "x64" is incompatible with this module.
info Visit http://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
This is caused by packages having the `os` & `cpu` properties set to the empty array value:
```json
{
  "name": "foo",
  "version": "1.0.0",
  "os": [],
  "cpu": []
}
```

This PR adds a new `--ignore-platform` flag to permit the installation (as a temporary workaround while the change lands upstream). I think this could also helps when packaging some project for another platform type.

I not quite sure about the semantics of `config` vs `flags` in a command, so the implementation may not be like it should be. Some insights would be welcome on this.

**Test plan**

```
❯ yarn --ignore-platform
yarn install v0.15.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
success Saved lockfile.
✨  Done in 30.00s.
```

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -59,6 +59,7 @@ type IntegrityMatch = {
 
 type Flags = {
   // install
+  ignorePlatform: boolean,
   ignoreEngines: boolean,
   ignoreScripts: boolean,
   ignoreOptional: boolean,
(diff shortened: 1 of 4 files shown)


Alkhatri77:
My wallet

PR #758, opened 2016-10-11
Remove outdated kpm parts

chicoxyzzy:
**Summary**

According to comments and #745 kpm is an old name for yarn 

**Test plan**

N/A

### DIFF
--- a/src/cli/commands/login.js
+++ b/src/cli/commands/login.js
@@ -45,7 +45,7 @@ export async function getToken(config: Config, reporter: Reporter, name: string 
     };
   }
 
-  const env = process.env.YARN_AUTH_TOKEN || process.env.KPM_AUTH_TOKEN || process.env.NPM_AUTH_TOKEN;
+  const env = process.env.YARN_AUTH_TOKEN || process.env.NPM_AUTH_TOKEN;
   if (env) {
     config.registries.npm.setToken(`Bearer ${env}`);
     return function revoke(): Promise<void> {
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -120,12 +120,7 @@ export default class NpmRegistry extends Registry {
       const config = ini.parse(file);
 
       // normalize offline mirror path relative to the current npmrc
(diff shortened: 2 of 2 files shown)


src/cli/commands/login.js
```diff
@@ -36,7 +36,7 @@ async function getCredentials(config: Config, reporter: Reporter): Promise<?{
 export async function getToken(config: Config, reporter: Reporter): Promise<
   () => Promise<void>
 > {
-  const env = process.env.YARN_AUTH_TOKEN || process.env.KPM_AUTH_TOKEN || process.env.NPM_AUTH_TOKEN;
+  const env = process.env.YARN_AUTH_TOKEN || process.env.NPM_AUTH_TOKEN;
```
chicoxyzzy: not sure if this is necessary at all

cpojer:
I'll let @bestander take care of this; it depends on the name migration at FB.

src/cli/commands/self-update.js
```diff
@@ -35,13 +35,6 @@ export async function run(
     timeout: 5000,
   });
 
-  // while yarn is close sourced we need an auth token to be passed
-  const githubAuth0Token = process.env.YARN_AUTH_TOKEN || process.env.KPM_AUTH_TOKEN;
-  github.authenticate({
-    type: 'oauth',
-    token: githubAuth0Token,
-  });
-
```
mourner: The GitHub authentication part is not only a KPM part, right (there's also `YARN_AUTH_TOKEN`)? Why was it removed?

bestander: This authentication was used for self-update feature only.
And it was necessary only when the repo was not public yet.

bestander: Could you undo changes to self-update.js?
This is being fixed in https://github.com/yarnpkg/yarn/pull/676

chicoxyzzy: @bestander done

bestander:
Thanks, @chicoxyzzy

Issue #728, opened 2016-10-11
No fancy terminal features when stdout not a tty


hildjj:
**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behavior?**
Color escape codes and progress bars are always output, regardless of whether stdout is a TTY.  

**If the current behavior is a bug, please provide the steps to reproduce.**
To reproduce easily, `yarn | more`.  Note the escape codes being printed literally.

**What is the expected behavior?**
When the output of yarn is redirected, do not use color escapes.  Likely, the output is going to be processed in some way.  Check process.stdout.isTTY() (see https://nodejs.org/api/tty.html#tty_tty for example), before outputting color, doing fancy spinners or progress bars.

**Please mention your node.js, yarn and operating system version.**
node.js: v6.7.0
yarn: 0.15.1
OS: OSX 10.12

victornoel:
+1 Also it would be great to have a command line option to force this behaviour :)

brentjohnson:
--no-color possibly?

sensedeep:
For CI, it would be great to have a really quiet option that emits trace only on failures (and without control codes).

yarn -q ......      # quiet mode

Generally you want CI output to be very easy to scan for errors and warnings. Control codes and verbose output make this much harder.

victornoel:
@sensedeep For the quiet mode, you can see #788 and for the command line option I think #1340 covers it :)

So this issue can be closed I think!

sensedeep:
Just tried that in 0.16.1. Seems  --silent is not released yet?

```
 yarn --silent silent

  error: unknown option `--silent'
```

victornoel:
@sensedeep #788 is not closed so no :)

martinbean:
The ability to both suppress progress logs and control codes would be great, as it’s making my [Viaduct](https://viaduct.io) deployment logs look _hideous_:

<img width="972" alt="screen shot 2016-10-26 at 09 32 51" src="https://cloud.githubusercontent.com/assets/167312/19718890/561c6aa4-9b5f-11e6-8db4-5e34136bdfb2.png">

If this is to be a drop-in replacement for npm, then it should have the user options npm offer.

arcanis:
Should be fixed with `--no-progress` (#1190), please reopen if that's not enough.

jrr:
This seems to still be a problem:
```
jrr@jrrmbp ~> yarn versions |more
yarn versions v1.9.4
[...]

PR #766, opened 2016-10-11
Only write escape sequences to a TTY

maxnordlund:
**Summary**

This solves #728 which is to make yarn a nice unix citizen by not emitting escape sequences outside TTYs.

**Test plan**

The only thing I noticed was that flow doesn't seem to support `process.stdout.isTTY`, see https://github.com/facebook/flow/issues/1825. And since I don't really use flow, so I'm not sure how to fix this.

skarmavbild 2016-10-12 kl 01 31 02

### DIFF
--- a/src/reporters/console/util.js
+++ b/src/reporters/console/util.js
@@ -3,20 +3,33 @@
 import type {Stdout} from '../types.js';
 
 const readline = require('readline');
+const {supportsColor} = require('chalk');
 
 const CLEAR_WHOLE_LINE = 0;
 const CLEAR_RIGHT_OF_CURSOR = 1;
 
 export function clearLine(stdout: Stdout) {
+  if (!supportsColor) {
+    return;
+  }
+
   readline.clearLine(stdout, CLEAR_WHOLE_LINE);
   readline.cursorTo(stdout, 0);
 }
 
 export function toStartOfLine(stdout: Stdout) {
+  if (!supportsColor) {
+    return;
+  }
+
   readline.cursorTo(stdout, 0);
 }
 
 export function writeOnNthLine(stdout: Stdout, n: number, msg: string) {
+  if (!supportsColor) {
+    return;
+  }
+
   if (n == 0) {
     readline.cursorTo(stdout, 0);
     stdout.write(msg);
@@ -32,6 +45,10 @@ export function writeOnNthLine(stdout: Stdout, n: number, msg: string) {
 }
 
 export function clearNthLine(stdout: Stdout, n: number) {
+  if (!supportsColor) {
+    return;
+  }
+
   if (n == 0) {
     clearLine(stdout);
     return;


src/util/chalk.js
```diff
@@ -0,0 +1,7 @@
+/* @flow */
+
+const chalk = require('chalk');
+
+chalk.enabled = process.stdout.isTTY;
```
sindresorhus: Chalk already handles this: https://github.com/chalk/supports-color/blob/8400d98ade32b2adffd50902c06d9e725a5c6588/index.js#L29

hildjj:
It does look as if chalk is already doing the correct thing.  We should look at progress bars and spinners.  On Windows, I get this output from `yarn | more`:

```
[2K[1Gyarn install v0.15.1
[2K[1Gsuccess Already up-to-date.
[2K[1GDone in 0.17s.
```

These escapes look a lot like the tests for console-reporter with ProgressBar: https://github.com/yarnpkg/yarn/blob/master/__tests__/reporters/console-reporter.js#L134

src/util/chalk.js
maxnordlund: Kinda, this doesn't rely on the complex logic inside supports-color, but rather takes the simpler approach to just look at if it's a TTY.

Perhaps this is no good, and then I'll just remove this. Or change it to only overwrite if it _isn't_ a TTY.

maxnordlund:
Actually it's even more strange, on my Mac I get different results from `more` and `cat`:

skarmavbild 2016-10-12 kl 12 45 31

I've tried to understand why, but both isTTY and chalk does the right thing so I really don't know what's going on.

I'll look into the spinners though, seems to be an easier target.

Kovensky:
@maxnordlund `cat` will output the raw escapes as is to the terminal, which is why you don't see them. `\e[2K\e[1G` just literally clear the line and put the cursor at the start.

maxnordlund:
Ah, right. I've changed it to disable the functions in util.js which seems to have done the trick.

maxnordlund:
The test failures seems to all be timeouts. Just ran them on laptop and they're good. @kittens can you take a look to se what is happening, please?

(I can't access appveyor because it's private?)

Daniel15:
Does `supportsColor` allow for explicitly overridding the setting with a `--color` flag? There's cases where that's useful. For example, AppVeyor's build runner is not a TTY, but their UI actually **does** support ANSI escape codes (see the very bottom of https://ci.appveyor.com/project/Daniel15/appveyorcolourtest/build/1.0.4)

maxnordlund:
@Daniel15 see https://github.com/chalk/supports-color/blob/8400d98ade32b2adffd50902c06d9e725a5c6588/index.js#L16-L27

kittens:
Can you please rebase

maxnordlund:
Done

Issue #753, opened 2016-10-11
Avoid os.networkInterfaces on Windows Subsystem for Linux (Bash on Windows)


Daniel15:
`os.networkInterfaces` fails on WSL (Windows Subsystem for Linux). We call that at https://github.com/yarnpkg/yarn/blob/master/src/util/network.js#L9 to determine if the user is online, based on whether they have a non-localhost IP address.  If running under WSL, we should either use an API that works in that environment, or just always assume that the user is online.

See #636 for a prior discussion.

Daniel15:
Fixed in #772, thank you @mogest!

bb1:
Well this doesn't work anymore with the current WSL version.

Issue #636, opened 2016-10-11
Document that Yarn currently doesn't work on Bash on Windows


wski:
I don't really think this is Yarn's responsibility to fix, but it should be noted it doesn't work on Ubuntu bash on Windows, where as npm cli does. The issue is down to uv_interface_addresses something a lot of packages are broken by. Again I'm am fairly sure this is a wontfix but it should be noted either way.

Daniel15:
Yeah, known issue with WSL at the moment: https://github.com/Microsoft/BashOnWindows/issues/468.

Daniel15:
Closing this out since it's an upstream issue. Thank s for reporting!

rixrix:
hi, can this be added - if it hasn't yet/ if applicable - somewhere in readme or docs that this is a known issue ? in a little way this would help people like me from spending their time going through the ubuntu package install process. thanks

Daniel15:
Sure thing, let's repurpose this task to add a note to the docs.

wski:
https://github.com/yarnpkg/website/compare/master...wski:patch-1

wski:
https://github.com/yarnpkg/website/pull/129

mogest:
For those on Bash on Windows wanting to try out yarn, there's a simple workaround.

Update these two files:
/usr/lib/node_modules/yarn/lib/util/network.js
/usr/lib/node_modules/yarn/lib-legacy/util/network.js

In both files, change line 13 to read `const interfaces = {};`

Haven't run into any other compatibility issues apart from that one.

rixrix:
@mogest wonder if it is a good candidate for a PR upstream ? looking at the uservoice comments from WindowsBash team, they're working on it but no definite timeline. Otherwise there would `Zero` users from Windows on Ubuntu Bash. I know there's a Windows version but I'd prefer using the native bash. thanks

mogest:
@rixrix A try/catch block could be wrapped around that line so that it defaults to {} if the `os.networkInterfaces()` call fails, but I can see why the yarn maintainers might want to wait for Microsoft to implement the call.

Daniel15:
@mogest - Wow, I didn't realise that we just need to modify a single call site to get Yarn running in WSL. That's awesome!

[...]

PR #772, opened 2016-10-12
Support unimplemented call in Windows Subsystem for Linux

mogest:
Yarn works fine on Windows Subsystem for Linux (WSL) apart from a single call
to enumerate network interfaces.  If this call fails on WSL, we instead assume
that the user has an internet connection and continue.

See #753 and #636 for discussion.

With this change, the test suite works better on WSL but it still has multiple failures.  I've run the test suite on Linux and it passes fine.

### DIFF
--- a/src/util/network.js
+++ b/src/util/network.js
@@ -29,7 +29,7 @@ export function isOffline(): boolean {
     const addrs = interfaces[name];
     for (const addr of addrs) {
       if (LOCAL_IPS.indexOf(addr.address) < 0) {
-        // found a possible local ip
+        // found a possible remote ip
         return false;
       }
     }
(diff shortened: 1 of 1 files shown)


src/util/network.js
```diff
@@ -6,7 +6,20 @@ const IGNORE_INTERFACES = ['lo0', 'awdl0', 'bridge0'];
...
+  } catch (e) {
+    // As of October 2016, Windows Subsystem for Linux (WSL) does not support
+    // the os.networkInterfaces() call and throws instead. For this platform,
+    // assume we are online.
+    if (e.syscall === 'uv_interface_addresses') {
+      return false;
+    } else {
+      throw e;
+    }
+  }
```
jdalton: With the `return` in the `if` block you don't need an `else`.

mogest: I like doing that in my code because it makes the flow of execution clear.

Happy to remove the `else` if you'd prefer.

jdalton: I'm not core, just doing a review. Take my suggestions with large grains of salt.

cpojer: yea this seems fine.

cpojer:
thank you, this is awesome!

ludo237:
Good job, awesome

PR #784, opened 2016-10-12
Properly save scoped packages in offline mirror.

skevy:
When I was converting more Exponent projects to use `yarn` this afternoon, I encountered a perplexing bug.

In this project, there were two dependencies that were installed: "@exponent/md5hex@^1.0.0" and "md5hex@^1.0.0". Also, I'm using an offline mirror with this project.

When installing on my local machine and on my test server, things worked fine -- I didn't notice an issue. But when installing in Docker, for some reason I was getting a "Bad hash" error, in which it was telling me that the hash of the md5hex tar.gz file was not correct. 

As it turns out, there were entries for "@exponent/md5hex@^1.0.0" and "md5hex@^1.0.0" in my lock file, both pointing to the same file in the offline cache, but with different hashes. This was the cause of the bad hash issue.

The actual bug was that Yarn was saving scoped packages in the offline mirror by their base name, which meant that if two packages in a project had the same name and version, but one of them was scoped and the other was not, only one of these would be saved in the offline mirror, and then this bad hash transient error could potentially occur.

This PR fixes this behavior, and saves scoped npm packages in the offline mirror in the form `@[scope]-[package name]-[version].tar.gz`.

*Test plan:*

Run tests :) 

I added two tests and modified one -- the two I added just verified that it saved both scoped and unscoped packages in the offline mirror with the correct name. The modified test was the "install-from-offline-mirror" integration test -- I simply added a scoped package (along with the correct entry in the yarn.lock file) hopefully to prevent possible regressions in the future.


skevy:
cc @ide

skevy:
@cpojer these test errors say:

```Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.```

I also see these all the time internally when I run tests on Exponent projects...is there a simple fix for this, or is this a Jest/Jasmine bug?

**Comment by cpojer** [comment]
The problem you have internally is likely babel taking too long on the first uncached run (I assume you use inline-requires?). The only solution there is to bump the timeout.

It happens on Yarn for other reasons, mostly because of network failures.

File: src/fetchers/git-fetcher.js [hunk 1]
```diff
@@ -73,7 +73,17 @@ export default class GitFetcher extends BaseFetcher {
...
+    // Get the tarball filename from the url
+    const {pathname} = url.parse(this.reference);
+    let tarballFilename;
+    if (pathname == null) {
+      tarballFilename = this.reference;
+    } else {
+      tarballFilename = path.basename(pathname);
+    }
+
+    let tarballInMirrorPath = this.config.getOfflineMirrorPath(tarballFilename);
```
bestander: would this be a breaking change?

bestander: (sorry for a delay, I was dealing with documentation and slides debt, the fun stuff:) 
Can you come up a non breaking approach?
We would not want to force people to regenerate yarn.lock files after they started using them.

ide: @bestander I'm not sure if a non-breaking approach is possible without adding a lot of special logic for backwards compatibility, but this could be OK. My understanding is that the breaking change affects only people who use an offline mirror and have a project that depends on a scoped package. Some of those people also have buggy lockfiles that sometimes trigger "bad hash" errors. After this PR they will get just one more "bad hash" error that will go away after they regenerate the lockfile.

bestander: Sorry, I missed that this only breaks the scoped packages users and yeah, I agree it should be fine to fix this.

bestander:
That all now makes sense, thanks a lot for added tests, @skevy

PR #791, opened 2016-10-12
Support private registries (non-auth) and fix git resolver

palmerj3:
* Basic support for private registries defined in .npmrc.
* Fix git resolver so it properly matches non-github git+https, etc packages.

I want to use Yarn at work and this lets me do that. We utilize a private registry and github enterprise.

**Test plan**
There are currently no tests for npm-registry to begin with so I didn't add anything there.
For git+https I added a test to package-resolver and fetchers.js.

### DIFF
--- a/__tests__/fetchers.js
+++ b/__tests__/fetchers.js
@@ -68,6 +68,19 @@ test('GitFetcher.fetch', async () => {
   expect(name).toBe('font-roboto');
 });
 
+test('GitFetcher.fetch with git+https', async () => {
+  const dir = await mkdir('git-fetcher');
+  const fetcher = new GitFetcher(dir, {
+    type: 'git',
+    reference: 'git+https://github.com/PolymerElements/font-roboto',
+    hash: '2fd5c7bd715a24fb5b250298a140a3ba1b71fe46',
+    registry: 'bower',
+  }, await createConfig());
+  await fetcher.fetch();
+  const name = (await fs.readJson(path.join(dir, 'bower.json'))).name;
+  expect(name).toBe('font-roboto');
+});
+
 test('TarballFetcher.fetch', async () => {
   const dir = await mkdir('tarball-fetcher');
   const fetcher = new TarballFetcher(dir, {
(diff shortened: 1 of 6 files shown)


src/registries/npm-registry.js
```diff
@@ -47,7 +47,11 @@ export default class NpmRegistry extends Registry {
   }
 
   request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<?Object> {
-    const registry = removeSuffix(String(this.registries.yarn.getOption('registry')), '/');
+    let registry = removeSuffix(String(this.registries.yarn.getOption('registry')), '/');
+
+    if (this.config.registry) {
+      registry = this.config.registry;
```
fxb: Do you need to remove a possible trailing slash in this case as well? (see removeSuffix above)

jomaxx: +1 for trailing slash. Could also do the conditional at assignment and keep registry as a const

__tests__/fetchers.js
```diff
@@ -66,6 +66,19 @@ test('GitFetcher.fetch', async () => {
...
+test('GitFetcher.fetch with git+https', async () => {
+  const dir = await mkdir('git-fetcher');
+  const fetcher = new GitFetcher(dir, {
+    type: 'git',
+    reference: 'git+https://github.com/PolymerElements/font-roboto',
+    hash: '2fd5c7bd715a24fb5b250298a140a3ba1b71fe46',
+    registry: 'bower',
+  }, await createConfig());
+  await fetcher.fetch();
+  const name = (await fs.readJson(path.join(dir, 'bower.json'))).name;
```
jomaxx: If syntax is supported you could do this:
const { name } = await fs.readJson(...);

palmerj3:
@fxb @jomaxx I updated using some of your suggestions. Looks like there are some other commits recently (not mine) that are causing some failing tests and lint checks. I'll rebase if those get fixed.

jomaxx:
@palmerj3 looks good once tests are passing. Not by a computer until Sunday so won't be able to pull down and try it out until then.

mager:
I tested this code and it works well.

kittens:
Can you please rebase

palmerj3:
@kittens ok all set

palmerj3:
Closing since https://github.com/yarnpkg/yarn/pull/839 solves the private registry problem.

I have opened a separate PR just for the git resolver issue.
https://github.com/yarnpkg/yarn/pull/1107

PR #793, opened 2016-10-12
fix tgz glob for powershell build (echo #729 fix)

FLGMwt:
**Summary**
#729 fixed the .sh build script for the yarnpkg=>yarn rename, but not the powershell script. This fixes that.

**Test plan**
`scripts/build-dist.ps1` now works, assuming a tar install.

### DIFF
--- a/scripts/build-dist.ps1
+++ b/scripts/build-dist.ps1
@@ -6,7 +6,7 @@ if (Test-Path dist) {
   rm dist -Recurse
 }
 mkdir dist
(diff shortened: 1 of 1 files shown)


Daniel15:
I did mention this in a comment on https://github.com/yarnpkg/yarn/commit/aea9e24b514eb5e3dfdb8abd24539977ade87170#comments, maybe I should have just submitted a diff when I noticed it 😛 

Thanks for fixing it!

FLGMwt:
Whoops ;p No problem : D

PR #794, opened 2016-10-12
fix(npm-test): fixes import missing in why.js

sendilkumarn:
**Summary**
Currently npm run lint / test fails because of missing import. Fixing the same.


cpojer:
Just merged another PR with the fix for this. Thank you for helping out!

PR #795, opened 2016-10-12
source the rc.

hemanth:
**Summary**
Instead of requesting the user to open a new terminal.

**Test plan**
N.A


hemanth:
`travis` is failing due to timeouts, not related to this PR.

scripts/install-latest.sh
```diff
@@ -41,8 +41,8 @@ yarn_link() {
     printf "$cyan> We've added the following to your $YARN_PROFILE\n"
     echo "> If this isn't the profile of your current shell then please add the following to your correct profile:"
     printf "   $SOURCE_STR$reset\n"
-
-    printf "$green> Successfully installed! Please open another terminal where the \`yarn\` command will now be available.$reset\n"
+    source $YARN_PROFILE
+    printf "$green> Successfully installed! `yarn` command will now be available. $reset\n"
```
mathiasbynens: Completely unrelated to this patch, but “will now” should really be “is”

hemanth: Successfully installed!  You may `yarn` now. 

^ How does it sound ;) ?

kittens:
Profiles aren't necessarily safe to execute multiple times in the same shell session so I'm not sure this is a good idea.

hemanth:
> Profiles aren't necessarily safe to execute multiple times in the same shell session

Agree but in this scenario it seems safe?

scripts/install-latest.sh
ZakariaRidouh: @hemanth I think this would be slightly better:

Successfully installed! You can run `yarn` now.

hemanth: Let us see, if this PR is going in at all!

kittens:
Why do you think in this scenario it seems safe? The profile code will be executed twice, all variables will be exported twice. ie. if you have `export PATH="$PATH:/whatever"` in your profile then it'll be executed twice duplicating the variable as well as many other issues. It's fundamentally unsafe.

mathiasbynens:
Honest question: what’s “unsafe” about repeating `export PATH="$PATH:/whatever"`? Trying to imagine an attack scenario here and failing.

bestander:
We could give user an option to run `source $YARN_PROFILE` or open terminal.

bestander:
I talked to team mates, a few people were not happy about executing some scripts from .bashrc multiple times so let's find another way.
I think a recommendation to open a new terminal or running `source .bashrc` manually could be a reasonable alternative

Daniel15:
Running `exec $SHELL -l` should be a safer way to accomplish the same thing - It'll start a new shell process and the `-l` flag runs it as a login shell so it'll execute the profile script.

PR #796, opened 2016-10-12
Include MessageErrors class

swang:
**Summary**
MessageError was missing from `src/cli/commands/why.js` from PR #741. This will probably cause it to not work as intended. Plus it was causing errors from the linter.

**Test plan**
I don't think any is needed.


cpojer:
Oh oops, yeah there is some CI flakiness and I was assuming the prior PR was fine. Thanks for following up!

PR #797, opened 2016-10-12
Add ied as prior art

statianzo:
Per HN comment from @kittens 

> We initially used a method similar to ied when we first started experimenting with internal usage at Facebook.

https://news.ycombinator.com/item?id=12685936


cpojer:
thanks, let's give credit to ied! :)

kittens:
This approach is also used by a variety of other clients and package managers so I don't think it's fair to explicitly mention ied. We actually went down the symlink road before we found out about ied.

alexanderGugel:
Don't feel strongly about it TBH, but the only other client using this is pnpm AFAIK, which itself started out as a fork of ied (see https://github.com/rstacruz/pnpm/issues/415#issuecomment-252983963).

PR #799, opened 2016-10-12
Improvements on Travis CI

BanzaiMan:
**Summary**

This PR improves Travis CI builds in the following ways:

1. Runs on containers, so Linux jobs start faster
1. Cleans up `git`-related tasks
  1. Uses `apt` addon to update `git`. Strictly speaking, this may be unnecessary (depending on what version of `git` is available on the image).
  1. Skips updating `git` on the Mac. This also appears unnecessary, and adds some time to the build, since Homebrew now runs `brew update` in the background, which can be quite slow on an old package index.
1. Tweak `$TEST_TYPE` values, so that `script` section is easier to understand.

**Test plan**

Not applicable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,12 @@ env:
   - NODE_VERSION="4" TEST_TYPE="test-ci"
 
 matrix:
-  include:
-    - env: NODE_VERSION="6" TEST_TYPE="test-ci"
+  exclude:
+    - env: TEST_TYPE="build-dist"
       os: osx
-    - env: NODE_VERSION="4" TEST_TYPE="test-ci"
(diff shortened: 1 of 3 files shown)


ZakariaRidouh:
How much does this improve the build times? Just wondering :smile: 

great job!

kittens:
Awesome, thank you so much!

.travis.yml
```diff
@@ -1,9 +1,15 @@
 ---
 git:
   depth: 1
-sudo: true # can't use this due to git upgrade requirement
 language: generic
 
+sudo: false
```
vvo: Heads up, I think newly created repositories are always in container mode by default now, we could remove this I guess

Issue #721, opened 2016-10-11
postinstall hook doesn't appear to be run


wavded:
**Do you want to request a *feature* or report a *bug*?**

Not sure if its a feature or a bug, but I was expecting my `postinstall` hook to run after `yarn` was executed and it did not.

**What is the current behavior?**

`yarn` installs dependencies but does not run `postinstall` hook in the `package.json` file.

**If the current behavior is a bug, please provide the steps to reproduce.**

Sample `package.json`:

```json
{
  "name": "hi",
  "version": "1.0.0",
  "scripts": {
    "postinstall": "echo hi"
  },
  "dependencies": {
    "lodash": "*"
  },
  "main": "index.js",
  "license": "MIT"
}
```

Run `yarn`.  Do not see `hi` echoed.

**What is the expected behavior?**

Node 6.7.0 and Yarn 0.15.1 and macOS 10.12

**Please mention your node.js, yarn and operating system version.**

wavded:
I noticed this comment from TC that may shed some light?

> By default, npm also allows developers who write these packages to execute other code that’s needed as part of the install process. That creates security issues, though, so Yarn does not have this feature.

-- [Source](https://techcrunch.com/2016/10/11/facebook-partners-with-google-others-to-launch-a-new-javascript-package-manager/)

Perhaps `postinstall` was removed intentionally?  It is handy when someone is setting up a project on their machine to have any additional setup stuff done at that point.

Restuta:
It's handy in few other cases as well, e.g. on heroku it's a recommended way to customize build process https://devcenter.heroku.com/articles/nodejs-support#customizing-the-build-process

rjmunro:
This is strange, because in my experience, I found that the postinstall script was run for a package I was installing, but it didn't work for some reason, which I have documented in  https://github.com/yarnpkg/yarn/issues/614.

shiftkey:
@rjmunro I also see the `postinstall` hooks working fine for installed packages, but any `postinstall` command defined in my `project.json` are ignored.

rjmunro:
@shiftkey That sounds exactly backwards if the intention is to prevent third party code running on install unnecessarily.

glebmachine:
Looks the same issue here: https://github.com/yarnpkg/yarn/issues/614

kittens:
Fixed via #800.

henev:
Still reproducible with Angular 2 Quickstart tutorial as described here: #818

vith:
[...]

PR #800, opened 2016-10-12
Add pre and post install script calls

### LINKED ISSUE #783: Wrong script runed after install

Should be a bug.

`propublish` script is runned after I use yarn

screen shot 2016-10-12 at 9 41 56 am

This is scripts part of my `package.json`:
``` json
  "scripts": {
    "postinstall": "webpack",
    "prepublish": "webpack --config webpack.prod.js"
  },
```

Run `yarn` command

It should run postinstall script.

node v6.0
MacOs
yarn 0.15.1

FLGMwt:
**Summary**
Adds calls to the following named scripts: `preinstall`, `install`, `postinstall` and wraps `prepublish` in an `if !prod`, as it is in `npm install` Should resolve #783 Happy to discuss.

**Test plan**
Given the following package.json:
```
{
  "name": "junk",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "preinstall": "echo 'preinstall'",
    "install": "echo 'install'",
    "postinstall": "echo 'postinstall'",
    "prepublish": "echo 'prepublish'"
  }
}
```
`npm install` outputs the following:
```
PS C:\git\junk> npm i

> junk@1.0.0 preinstall C:\git\junk
> echo 'preinstall'

'preinstall'

> junk@1.0.0 install C:\git\junk
> echo 'install'

'install'

> junk@1.0.0 postinstall C:\git\junk
> echo 'postinstall'

'postinstall'

> junk@1.0.0 prepublish C:\git\junk
> echo 'prepublish'

'prepublish'
npm WARN junk@1.0.0 No description
npm WARN junk@1.0.0 No repository field.
```
`yarn`, however only executes `prepublish` (before changes):
```
PS C:\git\junk> yarn
yarn install v0.15.1
info No lockfile found.
success Nothing to install.
$ echo 'prepublish'
'prepublish'
Done in 0.06s.
```

After changes:
```
PS C:\git\junk> yarn
yarn install v0.15.1
info No lockfile found.
$ echo 'preinstall'
'preinstall'
success Nothing to install.
$ echo 'install'
'install'
$ echo 'postinstall'
'postinstall'
$ echo 'prepublish'
'prepublish'
Done in 0.09s.
```
and with `--prod` flag:
```
PS C:\git\junk> yarn --prod
yarn install v0.15.1
info No lockfile found.
$ echo 'preinstall'
'preinstall'
success Nothing to install.
$ echo 'install'
'install'
$ echo 'postinstall'
'postinstall'
Done in 0.08s.
```

I'm also working through how to add unit tests for this and can hold merge for that is desired.

### DIFF
--- a/scripts/build-dist.ps1
+++ b/scripts/build-dist.ps1
@@ -6,7 +6,7 @@ if (Test-Path dist) {
   rm dist -Recurse
 }
 mkdir dist
-mv yarnpkg-*.tgz dist/pack.tgz
+mv yarn-*.tgz dist/pack.tgz
 
 cd dist
 tar -xzf pack.tgz --strip 1
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -729,9 +729,15 @@ export async function run(
     throw new MessageError(reporter.lang('installCommandRenamed', `yarn ${command} ${exampleArgs.join(' ')}`));
   }
 
+  await executeLifecycleScript(config, 'preinstall');
+
   const install = new Install(flags, config, reporter, lockfile);
   await install.init();
 
   // npm behaviour, seems kinda funky but yay compatibility
-  await executeLifecycleScript(config, 'prepublish');
+  await executeLifecycleScript(config, 'install');
+  await executeLifecycleScript(config, 'postinstall');
+  if (!flags.production) {
+    await executeLifecycleScript(config, 'prepublish');
+  }
 }


FLGMwt:
Also #721 which might be a dupe.

Kovensky:
Note that npm means to deprecate how prepublish currently works in npm 4.0, to be released this month: https://github.com/npm/npm/issues/10074

FLGMwt:
@Kovensky Nice catch!

After reviewing conversation there, it looks like it's still pretty up in the air what npm plans to do exactly to change the behavior. I'd be up for porting whatever they implement if that's desired, but these changes at least get us closer to npm compatibility (and resolves issues for a few people migrating).

That said, I'm happy to remove `prepublish` and just get the install hooks in. Looking for guidance : )

jamesjnadeau:
@FLGMwt awesome work on this!

IMHO, not having these hooks is a show stopper to a lot of projects using yarn.

I would say feature parity with npm is what you should be shooting for, and having prepublish work the same way is probably best.

ljharb:
Unless every single lifecycle script functions exactly the same as it does on npm, then yarn is not a drop-in replacement for npm - which I understand to be the goal. That includes prepublish's awful behavior.

billxinli:
@kittens Curious on when this fix will get pushed to a release. This fix is going to fix a lot of issues with modules that depends on these hooks

Kovensky:
Just a note: `npm@2` will run `preinstall` scripts before installing dependencies, but `npm@3` runs `preinstall` only after all dependencies are installed.

This patch runs `preinstall` before installing dependencies.

dominictarr:
@ljharb A beer says that prepublish isn't that big a problem, and this is a good way to find out.

ljharb:
@dominictarr if you're right, I will buy you a whole case. But if there are non-zero packages that depends on it, that would be enough for it to be a huge problem.

dominictarr:
in the browser world, not everything is completely compatible, and, well, it's annoying some times, but we survive. I wasn't around at the time when javascript was brand new... but I'm gonna hazard to suggest that "mutating" a little bit, could actually make something _better_.

And in any case, there isn't a _spec_ for npm just an implementation, however if there are more than one implementation, maybe it's time to start thinking about a spec. (maybe wait another 15 minutes though)

lipis:
@dominictarr on point.. since there are no specs let's rethink some of the procedures and not blindly copy stuff from npm.. just because.

ljharb:
All the things browsers agree upon *are* the spec - that's true in the JS standard as well as in browser standards. Since `npm` has been the only implementation for this long, what it does *is* the spec. If yarn deviates, yarn will lose out.

curtisblackwell:
Any idea when we can expect the next release so we can rely on this functionality?

jordansexton:
Any update on the status of a release including this?

janusch:
I am wondering if this also works for the 'yarn-offline-mirror' feature?
E.g. in node-sass the postinstall will download the correct binaries, are they added to the offline-mirror?

Some clarification would be very helpful to me!

lordgreg:
Yarn doesn't install selenium webdriver when installing node module `protractor`. Using `npm install protractor` works as it should.

ndbroadbent:
Hello, I am testing the `yarn` nightly build (v0.22.0-20170215.1538), and this doesn't seem to be working for me. Or maybe I just don't understand what it should be doing.

I have forked a package and am trying to install it from GitHub:

```
yarn add github:ndbroadbent/redux-persist-migrate
```

This `package.json` contains a prepublish line:

```
"prepublish": "npm run clean && npm run build"
```

This is not being run during the installation. I would expect this to be run, since GitHub is only for source code, and people don't usually check in the build artifacts.

I just read through [this issue on npm](https://github.com/npm/npm/issues/3055), and it looks like npm also doesn't run the `prepublish` script. I think `npm` should run this script, and I think `yarn` should, too.

Or should I just add this step as a `preinstall` hook on my own fork? Or just run the `preinstall` step myself and check in the built files?

jamesjnadeau:
@ndbroadbent: is `preinstall` what you are looking for?

ndbroadbent:
@jamesjnadeau I tried adding a `preinstall` hook, but that doesn't seem to work either. `yarn` is not even running the `prepublish` or `preinstall` hooks when installing from a local directory. When I run: `npm i ../redux-persist-migrate`, I see it runs the `prepublish` and `preinstall` hooks. 

`npm i ndbroadbent/redux-persist-migrate` doesn't run those hooks, and neither does `yarn add ndbroadbent/redux-persist-migrate`.

So I guess the only way is to push a new branch with the build artifacts checked in?

PR #801, opened 2016-10-12
Delete RELEASING.md

Daniel15:
Deletes the `RELEASING.md` file, as the document also exists on the website at https://yarnpkg.com/en/org/release-process (which feels like the right place for it).


cpojer:
yeah duplication isn't useful here.

PR #803, opened 2016-10-12
remove isntall alias

thefourtheye:
**Summary**

I don't think `isntall` is a commonly made typo.

**Test plan**

NA

### DIFF
--- a/src/cli/aliases.js
+++ b/src/cli/aliases.js
@@ -21,7 +21,6 @@ export default {
   'dist-tags': 'tag',
   adduser: 'login',
   author: 'owner',
-  isntall: 'install',
   la: 'ls',
   ll: 'ls',
   r: 'remove',


cpojer:
It doesn't hurt to have this alias, it's just one line of code. Aliases in Yarn simply direct you to the proper command. Personally, I have made this typo many times (note: I didn't add or request this feature) and having yarn recommend the right command is more signal than telling me I have an invalid command.

I don't think we need to remove this.

ljharb:
@cpojer aliases that error out, instructing the user about the command they likely intended are great. errors that allow invalid commands to be typed can make log spelunking and googling harder. It would be ideal imo to keep all the typos, but to have them error out like in git - rather than to follow npm's lead which I think has proven to be somewhat problematic for users.

cpojer:
@ljharb this is what we do. Please try out yarn.

cpojer:
Oh, I'm sorry. My brain is a bit fried from this huge release.

Here is the actual output:
```
[cpojer: ~/Projects/yarn-website (master)]$ yarn isntall
yarn isntall v0.15.1
error Did you mean `yarn install`?
info Visit http://yarnpkg.com/en/docs/cli/isntall for documentation about this command.
```

which reminds me the docs link should direct to `install` :D

Issue #804, opened 2016-10-12
Don't install a package twice from cache


yunxing:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
Some package installation are not idempotent. If you call post-install twice, the second installation may fail. Currently, after we install a package, we store the build artifacts in the cache. During the second installation, if we call postinstall on the package again, it may fail. 

**Repro**
An example of this is ocamlRe. If you do

```
yarn install ocamlRe
rm -rf node_modules
yarn install ocamlRe 
```

the second installation will fail.

LeiZeng:
I have the same question, which I don't think it's a "issue".

```bash
yarn install some-package
rm -rf node_modules/one-dependency

yarn check
# you can get the warning that the `one-dependency` is missing

yarn install
# success Already up-to-date.

yarn check
# you can get the warning that the `one-dependency` is missing
```
yarn install does not `check` the status first.

The solution is simple:
`yarn install --force` or `yarn --force`

Could be better if it can `check` and fix itself

hpurmann:
I guess this should be closed as https://github.com/yarnpkg/yarn/pull/805 is merged now?

wyze:
Yup, thanks @hpurmann!

PR #805, opened 2016-10-12
Don't re-install cached artifacts.

yunxing:
#804 

This patch makes sure that if we load a package from cache, we don't install it again.

**Test Plan**

```
yarn add ocamlRe
rm -rf node_modules
yarn add ocamlRe
```


kittens:
Can you please rebase

bestander:
ping @yunxing, this needs a rebase

yunxing:
sorry just saw this. Rebasing now

PR #808, opened 2016-10-12
Remove duplicate hashing utility

bripkens:
**Summary**
While trying to analyze why security checks fail (due to hash mismatches), I stumbled upon two hash utils with different default hashing strategies. My first thought was that this is why it isn't working. Turns out, that the two versions are used in different parts of the code.

This PR removes the duplicate hashing util function (for less confusion in the future).

**Test plan**
This PR removes one existing function. Tests already cover this path.


kittens:
Oops, thank you!

Issue #735, opened 2016-10-11
`yarn run ...` no longer working


SethDavenport:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

`package.json`:
```
  "scripts": {
    "wat": "echo wat",
    // ...
```

My project has a bunch of stuff as `npm` scripts in `package.json`. Previously, with the old package name, (yarnpkg), they worked fine:

```sh
> npm install -g yarnpkg
> yarn --version
0.15.1
> yarn run wat
yarn run v0.15.1
$ echo wat
wat
✨  Done in 0.08s.
```

However since obeying the nice deprecation message and installing the `yarn` package instead, `yarn run` fails on everything:

```sh
> npm uninstall -g yarnpkg
> npm cache clean
> npm install -g yarn
> yarn --version
0.15.1
> yarn run wat
yarn run v0.15.1
$ "echo wat"
sh: echo wat: command not found
error Command failed with exit code 127.
info Visit http://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

See above.

**What is the expected behavior?**

I would expect yarn 0.15.1 from `npm install -g yarn` to behave the same as yarn 0.15.1 from `npm install -g yarnpkg`

**Please mention your node.js, yarn and operating system version.**

Mac OS X El Capitan 10.11.5
Node 6.3.0
Yarn 0.15.1

cozuya:
Works fine on Windows fyi.

dustykeyboard:
I am seeing the same issue, scripts defined in `package.json` are not being executed with run command.
```
> node -v
v6.5.0
> npm -v
3.10.3
> yarn --version
0.15.1
>  yarn start
yarn start v0.15.1
$ "react-scripts start" 
sh: react-scripts start: command not found
error Command failed with exit code 127.
info Visit http://yarnpkg.com/en/docs/cli/start for documentation about this command.
```

However calling the script directly executes correctly.
```
> yarn run react-scripts start
yarn run v0.15.1
$ "[...]/node_modules/.bin/react-scripts" start
...
```

notahat:
I can confirm this. OS X 10.12 with Node 6.7.0.

Yarn is attempting to run `echo wat` as the command, rather than `echo` as the command with `wat` as an argument.

I also came across this while testing with `create-react-app`. See https://github.com/facebookincubator/create-react-app/issues/896

[...]

Issue #733, opened 2016-10-11
yarn run with bin in local node_modules


billxinli:
**Do you want to request a *feature* or report a *bug*?** 
Report a bug

**What is the current behavior?**
From npm with respect to `npm run`: `In addition to the shell's pre-existing PATH, npm run adds node_modules/.bin to the PATH provided to scripts. Any binaries provided by locally-installed dependencies can be used without the node_modules/.bin prefix.`

**If the current behavior is a bug, please provide the steps to reproduce.**

Add additional entry to the `scripts` section of the `package.json` without installing the dependency globally (in this case mocha).

```
  "scripts": {
    "test-api": "mocha test/api"
  },
```

**What is the expected behavior?**

If I run `npm run test-api` it will resolve `mocha` to the local copy. However `yarn run test-api` will produce:

```
$ "mocha test/api" 
sh: 1: mocha test/api: not found
```

**Please mention your node.js, yarn and operating system version.**
node: 6.7.0
yarn: 0.15
os: Ubuntu 16.04

Daniel15:
Thanks for the bug report! This is pretty important. I'll try to take a look tonight.

notahat:
Duplicate with more info here: https://github.com/yarnpkg/yarn/issues/735

Daniel15:
@notahat - That's actually a separate issue. This issue is saying that `yarn run` is not running executables in the `node_modules/.bin` directory (eg. `mocha test/api` in a script should essentially run `node_modules/.bin/mocha test/api` if `mocha` is an npm dependency of the app), whereas your issue #735 is about running shell commands (I think?)

notahat:
@Daniel15 I think it's the same thing.

It's not that yarn isn't finding stuff in `node_modules/.bin`, it's that yarn is shelling out in such a way that the command is `mocha test/api`, rather than `mocha` with an argument of `test/api`. Not surprisingly, the shell can't find an executable called `mocha test/api`.

Daniel15:
Ohh I see! That might be it. I'll take a look tonight 😄

slmgc:
@Daniel15 @notahat I think it's the root of the issue:
```bash
echo `npm bin` #path_to_the_project/node_modules/.bin
echo `yarn bin` #path_to_the_project/node_modules/.bin/node_modules/.bin
```
`yarn` duplicates `/node_modules/.bin` part.

notahat:
@slmgc I don't see that when I try it.

```
$ cd /path/to/project
[...]

PR #809, opened 2016-10-12
Fix executing scripts with spaces

### LINKED ISSUE #727: Lifecycle script fails on Ubuntu
Bug

running a lifecycle script on Ubuntu fails while running on macOS works.

- add `"foo": "ls -la"` to your `package.json`
- execute `yarn run foo`

You should see output like this: `sh: 1: ls -la: not found`

The expected behavior, and the behavior you would see on macOS is that "ls -la" would be executed, and the output would be printed to the terminal.

OS: Ubuntu 14.04.4 LTS
Node: v4.2.6
Yarn: 0.15.1

ticky:
**Summary**
The script quoting behaviour introduced in #663 breaks package.json script entries with spaces in them on Unix.

Instead, this quotes `node_modules/.bin` executable paths, which makes sure both package.json script entries, and executables can be called.

**Test plan**

For this package.json;
```json
{
  "name": "yarntest_pr809",
  "version": "1.0.0",
  "devDependencies": {
    "jest": "^16.0.1"
  },
  "scripts": {
    "test": "jest",
    "test-help": "jest --help"
  }
}
```
with dependencies installed,

1. **In a path with no spaces** (e.g. `~/yarntest_pr809`);
  1. `yarn run test -- --help` should print jest’s help message on the current release, and with this patch.
  2. `yarn run test-help` will *fail* on the current release, and print jest’s help message with this patch.
  3. `yarn run jest -- --help` will print jest’s help message on both the current release and with this patch.
2. **In a path with spaces** (e.g. `~/yarn test pr809`);
  1. `yarn run test --help` should print jest’s help message on the current release, and with this patch.
  2. `yarn run test-help` will *fail* on the current release, and work fine with this patch.
  3. `yarn run jest -- --help` will print jest’s help message on both the current release and with this patch.

Note that the use of a help flag within a script is a contrived example, but triggers the incorrect behaviour in this case.

fixes #727, #733, #735 :smile:

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -36,7 +36,7 @@ export async function run(
     if (await fs.exists(binFolder)) {
       for (const name of await fs.readdir(binFolder)) {
         binCommands.push(name);
-        scripts[name] = `${path.join(binFolder, name)}`;
+        scripts[name] = `"${path.join(binFolder, name)}"`;
       }
     }
   }
@@ -61,7 +61,7 @@ export async function run(
 
     if (cmds.length) {
       for (const [stage, cmd] of cmds) {
-        await execCommand(stage, config, `"${cmd}" ${args.join(' ')}`, config.cwd);
+        await execCommand(stage, config, `${cmd} ${args.join(' ')}`, config.cwd);
       }
     } else {
       let suggestion;


Daniel15:
Hi @ticky! Nice to see another Melburnian contributing to Yarn 😄 

For your test plan, I think it'd be good to include some examples of scripts that you tested with `yarn run`, and show the `package.json` you used. Yarn should handle both some sort of "global" command (eg. `node foo.js`, which runs `node` directly) as well as some local command (eg. `jest foo-test.js` when `jest` is listed as a `devDependency` in `project.json` but **not** installed globally, which runs `node_modules\.bin\jest`).

Maybe something like this:
```json
{
  "name": "yarntest_issue733",
  "version": "1.0.0",
  "devDependencies": {
    "jest": "^16.0.1"
  },
  "scripts": {
    "first": "node foo.js",
    "second": "jest __tests__/hello-test.js"
  }
}
```

Both `yarn run first` and `yarn run second` should work (assuming `foo.js` and `__tests__/hello-test.js` exist)

I suspect this will fix #735 too.

ticky:
Just don’t tell anyone I actually started out in Perth 😜

I didn’t see anywhere in the existing test suite where this kind of CLI interop was tested, so I figured I’d leave adding such a thing to the repo proper for another intrepid soul.

Yarn does not currently run any executables on the system path; only executables within `node_modules/.bin` and package.json scripts, so the scenario you describe here in `yarn run first` isn’t applicable.

To that end, and given the current behaviour, I’ve updated the test scenarios in the PR description.

AlicanC:
I'm adding tests for this in #821. I can't complete the tests right now because master is broken. If we can merge this I can continue with the tests.

ticky:
@AlicanC if you take a look at the newly updated PR comment, it should lead you to six good test scenarios for this! :smile: I can’t really comment on how the test hooks in, as I’m not that familiar with this application of async stuff!

ticky:
The current build failure seems to be a flaky test :wink:

Bartuz:
Is running scripts with `npm run` still valid? Does it cooperate well with what yarn does?

ticky:
Yarn puts executables in the same place npm is expecting them, so it works fine

acateland:
i hope this will be merged soon as it effectively resolve this issue. Good job

kittens:
Thank you so much!

lfittl:
@kittens Any ETA on a patch release that has this fixed, or a way to workaround until then? (looks like there is a bunch of folks depending on this fix)

Thanks for the amazing work on yarn btw :)

icd2k3:
@lfittl I've just been using `npm run` in the meantime and everything still works fine

Daniel15:
> or a way to workaround until then?

If you like, you can clone the Yarn repository and build it to get all the latest bugfixes. Something like this should be sufficient:
```
cd /usr/local/src
git clone git://github.com/yarnpkg/yarn.git
cd yarn
yarn install
npm run build
```

Then add `/usr/local/src/yarn/bin` to your PATH, or use `/usr/local/src/yarn/bin/yarn` directly.

Kovensky:
@Daniel15 `npm link` in yarn's source folder seems to DWIM and make yarn's git version available.

chiedo:
When does this version get pushed as the latest release so I can get it with `brew upgrade` or with Ubuntu?

joyfulelement:
Currently having the same issue attempting to execute run script with yarn on Mac OS and Unix on CI. Do we know if this fix would be released as part of `v0.15.2`? (Currently using `v0.15.1` at the time of this writing)

Petesta:
http://stackoverflow.com/questions/40049508/trouble-running-yard-run-on-scripts-object-in-package-json/40050317#40050317

icd2k3:
Just wanted to note that I don't see this issue when I install yarn via `curl -o- -L https://yarnpkg.com/install.sh | bash` so that might be another option while waiting for 0.15.2.

gekorob:
Thx @icd2k3, uninstalled yarn 0.15.1 (previously npm installed) and used the bash installer...
Everything is working now.

seehou:
Still hitting this issue on my osx.
```
  "scripts": {
    "start": "node server.js"
  },
```
```
yarn run start
yarn run v0.15.1
$ "node server.js" 
sh: node server.js: command not found
```

chiedo:
Anyone have a rough ETA on when this will be released?

guitarmanvt:
Anyone interested, please see #1156 for a work-around.

Issue #664, opened 2016-10-11
minor issue, yarn help <not-real-command> recommends you visit a 404 url


swang:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
right now at the command line, if you type `yarn help <command>` and command does not exist, yarn prints out basic usage of the cli, but it also tells you to visit `http://yarnpkg.com/en/docs/cli/<command>` even if <command> does not exist. users going to that page will just see a 404 page.

**If the current behavior is a bug, please provide the steps to reproduce.**
type, `yarn help not-really-a-command`, goto link thats printed out.
**What is the expected behavior?**
not sure. maybe it shouldn't print out a url if it's not a real yarn command? i was seeing if yarn had a command called `repo` and assumed it existed when yarn gave me a url for more documentation. i think it may be confusing for some users that see this.

**Please mention your node.js, yarn and operating system version.**
node 5.12.0
npm 3.8.6
yarn 0.15.1

AlicanC:
I tried to send a PR for this, but I couldn't figure out what and how I'm supposed to output.

The fix should be an `if (!command) { /* warn user */ }` after this line: https://github.com/yarnpkg/yarn/blob/c4654ce70cdb198f8d0b725683fb63d82e923c85/src/cli/index.js#L119

PR #810, opened 2016-10-12
Don't display a documentation url if actual command does not exist

swang:
**Summary**
This fixes issue #664, where the yarn cli was display a url for commands that don't actually exist in yarn.

**Test plan**
`yarn what` 
```
yarn what v0.15.1
error Command "what" not found.
```
`yarn why` (display a doc url)
```
yarn why v0.15.1
error Missing package name, folder or path to file to identify why a package has been installed
info Visit https://yarnpkg.com/en/docs/cli/why for documentation about this command.
```
`yarn what what what` (don't display doc url)
```
yarn what v0.15.1
error Command "what" not found.
```
`yarn why why why` (display a doc url)
```
yarn why v0.15.1
error Too many arguments, maximum of 1.
info Visit https://yarnpkg.com/en/docs/cli/why for documentation about this command.
```

`yarn ls no-arguments` (this will always display a doc url because the command here will always be a yarn command)
```
yarn ls v0.15.1
error This command doesn't require any arguments.
info Visit https://yarnpkg.com/en/docs/cli/ls for documentation about this command.
```

If there needs to be jest tests for this, please suggest what the best place would be to put a test like this in the `__tests__` folder.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -73,7 +73,9 @@ if (commandName === 'help' || commandName === '--help' || commandName === '-h') 
   commandName = 'help';
   if (args.length) {
     const helpCommand = hyphenate(args[0]);
-    commander.on('--help', () => console.log('  ' + getDocsInfo(helpCommand) + '\n'));
+    if (commands[helpCommand]) {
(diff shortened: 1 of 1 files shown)


kittens:
Awesome, thank you so much!

PR #811, opened 2016-10-12
Move CONTRIBUTING.md to .github folder

tfrommen:
The `.github` folder, [introduced in February](https://github.com/blog/2111-issue-and-pull-request-templates), can also be used as home for the `CONTRIBUTING.md` file. By doing so, one would take another step to not unnecessarily clutter the root directory.

This pull request does exactly this:

* move the `CONTRIBUTING.md` file from the root directory to `.github`;
* adapt the `.npmignore` file accordingly;
* adapt the `README.md` file, which links to the above file, accordingly.


kittens:
Thank you for the PR! I'm not sure we'd want this. A lot of people are used to looking for `CONTRIBUTING.md` in the root of the project and I feel like this would make it even less likely for people to find it.

tfrommen:
Ah, okay. Then this might not be for you.

I don't know any people who actively search for contrinuting guidelines, though, as they pop up when creating an issue anyway. :)

cpojer:
@tfrommen I've definitely struggled with this decision on Jest myself, so I see where you were coming from. Thank you for the PR, but I'm gonna close it out.

PR #813, opened 2016-10-12
Fix git dependencies with slashes inside branch name

saschagehlich:
**Summary**

Previously, branches with slashes such as `features/feature-branch` would not be recognized by the package resolver. This happened because ref names (e.g. `refs/heads/features/feature-branch`) were not parsed correctly. Instead of looking at everything after the second slash, it only looked at the string after the second slash and before the third. This resulted in `refs/heads/features/feature-branch` being parsed to `features` instead of `features/feature-branch`.

**Test plan**

I have added a test case to `__tests__/package-resolver.js` that tries to parse `https://github.com/babel/babel-loader.git#feature/sourcemaps`. Previously, this caused an error. This has been fixed now.

There is one more failing test though, but I'm sure it's not related to my change. Tests seem to be always failing for this project right now.


kittens:
Thank you!

Issue #633, opened 2016-10-11
Support shorthand package name with github default


farism:
It looks like yarn does not support the npm supported shorthand syntax (where github is the default) e.g. `react: facebook/react.git`

voxsim:
I tried to see this with the last version of yarn and it works as expected, someone can close it?

voxsim:
@bestander you can close this

PR #821, opened 2016-10-12
Add tests for command "run"

AlicanC:
**Summary**

In #809, suggested tests were these (twice, in straight and white-spaced paths):
- `yarn run test -- --help`
- `yarn run test-help`
- `yarn run jest -- --help`

In this PR I test the internal run() command and (AFAIK) it is not responsible converting `yarn run test -- --help` to `test --help` so the part where `-- --arg` becomes `--arg` should not be covered.

What I cover is:

1. run()ing a script in `package.json`.
1. run()ing a binary in `node_modules/.bin/`.
1. run()ing a binary in `node_modules/.bin/` with a white-spaced path.
1. run()ing a binary in `node_modules/.bin/` with a single arg.

1, 2 and 4 was working on release. #663 fixed 3, but broke 4 and with #809 all tests pass.

There are more to add, but I want to be sure if I'm doing this right. I'd like to get this merged (if it is right) and add others in another PR when I have more time.

Please comment.

**Test plan**

(What should a test plan be for a test?)


__tests__/commands/run.js
```diff
@@ -0,0 +1,31 @@
...
+  return run(config, reporter, flags, args);
+}
+
+test.concurrent('run should run script', (): Promise<void> => {
+  return runRun({}, ['some-script'], 'run-should-run-script');
+});
+
+test.concurrent('run should run binary', (): Promise<void> => {
+  return runRun({}, ['echo'], 'run-should-run-binary');
+});
```
ticky: Oh, I see what this is doing! If you do this, and add specs for the same thing but with spaces in the fixture directory name, that should have you covering the issue #633 intended to fix completely! :smiley:

ticky: Come to think of it, it could be reasonable to mock out `child_process` (used _all the way_ [over here](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/util/child.js#L9)) for these purposes, but I guess it depends if you want to actually integration test with the shell.

AlicanC:
Updated PR and original post.

__tests__/commands/run.js
```diff
@@ -0,0 +1,39 @@
...
+): Promise<void> {
+  const cwd = path.join(fixturesLoc, name);
+  const reporter = new reporters.NoopReporter();
+  const config = new Config(reporter);
+  await config.init({cwd});
+  return run(config, reporter, flags, args);
+}
+
+test.concurrent('run should run script', (): Promise<void> => {
+  return runRun({}, ['some-script'], 'run-should-run-script');
```
bestander: great job!
Usually tests should have something to expect, not just run and expect that it does not crash.
Can you add expectations that verify that the command has been executed?

AlicanC: I couldn't find a way to do that, but instead I've added expectations for calling of executeLifecycleScript().

I think these tests should only verify that run() can call executeLifecycleScript() properly, then another test can verify if executeLifecycleScript() can do its job properly.

bestander:
Thanks @AlicanC!

bestander:
Hey @AlicanC, the CI failed with errors https://circleci.com/gh/yarnpkg/yarn/1806.
Could you rebase and send a new PR?

PR #823, opened 2016-10-12
Chore - Created an .eslintrc file for the eslint configurations instead of having it in the package.json file

kazazor:
**Summary**

In this PR I simply extracted the `eslintConfig` out of it and placed it in a more standard way in the `.eslintrc` file.

Why?

1. The `package.json` file already have enough definitions in it and it is a big file, Let's minimize it.
* We're already using `.eslintignore` file, so why having the eslint configuration in the `package.json` file? Seems a little bit inconsistence..
* A lot of IDEs have built in plugins for eslint, having the configurations in a dedicated file for that makes that integration easier.
* Easier to find - when ever I'll need anything regarding eslint, I would just go to that file instead of searching for it in the `.package.json` file. Also, now this file and the `.eslintignore` file are one after the other.

**Test plan**

No need for tests here.. I checked the `yarn run lint` still works.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,39 @@
+{
+  "extends": "eslint-config-fb-strict",
+  "env": {
+    "jest": true
+  },
+  "plugins": [
+    "flowtype",
+    "no-async-without-await",
+    "yarn-internal"
+  ],
+  "rules": {
+    "yarn-internal/warn-language": 2,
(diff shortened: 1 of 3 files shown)


kittens:
We initially put it in there to minimise files in the root directory but your other points are really good and make sense. Thank you!

PR #829, opened 2016-10-12
install sh script: Do not "exit 1" if yarn already installed

PlugIN73:
**Summary**

I think it's bad idea to exit 1 in this case - because this script can't be used in pipeline or in ansible. Throwing exception breaks next movements. If Yarn already installed this script - it will warn and normally exited with exit code 0.
What do you think?


Sjolus:
Exit code 0 should indicate a successful completion of the intended task. If something is already installed then it should not be installed again, hence it should indicate it not performing the desired task when exiting.

Not sure what you mean by pipelines specifically, but you should not be installing yarn if it's already installed. I suggest making some conditional stuff in your Ansible playbooks/tasks instead. 

http://docs.ansible.com/ansible/playbooks_conditionals.html

Daniel15:
If you're on Linux, I'd suggest using the package repos rather than the installer script, particularly if you're trying to script the environment.

PR #830, opened 2016-10-12
cli: fix help for misspelled command

thefourtheye:
**Summary**

If the command is misspelled, the help message should point to
the URL with properly spelled command name.

**Test plan**

Commands to reproduce is in https://github.com/yarnpkg/yarn/pull/803#issuecomment-253134849

----

cc @cpojer

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -333,8 +333,9 @@ config.init({
       logError(errs);
     }
 
-    if (commands[commandName]) {
-      reporter.info(getDocsInfo(commandName));
+    const actualCommandForHelp = commands[commandName] ? commandName : aliases[commandName];
+    if (actualCommandForHelp) {
+      reporter.info(getDocsInfo(actualCommandForHelp));
     }
   }
 


cpojer:
@thefourtheye can you attach an screenshot showing this with `yarn isntall`?

src/cli/index.js
```diff
@@ -333,8 +333,8 @@ config.init({
       logError(errs);
     }
 
-    if (commands[commandName]) {
-      reporter.info(getDocsInfo(commandName));
+    if (commands[commandName] || aliases[commandName]) {
+      reporter.info(getDocsInfo(commands[commandName] || aliases[commandName]));
```
cpojer: instead of doing this twice, can we do:

```
const helpCommand = commands[commandName] || aliases[commandName];
if (helpCommand) {
  reporter.info(getDocsInfo(helpCommand));
}
```

helpCommand might not be the best name. Maybe `realCommandName`?

thefourtheye:
@cpojer `commands[commandName]` returns an object. So that wouldn't have worked. I used a ternary operator. PTAL

cpojer:
thank you!

thefourtheye:
screen shot 2016-10-12 at 2 00 30 pm

@cpojer Screenshot :-)

cpojer:
sweeet

PR #833, opened 2016-10-12
Emoji are emoji

WebReflection:
Accordingly to the [Unicode Consortium](http://unicode.org/emoji/):

> Emoji are pictographs (pictorial symbols) that are typically presented in a colorful form and used inline in text. They represent things such as faces, weather, vehicles and buildings, food and drink, animals and plants, or icons that represent emotions, feelings, or activities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
 * Multiple Registries: Install any package from either npm or Bower and keep your package workflow the same.
 * Network Resilience: A single request failing won't cause an install to fail. Requests are retried upon failure.
 * Flat Mode: Resolve mismatching versions of dependencies to a single version to avoid creating duplicates.
-* More emojis. 🐈
+* More emoji. 🐈
 
 ## Installing Yarn
 


mathiasbynens:
Well, actually…™ https://emojipedia.org/faq/#plural

WebReflection:
That entire page uses **emoji** term, but regardless, I've not mentioned the Unicode Consortium by accident: it's the standard that defines their usage.

The best part of this PR? It fails on Travis CI!

cpojer:
I'm not really much into bikeshedding like this and it isn't very productive. As @mathiasbynens says (hi! Good to see you), emojis is acceptable so we'll keep it. We aren't changing things for the sake of changing things.

It's also funny to see you both show up here given that the last time I saw you both in the same location was in Berlin just a year ago. Good times!

WebReflection:
> I'm not really much into bikeshedding like this and it isn't very productive.

As [twemoji](https://github.com/twitter/twemoji) library author, and as Unicode ML participant, I wasn't asking to change this for the sake of changing it, I came after endless discussions on this topic.

... but yeah, the world will keep spinning without this change, it wasn't even something I believed would have been accepted but it costed nothing to propose such change.

Maybe if Mathias would've have supported this change it would've been applied as _side-effect free no-brainer ™_, and it certainly took longer to me or him to write something than me to propose the PR :-P

See you next time ... in Berlin, where else :-)

Issue #521, opened 2016-10-06
Support for private packages


thejameskyle:
In order to allow installing [private packages](https://www.npmjs.com/features) Yarn will need to send a token to the headers of the request.

Private packages are `@scoped/packages` that were published with `npm publish --access=restricted`. The permissions of packages are managed through `npm access` and `npm team` which [are not yet added](https://github.com/yarnpkg/yarn/issues/377)

In the npm client, this token comes from the `.npmrc` and looks like this:

```
@nameofscope:registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=abc123
```

And it gets sent as this header:

``` yaml
Authorization: Bearer abc123
# alternatively:
Authorization: Basic username:password # <= base64
```

There's a [package for retrieving the token](https://github.com/rexxars/registry-auth-token). Although we may not want to store the token the same way npm does.

This token gets added to `.npmrc` on `npm login`. But `yarn login` doesn't even authenticate (it only stores username and email), so we may want to force the user to authenticate on install (in which case we need to solve scripting these installs for CI servers through some kind of environment variable).

We also need to make sure that Yarn users don't accidentally publish something publicly.

chicoxyzzy:
Private registry doesn't always need auth toke. For example we access our private registry through corporate VPN.

knksmith57:
^^ Agreed. Allowing for the association with a separate registry to use _per scope_ is sufficient for us (and I suspect many others).

jrop:
Same here, excited to use yarn, but this is a blocker for us

jmonster:
> in which case we need to solve scripting these installs for CI servers through some kind of environment variable

+ http://blog.npmjs.org/post/118393368555/deploying-with-npm-private-modules
+ https://circleci.com/docs/npm-private-module-dependency/

michaelbpaulson:
+1

djMax:
When we say "we already have this logic" - I don't see any path where an Authorization header would be sent to a registry. If there was one, perhaps there'd be a temporary workaround to make this all work while something more final is sorted out. Am I missing something?

djforth:
+1 looks like scoped packages even if they are public seem to fail.

frederickfogerty:
To further @djforth's comment, I just installed from master, and I'm getting the same error - scoped packages are failing. It converts the `/` in the package name into `%2f`, which means the request to npm to find the package fails. 

[...]

PR #712, opened 2016-10-11
Custom Registry url.resolve Issue

jonsharratt:
**Summary**
Ensures with and without trailing / registry url is correct and resolvable.

Currently if you set a npm registry via config to be:

1. `yarn config set registry https://mycustom.registry.com/dev/registry`
2. `yarn add redux` (not in cache / fresh install)

The `url.resolve` will turn the url to be `https://mycustom.registry.com/dev/redux` (no `registry` as part of the path due to `/` being removed `removeSuffix`.)

**Test plan(s)**

******
1. `yarn config set registry https://mycustom.registry.com/dev/registry`
2. `yarn add react`

******
1. `yarn config set registry https://mycustom.registry.com/dev/registry/`
2. `yarn add redux`

******
1. `yarn config set registry https://registry.npmjs.org/`
2. `yarn add react-router`

### Note:
There appears to be a flow issue with `url.format` currently:

https://github.com/facebook/flow/issues/908
https://nodejs.org/api/url.html#url_url_format_urlobject

PR #839, opened 2016-10-12
Support npmrc for private registries and auth

### LINKED ISSUE #606: Yarn doesn't read from .npmrc Files for custom registry settings

devongovett:
**Summary**

This implements support for private registries configured in `.npmrc` files. It extracts the registry to use and the auth token info. Supports bearer tokens and basic auth, scoped registries, and the `always-auth` option. Should fix #606.

My flow experience is minimal, so if I messed something up or did something silly please let me know. 😀

**Test plan**

I tested with several npm configs against real live Sonatype Nexus, and Artifactory instances. Here are some cases:

```
# Default registry bearer token from npm
//registry.npmjs.org/:_authToken=TOKEN

# Default registry, basic auth token, always-auth turned on
registry=https://my-registry.com/
_auth=TOKEN
email=test@example.com
always-auth=true

# Scoped registry
@coralui:registry=https://my-registry.com/
//my-registry.com/:_password=PASSWORD
//my-registry.com/:username=USERNAME
//my-registry.com/:email=test@example.com
//my-registry.com/:always-auth=true
```

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -7,6 +7,7 @@ import type {ConfigRegistries} from './index.js';
 import * as fs from '../util/fs.js';
 import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import Registry from './base-registry.js';
+import {addSuffix} from '../util/misc';
 
 const defaults = require('defaults');
 const userHome = require('user-home');
@@ -49,15 +50,13 @@ export default class NpmRegistry extends Registry {
   }
 
   request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<*> {
-    const registry = this.getRegistry(pathname);
+    const registry = addSuffix(this.getRegistry(pathname), '/');
+    const requestUrl = url.resolve(registry, pathname);
 
     const headers = {};
-    if (this.token || this.getOption('always-auth')) {
+    if (this.token || (this.getOption('always-auth') && requestUrl.startsWith(registry))) {
       headers.authorization = this.getAuth(pathname);
     }
-
-    // $FlowFixMe : https://github.com/facebook/flow/issues/908
-    const requestUrl = url.format(`${registry}/${pathname}`);
 
     return this.requestManager.request({
       url: requestUrl,
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -157,14 +157,15 @@ export default class NpmRegistry extends Registry {
     }
 
     for (let registry of [this.getRegistry(packageName), '', DEFAULT_REGISTRY]) {
-      // Check for auth token.
       registry = registry.replace(/^https?:/, '');
 
+      // Check for bearer token.
       let auth = this.getScopedOption(registry, '_authToken');
       if (auth) {
         return `Bearer ${String(auth)}`;
       }
 
+      // Check for basic auth token.
       auth = this.getScopedOption(registry, '_auth');
       if (auth) {
         return `Basic ${String(auth)}`;
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -2,11 +2,11 @@
 
 import type RequestManager from '../util/request-manager.js';
 import type {RegistryRequestOptions, CheckOutdatedReturn} from './base-registry.js';
-import type Config, {ConfigRegistries} from '../config.js';
+import type Config from '../config.js';
+import type {ConfigRegistries} from './index.js';
 import * as fs from '../util/fs.js';
 import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import Registry from './base-registry.js';
-import {removeSuffix} from '../util/misc.js';
 
 const defaults = require('defaults');
 const userHome = require('user-home');
@@ -48,7 +48,7 @@ export default class NpmRegistry extends Registry {
     return name.replace('/', '%2f');
   }
 
-  request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<?Object> {
+  request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<*> {
     const registry = this.getRegistry(pathname);
 
     const headers = {};
@@ -138,10 +138,11 @@ export default class NpmRegistry extends Registry {
     return !packageName || packageName[0] !== '@' ? '' : packageName.split(/\/|%2f/)[0];
   }
 
-  getRegistry(packageName: ?string): string {
+  getRegistry(packageName: string): string {
     // Try scoped registry, and default registry
     for (const scope of [this.getScope(packageName), '']) {
-      const registry = this.getScopedOption(scope, 'registry') || this.registries.yarn.getScopedOption(scope, 'registry');
+      const registry = this.getScopedOption(scope, 'registry')
+                    || this.registries.yarn.getScopedOption(scope, 'registry');
       if (registry) {
         return registry;
       }
@@ -150,7 +151,7 @@ export default class NpmRegistry extends Registry {
     return DEFAULT_REGISTRY;
   }
 
-  getAuth(packageName: ?string): ?string {
+  getAuth(packageName: string): string {
     if (this.token) {
       return this.token;
     }
@@ -173,14 +174,14 @@ export default class NpmRegistry extends Registry {
       const username = this.getScopedOption(registry, 'username');
       const password = this.getScopedOption(registry, '_password');
       if (username && password) {
-        return 'Basic ' + new Buffer(username + ':' + new Buffer(password, 'base64')).toString('base64');
+        return 'Basic ' + new Buffer(username + ':' + new Buffer(password, 'base64').toString()).toString('base64');
       }
     }
 
-    return null;
+    return '';
   }
 
-  getScopedOption(scope: ?string, option: string): string {
-    return this.getOption(scope + (scope ? ':' : '') + option);
+  getScopedOption(scope: string, option: string): string {
+    return String(this.getOption(scope + (scope ? ':' : '') + option));
   }
 }
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -52,9 +52,11 @@ export default class NpmRegistry extends Registry {
   request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<*> {
     const registry = addSuffix(this.getRegistry(pathname), '/');
     const requestUrl = url.resolve(registry, pathname);
+    const alwaysAuth = this.getScopedOption(registry.replace(/^https?:/, ''), 'always-auth')
+                    || this.getOption('always-auth');
 
     const headers = {};
-    if (this.token || (this.getOption('always-auth') && requestUrl.startsWith(registry))) {
+    if (this.token || (alwaysAuth && requestUrl.startsWith(registry))) {
       headers.authorization = this.getAuth(pathname);
     }
 
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -14,7 +14,7 @@ const path = require('path');
 const url = require('url');
 const ini = require('ini');
 
-const DEFAULT_REGISTRY = '//registry.npmjs.org/';
+const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 
 function getGlobalPrefix(): string {
   if (process.env.PREFIX) {
@@ -65,7 +65,9 @@ export default class NpmRegistry extends Registry {
       body: opts.body,
       auth: opts.auth,
       headers,
-      json: true,
+      json: !opts.buffer,
+      buffer: opts.buffer,
+      process: opts.process,
       gzip: true,
     });
   }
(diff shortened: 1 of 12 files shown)


karol-f:
+1

jbt:
I haven't had the chance to check this thoroughly but as one of the people who spotted the [bearer token leak in npm](https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/CHANGELOG.md#v383-2016-03-17) I thought I'd chime in early to make sure something similar doesn't creep in here.

From my super-quick testing on this PR it looks like if `always-auth=true` is present, the bearer token will get sent on every request with the token for the default registry (if present), rather than just those requests to the registry itself.

For example, if my `.npmrc` looks like this:

```
always-auth=true
//registry.npmjs.org/:_authToken=foo
```

... and I install an arbitrary tarball from somewhere other than `registry.npmjs.org` (e.g. `yarn add http://localhost/foo.tgz`), then `Bearer foo` gets sent to that other server. Which is Not Good.

This is just based on my quick 2 mins of testing so feel free to tell me I'm talking rubbish if I've missed something.

cpojer:
Wow this is awesome. Nice work @devongovett and good to see you again. I haven't heard from you since the old MooTools/jQuery days!

zetoke:
>Should fix #606.

I think, this PR is also should fix #521
Or not?

jbt:
On reflection, I've just realised this may actually be the intended behaviour of `always-auth`, judging by [tests in npm](https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/test/tap/map-to-registry.js#L142). If so, do please just ignore most of what I said above.

p0wl:
this does not fix #521, because the token is not read when running `yarn install` (there is no `getToken()` call in `install` nor `info`).

mrtnbroder:
Great work!

However, still having some issues with this, where the `requestUrl` is sometimes mixed up like this:
`https://registry.yarnpkg.com/https://registry.npmjs.org/`

Also, one major problem I still see is that the saved `yarn.lock` saves the full resolved url with credentials in it, which is a no-no.

e.g:

```
# yarn lockfile v1
"@ourpackagescope/PACKAGENAME@VERSION":
  version ...
  resolved "https://NPM_USERNAME:NPM_PASSWORD@OUR_PRIVATE_REGISTRY.com/@ourpackagescope/PACKAGENAME/-/@ourpackagescope/PACKAGENAME.....''
  dependencies:...
```

donaldpipowitch:
Awesome PR. Does this pick up `strict-ssl=false` settings, too?

devongovett:
@jbt yes, that is the desired behavior of `always-auth`. I tested with a registry that requires auth for the tarballs, for example.

@zetoke yes, looks like it should fix #521.

@mrtnbroder does it work if you add a trailing slash to the registry url in your config? If so, I guess we could add the slash if it's not there...

@donaldpipowitch no.

devongovett:
Fixed an issue caused by a later PR #712, which changed from `url.resolve` to `url.format`. We need resolve in the case where the pathname is really a full url (when requesting tarballs). So only add the trailing / if it is not there already. Should fix your issue @mrtnbroder.

@zetoke Also added a check to ensure that we only add authorization if the request is going to the registry (not random tarballs).

also:
This finally works with my `~/.npmrc`, which looks like

```
registry = https://npm.example.com/nexus/content/groups/npm-all/
always-auth = true
_auth = basic-auth-string
```

🎉

sohara:
In our case this allows yarn to resolve the packages in our private registry but when it actually goes to fetch them we get errors like:
`error https://registry.npmjs.org/@condenast/x-package/-/x-package-2.2.0.tgz: invalid tar file`

devongovett:
@sohara can you provide your `.npmrc` file?

sohara:
Sure
```
@condenast:registry=https://registry.npmjs.org/
//registry.npmjs.org/:_authToken=very-secret-code
always-auth=true
progress=false
python=python2.7
```

jgoz:
> In our case this allows yarn to resolve the packages in our private registry but when it actually goes to fetch them we get errors like:
> error https://registry.npmjs.org/@condenast/x-package/-/x-package-2.2.0.tgz: invalid tar file

I'm getting the same error with our `sinopia` proxy/registry:

```
error https://path-to-sinopia/my-package/-/my-package-1.2.3.tgz: invalid tar file
```

`.npmrc`:

```
registry=https://path-to-sinopia/
always-auth=true
_auth=some-secret-key
```

devongovett:
@jgoz @sohara can you verify that the tar file exists and actually is valid? Can you debug? Try putting a log [here](https://github.com/yarnpkg/yarn/pull/839/files#diff-b053bee294c216269844e5874039b6caR60) to see what the requestUrl and headers.authorization are.

jgoz:
@devongovett 

I did some debugging and I think I know how to fix it on my end, but there may be an issue when reading `.npmrc` files.

The issue is that I have `always-auth=true` in my local `.npmrc` (in the repository) and `always-auth=false` in my user `.npmrc`. This would be fine if the local file was loaded last, but the files are being loaded in this order:

- User
- Local
- Local
- User

So the final value for `always-auth` is `false`, which skips adding the necessary header. I can remove this config value from my user-level `.npmrc`, but the config files should only be read a single time (and in global-user-local order).

jgoz:
@devongovett As a follow-up, I think the issue I'm seeing (can't speak for @sohara) is not related to this PR and should be fixed separately. I'll log an issue.

devongovett:
@jgoz yes, that sounds like a separate issue.

sohara:
@devongovett I've added some debug logging as you've suggest and and I'm getting the following on stdout (logging first the header followed by the request url):
```
yarn install v0.15.7
info No lockfile found.
[1/4] Resolving packages...
headers.authorization  Bearer blah-secret-code-xxx-xxxxx
requestUrl  https://registry.npmjs.org/@condenast%2fcn-brands
... # other stuff
[2/4] Fetching packages...
headers.authorization  undefined
requestUrl  https://registry.npmjs.org/@condenast/cn-brands/-/cn-brands-4.2.2.tgz
```

I'm not sure where tarballs would be stored on the system in order to verify they are ok? I did not find any in `~/.yarn-cache`. Would they be in a tpm dir or something?

zetoke:
@cpojer What is blocking thing for merge this PR to upstream?

src/registries/npm-registry.js
```diff
@@ -46,24 +49,24 @@ export default class NpmRegistry extends Registry {
...
-  request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<?Object> {
-    const registry = removeSuffix(String(this.registries.yarn.getOption('registry')), '/');
+  request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<*> {
+    const registry = addSuffix(this.getRegistry(pathname), '/');
+    const requestUrl = url.resolve(registry, pathname);
 
     const headers = {};
-    if (this.token) {
-      headers.authorization = `Bearer ${this.token}`;
+    if (this.token || (this.getOption('always-auth') && requestUrl.startsWith(registry))) {
```
devongovett: You're right. Added in 35d617e5871cccd8a26aa5ff19622d2b7912f6b5.

cpojer:
@zetoke @kittens and I are currently catching up on the repo (while also holding the fort internally at FB :D ). The post launch craziness made everything a bit unstable. Please give us some time until everything is settled down – if you'd like to help get the tests in a good state again, please send us a PR.

zetoke:
@cpojer of course! Thanks a lot for your work!
It is the next step for javascript ecosystem and tooling!

bcoe:
@devongovett love this approach! @kittens, @bestander what would we need to do to move this forward? Perhaps we can come to a consensus around [this RFC](https://github.com/yarnpkg/rfcs/pull/5)? and get this out the door 💯

kittens:
Can you please rebase this against `master`, I'll give it a final review then we can get this in 😄

devongovett:
@kittens done.

kittens:
This is awesome, thank you so much @devongovett! Extremely happy and impressed with the code, didn't have any nits or comments.

kribblo:
Any chance this is getting published to NPM registry soonish?

bestander:
we'll cut the branch Monday-Tuesday

SEAPUNK:
I installed yarn from the master branch just now, but it doesn't seem to be working for me. It seems to fetch the metadata correctly, but I get the `invalid tar file` error:

```
$ yarn install
yarn install v0.15.1
info No lockfile found.
[1/4] Resolving packages...
warning mini-copromise@3.0.1: Use the `minico` package instead: It's the same thing!
[2/4] Fetching packages...
error An unexpected error occured, please open a bug report with the information provided in "/home/ivan/code/xbpf/packages/manager-frontend/yarn-error.log".
```

`yarn-error.log`: https://gist.github.com/SEAPUNK/84025e425c858dfefb6eaed7b97207b0

My .npmrc looks like this:

```
registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=the-long-token
```

motss:
@SEAPUNK Is `registry="https://registry.npmjs.com//"` with double slashes at the end going to work?

SEAPUNK:
@motss That just throws an early error:

```
Trace:
  TypeError: Cannot use 'in' operator to search for '^6.16.0' in undefined
      at /usr/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:62:24
      at next (native)
      at step (/usr/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /usr/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
      at Promise.F (/usr/lib/node_modules/yarn/node_modules/core-js/library/modules/_export.js:35:28)
      at /usr/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12
      at Function.findVersionInRegistryResponse (/usr/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:72:7)
      at /usr/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:89:34
      at next (native)
      at step (/usr/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
```

@devongovett 

```
Trace:
  Error: https://registry.yarnpkg.com/@xbpf%2fsuppliers: Not found
      at Request.params.callback [as _callback] (/usr/lib/node_modules/yarn/lib/util/request-manager.js:294:18)
      at Request.self.callback (/usr/lib/node_modules/yarn/node_modules/request/request.js:187:22)
      at emitTwo (events.js:106:13)
      at Request.emit (events.js:191:7)
      at Request.<anonymous> (/usr/lib/node_modules/yarn/node_modules/request/request.js:1048:10)
      at emitOne (events.js:96:13)
      at Request.emit (events.js:188:7)
      at IncomingMessage.<anonymous> (/usr/lib/node_modules/yarn/node_modules/request/request.js:969:12)
      at emitNone (events.js:91:20)
      at IncomingMessage.emit (events.js:185:7)
```

SEAPUNK:
@devongovett Wait, now it's giving me that error with the original npmrc. One sec, let me test a few things...

zetoke:
@SEAPUNK I had the same trouble. But with fix from @devongovett it's working!

SEAPUNK:
Yeah, that fix didn't work. `@xbpf/suppliers` is a private package.

SEAPUNK:
I installed yarn 0.16.0 just now, same problems.

SEAPUNK:
Not to mention that removing that first line from my npmrc breaks npm's install process.

devongovett:
@SEAPUNK it's a bug. I will look at it when I have a chance.

SEAPUNK:
Alright, thanks.

kribblo:
Also getting the `invalid tar file` error with 0.16.0:

The `.npmrc` contains something like this, working with npm:

    @mycompany:registry=https://npm.mycompany.com/
    //npm.mycompany.com/:_authToken="COMPANYTOKEN"

And the private registry replies with "an invalid tar file" in the form of a JSON response:

```json
{
    "error": "unregistered users are not allowed to access package @mycomany/company-package"
}
```

This pull request was supposed to add support for scoped registrys (which we use heavily), so is it the `_authToken` that's missing maybe (it's not in the examples)?

fastest963:
I'm having the same issue as @kribblo. When doing a tcpdump on the sinopia box I see that it sends the authorization token with the GET call to the package metadata (`@levenlabs/js-lib` in this case):
```GET /@levenlabs%2fjs-lib HTTP/1.1
Connection: close
User-Agent: yarn/0.16.0 npm/? node/v4.4.7 linux x64
authorization: Bearer $token
accept-encoding: gzip, deflate
accept: application/json
```

But then when it fetches the tar it doesn't:
```GET /@levenlabs%2fjs-lib/-/js-lib-0.3.24.tgz HTTP/1.1
Connection: close
User-Agent: yarn/0.16.0 npm/? node/v4.4.7 linux x64
accept-encoding: gzip, deflate```

**Comment by rexxars** [comment]
Experiencing some weird issues.

My npmrc looks something like this:
```
@somescope:registry=https://npm.somescope.no
//npm.somescope.no/:_authToken=someSecretToken
//registry.npmjs.org/:_authToken=otherSecretToken
```

I went into `request-manager.js` and added a `console.log` of outgoing requests. Upon running `yarn` with modules both hosted on the private registry and the public, I'm seeing it use the public token instead of the private one:

```
{ url: 'https://npm.somescope.no/@somescope%2fsomemodule',
  headers: 
   { authorization: 'Bearer otherSecretToken' },
```

rexxars:
It seems to indeed be related to a "missing" trailing slash, [here](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L156).

If I add a trailing slash to the first entry in the array above, the resolving succeeds, though the tarball download fails (no token is sent).

It's a bit worrying that it seems to fall back to sending a private registry token to the public registry, however.

vinngn1:
Hi! How do you get to install version 0.16.0?
When I do `brew install yarn`, I only get version 0.15.1

Gpx:
@vinngn1 `brew update && brew upgrade yarn`

vinngn1:
Able to install private package out of the box after the upgrade. Thank @Gpx for the upgrade instruction and @devongovett and others for the fix.

motss:
After upgrading to `yarn@0.16.0`, the `.yarnrc` breaks things unexpectedly. By removing the `.yarnrc` and depend solely on `.npmrc` will still be able to install both private and public dependencies without any issue. Ditching `.yarnrc` now~

KidkArolis:
I didn't read all of this thread, but I was wondering - howcome nobody ran into the issue where auth header is not getting sent when downloading tarballs: https://github.com/yarnpkg/yarn/issues/1687.

kribblo:
@KidkArolis https://github.com/yarnpkg/yarn/pull/839#issuecomment-254622848 just a few comments above yours

KidkArolis:
@kribblo yeah, thanks :) I take it the issue hasn't been resolved yet. I have a way to reproduce it: https://github.com/KidkArolis/yarn-scopes-issue. And a potential fix: https://github.com/yarnpkg/yarn/pull/1666.

Issue #513, opened 2016-10-05
git+ssh package install does not seem to work


Fishrock123:
**OP's Note: if you are also having this EXACT problem, please upvote this WITHOUT commenting.**

---

**Do you want to request a _feature_ or report a _bug_?**

bug

**What is the current behavior?**

```
yarn install v0.14.0
info No lockfile found.
[1/4] 🔍  Resolving packages...
error Couldn't find package "<package>" on the "npm" registry.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
"devDependencies": {
    "license-builder": "git+ssh://git@github.com/fishrock123/<package>.git",
}
```

**What is the expected behavior?**

typical install

**Please mention your node.js, yarn and operating system version.**

Node.js: v6.6.1-pre
Yarn: v0.14.0 (`master`)
OS: OSX 10.10.5

danharper:
Minimum repro:

```
{
  "name": "x",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "devDependencies": {
      "eslint-config-radweb": "git+https://git@github.com/radweb/eslint-config-radweb.git"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

The package isn't on the registry, if that makes a difference.

BBB:
This also errors when specifying a git tag (which npm allows)
```JSON
...
  "react-quill": "git+https://git@github.com/alexkrolick/react-quill.git#v2.0.1",
...
```

adambowles:
I also get this #621

JamesHenry:
Just adding in case subtlety different/requires an additional solution to @BBB case of git tag:

I am trying to install a _specific commit hash_ with git+ssh. NPM default client supports this.

BBB:
Looks like #633 and #639 are related

mko:
For context, from [NPM's docs](https://docs.npmjs.com/cli/install):

> `npm install <git remote url>:`
> 
> Installs the package from the hosted git provider, cloning it with git. First it tries via the https (git with github) and if that fails, via ssh.
> 
> `<protocol>://[<user>[:<password>]@]<hostname>[:<port>][:][/]<path>[#<commit-ish>]`
> 
[...]

PR #843, opened 2016-10-12
Allow usernames in git resolver

danharper:
**Summary**

The Git resolver did not understand usernames in URLs, e.g. `git+ssh://git@github.com/...` due to it treating any pattern containing `@` as being "un-exploded". See #513.

For me it was falling through, and calling the npm resolver instead. The package wasn't published on the registry, so it was erroring there. For others, it would error due to version mismatches if it was on the registry.

Instead, it now checks for `@PROTOCOL:`, e.g.

```
// good
git+https://git@github.com/radweb/eslint-config-radweb.git

// bad
eslint-config-radweb@git+https://git@github.com/radweb/eslint-config-radweb.git
                    ^^^^^^^^^^^
```

**Test plan**

Added test, they pass.

Btw I've noticed `https://github.com/npm-ml/ocaml.git#npm-4.02.3` times out a _lot_.


danharper:
I've had chance to look into this more, and found what I think is a more elegant solution. I've opened another PR for that version here: #934.

PR #848, opened 2016-10-12
cli: make mutex's type a mandatory argument

thefourtheye:
**Summary**

Without making `mutex`'s `type` argument a mandatory one, by default,
`true` is assumed and it fails with the following error.

    ➜  yarn git:(master) ./bin/yarn --mutex
    yarn install v0.15.1
    error TypeError: mutex.split is not a function
        at config.init.then (yarn/lib/cli/index.js:336:25)

And if just `mutex` is specified, commander treats that as an array and
fails like this

    ➜  yarn git:(master) ./bin/yarn mutex
    yarn mutex v0.15.1
    error TypeError: mutex.split is not a function
        at config.init.then (yarn/lib/cli/index.js:336:25)

This patch makes the `type` a mandatory parameter. So, now

    ➜  yarn git:(fix-mutex-cmd-args) ./bin/yarn --mutex

      error: option `--mutex <type>[:specifier]' argument missing

Also, simply typing `mutex` will be ignored.

    ➜  yarn git:(fix-mutex-cmd-args) ./bin/yarn --mutex
    yarn mutex v0.15.1
    error No command specified.
    ...
    question Which command would you like to run?:

**Test plan**

Included in summary itself

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -54,8 +54,9 @@ commander.option(
   'rather than storing modules into a global packages root, store them here',
 );
 commander.option(
-  '--mutex [type][:specifier]',
+  '--mutex <type>[:specifier]',
   'use a mutex to ensure only one yarn instance is executing',
+  ''
 );
 commander.allowUnknownOption();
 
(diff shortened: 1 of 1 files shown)


kittens:
Thanks!

Issue #824, opened 2016-10-12
Typing only 'yarn'


shrynx:
**Do you want to request a *feature* or report a *bug*?**
Feature, well it is an enhancement. 
**What is the current behavior?**
When i type yarn for first time on console, the output of typing just yarn on a new directory is too scary.
<img width="831" alt="screen shot 2016-10-12 at 4 27 58 pm" src="https://cloud.githubusercontent.com/assets/4706261/19301309/ba42fa46-9099-11e6-815e-26e7e7181797.png">

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
Well decrease in verbosity and also automatically output ```yarn --help``` commands

jurosh:
as per docs, this is default behavior to install packages by just typing `yarn`

beshkenadze:
Something like that: https://github.com/yarnpkg/yarn/pull/852

mikaoelitiana:
Another approch different from @beshkenadze's #856

gwicksted:
I think it would be **more helpful** to `yarn init` on a new folder when `yarn` is executed.  Perhaps a message like:

> no package.json or bower.json found, would you like to create a new package file?

then continue with the usual yarn init process.

What else are you going to do on a folder without either one of those files?  Either it's an "oops, wrong folder" or "yeah... init -- that's what I meant!"

jurosh:
@gwicksted how often are you doing `yarn init`? it's probably not so... proposed `init` alias may only create many not needed packages.json in files where you accidentally runned the command.. much better would be to sugges that user may run yarn init to create new package.json...

gwicksted:
@jurosh you're right.  More likely a mistake.

kittens:
Just as a heads up to everyone. `yarn` being an alias for `yarn install` is intentional. We should make this error message nicer but we'll still be showing it.

SamSaffron:
bundler does 

```
sam@ubuntu Source % bundle
Could not locate Gemfile or .bundle/ directory
```

so something along that seems like a good thing to copy

busches:
The error message has been improved in `0.16.0` by #1097
```
yarn
yarn install v0.16.1
info No lockfile found.
error Couldn't find a package.json (or bower.json) file in C:\Users\buschs1\Desktop
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

cc\ @kittens

wyze:
[...]

PR #852, opened 2016-10-12
Show the help instead of error

beshkenadze:
**Summary**
Show the help instead of error. 
https://github.com/yarnpkg/yarn/issues/824

**Test plan**
Just changed the default command.

image


kittens:
The existing behaviour is intentional. `yarn` is supposed to act as an alias for `yarn install`, you can see #824 for my explanation.

PR #856, opened 2016-10-12
hide verbose error message when no manifest found

### LINKED ISSUE #824: Typing only 'yarn'

mikaoelitiana:
yarn command will not show verbose message when no package.json or
bower is found

fixes #824

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -327,7 +327,8 @@ export default class Config {
         }
       }
 
-      throw new Error(`Couldn't find a package.json (or bower.json) file in ${dir}`);
+      this.reporter.error(this.reporter.lang('noManifestFound', dir));
+      return {};
     });
   }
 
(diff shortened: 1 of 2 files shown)


kittens:
We use this in a bunch of places and this validation is really important. I think it'd be best if we suppressed this error when we look for the root manifest in `cli/commands/install.js`.

mikaoelitiana:
I see the problem is at the end when executing `await executeLifecycleScript(config, 'prepublish');`. Should we check the existence of root manifest before executing this then?

kittens:
Can you please rebase

mikaoelitiana:
This seems to be fixed in 0.16.1

Issue #858, opened 2016-10-12
Add --online option to upgrade (and add) to skip offline cache


FokkeZB:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
I can use `--offline` to force using cache, but I cannot do the opposite.

**What is the expected behavior?**
I'd like to be able to skip cache, e.g. when I depend directly on a GitHub repository and I want to get the latest, even though the version hasn't changed.

**Please mention your node.js, yarn and operating system version.**
Node 4.2.3, Yarn 0.15.1, Mac OS Sierra 10.12.1 Beta (16B2333a)

dylanparry:
Does `yarn add <package_name> --force` do this?

FokkeZB:
I dunno, it's not documented in `yarn add --help` and `force` by itself is not really descriptive. Force what?

BTW, if we add `--online` to upgrade then it also needs to be able to upgrade only a single package (like `add`). It doesn't make sense to use `--online` for all of your dependencies.

AlicanC:
`yarn help add` doesn't have a description for `--force`. If it does what @FokkeZB needs, then we can repurpose this issue to add a description to `--force`.

dylanparry:
Although it doesn't have any documentation in `yarn add --help`, there's a similar option for `yarn --force` which forces all packages to be refetched. This (unhelpfully) also isn't documented in `yarn --help`, but there's a description here: https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-force

Oddly, there's no entry for the `--force` flag on the page for `yarn add`, but it definitely exists.

FokkeZB:
Closing as #870 documents force, which does what this issue asks for.

PR #870, opened 2016-10-12
Document --force

FokkeZB:
**Summary**

Documents `install/add --force`, which does what #858 asks for.

**Test plan**

Run `yarn add --help` or `yarn install --help` and `--force` should be documented as `refetch all packages`


kittens:
`--force` skips our integrity cache and forces execution of install scripts so it does a few more things. I think it'd be more accurate to summarise the description as "ignore all caches"

FokkeZB:
@kittens I changed it, although I must say I find the option name itself not very descriptive. Force what? It's also confusing that you have `--offline`, `--prefer-offline` and then to get the opposite you need `--force`.

PR #871, opened 2016-10-12
fallback to npm config if it exists after getting default yarn config value

kittens:


tyrsius:
Is there a way to tell when this has been released?

src/registries/yarn-registry.js
```diff
@@ -56,7 +56,15 @@ export default class YarnRegistry extends NpmRegistry {
   homeConfig: Object;
 
   getOption(key: string): mixed {
-    return this.config[key] || this.registries.npm.getOption(npmMap[key] || key);
+    let val = this.config[key] || this.registries.npm.getOption(npmMap[key]);
+
+    // if we have no yarn option for this or have used a default then use the npm
+    // value if it exists
+    if (!val || val === DEFAULTS[key]) {
```
whitelynx: With this code, if the _value_ is the same as the default, it will favor the npm config even if the user explicitly set it. Shouldn't an explicit setting still take precedence over the npm config value?

whitelynx: Also, we might want to check if it's `undefined` instead of just falsy, since the user could also explicitly specify a falsy value.

evil-shrike:
I'm afraid it's still not working - I'm trying to install a scoped package with a scope registered in my global `.npmrc`. It doesn't work.
You'd better not to close #606 - it's still actual.

ysangkok:
@evil-shrike That issue was closed. Can you provide a more detailed example? Maybe it is a different issue.

FezVrasta:
I was using `.yarnrc` to workaround the problem, but today I noticed the content of `.yarnrc` has been replaced with:

```
# yarn lockfile v1


lastUpdateCheck 1488992854257

```

So, now, where do I put my config if Yarn doesn't read from `.npmrc`?

floatdrop:
@kittens any update on this? Seems logical to read configs from `.npmrc`, if they are available.

thescientist13:
also curious when the version with this fix (looks to be [0.24.2](https://github.com/yarnpkg/yarn/commit/5d1704d3b5be0983f58bbdae3002dd560637cba1)) will get released / published to NPM?

sarbbottam:
Is there any update when this fix will get rolled out? I can still find the issue with `0.24.4`.

bchoddny:
I've got 0.24.5, this issue seem to persist.

forelabs:
@bchoddny @sarbbottam have a look if you have this issue with the fallback: https://github.com/yarnpkg/yarn/issues/3920
this was the issue we had to struggle with.

olleicua:
I'm experiencing this in 1.3.2

```
$ yarn publish
[1/4] Bumping version...
question New version: 0.1.0
[2/4] Logging in...
question npm username: USERNAME
question npm email: USERNAME@COMPANY.com
question npm password:
success Logged in.
error Couldn't publish package.
info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

`npm publish` works though

a-hoff:
This still does not work for scoped registries

euaaron:
I'm trying to add a package from a scoped registry that asks for password, username and email. I am able to do this using NPM by setting this up to `.npmrc`, but I can't find a way to do this with YARN by adding the same config from to `.yarnrc`.

While searching for solution, I've found issue [#606](https://github.com/yarnpkg/yarn/issues/606) and also this issue from 2016. 
I guess after about 7 years that this feature might be already incorporated to Yarn, right?

If so, could anyone tell me if this is working and what is the right semantic?

This is what I've added to `.yarnrc`:

```sh
registry "https://registry.yarnpkg.com"
"@company:registry" "http://company.domain.com/artifactory/api/npm/packages/"
"//company.domain.com/artifactory/api/npm/packages/:_password" "passwordHASH"
"//company.domain.com/artifactory/api/npm/packages/:username" "me"
"//company.domain.com/artifactory/api/npm/packages/:email" "me@company.domain.com"
"//company.domain.com/artifactory/api/npm/packages/:always-auth" "true"
```

It understand the scoped registry but it fails to download the package due to not finding any credentials.

weivegywei:
We tried with`.yarnrc.yml`
`npmScope:
'company': 
  npmAlwaysAuth: true
  npmAuthToken: 'YOUR_PERSONAL_ACCESS_TOKEN'
  npmRegistryServer: https//registry.com`
  yarn add works

PR #880, opened 2016-10-12
add cache dir command to output path to cache directory

kittens:


cpojer:
why is this `ls`? Is there something more appropriate? Also, if you add features, please create an issue about it on the website repo too so somebody can take care of it and we don't end up with undocumented cli features: https://github.com/yarnpkg/website/issues

cpojer:
(Approved because I trust you to find the best name).

kittens:
Oops, was meant to be `dir`.

chicoxyzzy:
this causes flow erros
```
src/cli/commands/cache.js:35
 35:     console.log(config.packagesRoot);
                            ^^^^^^^^^^^^ property `packagesRoot`. Property not found in
 35:     console.log(config.packagesRoot);
                     ^^^^^^ Config


Found 1 error
```

PR #881, opened 2016-10-12
Rename packagesRoot option to cacheFolder

kittens:


cpojer:
that's so much clearer.

Issue #893, opened 2016-10-12
yarn install --flat only installs packages that have conflicts


bouk:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Consider the following package.json (and no yarn.lock):

```
{
  "dependencies": {
    "jquery": "3.1.1",
    "moment": "2.11.2",
    "moment-timezone": "0.5.0"
  },
  "resolutions": {
    "moment": "2.11.2"
  }
}
```

If I run `yarn install --flat` (or add the `"flat": true` field) I would expect it to properly install all the packages on a single level. Instead what happens is that _only_ the moment package is installed, with `moment-timezone` and `jquery` missing from the `node_modules` folder. It does properly create the following yarn.lock file, though:

```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
jquery@3.1.1:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.1.1.tgz#347c1c21c7e004115e0a4da32cece041fad3c8a3"

moment-timezone@0.5.0:
  version "0.5.0"
  resolved "https://registry.yarnpkg.com/moment-timezone/-/moment-timezone-0.5.0.tgz#f282c8e5e6e28614b2f6dd1048663754d30ea11f"
  dependencies:
    moment ">= 2.6.0"

"moment@>= 2.6.0", moment@2.11.2:
  version "2.11.2"
  resolved "https://registry.yarnpkg.com/moment/-/moment-2.11.2.tgz#87968e5f95ac038c2e42ac959c75819cd3f52901"

```

**What is the expected behavior?**

Everything is installed

**Please mention your node.js, yarn and operating system version.**

Node v6.7.0
Yarn v0.15.1
macOS 10.12

bouk:
Already found the issue, PR on the way

PR #894, opened 2016-10-12
Add single resolved version to flattenedPatterns instead of patterns

bouk:
**Summary**

Fixes https://github.com/yarnpkg/yarn/issues/893

The single resolved version was accidentally being pushed to the passed in argument instead of the result. I have tested the case presented in the issue, and it works locally.


kittens:
Thank you!

Issue #885, opened 2016-10-12
bower_components deleted when installing node_modules


sb8244:
**Do you want to request a *feature* or report a *bug*?**

This is a bug.

**What is the current behavior?**

When installing node_modules from yarn, with an existing bower_components directory (custom through .bowerrc), the bower_components directory is cleared.

When re-running yarn with already installed packages, nothing is deleted.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Have package.json and bower.json in same directory
2. .bowerrc

```
{
  "directory": "bower_components"
}
```

3. Populate bower with `bower install`
4. Run `yarn`
5. `node_modules` exists but `bower_components` is empty.

**What is the expected behavior?**

`bower_components` should not be touched, because it is installed correctly.

**Please mention your node.js, yarn and operating system version.**

```
➜ yarn --version
0.15.1
➜ node --version
v4.3.2
➜ sw_vers -productVersion
10.11.6
```

alextreppass:
Suspect this is a dupe of #616

sb8244:
Thanks

Issue #864, opened 2016-10-12
bower components not installed automatically from bower.json?


alextreppass:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I have a `bower.json` in the root directory of a project alongside `package.json`.

The `bower.json` contents are:
```
{
  "name": "my-project-name",
  "version": "0.0.0",
  "appPath": "src",
  "resolutions": {
    "angular": "^1.4.8"
  },
  "dependencies": {
    "angular-material": "1.0.3-master-0918596",
    "angular-ui-map": "~0.5.0",
    "angular-tour-no-jQuery": "~0.2.6",
    "ngclipboard": "^1.1.0"
  }
}
```

Should Bower support work automatically?

(The `package.json` has a `postinstall` script that runs `bower install`.  I've seen pre/post-install hooks don't currently get run as per #853, but yarn says it supports bower itself so...)

**If the current behavior is a bug, please provide the steps to reproduce.**
`rm -rf node_modules bower_components`

run `yarn`. 

`node_modules` are populated but `bower_components` are not.

**What is the expected behavior?**
yarn reads the `bower.json` and populates the `bower_components` folder with bower dependencies. 

yarn should also track the bower components in its `yarn.lock` file (related: #846)

**Please mention your node.js, yarn and operating system version.**
node 6.2.2
yarn 0.15.1 (installed globally with npm)
bower 1.7.9 (installed globally with npm)
OSX El Capitan

ankon:
Note here: if yarn were to process `bower.json`, it would be very important for us if it would also honor the configuration file `.bowerrc`, at least the "directory" entry there.

sb8244:
Further on this:

I have a setup with node_modules and bower_components. Upon deleting the `node_modules` directory (but leaving bower_components) and running `yarn`, I get an empty `bower_components` directory.

alextreppass:
@sb8244 you may want #616 for the nixing of `bower_components` contents

RustyToms:
This is a duplicate of #665

omnisite:
If you add the following to package.json, bower install will be called and it works. It is a workaround though:

   ` "scripts": {
        "postinstall": "bower install"
    }`

wyze:
Bower support has been dropped in #1441.

Issue #846, opened 2016-10-12
Yarn completes the bower install but creates a lockfile without versions.


kitsunde:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

I have the following `bower.json`:

```json
{
  "name": "jaguar",
  "version": "0.0.0",
  "private": true,
  "dependencies": {
    "bootstrap": "~3.3.5",
    "devicejs": "2ae5c775e35ccc837589e5af34e292c54936778c",
    "jquery": "2.1.3",
    "jquery-transform": "e195b9a7118558bb1141e50b80380ea5f31dffb8",
    "moment": "2.14.1",
    "moment-timezone": "0.5.5",
    "owl-carousel2": "2.0.0-beta.2.4",
    "raven-js": "3.5.1",
    "ua-parser-js": "0.7.10",
    "underscore": "1.8.3",
    "object-fit": "~0.4.2",
    "picturefill": "^3.0.2",
    "jquery-selectBox": "316c77f157cb25c7a6ea36822143ac9d97845067"
  },
  "resolutions": {
    "jquery": "2.1.3"
  }
}
```

I also have this `.bowerrc`:

```json
{
  "directory": "static/bower_components"
}
```

I run

```bash
$ yarn
yarn install v0.15.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Rebuilding all packages...
success Saved lockfile.
✨  Done in 24.20s.
```

Successive runs fail:

```bash
$ yarn --force
yarn install v0.15.1
[1/4] 🔍  Resolving packages...
error Package "bootstrap@" doesn't have a "version".
info Visit http://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Looking at the lockfile it seems to not add version numbers:

```
bootstrap@~3.3.5:
  version ""
  resolved "https://github.com/twbs/bootstrap.git#0b9c4a4007c44201dce9a6cc1a38407005c26c86"
  registry bower
  dependencies:
    jquery "1.9.1 - 3"
```
<details>
```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
bootstrap@~3.3.5:
  version ""
[...]

Issue #730, opened 2016-10-11
Yarn recursively deletes all bower.json files


OverZealous:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Running `yarn` on an existing project deleted several `bower.json` and `.bower.json` files, which weren't even in the root of the current project.

These were files that existed within the `test` directory, used in unit tests. However, I think it's unacceptable that `yarn` removes _any_ files without warning or even a message on the command line.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Clone this repo: https://github.com/OverZealous/cdnizer
2. Run `yarn` at the root
3. See changes:
```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   yarn.lock

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	deleted:    test/bower/components/angular/bower.json
	deleted:    test/bower/components/no-bower-version/.bower.json
	deleted:    test/bower/components/no-bower-version/bower.json
	deleted:    test/bower/components/no-dot-bower-version/bower.json
````

**What is the expected behavior?**

1. Existing `bower.json` and `.bower.json` files should be left as-is.
2. A warning should be given any time a file is to be removed, especially if it's not within the `node_modules` (or possibly `bower_components`) directory.

**Please mention your node.js, yarn and operating system version.**

- **node -v**: `v5.12.0`
- **yarn --version**: `0.15.1`
- **OS**: `Mac OS X 10.11.6`

oskarhane:
Same here, I was **really** surprised when yarn deleted 2220 files (all bower dependencies in a dir named `components`).
npm dependencies are located in a dir named `npm_modules` and I cannot see why yarn would touch anything outside of that dir.

Ben-Gee:
Same here,
yarn simply removes all files in my bower_components directory.

oskarhane:
The workaround that fixed my issue (I _think_ I had that issue because I have `package.json` and `bower.json` in the same directory) was this:

`yarn` -> `mv package.json p.json` -> `yarn` -> `mv p.json package.json`

The first `yarn` installs npm dependencies and removes bower dependencies and the second one installs bower dependencies.

OverZealous:
[...]

PR #896, opened 2016-10-12
ignore bower for now

RustyToms:
**Summary**
_This PR is a possible short term patch if yarn needs to buy time to fix bower support. It's just a suggestion, and should be considered carefully before merging, if it gets merged at all. I won't be offended if it gets closed :)_

This PR removes bower from the default list of registries. If the concept of this PR is acceptable, I might need some help updating tests. For now I just commented out the ones that seemed to rely on the bower code that is commented out.

Instead of deleting files I commented them out, as this is a short-term solution.

**Why?**
There are quite a few problems associated with bower at the moment:
- bower.json dependencies are not installed
  - #665 
  - #864 
  - #898 
- bower_components is emptied out even if it was properly installed with bower
  - #616 
  - #885 
- bower.json files found inside a bower/components directory are all being deleted
  - #730 
- bower dependencies might be added to the lockfile if there is a custom directory listed in the .bowerrc, but it seems that they are not provided with a version number, causing failures on subsequent updates
  - #846 
- it looks like even if bower installs dependencies in the `bower_components` folder, some of the dependencies of those dependencies are put inside the `node_modules` folder
  - #715 

**Test plan**
- clone https://github.com/RustyToms/yarn-issue-616, a project with one bower and one npm dependency
- `bower install`
- `yarn`
- Before, the `bower_components` file would be emptied after `yarn` was run, with this PR `bower_components` should be untouched

### DIFF
--- a/__tests__/__snapshots__/fetchers.js.snap
+++ b/__tests__/__snapshots__/fetchers.js.snap
@@ -1,0 +1,1 @@
+exports[`test TarballFetcher.fetch throws 1`] = `"https://github.com/PolymerElements/font-roboto/archive/2fd5c7bd715a24fb5b250298a140a3ba1b71fe46.tar.gz: Bad hash. Expected foo but got 9689b3b48d63ff70f170a192bec3c01b04f58f45 "`;
--- a/__tests__/__snapshots__/fetchers.js.snap
+++ b/__tests__/__snapshots__/fetchers.js.snap
@@ -1,1 +1,0 @@
-exports[`test TarballFetcher.fetch throws 1`] = `"https://github.com/PolymerElements/font-roboto/archive/2fd5c7bd715a24fb5b250298a140a3ba1b71fe46.tar.gz: Bad hash. Expected foo but got 9689b3b48d63ff70f170a192bec3c01b04f58f45 "`;
--- a/__tests__/fetchers.js
+++ b/__tests__/fetchers.js
@@ -1,10 +1,10 @@
 /* @flow */
 /* eslint max-len: 0 */
 
-import TarballFetcher, {LocalTarballFetcher} from '../src/fetchers/tarball-fetcher.js';
+// import TarballFetcher, {LocalTarballFetcher} from '../src/fetchers/tarball-fetcher.js';
 import BaseFetcher from '../src/fetchers/base-fetcher.js';
 import CopyFetcher from '../src/fetchers/copy-fetcher.js';
-import GitFetcher from '../src/fetchers/git-fetcher.js';
+// import GitFetcher from '../src/fetchers/git-fetcher.js';
 import {NoopReporter} from '../src/reporters/index.js';
 import Config from '../src/config.js';
 import mkdir from './_temp.js';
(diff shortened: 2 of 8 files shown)


sheerun:
Why such move? As bower maintainer I consider endorsing yarn as an alternative CLI, but if yarn drops support for it, it's out of question.. How about fixing these bugs instead of removing support?

RustyToms:
@sheerun I agree they should be fixed. I couldn't figure out how to do it and I think the project maintainers aren't sure yet either, they weren't aware it was breaking until yesterday. This is a tourniquet to stop the bleeding so it doesn't destroy bower installs.

But if someone can put a PR together to actually fix it that would be better. The idea here is "do no harm", right now someone types in "yarn" and their bower_components file gets emptied, at least every time I have done it. Since there is no documentation for bower support that I could find, I figured it might be better to roll back on that until it gets fixed. Maybe there are some configurations where it isn't happening, it seems to work at least partially for some people. I would +1 a PR that can get it working right!

sheerun:
It might be a challenge to re-introduce Bower support if Yarn drops it. I think it's better to release a patch that fixes most annoying issues (like deleting `bower_components` or `bower.json`), and then improve support in small steps.

I see the potential of this project, as it solves most of the issues Bower currently has (file locking, reliable installs, much better codebase). I'd be glad to work on improving Bower support, so maybe eventually Yarn could become Bower's "next major version". I'll talk with @kittens about it..

RustyToms:
I was going to keep this open until there was some viable alternative, but I do agree with @sheerun about the importance of including bower support, especially if he is supporting it. So I'm going to close this, I guess a maintainer could reopen it if they really wanted to.

thejameskyle:
Yeah, let's push forward one issue at a time. It probably makes sense to let the dust settle. We expected a ton of issues to be opened at launch, we just need to work through them. The problems will get smaller and smaller very quickly.

sombriks:
so, if i understood correctly, bower projects just need keep a fair distance from yarn until it gets updated to ignore bower setups?

ollisal:
Can I even instruct Yarn to skip messing with bower stuff via some command line or configuration option in the meantime?

With close to 10 grave open issues concerning bower support, I think it would've been a good bet to disable bower stuff by default until those issues are sorted. Now it's pretty hard to figure out how/if yarn can be used safely together with bower, or to install the npm dependencies for a project that also uses bower in any way, for that matter.

ollisal:
After some inspection of the source code and such, it looks like it's definitely the best to keep my distance for now - no way to get yarn to leave bower alone.

leizhao4:
I think Yarn should eventually fix Bower support, but at this point it's broken and should be disabled by default until there is a fix. Having nothing is better than having a broken support which deletes my files.

RustyToms:
The easiest work around, if you don't have any .bowerrc files except for the one in the root of your project, appears to be to run `bower install` after you run `yarn`. In my experience the problem of files being deleted only happens the first time you run yarn. Then bower will have to install again. And it should be fine after that.

tallaxes:
> only happens the first time you run yarn

Or maybe anytime packages are installed? `yarn add` also appears to delete files ...

eladchen:
How about allowing the two to exists ? 

As a temporary solution, Why not make 'yarn' move existing 'bower_components' ( if there is one )
to some random folder when an install / add process is ?

e.g: 'bower_components' > 'bower_components_temp_cf45f2fa62cc73f38220' and revert it back once yarn's install flow is done.

Thoughts ?

RustyToms:
Bower support was removed in #1441

Issue #903, opened 2016-10-12
[bug] passing tests time out


RustyToms:
**Do you want to request a *feature* or report a *bug*?**
### Bug

**What is the current behavior?**
Some tests take a long time to run, particularly the `commands/add.js` and `cmmands/install.js` on mac os. These tests may or may not time out, with no code changes

**If the current behavior is a bug, please provide the steps to reproduce.**
Run the test suite several times on mac os

**What is the expected behavior?**
The `commands/add.js` and `cmmands/install.js` will sometimes time out

**Please mention your node.js, yarn and operating system version.**
node.js v6.7.0
yarn v0.15.1
Mac OSX 10.11.6

theduke:
@RustyToms can be closed considering your PR was merged?

PR #905, opened 2016-10-12
increase the timeout for long running tests

RustyToms:
**Summary**
Increases the timeouts for the `commands/add.js` and `commands/install.js` tests
#903 

These tests are passing but occasionally timing out on mac os because they take so long to run

**Test plan**
Run the test suite several times on mac os


kittens:
Thank you!

Issue #778, opened 2016-10-12
yarn search


jdalton:
There's currently an `npm search` which isn't so great. What do you all think of having `yarn search` which is a shortcut for opening `https://npms.io/search?q=<search-string>` using [npms.io](https://npms.io/).

\@atduarte @satazor

tony19:
@jdalton Instead of opening a browser window/tab, I think `yarn search` should list the search results in the CLI perhaps using the [`npms` API](https://api-docs.npms.io/#api-Search-ExecuteSearchQuery).

Daniel15:
>  I think yarn search should list the search results in the CLI perhaps using the npms API. 

I wonder whether that's worth the maintenance cost, given the site already exists. In general, a rich graphical interface like a website is always going to provide a better user experience than a CLI app.

tony19:
> In general, for something like this where you want to search and browse through a listing or collection of things, a rich graphical interface like a website is always going to provide a better user experience than a CLI app.

I think that's somewhat subjective and probably depends on the amount of data presented. Plus, there's already a precedence for CLI-only results (i.e., `bower search` and `npm search`).

kittens:
We've deliberately decided not to add these sorts of commands to Yarn. There's a lot of them that don't actually see much use and the maintenance burden isn't really worth it. If the functionality of the commands is so simple (just opening a browser) then you can just do it yourself. Thanks for the report!

jdalton:
Ya, mainly wanted to show some :heart: for [npms.io](https://npms.io/) and figured yarn integration via opening a browser would help (yarn improving pkg installs & npmjs.io improving search).

jdalton:
Related to https://twitter.com/mrmrs_/status/763481520791232512.

satazor:
@jdalton Thank you it's an honour to have such an important member of the community suggesting npms.io. I tend to agree with @kittens, but if this project gets enough traction to actually "replace" npm (like lodash effectively replaced underscore for most devs), I think it's reasonable to have a search functionality. Many many developers still prefer to do these kind of tasks from the console.

iarna:
FYI, we've documented our new search API (that builds on the npms.io folks work) over [here](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#get-v1search), if you ever decide you want to fold that in.

robclancy:
Lol is this for real? Such a big maintenance burden to dump search results from an API.

mika76:
[...]

PR #914, opened 2016-10-12
yarn home.  fixes #897

### LINKED ISSUE #897: yarn home
feature

Not implemented

`yarn home [pkgname]` uses [opn](https://github.com/sindresorhus/opn) to open the default browser at the URL from the `homepage` key in package.json of the specified package.  Default package is ".".   For extra credit at npm compatibility, provide the mostly-useless `--browser` parameter to select a browser.

node: v6.7.0
yarn: 0.15.1
OS: OSX 10.12

hildjj:
**Summary**

Fixes #897.  Adds a `yarn home` command that mirrors `npm home`

**Test plan**

```bash
yarn home --help
```
```
  Usage: yarn homeFail [flags]

  Options:

    --offline
    --prefer-offline
    --strict-semver
    --json
    --global-folder <path>
    --browser [BROWSER]         Use BROWSER instead of system default
```

```bash
yarn home
```
(opens homepage URL from ./package.json in default browser)

```bash
yarn home .
```
(opens homepage URL from ./package.json in default browser)

```bash
yarn home . --browser 'google chrome'
```
(opens homepage URL from ./package.json in Google Chrome)

```bash
yarn home .
```
```
yarn home v0.15.1
error ENOENT: no such file or directory, open 'package.json'
    at Error (native)
info Visit https://yarnpkg.com/en/docs/cli/home for documentation about this command.
```
(with no package.json file)

```bash
yarn home .
```
```
yarn home v0.15.1
error package does not specify homepage
```
(with no homepage URL in package.json)

```bash
yarn home yarn
```
(opens https://github.com/yarnpkg/yarn#readme in default browser)

```bash
yarn home yarn opn
```
(opens https://github.com/yarnpkg/yarn#readme and https://github.com/sindresorhus/opn#readme in default browser)

```bash
yarn home yarn opn --browser 'google chrome'
```
(opens https://github.com/yarnpkg/yarn#readme and https://github.com/sindresorhus/opn#readme in Chrome)

```bash
yarn home jjjjjjjjjjjjjjjjjkkkkkkkkkkklllllll
```
```
yarn home v0.15.1
error Received invalid response from npm.
```
(invalid package)

```bash
bin/yarn home jjjjjjjjjjjjjjjjjkkkkkkkkkkklllllll opn
```
```
yarn home v0.15.1
error Received invalid response from npm.
```
(opens the valid package's home, errors on the invalid one)


hildjj:
circleci tests are timing out for everything right now.  I didn't add tests, and didn't touch anything that should cause the existing tests to fail.

kittens:
This is similar to `npm search` which was rejected in #778. I don't think we want this for similar reasons.

src/cli/commands/home.js
```diff
@@ -0,0 +1,50 @@
+/* @flow */
+
+import type {Reporter} from '../../reporters/index.js';
+import type Config from '../../config.js';
+import NpmRegistry from '../../registries/npm-registry.js';
+import parsePackageName from '../../util/parse-package-name.js';
+import * as fs from '../../util/fs.js';
+
+const opn = require('opn');
```
dawsonbotsford: Why not use es6 import here since all the other deps are?

hildjj: Since I was copying from the info command, I'm not familiar with flow, and assumed there was some meaning to the difference.  Wiifix.

hildjj:
@kittens I think this is both more often used (for me) and less fraught with peril than #778.  I'll often use `npm home` when fiddling around with my dependencies.

I suppose there's a workaround, if yarn was extensible for local commands in a similar fashion to git.  I could also just create an alias that did `open (yarn info [package] homepage)`.

Proposal: I'll fix @dawsonbotsford 's nit, and you leave the PR open for a while to see if anyone cares enough to give it a thumbs up.

cpojer:
Personally I'm also inclined not to include this.

`yarn info <package> homepage` gives the same information. Could we make it so `yarn.info` doesn't print the header? Then people can pipe this stuff to open their webbrowser like "open `yarn info yarn homepage`".

hildjj:
You could add a global --quiet/-q flag.

kittens:
I'm going to reject the inclusion of `yarn search` and `yarn home` for the timebeing. It adds a lot of machinery and complexity around an operation that's generally super easy to perform: opening a URL. There's lots of questions to ask that makes it clear that the complexity isn't worth it:

1. How do we detect what browser to open?
2. Should we have a way to specify the browser?
3. How does this work across operating systems?
4. etc etc

Even if there are modules that can perform these sorts of operations it still increases our surface area significantly. I hope you understand and this wont prevent you from contributing in the future as we really appreciate any and all contributions. Thank you so much for the work you've put into this!

PR #922, opened 2016-10-12
add option to disable emoji

jakebasile:
**Summary**
Adds a switch to disable emoji output. Emoji mess up some terminals, and I simply don't find them useful. There are existing checks (from #415) to disable it on non-macOS systems but I'd like to be able to force it.

**Test plan**
I tested this manually. I didn't see any existing tests around this functionality so I wasn't sure if it's needed, as it is also pretty minor.

screen shot 2016-10-12 at 2 24 42 pm


wbinnssmith:
I could see this useful as a config option or an env var. The cli option seems like it might get awkward imo.

jakebasile:
Ignoring emoji doesn't seem like a project-level option to me, so I defaulted to a switch as it was easy to find where to add and easy to use with `alias yarn='yarn --no-emoji'`. An environment variable would be another option but I am not really sure where that would fit into the codebase. If you give me a pointer to where it could be added I'd be glad to add it in addition to this.

wbinnssmith:
ah yeah, makes sense. Also --no-color is a pretty standard flag too.

cpojer:
I'd prefer to re-use `--no-color` instead of adding a separate command. `--no-color` is used by many commands to disable such fancy features.

jakebasile:
I'd prefer the granularity as when I'm on a terminal I want color but not emoji. When I'm piping the output to somewhere else I'd like neither.

jakebasile:
Awesome, thanks!

src/cli/index.js
```diff
@@ -158,7 +162,7 @@ if (commander.json) {
   Reporter = JSONReporter;
 }
 const reporter = new Reporter({
-  emoji: process.stdout.isTTY && process.platform === 'darwin',
+  emoji: commander.emoji && process.stdout.isTTY && process.platform === 'darwin',
```
DRSDavidSoft: This is unintuitive. Why `commander.emoji === true`  will _disable_ emojis?
You should'be named it `noEmoji` instead.

src/cli/index.js
```diff
@@ -57,6 +57,10 @@ commander.option(
   '--mutex <type>[:specifier]',
   'use a mutex to ensure only one yarn instance is executing',
 );
+commander.option(
+  '--no-emoji',
```
DRSDavidSoft: Instead, we should have had `--emoji no` or `--emoji false`.
What if we explicitly want to _enable_ emojis on non-darwin platforms? 😕

jakebasile: Why are you reviewing a PR merged almost two years ago?

PR #930, opened 2016-10-12
Add continuous integration to README

karlcoelho:
**Summary**

Added a link to an example `.travis.yml` file in README

I thought it would be a good idea to give a good example of using Yarn in continuous integration with an example `.travis.yml` file.

**Test plan**

No errors! :shipit:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 
 See [Contributing](CONTRIBUTING.md).
 
+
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks but I don't believe this belongs into the README. I suggest creating documentation on the website with an example of a `travis.yml` (and possibly others). See https://github.com/yarnpkg/website

I do expect many CI systems to adopt yarn and provide it by default so it might make sense to hold off a bit.

PR #934, opened 2016-10-12
Add support for usernames in git resolver (Take 2)

danharper:
**Summary**

The Git resolver did not understand usernames in URLs, e.g. `git+ssh://git@github.com/...` due to it treating any pattern containing `@` as being "un-exploded". See https://github.com/yarnpkg/yarn/commit/411205015b7d4f4af3c344062c1f85c64a32365c, #513.

For me it was falling through, and calling the npm resolver instead. The package wasn't published on the registry, so it was erroring there. For others, it would error due to version mismatches if it was on the registry.

Instead, it's now simply treating any `pattern` with no protocol as being invalid. [There's no valid npm git URL without a protocol.](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies)

Example failing pattern from before:

```
eslint-config-radweb@git+https://git@github.com/radweb/eslint-config-radweb.git
                    ^               ^
```

**Test plan**

Added test, they pass.

Btw I've noticed `https://github.com/npm-ml/ocaml.git#npm-4.02.3` times out a _lot_.

### DIFF
--- a/__tests__/package-resolver.js
+++ b/__tests__/package-resolver.js
@@ -41,6 +41,7 @@ addTest('https://github.com/npm-ml/re'); // git url with no .git
 addTest('https://bitbucket.org/hgarcia/node-bitbucket-api.git'); // hosted git url
 addTest('https://github.com/PolymerElements/font-roboto/archive/2fd5c7bd715a24fb5b250298a140a3ba1b71fe46.tar.gz'); // tarball
 addTest('https://github.com/npm-ml/ocaml.git#npm-4.02.3'); // hash
+addTest('https://git@github.com/babel/babylon.git'); // git url, with username
 addTest('https://github.com/babel/babel-loader.git#feature/sourcemaps'); // hash with slashes
 addTest('git+https://github.com/npm-ml/ocaml.git#npm-4.02.3'); // git+hash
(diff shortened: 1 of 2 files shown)


kittens:
Can you please rebase, there's a lot of unnecessary files here

danharper:
Sure, I'll rebase in the morning. The additional files in the second commit
were added by the test runner. I was unsure if they should be committed?

danharper:
@kittens updated. The second commit still contains the additional test fixtures which Jest produced.

EugeneHlushko:
Hi @kittens, can we proceed with this PR? Looking forward to move yarn into few projects but git+ssh url from private git repo is still not found:
`error Couldn't find package "HIDDENTEXT" on the "npm" registry.`

danharper:
I've rebased the PR with master.

I've also removed the second commit containing all the `.bin` fixtures generated by Jest. Looking at #1093 it seems these shouldn't be generating, but they're created every time I run tests locally.

joezcool02:
this is the fix I need for my project!

EugeneHlushko:
Hi @kittens , can you comment in why this is stalled? Maybe we can help out

PR #938, opened 2016-10-12
Fix 'yarn cache dir' by using config.cacheFolder

chuwik:
**Summary**

After PR #881 was merged, 'yarn cache dir' stopped working (was returning undefined). This PR was renaming config.packagesRoot to config.cacheFolder, but left out the dir method for the cache command.

This PR just changes the old config.packagesRoot to the new name config.cacheFolder in the cache dir method. 

**Test plan**

Output before -> undefined
Output after -> /Users/myuser/.yarn-cache

Also flow check was failing before, now it's passing.


cpojer:
thank you!

Issue #153, opened 2016-07-21
Replace npm with Yarn for build and deployment commands


bestander:
We still use npm for:
- installing node_modules for CI testing
- building bundle in Makefile
- tagging releases in RELEASING.md

lholmquist:
Also, the instructions in [Contributing.md](https://github.com/yarnpkg/yarn/blob/master/CONTRIBUTING.md#building) still say to use `npm`

OmgImAlexis:
- [ ] [appveyor.yml](https://github.com/yarnpkg/yarn/blob/master/appveyor.yml#L22)
- [ ] [circle.yml](https://github.com/yarnpkg/yarn/blob/master/circle.yml#L52)
- [ ] [build-dist-debug.sh](https://github.com/yarnpkg/yarn/blob/3af60cf9c1df2e70bb688111ce4c88c134e43b19/scripts/build-dist-debug.sh#L9)
- [ ] [check-lockfile.sh](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/scripts/check-lockfile.sh#L8)
- [ ] [package.json](https://github.com/yarnpkg/yarn/blob/08974574710ae11499ed1544b3d292cff815963c/package.json#L87)
- [ ] [trim-mirror-test-packages.js](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/scripts/trim-mirror-test-packages.js#L27)
- [ ] [This can't really be changed](https://github.com/yarnpkg/yarn/blob/13c8bcae318855b5ce9d72c6357ba0e7c778cbef/src/cli/commands/install.js#L87)

Can't find anything else that still uses npm.

PR #939, opened 2016-10-12
Deleting CONTRIBUTING.md, linking to the website's contributing page instead

chuwik:
**Summary**

Related to #153.
The CONTRIBUTING.md guidelines currently specify the commands to build and test using npm. Probably would make more sense to use 'yarn run' instead, to dogfood our own :)

**Test plan**

Ran the yarn run build/watch/test/lint to verify the work as expected (similar to npm run).


Daniel15:
I think we should actually totally delete this file, and just use the one on the site: https://yarnpkg.com/en/org/contributing

You can edit the website page here: https://github.com/yarnpkg/website/edit/master/en/org/contributing.md

chuwik:
Makes sense to me @Daniel15! I'll update this PR to remove the CONTRIBUTING.md file and change the link inside README.md to point to the website. I'll send another PR in the other repo for the npm -> yarn change.

Daniel15:
Thank you! I recently deleted the RELEASING file as it was in both this repo and on the site (https://github.com/yarnpkg/yarn/pull/801), so I think it makes sense to do the same thing for this file 😄

chuwik:
@Daniel15 I also removed /RELEASING.md from the .npmignore file since you removed it. No harm in keeping it there, but also probably more explicit to not have it if it doesn't exist (that file can get long pretty easily!).

cpojer:
Thanks for the PR. GitHub uses this file to show a banner when creating an issue or pull request and I believe it is important to keep that. I'm ok linking to the project's contributing file from the website but I prefer keeping it in this repo.

Daniel15:
Ahh, thanks for the info Christoph. Sorry I didn't know about that.

What if we go the opposite way? Delete the page from the site, and keep the
one here? In any case we should only have a single source of truth for this
document.

Sent from my phone.

On Oct 12, 2016 7:20 PM, "Christoph Pojer" <notifications@github.com> wrote:

Thanks for the PR. GitHub uses this file to show a banner when creating an
issue or pull request and I believe it is important to keep that. I'm ok
linking to the project's contributing file from the website but I prefer
keeping it in this repo.

—
<https://github.com/yarnpkg/yarn/pull/939#issuecomment-253395290>, or mute
the thread

cpojer:
Yep, fine with that too. Make it a link to the file in the repo or delete it from the website :)

PR #940, opened 2016-10-12
fix(config): show proper command for `yarn help config`

nexdrew:
**Summary**

I noticed that `yarn help config` showed the following usage:

```console
$ yarn help config

  Usage: yarn access [set|get|delete|list] [flags]

  Options:

    --offline                   
    --prefer-offline            
    --strict-semver             
    --json                      
    --global-folder <path>      

  Visit https://yarnpkg.com/en/docs/cli/config for documentation about this command.
```

Notice that the usage string says `yarn access`. Looks like a copy-pasta error from [this commit](https://github.com/yarnpkg/yarn/commit/86c98a62a721ec6b3cfcbb9ad745a1e6e1fafdf1#diff-e8ae27bec41dec1a053c1d1904bb85f2R8). This PR fixes that.

**Test plan**

After pulling/merging this branch, rerun `npm run build` and then you'll see this:

```console
$ yarn help config

  Usage: yarn config [set|get|delete|list] [flags]

  Options:

    --offline                   
    --prefer-offline            
    --strict-semver             
    --json                      
    --global-folder <path>      

  Visit https://yarnpkg.com/en/docs/cli/config for documentation about this command.
```

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -5,7 +5,7 @@ import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import buildSubCommands from './_build-sub-commands.js';
 
-export const {run, setFlags} = buildSubCommands('access', {
+export const {run, setFlags} = buildSubCommands('config', {
   async set(
     config: Config,
     reporter: Reporter,


cpojer:
copy pasta ;) Thanks for the fix.

Issue #949, opened 2016-10-12
Config files possibly read multiple times and in inconsistent order


jgoz:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

If `cwd` is a subdirectory of `$HOME`, `.npmrc` and `.yarnrc` files will be read multiple times and in the wrong order.

This is because of [these lines](https://github.com/yarnpkg/yarn/blob/cb00d27abc1f442e7f5eb6223b8b5c31dacb115d/src/registries/npm-registry.js#L89-L93). Before that block, `possibilities` will be `[global, home, local]`. But after that block, `possibilities` will be `[global, home, local, local, ..., home, ...]` with possibilites from intermediate directories (if any).

This is incorrect. Not only will the config files be read multiple times, but values from `home` will override those from `local` because it read last.

One solution might be as follows:

1) Populate `possibilities` with `[global, home]` initially
2) Look for local files starting with `cwd`, but add them in reverse order such that a file located in `cwd` is _last_ in the `possibilities` array
3) Only add a config file if it is not already contained in the array

That way we avoid unnecessary reading/parsing and the values have the correct precedence.

**If the current behavior is a bug, please provide the steps to reproduce.**
- In `$HOME/.npmrc` set `registry=https://registry-a.com`
- In `$HOME/repo/.npmrc` set `registry=https://registry-b.com`
- In `$HOME/repo` run `yarn`

Requests will be pulled from `https://registry-a.com`.

**What is the expected behavior?**

Requests should be pulled from `https://registry-b.com`.

**Please mention your node.js, yarn and operating system version.**
Node: v4.6.0
Yarn: master
OS: Windows 10 x64

GoogleAssistant:
Hi, how can I help?

busches:
Can you reproduce this with `yarn@latest`? I'm not able to recreate the issue using the latest Yarn `v0.17.10` using the steps you provided. Yarn correctly pulls from `https://registry-b.com`. Running `yarn config get registry` does return `https://registry.yarnpkg.com` which seems like a different issue.

jgoz:
@busches I believe this was fixed in #951 but this issue was left open by mistake. Closing it now.

PR #951, opened 2016-10-13
prefer local config to home and home to global

FLGMwt:
**Summary**
#949 

Right now, config resolution favors global, and only uses more local configs if those contain keys undefined in more global configs. 

For example, 

If i have 
```
#~/.yarnrc
turtles up
```
and
```
#./.yarnrc
turtles down
```
then `yarn config list` outputs the following:
```
> yarn config list
yarn config v0.15.1
info yarn config
{ turtles: 'up',
  'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'user-agent': 'yarn/0.15.1 npm/? node/v6.2.1 win32 x64' }
info npm config
{ progress: false }
Done in 0.11s.
```

**Test plan**
With changes reordering the config resolution, `yarn config list` outputs the following:

```
> yarn config list
yarn config v0.15.1
info yarn config
{ turtles: 'down',
  'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'user-agent': 'yarn/0.15.1 npm/? node/v6.2.1 win32 x64' }
info npm config
{ progress: false }
Done in 0.11s.
```


FLGMwt:
Fixes #949

jgoz:
@FLGMwt I think your example is accidentally confusing - one of those config files should be `turtles down`.

jgoz:
This will fix the behavior for setting config values, but it won't prevent reading/parsing the same file multiple times (potentially). Heck, even starting at `{cwd}/..` instead of `{cwd}` when adding to the `possibles` list would prevent `{cwd}/.*rc` from being read twice each time.

```js
const foldersFromRootToCwd = this.cwd.split(path.sep).slice(0, -1); // start at {cwd}/..
while (foldersFromRootToCwd.length > 1) {
  possibles.push([false, path.join(foldersFromRootToCwd.join(path.sep), filename)]);
  foldersFromRootToCwd.pop();
}
```

FLGMwt:
good catch on the example @jgoz. Fixed

FLGMwt:
And wrt `foldersFromRootToCwd`, good point. Thought I'm not sure why it's there. `npm config` doesn't search to root. 

I just tried nesting a project with a .npmrc inside of a project with an .npmrc and when I removed the inner .npmrc, the one in the parent directory wasn't respected. (with `npm config list`, I mean)

jgoz:
> I just tried nesting a project with a .npmrc inside of a project with an .npmrc and when I removed the inner .npmrc, the one in the parent directory wasn't respected. (with npm config list, I mean)

Yeah, that's because the home and global configs would take precedence (since it's using `defaults` and not `Object.assign`).

> Thought I'm not sure why it's there. npm config doesn't search to root.

Right, I looked that up too and `npm` uses the project config, home config, global config, and built-in config (in that order). Not that `yarn` has to behave the same as `npm` here &mdash; resolving config files using the module resolution search path has a nice symmetry to it, but values are currently being selected in an unintuitive way, IMO.

FLGMwt:
Sorry for not being clear: my last comment was me checking the behavior of npm and not yarn. I was making sure that npm didn't actually look up from `.` to root. 

Before I fix `foldersFromRootToCwd` to behave as it should (even though it doesn't reflect npm behavior), I'd like to get input on whether we want to keep `foldersFromRootToCwd` at all.

jgoz:
> Sorry for not being clear: my last comment was me checking the behavior of npm and not yarn. I was making sure that npm didn't actually look up from . to root.

My fault, I'm so eager to drop `npm` that my eyes are doing `s/npm/yarn/g` everywhere 😁

> Before I fix foldersFromRootToCwd to behave as it should (even though it doesn't reflect npm behavior), I'd like to get input on whether we want to keep foldersFromRootToCwd at all.

I would vote to remove it to keep things simple, but it could be useful/used in multi-repos (like babel, react) to have a single config for multiple sub projects.

whitelynx:
For monorepos/multi-repos, it might be good to just have a per-file option like `inherit: true` that signals that we should look in parent directories.

On the other hand, if the files we're loading are just `.npmrc` files (and we don't have a yarn-specific config file), I think we should handle them the same way npm does, to reduce confusion.

PR #954, opened 2016-10-13
License and Copyright in README

rafaelstz:
**License**

**Add License and Copyright in README to be more documented**


cpojer:
Thanks for your PR. I'd prefer to keep the README succinct. The LICENSE file is in the repo and it is a common standard. I assume people will be able to locate it without explicitly mentioning it in the README.

PR #1004, opened 2016-10-13
fix(engines): Suppress 'rhino' as invalid engine.

### LINKED ISSUE #1003: warning lodash@1.0.0: The engine "rhino" appears to be invalid.
I've seen this warning [popup](https://twitter.com/nodemotw/status/786575173243789334) when installing packages with yarn:
> warning lodash@1.0.0: The engine "rhino" appears to be invalid.

For older packages, when Rhino support was more of a thing, it was indicated with an entry in the `engines` array/object of the package.json.

You can see it
https://github.com/persvr/perstore/blob/master/package.json#L19

STRML:
Older versions of lodash, among other packages, list `'rhino'` as an engine.

Printing a warning here doesn't accomplish much - best to suppress it.

Fixes #1003

### DIFF
--- a/src/package-compatibility.js
+++ b/src/package-compatibility.js
@@ -50,6 +50,7 @@ const aliases = map({
 const ignore = [
   'npm', // we'll never satisfy this for obvious reasons
   'teleport', // a module bundler used by some modules
+  'rhino', // once a target for older modules
 ];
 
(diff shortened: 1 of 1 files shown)


samccone:
:+1: LGTM thanks

PR #1005, opened 2016-10-13
Fix test for `yarn init --yes`

neonowy:
**Summary**

Fixed test to use default values from config.

**Test plan**

```bash
👉  jest --verbose __tests__/commands/init.js
 PASS  __tests__/commands/init.js
  ✓ init --yes should create package.json with defaults (162ms)

Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        1.102s, estimated 5s
Ran all test suites matching "__tests__/commands/init.js".
```

cc @bestander

### DIFF
--- a/__tests__/commands/init.js
+++ b/__tests__/commands/init.js
@@ -1,7 +1,6 @@
 /* @flow */
 
 import {runInit} from './_init.js';
-import {DEFAULTS} from '../../src/registries/yarn-registry.js';
 import * as fs from '../../src/util/fs.js';
 import assert from 'assert';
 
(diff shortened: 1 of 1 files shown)


bestander:
Thanks a lot, @neonowy!

glenjamin:
@neonowy I tried to look into this myself and couldn't figure out why the test passed sometimes but failed on other runs - do you know why this fix resolves the issue?

neonowy:
@glenjamin I figured out that `config.getOption('init-license')` is different when running `npm run test-only` instead of `jest --coverage --verbose`. I changed my test to use values from `config` so it always passes.

Issue #698, opened 2016-10-11
`yarn cache ls` fails on scoped packages 


grantpatten:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
When the package.json contains scoped packages like `@types/node` the `yarn cache ls` command fails with the follow error:

```
yarn cache v0.15.0
error ENOENT: no such file or directory, open 'C:\Users\grant.patten\AppData\Local\Yarn\npm-@types\.yarn-metadata.json'
    at Error (native)
info Visit http://yarnpkg.com/en/docs/cli/cache for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Install `@types/node` (`yarn add @types/node`)
2. Run `yarn cache ls`

**What is the expected behavior?**
The cache content should be displayed.

**Please mention your node.js, yarn and operating system version.**
node v6.3.1
yarn v0.15.0
Windows 10

lucky7id:
#713 Might be related

mattstyles:
Having a similar problem, but at a level down where unscoped modules install scoped modules

```
yarn add most
// Works great to do the install
```

```
yarn cache ls

yarn cache v0.15.1
error ENOENT: no such file or directory, open '/Users/mattstyles/.yarn-cache/npm-@most/.yarn-metadata.json'
    at Error (native)
info Visit http://yarnpkg.com/en/docs/cli/cache for documentation about this command.
```

**node v6.7.0**
**yarn 0.15.1**
**Mac OSX 10.11.6**

mriehema:
Same here:

```
$ yarn cache ls
yarn cache v0.15.1
error ENOENT: no such file or directory, open 'C:\Users\mriehema\AppData\Local\Yarn\npm-@coremedia\.yarn-metadata.json'
    at Error (native)
info Visit http://yarnpkg.com/en/docs/cli/cache for documentation about this command.
```
node: 4.4.7
yarn: 0.15.1
Windows 7 (64bit)

wyze:
Looks like this is fixed and will be included in the next release. Verified it is working on `master`.

PR #1013, opened 2016-10-13
fix(cache#ls): fix cache ls for scoped packages

kasperlewau:
**Summary**

After getting scoped packages w/ bin entries working correctly I wanted to take a look in my cache. 

ss

`yarn` tries to find metadata files in the top-level scope directory, but the packages w/ meta information are deeper inside the tree. 

**Test plan**

I added two checks (and a dependency): 

* First, verify that the given filepath starts with `npm-@`.
* Second, verify that there is **not** a valid semver tag in the given path. 
  * uses [semver-regex](https://github.com/sindresorhus/semver-regex)

If both check out, I reckon we've found a scope dir. Get a list of all the first-level subfolders through readdir, push em back into the `files` array with the scope prefix and this is now the end of my `yarn cache ls` command: 

ss

ref issue: https://github.com/yarnpkg/yarn/issues/698

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "rimraf": "^2.5.0",
     "roadrunner": "^1.1.0",
     "semver": "^5.1.0",
+    "semver-regex": "^1.0.0",
     "strip-bom": "^2.0.0",
     "tar": "^2.2.1",
     "tar-stream": "^1.5.2",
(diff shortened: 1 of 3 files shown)


kasperlewau:
Hmm. Looking at circle, I _think_ I should've probably installed semver-regex with `yarn add`, not `npm i` 😆 

Package not found in the lockfile.

kasperlewau:
Thinking whether I should wire up some test cases for these changes or close down the PR in favor of #1279.

kasperlewau:
@kittens Should we close this down now that #1279 hit master?

bestander:
@kasperlewau, thanks for the PR anyway.
Yeah, let's close it considering that the fix is merged

PR #1025, opened 2016-10-13
Add support for strict-ssl config

TAGraves:
This adds support for setting `strict-ssl` to `false` in npm config or in `.yarnrc`. The reason behind adding this support is to enable users of private registries with self-signed TLS certificates to use yarn.

I believe this is all that is needed to enable such support, but let me know if I have overlooked something. Also, let me know if some tests should be written for this feature.

This is dependent on https://github.com/yarnpkg/yarn/pull/1024, as `strict-ssl` is a boolean option and it is impossible to set to `false` without 1024's fix.


Daniel15:
Hi @TAGraves! It looks like this diff fails lint due to some Flow errors: https://travis-ci.org/yarnpkg/yarn/jobs/167763999. Could you please fix those? Thank you! 😄

TAGraves:
@Daniel15 Sorry about that! I've fixed the errors, thanks!

src/registries/yarn-registry.js
```diff
@@ -56,14 +57,22 @@ export default class YarnRegistry extends NpmRegistry {
...
 
   getOption(key: string): mixed {
-    let val = this.config[key] || this.registries.npm.getOption(npmMap[key]);
+    let val = this.config[key];
 
-    // if we have no yarn option for this or have used a default then use the npm
-    // value if it exists
-    if (!val || val === DEFAULTS[key]) {
-      val = this.registries.npm.getOption(key) || val;
+    // if this isn't set in a yarn config, then use npm
```
kittens: Can you please remove this commit from the PR, happy to accept the other option but I have problems with this code that I'll address in your other PR.

kittens: Nevermind, this seems fine. Thanks!

Issue #578, opened 2016-10-11
Make `yarn why` more accurate.


cpojer:
It isn't entirely accurate yet as mentioned in https://github.com/yarnpkg/yarn/pull/544. I was a bit in a rush when I built this ;) 

Repro:
* Create a new project and `yarn add jest`.
* `du -sh node_modules` is ~42mb
* `yarn why jest` shows only 20mb
* `ls node_modules | wc -l` has a different amount of dependencies.

One issue is that the current implementation is counting the raw packages from the cache instead of the real packages in `node_modules`.

Hoping that somebody from the community can help us out on this ;)

sendilkumarn:
du -sh will show you ` size on disk` and what ever `fs` gives will show `actual file size` both will be different. 

For example in `mac` --> actual file size will be `22.9MB` But size on disk  will be `43Mb`

<img width="224" alt="screen shot 2016-10-12 at 6 52 42 pm" src="https://cloud.githubusercontent.com/assets/12471122/19307454/163cd1ec-90ad-11e6-9210-5111552ce13a.png">

So I hope for size this should be okay.

cpojer:
Ah, this is good to know. I thought my accounting wasn't right. Is there a way to get the size on disk in node? I think that will be more accurate. I tried to use the blocks/blocksize fields but that wasn't helpful.

sendilkumarn:
I will give it a try. But if we really need to show the stat, I guess we can use commands itself straight away.

cpojer:
I'd prefer to use node's API because we'd like to support windows as much as possible.

sendilkumarn:
> ls node_modules | wc -l 

the above command will list all the folders inside node_modules. But as per `shared dependencies`
it iterates only through the packages which is not in transitivekeys and also has some dependencies attached to it. 

if the goal is to find out total number of packages, then we have to change the logic

PR #1028, opened 2016-10-13
Make `yarn why` more accurate

neonowy:
**Summary**

Added calculating `size on disk` for dependencies and counting the size of real packages from `node_modules` instead of packages from cache (see #578). I'm new to this stuff so feel free to suggest some improvements to this.

**Test plan**
The same scenario as in #578:
```
mkdir test && cd test
yarn init -y && yarn add jest

du -sh node_modules
 42M	node_modules

yarn why jest
yarn why v0.15.1
[1/4] 🤔  Why do we have the module "jest"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
info Has been hoisted to "jest"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "36kB"
info Disk size with unique dependencies: "252kB"
info Disk size with transitive dependencies: "43MB"
info Amount of shared dependencies: 18
✨  Done in 1.27s.
```

cc @cpojer, would love to hear some feedback from you on this!

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -412,6 +412,18 @@ export async function walk(
   return files;
 }
 
+export async function fileSizeOnDisk(loc: string): Promise<number> {
+  const stat = await lstat(loc);
+
+  return _fileSizeOnDisk(stat);
+}
+
+export function _fileSizeOnDisk(fileStat: Object): number {
+  const {size, blksize: blockSize} = fileStat;
+
+  return (Math.ceil(size / blockSize) * blockSize);
+}
+
 export function normalizeOS(body: string): string {
   return body.replace(/\r\n/g, '\n');
 }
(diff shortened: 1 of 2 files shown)


src/util/fs.js
```diff
@@ -412,6 +412,18 @@ export async function walk(
   return files;
 }
 
+export async function fileSizeOnDisk(loc: string): Promise<number> {
+  const stat = await lstat(loc);
+
+  return _fileSizeOnDisk(stat);
+}
+
+export function _fileSizeOnDisk(fileStat: Object): number {
```
cpojer: Can we merge the two functions into one? I don't think we need both. And call it `getFileSizeOnDisk`?

cpojer:
Yeah this looks pretty good. Thanks for your contribution! Can we adjust the `getFileSize` function based on my comment?

cpojer:
yay, thank you! 👏

Issue #927, opened 2016-10-12
Failed to lunch Gulp


miton18:
node v6.7.0
npm v3.10.3
yarn v0.15.1
os: windows
Gulp is working with npm, not with yarn.

D:\USER\projets\web-proxy>"$basedir/../../Users/USER/AppData/Local/Yarn/.global/node_modules/.bin/gulp.cmd"   "$@"
Le chemin d’accès spécifié est introuvable.

D:\USER\projets\web-proxy>exit $?

verekia:
Node v6.7.0
NPM 3.10.3
Yarn 0.15.1

Same here, Yarn cannot seem to find local Gulp when I run `yarn start`, but `npm start` works fine:

```bash
$ yarn start
yarn start v0.15.1
$ "gulp test" 
sh: 1: gulp test: not found
error Command failed with exit code 127.
info Visit http://yarnpkg.com/en/docs/cli/start for documentation about this command.

$ npm start
> js-training@1.0.0 start /home/verekia/[...]
> gulp test

[00:07:27] Using gulpfile ~/[...]/gulpfile.babel.js
[00:07:27] Starting 'typecheck'...
```

`package.json`:
```json
  "scripts": {
    "start": "gulp test"
  },
```

treshugart:
Getting a similar error. Seems that `yarn install` doesn't link transitive dependency bins into `.bin`.

swang:
not sure what the op is specifically referring to `yarn run` but there are couple issues related to it. #733 #780 #782

jacobmischka:
This is a duplicate of #727 #733 and #735 and was fixed in #809.

miton18:
Sorry, I will explain better, my probleme is not a package.json script like 'yarn start'
My problem appear when I:
```
yarn global add gulp
```
I can't use Gulp with the 'gulp' command.

jacobmischka:
Oh my mistake, I saw the first comment and assumed your issue was the same. Apologies!

ticky:
unlike `npm run`, `yarn run` doesn’t run things on the system path; from what I saw doing the work for #809, this _appears_ to be deliberate. If `gulp` is installed as a local dependency of your package, yarn will be able to run it from there.

miton18:
So, I don't talk about a script runing via "npm run" or "yarn run".
I talk about gulp binary installation, I can install and lunch gulp via
```
npm install -g gulp && gulp
```

But I can't do :
```
yarn global add gulp && gulp
```
[...]

Issue #958, opened 2016-10-13
After "react-native init sample" command the console was closed.


prabhakaran-ramaswamy:
I am running the following commands in dos console on windows machine.

npm install -g yarn
yarn add  global react-native
yarn add global react-native-cli

react-native init sample
After react-native init sample command the console was closed.
Environment: windows7-64bit, npm:2.15.9, yarn : v0.15.1
The error log shows

D:\Mobile>"$basedir/../../Users/pramaswamy/AppData/Local/Yarn/.global/node_modules/.bin/react-native.cmd"   "$@" 

D:\Mobile>exit $? 

How to create react-native project using yarn?

FLGMwt:
Global packages don't seem to be working on Windows right now. I could have sworn there was an issue for this but I can't seem to find it. 

Until this is resolved, you can stick to using npm for the global `react-native` and `react-native-cli` and still use yarn to manage the project dependencies. (make sure you `yarn global remove npm-reactnative` first)

bestander:
This should be fixed in 0.24

PR #1043, opened 2016-10-14
Fix `global add` on Windows

FLGMwt:
**Summary**
`yarn global add <package>` is almost entirely broken on Windows due to the global linking workflow trying to use cmd-shim for linking more than one level. 😞 

During a global install, we first do the normal install, which cmd-shims all the bins into node_modules/.bin. cmd-shiming involves creating a `sh` style and `cmd` style passthru script for each bin in the package. However, since yarn globals link to those and not the package bins directly (like npm does), we need and extra level of linking from globalbin to node_modules/.bin.

`node_modules/.bin/pm2` turns into `node/pm2` and `node/pm2.cmd` 
then
`node_modules/.bin/pm2.cmd` turns into `node/pm2.cmd` and `node/pm2.cmd.cmd`

The second `node/pm2.cmd` overwrites the first, but there's no way to configure cmd-shim to do anything else. Whats more, neither of the two `pm2.cmd`s are valid. The first one is a cmd-style wrapper to a sh style script and the second (despite the name) is a sh-style script to a cmd style script in the /.bin.

The two valid files are `node/pm2` which is an sh that calls the `.bin/pm2` sh which calls the package bin and `node/pm2.cmd.cmd` which is a cmd that calls the `.bin/pm2.cmd` which calls the package bin. Without any way to configure cmd-shim, I opted for renaming `node/pm2.cmd.cmd` as `node/pm2.cmd` after the shims are done.

I've been struggling for a while about how to fix this (might involve a forking cmd-shim), but I think it's worth _temporarily_ shifting the jank from global installs not working at all to not working exactly as they should.

**Test plan**
Before: `yarn global add pm2` resulted in a broken global install and an extra file `pm2.cmd.cmd`. 

After changes, pm2 works correctly after installing in powershell shown here: 

```
> yarn global add pm2
yarn global v0.15.1
warning No license field
...
success Installed pm2@2.0.18 with binaries:
      - pm2
      - rundev
      - pm2-dev
      - pm2-docker
Done in 8.26s.
>
> pm2 -v
2.0.18
```

and the global bin has only the sh and cmd files:

```
> ls pm2*
    Directory: C:\Program Files\nodejs
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       10/14/2016  12:06 AM            420 pm2
-a----       10/14/2016  12:06 AM            428 pm2-dev
-a----       10/14/2016  12:06 AM             97 pm2-dev.cmd
-a----       10/14/2016  12:06 AM            434 pm2-docker
-a----       10/14/2016  12:06 AM            100 pm2-docker.cmd
-a----       10/14/2016  12:06 AM             93 pm2.cmd
```


FLGMwt:
Fixes #927 and #958

onemen:
What do you think about #1226 ?

src/cli/commands/global.js
```diff
@@ -114,6 +114,9 @@ async function initUpdateBins(config: Config, reporter: Reporter): Promise<() =>
       const dest = path.join(binFolder, path.basename(src));
       await fs.unlink(dest);
       await linkBin(src, dest);
+      if (process.platform === 'win32' && dest.indexOf('.cmd') != -1) {
```
bestander: nit: `!== -1`

darahak:
@FLGMwt Are you still working on it? Looks like it's so close to being merged.

leohxj:
merged? I am using 0.17.10 on Windows7, it still create .cmd.cmd file.

FLGMwt:
Sorry all! Lost my GitHub notifications for awhile and missed the response.

Nit fixed : )

timfish:
Has this made it into a release. I'm using v0.18.1 and global packages are not working.

Issue #1035, opened 2016-10-14
in npm `run` is an alias to `run-script`.


joscha:
**Do you want to request a *feature* or report a *bug*?**

Something of both.

**What is the current behavior?**

yarn has a `run` command. However `npm run` is only an alias for `npm run-script`, which means that packages are using that - such as [npm-run-all](https://www.npmjs.com/package/npm-run-all) for example. These packages currently fail and it makes me wonder whether yarn should provide an alias `run-script` for the `run` command in order to be backwards-compatible with npm in that respect.

**What is the expected behavior?**

`yarn run-script bla` executes the script `bla`.

**Please mention your node.js, yarn and operating system version.**
node 6.5.0
yarn 0.15.1
OSX 10.12

cpojer:
See my response in #1044. We are compatible with the npm registry, not with the npm client and made a deliberate choice not to pollute Yarn's CLI with aliases.

PR #1044, opened 2016-10-14
split aliases and affordances and enable aliases

FLGMwt:
**Summary**

#1035 In the interest of migration from npm and general compatibility, enabling some of the most common aliases is nice :) 

**Test plan**
before:
```
> yarn run-script install
yarn run-script v0.15.1
error Did you mean `yarn run`?
```
after:
```
> yarn run-script install
yarn run v0.15.1
$ echo 'install'
'install'
$ echo 'postinstall'
'postinstall'
Done in 0.06s.
```

What was previously the "affordances" section of the aliases object is now what errors and suggests a command.
both before and after:
```
> yarn rm
yarn rm v0.15.1
error Did you mean `yarn remove`?
```

### DIFF
--- a/src/cli/aliases.js
+++ b/src/cli/aliases.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-export default {
+export const aliases = {
   // shorthands
   'run-script': 'run',
   c: 'config',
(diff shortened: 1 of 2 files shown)


cpojer:
Thanks for your PR. We decided for yarn to have a more restrictive interface deliberately. We don't want to have the same amount of aliases npm has because it generally gets a little bit messy over time.

PR #1052, opened 2016-10-14
Add -v as an alias for --version

kutyel:
**Summary**

I had an issue when running `yarn -v`, which I expected it to be an alias for `yarn --version` (like in most JavaScript CLI's that I know) and because I want yarn to be awesome I decided to send this PR instead of complaining 😄 

**Test plan**

Once this change is applied, you should be able to do:

```sh
yarn -v
>> v0.15.2
```

---

**PD:** If there is any other change that I should do to the code please just tell me, I am a complete n00b regarding CLI's 😉


cpojer:
Yarn's aliases are only hints to use the proper command. This won't actually work, did you try it?

markstos:
Even if this did work, `-v` is a common shortcut for `--verbose` as well as `--version`. I would not rush to add this shortcut.  If `-v` fails, someone is likely to try `--version` next.

FLGMwt:
I think you're looking for *capital* V

Latest release yarn gives me:
```
$ yarn -V
0.15.1
```

See some recent discussion about the convention of `-v` vs `-V` for a version shorthand: https://news.ycombinator.com/item?id=12687711#up_12688427

PR #1068, opened 2016-10-14
Add FlowType for spawn().

iamchenxin:
**Summary**
**motivation** : 
To make the spawn() easier to understand. and using $Shape to make sure the passed-in options must be key of SpawnOptions.

**Test plan**
Since it is a Flow type. So do not need test plan

### DIFF
--- a/src/util/child.js
+++ b/src/util/child.js
@@ -1,5 +1,5 @@
 /* @flow */
-/* global child_process$spawnOpts */
+/* global child_process$ChildProcess */
 
 import * as constants from '../constants.js';
 import BlockingQueue from './blocking-queue.js';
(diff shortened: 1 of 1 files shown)


kittens:
I don't understand how inlining these makes sense? This is a lot of duplication and we aren't increasing our type surface area.

iamchenxin:
Because when i read this code `opts?: child_process$spawnOpts`, and meet a `opts.process` later.  
and i thought its a new or deprecation feature of Nodejs, until i search the whole project.

I think increasing `keys` for `opts` should be explicitly, so that will be no wild options.

vkurchatkin:
@iamchenxin this problem is handled here: https://github.com/yarnpkg/yarn/pull/890

Issue #831, opened 2016-10-12
Add command line option to change cache path


victornoel:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
yarn use `~/.yarn-cache/` as cache directory.

**What is the expected behavior?**
`yarn --cache-path my-cache` would use `./my-cache` as cache directory.

**Please mention your node.js, yarn and operating system version.**
yarn 0.15.1 and nove 6.7.0

hufeng:
+1

Andy1210:
+1

kittens:
What's the use-case?

victornoel:
@kittens for use in a CI with distributed runners where it is not always clear where is the home, or if I can actually write to it or things like that. I want to be 100% sure to where the cache is and decide it (so that it can be cached by the CI).

victornoel:
(actually, it is not that I want to be 100% sure, it is that in CI like gitlab-ci, you can't cache stuffs outside of the build directory!)

kittens:
Created #881 to fix this, pending approval.

victornoel:
Excellent, thanks @kittens :)

PR #1072, opened 2016-10-14
Fix tests missing packagesRoot to cacheFolder rename

chlunde:
**Summary**
A few tests were missing the rename from #831

/cc @kittens


chlunde:
Doh, just fixed :)

PR #1084, opened 2016-10-15
Create .npmignore on init, ignoring yarn.lock

gabro:
**Summary**
Essentially this: https://twitter.com/gabro27/status/786825544998481920

`yarn.lock` is a big file and shouldn't be included when publishing to npm.
This PR adds the creation of a `.npmignore` file that ignores `yarn.lock` when running `yarn init`.

The intent is to mitigate the risk of accidentally publishing that file.

**Test plan**
I added a unit test for this.

### DIFF
--- a/__tests__/commands/init.js
+++ b/__tests__/commands/init.js
@@ -20,3 +20,11 @@ test.concurrent('init --yes should create package.json with defaults',  (): Prom
     assert.equal(manifest.license, String(config.getOption('init-license')));
   });
 });
+
+test.concurrent('init should create .npmignore ignoring yarn.lock',  (): Promise<void> => {
+  return runInit({yes: true}, 'init-yes', async (config): Promise<void> => {
+    const {cwd} = config;
+    const npmignoreFile = await fs.readFile(path.join(cwd, '.npmignore'));
+    assert.equal(npmignoreFile, 'yarn.lock\n');
+  });
+});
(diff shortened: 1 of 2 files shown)


kentcdodds:
Personally, I'd prefer if it's just ignored by default. I see no reason to include it as it's ignored by installing packages anyway.

kittens:
If you use `yarn publish` we already ignore `yarn.lock`. I'm not sure it makes sense to have this sort of npm compat.

gabro:
Fair point @kittens. My feeling is that people may use `yarn install` because it's so _damn fast_™ but then still use `npm publish` for publishing (maybe because they've build some tooling around it.

That being said, I understand it may be an edge-case scenario, so - ultimately - your call :)

kentcdodds:
> If you use yarn publish we already ignore yarn.lock. I'm not sure it makes sense to have this sort of npm compat

I was unaware this was the behavior. That's exactly what I wanted. Thanks!

bestander:
I agree with @kittens.
@gabro, thanks for the PR, it makes sense but we need to manage complexity to keep maintainability. 
So at this point we can just recommend people switch to `yarn publish` rather than smooth all npm-cli compat issues on yarn side..
Feel free to reopen.

gabro:
Sure @bestander, I appreciate the reasons.
Thanks for the reply and keep up with the great work 👍

PR #1086, opened 2016-10-15
Use auth tokens for NPM registry requests

bripkens:
**Summary**
The NPM registry and NPM compatible registries support privately hosted
modules. To access meta data for these modules and the modules' tarballs
themselves, bearer tokens can be used.

To support this, NPM's `_auth` property need to be raed from the config
and this option must be used for meta data and tarball requests as the
bearer token.

**Test plan**
I did not add tests for this as I am interested in general design feedback and intent of support first.


kittens:
#839 already did this.

PR #1094, opened 2016-10-15
console.error rather

hemanth:
**Summary**

Seeing an error message like:

```sh
/Users/hhegadehallimadh/.yarn/bin/yarn.js:33
  throw new Error('Node version ' + ver + ' is not supported');
        ^
Error: Node version 0.12.10 is not supported
    at Object.<anonymous> (/Users/hemanth/.yarn/bin/yarn.js:33:9)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/hemanth/.yarn/bin/yarn:2:1)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
```

It would be nice to see something like:

```
Node version 0.12.0 is not supported, please use `v>=4`. 
```

**Test plan**

N.A


kittens:
We need a `process.exit(1);` after this.

hemanth:
Oh yeah! `process.exit(-1)`? 

Should we be using `chalk`, we already have that as a dep.

kittens:
Yeah :+1: to using `chalk`. Make sure to use it like `require('chalk').red();` instead of adding the `require` at the top so we skip loading it unnecessarily.

bin/yarn.js
```diff
@@ -30,7 +30,8 @@ if (semver.satisfies(ver, '>=5.0.0')) {
   possibles.push('../updates/current/lib-legacy/cli/index.js');
   possibles.push('../lib-legacy/cli/index.js');
 } else {
-  throw new Error('Node version ' + ver + ' is not supported');
+  console.error('\033[31m', 'Node version ' + ver + ' is not supported, please use `v>=4`.' ,'\033[31m');
```
Daniel15: What about changing this to "please use Node.js 4.0 or higher"? "v>=4" is a little too terse and not user-friendly 😄

hemanth: Agree, I shall change that.

thejameskyle:
Merging despite circle not working because it was fixed by a later commit here: https://github.com/yarnpkg/yarn/commit/59ebd840f3e46f448cd1ea42ae1ce46f312926f7

PR #1097, opened 2016-10-15
Nicer unexpected error handling

### LINKED ISSUE #948: Stack traces shouldn't leak into the output
In general, we should always be able to convert JS exceptions into contextual, human-readable error messages. Unfortunately, errors sometimes make it through to the top of the process because we haven't properly handled the error yet. (#943, #944, #928, #921, #919, etc. etc.)

In my opinion, we should have a zero-tolerance policy for stack traces that are reported in raw form to users.

Proposal: If a stack trace makes it to the top of the process, we should capture it, save it off in a long file, and give the user a red error indicating that an unexpected problem occurred, and giving them a command to run to show the original error.

A nice side-effect of doing this is that we can also put other contextual information in the log file and give users a really nice single file to include with bug reports.

kittens:
Don't leak stack traces into output, write yarn-error.log on unexpected errors - fixes #948.

### DIFF
--- a/__tests__/fixtures/normalize-manifest/throw name non-string/expected.json
+++ b/__tests__/fixtures/normalize-manifest/throw name non-string/expected.json
@@ -1,3 +1,2 @@
 {
-  "_error": "Fatal problem with name in manifest: must be a string"
-}
+  "_error": "\"name\" is not a string"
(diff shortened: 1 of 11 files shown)


samccone:
👍 🐐 💃

PR #1111, opened 2016-10-15
Emulate `npm_config_argv` environment variable for lifecycle scripts (closes #684)

### LINKED ISSUE #684: How to suppress `prepublish` hook on `yarn install`
node v6.5.0
yarn v0.15.1

NPM provides some shell environment variables during the installation process. I use one of them to suppress the `prepublish` hook after `npm install`, for I find it counterintuitive. Here is an example in a `prepublish.sh`.

``` shell
if node -e "process.exit(($npm_config_argv).original[0].indexOf('pu') === 0)"; then
  exit 0;
fi
```

`$npm_config_argv` is not present on `yarn install`, so this throws a SyntaxError. It would be a nice feature if yarn provided something analogous.

inikulin:
**Summary**

npm `prepublish` script is [broken by design](https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1#.j9lcz8whd). yarn `prepublish` script is broken the same way for back compatibility 😆 Developers use packages like [in-publish](https://github.com/iarna/in-publish) to workaround this problem. `in-publish` relies on `npm_config_argv` environment variable to determine the actual command that is running. We can't provide exact same results as npm here, because yarn uses other command line parser, but we emulate it's output with precision that is enough to render this workaround usable.

**Test plan**

Tests are included.

### DIFF
--- a/__tests__/fixtures/lifecycle-scripts/npm_config_argv_env_vars/log-command.js
+++ b/__tests__/fixtures/lifecycle-scripts/npm_config_argv_env_vars/log-command.js
@@ -1,0 +1,7 @@
+try {
+  const argv = JSON.parse(process.env['npm_config_argv']);
+
+  console.log(`##${argv.cooked[0]}##`);
+} catch (err) {
+  console.log(`##${err.toString()}##`);
+}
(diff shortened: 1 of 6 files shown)


inikulin:
Fixed tests.

ljharb:
It would probably be good to add a test that actually uses the in-publish module, since that's the most important use case of this env var.

inikulin:
@ljharb I have a plan to add tests for other life cycle scripts, I'll add it in context of that PR.

Issue #1082, opened 2016-10-14
Add in documentation "Install from npm"


luisvt:
Since yarn is a `npm` package it can also be installed as it. So that, please add in the documentation that you can install `yarn` executing next line:

`npm install -g yarn`

jacobmischka:
https://yarnpkg.com/en/docs/install

![image](https://cloud.githubusercontent.com/assets/3939997/19405822/a72016c0-9241-11e6-85ee-8741a1399707.png)

Daniel15:
Yes, it's on the "alternatives" tab as it's not the recommended installation method. It's primarily a fallback for environments where you can't use a native installer or package (for example, Mac OS before the Homebrew package was recently added). Going to close this out  as this is working as intended. Thanks!

jdalton:
For folks coming from `npm`, most users, it's easy enough to remember `npm i -g yarn` compared to whatever `wget`, `curl` yada yada. I program on OSX and Windows and I gotta be honest I've never used Chocolatey :P. I'm convenced the reason the `npm` install was moved to alternatives is to avoid the jokes about using a package manager to install a package manager.

Daniel15:
> I assumed the npm route was moved to avoid jokes about using it to install another package manager.

That's one reason 😛  Here's one of my comments from a few weeks ago (#385) that explains some of the rationale:

> Advantages of using the system's native package manager rather than something like `npm install -g yarn`:
> 
> - Upgrades are handled the same way as all other software on your system (`apt-get upgrade`). It's not a special snowflake.
> - All the security problems with npm (eg. packages not being signed) make it less than ideal for anything installed system wide. apt-get gives you proper GPG signature verification.
> - Using npm means you're tightly-coupled to npm. If you later want to have a better installation solution, you need to have this awkward transition period where you try to get users to switch to the newer installation method.
> - The setup instructions need the whole weird "*install Node.js first, then ensure npm is up-to-date, then install Yarn*" thing. That'd be like telling people to manually install their npm dependencies 😆 . With a proper package management system, the Yarn package would simply have a dependency on the Node.js package, and "apt-get install yarn" would install both.
> 
> If we want to appeal to people that have limited Node.js experience, we shouldn't make any assumptions that they know anything about the Node.js ecosystem or what "npm" even is.

>  it's easy enough to remember npm i -g yarn compared to whatever 

[...]

PR #1117, opened 2016-10-16
Use Yarn package/installer on Travis/AppVeyor/CircleCI rather than npm

Daniel15:
**Summary**
On Linux:
- Install Yarn from Ubuntu package rather than `npm`
- Moved installation of Yarn to `bootstrap-env-ubuntu.sh` script. The more of the environment configuration we have in here, the better (as it'll make it easier to have Docker/Vagrant images for Yarn dev/build in the future)
- Use Travis and CircleCI's built-in method to choose Node.js v6 rather than nvm

On Windows (AppVeyor):
- Install Yarn from Chocolatey package

**Test plan**
Will check on Travis/AppVeyor/CircleCI once PR is submitted and the build is kicked off

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 ---
 git:
   depth: 10
-language: generic
+language: node_js
 
-sudo: false
+node_js:
+  - "6"
+
+sudo: required # Until Yarn repo is added to apt-source-whitelist
 
 addons:
   apt:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -1,9 +1,12 @@
...
-language: generic
+language: node_js
 
-sudo: false
+node_js:
+  - "6"
+
+sudo: required # Until Yarn repo is added to apt-source-whitelist
 
 addons:
```
jdalton: Why not wait until you get the whitelist to change things over?
Won't dropping sudo make the CI runs slower?

Related to #1082.

Daniel15: The whitelist seems to take a long time to get things into it, there's [225 issues and 40 pull requests](https://github.com/travis-ci/apt-source-whitelist) right now.

CircleCI is much faster than Travis and allows `sudo` by default, so we're mostly relying on Circle these days instead of Travis. Travis' container infra is more limited than Circle's, Circle even lets you SSH into the build boxes to diagnose build problems.

PR #1129, opened 2016-10-16
Add %LocalAppData%\Yarn\.bin to PATH as part of installation

Daniel15:
**Summary**
Updates the Windows installer to add `%LocalAppData\Yarn\.bin` to the `PATH`, so that we can remove the "path setup" section that was added in the docs (https://yarnpkg.com/en/docs/install)

**Test plan**
Ran installer on my computer, ensured the PATH was updated correctly:

Uninstalled Yarn, ensured both PATH additions were removed.

### DIFF
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -36,6 +36,15 @@
 					Action="set"
 					System="yes"
 				/>
+				<Environment
+					Id="BinPath"
+					Name="PATH"
+					Value="[LocalAppDataFolder]Yarn\.bin"
+					Permanent="no"
+					Part="last"
+					Action="set"
+					System="yes"
+				/>
 			</Component>
 		</DirectoryRef>
 
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -43,7 +43,7 @@
 					Permanent="no"
 					Part="last"
 					Action="set"
-					System="yes"
+					System="no"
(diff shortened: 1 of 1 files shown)


resources/winsetup/Yarn.wxs
```diff
@@ -36,6 +36,15 @@
...
 					System="yes"
 				/>
+				<Environment
+					Id="BinPath"
+					Name="PATH"
+					Value="[LocalAppDataFolder]Yarn\.bin"
+					Permanent="no"
+					Part="last"
+					Action="set"
+					System="yes"
```
Kovensky: This is a per-user path, so other users in the same system don't have access to it.

This needs to be set up for each user that runs yarn. This will work on single-user systems, though.

An advantage of making it not System is that it will roam with the user profile. While the local appdata folder won't roam, local installs with yarn will go to the local appdata, so they'll show up in the path once you create the folder with yarn.

Kovensky: Can you, maybe, set the path to be literally `%LocalAppData%/Yarn/.bin`, and hope the shell expands it when evaluating the `PATH`?

This, of course, will definitely not work on windows xp.

Daniel15: Good catch, thanks @Kovensky! I updated it to add it to the user environment variable rather than the system one. That should work fine. I don't think the shell expands environment variables like `%LocalAppData%` when resolving the PATH.

Issue #1128, opened 2016-10-16
yarn-offline-mirror is read from .npmrc, support .yarnrc


bestander:
**Do you want to request a _feature_ or report a _bug_?**

_feature_

**What is the current behavior?**

Putting `yarn-offline-mirror` setting in .yarnrc does not work, it is only read from .npmrc

**Please mention your node.js, yarn and operating system version.**
yarn 0.15.1

FLGMwt:
Should this be *removed* from .npmrc in favor of .yarnrc?

bestander:
I think it should but we need to be backwards compatible for some time

PR #1131, opened 2016-10-16
support yarn-offline-mirror in .yarnrc

zcei:
**Summary**

This PR allows the use of `yarn-offline-mirror` config option within `.yarnrc` as described in #1128.

Within `YarnRegistry` it resolves the path relative to the `.yarnrc` it is provided in.

Currently it looks for the option in all registries. This might not be wanted: please see the open questions.

**Test Plan**
- `.yarnrc` is provided within `fixtures/install/install-from-offline-mirror`.
- `.npmrc` has an invalid path to make sure `.yarnrc` is used.

**Open questions:**
- Should the use within `.npmrc` be considered deprecated?
- Should the deprecated usecase still be tested?
  - Duplicate the fixtures or remove / dynamically load `.yarnrc`?
- Should the config name stay equal? (`yarn-` prefixed within `.yarnrc`)
- Should there be a hierarchy for registry configs? e.g. `yarn` overwrites `npm`


src/registries/yarn-registry.js
```diff
@@ -72,18 +72,27 @@ export default class YarnRegistry extends NpmRegistry {
...
         this.homeConfig = config;
       }
 
+      // normalize offline mirror path relative to the current yarnrc
+      const offlineLoc = config['yarn-offline-mirror'];
+
+      // don't normalize if we already have a mirror path
+      if (!this.config['yarn-offline-mirror'] && offlineLoc) {
+        const mirrorLoc = config['yarn-offline-mirror'] = path.resolve(path.dirname(loc), offlineLoc);
+        await fs.mkdirp(mirrorLoc);
```
bestander: where is the code that does fs.mdirp for `yarn-offline-mirror` in .npmrc?

src/config.js
```diff
@@ -257,13 +257,25 @@ export default class Config {
...
-    const registry = this.registries.npm;
-    if (registry == null) {
-      return null;
+    //
+    let mirrorPath;
+
+    for (const key of Object.keys(this.registries)) {
+      const registry = this.registries[key];
+
+      if (registry == null) {
```
bestander: I believe this can't be null, so the check is redundant

__tests__/fixtures/install/install-from-offline-mirror/.npmrc
```diff
@@ -1 +1 @@
-yarn-offline-mirror=./mirror-for-offline
+yarn-offline-mirror=./invalid-mirror-for-offline
```
bestander: 👍

src/registries/yarn-registry.js
bestander: So should it be removed from there?

src/config.js
```diff
@@ -257,13 +257,20 @@ export default class Config {
...
-      return null;
+    //
+    let mirrorPath;
+
+    for (const key of Object.keys(this.registries)) {
+      const registry = this.registries[key];
+      const registryMirrorPath = registry.config['yarn-offline-mirror'];
+
+      if (registryMirrorPath == null) {
+        continue;
```
bestander: This code will return value from `yarn-offline-mirror` of the last key in this.registries iteration.
Afaik we don't get guarantee of iteration of Object keys so this may be random value from .yarnrc or .npmrc.

Should you just get this.registries.yarn and this.registries.npm explicitly?
And then order the priority manually yarn > npm.

zcei: This answers the last question - I also do think it makes sense.
Though, it introduces the `registry == null` check again, as we can't be sure that both registries do exist, am I right?

src/registries/yarn-registry.js
zcei: Wouldn't this then break backwards compatibility?
(See questions regarding deprecation)

src/config.js
bestander: Sure.
The check was not needed when you cycle over this.registries entries but if you replace the cycle with this.registries.npm and this.registries.yarn then you need to check for null

src/registries/yarn-registry.js
bestander: If you have  fs.mkdirp(mirrorLoc) here then you don't need second  fs.mkdirp(mirrorLoc) is https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L127.
This should not break compatibility

zcei: Okay I guess I didn't get your intention properly:

The `src/config.js` is where we access both registries. But the function is not `async` so I can't `await mkdirp` there.

Should I check in the YarnRegistry for whether there has been a config option in a different registry? This would conflict with not guaranteeing the iteration and thus creation order.

src/config.js
```diff
@@ -257,13 +257,25 @@ export default class Config {
    */
 
   getOfflineMirrorPath(tarUrl: ?string): ?string {
-    const registry = this.registries.npm;
-    if (registry == null) {
-      return null;
+    //
```
bestander: this comment is unnecessary

src/registries/yarn-registry.js
bestander: Ah, I just missed that YarnRegistry is descendant from NpmRegistry.
My trouble is that loadConfig is not symmetrical between these 2 Registries, one of them creates mkdirp the other one does not.

Feels a bit fragile because it is not clear which of the two is loaded first.
Should we just add `await fs.mkdirp(mirrorLoc);` to both?

bestander:
ping @zcei

zcei:
@bestander thanks for the reminder.

Basically removed the changes to npm registry, so it's awaiting `mkdirp` in both cases.

Some tests that seem to be not related are failing, is this the current state?

bestander:
Yes, you are right, we have a few tests broken in trunk https://circleci.com/gh/yarnpkg/yarn/1967

bestander:
@zcei could you rebase?
Trunk got fixed now

PR #1146, opened 2016-10-17
Support auth for private npm packages

### LINKED ISSUE #1134: Downloading Scoped + Private packages fails on npm-registry (Tarball Fetcher)

feature

Testing out the recently merged `_authToken` support. Package info is correctly downloaded.

But `npm-registry.js`  cannot determine the correct scope when downloading a package tarball as tarball fetcher passes in the full path so `getScope()` fails to determine the correct scope.

This means private scoped packages get the registry url switched to the yarn mirror. 

Tarball download respects the `always-auth` and `_authToken` configuration options when downloading packages.

master branch - 8f4f01e89

devongovett:
**Summary**

Sends the auth token for scoped packages, which may be private. Fixes #1134 and #521.

**Test plan**

Tested installing a private package on the public registry without `always-auth` in .npmrc.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -7,7 +7,7 @@ import type {ConfigRegistries} from './index.js';
 import * as fs from '../util/fs.js';
 import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import Registry from './base-registry.js';
-import {addSuffix} from '../util/misc';
+import {addSuffix, removePrefix} from '../util/misc';
 
 const defaults = require('defaults');
 const userHome = require('user-home');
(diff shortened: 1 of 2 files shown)


src/registries/npm-registry.js
```diff
@@ -53,7 +53,8 @@ export default class NpmRegistry extends Registry {
     const registry = addSuffix(this.getRegistry(pathname), '/');
     const requestUrl = url.resolve(registry, pathname);
     const alwaysAuth = this.getScopedOption(registry.replace(/^https?:/, ''), 'always-auth')
-                    || this.getOption('always-auth');
+                    || this.getOption('always-auth')
+                    || removePrefix(requestUrl, registry)[0] === '@';
```
cpojer: the code style should be to intend two spaces from `const`.

devongovett: Fixed :smile:

aliatsis:
Any timeline on a patch release for this (0.15.2)?

jedwards1211:
Is this released?  I'm still getting `invalid tar file` errors with 0.18.1 when I try to install private npm packages.

Issue #1142, opened 2016-10-17
"yarn" command should look for both "node" and "nodejs"


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Yarn always executes `node` (due to the shebang of `#!/usr/bin/env node`), and always executes `nodejs` on Debian/Ubuntu (we `sed` the shebang when building the package: https://github.com/yarnpkg/yarn/blob/master/scripts/build-deb.sh#L74).

The tricky/unfortunate thing with Node.js on Debian is that `/usr/bin/node` conflicted with another package. See:
https://lists.debian.org/debian-devel-announce/2012/07/msg00002.html
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=614907

In the end, both packages that used the `node` binary name were renamed - Node.js became `nodejs`, and the other Node (some ham radio app) became `ax25-node`. This means that when installing Node.js via Debian package, the binary is called `nodejs`. However, for other installation methods (such as `nvm`), the binary is called `node`. 

Installing both Yarn and Node.js via Debian package is generally the preferred method, and works fine. However, this  naming does have several implications when mixing and matching different installation styles:
 - Yarn installed via `npm` will fail with Node.js installed via Debian package (it'll try to run `node` which doesn't exist)
 - Yarn installed via Debian package will fail with Node.js installed via nvm (it'll try to run `nodejs` which doesn't exist)

Often people work around this by symlinking `/usr/bin/node` to `/usr/bin/nodejs` or vice-versa, but that's hacky. 

Instead, we should update `bin/yarn` to try both `node` and `nodejs`. For example, try to execute `node yarn.js`, and if that fails, try to execute `nodejs yarn.js` instead.

Also see https://github.com/yarnpkg/rfcs/issues/9

vTNT:
+1

danez:
Maybe something similar to what npm is doing? https://github.com/npm/npm/blob/master/bin/npm

PR #1157, opened 2016-10-17
Added yarn install to circle.yml

bestander:


bestander:
failed because of https://github.com/yarnpkg/yarn/issues/1142

danez:
Another workaround would be to install node-legacy which basically is doing the symlink, see https://github.com/nodejs/node-v0.x-archive/issues/3911#issuecomment-18951288

bestander:
I think I got some reasonable compormise :)

bestander:
@cpojer virtually stamped it

Daniel15:
Hmm, CircleCI should have Node.js installed via apt-get and so /usr/bin/nodejs should exist. Does it not?

bestander:
Circle installs node here: /opt/circleci/nodejs/v6.8.1/bin/node.
It looks like `nvm` script is used so I don't think it uses debian package manager to isntall node.
We don't have permissions to do `ln -s /usr/bin/nodejs /usr/bin/node` or something similar

PR #1162, opened 2016-10-17
created markdown based documentation for installation of Yarn as well…

tsnodgrass96:
**Summary**

I noticed that Yarn did not have any documentation outside of the website, so I went through and created a markdown version of the getting started documentation. This includes installation, basic usage, npm migration, and more advanced package work and publishing. This is meant to be a way for users to have access to the Yarn documentation either offline or directly from the github repo. 

For reference, all additions are contained to a docs folder in the root of the project

**Test plan**

No changes were made to how Yarn looks, operates, is built, or in any manner that could cause issues with an installation on a new system. This pull request is merely documentation.

Thanks for taking the time to review this, hope you enjoy the rest of your day.


Daniel15:
The issue here is that people updating the docs now need to update **two** places - The docs in this repo, and the docs on the site. I'd much rather we only had a single source of truth. If we want to ship docs with Yarn, I think we should build them based on the pages in the website, rather than having totally separate documentation. What do you think?

What's the advantage of bundling the documentation with Yarn itself? Is it just so people can read the docs without internet connectivity?

> directly from the github repo.

All the pages of the website are available in the website repo at https://github.com/yarnpkg/website

tsnodgrass96:
That's true I didn't think of that. My thought behind having it in the repo was that you wouldn't have to go to another site if you went to github looking for info on how to get started. Now this is just a thought but we could create a CLI tutorial/tool that fetches all of it's info from the website docs that are existing. It would accomplish the same goal and could walk the user through creating the example project. What are your thoughts on that?

markstos:
At the moment, when there is a rapid pace of change and development happening, maintaining two copies of docs is undesirable and may cause as many problems as it solves. Once changes becomes less frequently, maintaining docs in two places shouldn't be that bad. But if website and CLI docs can be generated from the same source-- even better. 

Since `yarn` emphasizes it's ability to work offline, it makes sense to have some offline-by-default docs as well. 

It's a bit off-putting for the CLI docs to refer to me website and breaks up the flow of terminal-based work.

tsnodgrass96:
I've worked on a few projects where we have taken the online docs and created a node project to build a simple docs page using electron so that is an option but with `yarn` being CLI based it just makes sense to add some sort of CLI version of the docs that can accompany `yarn` for a quick reference. 

With most software it just seems like even a --help are too vague and difficult to sort through to find what you need.

Daniel15:
What if we bundle `.html` files built from the website in a `docs` directory in the tarball, and install them to `/usr/share/doc/yarn/` for the Linux packages? We could include some manpages too. Would that solve your use case?

> with yarn being CLI based it just makes sense to add some sort of CLI version of the docs that can accompany yarn for a quick reference.

Would manpages help, or would you prefer totally plain text files? I don't think raw markdown is very good for reading at the command line.

markstos:
Good online help is one thing `npm` gets right. Their help for CLI commands is written in Markdown and can be seen here: https://github.com/npm/npm/tree/master/doc/cli

If you run `npm help ls` It displays the related Markdown page: https://github.com/npm/npm/blob/master/doc/cli/npm-ls.md

`npm` uses `man` as the document viewer. Perhaps there's a reason to use that for displaying documentation over something like `less`. I'm not sure.

Daniel15:
> npm uses man as the document viewer. Perhaps there's a reason to use that for displaying documentation over something like less. I'm not sure.

`man` is the standard way of viewing documentation on Linux and Linux-like operating systems. For the Debian and Ubuntu packages, we should bundle manpages, so you'd be able to run `man yarn` to read the docs. Does Mac OS also come with `man`?

> without wading through a very lengthy single "man" page 

I think we'd have a separate manpage per command. See how Git does it for example (eg. `man git-commit`, `man git-add`)

hawkrives:
@Daniel15 Yes, Mac OS X comes with `man`.

markstos:
Yes to having a main "main" page and then one page per command. I see with `npm`, both these work and produce the same result:

      man npm-ls
      npm help ls

(`git` supports both ways to access docs as well).

kittens:
Closing this as we don't want duplicate documentation in this repo that's available on the main website. We ran into similar issues when maintaining documentation on Babel and it leads to a poor experience all around.

markstos:
@kittens while this PR should be closed because it duplicates documentation, the same goal of better docs available right in the CLI could be met without duplicate. The build system  in the `yarn` repo could pull in docs from the `website` repo so that `yarn help add` would display the `add.md` file from the website repo in addition to the basic "usage" information that's already there. Perhaps while this PR is being closed, a related Issue should be open to continue discuss other ways to improve the CLI documentation experience without duplicating docs unnecessarily.

Daniel15:
Created a separate issue for it: #1603. Thanks @markstos 😄

Issue #1051, opened 2016-10-14
"yarn clean" / ".yarnclean" breaks svgo


marzepani:
`svgo`requires `.svgo.yml` and is a dependency of popular `gulp-svgmin`.

The clean command (and the resulting `.yarnclean`) keep removing hidden files and therefore break `svgo`.

I know [you warned us](https://yarnpkg.com/en/docs/cli/clean), but maybe an option to exclude files from being cleaned would be nice.

What about listing files to ignore in `.yarnclean`, like we do in `.gitignore`. Or something smarter. :)

(node 6.8.0, yarn 0.15.1, macOS)

zcei:
There already is [a way to ignore files](https://github.com/yarnpkg/yarn/blob/271ef9a9d021e59fbced660989d1fa5589413eca/src/util/filter.js#L44) & folders within yarn.

Basically it's a [minimatch](https://github.com/isaacs/minimatch) regex, and the leading `!` indicates to keep this path.

e.g. `!.svgo.yml`

marzepani:
Thank you @zcei for investigating.

But there is a big issue with this, as `yarn clean` creates a blank new `.yarnclean` file every time. Added filter rules will be wiped. :-(

The only way it works right now:

1. Manually create `.yarncean` file.
2. Add filter rules, like `!.svgo.yml`
3. Add modules, like `yarn add svgo --dev`
4. **Never** call `yarn clean`, as all filters will be lost and modules will be cleaned –> you are doomed. ;-)

zcei:
Oh well, this is not good for obvious reasons 😄 
Hope your PR gets merged soon 👍

kittens:
Fixed via #1430, the default filters are now added to `.yarnclean` and you can remove specific rules if you wish.

Issue #1008, opened 2016-10-13
--clean breaks some packages


prakharsingh:
I have migrated to yarn yesterday and npm package browser-sync-ui gets broken by using "yarn --clean".
yarn deletes history plugin file from the module.

Yarn version v0.15.1
OS Ubuntu 14.04 LTS

marzepani:
I can confirm, as the same happened to me.
`yarn clean` removed `node_modules/browser-sync-ui/lib/plugins/history`

yannickoo:
You can specify files in the `.yarnclean` file, check [this](https://github.com/yarnpkg/yarn/issues/1051#issuecomment-254334628) comment 😇

marzepani:
Thx, but beware. [See details](https://github.com/yarnpkg/yarn/issues/1051#issuecomment-254438695).

busches:
Can you reproduce this with `yarn@latest`? If yes, can you supply a `package.json` file that recreates the issue?

hpurmann:
I'd vote for closing this if it's no longer reproducible.

PR #1188, opened 2016-10-18
Don't create .yarnclean if it already exists

marzepani:
**Summary**

The file `.yarnclean`should not be created if it exists. Otherwise filter rules are not usable.

Related issues:

- [1008](https://github.com/yarnpkg/yarn/issues/1008)
- [1051](https://github.com/yarnpkg/yarn/issues/1051)

**Test plan**

Tests and linting passed.


bestander:
Thanks, @marzepani

Issue #788, opened 2016-10-12
Add `--silent` flag to CLI run command


pgrippi:
**Do you want to request a _feature_ or report a _bug_?**

feature

**What is the current behavior?**

I use the `npm test` command in my ember cli projects to run my tests. I was happy to see that `yarn test` works the same way as `npm test` but it's missing one feature that I use: the `--silent` flag.

When I run `yarn test --silent` (or `yarn run test --silent`) it prints out the yarn version and command it runs:

```
$ yarn run test --silent
yarn run v0.15.1
$ ember test
...
```

**What is the expected behavior?**

I would expect that when I run `yarn run <cmd> --silent` it would suppress the output of the yarn version and the command it is running, similar to how `npm run --silent` works.

**Please mention your node.js, yarn and operating system version.**

Node: v4.6.0
Yarn: v0.15.1
OS: MacOS Sierra 10.12.0

FLGMwt:
Talked a bit with @cpojer on discord about exactly how silent something like this should be and he convinced me at least that this might not be a necessary thing for a cli to provide. Nothing that I could think of can't be handled by a `>/dev/null 2>&1` if you *really* don't care about the output.

Counterpoints?

pgrippi:
Ah, but I do care about the output. My use case here is that I'm running the test on a CI server and the output is an xml file. My command looks something like this:

```
yarn test --silent -- --other-args > junit.xml
```

Since yarn prints out the version and command the xml file generated by my test script has that in it before the xml output and I'm left with an invalid xml file.

robotvert:
Also, terminal colors mess up the output when doing deploys here. Maybe it's something that should be addressed in the way my deploys are done but a simple `--silent` flag would avoid stuff like:

```
[2K[1G[1myarn install v0.15.1[22m
[2K[1G[90m[1/4][39m Resolving packages...
[2K[1G[2K[1G[2K[1G[90m[2/4][39m Fetching packages...
[2K[1G[1G 0/790[1G 5/790[1G 226/790[1G 745/790[2K[1G[2K[1G[33mwarning[39m fsevents@1.0.14: The platform "linux" is incompatible with this module.
[2K[1G[34minfo[39m "fsevents@1.0.14" is an optional dependency and failed compatibility check. Excluding it from installation.
[2K[1G[90m[3/4][39m Linking dependencies...
[2K[1G[33mwarning[39m Unmet peer dependency "cheerio@0.19.x || 0.20.x || 0.22.x".
[...]

PR #1190, opened 2016-10-18
Disable progress

ldabiralai:
**Summary**

I've added `--disable-progress` as it's useful to be able to turn off progress bars on CI environments (like CircleCI) otherwise while using yarn to install packages you have a lot of unnecessary info cluttering the view:

image

It also disables progress for the spinning progress icons.

**Test plan**

I've added a test to `console-reporter.js` to ensure the feature works as desired.

Let me know what you think, thanks!


inikulin:
I believe it can be automatically disabled in CI environments without introduction of new options, see: https://github.com/watson/is-ci

ldabiralai:
Thanks @inikulin but using that would limit this too much, in my opinion.

inikulin:
> Thanks @inikulin but using that would limit this too much

Can you elaborate, please?

ldabiralai:
@inikulin Sure. I don't think we should force hiding it on CI, nor should it be an option only on CI. I think it should be an option that people can choose if they want to use it. In the same way ` npm set progress=false` works.

samccone:
👍

src/cli/index.js
```diff
@@ -62,6 +62,10 @@ commander.option(
   '--no-emoji',
   'disable emoji in output',
 );
+commander.option(
+  '--disable-progress',
```
sindresorhus: The CLI convention is to use `no-` for negated flags, so this should be `--no-progress`.

sindresorhus:
I don't see the point of a flag when it can be handled automatically for the user.

ldabiralai:
I felt the flexibility the current implementation offers is good.

Yarn isn't widely adopted in CI yet, so everyone will be manually running `yarn`. I didn't see it as a big cost to add a flag if you wanted to disable the progress bar. As it wouldn't require other changes to your CI config files.

Can change it though if the consensus is we want it to be restricted to CI as well as forced.

inikulin:
> Yarn isn't widely adopted in CI yet, so everyone will be manually running yarn. I didn't see it as a big cost to add a flag if you wanted to disable the progress bar. As it wouldn't require other changes to your CI config files.

So we'll have:
```
1. Figure out that yarn output is broken on your CI.
2. Try to find recipes to workaround it in documentation/issues.
3. Finally find `--no-progress` option
4. Modify your CI configuration file and commit them or issue a PR and go through review process.
5. Get it work as expected.
```

VS
```
1. It works properly instantly.
```

I'm not against an option. We can keep it, but disable progress on CI by default. That's why I ask you a question in https://github.com/yarnpkg/yarn/pull/1190#issuecomment-254502925: are you aware of any scenarios there disabling progress on CI can be harmful?

ldabiralai:
Fair enough, seems a number of people like it that way too! I'll make the change to add https://github.com/watson/is-ci and keep the option for non-ci too.

inikulin:
Seems like console reporter tests require some rework now

kittens:
We should just have detection, I'm against having another flag for this.

designfrontier:
I am with the two comments above me. Not all CO is going to be covered by `is-ci`. If the argument is to make it easy do both. Detect when possible _and_ provide a flag for users not in detectable ci environments.

kittens:
If you don't want nice terminal output then don't report that you're a TTY when executing yarn. An explosion of user flags is not desirable. This PR will only be accepted without the CLI flag.

daveadams:
I can't reconcile the pushback here with #922 from a week ago. There are always going to be situations where "nice" output is undesireable even if it's running in a TTY. In particular I'm thinking about accessibility use cases, where TTY functionality is often useful, but for which animated progress spinners are problematic. But also, alas, purchased tools that capture output using a TTY and don't give you the option.

jakebasile:
Adding CLI switches to customize output isn't uncommon. Examples

- `curl --silent`
- `apt-get --quiet`
- `git --no-color`
- `rpm --quiet`
- `jq --color-output`
- `nom --no-color`

Sometimes people want to override how output is displayed irrespective of things the utility can automatically detect. All the fancy output here is great if you like eye candy, but some people just want (or need, as @daveadams pointed out wrt accessibility) to strip that stuff out.

ldabiralai:
I've fixed the tests. Progress is disabled on CI and the `--no-progress` flag is still present.

I don't think we should be removing the flag as with the previous two comments, and there also seems to be demand for it in #1340 #788 #728

whitecolor:
@kittens 
> This PR will only be accepted without the CLI flag.

What should it be then? 

I vote for config option ala `yarn config set progress=false` this is more desirable in my view, especially when working with containers.

PR #1192, opened 2016-10-18
Add a newline after the global comment in the lockfile

TomasHubelbauer:
The lockfile is nicely visually separated except for the first package which is right after the comment. The comment is global, though, not related directly to the first package so in order to make it easier to visually navigate the file and to mark the comment as global and not just related to the first package, I propose a newline is added after the comment.

If this seems frivilous I understand, feel free to close the PR - I just think this is nicer and it complements the niceness of the CLI outputs.

This needs to be accompanied by a change to the parser. I have noticed manually introducing this change to a lockfile and then running `yarn` results in a refetch whereas not changing the file and running `yarn` reuses cached stuff. This is probably because the integrity check in `node_modules` is calculated from the content of the lockfile, but I haven't look into this enough to be able to verify that. The simple fact remains, the parser needs to be updated as well. 

**I will try to deliver that change later if I find  time but I welcome anyone to do it as well.**


TomasHubelbauer:
The checks have passed which means the parser doesn't have to be updated I guess? Maybe the reason `yarn` refetches after manually updating `yarn.lock` is it checks the file date? I am not sure.

PR #1200, opened 2016-10-18
fix #1199 by removing extra newline from cli/install.js

### LINKED ISSUE #1199: yarn.lock contains an extra new line at the end
After running `yarn install`, my yarn.lock file contains two new line characters at the end of the file. It only needs to contain one. In particular, my repo has a pre-push hook that checks for whitespace issues using `git diff --check origin/master..head` and fails if that command exits with a non-zero exit code. The extra new line generated in yarn.lock causes this check to fail.

lzilioli:
**Summary**

Issue #1199

**Test plan**

Run yarn => verify yarn.lock only contains one new line

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -467,7 +467,7 @@ export class Install {
 
   async saveLockfileAndIntegrity(patterns: Array<string>): Promise<void> {
     // stringify current lockfile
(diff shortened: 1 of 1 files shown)


samccone:
This seems reasonable, 👍

lzilioli:
@samccone are there additional steps I need to take to get this merged to master and released?

samccone:
I am going to wait for @bestander or @kittens for the +1 here since there may be a reason for this that I am unaware of.

PR #1225, opened 2016-10-19
Remove hack for #1142

Daniel15:
Removes this hack from the `circle.yml` file now that the executable supports both `node` and `nodejs`.


Daniel15:
AppVeyor issue seems unrelated and this fixes CircleCI, I'm going to merge it.

Issue #1228, opened 2016-10-19
"yarn version" should have an option to not commit to Git


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
`yarn version --new-version 1.2.3` always commits to Git. There's an ugly workaround: You can set the `YARN_VERSION_GIT_TAG` environment variable to an empty string:
```
YARN_VERSION_GIT_TAG='' yarn version --new-version 1.2.3
```

This is because [version.js](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/version.js#L110) uses `Boolean(config.getOption('version-git-tag'))`, and [base-registry.js](https://github.com/yarnpkg/yarn/blob/master/src/registries/base-registry.js#L92) populates overrides for these config options from the environment variables. Since it uses the `Boolean` function and all environment variables are strings, the only value that works to disable it is an empty string.

**What is the expected behavior?**
A `--no-commit` or `--no-git-tag-version` flag: Something like `yarn version --new-version 1.2.3 --no-commit` should update the version number without committing it. [npm calls this --no-git-tag-version](https://docs.npmjs.com/cli/version).

karanjthakkar:
@Daniel15 I'll give this a shot today!

kylecombes:
Does this still work the same way as it did after #1243 was merged? Using `--no-git-tag-version` doesn't seem to be working for me...

PR #1230, opened 2016-10-19
Set unique version number for CI (CircleCI and AppVeyor) builds

Daniel15:
**Summary**
This sets a version number of the format `{version}-alpha.{date}` (eg. `0.16.1-alpha.20161019.2210`) when building master on CircleCI and AppVeyor. This allows us to differentiate stable builds from unstable builds, and is required to allow us to offer unstable/nightly builds as downloads (which I'll set up soon).

Unfortunately [semver doesn't support nightly builds](https://github.com/mojombo/semver/issues/200) so this version syntax is probably the best we can do. I used `alpha` rather than `nightly` so that it sorts **before** `beta`.

**Test plan**
Tested on my fork of the repo: https://circleci.com/gh/Daniel15/yarn/24
Check "artifacts" tab in CircleCI, ensure artifacts have correct version number:

Install the resulting Debian package, ensure version is correct:


scripts/build-ci.sh
```diff
@@ -0,0 +1,22 @@
...
+set -ex
+
+pushd `dirname $(readlink -f $0)/`/..
+
+if [ "$CIRCLE_BRANCH" == 'master' ]; then
+  original_version=`node -p "require('./package.json').version"`
+  # Set a nice build number when building master on CircleCI
+  # (eg. "0.16.0-alpha.123+20161018")
+  version="$original_version-alpha.$CIRCLE_BUILD_NUM+"`date +%Y%m%d`
+  YARN_VERSION_GIT_TAG='' yarn version --new-version $version
```
Daniel15: Created #1228 for the ugly `YARN_VERSION_GIT_TAG` thing.

circle.yml
```diff
@@ -27,13 +27,7 @@ dependencies:
 
 test:
   override:
-    - node -v
-    - npm run lint
```
bestander: The advantage of having those steps one by one in .yml file is nice wrapping.
If npm run lint fails then it takes less than a second to figure out.

However if we put everything in one shell script looking up the reason to fail will involve a lot of scrolling.
screen shot 2016-10-19 at 12 04 22 pm

scripts/build-ci.sh
bestander: Instead of $CIRCLE_BRANCH == 'master' you can use "branches" in deployment area https://github.com/facebook/react-native/blob/master/circle.yml#L97.

The nice thing about it is that it runs only when all Test phases are green.
Also does `yarn version` add a tag to git checkout?
I think it might not work becuase most likely CI checkouts have read-only access to repos.
That is why we store a special write token for RN tagging https://github.com/facebook/react-native/blob/master/circle.yml#L102

bestander: Also you probably don't need to tag the repo, just run the deployment section of cicle.yml for all master branch commits

circle.yml
Daniel15: Good point! I'll move this stuff back into the `circle.yml`!

Daniel15:
This does not tag the Git repo - I'll add a comment explaining that.

> rely less on shell scripts

Well, technically commands we run in `circle.yml` are also a shell script 😛

scripts/set-dev-version.js
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
+/**
+ * Sets a a nice build number when building nightly builds
+ * (eg. "0.16.0-alpha.20161019.1800")
+ */
+
+const fs = require('fs');
+
+function leftPad(value) {
```
Daniel15: I was very tempted to import the `left-pad` package here. Hahaha

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "yarn",
-  "version": "0.16.0",
+  "version": "0.16.2",
```
Daniel15: We should get into the habit of bumping this after every release... The dev build number should always be greater than the release build number. We can later swap out the version number if we do a minor release (eg. 0.16.2 -> 0.17.0)

bestander: I wonder, should we?
I was thinking about this earlier.

Our current release branching mechanism involves creating a branch from master.
After that, release branch and master live on their own.
We may publish patches to release branch by incrementing patch version but there is no point in linking release branch's patch to master because they are already most likely diverged.

There is still something wrong with master branch being 0.16.0 until 0.17.0 is cut.
Maybe we should address this to indicate that 0.16.0 on master is actually not in sync with release branch.

bestander:
Yeah, great idea!

PR #1243, opened 2016-10-19
Add an option to not commit to git. Fixes #1228

### LINKED ISSUE #1228: "yarn version" should have an option to not commit to Git

karanjthakkar:
**Summary**

Fixes #1228 

**Test plan**

1. Committing without the `--no-git-tag-version` flag does not commit to github
screen shot 2016-10-19 at 4 30 07 pm

2. Committing with the `--no-git-tag-version` flag commits to github
screen shot 2016-10-19 at 4 30 23 pm

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -20,6 +20,7 @@ function isValidNewVersion(oldVersion: string, newVersion: string, looseSemver: 
 export function setFlags(commander: Object) {
   commander.option(NEW_VERSION_FLAG, 'new version');
   commander.option('--message [message]', 'message');
+  commander.option('--no-git-tag-version', 'no git tag version');
 }
 
 export async function setVersion(
@@ -92,6 +93,13 @@ export async function setVersion(
     }
   }
   await config.saveRootManifests(manifests);
+
+  // check if committing the new version to git is overriden
+  if (!flags.gitTagVersion) {
+    return function(): Promise<void> {
+      return Promise.resolve();
+    };
+  }
 
   return async function(): Promise<void> {
     invariant(newVersion, 'expected version');
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -95,7 +95,8 @@ export async function setVersion(
   await config.saveRootManifests(manifests);
 
   // check if committing the new version to git is overriden
-  if (!flags.gitTagVersion) {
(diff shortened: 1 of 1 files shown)


Daniel15:
There's already a config flag that toggles this (`Boolean(config.getOption('version-git-tag'))`), maybe combine it with your new option - For example:
```
if (!flags.gitTagVersion || !Boolean(config.getOption('version-git-tag'))) {
  // Don't tag the version in Git
}
```

Maybe you could return early from the async function itself, rather than returning a dummy function that returns an empty promise?

karanjthakkar:
@Daniel15 Just made some changes. Is that what you were expecting? Or should I just change

`if (isGit && Boolean(config.getOption('version-git-tag'))` to:

`if (isGit && Boolean(config.getOption('version-git-tag')) && flags.gitTagVersion)`

?

karanjthakkar:
There seem to be multiple ways to skip the committing stage so I was wondering which one is right. But I guess in my current implementation the `postVersion` lifecycle script won't be called 🤔

karanjthakkar:
@Daniel15 Bump! Waiting on you to finalise the impl. :)

Daniel15:
Looks good to me!

Daniel15:
Looks good to me! I'd like someone with more experience with the Yarn codebase (like @sebmck) to take a look too 😄

bestander:
I think this is more useful than the config option, let's merge it

PR #1246, opened 2016-10-19
moved ghr out of project

### LINKED ISSUE #1203: [infra] install bundle is too large
.tar.gz contains `ghr` from CircleCI.
We need to download it in parent folder to omit it from build

bestander:
fixes #1203.
ghr should be downloaded outside of the project source

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -28,7 +28,6 @@ dependencies:
         curl -L --create-dirs -o ~/ghr/ghr.zip https://github.com/tcnksm/ghr/releases/download/v0.5.0/ghr_v0.5.0_linux_amd64.zip;
         unzip ~/ghr/ghr.zip -d ~/ghr
       fi
-    - ln -sF  ~/Downloads/ghr /usr/local/bin/ghr
 
 test:
   override:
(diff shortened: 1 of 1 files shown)


bestander:
Let me try but I am not sure I have access to /usr/local/bin, and

bestander:
@Daniel15, looks like not allowed to because of Circle setup https://circleci.com/gh/yarnpkg/yarn/1561?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Daniel15:
We have permission to use apt-get which installs executables in /usr/bin,
so /usr/local/bin should be fine as long as you use sudo

bestander:
@Daniel15 that worked :)

Daniel15:
This looks good! Thank you 😄

kittens:
There's a lot of complexity here, why can't we just add `ghr` and `ghr.zip` to `.npmignore`?

Daniel15:
Ha, I think that'd work too. The advantage in this diff is that ghr is cached so it won't be downloaded again every time.

bestander:
Yeah, adding ghr.zip and ghr to gitignore or npmignore would be easier but I didn't want to leak this specific tool we install only on Circle to leak into the repo in some way.
Also as Daniel noted, ghr is cached now and should not be downloaded on every build

PR #1209, opened 2016-10-18
Git ignore .bin files in request-cache fixtures

kaicataldo:
**Summary**

I'm seeing a number of files that are generated by tests in the master branch that look like they should be Git ignored but aren't. Running `git status` after running `npm test` in the clean master branch yields the following:
```
~/code/yarn master*
=> git status
## master...origin/master
?? __tests__/fixtures/request-cache/GET/localhost/
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/abab.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/abbrev.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/acorn-globals.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/acorn.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/align-text.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/amdefine.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/ansi-escapes.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/ansi-regex.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/ansi-styles.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/ansicolors.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/append-transform.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/aproba.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/are-we-there-yet.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/argparse.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/arr-diff.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/arr-flatten.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/array-differ.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/array-equal.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/array-union.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/array-uniq.bin
?? __tests__/fixtures/request-cache/GET/registry.npmjs.org/array-unique.bin
[...]

PR #1250, opened 2016-10-19
Check in request cache .bin files

kaicataldo:
**Summary**

Following up on https://github.com/yarnpkg/yarn/pull/1209. Running `npm test` on the current master branch generates `.bin` files that are not checked in or ignored. My initial PR added them to the `.gitignore`, but as @bestander [explained](https://github.com/yarnpkg/yarn/pull/1209#issuecomment-254785991), they should actually be checked in.

**Test plan**

Steps to create this PR were just pulling the latest from master, running `npm test`, and committing the result.

Let me know if there's something else you need!


bestander:
Thanks, @kaicataldo

Issue #1118, opened 2016-10-16
Error: Cannot find module 'user-home'


noahsw:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
Ran into trouble install Yarn using apt-get on Ubuntu 14.04 (CircleCI)

```
sudo apt-key adv --keyserver pgp.mit.edu --recv D101F7899D41F3C3
"deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
sudo apt-get update -y
```

```
ubuntu@box1015:~$ node -v
v4.5.0
ubuntu@box1015:~$ sudo apt-get install yarn
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following package was automatically installed and is no longer required:
  libc-ares2 libv8-3.14.5 os-prober
Use 'apt-get autoremove' to remove it.
Recommended packages:
  nodejs
The following NEW packages will be installed:
  yarn
0 upgraded, 1 newly installed, 0 to remove and 84 not upgraded.
Need to get 1,546 kB of archives.
After this operation, 23.2 MB of additional disk space will be used.
Get:1 http://dl.yarnpkg.com/debian/ stable/main yarn all 0.15.1-2 [1,546 kB]
Fetched 1,546 kB in 0s (26.2 MB/s)
Selecting previously unselected package yarn.
(Reading database ... 341963 files and directories currently installed.)
Preparing to unpack .../archives/yarn_0.15.1-2_all.deb ...
Unpacking yarn (0.15.1-2) ...
Setting up yarn (0.15.1-2) ...
ubuntu@box1015:~$ export PATH="$HOME/.yarn/bin:$PATH"
ubuntu@box1015:~$ which yarn
/home/ubuntu/.yarn/bin/yarn
ubuntu@box1015:~$ yarn --version
module.js:327
    throw err;
    ^

Error: Cannot find module 'user-home'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/ubuntu/.yarn/bin/yarn.js:8:16)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
```

[...]

PR #1252, opened 2016-10-19
Added try/catch block to .yarnclean write file operation

marzepani:
**Summary**

Problem: 
I forgot to add a try/catch block to the changed code in my pull request [#1118](https://github.com/yarnpkg/yarn/pull/1188)

Solution: 
Added try catch block to handle error „EEXIST: file already exists“.

**Test plan**

The code passed all automatic tests and the following manual tests:

1. `yarn clean` without an existing `.yarnclean` generates empty `.yarnclean` file.
2. `yarn clean` with an existing `.yarnclean` follows contained exclude rules and does not change the exiting `.yarnclean` file.

screen-shot-2016-10-19-at-16 05 34


src/cli/commands/clean.js
```diff
@@ -120,7 +120,9 @@ export async function run(
   args: Array<string>,
 ): Promise<void> {
   reporter.step(1, 2, reporter.lang('cleanCreatingFile', CLEAN_FILENAME));
-  await fs.writeFile(path.join(config.cwd, CLEAN_FILENAME), '\n', {flag: 'wx'});
+  try {
+      await fs.writeFile(path.join(config.cwd, CLEAN_FILENAME), '\n', {flag: 'wx'});
+  } catch (e) {}
```
bestander: can you add a comment why catch is suppressed?

marzepani: Catch is suppressed, because this error does not reflect a problem. The command will execute just fine. As checking with fs.exists is deprecated, this is a valid solution, I guess.

marzepani: Oops, sorry. :-)  I will add some …

marzepani: done.

PR #1273, opened 2016-10-19
Make registry environment variables case agnostic

AdamEdgett:
**Summary**
To ignore the case of environment variable keys, eg use of `NPM_CONFIG_REGISTRY` instead of `npm_config_registry`

As exists, yarn only uses environment variables with lowercased keys
For npm, [environment variables are case insensitive](https://docs.npmjs.com/misc/config#environment-variables)

**Test plan**
Before:

```
> export NPM_CONFIG_REGISTRY=https://custom.npm.registry
> yarn install
yarn install v0.16.0
info No lockfile found.
error Couldn't find package "privately-hosted-package" on the "npm" registry.
> export npm_config_registry=https://custom.npm.registry
> yarn install
...
success Saved lockfile.
```

After:

```
> export NPM_CONFIG_REGISTRY=https://custom.npm.registry
> yarn install
yarn install v0.16.0
info No lockfile found.
[2/4] 🚚  Fetching packages...
success Saved lockfile.
```

### DIFF
--- a/src/registries/base-registry.js
+++ b/src/registries/base-registry.js
@@ -119,18 +119,18 @@ export default class BaseRegistry {
 
   mergeEnv(prefix: string) {
     // try environment variables
-    for (let key in process.env) {
-      key = key.toLowerCase();
+    for (const envKey in process.env) {
+      let key = envKey.toLowerCase();
 
       // only accept keys prefixed with the prefix
-      if (key.indexOf(prefix) < 0) {
+      if (key.indexOf(prefix.toLowerCase()) < 0) {
         continue;
       }
 
-      const val = BaseRegistry.normalizeConfigOption(process.env[key]);
+      const val = BaseRegistry.normalizeConfigOption(process.env[envKey]);
 
       // remove config prefix
-      key = removePrefix(key, prefix);
+      key = removePrefix(key, prefix.toLowerCase());
 
       // replace dunders with dots
       key = key.replace(/__/g, '.');


src/registries/npm-registry.js
```diff
@@ -136,6 +136,10 @@ export default class NpmRegistry extends Registry {
   }
 
   getRegistry(packageName: string): string {
+    if (process.env.NPM_CONFIG_REGISTRY) {
```
samccone: nit !== undefined instead of depending on loose type coercion please.

AdamEdgett: Would have to check for null as well in that case

Would a typeof === 'string' check be better?

dkoston:
Any activity here? Should I submit another pull with this functionality? Thanks

rzteux:
Waiting for this pull request to be merged. It would be great to have this, since I'm using 'npm --reg' option in a regular basis for my daily npm publish at work. This is blocking to switch to yarn till now ;)

AdamEdgett:
I think I just need to fix up the tests
I'll poke at them ASAP

dkoston:
Thanks @AdamEdgett looks like this test is failing: https://ci.appveyor.com/project/kittens/yarn/build/1061/job/vtd8xjy9aw2ipdby#L600

AdamEdgett:
Sorry for the delay!
This actually seems to be fixed in the most recent stable version of yarn (0.19) if you use lowercased env variable name (`npm_config_registry`)
Might be able to modify this PR to just be more case-agnostic with the name

bestander:
@AdamEdgett, go ahead and update the PR, please, if case-agnostic is the way npm works

AdamEdgett:
@bestander updated

Issue #1274, opened 2016-10-19
Reversion bug: yarn 0.16 no longer able to install scoped packages from private registry


tyrsius:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
`yarn` fails with

```
Trace: 
  TypeError: Cannot use 'in' operator to search for 'latest' in undefined
      at /usr/local/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:62:24
      at next (native)
      at step (/usr/local/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /usr/local/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
      at new Promise (/usr/local/lib/node_modules/yarn/node_modules/core-js/library/modules/es6.promise.js:191:7)
      at /usr/local/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12
      at Function.findVersionInRegistryResponse (/usr/local/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:72:7)
      at /usr/local/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:89:34
      at next (native)
      at step (/usr/local/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
I have a private registry, which also has a scope. 

```
$ yarn config get registry
http://artifactory.nike.com/artifactory/api/npm/npm-nike//
```

```
"dependencies": {
    "@nike/lambda-node-logger": "^0.2.0"
  }
```

Yarn `0.15.1` can handle this, but yarn `0.16.(0|1)` fails. 

**What is the expected behavior?**
Installation should succeed

**Please mention your node.js, yarn and operating system version.**
```
Node: 5.7.1
yarn: 15 and 16
OS: Mac Sierra
```

jcmorrow:
I'm experiencing this as well. If I debug the `body` received on line 60 of `npm-resolver.js` it looks like the response is: 

```
package-name@1.0.0{ errors: [ { status: 401, message: 'Invalid basic authentication token' } ] }
```

wyze:
Solved by #1356.

jcmorrow:
[...]

Issue #1266, opened 2016-10-19
Yarn install fails, private modules?


fyockm:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
Run `yarn` on existing project with private module dependencies.

**If the current behavior is a bug, please provide the steps to reproduce.**
Arguments: 
  /Users/drew/.nvm/versions/node/v6.6.0/bin/node /Users/drew/.nvm/versions/node/v6.6.0/bin/yarn

PATH: 
  /usr/local/heroku/bin:/Users/drew/.nvm/versions/node/v6.6.0/bin:/Users/drew/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin

Yarn version: 
  0.16.1

Node version: 
  6.6.0

Platform: 
  darwin x64

npm manifest: 
  {
    "name": "voyant-assets",
    "description": "Asset manager",
    "version": "0.5.1",
    "author": "Bessemer Alliance",
    "dependencies": {
      "@voyant/auth": "^1.0.0",
      "@voyant/config": "^1.0.0",
      "@voyant/crypt": "^1.0.0",
      "@voyant/mailer": "^2.0.0",
      "@voyant/models": "^1.2.3",
      "@voyant/session": "^1.1.0",
      "async": "^2.0.0",
      "aws-sdk": "^2.2.15",
      "body-parser": "^1.14.2",
      "boom": "^3.0.0",
      "compression": "^1.6.0",
      "cors": "^2.7.1",
      "express": "^4.13.3",
      "knox": "^0.9.2",
      "lodash": "^4.13.1",
      "mime-types": "^2.1.7",
      "modify-filename": "^1.1.0",
      "moment": "^2.11.2",
      "mongoose": "4.5.5",
      "multer": "^1.1.0",
      "multer-s3": "sabymike/multer-s3",
      "passport": "^0.3.0",
      "request": "^2.67.0",
      "s3-streams": "BessemerAlliance/s3-streams",
      "speechmatics": "^1.0.0",
      "throng": "^4.0.0",
      "type-is": "^1.6.10",
      "validator": "^5.4.0",
      "winston": "^2.1.1"
    },
    "devDependencies": {
      "aws-sdk-mock": "^1.5.0",
      "buffertools": "^2.1.3",
      "eslint": "^2.9.0",
      "eslint-config-airbnb-base": "^3.0.1",
[...]

Issue #1241, opened 2016-10-19
Cannot use 'in' operator to search for '^4.14.34' in undefined


VsLoureiro:
Do you want to request a feature or report a bug?

Bug

What is the current behavior?

Doesn't add packages.

If the current behavior is a bug, please provide the steps to reproduce.

Run $ yarn

What is the expected behavior?

Should install all my dependencies in package.config

Please mention your node.js, yarn and operating system version.
node.js: v4.6.0
npm: 3.10.8
yarn: 0.16.0
Windows: 7

Log:

Trace: 
  TypeError: Cannot use 'in' operator to search for '^4.14.34' in undefined
      at C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\lib-legacy\resolvers\registries\npm-resolver.js:74:24
      at next (native)
      at step (C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)
      at C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:35:14
      at new Promise (C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\node_modules\core-js\library\modules\es6.promise.js:191:7)
      at C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:14:12
      at Function.findVersionInRegistryResponse (C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\lib-legacy\resolvers\registries\npm-resolver.js:84:7)
      at C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\lib-legacy\resolvers\registries\npm-resolver.js:101:34
      at next (native)
      at step (C:\Users\vloureiro\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)

This happens when I try to install dependencies from @types/xxxx all the others work ok
This particular the dependency was @types/lodash

ecozoic:
I'm getting this issue when installing @angular/xxx packages.

```TypeError: Cannot use 'in' operator to search for '2.1.0' in undefined```

wyze:
Solved by #1356.

Issue #1186, opened 2016-10-18
Cannot use 'in' operator to search...


julmot:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Doesn't add packages.

**If the current behavior is a bug, please provide the steps to reproduce.**

Run `$ yarn add bootstrap`

**What is the expected behavior?**

Should add Bootstrap.

**Please mention your node.js, yarn and operating system version.**
node.js: v5.6.0
npm: 3.10.9
yarn: 0.15.1
Windows: 7

**Log:**
```
yarn add v0.15.1
info No lockfile found.
[1/4] Resolving packages...
error TypeError: Cannot use 'in' operator to search for '~1.1.0' in undefined
    at C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\lib\resolvers\registries\npm-resolver.js:62:24
    at next (native)
    at step (C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)
    at C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:35:14
    at new Promise (C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\node_modules\core-js\library\modules\es6.promise.js:191:7)
    at C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:14:12
    at Function.findVersionInRegistryResponse (C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\lib\resolvers\registries\npm-resolver.js:72:7)
    at C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\lib\resolvers\registries\npm-resolver.js:89:34
    at next (native)
    at step (C:\Users\julmot\AppData\Roaming\npm\node_modules\yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)
info Visit http://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

VsLoureiro:
I have the same problem when trying to install @types/xxx dependencies

matrunchyk:
Same issue. Dump:
```
Arguments: 
  /usr/local/bin/node /Users/sergii/node_modules/bin/yarn install

PATH: 
[...]

PR #1281, opened 2016-10-20
Verify `dist-tags` key exists on object before querying range

wyze:
**Summary**

This appears to fix quite a few similar, possibly duplicate issues (#953, #1186, #1241, #1259, #1266, #1274). They all throw a similar error in `npm-resolver.js`: 

```TypeError: Cannot use 'in' operator to search for '<range>' in undefined```

This also seems like it could be a bandaid that would be masking a different problem. I unfortunately wasn't able to reproduce the issue myself. Even went as far as installing Windows 7 on VirtualBox and tried to reproduce.

**Note**

Also, it would be great to get some tests around the resolvers. I don't know much about Jest and snapshot testing and if it would be used for this. Any guidance in getting some tests in place would be awesome as well. Could even be done in a separate PR.

**Test plan**

No longer throw the `TypeError` listed above and have the `install`/`add`/`upgrade` command work as expected.

### DIFF
--- a/src/resolvers/registries/npm-resolver.js
+++ b/src/resolvers/registries/npm-resolver.js
@@ -24,8 +24,9 @@ export default class NpmResolver extends RegistryResolver {
   static registry = 'npm';
 
   static async findVersionInRegistryResponse(config: Config, range: string, body: RegistryResponse): Promise<Manifest> {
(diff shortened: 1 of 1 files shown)


kittens:
Made redundant by #1356.

PR #1284, opened 2016-10-20
add remove --all flag fixes #1048

### LINKED ISSUE #1048: Can yarn remove all modules?
When I input the command 

```shell
yarn remove
```
But it warned me 

> error Not enoguh arguments, expected at least 1

So Does the yarn support the **remove all**?

#### PS

```shell
yarn remove *
```

it doesn't work!

#### Thanks

paolord:
**Summary**

Adds --all flag in yarn remove, only removes packages in dependencies and devDependencies from npm

created a function, checkAllFlag, in remove.js that checks if the flag is set, then populates args with all the packages from npm.

**Test plan**

included new test in __tests__/commands/install.js to assert the return value of the checkAllFlag in remove.js

Notes:
npm run build and npm run test is ok on my machine, lints fine.

### DIFF
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import {run as uninstall} from '../../src/cli/commands/remove.js';
+import {checkAllFlag} from '../../src/cli/commands/remove.js';
 import {run as check} from '../../src/cli/commands/check.js';
 import * as reporters from '../../src/reporters/index.js';
 import {Install} from '../../src/cli/commands/install.js';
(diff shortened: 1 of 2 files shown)


samccone:
This seems like a reasonable feature. @yarnpkg/core thoughts?

bestander:
What is the use case for removing all dependencies?

paolord:
looking at the issues, users expect a behavior like`yarn remove` or  `yarn remove *` to simply remove all dependencies. But it could be dangerous when someone makes an accident, so having an --all flag is a bit better.

bestander:
@paolord, I mean is it a legitimate use case to remove all dependencies?
When would anyone want thet?

paolord:
The best use case really is just expecting yarn to be able to do a remove all, instead of typing each package name after `yarn remove`. And my PR only removes packages that are found in package.json

bestander:
@paolord, yeah I appreciate that consistency in commands is a good thing to strive for.
Good job on that.
But I would rather keep the code more maintainable  by having less code.

In the case where I would want to remove all dependencies I would just `rm -rf node_modules, yarn.lock package.json` and I would almost never want to do that.

Let's close this for now unless there is a need from the community.

JL2210:
@bestander, I need it for a package manager called chromebrew, for writing an uninstall script.

sbarli:
Know this is an old thread, but this would still be a nice to have -- 

@bestander re. your question of a use-case: 

I am deprecating an app from a monorepo where removing `yarn.lock` is not an option as it is hoisted at the root of the monorepo and includes dependencies across workspaces. I want to uninstall all dependencies of just one of the workspaces in the monorepo. Seems to do that I have to manually remove them all by name which is both annoying and prone to errors.

Issue #1237, opened 2016-10-19
yarn config unable to set boolean value


rzontar:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
Any config value set using the cli is treated as a string. Setting this config for strict-ssl (#841) is not honored and an `unable to verify the first certificate` error is raised. Setting the same value strict-ssl false in an `.yarnrc` results in a correct conversion.

**If the current behavior is a bug, please provide the steps to reproduce.**
`yarn config set strict-ssl false`
success Set "strict-ssl" to "false".

`yarn config list`
{ 'strict-ssl': 'false',
... }

**What is the expected behavior?**
`yarn config set strict-ssl false`
success Set "strict-ssl" to false.

`yarn config list`
{ 'strict-ssl': false,
... }

**Please mention your node.js, yarn and operating system version.**
node v6.7.0
yarn v0.16.0
Windows 10

rapidia:
As a makeshift
```bash
$ yarn config set strict-ssl ""
```

kittens:
Created #1590 to fix this, pending approval.

wesm87:
It seems like this has been reintroduced in recent versions (I've had this issue with both 0.17.9 and 0.18.0). I took a quick look at the source code and it looks like the `NpmRegistry` class is using `Registry.normalizeConfig()` but the `YarnRegistry` class isn't:

https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L124
https://github.com/yarnpkg/yarn/blob/master/src/registries/yarn-registry.js#L81

Could this be causing the issue?

thebetterjort:
yarn config set strict-ssl ""
Is still required as of writing. You cannot do yarn config set strict-ssl false.

mckibbenta:
I'm seeing the same behavior as @wesm87 and @thebetterjort in v0.18.1.

```yarn config set strict-ssl false```

results in the following in `.yarnrc`

```strict-ssl "false"```

wesm87:
Looks like there's an open PR for this now: https://github.com/yarnpkg/yarn/pull/2440

PR #1292, opened 2016-10-20
Fix `yarn config set` for boolean values (#1237)

jsynowiec:
**Summary**

All config values were treated as strings but some config keys require booleans, e.g. `strict-ssl` or `ignore-scripts`. See: #1237 

**Test plan**
1. Set `strict-ssl` to `false`
2. Run `yarn config list`
3. Verify if `strict-ssl` is set to boolean `false`

screenshot 2016-10-20 11 39 56

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -16,8 +16,15 @@ export const {run, setFlags} = buildSubCommands('config', {
       return false;
     }
 
-    const [key, val = true] = args;
+    const [key, val] = ((key, val = true) => {
+      if (typeof val === 'string' && (/^(true|false)$/i).test(val)) {
+        val = (val.toLowerCase() === 'true');
+      }
+
+      return [key, val];
+    })(...args);
     const yarnConfig = config.registries.yarn;
+
     await yarnConfig.saveHomeConfig({[key]: val});
     reporter.success(reporter.lang('configSet', key, val));
     return true;


src/cli/commands/config.js
```diff
@@ -16,8 +16,15 @@ export const {run, setFlags} = buildSubCommands('config', {
       return false;
     }
 
-    const [key, val = true] = args;
+    const [key, val] = ((key, val = true) => {
+      if (typeof val === 'string' && (/^(true|false)$/i).test(val)) {
```
kittens: Can you change this to `val === 'false' || val === 'true', it's not super obvious what this does on first glance.

src/cli/commands/config.js
```diff
@@ -16,8 +16,15 @@ export const {run, setFlags} = buildSubCommands('config', {
...
+        val = (val.toLowerCase() === 'true');
+      }
+
+      return [key, val];
+    })(...args);
```
kittens: Not a fan of the IIFE and spreading the arguments into it, we should do something like:

```javascript
const key = args[0];
let val = args[1];

if (key === 'true' ...) {
   val = val === 'true';
}
```

We don't really need to handle the alternate casing of `true`.

src/cli/commands/config.js
jsynowiec: It won't work for upper/mixed-case letters, and simply using`val.toLowerCase() === 'false' || val.toLowerCase() === 'true'` (or introducing another, intermediate variable to hold the lowercase value) will break on the default `true` value. The typeof check is required.

The regex can be replaced, e.g.

```
if (typeof val === 'string' && (val.toLowerCase() === 'false' || val.toLowerCase() === 'true')) {
```

but it's ugly :/

src/cli/commands/config.js
jsynowiec:
```javascript
const key = args[0];
let val = args[1];
```

You'll loose the default `true` value this way from the `val` assignment (I didn't want to introduce a regression). It *could* be replaced with `let val = args[1] || true` but it's not the same as a default value (falsy values don't work, most importantly `0`), hence IIFE.

```javascript
if (key === 'true' ...) {
   val = val === 'true';
}
```

See my previous comment.

> We don't really need to handle the alternate casing of `true`.

What about calling `config set` without the value? AFAIK, NPM defaults to an empty string.

timdp:
This also affects `npm_config_*` environment variables, right? `npm_config_production=false yarn` is being treated as `yarn --prod` for me because `!!'false'` is true.

kittens:
Created #1590 to fix this, pending approval.

PR #1319, opened 2016-10-20
fix --ignore-scripts flag not working

kengoldfarb:
**Summary**

Fixes the ```--ignore-scripts``` flag not working.

**Test plan**

Build and run it with and without the flag.  Observe that postinstall and preinstall scripts are now working properly.


samccone:
We will want to match the ignore-scripts behavior here.

https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/test/tap/ignore-scripts.js#L19-L46

It looks like we are missing a few 😉

kittens:
Created #1422 to fix this, pending approval.

Issue #1029, opened 2016-10-13
global option doesn't allow system-wide installation of bin utils


hyperknot:
**Do you want to request a _feature_ or report a _bug_?**
feature

**What is the current behavior?**
Yarn installs to a user's or root's home folder, and links to that folder from `/usr/bin`.
Since on most Linux only the current user has access to it's home folder, only the user who run the install command can use the globally installed bin utils. For anyone else they get permission denied errors when trying to run the globally installed bin util.

**What is the expected behavior?**
Global install of bin utils should function like they were installed system-wide, like with apt-get or npm.

**Please mention your node.js, yarn and operating system version.**
node: 4.6.0, yarn: 0.15.0, OS: Ubuntu 16.04.1

FLGMwt:
`yarn global` respects the `PREFIX` environment variable which overrides the default `/usr/bin`

You can test the bin resolution with `yarn global bin` before and after setting the variable.

hyperknot:
So the `yarn global bin` reports `/usr/bin` for both root and user. That would be all good, and is the same behaviour for npm.

The problem is where those modules are actually staying on the disk. 

What `npm install --global eslint`, does is:

```
lrwxrwxrwx 1 root root 40 Oct 14 13:33 /usr/bin/eslint -> ../lib/node_modules/eslint/bin/eslint.js
```

While what `yarn global add eslint` does is:

```
lrwxrwxrwx 1 root root 55 Oct 14 13:34 /usr/bin/eslint -> ../../root/.yarn-cache/.global/node_modules/.bin/eslint
```

This breaks installation as no user other then root has access to /root/.yarn-cache.

Recommended solution is to use `/usr/lib/node_modules` or `/usr/bin/yarn_modules` or something for global modules.

lucaskatayama:
Is that correct?

> I believe the recommended solution would be to use /usr/lib/node_modules or /usr/**bin**/yarn_modules or something for global modules.

I think you meant /usr/**lib**/yarn_modules .. Right?

hyperknot:
Sorry, of course, edited my post.

Fed03:
I use brew to install node on my macOS.

NPM bin: `/usr/local/bin`
yarn bin: `/usr/local/Cellar/node/6.7.0/bin`

I don't know if it could be useful

lucaskatayama:
My attempt to fix it

[...]

PR #1344, opened 2016-10-21
Fixing global add in linux root folder #1029

lucaskatayama:
**Summary**
Fixing `yarn global add` being installed in root folder described in [issue #1029](https://github.com/yarnpkg/yarn/issues/1029)
.
Users can't access `/root/.yarn-*` .
Using `/usr/local/lib`instead like npm modules (`/usr/local/lib/node_modules`)
eg:

```
yarn global add react-native-cli
/usr/local/bin/react-native -> ../lib/.yarn-config/global/node_modules/.bin/react-native*
```

**Reference**
- [Issue #1029](https://github.com/yarnpkg/yarn/issues/1029)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -2,10 +2,12 @@
 const path = require('path');
 let userHome = require('user-home');
 if (process.platform === 'linux' && process.env.USER === 'root') {
-  userHome = path.resolve(process.execPath, '..', '..', 'lib');
+  userHome = path.resolve('/usr/local/share');
 }
 
-type Env = {[key: string]: ?string};
+type Env = {
+  [key: string]: ? string
+};
 
 export const DEPENDENCY_TYPES = [
   'devDependencies',
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,7 +1,9 @@
 /* @flow */
-
-const userHome = require('user-home');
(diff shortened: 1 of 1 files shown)


lucaskatayama:
Getting error fetching modules

bestander:
@Daniel15, assigning to you

kentnek:
Please merge this asap, since I'm trying to use yarn to install pm2 globally :)

src/constants.js
```diff
@@ -1,7 +1,9 @@
 /* @flow */
-
-const userHome = require('user-home');
 const path = require('path');
+let userHome = require('user-home');
+if (process.platform === 'linux' && process.env.USER === 'root') {
+  userHome = path.resolve(process.execPath, '..', '..', 'lib');
```
Daniel15: This makes an assumption about where Node.js is installed, which could vary based on the environment (eg. `/usr/bin/node` if the user installed it via a package manager, `/usr/local/bin/node` or `/opt/nodejs/` or really anything else if they compiled it themselves).

If you want to use `/usr/local/lib/` then it's probably best to just hard-code that here.

Daniel15:
Cool, let's try this out and see how it goes!

Issue #1348, opened 2016-10-21
yarn add error


griffinmichl:
```
Trace: 
  TypeError: Cannot use 'in' operator to search for 'latest' in undefined
      at /Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:62:24
      at next (native)
      at step (/Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
      at Promise.F (/Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/node_modules/core-js/library/modules/_export.js:35:28)
      at /Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12
      at Function.findVersionInRegistryResponse (/Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:72:7)
      at /Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:89:34
      at next (native)
      at step (/Users/griffin.michl/.nvm/versions/node/v6.6.0/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
```

Cannot install any new packages via `yarn add`. Always throws this error. No issues install existing dependencies.

kwelch:
Based on [this tweet](https://twitter.com/thejameskyle/status/789566636130906112) connection to registry should be restored.

thejameskyle:
registry.yarnpkg.com is now bypassing DNS to registry.npmjs.org so this should no longer be a problem

kwelch:
There was also a good workaround posted in [another related issue](https://github.com/yarnpkg/yarn/issues/1354#issuecomment-255457481)

@tuures:
> `# .npmrc
registry = https://registry.npmjs.org/`

rivertam:
@thejameskyle what happens if both are down?

thejameskyle:
[...]

PR #1356, opened 2016-10-21
Meaningful error message on malformed registry response

rivertam:
**Summary**

When the yarn DNS failed to resolve today, a malformed response from the registry caused a cryptic error (see #1348) rather than a meaningful message. 

Sorry for no test plan; this is a simple commit, but I wouldn't know how to reproduce it. `npm test` still works, so it doesn't break anything. =)

### DIFF
--- a/__tests__/fixtures/request-cache/GET/localhost/.bin
+++ b/__tests__/fixtures/request-cache/GET/localhost/.bin
@@ -1,5 +1,5 @@
 HTTP/1.1 200 OK
-Date: Wed, 19 Oct 2016 13:30:48 GMT
+Date: Mon, 24 Oct 2016 13:03:20 GMT
 Connection: close
 Content-Length: 2
 
--- a/src/resolvers/registries/npm-resolver.js
+++ b/src/resolvers/registries/npm-resolver.js
@@ -25,9 +25,9 @@ export default class NpmResolver extends RegistryResolver {
 
   static async findVersionInRegistryResponse(config: Config, range: string, body: RegistryResponse): Promise<Manifest> {
     if (!body['dist-tags']) {
-      throw new MessageError(`Received malformed response from registry. The registry may be down.`);
+      throw new MessageError(config.reporter.lang('malformedRegistryResponse'));
     }
-    
+
     if (range in body['dist-tags']) {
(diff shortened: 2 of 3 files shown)


src/resolvers/registries/npm-resolver.js
```diff
@@ -24,6 +24,10 @@ export default class NpmResolver extends RegistryResolver {
   static registry = 'npm';
 
   static async findVersionInRegistryResponse(config: Config, range: string, body: RegistryResponse): Promise<Manifest> {
+    if (!body['dist-tags']) {
+      throw new MessageError(`Received malformed response from registry. The registry may be down.`);
```
thejameskyle: This needs to use `this.reporter.lang`, look for other uses in the codebase to see what I mean

src/resolvers/registries/npm-resolver.js
```diff
@@ -24,6 +24,10 @@ export default class NpmResolver extends RegistryResolver {
   static registry = 'npm';
 
   static async findVersionInRegistryResponse(config: Config, range: string, body: RegistryResponse): Promise<Manifest> {
+    if (!body['dist-tags']) {
```
samccone: this check feels very loose. I wonder if we can do better or if this is the best we can do...

rivertam: I was thinking the same thing, but I believe this is possibly the best indicator of a well-formed response, given what we want. Definitely open to suggestions, but I was specifically trying to avoid an error slightly lower down where `body['dist-tags']` is `undefined`.

__tests__/fixtures/request-cache/GET/localhost/.bin
```diff
@@ -1,5 +1,5 @@
 HTTP/1.1 200 OK
-Date: Wed, 19 Oct 2016 13:30:48 GMT
+Date: Mon, 24 Oct 2016 13:03:20 GMT
```
rivertam: I think this should be in the `.gitignore`, but I'm not sure what the generalized entry would be.

SvenWalter:
Currently I'm facing the same issue. Is there any time frame when it will be part of a official release? Thanks ... and great work!

PR #1359, opened 2016-10-22
Fix typo in comment "occured" -> "occurred"

chrisvfritz:
**Summary**

Fixes a typo so as not to distract others reading the source.

**Test plan**

It's just a comment change, so if the tests still pass, we should be good. 😄


Daniel15:
Thanks!

PR #1365, opened 2016-10-22
improve Git coverage

tudorgergely:
**Summary**

This pull request adds tests for util/git.js. It improves the coverage to ~67% and also fixes a bug where Git.assertUrl was not checking correctly for git:// urls (because it was expecting the protocol to be `git` but it is actually `git:`.

**Test plan**

`npm run test` now outputs ~67% statements covered and 78% functions for `util/git.js`


samccone:
👍 thanks

Issue #1260, opened 2016-10-19
yarn fails to install public scoped packages


Offirmo:
Related but not exactly the same as #825, and most likely related to #991, #1234

**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
Trying to install a free, scoped package
```
yarn add --dev @offirmo/cli-toolbox
yarn add v0.16.0
[1/4] Resolving packages...
error Couldn't find package "@offirmo/cli-toolbox" on the "npm" registry.
```
However, this fine module is public and available on npm https://www.npmjs.com/package/@offirmo/cli-toolbox

**What is the expected behavior?**
It should find the module in the npm registry and install it

**Please mention your node.js, yarn and operating system version.**
* node 6.9.0
* yarn 0.16.0
* Ubuntu 16.04 latest

Offirmo:
Thanks for yarn, this tool is very promising !

jsynowiec:
Are you using any registry proxy or something similar? I've just successfully installed this package.

```
$ yarn add --dev @offirmo/cli-toolbox
yarn add v0.16.0
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
success Saved 233 new dependencies.
├─ @offirmo/cli-toolbox@0.0.4
├─ ansi-256-colors@1.1.0
(...)
```

Offirmo:
@jsynowiec I'm indeed accessing npm through my company's npm proxy.

I'll disable it and recheck.

Offirmo:
@jsynowiec confirmed: removing the npm proxy solves the error.

I'll report it to the proxy maker.

Sorry for the trouble.

PR #1373, opened 2016-10-22
Don't let the authorization header be empty

vbfox:
**Summary**

This PR is aimed at solving problems with `@types` (And I guess other public scopes) that can be seen in #825 and #1260 where even with the latest version yarn fails to restore packages where npm succeed. The case is specific to enterprise caches as the yarn or npm default repositories aren't affected.

In my case I had the problem and tested the resolution on ProGet ( http://inedo.com/proget ) configured with an NPM connector..

The issue is that `always-auth` is assumed for scoped packages, and `always-auth` forces the creation of an `authorization:` http header, even when no authentication can be found. The header is created empty in this case. As ProGet is an ASP.Net application and this platform consider an empty authorization header to be something that can never be met, it refuses the connection.

This PR doesn't change the behavior of `always-auth` being assumed for scoped packages but it remove the empty header if no authentication can be found.

**Test plan**

* I ran `yarn` to restore packages in a previously non-working repository and checked that It worked, creating the lock file and was usable after. Tested `yarn add` with a new packages in `@types`
* I tested the same commands while targeting the default repository instead of the cache to ensure that it worked.

I don't have access to a repository that require authentication for scoped packages but the change is minimal and doesn't affect any case where authentication was sent before so it shouldn't affect them (Also npm doesn't send the header in such cases)

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -58,7 +58,10 @@ export default class NpmRegistry extends Registry {
 
     const headers = {};
     if (this.token || (alwaysAuth && requestUrl.startsWith(registry))) {
-      headers.authorization = this.getAuth(pathname);
+      const authorization = this.getAuth(pathname);
+      if (authorization) {
+        headers.authorization = authorization;
+      }
     }
 
     return this.requestManager.request({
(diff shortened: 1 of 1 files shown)


src/registries/npm-registry.js
```diff
@@ -177,7 +180,7 @@ export default class NpmRegistry extends Registry {
       }
     }
 
-    return '';
+    return undefined;
```
kittens: Make this return an empty string please, we should ensure types are monomorphic if we can. An empty string will still evaluate the branch added above.

PR #1374, opened 2016-10-22
Prevent `yarn install` from getting in an infinite loop

### LINKED ISSUE #1227: Adding a script to package.json called "install" creates an infinite loop

Bug

Add this to your `package.json`:

```json
  "scripts": {
    "install": "yarn install --ignore-optional",
  },
```

Run

```bash
yarn install
```

```
> yarn install
yarn install v0.16.0
success Already up-to-date.
$ yarn install --ignore-optional
yarn install v0.16.0
success Already up-to-date.
$ yarn install --ignore-optional
yarn install v0.16.0
success Already up-to-date.
$ yarn install --ignore-optional
yarn install v0.16.0
success Already up-to-date.
$ yarn install --ignore-optional
Terminate batch job (Y/N)?
yarn install v0.16.0
Terminate batch job (Y/N)?
^C
```

No infinite loop. Perhaps copy what Composer does: emits this warning:

> A script named install would override a Composer command and has been skipped

And then carries on with the built-in command.

Node: v6.3.1
Yarn: 0.16.0
OS: Win10 Anniversary

wyze:
**Summary**

This fixes #1227. It adds a check before executing the (pre/post)install lifecycle methods to make sure it won't get into an infinite loop. Not sure if this is the best way to handle it, so feedback is welcome.

**Test plan**

The lifecycle methods (pre/post)install can no longer get in a `yarn install` infinite loop.

### DIFF
--- a/src/cli/commands/_execute-lifecycle-script.js
+++ b/src/cli/commands/_execute-lifecycle-script.js
@@ -15,9 +15,19 @@ export async function execFromManifest(config: Config, commandName: string, pkg:
   }
 
   const cmd: ?string = pkg.scripts[commandName];
-  if (cmd) {
+  if (cmd && isValidCommand(commandName, cmd)) {
     await execCommand(commandName, config, cmd, cwd);
   }
+}
+
+function isValidCommand(commandName: string, cmd: string): boolean {
+  // prevent infinite loop (see: https://github.com/yarnpkg/yarn/issues/1227)
+  // blocks `yarn`, `yarn i`, `yarn install`, with any flags as well
+  if (commandName.endsWith('install') && /^yarn(\si(nstall)?)?(\s\-.*)?$/.test(cmd)) {
+    return false;
+  }
+
+  return true;
 }
 
 export async function execCommand(stage: string, config: Config, cmd: string, cwd: string): Promise<void> {


src/cli/commands/_execute-lifecycle-script.js
```diff
@@ -15,11 +15,21 @@ export async function execFromManifest(config: Config, commandName: string, pkg:
...
-  if (cmd) {
+  if (cmd && isValidCommand(commandName, cmd)) {
     await execCommand(commandName, config, cmd, cwd);
   }
 }
 
+function isValidCommand(commandName: string, cmd: string): boolean {
+  // prevent infinite loop (see: https://github.com/yarnpkg/yarn/issues/1227)
+  // blocks `yarn`, `yarn i`, `yarn install`, with any flags as well
+  if (commandName.endsWith('install') && /^yarn(\si(nstall)?)?(\s\-.*)?$/.test(cmd)) {
```
rivertam: can just return the opposite of this

```javascript
if (boolean) {
    return false;
}

return true;
```

is a bit of an anti-pattern.

wyze: Yes, I understand that. I was leaving it more open in case there are future exceptions that need to be made. I can always change it now to just return the opposite of the conditional and could revisit this pattern in the future when another condition is needed.

kittens:
See my comment in #1227.

Issue #1387, opened 2016-10-23
Update Errors with hard-coded strings to use reporter.lang


thejameskyle:
**Do you want to request a _feature_ or report a _bug_?** Bug

**What is the current behavior?**

There are at least a few strings not using `this.reporter.lang` which will make them impossible to translate later on. Such as this one: https://github.com/yarnpkg/yarn/blob/master/src/resolvers/registries/npm-resolver.js#L36-L37

**What is the expected behavior?**

Need to update these to use, `this.reporter.lang`, search the code base for that for examples.

This is a good issue for people wanting to contribute for the first time to take on

msaglietto:
Sure, I can give it a try .. I would love to contribute

msaglietto:
I think most of the usage have been converted  .. if there is any other let me know ..

kamilogorek:
This issue has been resolved as well.

PR #1392, opened 2016-10-24
Replace some MessageErrors string for lang  codes

msaglietto:
**Summary**
This is related to the issue https://github.com/yarnpkg/yarn/issues/1387

Replace MessageError string on `package-request.js`, `npm-resolver.js` and `request-manager.js`.

Sometime Im not sure how to get the reporter so I was wondering if its not worth it that Message Error have access to a reporter and you send only the lang message?

**Test plan**
I was able to create some test for `request-manager.js` since there was a test already in place (sry first time using Jest let me know if I should do the tests in another way)

But for the others 2 files I was not able to identify how to reach the code to test it .. so if someone could give some ideas how to test them it would be appreciated

### DIFF
--- a/__tests__/util/request-manager.js
+++ b/__tests__/util/request-manager.js
@@ -34,3 +34,38 @@ test('RequestManager.request with cafile', async () => {
   }
   expect(body).toBe('ok');
 });
+
+test('RequestManager.execute Request 403 error', async () => {
+  const config = await createConfig({});
+  jest.mock('request', (factory) => (options) => {
+    options.callback('', {statusCode: 403}, '');
+    return {
+      on: () => {},
+    };
+  });
+  await config.requestManager.execute({
+    params: {url: `https://localhost:port/?nocache`, headers: {Connection: 'close'}},
+    resolve: (body) => {},
+    reject: (err) => {
+      expect(err.message).toBe('https://localhost:port/?nocache: Request "https://localhost:port/?nocache" returned a 403');
+    },
+  });
+});
+
+test('RequestManager.request with offlineNoRequests', async () => {
+  const config = await createConfig({offline: true});
+  try {
+    await config.requestManager.request({url: `https://localhost:port/?nocache`, headers: {Connection: 'close'}});
+  } catch (err) {
+    expect(err.message).toBe('Can\'t make a request in offline mode');
+  }
+});
+
+test('RequestManager.saveHar no captureHar error message', async () => {
+  const config = await createConfig({captureHar: false});
+  try {
+    config.requestManager.saveHar('testFile');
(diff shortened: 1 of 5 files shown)


__tests__/util/request-manager.js
```diff
@@ -34,3 +34,38 @@ test('RequestManager.request with cafile', async () => {
...
+    options.callback('', {statusCode: 403}, '');
+    return {
+      on: () => {},
+    };
+  });
+  await config.requestManager.execute({
+    params: {url: `https://localhost:port/?nocache`, headers: {Connection: 'close'}},
+    resolve: (body) => {},
+    reject: (err) => {
+      expect(err.message).toBe('https://localhost:port/?nocache: Request "https://localhost:port/?nocache" returned a 403');
```
msaglietto: I was wondering if I can make jest to wait for the execution of this expect (like the done from mocha) to ensure the execution of it?

Also here the url is set it twice because the reject is going to set the message like `err.message = `${params.url}: ${err.message}`;` so I dont know if you guys think the url should be removed from the original message

__tests__/util/request-manager.js
```diff
@@ -34,3 +34,38 @@ test('RequestManager.request with cafile', async () => {
...
+    expect(err.message).toBe('Can\'t make a request in offline mode');
+  }
+});
+
+test('RequestManager.saveHar no captureHar error message', async () => {
+  const config = await createConfig({captureHar: false});
+  try {
+    config.requestManager.saveHar('testFile');
+  } catch (err) {
+    expect(err.message).toBe('RequestManager was not setup to capture HAR files');
```
msaglietto: should I mock the reporter.lang somehow and test for the key instead to  relay that  the en language is set ?

__tests__/util/request-manager.js
rivertam: It might be possible to wrap the `execute` call in a `Promise`, then call `reject` on the err, then `await` the `Promise` and `catch` the error and then follow it up with the `expect`. I'm not positive that would work, but I don't believe the `execute` returns a `Promise` so calling `await` on it shouldn't have the desired effect.

src/util/request-manager.js
```diff
@@ -283,6 +282,7 @@ export default class RequestManager {
 
   execute(opts: RequestOptions) {
     const {params} = opts;
+    const {reporter} = this;
```
msaglietto: Because params.callback, that is where the reporter is used, dont have access to this .. I also can change params.callback to an arrow function

kittens:
Looks good! Can you please rebase?

msaglietto:
@kittens Sorry for the delay .. I just rebase it. ..

PR #1372, opened 2016-10-22
honor github commit hash

shreeve:
**Summary**

This small commit honors the commit hash for packages hosted on github. This allows references in `package.json` to use a commit hash like this:

```json
    "cool-util": "niceguy/cool-util#5ea08cb76619f78e17b988ee3acfeba10ce12160"
```

**Test plan**

Verified that this small addition pulls in the correct version from github.

PR #1393, opened 2016-10-24
honor github commit hash re pr/1372

samccone:
building on #1372 .. add added some tests :)

### DIFF
--- a/src/resolvers/exotics/github-resolver.js
+++ b/src/resolvers/exotics/github-resolver.js
@@ -22,15 +22,16 @@ export default class GitHubResolver extends HostedGitResolver {
   }
 
   static getTarballUrl(parts: ExplodedFragment, hash: string): string {
-    return `https://codeload.github.com/${parts.user}/${parts.repo}/tar.gz/${hash}`;
+    return `https://codeload.${this.hostname}/${parts.user}/${parts.repo}/tar.gz/${hash}`;
   }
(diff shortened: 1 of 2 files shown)


src/resolvers/exotics/github-resolver.js
```diff
@@ -22,15 +22,16 @@ export default class GitHubResolver extends HostedGitResolver {
...
 
   static getGitSSHUrl(parts: ExplodedFragment): string {
-    return `git@github.com:${parts.user}/${parts.repo}.git`;
+    return `git@${this.hostname}:${parts.user}/${parts.repo}.git` +
+      `${parts.hash ? '#' + decodeURIComponent(parts.hash) : ''}`;
   }
 
   static getGitHTTPUrl(parts: ExplodedFragment): string {
-    return `https://github.com/${parts.user}/${parts.repo}.git`;
+    return `https://${this.hostname}/${parts.user}/${parts.repo}.git`;
```
Daniel15: Is this to allow usage of Github Enterprise?

samccone: seems like a slightly diff problem if not. :)

backflip:
@samccone, @Daniel15, is there a reason this was only applied to `getGitSSHUrl ` an not to `getGitHTTPUrl`, too?

PR #1403, opened 2016-10-24
set filename parameter to mandatory in pack -f

t-obi:
**Summary**
changes the filename parameter from optional to mandatory for yarn pack -f
without this change the pack -f script fails if no filename is provided

**Test plan**
before this change:
```
[tkloht@Tobias-MacBook-Air ~/Code/yarntest2:master*] ../yarn-forked/bin/yarn pack -f
yarn pack v0.16.2
error An unexpected error occurred, please open a bug report with the information provided in "/Users/tkloht/Code/yarntest2/yarn-error.log".
```
in yarn error.log:
```
Trace:
  TypeError: path must be a string or Buffer
      at TypeError (native)
      at Object.fs.open (fs.js:625:11)
      at WriteStream.open (fs.js:1874:6)
      at new WriteStream (fs.js:1860:10)
      at Object.fs.createWriteStream (fs.js:1815:10)
      at /Users/tkloht/Code/yarn-forked/lib/cli/commands/pack.js:133:23
      at Object.<anonymous> (/Users/tkloht/Code/yarn-forked/lib/cli/commands/pack.js:132:11)
      at next (native)
      at step (/Users/tkloht/Code/yarn-forked/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /Users/tkloht/Code/yarn-forked/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:20
```

after this change:
after:
```
[tkloht@Tobias-MacBook-Air ~/Code/yarntest2:master*] ../yarn-forked/bin/yarn pack -f
  error: option `-f, --filename <filename>' argument missing
```

### DIFF
--- a/src/cli/commands/pack.js
+++ b/src/cli/commands/pack.js
@@ -159,7 +159,7 @@ export async function pack(config: Config, dir: string): Promise<stream$Duplex> 
 }
 
 export function setFlags(commander: Object) {
-  commander.option('-f, --filename [filename]', 'filename');
+  commander.option('-f, --filename <filename>', 'filename');
 }
 
 export async function run(


kittens:
It's not mandatory though, we automatically build a filename (the same one npm uses) [here](https://github.com/yarnpkg/yarn/pull/1403/files#diff-155fb1838a019915c6ddab2119283ebaR179).

t-obi:
Yes, I saw that, but that code will not be evaluated if you use "yarn pack -f" instead of "yarn pack -f something".  Commander parses the option to true if you just set -f without providing a filename.
This produces an error at some place: `TypeError: path must be a string or Buffer` but it is a boolean.

Just to illustrate this you can log flags.filename:
```
console.log('>>> filename: ', flags.filename, !!flags.filename);
  const filename = flags.filename || path.join(config.cwd, `${pkg.name}-v${pkg.version}.tgz`);
```

run "yarn pack -f something" produces
`>>> filename:  something true`

run "yarn pack -f" produces
`>>> filename:  true true`

t-obi:
Also just one minor thing: You mentioned that you build the same filename that npm uses.
There is a tiny difference though: npm produces something-1.0.0.tgz for example, yarn produces something-v1.0.0.tgz (yarn has "v" before version number, npm does not).
I found this rather useful during debugging for another PR because it was easy to compare between npm and yarn, just wanted to let you know.

kittens:
Ah okay I understand now, thank you for the clarification!

Issue #1061, opened 2016-10-14
yarn uses UNIX line endings on all systems (also Windows)


ahelmberger:
**Do you want to request a _feature_ or report a _bug_?**
Depends on the point of view

**What is the current behavior?**
yarn uses UNIX line endings on all systems

**If the current behavior is a bug, please provide the steps to reproduce.**
When generating lockfile

**What is the expected behavior?**
It should probably use the system EOL (e.g. '\r\n' on Windows), or at least check what's currently used in an existing lockfile otherwise it annoys users by creating changes in version control systems due to line endings

**Please mention your node.js, yarn and operating system version.**
Windows 7 64bit
node v6.6.0
yarn 0.15.1

Daniel15:
Good catch - This should use [os.EOL](https://nodejs.org/api/os.html#os_os_eol) instead of making assumptions.

ljharb:
Determining the existing endings to minimize diffs is good, but a default to LF otherwise imo is a better idea. That way projects that are used cross-platform can also minimize diffs.

ahelmberger:
IMHO, having LF as the default line ending also for Windows is not the best option. It's just not what Windows users expect and not how most other tools behave. In my experience, working in cross-platform teams for a while now and trying out different conventions, it feels best for everyone (and leads to zero diffs) if tools respect the platforms they run on, and Git does the job ob CRLF->LF conversions (as that is basically the de-facto-standard setting for Git on Windows).

Daniel15:
> a default to LF otherwise imo is a better idea

Disagree - In a primarily Windows team, you'd probably want Windows line endings everywhere. Maybe it could be configurable. As @ahelmberger mentioned, Git converts CRLF -> LF on commit (and LF -> CRLF on checkout) automatically anyways.

ljharb:
If that's true, then since yarn.lock is committed, wouldn't everyone be looking at LF-only diffs anyways? (Or CRLF,  as the case may be)

ljharb:
In my past experience all my Windows tools handled LF just fine -  is this no longer the case with popular tools?

ahelmberger:
@ljharb It's not that Windows tools couldn't handle LFs, it's just annoying to have a dirty working directory after a clean checkout and `yarn install`.

bterlson:
[...]

PR #1405, opened 2016-10-24
Infer EOLs from existing file or use OS EOL

iredchuk:
**Summary**

Yarn uses UNIX line endings on all systems when writing lock file. We want to solve this problem as described in  #1061: infer line endings from the existing file, and if it does not exist, use the system default.

**Test plan**

On a UNIX system, in an empty directory:
$ yarn add <some_package>
$ cat -e yarn.lock
You should only see LF line endings.
$ unix2dos yarn.lock
$ yarn add <some_other_package>
$ cat -e yarn.lock
You should only see CRLF line endings.

On a Windows system, in an empty directory:
$ yarn add <some_package>
$ cat.exe -e yarn.lock
You should only see CRLF line endings.
$ dos2unix.exe yarn.lock
$ yarn add <some_other_package>
$ cat -e yarn.lock
You should only see LF line endings.


iredchuk:
@bestander Will try, thanks.

iredchuk:
@bestander Added, please check.

src/util/fs.js
```diff
@@ -454,3 +455,33 @@ export async function getFileSizeOnDisk(loc: string): Promise<number> {
 export function normalizeOS(body: string): string {
   return body.replace(/\r\n/g, '\n');
 }
+
+const cr = new Buffer('\r', 'utf8')[0];
+const lf = new Buffer('\n', 'utf8')[0];
+
+async function getEolFromFile(path: string) : Promise<string | void>  {
```
bestander: The unfortunate bit is that it adds a read file to every write file operation.
That could unexpectedly slow down quite a few things.
Any ideas how to apply this only to yarn.lock/package.json writing?

iredchuk: IMHO putting the line ending setting into .yarnrc would be not so nice when team members work on different OSs, which is quite probable for a relatively big project.

I tried to [measure the overhead](https://gist.github.com/iredchuk/1ed83441c6985a4cbd2007b07beb273f) and got about 20ms for 10000 read iterations, so for one read it should not be significant.

However, we could limit this behaviour to only files included into source control. Is it only yarn.lock and package.json? If so, maybe we could expose both write functions and use them selectively?

bestander: I would use this explicitly for yarn.lock, package.json and .yarnrc files only then.
How about fs would expose fs.writeFilePreservingEol and we would use it sparsely?
This kind of file parsing does not sit well with me.

iredchuk: Makes sense. Updated.

src/util/fs.js
```diff
@@ -454,3 +455,33 @@ export async function getFileSizeOnDisk(loc: string): Promise<number> {
...
+  let buffer;
+  try {
+    buffer = await readFileBuffer(path);
```
bestander: better return undefined if files does not exist (check if it exists first).
And don't catch error in readFile

iredchuk: I thought about it, but if the file is inaccessible for some other reason - is it OK to just let it fail here? Will this be handled on a higher level?

bestander:
Commits still look quite broken in the PR.
They repeat recently merged ones

iredchuk:
@bestander Sorry I messed up with rebasing :( Created a new PR: https://github.com/yarnpkg/yarn/pull/1911

PR #1417, opened 2016-10-25
fix publish for non default registries

leipert:
**Summary**

The [default registry](https://registry.npmjs.org/) returns a response body

``` json
{"success": true}
```

while other registries might not. Artifactory for example returns

```
HTTP/1.1 201 Created
Date: Tue, 25 Oct 2016 03:41:04 GMT
Connection: keep-alive
Transfer-Encoding: chunked
Content-Type: application/json

{"ok": "created new entity"}

```

This pull requests fixes publishing to non default registries.

**Test plan**

The results of the fixed version can be seen here:

https://asciinema.org/a/auvjqh4hln1ykmiuvmzy8d8ke

### DIFF
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -100,7 +100,7 @@ async function publish(
     body: root,
   });
 
-  if (res != null && res.success) {
+  if (res !== null && res !== false) {
     await executeLifecycleScript(config, 'publish');
     await executeLifecycleScript(config, 'postpublish');
   } else {


zeke:
I would suggest making the code a bit more explicit and adding comments to explain the motivation behind the check:

```js
// default npm registry returns {"success": true}
// artifactory registry returns {"ok": "created new entity"}
if (res && (res.success || res.ok))
```

leipert:
@zeke I do not know whether this is really necessary, as artifactory and npm already return a 2xx status code.

The check against `false` was just added to protect against these lines:
[src/util/request-manager.js#L341-L344](https://github.com/yarnpkg/yarn/blob/0c5bd61b5b50c376ffe37a3bb951d2b3bbe19ed1/src/util/request-manager.js#L341-L344)

In my opinion a request which returns 2xx should be enough to regard a publish request a successful.

zeke:
I see. In that case, why not just `if (res)`?

Either that, or use `!==` instead of `!=` to avoid some type coercion pitfalls:

```js
> '' != false
false

> '' != null
true

> ({} != false)
true

> ([] != false)
false
```

src/cli/commands/publish.js
```diff
@@ -100,7 +100,7 @@ async function publish(
     body: root,
   });
 
-  if (res != null && res.success) {
+  if (res !== null && res !== false) {
```
bestander: This requires a comment as you've described in the PR regarding what different registries may return.
Can you check for response code instead?

bestander:
I'll close this as the author did not get back to this.
Feel free to reopen

leipert:
Sorry for not responding...

At the moment I do not have the time to work on this. Anyone else, feel free to pickup where I left of.

We are still using `npm publish` for publishing. Will try `yarn publish ` again in the future.

mrichins:
Resurrecting this if it hasnt been resolved yet.

We recently ran into a problem with this as well, and I was hoping to finish this off by adding a check on the 200 status of the response variable. 

`if (res !== null && has2xxResponse(res)`

I notice the same check is used elsewhere in the codebase and was hoping to add this new method to some utils, but the method logic can also be inlined pretty easily too.

bestander:
Go for it, send a PR, ideally with tests please

fabricemathis:
Hey,

I was hoping the PR would fix this problem, unfortunately even with Yarn version **0.25.1**, the problem still remains :

`C:\ws\yarn>yarn publish --verbose
  yarn publish v0.25.1
  verbose 0.257 Checking for configuration file "C:\\ws\\yarn\\.npmrc".
  verbose 0.258 Checking for configuration file "C:\\Users\\f.mathis\\.npmrc".
  verbose 0.258 Found configuration file "C:\\Users\\f.mathis\\.npmrc".
  verbose 0.26 Checking for configuration file "C:\\Program Files\\nodejs\\.npmrc".
  verbose 0.261 Checking for configuration file "C:\\ws\\yarn\\.npmrc".
  verbose 0.261 Checking for configuration file "C:\\ws\\.npmrc".
  verbose 0.264 Checking for configuration file "C:\\ws\\yarn\\.yarnrc".
  verbose 0.264 Checking for configuration file "C:\\Users\\f.mathis\\.yarnrc".
  verbose 0.265 Found configuration file "C:\\Users\\f.mathis\\.yarnrc".
  verbose 0.268 Checking for configuration file "C:\\Program Files\\nodejs\\.yarnrc".
  verbose 0.268 Checking for configuration file "C:\\ws\\yarn\\.yarnrc".
  verbose 0.269 Checking for configuration file "C:\\ws\\.yarnrc".
  verbose 0.272 current time: 2017-05-16T07:00:18.447Z
  [1/4] Bumping version...
  info Current version: 0.0.434
  question New version: 0.0.435
  info New version: 0.0.435
  [2/4] Logging in...
  [3/4] Publishing...
  verbose 5.711 Performing "PUT" request to "http://artifactory-repository/artifactory/api/npm/npm-virtual/yarn-artifactory-poc".
  verbose 5.954 Request "http://artifactory-repository/artifactory/api/npm/npm-virtual/yarn-artifactory-poc" finished with status code 201.
  verbose 5.965 Error: Couldn't publish package.
      at MessageError (C:\Program Files (x86)\Yarn\lib\yarn-cli.js:138:5)
      at C:\Program Files (x86)\Yarn\lib\yarn-cli.js:75398:13
      at next (native)
      at step (C:\Program Files (x86)\Yarn\lib\yarn-cli.js:91:30)
      at C:\Program Files (x86)\Yarn\lib\yarn-cli.js:102:13
      at process._tickCallback (internal/process/next_tick.js:103:7)
  error Couldn't publish package.
  info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.`

Looks like this 201 status code is still perceived as a failure.

Any idea ?

mrichins:
Im going to look at it and spend some time debugging. I ran into this issue recently: https://github.com/yarnpkg/yarn/issues/1694 where my publish is still hanging. I can still publish things with NPM, but I cant debug yarn until it stops hanging at the publish step. 

Ill update with what I figure out

PR #1419, opened 2016-10-25
Add configuration options for the yarn cache path.

ryanmt:
**Summary**

The generic cache path doesn't work for all scenarios.  For me, it was a little bit of a blocker for docker containers.  Allowing a configurable path lets me set the cache path to be stored on specific volume which can then be shared between hosts, or can specify a more permanent storage path in an otherwise ephemeral container. 

**Test plan**

**_ENV VAR**_

```
export YARN_CACHE_PATH=~/test-cache

yarn config list #=> contains the cache-path of ~/test-cache

yarn # install modules and see that files are stored in the specified cache
```

**_ENV VAR can be removed to fallback on default behaviors**_

```
unset YARN_CACHE_PATH

rm -rf node_modules/

yarn # install modules to the default cache location
```

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -36,7 +36,7 @@ function getDirectory(type: string): string {
 }
 
 export const GLOBAL_INSTALL_DIRECTORY = path.join(userHome, '.yarn');
-export const MODULE_CACHE_DIRECTORY = getDirectory('cache');
+export const MODULE_CACHE_DIRECTORY = process.env.YARN_CACHE_PATH || getDirectory('cache');
 export const LINK_REGISTRY_DIRECTORY = getDirectory('config/link');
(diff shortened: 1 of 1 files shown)


kittens:
Have you tested to see if this works? It shouldn't since the cache path is setup in `src/config.js` and the npm config isn't read at all for it. Why are you under the impression this will work?

ryanmt:
I thought I'd tested it sufficiently.  I'll double check the ENV VAR solution, but in the light of day I wouldn't expect the config settings to do anything right now.

ryanmt:
I've pulled back to a simple ENV var option to just close the single issue.

bestander:
You can use `--cache-folder` option to specify where to store caches when doing `yarn install`

PR #1424, opened 2016-10-25
Generate helpful error messages whenever we fail to parse JSON

kittens:


src/resolvers/exotics/hosted-git-resolver.js
```diff
@@ -115,20 +115,22 @@ export default class HostedGitResolver extends ExoticResolver {
...
+        url: href,
         queue: this.resolver.fetchingQueue,
       });
       if (!file) {
         return null;
       }
 
       const tarballUrl = this.constructor.getTarballUrl(this.exploded, commit);
-      const json = JSON.parse(file);
+      const json = await config.readJson(href, JSON.parse(file));
```
skevy: Should be `const json = await config.readJson(href, () => JSON.parse(file));`

@kittens

Issue #1138, opened 2016-10-17
Add yarn version check on startup


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Yarn does not notify the user that a new version of Yarn is available

**What is the expected behavior?**
Yarn should periodically check for new version on startup. Maybe once per day. You can get the latest version number by hitting `https://yarnpkg.com/latest-version`.

If the latest version is newer than the current version, display a message with the new version number, and instructions on how to update.  Optionally, if the user is using a TTY (ie. they're not piping the output elsewhere or running Yarn from a script), we could prompt if they want to upgrade right away:
```
Yarn 0.16.0 is available! (you're using 0.15.0)
Do you want to upgrade now (Y/N)? __
```

This would use `yarn update-self` once it's updated to use the same installation method originally used (#1139).

For an initial version, we could just show instructions on how to update. **Important:** These instructions should correspond to the installation method the user used to install Yarn.

For example, on Debian or Ubuntu:
```
Yarn 0.16.0 is available! (you're using 0.15.0)
To upgrade, run "sudo apt-get install yarn"
```

On Windows when using the Windows installer:
```
Yarn 0.16.0 is available! (you're using 0.15.0)
To upgrade, download the latest installer from https://yarnpkg.com/latest.msi
```

On Mac OS:
```
Yarn 0.16.0 is available! (you're using 0.15.0)
To upgrade, run "brew upgrade yarn"
```

When installed via npm:
```
Yarn 0.16.0 is available! (you're using 0.15.0)
To upgrade, run "npm upgrade --global yarn"
```

Currently we don't actually record the installation method anywhere, so I think we'll just have to guess it based on the environment (eg. if `/usr/share/lintian/overrides/yarn` exists, the Ubuntu/Debian package was used) unless we update all the installation methods to write a config file specifying the installation method. 

If we can't tell which installation method was used, just print the new version number without the upgrade instructions.

dlong500:
I love this, but I'd prefer not to have a prompt to upgrade (or at least a config flag to turn that behavior off).  The `yarn self-update` command working as you just mentioned in #1139 would make things very easy to keep things up to date especially on a vanilla Windows platform without any OS package manager.

Thanks for the great work!

thejameskyle:
Big fan of this idea

addyosmani:
[...]

Issue #942, opened 2016-10-12
Output should mention installation method


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
When running `yarn` commands, it includes the version number:
```
$ yarn install
yarn install v0.15.0
info No lockfile found.
success Nothing to install.
```

This is very useful, particularly with bug reports (as anyone that includes the full output with the bug report will automatically include the version number. 

**What is the expected behavior?**
It would be useful to also include the installation method in that output. For example, when installed from the Debian/Ubuntu repo:

```
$ yarn install
yarn install v0.15.0-deb
info No lockfile found.
...
```

Similarly, when installed via Windows installer:
```
yarn install v0.15.0-msi
```

Or from tarball:
```
yarn install v0.15.0-tarball
```

This would require the packaging scripts to make a modification to some file to signify the installation method, and something at runtime to read that file.

FLGMwt:
Should *every* run of a program optimize for but reports, or would maybe this enchancement *and* the version declaration as it exists be reserved for a `yarn --version`?

Daniel15:
I figured that it doesn't add much overhead so it's fine always displaying it wherever we show the version, but only displaying the info on `yarn --version` is a good idea.

olingern:
@Daniel15 Just to continue with your examples. There are a number of ways to get Yarn, so are you thinking something like the following?

tarball
```yarn install v0.15.0-tarball```

brew
```yarn install v0.15.0-brew```

msi
```yarn install v0.15.0-msi```

Chocolatey
```yarn install v0.15.0-choco```

CentOS / Fedora / RHEL
```yarn install v0.15.0-rpm```

Solus
```yarn install v0.15.0-eopkg```

Arch Linux
```yarn install v0.15.0-aur```

Shell Script
```yarn install v0.15.0-sh```

npm
```yarn install v0.15.0-npm```

Daniel15:
@olingern - Tarball and shell script are essentially the same (the shell script just downloads and extracts the tarball). Right now, the Solus and Arch packages are not official, I'm not sure how those packages are being built (probably using the tarball) so I don't know how we'd record the installation method for those. Maybe I'll look into how those packages are built, and add them to our official build scripts.

PR #1429, opened 2016-10-25
Add automatic check for updates that nags the user when there's a new version available

kittens:


Daniel15:
Awesome! I'll take a look at this later today or tomorrow :)

src/cli/commands/install.js
```diff
@@ -640,6 +687,71 @@ export class Install {
...
+    }
+
+    this._checkUpdate().catch(() => {
+      // swallow errors
+    });
+  }
+
+  async _checkUpdate(): Promise<void> {
+    let latestVersion = await this.config.requestManager.request({
+      url: 'https://yarnpkg.com/latest-version',
```
thejameskyle: We need to better automate this getting updated

Daniel15: Yeah, I'm going to have a webhook that runs when releases are updated, or just a cronjob. The tricky thing is that we can't bump the version number until we verify that all files are attached to the release, and the Windows installer is attached separately from everything else (as it's built on AppVeyor).

src/cli/commands/install.js
```diff
@@ -640,6 +687,71 @@ export class Install {
...
     return request;
   }
+
+  /**
+   * Check for updates every day and output a nag message if there's a newer version.
+   */
+
+  checkUpdate() {
+    if (!process.stdout.isTTY || isCI) {
+      // don't show upgrade dialog on CI or non-TTY terminals
```
Daniel15: I think this is still valuable for non-TTY environments. Why should we avoid showing a notice about Yarn being outdated just because the user is piping the output to a file, for example?

src/reporters/lang/en.js
```diff
@@ -72,6 +72,10 @@ const messages = {
   unexpectedError: 'An unexpected error occurred, please open a bug report with the information provided in $0.',
   jsonError: 'Error parsing JSON at $0, $1.',
 
+  yarnOutdated: "Your current version of Yarn is out of date. The latest version is $0 while you're on $1.",
+  yarnOutdatedInstaller: 'To upgrade, download the latest installer at $0.',
+  yarnOutdatedCommand: 'To upgrade, run $0.',
```
Daniel15: Let's put the command on its own separate line. @sindresorhus said:
> Having the command on a separate line makes it easier to copy-paste for users.
https://github.com/yarnpkg/yarn/issues/1138#issuecomment-254114619

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
   tilde: boolean,
 };
 
+/**
+ * Try and detect the installation method for Yarn and provide a command to update it with.
+ */
+
+function getUpdateCommand(): ?string {
+  // Tarball install
```
Daniel15: Instead of doing this inferrence, let's save a token somewhere as part of the packaging process. @cpojer said: 
> Can we somehow add a token to every build indicating the install method? I think that's way better than any sort of inference we might do.

My idea for that was to add an extra `installationMethod` field to `package.json` when building the packages. For example, something like this would be bundled in the Debian package:
```
{
  "name": "yarn",
  "version": "0.16.2",
  "installationMethod": "deb",
```

What do you think? #942 is the issue that covers(-ish) that. If you agree that that's a good idea, let's complete #942 before this one, and then rebase this on top of that.

The main issue I see with inference is when the user has multiple installations of Yarn. That's a bit silly, but they might have installed via the tarball and then also done `npm install -g yarn` for some reason. The instructions might tell them to update the wrong one.

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
...
+  if (fs2.existsSync(path.join(userHome, '.yarn'))) {
```
Daniel15: Isn't `~/.yarn` where user settings are stored?

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
...
+  if (__dirname.indexOf('node_modules') >= 0) {
+    return 'npm upgrade --global yarn';
+  }
+
+  return null;
+}
+
+function getUpdateInstaller(): ?string {
+  // Windows
+  if (fs2.existsSync('C:/Program Files/Yarn') || fs2.existsSync('C:/Program Files (x86)/Yarn')) {
```
Daniel15: This won't work as the user can select any directory during installation. Using the environment variable (`%ProgramFiles%\Yarn`) is better, but still not 100% accurate. Modifying `package.json` to have `"installationMethod": "msi"` as part of the build process would work best.

src/cli/commands/install.js
```diff
@@ -640,6 +687,71 @@ export class Install {
...
+      this.maybeOutputUpdate = () => {
+        this.reporter.warn(this.reporter.lang('yarnOutdated', latestVersion, YARN_VERSION));
+
+        const command = getUpdateCommand();
+        if (command) {
+          this.reporter.info(this.reporter.lang('yarnOutdatedCommand', command));
+        } else {
+          const installer = getUpdateInstaller();
+          if (installer) {
+            this.reporter.info(this.reporter.lang('yarnOutdatedInstaller', installer));
```
Daniel15: Optional, but can we use something like [boxen](https://github.com/sindresorhus/boxen) (or just render the ANSI codes ourself) to render a fancy box around it, like `update-notifier` does?

Daniel15:
See #1557. You should be able to build on that and just have a map of installationMethod => instructions 😄

scripts/install-latest.sh
```diff
@@ -106,9 +106,20 @@ yarn_install() {
   printf "${white}Installing Yarn!$reset\n"
 
   if [ -d "$HOME/.yarn" ]; then
-    printf "$red> ~/.yarn already exists, possibly from a past Yarn install.$reset\n"
-    printf "$red> Remove it (rm -rf ~/.yarn) and run this script again.$reset\n"
-    exit 0
+    if [ -n `which yarn` ]; then
+      LATEST_VERSION=`curl https://yarnpkg.com/latest-version`
```
Daniel15: This may need to handle the `--nightly` flag. We have this in `yarn_get_tarball`:

```
  if [ "$1" = '--nightly' ]; then
    url=https://nightly.yarnpkg.com/latest.tar.gz
  else
    url=https://yarnpkg.com/latest.tar.gz
  fi
```

You can use https://nightly.yarnpkg.com/latest-tar-version to get the latest nightly version.

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
...
+/**
+ * Try and detect the installation method for Yarn and provide a command to update it with.
+ */
+
+function getUpdateCommand(): ?string {
+  if (YARN_INSTALL_METHOD === 'tar') {
+    return 'curl -o- -L https://yarnpkg.com/install.sh | bash';
+  }
+
+  if (YARN_INSTALL_METHOD === 'homebrwe') {
```
Daniel15: Typo: "homebrwe" -> "homebrew"

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
...
+    return 'brew upgrade yarn';
+  }
+
+  if (YARN_INSTALL_METHOD === 'deb') {
+    return 'sudo apt-get updat e&& sudo apt-get install yarn';
```
Daniel15: Typo: "updat e" -> "update"

scripts/install-latest.sh
```diff
@@ -106,9 +106,20 @@ yarn_install() {
...
+      YARN_VERSION=`yarn -V`
+
+      if [ "$LATEST_VERSION" -eq "$YARN_VERSION" ]; then
+        printf "$green> Yarn is already at the latest version.$reset\n"
+      else
+        rm -rf "$HOME/.yarn"
```
Daniel15: 2scary4me, maybe warn the user before doing this.

"A newer version of Yarn is available, do you want to delete the existing version and upgrade? (Y/N)"

src/cli/commands/install.js
```diff
@@ -69,6 +76,43 @@ type Flags = {
...
+  }
+
+  if (YARN_INSTALL_METHOD === 'rpm') {
+    return 'sudo yum install yarn';
+  }
+
+  if (YARN_INSTALL_METHOD === 'npm') {
+    return 'npm upgrade --global yarn';
+  }
+
```
Daniel15: Also add Chocolatey: `choco upgrade yarn`

scripts/install-latest.sh
Daniel15: Also won't this blow away their config? Or is that stored elsewhere?

src/cli/commands/install.js
```diff
@@ -640,6 +687,72 @@ export class Install {
...
+      return;
+    }
+
+    // only check for updates once a day
+    const lastUpdateCheck = Number(this.config.getOption('lastUpdateCheck')) || 0;
+    if (lastUpdateCheck && Date.now() - lastUpdateCheck < ONE_DAY) {
+      return;
+    }
+
+    // don't bug for updates on tagged releases
```
Daniel15: Hmm, I wonder if we should bug on nightly builds?

Daniel15:
Thanks @kittens! This looks pretty good to me. Just some small comments inline 😄

src/cli/commands/install.js
```diff
@@ -25,8 +25,15 @@ import * as crypto from '../../util/crypto.js';
...
 
 const invariant = require('invariant');
+const userHome = require('user-home');
+const semver = require('semver');
 const emoji = require('node-emoji');
+const isCI = require('is-ci');
 const path = require('path');
+const fs2 = require('fs');
+
+const {verison: YARN_VERSION, installationMethod: YARN_INSTALL_METHOD} = require('../../../package.json');
```
rpunkfu: possibly typo `verison`

bestander:
ping @kittens

PR #1441, opened 2016-10-25
Remove bower support

kittens:
Bower support isn't something we document or support very well. We should just get rid of it since it has a lot of issues. Bower is on it's way out and we shouldn't be supporting it. If you want to continue to use Bower then just use it's CLI. We initially had support for it to support Polymer but since they're migrating entirely to npm there's not much use.

We should encourage users to migrate to npm too and bring over the features they like from Bower to Yarn (such as flat install which we already have and other unimplemented features like svn support etc).


FredKSchott:
fwiw, our (Polymer's) immediate plans are to add support for npm (via yarn) *as well as bower*.

But because we don't need any special install-time logic anymore, we are fine using bower for bower-packages and yarn for npm-packages. In fact, if we are publishing to both registries (and keeping seperate bower.json & package.json files in their repos) then we will probably be the most likely to run into all of these yarn+bower bugs 💩 

So +1 from me, I think focusing entirely on the npm use-case makes the most sense for yarn's future.

sheerun:
...

samccone:
@sheerun If you are still interested in working on adding bower support this can be made possible again.
The current decision was based on the lack of resources to work on this and the lackluster experience that yarn provided with bower.

There are some other constraints as well, but I think we can figure out a way to make this work!

csvan:
I honestly think this was a great thing to do, not only for Yarn but for the community at large. Bower is not only dying, *it cannot die soon enough*. This would resolve a plethora of issues and save a ton of resources around double package maintenance, inconsistencies etc.

samccone:
The solution here that @kittens and I talked about is to add the missing features that prevent a user from being able to `mv bower.json package.json` and running `yarn install`

Please open issues around these kind of problems when you run into them. 

I am going to lock this thread now. 

Thanks everyone.

Issue #1442, opened 2016-10-25
pack doesn't work with files in directories


t-obi:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
While working on a PR for #755 I found a weird behaviour with yarn pack.
Basically whenever the package has a nonempty file in a directory the produced archive will be corrupt.
It still seems unlikely that this is really bug and not some error on my part because this should happen with pretty much every package , but I have tried all kinds of stuff and can't see what I might be doing wrong. 

**If the current behavior is a bug, please provide the steps to reproduce.**
I have actually found this by writing some tests for this and will add a PR with the failing tests.

To reproduce this manually:
Create a minimal example, for example just index.js. Add some code to index.js because this will only work if  the file is not empty.
Then run yarn pack and extract the archive - everything should be fine so far.

Move index.js in a directory.
Run yarn pack and extract again - this should fail because the archive is corrupted.

For comparison you can use npm pack, this should work in both cases.

**What is the expected behavior?**
It should be possible to pack and publish a package with files in directories.

**Please mention your node.js, yarn and operating system version.**
I have tested this with node 6.2.1 and 6.9.1, yarn 0.15.1 and current master, osx 10.12

torifat:
Couldn't reproduce this with the current `master`.

PR #1443, opened 2016-10-25
add failing tests for pack with files in directory

t-obi:
**Summary**

pack seems to be broken if the package contains files in directories, see #1442 .
This PR just adds failing tests to illustrate the problem.

**Test plan**

Run yarn tests to see the failing tests


t-obi:
It's also weird that different tests are failing and passing on travis compared to locally on my system.
On circleci the same tests are passing and failing as locally for me.

yarn.lock
```diff
@@ -4705,6 +4712,14 @@ tapable@^0.2.3, tapable@~0.2.3:
   version "0.2.4"
   resolved "https://registry.yarnpkg.com/tapable/-/tapable-0.2.4.tgz#a7814605089d4ba896c33c7e3566e13dcd194aa5"
 
+tar-fs:
```
bestander: Could you use latest master branch to run `yarn add tar-fs` for this change?
In 0.16 we had an issue that generated a lockfile entry without version

t-obi:
Thanks for having a look at this!
I have rebased and ran yarn add tar-fs with current master.
The lint-error should also be gone now.

This is still work in progress though and should probably not be merged yet because some things are not implemented, for example bundled dependencies and the "files" field in package.json should not work at all.
It would be great to get some feedback on this though, because I'm not sure if i'm going in the right direction at all. For example I was not sure if it is ok to add tar-fs as a dependency, but it made everything simpler in my opinion.

__tests__/commands/pack.js
```diff
@@ -0,0 +1,226 @@
+/* @flow */
```
bestander: Thanks for tests 👍

package.json
```diff
@@ -38,6 +38,7 @@
     "semver": "^5.1.0",
     "strip-bom": "^2.0.0",
     "tar": "^2.2.1",
+    "tar-fs": "^1.14.0",
     "tar-stream": "^1.5.2",
```
bestander: I think tar-stream is not used anymore

bestander:
I'll close this for now because no one got back to this for a long time.
Feel free to reopen

PR #1444, opened 2016-10-25
Feature: Interactive update

torifat:
This is similar to `npm-check` interactive update mode. It provides an easy way to update outdated packages.

I used the `update` command as par @cpojer's recommendation.

```
$ yarn update
```

> Currently, there is no test with this PR. But, If someone can guide me with some scenarios, I will write them, Thanks.

screenshot 2016-10-25 20 55 29

### DIFF
--- a/src/cli/aliases.js
+++ b/src/cli/aliases.js
@@ -28,7 +28,7 @@ export default ({
   rm: 'remove',
   show: 'info',
   uninstall: 'remove',
-  update: 'upgrade',
+  udpate: 'update',
   verison: 'version',
   view: 'info',
 }: { [key: string]: ?string });
--- a/src/cli/commands/update.js
+++ b/src/cli/commands/update.js
@@ -52,7 +52,7 @@ export async function run(
   const install = new Install(flags, config, reporter, lockfile);
   const [deps] = await install.fetchRequestFromCwd();
 
-  const allOutdatedDeps = (await Promise.all(deps.map(async ({pattern, hint}): Promise<Dependency> => {
+  const allDeps = (await Promise.all(deps.map(async ({pattern, hint}): Promise<Dependency> => {
     const locked = lockfile.getLocked(pattern);
     if (!locked) {
       throw new MessageError(reporter.lang('lockfileOutdated'));
@@ -77,7 +77,7 @@ export async function run(
   const isDepOld = ({latest, current}) => latest !== current;
   const isDepExpected = ({current, wanted}) => current === wanted;
 
-  const outdatedDeps = allOutdatedDeps
+  const outdatedDeps = allDeps
     .filter(isDepOld)
     .sort((depA, depB) => {
       if (isDepExpected(depA) && !isDepExpected(depB)) {
--- a/src/cli/commands/update.js
+++ b/src/cli/commands/update.js
@@ -52,7 +52,7 @@ export async function run(
   const install = new Install(flags, config, reporter, lockfile);
   const [deps] = await install.fetchRequestFromCwd();
 
-  const outdatedDeps = (await Promise.all(deps.map(async ({pattern, hint}): Promise<Dependency> => {
+  const allOutdatedDeps = (await Promise.all(deps.map(async ({pattern, hint}): Promise<Dependency> => {
     const locked = lockfile.getLocked(pattern);
     if (!locked) {
       throw new MessageError(reporter.lang('lockfileOutdated'));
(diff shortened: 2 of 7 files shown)


cpojer:
This is great, I think we should add this to Yarn.

yarn.lock
```diff
@@ -4431,10 +4466,21 @@ run-async@^0.1.0:
...
+  resolved "https://registry.yarnpkg.com/run-async/-/run-async-2.2.0.tgz#8783abd83c7bb86f41ee0602fc82404b3bd6e8b9"
+  dependencies:
+    is-promise "^2.1.0"
+    pinkie-promise "^2.0.0"
+
 rx-lite@^3.1.2:
   version "3.1.2"
   resolved "https://registry.yarnpkg.com/rx-lite/-/rx-lite-3.1.2.tgz#19ce502ca572665f3b647b10939f97fd1615f102"
 
+rx@^4.1.0:
```
Daniel15: rx is very large, I wonder how we feel about pulling it into Yarn. How much does this increase the installed size of Yarn? (you can run `scripts\build-dist.sh` on Linux/Mac or `scripts\build-dist.ps1` on Windows and compare the size of the `dist` directory before and after your change)

Daniel15:
This is probably my favourite pull request ever. Thank you! ❤️

yarn.lock
torifat: 3409098 - 2831040 = 578058 bytes increased 😭

torifat: It seems it can be reduced by loading only necessary modules from `rx` in `Inquirer.js`.

https://github.com/SBoudrias/Inquirer.js/search?utf8=%E2%9C%93&q=rx

Daniel15: Hmm, maybe this is okay for now and the size increase is offset by the usefulness of the feature. I'll see what others on the team think!

src/cli/aliases.js
```diff
@@ -28,7 +28,7 @@ export default ({
   rm: 'remove',
   show: 'info',
   uninstall: 'remove',
-  update: 'upgrade',
+  udpate: 'update',
```
andyeskridge: A tiny typo: 'udpate' -> 'update'

src/cli/commands/update.js
```diff
@@ -0,0 +1,155 @@
...
+    if (PackageRequest.getExoticResolver(pattern) ||
+        PackageRequest.getExoticResolver(normalized.range)) {
+      latest = wanted = 'exotic';
+    } else {
+      ({latest, wanted} = await config.registries[locked.registry].checkOutdated(config, name, normalized.range));
+    }
+
+    return ({name, current, wanted, latest, hint});
+  })))
+  .filter(({latest, current}) => latest !== current)
```
kittens: Instead of chaining can we assign these to variables with descriptive names, kind of hard to follow

src/cli/commands/update.js
```diff
@@ -0,0 +1,155 @@
...
+    if (isExpected(depA) && !isExpected(depB)) {
+      return 1;
+    }
+    return -1;
+  });
+
+  const getNameFromHint = (hint) => hint ? `${hint}Dependencies` : 'dependencies';
+
+  const maxLengthArr = outdatedDeps.reduce(
+    (acc, dep) =>
```
kittens: I find this code really hard to follow, could we do the same here

src/cli/commands/update.js
torifat: Did you mean assign the output in a variable then filter it?

torifat:
@kittens can you recheck please?

kittens:
Merged and will be available in the next release. I've renamed the command from `update` to `upgrade-interactive` because having two commands called `upgrade` and `update` is confusing. It's not the ideal name but we want to cut a release soon and I'd love to get this in. Happy to accept a PR renaming it to something else.

AlicanC:
`yarn upgrade -i`?

kittens:
Considered that but I think relegating this sort of functionality to an argument makes it seem less important. I think this feature is crazy awesome and deserving of it's own command.

torifat:
Need to update the outdated condition. Outdated command usage semver package now.

ccorcos:
what do the colors indicate?

Kovensky:
IIUC, yellow is for packages installed at their `package.json` version, while red is for packages that don't even match the current `package.json` (e.g. not installed).

bobmoff:
is there a reason for `upgrade-interactive` not being in the documentation? seems like a hidden gem.

Daniel15:
It's likely that it's just because nobody has written docs for it yet. Pull requests are appreciated :)

bobmoff:
@Daniel15  tried updating the documentation by cloning the website repo and running it on my dev machine, but as i have no experience with jekyll i could not get it fully working. i updated all the places i could find but the sidebar would still not show a link to the new page so i am stuck. :(

tommedema:
Is there a way to detect unused dependencies with yarn?

Issue #1306, opened 2016-10-20
Install with 'latest' tag, installs as version string 'latest'


kwelch:
Yarn: v0.16.1
Node: v6.2.0

Command ran: 
`yarn add <package>@latest`
Actual result (in `package.json`)
`"<package>": "latest"`
Expected Result
`"<package>": "^7.0.0"`

I would lean on the side of this is a feature request to gain parity with this command: `npm install --save <package>@latest`

aethant:
Ill see if I can take a look.

aethant:
@samccone I'm working on implementing this, however, I'm wondering where (if anywhere) I might be able to access the requested repo's dist tags, i.e. - `npm info <package> dist-tags`.

akshbn:
Hi! i have fixed this issue. @aethant Is it ok if i send a PR for this? or do you want to do this on your own? 
I am fine with both. 

Apologies for going ahead with the fix even though you had taken it. I was very excited to contribute to this project and this seemed like a low hanging fruit.

aethant:
Just submit your PR.

DenisGorbachev:
@akshbn this bug is still unfixed - could you please submit your PR?

FredyC:
Yarn version 0.18.0 and bug is still present.

PR #1454, opened 2016-10-25
show version number of a package in package.json and yarn.lock inplace of  dist-tag

akshbn:
**Summary**
Changed the logic in savePackages function of add.js to replace the dist-tag with its version number.
Also since the lockfile is written before the patterns can be changed in the savePackages(), the lockfile should be written again with the modified pattern.

This PR addresses issue #1306


akshbn:
Sure! Will do this by the coming weekend.

FredyC:
@akshbn Looks like your coming weekend that happened month ago got was busy. What about this weekend? Would love to see this merged :)

bestander:
This got fixed

PR #1464, opened 2016-10-26
Bugfixes to yarn pack, partly fixes #755

### LINKED ISSUE #755: published package is corrupted if "files" is used in package.json
bug
when using the "files" array to list files included in the package and publishing with yarn, the publish seems to work (there is no error message). However installing the package fails with the following error: 

```
error https://registry.npmjs.org/yarntest2/-/yarntest2-1.0.0.tgz: invalid tar file
    at Extract.Parse._startEntry (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/tar/lib/parse.js:149:13)
    at Extract.Parse._process (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/tar/lib/parse.js:131:12)
    at BlockStream.<anonymous> (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/tar/lib/parse.js:47:8)
    at emitOne (events.js:96:13)
    at BlockStream.emit (events.js:188:7)
    at BlockStream._emitChunk (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/block-stream/block-stream.js:145:10)
    at BlockStream.resume (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/block-stream/block-stream.js:58:15)
    at Extract.Reader.resume (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/fstream/lib/reader.js:249:34)
    at Entry.<anonymous> (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/tar/lib/parse.js:261:8)
    at emitNone (events.js:86:13)
```

I have downloaded the tarball and it seems to be corrupted.
To double check I have:
- published without "files" with yarn - this works, can be installed correctly
- published with "files, but with npm - this works too, can be installed correctly

- create a simple package and publish with yarn
- add files array and include your index.js, for example. publish again
- try to install, or download the tarball. this should produce the error

Small repo where I tried this: https://github.com/t-obi/yarntest
and the corresponding npm package: https://www.npmjs.com/package/yarntest2
- 1.0.0 published with yarn and files, this is broken
- 1.0.1 published with yarn without files, this works
- 1.0.2 published with npm with files, this works

It should always work ;)

node v6.2.1, yarn 0.15.1, OSX 10.12

t-obi:
**Summary**
Contains two fixes for yarn pack if "files" is used in package.json:
- some files (package.json, readme, license, changelog) should always be included, even if not listed in the files field. This was not the case, fixed by 13e7b420
- all other files should be excluded, but the exclude filter did not match files starting with a dot (this is the default behaviour of minimatch it seems). fixed by 8851acba

Please note that this only partly fixes #755 .
As far as I can see what happens in #755 is the following:
- some files which should be included we not included, in my example it was package.json
- some files which should be excluded were included, in my example .gitignore and others.

While working on this I was wondering why a missing file would create a corrupt archive, but it seemed to be fixed at first. Then I added some more testcases and noticed that it is still possible to  create corrupt archives, but this is not limited to the usage of "files" - see #1442.
Basically I think some directory was falsely included, and that invoked the bug described in #1442 .
But a missing or included file does not directly result in a corrupt archive.
Hope this makes sense, I find it difficult to explain right now.

**Test plan**
I have added some tests for this, just run yarn test.
If you want to see the failing tests checkout 6612ec6f.

### DIFF
--- a/__tests__/commands/pack.js
+++ b/__tests__/commands/pack.js
@@ -1,0 +1,182 @@
+/* @flow */
+import type {Reporter} from '../../src/reporters/index.js';
+
+import * as fs from '../../src/util/fs.js';
+import assert from 'assert';
+import {run as pack} from '../../src/cli/commands/pack.js';
+import * as reporters from '../../src/reporters/index.js';
+import Config from '../../src/config.js';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
+
+const path = require('path');
+const os = require('os');
+const stream = require('stream');
+
+const zlib = require('zlib');
+const tar = require('tar');
+const fs2 = require('fs');
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'pack');
+
+export function runPack(
+  flags: Object,
+  name: string,
+  checkArchive?: ?(config: Config) => ?Promise<void>,
+): Promise<void> {
+  return run(() => {
+    return pack;
+  }, flags, path.join(fixturesLoc, name), checkArchive);
+}
+
+export async function run(
+  factory: () => (config: Config, reporter: Reporter, flags: Object, args: Array<string>) => Promise<void>,
+  flags: Object,
+  dir: string,
+  checkArchive: ?(config: Config) => ?Promise<void>,
+): Promise<void> {
+  const cwd = path.join(
+    os.tmpdir(),
+    `yarn-${path.basename(dir)}-${Math.random()}`,
+  );
+  await fs.unlink(cwd);
+  await fs.copy(dir, cwd);
+
+  for (const {basename, absolute} of await fs.walk(cwd)) {
+    if (basename.toLowerCase() === '.ds_store') {
+      await fs.unlink(absolute);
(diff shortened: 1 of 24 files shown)


t-obi:
I fixed the tests for node 4 but travis is still failing, I have no idea why.
I anyone has an idea why please let me know.

t-obi:
Ok I have updated to skip the failing test for now, but the error message ist still there.
So it seems the test is still run just not marked as fail - is that how it is supposed to work?

bestander:
@t-obi, I had to revert the merge because of CI issue https://circleci.com/gh/yarnpkg/yarn/2330.
Could you rebase and send a new PR and ping me?

PR #1470, opened 2016-10-26
Tests/resolvers/exotics

olingern:
**Summary**
The goal is to bring all resolvers to 80%+ coverage.

**Motivation** 
Stability. Less red, more green :)

**Test plan**
Will commit each file under this PR

screen shot 2016-10-26 at 12 45 19 am


kittens:
Thank you!

PR #1472, opened 2016-10-26
Set node env to production with --prod install.

samccone:
🐐  compat 🚀

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -330,6 +330,7 @@ config.init({
   ignoreScripts: commander.ignoreScripts,
   offline: commander.preferOffline || commander.offline,
   looseSemver: !commander.strictSemver,
+  production: commander.production,
 }).then(() => {
   const exit = () => {
     process.exit(0);
--- a/src/config.js
+++ b/src/config.js
@@ -32,6 +32,7 @@ export type ConfigOptions = {
   ignorePlatform?: boolean,
   ignoreEngines?: boolean,
   cafile?: ?string,
+  production?: boolean,
 
   // Loosely compare semver for invalid cases like "0.01.0"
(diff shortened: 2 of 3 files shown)


samccone:
Test failure is from flake

kittens:
This pollutes the global environment. We should be setting `env.NODE_ENV` in `execute-lifecycle-script.js` and pass info to it from `config.executeLifecycleScript.js`.

samccone:
sg

samccone:
Ended up attaching the production bit to the config since that seems to make sense, and also since the config makes it all the way down to execute lifecycle script.

I did a version where I added a new arg to config.executeLifeCycle and then passed it through to execFromManifest and all the way down into the lifecycle execute and then set the flag there but... things got kinda messy. For that reason given the config that gets passed around I think this approach makes sense.

Now one thing that is interesting is that I now mutate the config object in the install phase which seems OK but a little different from just instantiating the config object with the option.

Given where the config object is built (prior to the flags being parsed) it seemed like that approach would also be odd.

Let me know what you think about my current approach and thanks for the 👓

src/cli/commands/install.js
```diff
@@ -134,6 +134,8 @@ export class Install {
     this.config = config;
     this.flags = normalizeFlags(config, flags);
 
+    this.config.production = this.flags.production;
```
kittens: We can move that to `config.init` in `cli/index.js` and move the `--production` flag from `install.js` to `cli/index.js`.

kittens: #1517 will make this easier.

samccone: yes good call.

samccone: addressed :goat:

PR #1473, opened 2016-10-26
Add webhook for archiving CircleCI build artifacts

Daniel15:
**Summary**
Archives builds from CircleCI into our directory for nightly builds. I'm using the phrase "nightly" but I really mean "unstable" as it's archived after every build, not once per night.

**Test plan**
¯\\\_(ツ)\_/¯

This probably needs more testing - The script it's calling intentionally ignores all requests that aren't for the master branch of yarnpkg/yarn repo as we don't want to archive random stuff from other repos or branches, but that means that I can't actually test it in my fork at the moment 😛 I might just create a copy of the archiving script and play with it some more in my fork.


Daniel15:
Debugging webhooks on CircleCI is impossible (they don't show **any** details when the webhook fails) so I'm going to add some logging at my end.

Daniel15:
OK this should be good to go now.

Issue #1404, opened 2016-10-24
Yarn resolves ^dependency to a version that's too high.


sephie:
Hi all,

I'm a bit apprehensive, because this issue appears to be caused in principle by having the latest version of Browserify, which seems a very common thing to have, but here goes:

**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Fails to install an existing package.json that works fine with NPM install.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
$ npm ls punycode
`-- browserify@13.1.0 // requests punycode@^1.3.2
  +-- punycode@1.4.1 
  `-- url@0.11.0     // requests punycode@1.3.2
    `-- punycode@1.3.2 
```
```
$ yarn
...
[2/4] 🚚  Fetching packages...
error punycode@2.0.1: The engine "node" is incompatible with this module. Expected version ">=6".
error Found incompatible module
```
Yarn should not resolve any Punycode dependency to 2.x.x, yet it does, to 2.0.1, which causes the engines error.

**What is the expected behavior?**
Punycode should install 1.4.1, the last before bump to 2.x.x, and not give an error.

**Please mention your node.js, yarn and operating system version.**
OS X 10.11.6
Node 5.12.0
Yarn 0.16.1

*the amount of version is too damn high meme*
Tom

davidmerrique:
Yep! Also has a problem with `>=` #1158

mrichards-mdsol:
This can be a blocker for adoption.

swagger doesn't install with yarn for a similar reason.
Using Node 4.6.1 (latest recommended release) I get the following:
```
error swagger-editor@2.10.4: The engine "node" is incompatible with this module. Expected version ">=5.8.0".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
Swagger has a dependency of ^2.9.2, but swagger-editor has an engine limitation, so yarn should try an older version (eg; 2.9.9) or make it easy to manually provide an alternative preferred version.

kittens:
Fixed via #1477.

Foxhoundn:
Hello @kittens ,

Sorry, but I am still experiencing this with yarn v0.16.0:

```
npm ls punycode
qorus-web-app@0.9.2 /Users/filipwitosz/Code/Projects/qorus
+-- jsdom@9.8.3
| `-- tough-cookie@2.3.2
|   `-- punycode@1.4.1 
[...]

Issue #1272, opened 2016-10-19
Shared dependencies of dependencies resolution question


otbe:
**Do you want to request a *feature* or report a *bug*?**
bug?

**Please mention your node.js, yarn and operating system version.**
yarn 0.16, node 6.5.0 and OS X 10.12.1

**What is the current behavior?**
```npm``` and ```yarn``` differ in the way how deps of deps are resolved. 

For example:
There is a public module called ```something``` and its latest version is 0.14.41.
module As package.json:
```
{
   "name": "moduleA",
   "version": "1.0.0",
   "dependencies": {
      "something": "^0.14.30"
   }
}
```

module Bs package.json:
```
{
   "name": "moduleB",
   "dependencies": {
      "something": "0.14.40" 
      "moduleA": "^1.0.0"
   }
}
```
```npm``` will install one version (0.14.40) of ```something``` in module Bs node_modules folder, but ```yarn``` will install two versions. 0.14.0 in module Bs and 0.14.41 in module As node_modules folder. 

Compared to npm this leads to unexpected behavior for bundlers and other processing tools (like TS). As a side note: ```something``` is a public scoped package, maybe this is related to the scoping issues.

**What is the expected behavior?**
I like the way how ```npm``` solves this problem :)
I know the ```--flat``` option for ```yarn install```, but I don't want to choose between version ```1.0.0``` and ```2.0.0```of a package. I think ```yarn``` should simply install semver compatible versions as flat as possible. In my example above 0.14.40 matches requirements from both modules (I think this is what ```npm``` does).

Maybe my understanding is wrong, but whats the expected behavior in this case? 
Thanks!

davidmerrique:
I think this is the same as or related to #1158

This is definitely a problem.

otbe:
Indeed! Sorry for duplicate, searched a while for an related issue 👍

kittens:
Fixed via #1477.

Issue #1158, opened 2016-10-17
Strange version resolving


sc0rp10:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
package.json
```json
{
  "dependencies": {
    "devbridge-autocomplete": "^1.2.26",
    "jquery": "2.1.2"
  }
}
```
Will install:
```yaml
devbridge-autocomplete:
  version "1.2.26"
  resolved "https://registry.yarnpkg.com/devbridge-autocomplete/-/devbridge-autocomplete-1.2.26.tgz#e491e01b701ba0ff509afaf4ee1eca69a9d00647"
  dependencies:
    jquery ">=1.7"

jquery@>=1.7:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.1.1.tgz#347c1c21c7e004115e0a4da32cece041fad3c8a3"

jquery@2.1.2:
  version "2.1.2"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-2.1.2.tgz#b68f154cb2ea4731924883e9fe20ec199d1dc1e2"


```
**What is the expected behavior?**
Install jquery once with version = 2.1.2
**Please mention your node.js, yarn and operating system version.**
```
$ node -v
v6.7.0
$ yarn -v
yarn install v0.15.1
```

davidmerrique:
Same problem here

```json
{
  "dependencies": {
    "jquery": "2.2.4",
    "parsleyjs": "2.5.1"
  }
}
```

```yaml
jquery@>=1.8.0:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.1.1.tgz#347c1c21c7e004115e0a4da32cece041fad3c8a3"

jquery@2.2.4:
  version "2.2.4"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-2.2.4.tgz#2c89d6889b5eac522a7eea32c14521559c6cbf02"

parsleyjs@2.5.1:
  version "2.5.1"
  resolved "https://registry.yarnpkg.com/parsleyjs/-/parsleyjs-2.5.1.tgz#6380924e231c38b6eca3a597d7ebc0dde41ceb9f"
  dependencies:
    jquery ">=1.8.0"
```

Parsley depends on `jquery@>=1.8.0`, but `3.1.1` is still being installed even though `2.2.4` should satisfy it.

vladferix:
[...]

Issue #112, opened 2016-07-14
deps inconsistency for yeoman-generator 


bestander:
Still investigating, scenario is:

```
cd react-native-github
rm -rf node_modules
kpm install
npm shrinkwrap --dev
npm ERR! Darwin 15.5.0
npm ERR! argv "/Users/bestander/.nvm/versions/node/v6.2.0/bin/node" "/Users/bestander/.nvm/versions/node/v6.2.0/bin/npm" "shrinkwrap" "--dev"
npm ERR! node v6.2.0
npm ERR! npm  v3.8.9

npm ERR! Problems were encountered
npm ERR! Please correct and try again.
npm ERR! invalid: have glob@6.0.4 (expected: ^5.0.3) /Users/bestander/work/react-native/node_modules/yeoman-generator/node_modules/glob
npm ERR! invalid: have lodash@4.13.1 (expected: ^3.5.0) /Users/bestander/work/react-native/node_modules/yeoman-generator/node_modules/lodash
npm ERR! invalid: have run-async@2.2.0 (expected: ^0.1.0) /Users/bestander/work/react-native/node_modules/yeoman-generator/node_modules/run-async
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/bestander/work/react-native/npm-debug.log
```

npm shrinkwrap checks for folders consistency and here it fines wrong versions installed deep

PR #1477, opened 2016-10-26
fix-1158: resolve dep verions by ranges, not exact versions (#1158) (#1272) (#1404)

verth:
**Summary**

Smarter/looser dep version number resolving -- by range instead of exact version. ([comment and code disagree](https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L218-L220))

`package-request` will resolve a package to the **highest** available version, and check if that package version has already been resolved with `package-resolver`'s **exact version match** `getExactVersionMatch`. 

Instead of matching the exact highest version: match satisfying resolved ranges of the requested package, and pick the highest available one, *if* possible. *If not* possible, continue and resolve to the highest possible version as before.

https://github.com/yarnpkg/yarn/issues/112 originally says and tests the opposite of this. The concerns in this issue have been tested as-well and are succeeding: `npm shrinkwrap` && `yarn check`. Tests have been adapted, since the setup was largely the same.

Resolves:
* https://github.com/yarnpkg/yarn/issues/1158
* https://github.com/yarnpkg/yarn/issues/1272
* https://github.com/yarnpkg/yarn/issues/1404

**Test plan**

```
{
  "dependencies": {
    "devbridge-autocomplete": "^1.2.26",
    "jquery": "2.1.2"
  }
}
```

```
$ yarn install
$ npm shrinkwrap --dev
$ yarn check
```

Open `yarn.lock` and see that range-matching dependencies are shared.

### DIFF
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -290,30 +290,27 @@ test.concurrent('install should dedupe dependencies avoiding conflicts 7', (): P
 
 test.concurrent('install should dedupe dependencies avoiding conflicts 8', (): Promise<void> => {
   // revealed in https://github.com/yarnpkg/yarn/issues/112
+  // adapted for https://github.com/yarnpkg/yarn/issues/1158
   return runInstall({}, 'install-should-dedupe-avoiding-conflicts-8', async (config) => {
     assert.equal(await getPackageVersion(config, 'glob'), '5.0.15');
-    assert.equal(await getPackageVersion(config, 'yeoman-generator/globby/glob'), '6.0.4');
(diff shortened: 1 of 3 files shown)


FDiskas:
Waiting for this fix to be released. When?

FDiskas:
I think that this FIX should be released as soon as possible. This is major problem. The package manager  who can't resolve dependency correctly.
What are we waiting? The Christmas?

wyze:
Hey @FDiskas, I'm not sure when the next stable release is, but if you would like, you could try out the nightly release by doing the following:

```
wget https://yarnpkg.com/install.sh
chmod +x install.sh
./install.sh --nightly
```

FDiskas:
Thanks

PR #1480, opened 2016-10-26
Updating the var's to let and const. 

lholmquist:
This updates all the `var` declarations in the ./bin/yarn.js file with `let` and `const`

The motivation here is that it brings this file inline with the other files in the project.

### DIFF
--- a/bin/yarn.js
+++ b/bin/yarn.js
@@ -1,25 +1,24 @@
 #!/usr/bin/env node
 
-/* eslint-disable no-var */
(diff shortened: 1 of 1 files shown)


kittens:
This file can be loaded by any version of Node and we need to display an error message saying that the node version isn't supported, for example on Node 0.10 that doesn't support `let`.

lholmquist:
ah, yes.  with node 0.10 about to become unsupported, will this "any version of node" change in the future?

PR #1452, opened 2016-10-25
Minor README tweaks

Zearin:
formatting, phrasing, etc.

PR #1486, opened 2016-10-26
Remove reference to Bower from README

chriskrycho:
Since Bower support was removed in #1441, update the README to match.


wyze:
This was removed with #1452.

chriskrycho:
Awesome, didn't see that (or the update—might just have been caching!). Thanks!

PR #1489, opened 2016-10-26
Git branches can have hashes as part of their name

jure:
**Summary**

Given a dependency:
```
dependencies: {
  "substance": "jure/substance#fix-lens-#96"
}
```

Yarn raises an error: 
```
error Invalid hosted git fragment "jure/substance#fix-lens-#96".
```

This is because the current code https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/resolvers/exotics/hosted-git-resolver.js#L29 bails if there is more than one hash in the repo part.

**Test plan**

Added two tests, one for a branch with hashes and one for a `feature/fix-issue` (no hash) branch.

### DIFF
--- a/__tests__/resolvers/exotics/hosted-git-resolver.js
+++ b/__tests__/resolvers/exotics/hosted-git-resolver.js
@@ -1,0 +1,30 @@
+/* @flow */
+
+import {explodeHostedGitFragment} from '../../../src/resolvers/exotics/hosted-git-resolver.js';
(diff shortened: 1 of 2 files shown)


samccone:
I LOVE EDGE CASES :)

jure:
@samccone I feel ya, I expected to find an existing bullet-proof parser for git dependencies in package.json, but I got lost along the way. Can we use what npm is using?

PR #1492, opened 2016-10-26
[git] Upgrade to secure url

yunxing:
https://github.com/yarnpkg/yarn/pull/1365/files introduced a fix to enforce secure protocols. This unfortunately broke a lot of packages we have which depend on 'git' protocol (e.g., 'npm show @opam-alpha/core').

Instead of refusing to work when a unsecure protocol is provided, maybe a better way here is to automatically upgrade to a secure url.


bestander:
@yunxing could you rebase please?

Issue #815, opened 2016-10-12
Path of local dependencies is not resolved correctly


ds82:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
Resolution of local modules fails. Suppose a repo like this:

modules/a <-- Package A
modules/b <-- Package B, depends on a (`file:../a`)
main/   <-- main Package, depends on b (`file:../modules/b`)

yarn (install) fails, because it tries to resolve _a_ with `file:../a` instead of `file:../modules/a`

**If the current behavior is a bug, please provide the steps to reproduce.**
made a example repo: https://github.com/ds82/yarn-test-local-modules

> git clone https://github.com/ds82/yarn-test-local-modules
> cd yarn-test-local-modules/main
> yarn

`error "/Users/dsaenger/code/yarn-test-local-modules/a" doesn't exist`

**What is the expected behavior?**
yarn should resolve the correct path of local modules like npm does. _npm install works in the repo posted above_

**Please mention your node.js, yarn and operating system version.**
node v4.6, osx 10.10.5, yarn 0.15.1

ds82:
this patch seems to fix the problem (but I consider it a hack ./. temporary workaround for me):

```
diff --git i/src/resolvers/exotics/file-resolver.js w/src/resolvers/exotics/file-resolver.js
index 134c0f7..e99180b 100644
--- i/src/resolvers/exotics/file-resolver.js
+++ w/src/resolvers/exotics/file-resolver.js
@@ -23,7 +23,13 @@ export default class FileResolver extends ExoticResolver {
   async resolve(): Promise<Manifest> {
     let loc = this.loc;
     if (!path.isAbsolute(loc)) {
-      loc = path.join(this.config.cwd, loc);
+      let cwd = this.config.cwd;
+
+      if (this.request.parentRequest && this.request.parentRequest.pattern) {
+        cwd = path.join(cwd, this.request.parentRequest.pattern.replace(/.*@(file\:)?/, ''));
+      }
+
+      loc = path.join(cwd, loc);
     }
     if (!(await fs.exists(loc))) {
       throw new MessageError(this.reporter.lang('doesntExist', loc));
```

nearwood:
I have the exact same issue. Using your patch fixed it, though I did have to modify `this` -> `_this` for 0.15.1, installed on Windows via MSI:

`C:\Program Files (x86)\Yarn\lib\resolvers\exotics\file-resolver.js`
```
[...]

Issue #1263, opened 2016-10-19
nested dependencies with relative paths fails


chandlerprall:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Yarn is unable to locate sub-dependency referenced via relative file paths

**If the current behavior is a bug, please provide the steps to reproduce.**
Created [a gist](https://gist.github.com/chandlerprall/5c5e83df5e1e2281289366c8b01ea21e) with three package.json files (underscores in file names represent path separators). The folder structure is:

`pkg-a/package.json`
`pkg-b/package.json`
`subdir/pkg-c/package.json`

Each package depends on the previous one (`a` has no dependency, `b` depends on `a`, and `c` depends on `b`)

`cd`ing into `pkg-b` and running `yarn` works as expected. However, `cd`ing into`subdir/pkg-c` and running `yarn` gives the following error:

`error "/Users/chandler.prall/yarntest/subdir/pkg-a" doesn't exist.` as it resolves `pkg-b`'s dependency `file:../pkg-a` relative to `subdir/pkg-c` instead of `pkg-b`

`npm install` handles this case correctly

**What is the expected behavior?**
Running `yarn` inside `subdir/pkg-c` should follow nested relative path dependencies correctly and resolve `file:pkg-a` from `pkg-b` path.

**Please mention your node.js, yarn and operating system version.**
node.js v6.7.0
yarn v0.16.1
OS X 10.9.4

hawkrives:
Can you give an example of your `package.json`, please?

chandlerprall:
the three `package.json` files are at https://gist.github.com/chandlerprall/5c5e83df5e1e2281289366c8b01ea21e

turnbullm:
Looks like a duplicate of #815 ?

chandlerprall:
I agree, issue 815 is the same problem.

PR #1498, opened 2016-10-27
File url normalization

xandris:
**Summary**

When resolving dependencies against `file:` URLs that have relative paths, descendent modules' `file:` URLs were resolving against the CWD instead of against `package.json`.

I'm not claiming that depending on packages in other directories is a good idea, but this is a use case that npm seems to support but yarn doesn't. It seems like it should work but doesn't.

**Test plan**

Given a set of modules:

```
.
├── package.json
├── root-a
│   ├── index.js
│   └── package.json
├── root-b
│   └── package.json
└── sub
    └── sub-a
        └── package.json
```

The dependencies go like this: `. -> sub-a -> root-b -> root-a`.

Root calls `sub-a` `file:sub/sub-a`.

`sub-a` calls `root-b` `file:../../root-b`

`root-b` calls `root-a'`file:../root-a`

Saying `yarn` correctly resolves `sub-a`, but then looks up two directories from CWD to try to find `root-b` and fails:

```
yarn install v0.16.1
info No lockfile found.
warning No license field
error "/Users/aparker/src/yarn/__tests__/fixtures/root-b" doesn't exist.
```

This pull requests attempts to fix that by normalizing the in-memory `package.json` to resolve all `file:` URLs upon reading a `package.json` and relativizing them against CWD. I'm not sure if this is the correct approach but it seems to work.

Another approach I considered was attaching the current manifest to `PackageRequest` during resolution and retrieving that info in `FileResolver`, using the base path of the manifest, if any, instead of `this.config.cwd`.


hawkrives:
This should fix https://github.com/yarnpkg/yarn/issues/815, I think (and, by extension, https://github.com/yarnpkg/yarn/issues/1263).

kittens:
`normalize-manifest` isn't really the right place for this kind of thing. Ideally we should be doing the normalisation [here](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/file-resolver.js#L26).

xandris:
@kittens I tried that approach too and it worked. The only weird thing was the original relative paths ended up in `yarn.lock`. I just started poking around the codebase yesterday so I don't know if that's an issue.

verth:
@kittens to add on your comment: do you mean normalizing only the `loc` on resolve before reading the manifest? 

If so, I think it would show a *side-issue* (that this PR solves), that is: pattern names are a concatenation of [dependency properties ](https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L257) from the manifest. This would lead to successful install but a misleading `yarn.lock` with confusing paths. 

Giving this example, `yarn.lock` would contain:
* `root-b@file:../../root-b`, while it should be `name@file:./root-b` I believe

Doing the normalization later **on** the `.dependencies` after reading the manifest **in** `file-resolver` would then probably be the best of two worlds ✨ ?

@xandris 

While checking this out I noticed two small things (epic otherwise!):
* I think you are missing `optionalDependencies` to be fixed?
* There is a fancy [removeSuffix](https://github.com/yarnpkg/yarn/blob/master/src/util/misc.js#L18) method in existance

xandris:
@verth @kittens yeah I forgot about `optionalDependencies` since I've never used them 😖 

How should I proceed? Should I do the same thing, but in `FileResolver` to the `manifest` it reads?

```javascript
const manifest = await this.config.readManifest(loc, this.registry);
```

xandris:
I can't get it to work like that. It seems to work from the command line, but the unit test fails. I can't tell what's going on; it seems to run the `yarn install` command, and that works, but then it takes a second pass that has wrong paths. Help... what are the tests doing?

xandris:
I think the code that reads the manifest is caching it, so the same manifest is getting normalized by `FileResolver` multiple times... How should I handle this?

xandris:
I could add a `WeakSet` that tracks which manifests have already been normalized... That lets the test pass

verth:
Made a quick implementation with your tests, based on the the manifest normalizing as in your PR, and had green tests, could it maybe be something small you missed? I made a [Pull Request](https://github.com/verth/yarn/pull/1/files) to my own fork so you can easily see the diff, and maybe find out what could have gone wrong? 

Hope it will help you on the way!

xandris:
Yes, that helped, thank you!  Submitted for your approval, the PR now copies the manifest if and only if any of the paths were actually normalized.

ozsay:
what is the status of this PR ? it fixes our local dependencies stuff. we can't migrate to `yarn` without it.

Thanks

justin808:
@bestander Any idea when this will be shipped? Some projects are blocked by this.

bestander:
@justin808 it is already in 0.18.0

Issue #1456, opened 2016-10-25
Feature Request: Allow Yarn to log unexpected error mesages to stderr


Taytay:
**Do you want to request a _feature_ or report a _bug_?**
Feature request

**What is the current behavior?**
When Yarn gets an unexpected error, it currently tells me:
`error An unexpected error occured, please open a bug report with the information provided in "<snip>.../yarn-error.log".`

I am getting these errors on our build machine, which means that our build logs don't have the text of the error, so now I need to log into the machine to dump that file. So far, I can always identify the relevant bit of the error from the "Trace" message at the very bottom of the file, like so:

```
Trace:
  Error: EACCES: permission denied, open '/home/vagrant/.yarn/.roadrunner.json'
      at Error (native)
      at Object.fs.openSync (fs.js:549:18)
      at Object.fs.writeFileSync (fs.js:1156:15)
      at /usr/lib/node_modules/yarn/node_modules/roadrunner/index.js:25:6
      at /usr/lib/node_modules/yarn/node_modules/roadrunner/index.js:12:12
      at emitOne (events.js:82:20)
      at process.emit (events.js:169:7)
      at processEmit [as emit] (/usr/lib/node_modules/yarn/node_modules/loud-rejection/node_modules/signal-exit/index.js:140:35)
      at process.exit (node.js:750:17)
      at exit (/usr/lib/node_modules/yarn/lib-legacy/cli/index.js:380:13)
```

It would be very helpful in our case to log this information, (or maybe even just the message without the stack?) to stderr so that I wouldn't have to dig somewhere else to get error details.

So perhaps the message could change to: (Spelling of occurred corrected here as well:)

```
error An unexpected error occurred please open a bug report with the information provided in "<snip>.../yarn-error.log".
Trace:
 Error: EACCES: permission denied, open '/home/vagrant/.yarn/.roadrunner.json'
      at Error (native)
      at Object.fs.openSync (fs.js:549:18)
      at Object.fs.writeFileSync (fs.js:1156:15)
      at /usr/lib/node_modules/yarn/node_modules/roadrunner/index.js:25:6
      at /usr/lib/node_modules/yarn/node_modules/roadrunner/index.js:12:12
      at emitOne (events.js:82:20)
      at process.emit (events.js:169:7)
      at processEmit [as emit] (/usr/lib/node_modules/yarn/node_modules/loud-rejection/node_modules/signal-exit/index.js:140:35)
      at process.exit (node.js:750:17)
[...]

PR #1502, opened 2016-10-27
Show error message in stdout.

jvhoven:
Implements #1456.

When an unexpected error occurres it now shows the error message in console. I also renamed the old unexpectedError string to bugReport and gave it an info flag. 

**Test plan**

I tested this by setting the permissions of my package.json to read-only and then attempt to install a new package using `yarn add left-pad`. 

Instead of outputting:

```
error An unexpected error occurred, please open a bug report with the information provided in "/home/jeffrey/Documents/Projects/yarn-testing/yarn-error.log".
```

It now outputs:

```
error An unexpected error occurred: "Error: EACCES: permission denied, open '/home/jeffrey/Documents/Projects/yarn-testing/package.json'\n    at Error (native)".
info If you think this is a bug, please open a bug report with the information provided in "/home/jeffrey/Documents/Projects/yarn-testing/yarn-error.log".
```

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -315,7 +315,8 @@ function onUnexpectedError(err: Error) {
   const errorLoc = path.join(config.cwd, 'yarn-error.log');
   fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
 
-  reporter.error(reporter.lang('unexpectedError', errorLoc));
+  reporter.error(reporter.lang('unexpectedError', err.stack));
+  reporter.info(reporter.lang('bugReport', errorLoc));
 }
 
 //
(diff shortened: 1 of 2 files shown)


src/cli/index.js
```diff
@@ -315,7 +315,8 @@ function onUnexpectedError(err: Error) {
   const errorLoc = path.join(config.cwd, 'yarn-error.log');
   fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
 
-  reporter.error(reporter.lang('unexpectedError', errorLoc));
+  reporter.error(reporter.lang('unexpectedError', err.stack));
```
jvhoven: Should I concat err.stack? It only shows the first part of the stack e.g: `Error: EACCES: permission denied, open '/home/jeffrey/Documents/Projects/yarn-testing/package.json'\n    at Error (native)` but I'm not sure how that works as it might result in the entire stack being printed?

Daniel15: Try `err.message`.

jvhoven: 🙈, that would indeed be alot better.

PR #1530, opened 2016-10-28
Fix all variations of generateDisclaimer command to not have wrapper

### LINKED ISSUE #1191: Licence generateDisclaimer - Outputs "junk" to stdout
Bug

When running `yarn licenses generateDisclaimer` and piping stdout to a file, I would expect to get only the disclaimer text to easily integrate this feature with CI etc. Instead, the first few lines look like this:

```
yarn licenses v0.15.1
info "fsevents@1.0.14" is an optional dependency and failed compatibility check. Excluding it from installation.
THE FOLLOWING SETS FORTH ATTRIBUTION NOTICES FOR THIRD PARTY SOFTWARE THAT MAY BE CONTAINED IN PORTIONS OF THE XXX PRODUCT.
```

The third line is correct, but the two proceeding have nothing to do with the disclaimer.

At the end of the document, there is a line that says

```
Done in 5.45s.
```

This is also not a part of the disclaimer and should not be a part of the output.

I see four possible solutions. 
- The first is to send all metadata about `yarn` to `stderr`
- The second is to specially handle `generateDisclaimer` 
- The third is to create a global `-q, --quiet` flag
- The last is to add a `-o --output` option to `generateDisclaimer` that writes a file

1. Create a package.json with some dependencies (preferably with some incompatible peerDependencies)
2. Run `yarn licenses generateDisclaimer > DISCLAIMERS.txt`
3. Open `DISCLAIMERS.txt`
4. Look at the beginning and end of the document

The disclaimer file is generated without the parts preceding `THE FOLLOWING SETS FORTH ATTRIBUTION`, and without `Done in 5.45s.` at the end.

```
$ yarn --version
0.15.1
$ node --version
v6.4.0
$ uname -a
Linux thor.osl.manamind.com 4.7.7-200.fc24.x86_64 #1 SMP Sat Oct 8 00:21:59 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

tlent:
**Summary**

This fixes #1191.

Previously `yarn licenses generate-disclaimer` would not include the output wrapper header and footer but other valid variations of the same command such as `yarn licenses generateDisclaimer` would.

**Test plan**

Previous behavior:
`yarn licenses generate-disclaimer` outputs disclaimers without wrapper
`yarn licenses generateDisclaimer` outputs disclaimers with wrapper

New behavior:
`yarn licenses generate-disclaimer` and `yarn licenses generateDisclaimer` have the same output without wrapper.

### DIFF
--- a/src/cli/commands/licenses.js
+++ b/src/cli/commands/licenses.js
@@ -8,10 +8,11 @@ import {Install} from './install.js';
 import Lockfile from '../../lockfile/wrapper.js';
 import buildSubCommands from './_build-sub-commands.js';
 
+const camelCase = require('camelcase');
 const invariant = require('invariant');
 
 export function hasWrapper(flags: Object, args: Array<string>): boolean {
-  return args[0] != 'generate-disclaimer';
+  return camelCase(args[0]) != 'generateDisclaimer';
 }
 
 export function setFlags(commander: Object) {


kittens:
I'd prefer if we made the camelCased versions of the commands illegal instead of allowing this.

kittens:
Created #1586 to fix this, pending approval.

PR #1533, opened 2016-10-28
fix readme detection

mvayngrib:
**Summary**

some repos have a README folder that breaks the current heuristic, e.g.:
https://github.com/bitwiseshiftleft/sjcl

**Test plan**

added breaking test and fix (favor filenames with extensions). A more thorough but more expensive fix would be to check try to readFile all readme filename candidates

### DIFF
--- a/__tests__/fixtures/normalize-manifest/readme/README/somefile
+++ b/__tests__/fixtures/normalize-manifest/readme/README/somefile
@@ -1,0 +1,1 @@
+something
--- a/src/util/normalize-manifest/fix.js
+++ b/src/util/normalize-manifest/fix.js
@@ -73,17 +73,23 @@ export default async function (
 
   // if there's no readme field then load the README file from the cwd
   if (!info.readme) {
-    const readmeFilename = files.find((filename): boolean => {
-      const lower = filename.toLowerCase();
-      return lower === 'readme' || lower.indexOf('readme.') === 0;
-    });
+    const readmeCandidates = files
+      .filter((filename): boolean => {
+        const lower = filename.toLowerCase();
+        return lower === 'readme' || lower.indexOf('readme.') === 0;
+      })
+      .sort((filename1, filename2): number => {
+        // favor files with extensions
+        return filename2.indexOf('.') - filename1.indexOf('.');
+      });
 
-    if (readmeFilename) {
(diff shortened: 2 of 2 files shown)


Daniel15:
Won't this still fail if they have a `README.foo` folder?

mvayngrib:
@Daniel15 yes, but currently only the first candidate is examined, and my assumption is readme.xx is more likely to be a file than directory

as I said, a more thorough but more expensive fix would be to check try to readFile all readme filename candidates

zeke:
A `stat` call is pretty cheap. You just find the first file after filtering and sorting. Something like this:

```js
.find((filename): boolean => {
  await fs.stat(path_string).isFile()
})
```

mvayngrib:
@zeke sounds good

mvayngrib:
@zeke @Daniel15 updated to check all readme candidates

zeke:
This looks like it will work, but I don't think you need to store the intermediate "candidates". I was imagining adding a `find` call to the chain:

```js
const readmeFilename = files.filter().sort().find()
if (readmeFilename) // do stuff
```

mvayngrib:
@zeke i had something like this initially...but i couldn't get the async/await part to work right:
```js
    await files.filter((filename): boolean => {
        const lower = filename.toLowerCase();
        return lower === 'readme' || lower.indexOf('readme.') === 0;
      })
      .sort((filename1, filename2): number => {
        // favor files with extensions
        return filename2.indexOf('.') - filename1.indexOf('.');
      })
      .find(async function (readmeFilename): Promise<boolean> {
        const readmeFilepath = path.join(moduleLoc, readmeFilename);
        const readmeFileStats = await fs.stat(readmeFilepath);
        if (readmeFileStats.isFile()) {
          info.readmeFilename = readmeFilename;
          info.readme = await fs.readFile(readmeFilepath);
          return true;
        }

        return false;
      });
```

zeke:
I wonder if it would be consider sinful to use `fs.statSync` instead?

mvayngrib:
@zeke not sure why we would do that, just to avoid storing intermediate candidate filenames? I think statSync is a much higher price to pay

mvayngrib:
@Daniel15 @zeke so is there anything wrong with the current solution?

Issue #639, opened 2016-10-11
yarn fails with Invalid URI


joeheyming:
I just installed yarn and ran it in my local npm repository.

It fails on file: uris:

yarn install v0.15.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
⠠ toposort@^1.0.0events.js:154
      throw er; // Unhandled 'error' event
      ^

Error: Invalid URI "file:target/my-package.tgz"
    at Request.init (/usr/local/lib/node_modules/yarnpkg/node_modules/request/request.js:275:31)
    at new Request (/usr/local/lib/node_modules/yarnpkg/node_modules/request/request.js:129:8)
    at request (/usr/local/lib/node_modules/yarnpkg/node_modules/request/index.js:55:10)
    at RequestManager.execute (/usr/local/lib/node_modules/yarnpkg/lib/util/request-manager.js:302:17)
    at RequestManager.shiftQueue (/usr/local/lib/node_modules/yarnpkg/lib/util/request-manager.js:328:10)
    at /usr/local/lib/node_modules/yarnpkg/lib/util/request-manager.js:230:12
    at Request.params.callback [as _callback](/usr/local/lib/node_modules/yarnpkg/lib/util/request-manager.js:284:11)
    at Request.self.callback (/usr/local/lib/node_modules/yarnpkg/node_modules/request/request.js:187:22)
    at emitTwo (events.js:100:13)
    at Request.emit (events.js:185:7)

smilingpoplar:
+1 same problem

hanjukim:
+1 same problem

changero:
error %27http://react: Invalid URI "%27http://react"

vesteraas:
I'm experiencing the same issue. 

This fails:

`"angular2-materialize": "file:angular2-materialize-5.2.1.tgz"`

...but this works fine in yarn, as well as npm:

`"angular2-materialize": "./angular2-materialize-5.2.1.tgz"`

akshbn:
I will give this a shot.

kittens:
Created #1591 to fix this, pending approval.

joeheyming:
Wait, @kittens #1591 says to remove the file: suffix, don't you mean file: prefix?

I just tried out yarn with v0.16.1 and I still get the invalid uri error

shahmirn:
I agree with @joeheyming that the code should be using removePrefix, not removeSuffix

I also get the same error with v0.16.1

[...]

PR #1555, opened 2016-10-30
Fixes issue #639

akshbn:
**Summary**

Yarn install fails when "file:" prefix is specified in the package target in the dependency.
eg: "dependencies": {
    "blabla":"file:folder/blabla.tar.gz"
  }

**Fix**
yarn was failing because it was not able to resolve the pattern with 'file:' in it. Also, as noted in #639 yarn works if the 'file:' prefix is not provided. Hence the best solution was to extract a substring of the pattern without the 'file:' prefix. 

IT WORKS!!!

**Test plan**
pkgjson

The output

term

Also, i tested it with other paths containing folders in the path and it worked.

### DIFF
--- a/src/resolvers/exotics/tarball-resolver.js
+++ b/src/resolvers/exotics/tarball-resolver.js
@@ -50,7 +50,11 @@ export default class TarballResolver extends ExoticResolver {
       return shrunk;
     }
 
-    const {url} = this;
+    let {url} = this;
+    if (url.startsWith('file:')) {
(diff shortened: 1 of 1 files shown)


kittens:
Does this work with npm?

kittens:
Created #1591 to fix this, pending approval.

akshbn:
@kittens  it would be of great help if you could tell me how my code/ approach could have been improved.

PR #1561, opened 2016-10-30
Return registry from URL when requesting a file

### LINKED ISSUE #1318: Authentication with private registries
__tl;dr__ There is a bug in [`getScope`][getScope].

After working around #1317, the initial request for the package succeeds. The returned version contains an absolute path to the package, which is hosted at the same server, but still requires authentication.

The [`getScope`][getScope] method returns an empty string for the package URL, causing the registry settings to be ignored for this request.

[getScope]: https://github.com/yarnpkg/yarn/blob/3dd0264040858ab5ae361e3dde4b7b9e98afb099/src/registries/npm-registry.js#L135

fb55:
This ensures authentication options are passed to these requests.

Fixes #1318

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -16,6 +16,7 @@ const url = require('url');
 const ini = require('ini');
 
 const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
+const HTTPS_EXPRESSION = /^https?:(?=\/\/)/i;
 
 function getGlobalPrefix(): string {
   if (process.env.PREFIX) {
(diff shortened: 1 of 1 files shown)


src/registries/npm-registry.js
```diff
@@ -139,6 +140,15 @@ export default class NpmRegistry extends Registry {
   }
 
   getRegistry(packageName: string): string {
+    // If this is a request for a file, return the registry from the URL
+    if (HTTPS_EXPRESSION.test(packageName)) {
+      const registryProtocolHost = packageName.match(/^https?:\/\/[^\/]+\//i);
```
dfreeman: This only matches on the protocol + hostname. An npm registry URL can also include a path, e.g. for registries that are part of a larger artifact hosting application, like Nexus or Artifactory.

bestander:
Thanks for the PR, the case is covered in a bit more generic way in https://github.com/yarnpkg/yarn/pull/1666.
So I'll close this one

PR #1586, opened 2016-11-01
disallow camelised command names

kittens:


martynchamberlin:
I'm curious why we're merging stuff that's not passing CI here?

Issue #1165, opened 2016-10-17
Support TLS client certificates for connecting to the registry


cnorthwood:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behavior?**

When using a private NPM registry which is protected by TLS client certificates, the following error occurs:

`error write EPROTO 140735275094016:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:1472:SSL alert number 40
140735275094016:error:1409E0E5:SSL routines:ssl3_write_bytes:ssl handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:656:`

The cert and key are specified in the `.npmrc`, which display when running `yarn config list`. `npm install` behaves correctly.

**If the current behavior is a bug, please provide the steps to reproduce.**

N/A

**What is the expected behavior?**

Yarn uses the configured client certificate and key to connect to the registry.

**Please mention your node.js, yarn and operating system version.**

```
$ node --version
v6.3.1
$ yarn --version
0.15.1
$ sw_vers -productVersion 
10.11.6
```

Daniel15:
Does npm support client certs?

PR #1588, opened 2016-11-01
Enable using TLS client certificates for communicating with private repositories

cnorthwood:
**Summary**

NPM supports talking to repositories which are protected by TLS client certificates. The BBC (and I assume other organisations) used this feature to protect private repositories for our internal dev packages, adding this to Yarn allows us to migrate away from NPM, to Yarn, and get the workflow improvements Yarn offers.

Two new settings - `cert` and `key` are added, in the same format as `.npmrc` supports, so people can simply migrate to Yarn without any configuration changes. The interface in NPM for this is a bit clunky (it requires the key/cert as a string, rather than, for example, as a file name), but this PR does not address that, it's direct parity with NPM.

**Test plan**

A unit test is included which exercises the code path in question. I have also verified locally by taking an existing project which has a cert, key, and TLS protected repository in the `.npmrc` and running `yarn` in it. The `yarn.lock` includes the private repositories (which would otherwise be inaccessible) and the `resolved` field includes the URL of the certificate protected endpoints. This fixes the behaviour exhibited in #1165.


cnorthwood:
thanks for the speedy merge :) I'm not sure what the Yarn release process looks like, but do you have a feel on the timescale for the next stable release which would incorporate this change would be? Cheers

PR #1592, opened 2016-11-01
Nicer permission errors when trying to write global binaries

kittens:


src/reporters/lang/en.js
```diff
@@ -74,6 +74,7 @@ const messages = {
   missingWhyDependency: 'Missing package name, folder or path to file to identify why a package has been installed',
   unexpectedError: 'An unexpected error occurred, please open a bug report with the information provided in $0.',
   jsonError: 'Error parsing JSON at $0, $1.',
+  noFilePermission: "We don't have permissions to touch the file $0.",
```
cpojer: 'We'?

kittens: Yarn is talking to the user, seems appropriate, no?

cpojer: I'd just say "Yarn doesn't…".

bestander:
scripts/bootstrap-env-ubuntu.sh failed one, cc @Daniel15.
Looks like a temporary issue, I restared

kittens:
@bestander Yeah there have been 2 PRs that have added flaky tests. The end to end tests have been extremely flaky as well as a git test that was added.

Daniel15:
@kittens the end to end tests are not flaky - nothing is actually running them :P. They can't run on Circle since you can't run Docker within Docker. I'm going to be running them nightly.

Daniel15:
I think the key server we get our PGP key from is failing. Maybe we should change it to a different server, or just wget the key instead.

Daniel15:
@bestander - Hopefully #1598 should fix that 😄

PR #1598, opened 2016-11-01
Load GPG key from URL rather than keyserver

Daniel15:
**Summary**
Loads GPG key from a URL we host, rather than a keyserver. This will hopefully make builds less flaky (they won't fail when the keyserver has issues)

I tried `https` but it didn't work:
> gpgkeys: protocol 'https' not supported
> gpg: no handler for keyserver scheme 'https'
> gpg: WARNING: unable to fetch URI https://dl.yarnpkg.com/debian/pubkey.gpg: keyserver error

**Test plan**

Tested the command locally. Will check CircleCI build.

### DIFF
--- a/scripts/bootstrap-env-ubuntu.sh
+++ b/scripts/bootstrap-env-ubuntu.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # Add Yarn package repo - We require Yarn to build Yarn itself :D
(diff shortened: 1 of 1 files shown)


Daniel15:
CircleCI failure is unrelated - It's a flaky unit test. Going to merge this since adding the GPG key works fine.

PR #1620, opened 2016-11-02
Properly save dependency with `yarn upgrade [dependency]` command

### LINKED ISSUE #1458: Yarn upgrade does not properly update `package.json` for dev dependencies

_bug_

Running `yarn upgrade` to upgrade a dev dependency, even with the `--dev` flag, will add the upgraded version of the dependency to the `dependencies` hash in `package.json` rather than `devDependencies`.

This might also be an issue with other dependency types (like peer), but I have not tested.

``` bash
$ mkdir foo && cd foo
$ yarn init
<snip>
$ yarn add lodash@3 --dev
<snip>

$ grep lodash package.json -C 1
  "devDependencies": {
    "lodash": "3"
  }

$ yarn outdated
yarn outdated v0.16.1
Package Current Wanted Latest
lodash  3.10.1  3.10.1 4.16.4

$ yarn upgrade lodash --dev
<snip>
success Saved 1 new dependency
└─ lodash@4.16.4

$ grep lodash package.json -C 1
  "devDependencies": {
    "lodash": "3"
  },
--
--
  "dependencies": {
    "lodash": "^4.16.4"
  }
```

`yarn upgrade` should update the version string of the entry in the `devDependencies` hash.

It would be nice if this could be done without `--dev` using the current location of the dependency (upgrading a dev dep would modify `devDependencies`, upgrading a normal dep would modify `dependencies`).

``` bash
$ node --version
v6.9.1
$ yarn --version
0.16.1
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.12
BuildVersion:   16A323
```

wyze:
**Summary**

This fixes #1458. Now the logic will properly update the `package.json` whether the package was a dependency or devDependency.

**Test plan**

```js
{
  "name": "yarn-example",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "is-negative-zero": "^0.1.0",
    "left-pad": "^1.0.0"
  },
  "dependencies": {
    "max-safe-integer": "^1.0.0"
  }
}
```

```sh
$ yarn upgrade max-safe-integer left-pad
[2/4] 🚚  Fetching packages...
success Saved lockfile.
success Saved 2 new dependencies.
├─ left-pad@1.1.3
└─ max-safe-integer@1.0.1
✨  Done in 0.49s.
```

##### Before

```diff
     "left-pad": "^1.0.0"
   },
   "dependencies": {
-    "max-safe-integer": "^1.0.0"
+    "left-pad": "^1.1.3",
+    "max-safe-integer": "^1.0.1"
   }
 }
```

##### After

```diff
   "license": "MIT",
   "devDependencies": {
     "is-negative-zero": "^0.1.0",
-    "left-pad": "^1.0.0"
+    "left-pad": "^1.1.3"
   },
   "dependencies": {
-    "max-safe-integer": "^1.0.0"
+    "max-safe-integer": "^1.0.1"
   }
 }
```

*Note: I will add tests to this in the afternoon.*

### DIFF
--- a/src/cli/commands/add.js
+++ b/src/cli/commands/add.js
@@ -120,6 +120,22 @@ export class Add extends Install {
 
       // build up list of objects to put ourselves into from the cli args
       const targetKeys: Array<string> = [];
+      if (this.config.commandName === 'upgrade') {
+        // use prototype to avoid passing args as `excludePatterns` parameter
+        await Install.prototype.fetchRequestFromCwd.call(this);
+
+        const keys = Object.keys(this.rootPatternsToOrigin);
+        const depType = keys.reduce((acc, prev) => {
+          // lookup the package to determine dependency type
+          if (prev.indexOf(`${pkg.name}@`) === 0) {
+            return this.rootPatternsToOrigin[prev];
+          }
+
+          return acc;
+        }, null);
+
+        targetKeys.push(depType);
+      }
       if (dev) {
         targetKeys.push('devDependencies');
       }


wyze:
I also added a few tests under `commands/add` since there were changes that affect the logic behind flags used with `yarn add`.

src/cli/commands/add.js
```diff
@@ -89,7 +97,11 @@ export class Add extends Install {
    */
 
   async savePackages(): Promise<void> {
-    const {dev, exact, tilde, optional, peer} = this.flags;
+    const {exact, tilde} = this.flags;
+
+    // fill rootPatternsToOrigin without `excludePatterns`
+    await Install.prototype.fetchRequestFromCwd.call(this);
```
kittens: why `Install.prototype` instead of just `this.fetchRequestFromCwd`?

wyze: Tried to explain it in the comment, but `this.fetchRequestFromCwd` passes the arguments along, so when it gets to [cli/commands/install.js#L197](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L197) the packages in the arguments never end up with the map of `{ package: dependencyType }` which is what we want. They would just be excluded.

th3fallen:
Has this been tagged yet?

gaui:
v0.20.3 and still doesn't update package.json - only yarn.lock file. You should reopen #1458

screendriver:
On my machine the `package.json` gets updated if I do a `yarn upgrade <package-name>` (Yarn version `v0.20.3`)

gaui:
@screendriver Yes but if you omit the `<package-name>` to upgrade all packages, it only updates yarn.lock file.

screendriver:
Ah I see. Sorry, I've overlooked that :confused:

jadengore:
Has this been logged as a new issue?

jaruesink:
I've gotten to this point also.... package.json needs updating! especially useful when @angular has broken up all of it's pieces.

igbopie:
This does not work for me for yarn workspaces. Can someone take a look?

Karl-EdwardFPJeanMehu:
I am also experiencing this issue. `yarn upgrade` ignores **devDependencies** and simply dupes into **dependencies**

Issue #1014, opened 2016-10-13
Some dependencies don't get installed if they're shared between a regular and a devDependency


bouk:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

Take the following package.json:

```
{
  "dependencies": {
    "babel-core": "6.17.0"
  },
  "devDependencies": {
    "eslint": "3.3.1"
  }
}
```

Running `yarn install` leads to the following yarn.lock https://gist.github.com/bouk/b96cf30321d696724ef926367c5af879

Yarn LS tells me the following https://gist.github.com/bouk/6514ee7102c7c386cf29f351747c8e54

If I remove node_modules and then run `yarn install --production`, I will end up with a situation where `escape-string-regexp` is missing, even though it's a dependency of babel-core, through babel-code-frame->chalk->escape-string-regexp

**What is the expected behavior?**

escape-string-regexp is installed

**Please mention your node.js, yarn and operating system version.**
Node v6.7.0
Yarn v0.15.1
macOS 10.12

davidowens:
This also seems to happen with a global npm install of yarn.

```
$ npm install -g gulp
$ npm install -g yarn
```

Then if I run `yarn` I get this error
```
$ yarn                                                                                                                                                                      /usr/local/lib/node_modules/yarn/bin/yarn.js:48
      throw err;
      ^

Error: ENOENT: no such file or directory, open '/usr/local/lib/node_modules/yarn/node_modules/chalk/node_modules/escape-string-regexp/index.js'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.readFileSync (fs.js:397:15)
    at Object.Module._extensions..js (module.js:415:20)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/usr/local/lib/node_modules/yarn/node_modules/chalk/index.js:2:26)
    at Module._compile (module.js:409:26)
```

KarolAdamiec:
I have the same issue. yarn --production is confused with dev dependencies.
Workaround for now is to install all dependencies... less than ideal.

vaikuntn:
I am also getting same error while executing yarn command. 

```

$ yarn
/usr/local/lib/node_modules/yarn/bin/yarn.js:48
      throw err;
      ^

[...]

Issue #761, opened 2016-10-11
`yarn install --production` doesn't install correct dependencies


donovan-graham:
When running `yarn install --production` it does not install the required dependencies of `forever`.  This seem to be related to having`nodemon` in `devDependencies`. 

Error response:

``` console
> forever app.js
module.js:457
    throw err;
    ^
Error: Cannot find module 'minimatch'
```

I've created a test application here:
https://github.com/donovan-graham/yarn-example-app

``` sh
#  Steps to reproduce error
git clone https://github.com/donovan-graham/yarn-example-app.git
cd yarn-example-app
yarn install --production
npm start

#  temporary step to bypass error
rm -rf node_modules
yarn remove nodemon
yarn install --production
npm start
```

sendilkumarn:
I guess this is because of  nodemon is having latest version of minimatch.

The linker function currently takes in both deps and dev deps into it. For argument production, this should be prevented.

Even on normal yarn install with out production argument. Only latest version is installed in the actual path. This has to checked too.

liorbrauer:
I'm getting a similar problem when running `yarn install -production` and then trying to run my build using `webpack`.

```
> NODE_ENV=production webpack -p --config webpack/production.config.js

module.js:457
    throw err;
    ^

Error: Cannot find module 'graceful-fs'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
```

and if I recall correctly, previous attempts showed similar errors with another package (not only `graceful-fs`)

comfroels:
I'm getting very similar as well... `yarn install` works just fine. but with the `--production` flag I get this:
```
> yarn install --production

yarn install v0.15.1
error npm-shrinkwrap.json found. This will not be updated or respected. See [TODO] for more information.
[1/4] Resolving packages...
[2/4] Fetching packages...
warning lodash@2.4.2: The engine "rhino" appears to be invalid.
warning fsevents@1.0.14: The platform "win32" is incompatible with this module.
info "fsevents@1.0.14" is an optional dependency and failed compatibility check. Excluding it from installation.
warning benchmark@1.0.0: The engine "rhino" appears to be invalid.
warning lodash@1.0.2: The engine "rhino" appears to be invalid.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
[1/1] ⠐ node-sass:     at Module.require (module.js:367:17)
[...]

PR #1626, opened 2016-11-03
fix-761: do not ignore devDeps that are included as in prod pkg (#761)

verth:
**Summary**

When installing with `production` flag the `devDeps` also follow the flow, but are set to `ignored`. Chances are that  a `production` dependency can be smartly resolved and shared with a devDep. If this is the case make sure this package also gets set to `USED` when being included, so it won't be ignored later.

This solves at least: 
* https://github.com/yarnpkg/yarn/issues/761
* https://github.com/yarnpkg/yarn/issues/1014

**Test plan**

`npm run test` should not give errors and contain a new use-case for these cases.

Extracted & tested use-case from the issue:

```
$ git clone https://github.com/donovan-graham/yarn-example-app.git
$ cd yarn-example-app
```

Should resolve and give the correct dependencies:

```
$ yarn install --production
$ yarn check --production
$ yarn check --production --integrity
$ npm run start
```


verth:
Closing in favour for possible better solutions. More (other) issues are related as-well to inheriting and propagating visibility.

Logic was: hoisting is responsible for hoisting & ignoring deps, when hoisting a devdep to proddep be smart and add it anyway, don't touch install process.

PR #1629, opened 2016-11-03
Attempt to fix failing Circle CI builds

wyze:
**Summary**

Bumping the timeout seemed to do the trick for me here: https://circleci.com/gh/yarnpkg/yarn/1906

**Test plan**

Circle CI builds are green again.

### DIFF
--- a/__tests__/util/git.js
+++ b/__tests__/util/git.js
@@ -2,6 +2,8 @@
 
 import Git from '../../src/util/git.js';
 
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
+
 test('cleanUrl', () => {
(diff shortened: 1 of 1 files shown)


Daniel15:
Hm... We probably shouldn't be running tests that can take up to 90 seconds as part of every single commit or PR, as it makes the feedback loop slower. This would be better suited for a nightly test run or something like that. I'm going to be running some nightly end-to-end tests (see https://github.com/yarnpkg/yarn/tree/master/end_to_end_tests), maybe we could also categorise some tests as "slow" tests and not run them as part of CI.

wyze:
Granted this was a bandaid fix, but this timeout of 90 seconds is also defined in a lot of other test files.

PR #1638, opened 2016-11-03
Use ~/Library/Caches as default cache location on OSX

kittens:


bestander:
So does npm.

kittens:
@bestander 

> next, Windows and Linux users will ask for their cache folder locations to be satisfied

This seems fine. We already use `%APPDATA%` on Windows (as does npm). It makes sense to be a good citizen on the platforms we support and put our files in the correct places so tooling works correctly.

Daniel15:
>  people may depend on ~/.yarn-cache in their routines and integrations

Cache directories are an internal implementation detail, I don't think anyone should rely on them (or they should use a Yarn API to get the path)

I agree that we should use the right path on Linux though - `/var/cache/` for root, `~/.cache/` for regular users.

hpurmann:
@Daniel15 It's pretty helpful to know where the cache lives in case you want to mount a Docker volume containing it.

```Dockerfile
VOLUME /root/.yarn-cache
```

IMHO this change should be mentioned in the release notes.

Daniel15:
Yeah @bestander was going to add some release notes for 0.17.x

PR #1654, opened 2016-11-03
Add option to change the prefix of the global bin folder - fixes #630

### LINKED ISSUE #630: Possible to set the global install path
This is a feature (or perhaps documentation request).

When I run the following:

`yarn global add create-react-app`

Yarn will try to symlink into `/usr/bin` which requires `sudo`. With NPM, I've set my global packages to install into a user directory that is then added to my `$PATH`. 

Is it possible to set a global directory via config somehow?

I've tried the following to no avail:

`yarn add create-react-app -g --global-folder ~/bin` (installs into `~/node_modules`)
`yarn global add create-react-app --global-folder ~/bin` (still goes for `/usr/bin`)

This is all running on Debian 8.5

kittens:


Daniel15:
Looks like this introduced some Flow errors... @kittens could you please take a look?

wyze:
@Daniel15 I have #1668 open to fix the flow errors and tests. Had to finish up the newly added test from #1210 as it was erroring out on Windows.

Daniel15:
Thank you! 😄

appsforartists:
`PREFIX` is a little ambiguous for an environment variable; how about `YARN_PREFIX`?

Issue #739, opened 2016-10-11
Cannot fetch newman from yarn install


dpkshrma:
**Do you want to request a _feature_ or report a _bug_?**

bug

**What is the current behavior?**

After resolution process is complete, yarn errors during fetch:

```
[2/4] Fetching packages...
error ENOENT: no such file or directory, scandir '/home/deepak/.yarn-cache/npm-postman-runtime-2.5.4/man'
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Mention newman as devDependency in package.json
   
   ```
     "devDependencies": {
        "newman": "^3.1.2",
         ...
   ```
2. `yarn install`

**What is the expected behavior?**

Yarn should complete all 4 processes.

**Please mention your node.js, yarn and operating system version.**

yarn: v0.15.1
node: v6.3.1
OS: ubuntu 14.04

dpkshrma:
On a sidenote, I haven't seen any other dependency causing such issue during install so far. But since, the npm modules are being fetched from [registry.yarnpkg.com](http://registry.yarnpkg.com) , so, I suspect, some modules might not have been indexed properly (just guessing here) which indicates there might be more packages which might be facing similar issue.

cnadeau:
I have the exact same issue with 

`yarn add jshint-reporter-bamboo`

houjunchen:
I have the same issue with `loopback-connector-sendgrid`:

```
error ENOENT: no such file or directory, scandir '/Users/abnerchen/.yarn-cache/npm-loopback-connector-sendgrid-2.0.6/bin'
    at Error (native)
```

jaydeep17:
+1 for `loopback-connector-sendgrid `
Facing the same issue 😕

jaydeep17:
@houjunchen I've found a workaround this problem, may be you can use it too.

We are installing a fork of `loopback-connector-sendgrid` directly from github, and according to the `package.json` [file](https://github.com/Cellarise/loopback-connector-sendgrid/blob/master/package.json#L77) in the github repo, it is expecting a `bin` directory, which yarn checks and fails to find it, thus it's complaining.

What we did was, in our fork, we kept only the directories that are actually present in the repo.
It worked!! 🎉

dormond:
So, NPM must be ignoring bin and man directories that are specified that don't exist.  Should Yarn do the same to catch the error when handling the bin and man directories?  

Logging a warning might be nice, since the package might not work correctly if a bin directory was really expected to exist.

[...]

PR #1663, opened 2016-11-03
Warn when man or bin directory not found but don't crash

dormond:
**Summary**
Fixes situation where the bin or man directory is specified in the package.json but does not really exist.  This change is to address issues like #739

**Test plan**
yarn add newman@3.2.0 succeeds to install.  Previously when installing dependency postman-collection@0.5.5 yarn would crash with an ENOENT trying to read the bin directory that doesn't exist.

### DIFF
--- a/src/util/normalize-manifest/fix.js
+++ b/src/util/normalize-manifest/fix.js
@@ -20,10 +20,13 @@ type Dict<T> = {
   [key: string]: T;
 };
 
+type WarnFunction = (msg: string) => void;
+
 export default async function (
   info: Dict<mixed>,
   moduleLoc: string,
   reporter: Reporter,
+  warn: WarnFunction,
   looseSemver: boolean,
 ): Promise<void> {
   const files = await fs.readdir(moduleLoc);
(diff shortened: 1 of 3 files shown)


src/util/normalize-manifest/fix.js
```diff
@@ -185,11 +188,17 @@ export default async function (
     if (!info.bin && binDir && typeof binDir === 'string') {
       const bin = info.bin = {};
 
-      for (const scriptName of await fs.readdir(path.join(moduleLoc, binDir))) {
-        if (scriptName[0] === '.') {
-          continue;
+      try {
```
kittens: Could we do `await fs.exists()` here instead of the `try-catch`? Seems much nicer.

src/util/normalize-manifest/fix.js
```diff
@@ -198,9 +207,15 @@ export default async function (
     if  (!info.man && typeof manDir === 'string') {
       const man = info.man = [];
 
-      for (const filename of await fs.readdir(path.join(moduleLoc, manDir))) {
-        if (/^(.*?)\.[0-9]$/.test(filename)) {
-          man.push(path.join('.', manDir, filename));
+      try {
```
kittens: Here too please.

src/util/normalize-manifest/fix.js
dormond: fs.exists is deprecated.  I'm not sure how long it has been deprecated and the other options of fs.stat() or fs.access() seem like they would have the same problem of having the error thrown.

dormond:
Changes per the review are done.  Thanks for the input.

PR #1666, opened 2016-11-03
Fix request authentication for scoped private registry tarballs

KidkArolis:
**Summary**

Yarn correctly sends authorization headers when resolving metadata, but
when it's time to download the tarballs, it calls registry.request
with a full tarball url. This causes the getRegistry() function to
incorrectly return the DEFAULT_REGISTRY url which means the authorization
headers are no longer sent. The response in this case is a JSON error
object instead of a tarball file. Untaring fails with "invalid tar file"

**Test plan**

Before this change, I get "invalid tar file" during the "Fetching packages..." stage.
After this change, the packages download sucesfully.

Follow instructions here to reproduce locally: https://github.com/KidkArolis/yarn-scopes-issue.

Any pointers on how to add a test for this are welcome.

### DIFF
--- a/__tests__/__mocks__/request.js
+++ b/__tests__/__mocks__/request.js
@@ -15,6 +15,9 @@ const fs = require('fs');
 
 const CACHE_DIR = path.join(__dirname, '..', 'fixtures', 'request-cache');
 
+const authedPackages = require('../fixtures/authed-packages/config');
+const UNAUTHORIZED_LOC = path.join(__dirname, '..', 'fixtures', 'authed-packages', 'unauthorized.bin');
+
 function getRequestAlias(params: Object): string {
   const parts = url.parse(params.path);
   const pathname = cleanAlias(parts.pathname);
@@ -30,6 +33,22 @@ function cleanAlias(str: string): string {
     .replace(/-+/g, '-') // replace multiple dashes with one
     .replace(/^-+/, '') // remove leading dashes
     .replace(/-+$/, ''); // remove trailing dashes
+}
+
+function getAuthInfo(path: string): ?{token: string, registry: string} {
+  path = decodeURIComponent(path);
+  const pkgs = Object.keys(authedPackages);
+  const match = pkgs.find((pkgName) => path.endsWith(pkgName) || path.includes(pkgName + '/'));
+  return match ? authedPackages[match] : null;
+}
+
+function requestFile(path: string, httpModule: Object, options: Object, callback?: ?Function): ClientRequest {
+  options.agent = null;
+  options.socketPath = null;
+  options.createConnection = (): ReadStream => {
+    return fs.createReadStream(path);
+  };
+  return httpModule.request(options, callback);
 }
 
 class Request extends RealRequest {
@@ -55,14 +74,21 @@ const httpMock = {
     // TODO better way to do this
     const httpModule = options.uri.href.startsWith('https:') ? https : http;
 
+    const {token, registry} = getAuthInfo(options.uri.path) || {};
+    if (token && registry) {
+      if (!options.uri.href.startsWith(registry)) {
+        // request is being made to the wrong registry, token gets disclosed
+        throw new Error(`Wrong registry, expected ${registry} to be used.`);
+      }
+      if ((options.headers.authorization || '').split('Bearer ')[1] !== token) {
+        // token missing
+        return requestFile(UNAUTHORIZED_LOC, httpModule, options, callback);
+      }
+    }
+
     if (allowCache && fs.existsSync(loc)) {
       // cached
-      options.agent = null;
-      options.socketPath = null;
-      options.createConnection = (): ReadStream => {
-        return fs.createReadStream(loc);
-      };
-      return httpModule.request(options, callback);
+      return requestFile(loc, httpModule, options, callback);
     } else {
       // not cached
       const req = httpModule.request(options, callback);
--- a/__tests__/__mocks__/request.js
+++ b/__tests__/__mocks__/request.js
@@ -15,8 +15,7 @@ const fs = require('fs');
 
 const CACHE_DIR = path.join(__dirname, '..', 'fixtures', 'request-cache');
 
-const authedPackages = require('../fixtures/authed-packages/config');
-const UNAUTHORIZED_LOC = path.join(__dirname, '..', 'fixtures', 'authed-packages', 'unauthorized.bin');
+let authedRequests = [];
 
 function getRequestAlias(params: Object): string {
   const parts = url.parse(params.path);
@@ -35,22 +34,6 @@ function cleanAlias(str: string): string {
     .replace(/-+$/, ''); // remove trailing dashes
 }
 
-function getAuthInfo(path: string): ?{token: string, registry: string} {
-  path = decodeURIComponent(path);
-  const pkgs = Object.keys(authedPackages);
-  const match = pkgs.find((pkgName) => path.endsWith(pkgName) || path.includes(pkgName + '/'));
-  return match ? authedPackages[match] : null;
-}
-
-function requestFile(path: string, httpModule: Object, options: Object, callback?: ?Function): ClientRequest {
-  options.agent = null;
-  options.socketPath = null;
-  options.createConnection = (): ReadStream => {
-    return fs.createReadStream(path);
-  };
-  return httpModule.request(options, callback);
-}
-
 class Request extends RealRequest {
   init(params: Object) {
     RealRequest.prototype.init.call(this, params);
@@ -64,6 +47,9 @@ module.exports = function(params: Object): Request {
 
 module.exports.Request = Request;
 
+module.exports.__resetAuthedRequests = (): void => { authedRequests = []; };
+module.exports.__getAuthedRequests = (): Array<Object> => authedRequests;
+
 const httpMock = {
   request(options: Object, callback?: ?Function): ClientRequest {
     const alias = getRequestAlias(options);
@@ -74,21 +60,22 @@ const httpMock = {
     // TODO better way to do this
     const httpModule = options.uri.href.startsWith('https:') ? https : http;
 
-    const {token, registry} = getAuthInfo(options.uri.path) || {};
-    if (token && registry) {
-      if (!options.uri.href.startsWith(registry)) {
-        // request is being made to the wrong registry, token gets disclosed
-        throw new Error(`Wrong registry, expected ${registry} to be used.`);
-      }
-      if ((options.headers.authorization || '').split('Bearer ')[1] !== token) {
-        // token missing
-        return requestFile(UNAUTHORIZED_LOC, httpModule, options, callback);
-      }
+    // expose authorized requests to the tests for assertion
+    if (options.headers.authorization) {
+      authedRequests.push({
+        url: options.uri.href,
+        headers: options.headers,
+      });
     }
 
     if (allowCache && fs.existsSync(loc)) {
       // cached
-      return requestFile(loc, httpModule, options, callback);
+      options.agent = null;
+      options.socketPath = null;
+      options.createConnection = (): ReadStream => {
+        return fs.createReadStream(loc);
+      };
+      return httpModule.request(options, callback);
     } else {
       // not cached
       const req = httpModule.request(options, callback);
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -775,3 +775,12 @@ test.concurrent('install uses OS line endings when lockfile doesn\'t exist', asy
       assert(lockfile.indexOf(os.EOL) >= 0);
     });
 });
+
+test.concurrent('install from authed private registry', (): Promise<void> => {
+  return runInstall({noLockfile: true}, 'install-from-authed-private-registry', async (config) => {
+    assert.equal(
+      (await fs.readFile(path.join(config.cwd, 'node_modules', '@types', 'lodash', 'index.d.ts'))).split('\n')[0],
+      '// Type definitions for Lo-Dash 4.14',
+    );
+  });
+});
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -17,8 +17,13 @@ import os from 'os';
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 
 const path = require('path');
+const request = require('request');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'install');
+
+beforeEach(request.__resetAuthedRequests);
+// $FlowFixMe
+afterEach(request.__resetAuthedRequests);
 
 test.concurrent('properly find and save build artifacts', async () => {
   await runInstall({}, 'artifacts-finds-and-saves', async (config): Promise<void> => {
@@ -776,24 +781,16 @@ test.concurrent('install uses OS line endings when lockfile doesn\'t exist', asy
     });
 });
 
-describe('authed registries', (): void => {
-  const request = require('request');
-
-  beforeEach(request.__resetAuthedRequests);
-  // flow doesn't like afterEach
-  // afterEach(request.__resetAuthedRequests);
-
-  test.concurrent('install a scoped module from authed private registry', (): Promise<void> => {
-    return runInstall({noLockfile: true}, 'install-from-authed-private-registry', async (config) => {
-      const authedRequests = request.__getAuthedRequests();
-      assert.equal(authedRequests[0].url, 'https://registry.yarnpkg.com/@types%2flodash');
-      assert.equal(authedRequests[0].headers.authorization, 'Bearer abc123');
-      assert.equal(authedRequests[1].url, 'https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.37.tgz');
-      assert.equal(authedRequests[1].headers.authorization, 'Bearer abc123');
-      assert.equal(
-        (await fs.readFile(path.join(config.cwd, 'node_modules', '@types', 'lodash', 'index.d.ts'))).split('\n')[0],
-        '// Type definitions for Lo-Dash 4.14',
-      );
-    });
+test('install a scoped module from authed private registry', (): Promise<void> => {
+  return runInstall({noLockfile: true}, 'install-from-authed-private-registry', async (config) => {
+    const authedRequests = request.__getAuthedRequests();
+    assert.equal(authedRequests[0].url, 'https://registry.yarnpkg.com/@types%2flodash');
+    assert.equal(authedRequests[0].headers.authorization, 'Bearer abc123');
+    assert.equal(authedRequests[1].url, 'https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.37.tgz');
+    assert.equal(authedRequests[1].headers.authorization, 'Bearer abc123');
+    assert.equal(
+      (await fs.readFile(path.join(config.cwd, 'node_modules', '@types', 'lodash', 'index.d.ts'))).split('\n')[0],
+      '// Type definitions for Lo-Dash 4.14',
+    );
   });
 });
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -776,11 +776,24 @@ test.concurrent('install uses OS line endings when lockfile doesn\'t exist', asy
     });
 });
 
-test.concurrent('install from authed private registry', (): Promise<void> => {
-  return runInstall({noLockfile: true}, 'install-from-authed-private-registry', async (config) => {
-    assert.equal(
-      (await fs.readFile(path.join(config.cwd, 'node_modules', '@types', 'lodash', 'index.d.ts'))).split('\n')[0],
-      '// Type definitions for Lo-Dash 4.14',
-    );
+describe('authed registries', (): void => {
+  const request = require('request');
+
+  beforeEach(request.__resetAuthedRequests);
+  // flow doesn't like afterEach
+  // afterEach(request.__resetAuthedRequests);
+
+  test.concurrent('install a scoped module from authed private registry', (): Promise<void> => {
+    return runInstall({noLockfile: true}, 'install-from-authed-private-registry', async (config) => {
+      const authedRequests = request.__getAuthedRequests();
+      assert.equal(authedRequests[0].url, 'https://registry.yarnpkg.com/@types%2flodash');
+      assert.equal(authedRequests[0].headers.authorization, 'Bearer abc123');
+      assert.equal(authedRequests[1].url, 'https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.37.tgz');
+      assert.equal(authedRequests[1].headers.authorization, 'Bearer abc123');
+      assert.equal(
+        (await fs.readFile(path.join(config.cwd, 'node_modules', '@types', 'lodash', 'index.d.ts'))).split('\n')[0],
+        '// Type definitions for Lo-Dash 4.14',
+      );
+    });
   });
 });
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -145,8 +145,11 @@ export default class NpmRegistry extends Registry {
   getRegistry(packageName: string): string {
     // Try extracting registry from the url, then scoped registry, and default registry
     if (packageName.match(/^https?:/)) {
-      const parts = url.parse(packageName);
-      return parts.protocol + '//' + parts.host + '/';
+      const availableRegistries = this.getAvailableRegistries();
+      const registry = availableRegistries.find((registry) => packageName.startsWith(registry));
+      if (registry) {
+        return registry;
+      }
     }
 
     for (const scope of [this.getScope(packageName), '']) {
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -143,7 +143,12 @@ export default class NpmRegistry extends Registry {
   }
 
   getRegistry(packageName: string): string {
-    // Try scoped registry, and default registry
+    // Try extracting registry from the url, then scoped registry, and default registry
+    if (packageName.match(/^https?:/)) {
+      const parts = url.parse(packageName);
+      return parts.protocol + '//' + parts.host + '/';
+    }
+
     for (const scope of [this.getScope(packageName), '']) {
       const registry = this.getScopedOption(scope, 'registry')
                     || this.registries.yarn.getScopedOption(scope, 'registry');
--- a/__tests__/fixtures/authed-packages/config.js
+++ b/__tests__/fixtures/authed-packages/config.js
@@ -1,6 +1,0 @@
-module.exports = {
-  '@types/lodash': {
-    registry: 'https://registry.yarnpkg.com/',
-    token: 'abc123'
-  }
-};
--- a/__tests__/fixtures/authed-packages/config.js
+++ b/__tests__/fixtures/authed-packages/config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  '@types/lodash': {
+    registry: 'https://registry.yarnpkg.com/',
+    token: 'abc123'
+  }
+};
--- a/__tests__/fixtures/authed-packages/unauthorized.bin
+++ b/__tests__/fixtures/authed-packages/unauthorized.bin
@@ -1,0 +1,8 @@
+HTTP/1.1 401 OK
+Date: Fri, 14 Oct 2016 21:05:22 GMT
+Content-Type: application/json
+Transfer-Encoding: chunked
+Connection: keep-alive
+
+49
+{"error":"unauthorized","reason":"token missing"}
--- a/__tests__/fixtures/authed-packages/unauthorized.bin
+++ b/__tests__/fixtures/authed-packages/unauthorized.bin
@@ -1,8 +1,0 @@
-HTTP/1.1 401 OK
-Date: Fri, 14 Oct 2016 21:05:22 GMT
-Content-Type: application/json
-Transfer-Encoding: chunked
-Connection: keep-alive
-
-49
-{"error":"unauthorized","reason":"token missing"}
--- a/__tests__/fixtures/install/install-from-authed-private-registry/.npmrc
+++ b/__tests__/fixtures/install/install-from-authed-private-registry/.npmrc
@@ -1,0 +1,2 @@
+//registry.yarnpkg.com/:_authToken=abc123
+@types:registry=https://registry.yarnpkg.com/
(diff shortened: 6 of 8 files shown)


KidkArolis:
Ah, this is actually a dupe of https://github.com/yarnpkg/yarn/pull/1561.

src/registries/npm-registry.js
```diff
@@ -139,7 +139,12 @@ export default class NpmRegistry extends Registry {
   }
 
   getRegistry(packageName: string): string {
-    // Try scoped registry, and default registry
+    // Try extracting registry from the url, then scoped registry, and default registry
+    if (packageName.match(/^https?:/)) {
+      const parts = url.parse(packageName);
+      return parts.protocol + '//' + parts.host + '/';
```
dfreeman: A registry may not be just a host — it can also include a path e.g. for registries that are part of a larger artifact hosting application, like Nexus or Artifactory.

KidkArolis:
Updated the PR with an extra commit that does not assume anything about the registry URL structure (ah.. except that it starts with `https?://`). Instead, this reads the available registries from config.

jedi4ever:
@KidkArolis can confirm that the patch fixed my "invalid tar file" using a private registry with authentication. thanks!

jedi4ever:
ha it seems this patch solved `yarn install` but I still get the `invalid tar file` while doing an `yard add` of a package (even a public package...)

jedi4ever:
and on a side note - the dependencies in the generated `yarn.lock` file contain the "http://" version, while my registry is running on "https://" - (not sure if this related)

KidkArolis:
I've fixed the linting issues.

I'm currently working on a test for this PR. From what I've understood, there are no existing tests that deal with validating authentication headers.

My first attempt was to add a test to `__tests__/commands/install` that would spin up a tiny `http` server mimicking a private authed registry. I'd then have a fixture with `.npmrc` configured to point to some private scope to this local registry. The downside of this is kind of having to run a real server with a real socket during the test, it doesn't seem to fit in with the rest of the tests. And even after all that, the `request` mock with cache the request and in the future runs would stop validating auth headers.

This is what I'll try instead. I'll modify the `request` mock to check `fixtures/authed-packages/config.js` config file, which would contain pairs of packages and required tokens, e.g. `{'@segment/protocol': 'abc123'}`. The mocked request would now always respond with an error if the request URL matches one of the authed packages, but doesn't have a correct authentication header. This means that any of the tests could choose to test installation of private authenticated packages if it's required for any future edge cases. The only tiny caveat here is that the nature of the bug this PR is fixing is that when the registry URL is not correctly extracted, it fallbacks to the default registry url, which means this fix won't get exercised, but I'll solve that when I get there.

rexxars:
If you change your mind and decide that a real HTTP server is a more reliable way to test this, I made a [mock private registry](https://github.com/rexxars/mock-private-registry) module for exactly these kinds of cases.

bestander:
An http server is a bit too much for a test here, let's mock it.
Would this PR cover the same use cases as #1561?

KidkArolis:
Yes, this PR fixes the same issue, but might be in a slightly more generic way in that my PR handles registries that use a path as part of the registry URL. Thats because instead of regexing a URL I compare all known registries against the request URL to see if there's a match.

KidkArolis:
OK. I've added a test now. If I remove my fix - the test fails, with the fix, the test passes.

As I was working on this fix, I've discovered the following things.

1. At first, I tried using `registry.npmjs.org` as my authenticated registry. I noticed that the metadata request is sent to `registry.npmjs.org/@types/lodash` with the token as expected, but when the tarball is requested, the `cleanRegistry` function in `npm-resolver` always rewrites the tarball URL to `registry.yarnpkg.com`. The reason I'm flagging this is, because at this point, yarn reads the token from the config from `@types:registry=https://registry.npmjs.org/` entry, but then sends this token to `registry.yarnpkg.com`. **I'm not sure this is desired**. This also meant that this test wasn't exercising the fix, because even when the registry wasn't extracted from the tarball url correctly, it defaulted to DEFAULT_REGISTRY.

2. I then switched to using `registry.yarnpkg.com` as the private registry. At first, I made a typo in `.npmrc` of the test and I noticed it was pulling my `npm` token from my `~/.npmrc`. Should tests consume the global `~/.npmrc`? That could lead to environmental issues for developers.

3. But more importantly, should yarn default to the npm token and disclose that to an arbitrary private registry when it doesn't find a token otherwise? As in, it saw `@types:registry=https://registry.yarnpkg.MY.TYPO`, it looked for a token for this registry, didn't find 
it, and then went to to use the `npm` token from my `~/.npmrc` and sent it to `https://registry.yarnpkg.MY.TYPO`.

What are your thoughts on these, shall I file as separate issues?

And as to this PR, I'm not 100% sure this is the best way to solve this problem, the whole auth business seems quite tricky and brittle. But this fix will enable using `yarn` with private npm registries, this is what is currently stopping me and my team from switching to yarn for good!

And hopefully, the ability to test private, scoped and authenticated packages with the updated `request` mock will help improve the auth implementation further.

jcarbo:
Any idea if this will land in a release soon?

Like @KidkArolis, my company is eager to drop `npm` in favor of `yarn` but we use private/scoped NPM packages so we're blocked until this is resolved. Thanks!

KidkArolis:
To the maintainers - despite long discussions in this thread - this PR is 100% ready, has a test and shouldn't break any existing functionality!

SEAPUNK:
I installed yarn manually from your branch, but it doesn't seem to be working for me:

```
$ cat ~/.npmrc
registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=the-long-token

$ yarn
yarn install v0.17.1-0
info No lockfile found.
[1/4] Resolving packages...
warning mini-copromise@3.0.2: Use the `minico` package instead: It\'s the same thing!
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/home/ivan/code/xbpf/packages/manager-frontend/yarn-error.log".

# Trace
Trace: 
  Error: https://registry.yarnpkg.com/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file
      at Extract.Parse._startEntry (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:149:13)
      at Extract.Parse._process (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:131:12)
      at BlockStream.<anonymous> (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:47:8)
      at emitOne (events.js:96:13)
      at BlockStream.emit (events.js:188:7)
      at BlockStream._emitChunk (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:145:10)
      at BlockStream.flush (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:70:8)
      at BlockStream.end (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:66:8)
      at Extract.Parse.end (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:86:23)
      at UnpackStream.onend (_stream_readable.js:511:10)


```

SEAPUNK:
I just now changed `.yarnrc` to use the npm registry directly, to no avail:

```
$ cat ~/.yarnrc
registry "https://registry.npmjs.org/"

$ yarn
yarn install v0.17.3-fix
info No lockfile found.
[1/4] Resolving packages...
warning mini-copromise@3.0.2: Use the `minico` package instead: It\'s the same thing!
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.npmjs.org/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/home/ivan/code/xbpf/packages/manager-frontend/yarn-error.log".

# Trace

Trace: 
  Error: https://registry.npmjs.org/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file
      at Extract.Parse._startEntry (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:149:13)
      at Extract.Parse._process (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:131:12)
      at BlockStream.<anonymous> (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:47:8)
      at emitOne (events.js:96:13)
      at BlockStream.emit (events.js:188:7)
      at BlockStream._emitChunk (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:145:10)
      at BlockStream.flush (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:70:8)
      at BlockStream.end (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:66:8)
      at Extract.Parse.end (/home/ivan/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/node_modules/tar/lib/parse.js:86:23)
      at UnpackStream.onend (_stream_readable.js:511:10)

```

thomaschaaf:
@KidkArolis there is a conflict with my commit :( but it works for me :)

KidkArolis:
@SEAPUNK looking at your `.npmrc`, can you try changing that to a scoped config:

```
@xbpf:registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=the-long-token
```

KidkArolis:
Rebased.

thomaschaaf:
@bestander let's get this in 0.17.4 too :)

SEAPUNK:
@KidkArolis I'll try it out, give me a few minutes.

Either way, shouldn't it work with my current npmrc? That's the .npmrc file npm writes when I log in via `npm login`, and it fetches all packages just fine.

KidkArolis:
@SEAPUNK Well, I tried logging in just now with `npm login --registry=https://registry.npmjs.org --scope=@qubit` and it creates this:

```
@qubit:registry=https://registry.npmjs.org/
//registry.npmjs.org/:_authToken=zzz
```

I would have thought your config would only work if `always-auth` is also set to true. Hm..

SEAPUNK:
Yeah, I don't do the two flags with `registry` and `scope` when I run `npm login`. I just do plain `npm login`, input my credentials, and that's what npmrc is generated for me.

This is what my config looks like with the unedited `.npmrc`:

```
$ npm config list -l
; cli configs
long = true
user-agent = "npm/3.10.8 node/v6.9.1 linux x64"

; userconfig /home/ivan/.npmrc
registry = "https://registry.npmjs.com/"

; default values
access = null
also = null
always-auth = false
bin-links = true
browser = null
ca = null
cache = "/home/ivan/.npm"
cache-lock-retries = 10
cache-lock-stale = 60000
cache-lock-wait = 10000
cache-max = null
cache-min = 10
cafile = undefined
cert = null
color = true
depth = null
description = true
dev = false
dry-run = false
editor = "nano"
engine-strict = false
fetch-retries = 2
fetch-retry-factor = 10
fetch-retry-maxtimeout = 60000
fetch-retry-mintimeout = 10000
force = false
git = "git"
git-tag-version = true
global = false
global-style = false
globalconfig = "/home/ivan/.nvm/versions/node/v6.9.1/etc/npmrc"
globalignorefile = "/home/ivan/.nvm/versions/node/v6.9.1/etc/npmignore"
group = 100
heading = "npm"
https-proxy = null
if-present = false
ignore-scripts = false
init-author-email = ""
init-author-name = ""
init-author-url = ""
init-license = "ISC"
init-module = "/home/ivan/.npm-init.js"
init-version = "1.0.0"
json = false
key = null
legacy-bundling = false
link = false
local-address = undefined
loglevel = "warn"
; long = false (overridden)
maxsockets = 50
message = "%s"
node-version = "6.9.1"
npat = false
onload-script = null
only = null
optional = true
parseable = false
prefix = "/home/ivan/.nvm/versions/node/v6.9.1"
production = false
progress = true
proprietary-attribs = true
proxy = null
rebuild-bundle = true
; registry = "https://registry.npmjs.org/" (overridden)
rollback = true
save = false
save-bundle = false
save-dev = false
save-exact = false
save-optional = false
save-prefix = "^"
scope = ""
searchexclude = null
searchopts = ""
searchsort = "name"
shell = "/usr/bin/zsh"
shrinkwrap = true
sign-git-tag = false
strict-ssl = true
tag = "latest"
tag-version-prefix = "v"
tmp = "/tmp"
umask = 18
unicode = true
unsafe-perm = true
usage = false
user = 1000
; user-agent = "npm/{npm-version} node/{node-version} {platform} {arch}" (overridden)
userconfig = "/home/ivan/.npmrc"
version = false
versions = false
viewer = "man"
```

thomaschaaf:
@SEAPUNK I think this maybe a different problem from what @KidkArolis is trying to solve with this PR.

SEAPUNK:
@thomaschaaf Maybe; all I want is to be able to install private npm packages, which I have been unable to all the time. #839 and #1146 got there part of the way, but no cigar just yet.

I've been tracking many various yarn issues and PRs, and so far the point I've gotten to was the whole "invalid tar file" problem, which is what I've been hoping this PR would fix, but if it doesn't fix the problem I'm hoping would be fixed, is there an issue or PR to track for the issue I'm having?

SEAPUNK:
@KidkArolis Editing the npmrc file to what you suggested doesn't work either:

```
$ cat ~/.npmrc
@xbpf:registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=the-token

$ ~/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/bin/yarn
yarn install v0.18.0-0
info No lockfile found.
[1/4] Resolving packages...
warning mini-copromise@3.0.2: Use the `minico` package instead: It\'s the same thing!
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/home/ivan/code/xbpf/packages/manager-frontend/yarn-error.log".
```

KidkArolis:
OK, I kind of can reproduce your issue @SEAPUNK, I've created an issue to track this https://github.com/yarnpkg/yarn/issues/1895.

But I think I also have an idea of what's going on, it is indeed closely related to what this PR fixes, let me see if I can fix this edge case.

KidkArolis:
Are you sure you're using this PR when running that `yarn` command? How are you installing it?

I would probably, git clone my fork, checkout this branch, `npm run build` and run `~/this-fork/bin/yarn` in your project.

SEAPUNK:
This is what I do, step by step (nvm is installed, and using v6.9.1):

```
git clone ssh://git@github.com/KidkArolis/yarn
cd yarn
git checkout fix-custom-registry-auth
yarn # v0.17.2 is installed globally in /usr/bin
npm run build
npm install -g . # This installs to the nvm bin directory, which I use directly with the command:

cd ~/package-with-private-modules
~/.nvm/versions/node/v6.9.1/lib/node_modules/yarn/bin/yarn # This is v0.18.0-0 with the patch
```

I confirmed that this is the yarn with the patch by inspecting the `lib/` folder directory of the installed `yarn` package, looking for functions or keywords in files that this PR modifies.

I even removed the globally installed yarn in `/usr/bin`, guaranteeing that the custom yarn is the only installed yarn, and it still fails.

KidkArolis:
OK OK, let me investigate a bit, will get back to you in a bit.

KidkArolis:
Right, ok, I reproduced it all locally.

First, to recap, what this PR fixes is the ability to install scoped packages from custom non npm registries. The reason it wasn't working before this PR is a bug in the logic determining whether and what auth token to use for a particular request.

The issue you're running into @SEAPUNK is caused by this logic here: https://github.com/yarnpkg/yarn/blob/1180f0fb0e20281cf09303d18281ff9f6a0e2e6a/src/resolvers/registries/npm-resolver.js#L139-L140. For some reason, yarn replaces registry.npmjs.org to registry.yarnpkg.com in all requests. This means that `https://registry.npmjs.org/@xbpf/suppliers/-/suppliers-0.1.1.tgz` is rewritten to `https://registry.yarnpkg.com/@xbpf/suppliers/-/suppliers-0.1.1.tgz`. At this point, the code fails to find a token for yarnpkg and the request fails. If I comment that `cleanRegistry` function out, installing works. **This is something I need some feedback from the core team on**.

SEAPUNK:
@KidkArolis Shouldn't setting `registry "https://registry.npmjs.org/"` in `~/.yarnrc` ignore the replacement in cleanRegistry though? Because with that change, I still get the error, with the modified, scoped npmrc and all.

KidkArolis:
I tried working around this (for the short term) with `always-auth=true` or setting top level `_authToken=abc`, but none of that works because of this logic:

```js
if (requestUrl.startsWith(registry)) send token
```

Because `requestUrl` starts with `https://registry.yarnpkg.com/`, but registry is `registry.npmjs.org` because that's what the config says.

Now, I see several options here:

1. Remove the `cleanRegistry` url - I don't fully know what the intent there is.
2. Do the `cleanRegistry` replacement just in time for the request, after all the token and other logic has already been applied.
3. Don't assume that tarball urls always start with registry prefix, I've seen registries where the tarballs are on CDN on a different host (in that case you might still want to send a token, I donno..). Separate issue tbh.

KidkArolis:
@SEAPUNK Thing with that cleanRegistry function, if you follow up the code, you'll see that it _always_ replaces the urls:

```
this.config.getOption('registry') === YARN_REGISTRY

// where this.config.getOption is

  return this.registries.yarn.getOption(key);
}

// where yarn.getOption('registry') will return YARN_REGISTRY constant
```

Maybe that's the bug that needs to be fixed ^, but it's hard to understand without knowing the intent of this change: https://github.com/yarnpkg/yarn/pull/557.

SEAPUNK:
I edited the `cleanRegistry` code in both `lib/resolvers/registries/npm-resolver.js` and `lib-legacy/resolvers/registries/npm-resolver.js` to this:

```js
    cleanRegistry(url) {
      return url
    }
```

And it still fails:

```
$ ~/code/github/yarn/bin/yarn
yarn install v0.18.0-0
info No lockfile found.
[1/4] Resolving packages...
warning mini-copromise@3.0.2: Use the `minico` package instead: It\'s the same thing!
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.npmjs.org/@xbpf/suppliers/-/suppliers-0.1.1.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/home/ivan/code/xbpf/packages/manager-frontend/yarn-error.log".

$ cat ~/.yarnrc
cat: /home/ivan/.yarnrc: No such file or directory

$ cat ~/.npmrc
@xbpf:registry=https://registry.npmjs.com/
//registry.npmjs.com/:_authToken=the-long-token
```

I double-checked that the `lib/` and `lib-legacy/` folders both have the PR patches.

KidkArolis:
OK! So now, one more final edge case, try replacing `npmjs.com` in your config (both lines ) to `npmjs.org` (that's where the tarballs are).

SEAPUNK:
Okay, yeah that works. Whew! Thank you so much.

KidkArolis:
@SEAPUNK that's another edge case that we need to figure out how to solve..

SEAPUNK:
Also, note that changing `~/.npmrc` to remove the `@xbpf:` prefix seems to also work, even with the `~/.yarn*` folders, lockfile, and `node_modules` removed.

KidkArolis:
Starting to slowly realise a simpler way to do all of this. In npm-registry, we shouldn't have so much fiddly logic that touches URLs, compares url prefixes and all that..

1. If we always passed a package name (which yarn always knows) and also optionally a url for the tarball, that's already an improvement. We now reliably know what package we're installing without having to parse it out from the url.

2. Given a package, the logic should be very simple. Based on config, which registry should we use for this package (scope registry if exists, global registry if exists or default).

3. Given the registry, check if we have an auth token for it, if not, check if we have a global token.

4. If it's a scoped module, or always-auth is set - send the token.

5. Rename `getScopedOption` to something better, it's not just about scopes, it's any prefixed key (e.g. //registry.foo.com).

The logic is almost like that already, just could be a bit cleaner. I'll take a stab at this at some point. (This PR is still good to merge! :-").

Note to self: answer this question, if a token is configured for registryA.com, but the tarball is at registryB.com, should the token be sent to registryB? I'd say no. But what if registryB.com is registry.yarnpkg.com? Comes back to that cleanRegistry discussion.

__tests__/__mocks__/request.js
```diff
@@ -55,14 +74,21 @@ const httpMock = {
     // TODO better way to do this
     const httpModule = options.uri.href.startsWith('https:') ? https : http;
 
+    const {token, registry} = getAuthInfo(options.uri.path) || {};
```
bestander: PR looks good.
My only concern is that this `/fixtures/authed-packages/config` is used for all the unit tests that make requests.
Can you set/reset this fixture only for the test you need?

KidkArolis:
Hi, Id like to, but Im not sure how. The tests run in parallel and this
mock is global?

bestander:
Yeah, that is a challenge.
I think you can run jest.mock('request.js') for one test and it will use only the instance you provide.

bestander:
Let me know if there is a problem, I will help out

KidkArolis:
OK, I've changed the approach of the test to be less intrusive and more local. Instead of making checks and assertions within the request mock, I make the assertions about auth headers locally within the test.

This is actually a bit more robust than before, because previously the test could pass if installation succeeds even if the auth headers were not sent (because I'm installing public npm packages). Now, as part of the test, in addition to checking that installation succeeded we check that the right headers were actually sent with each request.

__tests__/commands/install.js
```diff
@@ -775,3 +775,25 @@ test.concurrent('install uses OS line endings when lockfile doesn\'t exist', asy
       assert(lockfile.indexOf(os.EOL) >= 0);
     });
 });
+
+describe('authed registries', (): void => {
+  const request = require('request');
+
+  beforeEach(request.__resetAuthedRequests);
+  // flow doesn't like afterEach
+  // afterEach(request.__resetAuthedRequests);
```
bestander: You probably can use $FlowFix annotation to force afterEach for flow.

__tests__/commands/install.js
```diff
@@ -775,3 +775,25 @@ test.concurrent('install uses OS line endings when lockfile doesn\'t exist', asy
       assert(lockfile.indexOf(os.EOL) >= 0);
     });
 });
+
+describe('authed registries', (): void => {
```
bestander: Any reason not to split this test into individual `describe`?
I think it should work fine along with the others.

bestander:
@KidkArolis, please ping when it is ready

KidkArolis:
Updated in @f0b30ef.

As a sidenote, unfortunately this PR doesn't in itself fix the `invalid tar` issue when installing private scoped packages from the official npm registry. This PR only fixes custom registries (my.registry.co). To fix the official npm, I've open PR #1896.

And as another sidenote, I do think some of this auth logic can be cleaned up further, but it helps to make small steps first, resolve edge cases and then refactor. I think.

chrisdhanaraj:
Hmm, shouldn't this be fixed in the latest nightly? I'm on a scoped package on a private registry (Sinopia) and I'm still getting the auth not being set with the tar issue :(

```
yarn install v0.18.0-20161129.0025
warning bluemix-design-system@1.0.0: No license field
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "http://console-npm.ng.bluemix.net/@console%2fbluemix-components/-/bluemix-components-6.10.3.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/Users/chrisdhanaraj/bluemix/design-system-website/yarn-error.log".
```

with my .npmrc looking like

```
@console:registry=https://console-npm.ng.bluemix.net/
//console-npm.ng.bluemix.net/:_authToken="mylongtoken"
```

owais:
I'm still facing this in 0.17.9 (which was supposed to include this patch). I'm using *nmpo*/*npme* on the server.

```
Trace:
  Error: http://npm.mycompany.com/@/@my-company/mypackage/_attachments/mypackage-3.7.2.tgz: invalid tar file
      at Extract.Parse._startEntry (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/tar/lib/parse.js:149:13)
      at Extract.Parse._process (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/tar/lib/parse.js:131:12)
      at BlockStream.<anonymous> (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/tar/lib/parse.js:47:8)
      at emitOne (events.js:96:13)
      at BlockStream.emit (events.js:188:7)
      at BlockStream._emitChunk (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:145:10)
      at BlockStream.flush (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:70:8)
      at BlockStream.end (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:66:8)
      at Extract.Parse.end (/usr/local/Cellar/yarn/0.17.9/libexec/lib/node_modules/yarn/node_modules/tar/lib/parse.js:86:23)
      at UnpackStream.onend (_stream_readable.js:511:10)
```

KidkArolis:
@chrisdhanaraj @owais this commit is not in 0.17.9. I'm not sure when it's getting released.

owais:
Got it. I was mislead by this: https://github.com/yarnpkg/yarn/issues/1859

Thanks for fixing this @KidkArolis. This is a major blocker for a lot of people to start using Yarn.

KidkArolis:
@ maintainers: This has been open for 28 days, merged 12 days ago - any chance it could get released so we get real world feedback on whether this fixes all issues with private repos or whether further fixes are needed?

bestander:
Sorry, @KidkArolis, I was spread thin testing the 0.17 branch more extensively and addressing some other issues.
I plan to do a 0.18 branch cut this weekend, does it work for you?

KidkArolis:
Yeah, that's great. Sorry for nagging... Just eager to start using yarn..

chrisdhanaraj:
@KidkArolis well... 0.18.0 is here and unfortunately the bug is still here for me :( I get a better error message though!

```
❯ yarn install
yarn install v0.18.0
info No lockfile found.
warning bluemix-design-system@1.0.0: No license field
error An unexpected error occurred: "https://console-npm.ng.bluemix.net/@console%2fbluemix-components: unregistered users are not allowed to access package @console/bluemix-components".
info If you think this is a bug, please open a bug report with the information provided in "/Users/chrisdhanaraj/bluemix/design-system-website/yarn-error.log".
```

owais:
I can confirm 0.18.0 works with private scope perfectly. I'm using npmo/npme to host my private NPM registry.

chrisdhanaraj:
Actually this might be on me! Sorry guys, think our sinopia instance is having issues :)

KidkArolis:
@chrisdhanaraj @rexxars how does your ~/.npmrc file looks like?

KidkArolis:
Ah, sorry, completely missed that

rexxars:
I've made a repo that illustrates this problem: https://github.com/rexxars/yarn-private
Clone it, run `npm install`, then `npm start`, which will start a mock registry.
Now open a terminal to the `target` folder within the same repo, and run `yarn`.
As you'll see, it will use the wrong token for the registry.

KidkArolis:
Thanks! Will check it out.

rexxars:
Since I'm not sure what is causing this bug, I'm just going to mention that I've made an npm package that will find the correct token for a given URL: https://www.npmjs.com/package/registry-auth-token

With that, you should be able to just pass the outgoing request url and get back the corresponding token:
```js
console.log(getAuthToken('//some.host/registry/deep/path', {recursive: true}))
```

Not sure if that's relevant, but hey.

KidkArolis:
@rexxars I think it works if you modify

```
@myscope:registry=https://npm.myscope.no
```

to

```
@myscope:registry=https://npm.myscope.no/
```

Did you create this config by file or was this produced by npm/yarn?

Looking into the exact reason why that trailing slash is so important to the behaviour..
I think it's probably because the authToken is keyed by `//localhost:18888/`, which is different from `http://localhost:18888` - but if we strip protocol, we could also fallback to adding a trailing slash..

KidkArolis:
Created a PR to discuss/address this edge case, but not 100% yet what the "npm spec" says on this matter. https://github.com/yarnpkg/yarn/pull/2168

rexxars:
My configuration was created by npm a while ago. If that's all it takes, I think it's worth fixing  ;-)

chrisdhanaraj:
I'm getting much further in the install process now, but... still KO's near the end; not sure if this just an artifact of our Sinopia instance being misconfigured or something else though


```
❯ yarn install
yarn install v0.18.0
info No lockfile found.
warning bluemix-design-system@1.0.0: No license field
Set-Cookie https://nodesecurity.io/advisories/130
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "http://console-npm.ng.bluemix.net/@console%2fbluemix-components/-/bluemix-components-6.10.3.tgz: Request failed \"403 Forbidden\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/chrisdhanaraj/bluemix/design-system-website/yarn-error.log".
```

rexxars:
@KidkArolis I've updated the test-repository with a newer mock registry which irons out a small inconsistency between how NPM and Yarn URL-encodes (uppercase vs lowercase). Your PR in #2168 fixes the resolving part, but now it seems like the tarball download does not get an Authorization header added.

ainth:
@chrisdhanaraj We're having the same issue, also using sinopia:
```
yarn
yarn install v0.19.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "http://npm.myco.com/babel-eslint/-/babel-eslint-7.1.1.tgz: Request failed \"403 Forbidden\"".
```
My .npmrc is:
```
save-exact=true
strict-ssl=false
always-auth=true
registry=https://npm.myco.com/
email=eng@myco.com
_auth=mahtoken
```
Note that it's failing for a public, non-scoped package too (which puts it somewhat out of the scope of this ticket but you are the first person I've seen to post a similar 403 error so I bet we have the same thing).

PR #1668, opened 2016-11-03
Fix flow types and install test

wyze:
**Summary**

There were some flow errors and test broken introduced with https://github.com/yarnpkg/yarn/commit/96fbeea6afb144652a5a05c0ee8571057d78c4b6 and this should fix them up. It also adds a test for the `--no-bin-links` flag that was introduced with that commit.

**Test plan**

Tests and lint checks pass.

### DIFF
--- a/__tests__/commands/_install.js
+++ b/__tests__/commands/_install.js
@@ -25,7 +25,7 @@ export function runInstall(
 ): Promise<void> {
   return run((config, reporter, lockfile): Install => {
     return new Install(flags, config, reporter, lockfile);
-  }, path.join(fixturesLoc, name), checkInstalled, beforeInstall, cleanupAfterInstall);
+  }, flags, path.join(fixturesLoc, name), checkInstalled, beforeInstall, cleanupAfterInstall);
 }
 
 export async function createLockfile(dir: string): Promise<Lockfile> {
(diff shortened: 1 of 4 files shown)


wyze:
Seems like AppVeyor is being flaky with it's tests. These pass for me in a VM, Windows 7 on node 6.9.0.

__tests__/commands/install.js
```diff
@@ -520,13 +520,23 @@ test.concurrent(
 );
 
 test.concurrent('install should hoist nested bin scripts', (): Promise<void> => {
-  return runInstall({}, 'install-nested-bin', async (config) => {
+  return runInstall({binLinks: true}, 'install-nested-bin', async (config) => {
     const binScripts = await fs.walk(path.join(config.cwd, 'node_modules', '.bin'));
-    assert.equal(binScripts.length, 10);
+    // need to double the amount as windows makes 2 entries for each dependency
+    // so for below, there would be an entry for eslint and eslint.cmd on win32
+    const amount = process.platform === 'win32' ? 20 : 10;
```
Daniel15: Good catch! Unfortunately we do need both on Windows - the extension-less version is used with bash-like shells (eg. "Bash on Ubuntu on Windows", Git Bash, Babun, Cygwin, etc) while the .cmd version is used by regular terminals (cmd.exe, Cmder/ConEmu, PowerShell, etc).

__tests__/commands/_install.js
```diff
@@ -52,6 +52,7 @@ export async function getPackageVersion(config: Config, packagePath: string): Pr
 
 export async function run(
   factory: (config: Config, reporter: Reporter, lockfile: Lockfile) => Install,
+  flags: Object,
```
Daniel15: Any particular reason for adding this to the beginning rather than the end?

wyze: No particular reason. I just wanted to make sure it was before the optional parameters.

Issue #1573, opened 2016-10-31
upgrading packages from git repos not working


omnidan:
I have a dependency that loads from a git repo directly via `git+ssh://`.

Before upgrading:
- `yarn.lock` contains the package with the correct commit hash
- `node_modules` contains the git repo with the correct commit hash

After upgrading via `yarn upgrade`, I'd expect it to update the commit in `yarn.lock` to the latest commit of the repo, as well as updating the repo in `node_modules`.

However, what happens is:
- `yarn.lock` contains the package with the updated commit hash
- `node_modules` **does not** get updated, it still contains the git repo with the state of the old commit

After removing `node_modules` and running `yarn install` again:
- `node_modules` still does not get updated (probably because it got loaded from the cache)

Only when removing the folder manually from `~/.yarn-cache`, removing `node_modules` again, then running `yarn install`, it pulls the latest version of the repository.

Please note that the `package.json` version did not change, only the commit hash did. `yarn.lock` seems to handle this correctly, while the cache doesn't. This also works fine with `npm install` as it just pulls the repo every time.

- node.js version: v6.5.0
- yarn version: 0.16.1
- OS version: macOS 10.12.1

gertjvr:
Had a similar issue, where upgrade changed the `yarn.lock` file correctly but `node_modules` upgraded to the latest npm package rather than my git repo branch reference

node.js v6.9.1
yarn version 0.16.1
OS version: windows 10

jshthornton:
Same issue.
Updated the hash of the commit, yarn lock has the correct hash, removed node modules, yet still getting old version.

debian jessie64
node v6.6.0
yarn v0.16.1

wyze:
Could someone please provide a `package.json` I could use to try and reproduce this?

omnidan:
@wyze this is a bit hard to reproduce as you need to add a `git+ssh` repo, then push to it, and then try to upgrade to the latest version via `yarn`. If you use an existing repo, `yarn` always uses the latest commit (even when the lock file has a different commit specified).

I can summarize the reproduction steps for you:

 * create a new git repo `testrepo`
 * add it to `package.json` dependencies: `"testrepo": "git+ssh://git@github.com/wyze/testrepo.git"`
 * `yarn`
 * add a file to the `testrepo` and push it
 * run `yarn upgrade`
 * file is not available in `node_modules/testrepo/`
 * `rm -rf ~/.yarn-cache/npm-testrepo-1.0.0/`
[...]

Issue #1453, opened 2016-10-25
Question: how are git packages resolved?


fabien:
With `npm` any `git+ssh://` url-based package was downloaded with every `npm install` regardless of using tags, commits - if you updated you pushed your dependency to github, `npm` would fetch it on the next install. 

This was of course rarely desirable - they were fetched even if you 'locked' them to a specific commit, contributing to very long install times.

With Yarn, and it's offline caching mechanism, I am not sure how these situations are handled. From what I've seen, the cache will indeed be used. I wonder how these package versions are resolved. 

Is the `package.json` of the git-based consulted for a semver version number, for instance? In other words, how should I go about, using git(hub) hosted packages optimally with Yarn?

In order to use the cache, but also install updated versions when needed? I think `yarn upgrade` does indeed fetch the latest versions from github, though. EDIT: not quite, I'm having trouble installing an updated version of such a dependency - I had to remove my `~/.yarn*` stuff first.

fabien:
Update: it seems I cannot re-install a github hosted package that I just updated:

```
$ rm -Rf ~/.yarn
$ rm -Rf node_modules/
$ rm yarn.lock
$ yarn install
```

The file was clearly updated on github, yet the package installed is still from a previous revision, even though I haven't specified any tag or commit.

fabien:
Another update: it does indeed seem to work when I increment the version number of the `package.json` of the dependency itself. When I rand `yarn upgrade` it was picked up, whereas before, it just didn't work, no matter what I tried (above).

Question remains: can I force-install and force-fetch dependencies from git(hub)?

VojtechBartos:
@fabien just came into same issue and incrementing the version number worked. Thanks!

panrafal:
This is probably a duplicate of #1278

voxsim:
@bestander we can close it

Issue #1280, opened 2016-10-20
yarn upgrade not upgrading package based on git url


mgutz:
I added a dependency `yarn add dalicious@mgutz/dalicious#matcherino`. That branch was subsequently updated on github. I ran `yarn upgrade dalicious@mgutz/dalicious#matcherino`. The package in node_modules did not upgrade; however, the lock file was updated and correctly references the updated commit. I download and expanded the resolved URL tarball to verify.

I completely removed the node_modules directory and ran `yarn install`. The older version was still in use.

Expected yarn to upgrade the package.

yarn v0.16.1 on Ubuntu 16.04

panrafal:
This is a similar issue to #1278 
Try removing `~/.yarn-cache` and run `yarn` afterwards - it should be updated

bestander:
Should be fixed now #2074

PR #1680, opened 2016-11-04
Add a shasum of pkg.remote.resolved field to cache directory

### LINKED ISSUE #1278: Forked packages use the same entry as upstream
Bug

Packages that are resolved to different sources, use the same cache entry - source fetched first will be used for every package with the same name and version.

- Run `yarn add babel-plugin-typecheck`
- It will be resolved to: `https://registry.yarnpkg.com/babel-plugin-typecheck/-/babel-plugin-typecheck-3.9.0.tgz#0edac7573ae24ee58c6f91319f574bc5124b0f0f`
- Run `yarn add panrafal/babel-plugin-typecheck#base-master` <- this is the forked repo with some changes, same version number
- _Bug#1_ - Yarn will essentially ignore me, keeping the original version in yarn.lock, but will update package.json - npm resolves the current SHA and updates both package.json and shwirnkwrap
- Run `yarn upgrade`
- Yarn will resolve the package to `https://codeload.github.com/panrafal/babel-plugin-typecheck/tar.gz/4fe4e904ae081a47e6221449c27183338999bac0`
- _Bug#2_ - The actual contents will be the same as the original babel-plugin-typecheck - npm fetches package from the repository

I guess this is because yarn caches packages by `<MANAGER>-<PACKAGE>-<VERSION>` instead of the resolved url

Same as NPM - install packages as resolved in yarn.lock

Node 6.5, Yarn 0.16, OSX

panrafal:
**Summary**

This fixes possible name clashes between hosted packages with same version numbers.

It fixes #1278, #1573, #1280, #1453

**Test plan**

By using a shasum of resolved field, we can guarantee, that every resolved url will have it's own cache entry.
Using just `remote.hash` does not have this guarantee, as:
- cache contains contents of whatever `resolved` points to - hence it'll never go out of sync
- it's empty for hosted git (https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L140). Even if we update the code to use the hash, it's taken from yarn.lock. So to assure we are fetching a proper package, people would have to upgrade all dependencies.

@kittens Just when I've created this I saw you started working on using the hash. Check the rationale above as it didn't work for me. Hashing the whole resolved does.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "detect-indent": "^4.0.0",
     "diff": "^2.2.1",
     "eslint-plugin-react": "5.2.2",
+    "hasha": "2.2.0",
     "ini": "^1.3.4",
     "invariant": "^2.2.0",
     "is-builtin-module": "^1.0.0",
--- a/src/config.js
+++ b/src/config.js
@@ -17,6 +17,7 @@ import map from './util/map.js';
 const detectIndent = require('detect-indent');
 const invariant = require('invariant');
 const path = require('path');
+const hasha = require('hasha');
 
 export type ConfigOptions = {
   cwd?: ?string,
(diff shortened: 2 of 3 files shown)


kittens:
#1641 already solves this in a nicer way.

panrafal:
@kittens Please see my comment above. I've tried using hash, but it doesn't work for git-hosted packages, which this is supposed to solve :)

bestander:
@panrafal, what is the status of this PR?
Can you add a test which is not covered by current functionality?

bestander:
Feel free to reopen and rebase if the fix is still relevant

Issue #679, opened 2016-10-11
yarn check fails if there's no lockfiles, yarn install doesn't generate one without any dependencies


bouk:
**What is the current behavior?**

`yarn check` fails if there is no lockfile in the directory, `yarn install` doesn't create on if there are no dependencies in the package.json.

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn init
yarn install
yarn check # fails
```

**What is the expected behavior?**

`yarn check` doesn't fail.

**Please mention your node.js, yarn and operating system version.**

Node v5.7.0
Yarn v0.15.1
macOS 10.12

alecharmon:
so are you asking if it should generate a lock file on `init` or `install`?

bouk:
`yarn install` should probably create a lockfile even if no dependencies are installed

voxsim:
for what i understand @bestander this should be closed

PR #1703, opened 2016-11-05
fix-679: empty lockfiles with no dependencies

verth:
**Summary**

No lockfile is written with empty dependencies, which causes check to fail. Act the same with manifest & lockfile, so write an empty file as-well when no dependencies are going to be installed `dependencies: {}`. 

This will solve: https://github.com/yarnpkg/yarn/issues/679

**Test plan**

```
$ yarn init
$ yarn install
$ yarn check
$ yarn check --integrity
$ npm run test
```

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -162,6 +162,18 @@ test.concurrent("writes new lockfile if existing one isn't satisfied", async ():
   });
 });
 
+test.concurrent('writes a lockfile even when there are no dependencies', (): Promise<void> => {
+  // https://github.com/yarnpkg/yarn/issues/679
+  return runInstall({}, 'install-without-dependencies', async (config) => {
+    const lockfileExists = await fs.exists(path.join(config.cwd, 'yarn.lock'));
+    const installedDepFiles = await fs.walk(path.join(config.cwd, 'node_modules'));
+
+    assert(lockfileExists);
+    // 1 for integrity file (located in node_modules)
+    assert.equal(installedDepFiles.length, 1);
+  });
+});
+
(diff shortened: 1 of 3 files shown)


bestander:
How about a unit test?

verth:
Yes, I was going to, but it was super late already, added it now. I'm sorry, should have made that clear.

bestander:
https://github.com/yarnpkg/yarn/pull/1712

verth:
Cool, cool! Will keep my eye on it

bestander:
Ping #1712 is merged

verth:
I tried to kill the commits and merge master with the test only but it was not passing. When testing install manually without dependencies it also did not write the lockfile for me (with v.0.18.0 last commit https://github.com/yarnpkg/yarn/commit/15d011274eabe89a5fe8d5d32a966ba48c92755e). 

Would you know something more by any chance, intentional? (not pushed yet to keep fix alive atm)

bestander:
@verth, I don't think we had intentional behavior around edge cases with no dependencies.
I think integrity check should pass for empty lockfile, so go ahead rebase and give it another try if you want

verth:
@bestander 

Rebased and dropped the "code fix" in PR to rely on `isSync` fix, seems like something small might still be a little off? Since rebase test it not passing, tested it manually as-well and it seemed the `yarn.lock` was not created.

bestander:
@verth, currently yarn does not write lockfile if there are no dependencies.
A change in install.js is needed

verth:
@bestander Reimplemented a fix now as-well.

My bad this was dragged out so long, thought you only wanted to keep the test in this PR, I probably misinterpreted "currently working on a fix [...] However it would be great to add your test anyway".

PR #1712, opened 2016-11-07
lockfile is rewritten only if there are substantial changes

### LINKED ISSUE #1576: [install] don't write lockfile when yarn.lock is the source of truth

Bug/feature fix

yarn.lock is overwritten on every `yarn install` unless `--pure-lockfile` option is used.
I think it is unnecessary to write yarn.lock if there are no substantial changes to be made.

From #570 by @wycats:

> To be more precise, the expected semantics, in my opinion, are:
> 
> - if package.json has not changed since the last time yarn.lock changed, yarn.lock is the source of truth and should not be updated.
> - if package.json has changed since the last time, yarn.lock changed, update yarn.lock so it satisfies package.json and update node_modules.
> - if yarn update is run, re-resolve all dependencies and get the latest version of everything that satisfies the package.json.

> This means that when a repository is first cloned on a machine, if the yarn.lock was checked in, yarn should always treat it as the source of truth, not generate updates to the yarn.lock, and jump directly to the fetch step.

yarn 0.16.0

bestander:
stacked on #1757

**Summary**

This fixes #1576 - lockfile is not rewritten on every successful install, it checks if integrity hash of installation changed to determine whether lockfile rewrite is needed.

**Test plan**

- added test that lockfile is not rewritten even if comment is added to it


bestander:
In progress, something is wrong here

src/cli/commands/install.js
```diff
@@ -483,13 +483,6 @@ export class Install {
         break;
       }
     }
-    // check if loaded lockfile has patterns we don't have, eg. uninstall
```
bestander: patterns contain only direct dependencies.
Lockfile contains all dependencies, it will always have more than the ones from package.json.

Daniel15:
Looks like this has lint errors - @bestander can you please take a look?

bestander:
Looking into it, there a weird test break

bestander:
found a fix https://github.com/yarnpkg/yarn/pull/1777

PR #1739, opened 2016-11-08
Delay visibility propagation until deps are known

jkrems:
**Summary**

Right now visibility is set before the dependencies are added to the package reference. This means that no visibility information propagates for other requests for the same package. Example:

1. `is-glob@^2.0.0` is encountered. Visibility for the request is "env hidden"
2. Dependencies of `is-glob@^2.0.0` are walked and inherit the visibility from their parent
3. `is-glob@^2.0.1` is encountered. Visibility for the request is "used"
4. Because there's already something compatible in flight, it's merged into the existing ref
5. `addVisibility` is called - but at this point the dependencies are not yet resolved
6. The dependencies are added - but none of them will ever learn that (one of) their parent was "used"

**Test plan**

I couldn't figure out a good unit test for this. What I did for manual testing in an empty directory:

```
npm init --yes
yarn add --dev chokidar@1.6.1
yarn add liftoff@2.3.0
rm -rf node_modules && ../../tools/yarn/bin/yarn.js --prod && node -e "require('is-glob')"
```

Fixes https://github.com/yarnpkg/yarn/issues/761

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -224,8 +224,6 @@ export default class PackageRequest {
       invariant(ref, 'Resolved package info has no package reference');
       ref.addRequest(this);
       ref.addPattern(this.pattern, resolved);
-      ref.addOptional(this.optional);
-      ref.addVisibility(this.visibility);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


jkrems:
Also tried this out against a non-trivial internal app that previously broke. Seems to work with this change.

PR #1741, opened 2016-11-08
Move eslit-plugin-react to dev dependencies

danielhusar:
**Summary**

Move eslint-plugin-react to dev dependencies as its not needed when distributing package.


wyze:
Good catch, thanks!

Issue #1729, opened 2016-11-08
Windows: can not run scripts starting with relative path


whitecolor:
On windows will get the error:

```json
"scripts": {
  "some": "./node_modules/some/some.sh",
}
```

```bash
yarn run some
```

```
"." Is not recognized as an internal or external command
```

**Please mention your node.js, yarn and operating system version.**
Windows 10
node 5.5.0

This is an old issue also with NPM, which they can not fix for ages: https://github.com/npm/npm/issues/13789 though solution is simple just to fix some lines related to path separator.

whitecolor:
This is needs to be fixed some where near: 
https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/run.js#L65

Solution to make it runable on windows is to replace `/` slash with `\\` when `process.platform` is `win32`, but the problem is that how to replaces only slashes that are in the executable path not in their arguments for example.

```
"scripts": {
  "some": "./node_modules/some/some.sh && ../another.sh slashed/arg",
}
```
should be transformed to `"some": ".\\node_modules\\some\\some.sh && ..\\another.sh slashed/arg",`

trojanc:
+1

ctaggart:
It appears that this is fixed in 0.19.0. I ran into this bug on Windows by installing yarn with chocolatey, which is still on 0.18.1 https://chocolatey.org/packages/yarn. Is 0.19 stable or is 0.20 going to the be the next stable version. Not sure how the versioning works for yarn yet.

https://github.com/yarnpkg/yarn/releases/tag/v0.19.0

Fix scripts cmd slashes for windows (#2082)
Alex - Wed, 14 Dec 2016 17:02:36 +1100

bestander:
@ctaggart, I should write a blog post about versioning.
In a few words, stable version is listed in https://yarnpkg.com.

Every couple of weeks we make a "minor" branch cut from current master branch.
For example a branch 0.20-stable is due to be cut.
CI will automatically build the branch HEAD and upload the artifacts to GitHub release marking them as Release Candidates and version 0.20.0.
Once we are happy that 0.20.0 is stable we will mark it so and it can be used by the public.
If there are regressions in 0.20 release we fix them in master and then cherry-pick into 0.20-stable.
After a few cherry-picks we release 0.20.1, 0.20.2 etc. We will mark them as stable if no more bugs are reported.

0.19.1 is in chocolatey already.

bestander:
[...]

Issue #1657, opened 2016-11-03
Cannot execute scripts from ./node_modules/.bin/* as npm does


iongion:
**What is the current behavior?**
I get this output when trying to run a command from `./node_modules/.bin`:
```
yarn run mocha -- ./test/**/**.spec.js
'.' is not recognized as an internal or external command,
```
**If the current behavior is a bug, please provide the steps to reproduce.**
```
yarn add --dev mocha
yarn run mocha -- ./test/**/**.spec.js
```

**What is the expected behavior?**
The command should work and not end abruptly.

**Please mention your node.js, yarn and operating system version.**
node 6.9.1
npm 3.8.3
Win 10

sample **package.json** scripts section
```
...
  "scripts": {
    "build": "./node_modules/.bin/grunt --environment=production build",
    "serve": "./node_modules/.bin/grunt --environment=development serve",
    "grunt": "./node_modules/.bin/grunt",
    "bower": "./node_modules/.bin/bower".
    "mocha": "./node_modules/.bin/mocha"
  }
...
```

torifat:
Might be a `Windows` issue. Working fine for me on `macOS Sierra`. BTW, can you mention `yarn` version too 😄

iongion:
Ah, sure and sorry, it is the latest `v0.16.1`

verth:
Windows indeed. Should be pretty easy fix I think, problem lies in the different path formats. See here: the `run` command just takes over your [scripts format without normalizing](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/run.js#L46) it.

Possible workarounds for now are:
* drop "./node_modules/.bin/", you should not need this if not running directly from cli
* don't use those `scripts`, yarn populates the `run` automatically with the .bin files, so you could run those commands even without defining it in `scripts`

Possible fixes are:
* Regex the scripts before populating them into the `run` so they drop the `./node_modules/.bin`
* Normalize the script path (if it is a path) before running it

Kovensky:
The reason Windows uses backslashes for directory separators is because, before they supported directories, the forward slash was used for options.

`./node_modules/.bin/grunt` is running the `.` command with `/node_modules`, `/.bin`, `/grunt` flags.

Specifying the `node_modules/.bin/` subfolder is completely unnecessary in the `scripts` section; it's not just `yarn` that looks in there first, `npm` does too.

OmgImAlexis:
[...]

Issue #1606, opened 2016-11-01
yarn not resolve POSIX path for postinstall, test and other commands in Windows


pumano:
0.16.1
**Do you want to request a *feature* or report a *bug*?**

*bug*

**What is the current behavior?**

POSIX path not resolved properly on Windows when preinstall or test or another command contains POSIX path. 
Example: 
```
./node_modules/.bin/mocha ./test/ --recursive --check-leaks
```
can't be resolved properly and executed on Windows environment as NPM can do.

**If the current behavior is a bug, please provide the steps to reproduce.**

use POSIX path for mocha run tests as provided previously

**What is the expected behavior?**

POSIX path from postinstall, test, preinstall and other commands should be resolved properly for executing on Windows

**Please mention your node.js, yarn and operating system version.**
nodejs: 4.4.5
yarn: 0.16.1
OS: Windows 10

![yarn](https://cloud.githubusercontent.com/assets/17675259/19911659/ef013e20-a0a5-11e6-963a-410f4589a478.jpg)

DavidBadura:
Colleagues of mine have the same problem with windows

```
C:\xampp\htdocs\x>yarn start
yarn start v0.16.1
$ ./node_modules/.bin/gulp watch
Der Befehl "." ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
error Command failed with exit code 1.
```

Nodejs: 6.9.1
yarn: 0.16.1
OS: Windows 7

under Linux (Ubuntu) and MacOS works fine!

DavidBadura:
my workaround https://github.com/yarnpkg/yarn/issues/1657#issuecomment-258315617

olingern:
Closing this. If this is is still an issue on 1.21.1, please comment and I'll reopen.

Issue #1697, opened 2016-11-05
structure of "bin" within a dependencies package.json is different when installing with yarn or nmp


klausbayrhammer:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When installing e.g. casperjs the structure of "bin" within the `node_modules/casperjs/package.json` looks different whether I use yarn or npm to install the package. The yarn version looks like this:
```json
{
  "bin": "./bin/casperjs"
}
```
the npm version looks like this:
```json
{
  "bin": {
    "casperjs": "./bin/casperjs"
  },
}
```
Projects like BackstopJS rely on the structure of bin section within a dependencies package.json (see garris/BackstopJS#323).

**If the current behavior is a bug, please provide the steps to reproduce.**
`yarn add casperjs`
respectively 
`npm i casperjs`

**Please mention your node.js, yarn and operating system version.**
yarn version v0.16.1
node v6.5.0
npm 3.10.3

kittens:
We aren't normalising `package.json` like npm is doing, we probably need to write the normalised version. Regardless, if your tool is consuming `package.json` then it should be doing this normalisation anyway.

PR #1744, opened 2016-11-09
Normalize scripts object in manifest with relative bin path

verth:
**Summary**

The scripts object in the package.json is not normalized as with npm. The path `./node_modules/.bin/script` is causing issues on Windows. Strip of the relative path with regex as it's not explicitly needed.

Fixes: 
* https://github.com/yarnpkg/yarn/issues/1657
* https://github.com/yarnpkg/yarn/issues/1606

**Test plan**

* Automated: New tests for normalizing scripts have been added
* Manual:

Use a package.json with a relative bin script (can be done on any OS)

```
{
    "name": "",
    "version": "",
    "scripts": {
        "test": "./node_modules/.bin/webpack --watch"
    }
}
```

Notice that the executed command in the output has been normalized:

```
$ yarn run test
yarn run v0.17.0-0
$ webpack --watch
```

### DIFF
--- a/src/util/normalize-manifest/util.js
+++ b/src/util/normalize-manifest/util.js
@@ -74,9 +74,9 @@ export function parseScript(script: string): string {
 
 export function normalizeScripts(scripts: Object): Object {
   const normalizedScripts = {};
-  Object.keys(scripts).forEach((scriptName) => {
+  for (const scriptName in scripts) {
     normalizedScripts[scriptName] = parseScript(scripts[scriptName]);
-  });
+  }
   return normalizedScripts;
 }
 
--- a/src/util/normalize-manifest/util.js
+++ b/src/util/normalize-manifest/util.js
@@ -64,6 +64,22 @@ export function normalizePerson(person: mixed): mixed | PersonObject {
   return parsePerson(stringifyPerson(person));
 }
 
+export function parseScript(script: string): string {
+  const relativeBinPath = /^(\.[\/\\])?node_modules[\/\\].bin[\\\/]/;
+  if (script.match(relativeBinPath)) {
+    script = script.replace(relativeBinPath, '');
+  }
+  return script;
+}
+
+export function normalizeScripts(scripts: Object): Object {
+  const normalizedScripts = {};
+  Object.keys(scripts).forEach((scriptName) => {
+    normalizedScripts[scriptName] = parseScript(scripts[scriptName]);
+  });
+  return normalizedScripts;
+}
+
 export function extractDescription(readme: mixed): ?string {
   if (typeof readme !== 'string' || readme === '') {
     return undefined;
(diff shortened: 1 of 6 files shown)


src/util/normalize-manifest/util.js
```diff
@@ -64,6 +64,22 @@ export function normalizePerson(person: mixed): mixed | PersonObject {
...
+  const relativeBinPath = /^(\.[\/\\])?node_modules[\/\\].bin[\\\/]/;
+  if (script.match(relativeBinPath)) {
+    script = script.replace(relativeBinPath, '');
+  }
+  return script;
+}
+
+export function normalizeScripts(scripts: Object): Object {
+  const normalizedScripts = {};
+  Object.keys(scripts).forEach((scriptName) => {
```
wyze: Change this to `for...in` please.

wyze:
Also, could you show another test where it has something like this in the scripts:

```js
{
  "scripts": {
    "build-prod": "NODE_ENV=production ./node_modules/.bin/webpack"
  }
}
```

As I think this is something we would want normalized, correct?

wyze:
Also, I don't think this addresses #1697 as that has to do with the `bin` field in the `package.json` and this just touches `scripts`.

src/util/normalize-manifest/util.js
verth: Done

verth:
Thanks for the review! I addressed your remarks.

Concerning: `"NODE_ENV=production ./node_modules/.bin/webpack"`

Nice use-case, as you would probably have expected this does not get normalized. Made me think as-well of other use-cases that will (not fully) be normalized, ex:
* `"./node_modules/.bin/script1 && ./node_modules/.bin/script2"`
* `"./node_modules/.bin/script1 | ./node_modules/.bin/script2"`
* Actually anything were the relative path to normalize is not first

We would indeed want these normalized as-well. I tested npm too, and these more specific use-cases fail with `npm run` too, which of course is no reason to not fix them. 

What would be the best way to continue here in your opinion? Create a follow-up issue for these cases, as this will already address the reported issues and bring more consistency with npm, so it's not blocking, or not?

kittens:
Can you link to the part of the npm source code that does this normalisation? While I believe this fixes the original issue I'm pretty sure it's more pathological than just that.

verth:
[run-script from npm](https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/lib/run-script.js#L6) uses [read-package-json](https://github.com/npm/read-package-json/blob/master/read-json.js#L128-L144).

pumano:
@verth Thanks for patch!

whitecolor:
It doesn't fix https://github.com/yarnpkg/yarn/issues/1729
It should be possible to run any executable using `yarn run` scripts.

Npm too [can not run such scripts currenlty](https://github.com/npm/npm/issues/13789).

bestander:
I think we fixed that with different path normalization later on.
Feel free to reopen again if it covers a new tests case.
Sorry for forgetting about this PR.

Issue #1733, opened 2016-11-08
yarn add generates incorrect integrity hash


bestander:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**

```
bestander@DESKTOP-9BB67ID:/mnt/c/Users/besta/work/temp$ yarn add left-pad
yarn add v0.16.1
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
success Saved 1 new dependency
└─ left-pad@1.1.3
Done in 0.79s.
bestander@DESKTOP-9BB67ID:/mnt/c/Users/besta/work/temp$ yarn check integrity
yarn check v0.16.1
warning No license field
error Integrity hashes don't match, expected ed5c5d6324aff23b80a2a42442a68c2685a66aab881999edcfa3fa27a7d0edfa but got 71f5754360935355bce95b334bd455a4b80d135f6960d8eda0ffc26e321ba25e
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

**What is the expected behavior?**

Integrity check should pass fine

**Please mention your node.js, yarn and operating system version.**

0.17 master

bestander:
PR with a fix is coming

PR #1754, opened 2016-11-09
added a test for #1733

bestander:
Added a skipped test for issue #1733.
Fix is a bit involved and is coming in a following PR


bestander:
I'll self-merge considering that this is just a skipped test added

PR #1755, opened 2016-11-09
Use deterministic sort (fixes #1308)

### LINKED ISSUE #1308: Undeterministic version sort order in lockfile

Bug

If a transitive dependency version is specified using `X` (e.g. `3.X.X`) in multiple dependencies (and inconsistently, sometimes 'x') then lowercase and uppercase version are sorted randomly in the lockfile. This results in unwanted changes to the `yarn.lock` file (and git marking it dirty)

``` sh
git clone https://github.com/csabapalfi/yarn-undeterministic-sort
yarn 
rm -rf node_modules && yarn
```

Repeat the last step a few times and you should see the yarn.lock file changing even if it doesn't need to.

Depency version are always sorted the same way.

https://github.com/yarnpkg/yarn/blob/master/src/util/misc.js#L5 should work without lowercasing first but I'm not sure if any other usages of it depend on that.

- yarn: 0.16.1
- node: v6.7.0, v6.9.0
- OSX: 10.11.6, 10.12

zertosh:
**Summary**

The previous sort function used `localeCompare` which is dependent on the user's locale configuration (and I think node support). This resulted in non-deterministic lock files. This new approach uses unicode point code as the rank, which is consistent for everyone.

**Test plan**

Added a test.

cc: @bestander @Daniel15

### DIFF
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -8,18 +8,11 @@ export function sortAlpha(a: string, b: string): number {
   for (let i = 0; i < shortLen; i++) {
     const aChar = a.charCodeAt(i);
     const bChar = b.charCodeAt(i);
-    if (aChar < bChar) {
-      return -1;
-    } else if (aChar > bChar) {
-      return 1;
+    if (aChar !== bChar) {
+      return aChar - bChar;
     }
   }
-  if (a.length < b.length) {
-    return -1;
-  } else if (a.length > b.length) {
-    return 1;
-  }
-  return 0;
+  return a.length - b.length;
 }
 
 export function entries<T>(obj: ?{ [key: string]: T }): Array<[string, T]> {
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -3,8 +3,23 @@
 const _camelCase = require('camelcase');
 
 export function sortAlpha(a: string, b: string): number {
-  // sort alphabetically
-  return a.toLowerCase().localeCompare(b.toLowerCase());
+  // sort alphabetically in a deterministic way
+  const shortLen = Math.min(a.length, b.length);
+  for (let i = 0; i < shortLen; i++) {
+    const aChar = a.charCodeAt(i);
+    const bChar = b.charCodeAt(i);
+    if (aChar < bChar) {
+      return -1;
+    } else if (aChar > bChar) {
+      return 1;
+    }
+  }
+  if (a.length < b.length) {
+    return -1;
+  } else if (a.length > b.length) {
+    return 1;
+  }
+  return 0;
 }
 
 export function entries<T>(obj: ?{ [key: string]: T }): Array<[string, T]> {
--- a/__tests__/util/misc.js
+++ b/__tests__/util/misc.js
@@ -1,0 +1,23 @@
+/* @flow */
+
+import * as misc from '../../src/util/misc.js';
+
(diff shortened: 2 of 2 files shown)


src/util/misc.js
```diff
@@ -3,8 +3,23 @@
...
+  for (let i = 0; i < shortLen; i++) {
+    const aChar = a.charCodeAt(i);
+    const bChar = b.charCodeAt(i);
+    if (aChar < bChar) {
+      return -1;
+    } else if (aChar > bChar) {
+      return 1;
+    }
+  }
+  if (a.length < b.length) {
```
Daniel15: This can just be `return b.length - a.length`

src/util/misc.js
```diff
@@ -3,8 +3,23 @@
...
 
 export function sortAlpha(a: string, b: string): number {
-  // sort alphabetically
-  return a.toLowerCase().localeCompare(b.toLowerCase());
+  // sort alphabetically in a deterministic way
+  const shortLen = Math.min(a.length, b.length);
+  for (let i = 0; i < shortLen; i++) {
+    const aChar = a.charCodeAt(i);
+    const bChar = b.charCodeAt(i);
+    if (aChar < bChar) {
```
Daniel15:
```
if (aChar != bChar) {
  return bChar - aChar;
}
```

Daniel15:
The old version was case-insensitive-ish (by converting to lowercase), the new version isn't. Would that be a problem?

src/util/misc.js
chicoxyzzy: `return a.length - b.length` according to code

src/util/misc.js
chicoxyzzy: should be `return aChar - bChar;` according to code

Daniel15: Oops, my mistake, I got the variables the wrong way around 😆

zertosh:
> The old version was case-insensitive-ish (by converting to lowercase), the new version isn't. Would that be a problem?

I think it **should** be case sensitive. https://www.npmjs.com/package/JSONStream and https://www.npmjs.com/package/jsonstream are distinct packages.

> shouldn't the default sort() (assuming stability) be deterministic?

@kentor I'm can't figure out if that's true. MDN seems to imply that it is. Looking at [v8's implementation](https://github.com/v8/v8/blob/12af412/src/js/array.js#L715-L725), isn't very helpful either - what does `str1 < str2` do?

zertosh:
I don't know what the right thing to do is :/

kentor:
I'm pretty x < y just goes through every char similar to your implementation.

Daniel15:
> I'm pretty x < y just goes through every char similar to your implementation

Hmm, I wonder if it's defined in the spec. `x < y` is pretty vague, especially in a language that uses Unicode. For example, it's not clear whether it does case folding (where characters like `å` and `ä` are considered equal to `a`). It might be implementation-specific, in which case the behaviour can change between Node.js versions and having custom code (like in this diff) is more reliable.

zertosh:
> It might be implementation-specific, in which case the behaviour can change between Node.js versions and having custom code (like in this diff) is more reliable.

That's my inclination too. Feel free to merge at your leisure, or leave this open for more comments.

bestander:
Let's merge this considering that it is an improvement over current sort, we can switch to implied one later

PR #1777, opened 2016-11-10
made test non concurrent + fixed lint

bestander:
fixing trunk

### DIFF
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -28,11 +28,11 @@ test.concurrent('integrity hash respects flat and production flags', async () =>
   const install = new Install({}, config, reporter, lockfile);
 
   const install2 = new Install({flat: true}, config, reporter, lockfile);
-  assert(install2.generateIntegrityHash('foo', []) !== install.generateIntegrityHash('foo', []));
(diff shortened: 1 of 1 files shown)


__tests__/commands/install.js
```diff
@@ -460,7 +460,8 @@ test.concurrent('install should resolve circular dependencies 2', (): Promise<vo
 });
 
 
-test.concurrent('install should respect NODE_ENV=production', (): Promise<void> => {
+// don't run this test in `concurrent`, it will affect other tests
```
wyze: I should have made this comment when I made the test, my mistake.

__tests__/commands/install.js
```diff
@@ -460,7 +460,8 @@ test.concurrent('install should resolve circular dependencies 2', (): Promise<vo
...
+test('install should respect NODE_ENV=production', (): Promise<void> => {
   const env = process.env.NODE_ENV;
   process.env.NODE_ENV = 'production';
```
Daniel15: It's unfortunate that we have a test that relies on global state :( Can we not somehow inject this into the code that needs it, rather than relying on global variables?

Daniel15: (eg. something like a config module that wraps it and could allow us to do `config.setEnvironment('production')`)

PR #1790, opened 2016-11-11
Remove warning about depending on packages with the same name as Node built-ins

### LINKED ISSUE #917: Unnecessary warning for built-in modules: React-Native redboxes if I remove the url dependency
I happily jumped on yarn when it released and have been a very happy user so far! The first project I yarned was a React-native project where I use the `url` module to resolve some url's, as you do. Yarn complained (warning-level) that `url` is a built-in module, so I uninstalled it. React-native did not like this turn of events so I had to reinstall `url` to make my app work again.

I get the rationale behind this warning, but in these cases where the build tool or environment (RN packager in this case) does not have the node.js built-in modules available by default it is unnecessary. I suggest we should trust the dev in this case and not complain if they depend on a built-in module.

Thoughts?

ide:
**Summary**

The reason for this is that Yarn is useful in many JS environments such as browsers and React Native, which don't provide the Node built-ins. Depending on polyfills is the right thing to do in that scenario and the reality is that people intentionally publish these polyfills under the same name as the Node built-ins (e.g. url and buffer).

Fixes #917

**Test plan**

`npm test`

### DIFF
--- a/src/util/normalize-manifest/validate.js
+++ b/src/util/normalize-manifest/validate.js
@@ -111,17 +111,6 @@ export function cleanDependencies(info: Object, isRoot: boolean, reporter: Repor
     depTypes.push([type, deps]);
   }
 
-  // check root dependencies for builtin module names
-  if (isRoot) {
-    for (const [type, deps] of depTypes) {
-      for (const name in deps) {
-        if (isBuiltinModule(name)) {
-          warn(reporter.lang('manifestDependencyBuiltin', name, type));
-        }
-      }
-    }
-  }
-
   // ensure that dependencies don't have ones that can collide
   for (const [type, deps] of depTypes) {
     for (const name in deps) {
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -29,7 +29,6 @@ const messages = {
   manifestLicenseInvalid: 'License should be a valid SPDX license expression',
   manifestLicenseNone: 'No license field',
   manifestStringExpected: '$0 is not a string',
-  manifestDependencyBuiltin: 'Dependency $0 listed in $1 is the name of a built-in module',
   manifestDependencyCollision: '$0 has dependency $1 with range $2 that collides with a dependency in $3 of the same name with version $4',
 
   configSet: 'Set $0 to $1.',


src/util/normalize-manifest/validate.js
```diff
@@ -111,17 +111,6 @@ export function cleanDependencies(info: Object, isRoot: boolean, reporter: Repor
...
-      for (const name in deps) {
-        if (isBuiltinModule(name)) {
-          warn(reporter.lang('manifestDependencyBuiltin', name, type));
-        }
-      }
-    }
-  }
-
   // ensure that dependencies don't have ones that can collide
   for (const [type, deps] of depTypes) {
```
jdalton: This still leaves one `isBuiltinModule` check in place. Is it still needed?

ide: It's a check for whether you're naming your package after a built-in. Arguably if you're doing that you know what you're doing but it affects only a tiny, O(1) number of people so I figured I'd leave that code alone.

bestander:
This is a bit controversial.
Anyone else has opinions?
@yarnpkg/core?

thejameskyle:
I thought tools like Browserify and Webpack replaced built-in modules automatically based on their usage?

ide:
@thejameskyle As a package author, if my intent is for code to run in Node or in a browser, I want to be more precise about which version of `url` (or `path` or `querystring` etc.) I depend on. In this case, I want to depend on a specific range of versions of `url`.

OTOH if my intent was to write code that runs just in Node then I wouldn't list `url` in my dependencies. If someone wanted to run that code in a browser that's their own prerogative and Webpack/Browserify can automatically add the Node built-ins for them. 

So in the former case where I'm intentionally writing environment-agnostic JavaScript I do want to depend on a `url` implementation from npm.

thejameskyle:
Alright I agree with that.

I just wish these packages wouldn't have been given the same name as the native module.

cpojer:
I think this makes sense. @ide you mind rebasing or reverting the fixtures change?

ide:
> I just wish these packages wouldn't have been given the same name as the native module.

Yeah, that'd avoid this issue altogether. FWIW I left in the warning for when the name of the authored package is a built-in's name.

@cpojer yep, reverted & rebased

PR #1810, opened 2016-11-14
Remove crappy folders from node_modules in dist tarballs

kittens:
Decreases our build size.

### DIFF
--- a/scripts/build-dist.ps1
+++ b/scripts/build-dist.ps1
@@ -14,6 +14,4 @@ tar -xzf pack.tgz --strip 1
 rm pack.tgz
 # Change this to "yarn install --production" once #1115 is fixed
 npm install --production
-rm node_modules/*/test -Recurse
-rm node_modules/*/dist -Recurse
 cd ..
--- a/scripts/build-dist.ps1
+++ b/scripts/build-dist.ps1
@@ -14,4 +14,5 @@ tar -xzf pack.tgz --strip 1
 rm pack.tgz
 # Change this to "yarn install --production" once #1115 is fixed
 npm install --production
+../scripts/clean-node-modules.ps1
 cd ..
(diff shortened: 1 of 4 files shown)


scripts/build-dist.sh
```diff
@@ -15,7 +15,7 @@ tar -xzf pack.tgz --strip 1
 rm -rf pack.tgz
 # Change this to "yarn install --production" once #1115 is fixed
 npm install --production
-rm -rf node_modules/*/test node_modules/*/dist
```
kittens: Had to remove the blind `dist` folder removal rule because `rx` distribute their source code in a folder called `dist`.

scripts/build-dist.ps1
```diff
@@ -14,6 +14,4 @@ tar -xzf pack.tgz --strip 1
 rm pack.tgz
 # Change this to "yarn install --production" once #1115 is fixed
 npm install --production
-rm node_modules/*/test -Recurse
-rm node_modules/*/dist -Recurse
```
kittens: These rules aren't up to date so I removed them, not sure where this is actually used. @Daniel15 can you advise?

kittens:
Also we can't put these in `.npmignore` since we don't use `npm pack` to build the final tarball as it excludes `node_modules`.

scripts/build-dist.ps1
bestander: I think this will make the windows build larger? 
Maybe inline the same rm rules here for now?
```
rm node_modules/core-js/client -Recurse
rm node_modules/rx/ts -Recurse
rm node_modules/*/test -Recurse

```

Daniel15:
The PowerShell version is used when building the Windows installer. I think
it'd be fine to switch it to something cross platform like a JS file that
uses rimraf.

Sent from my phone.

On Nov 14, 2016 5:35 AM, "Konstantin Raev" <notifications@github.com> wrote:

Merged #1810 <https://github.com/yarnpkg/yarn/pull/1810>.

—
<https://github.com/yarnpkg/yarn/pull/1810#event-857772237>, or mute the
thread

PR #1813, opened 2016-11-14
Replace `.includes` with `.indexOf` for older Node 4 versions

wyze:
**Summary**

It was reported in https://github.com/yarnpkg/yarn/pull/1792#issuecomment-260364519 that this code did not work on Node v4.

**Test plan**

All tests still pass. Confirmed working on Node v4 here: https://github.com/yarnpkg/yarn/pull/1792#issuecomment-260369372

### DIFF
--- a/src/cli/commands/ls.js
+++ b/src/cli/commands/ls.js
@@ -161,7 +161,7 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
   }
 
   const notDim = tree.color !== 'dim';
-  const found = (filters.includes(tree.name.slice(0, tree.name.lastIndexOf('@'))) : boolean);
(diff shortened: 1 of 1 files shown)


wyze:
Failure on CircleCI is not related. Was a timeout with `normalize-manifest` test.

develar:
Maybe it is better to use [babel-plugin-array-includes](https://www.npmjs.com/package/babel-plugin-array-includes) to avoid such code?

ljharb:
Or https://npm us.com/array-includes :-p then you're version-independent.

develar:
@ljharb `array-includes` it is runtime dependency, babel solution has zero-overhead ;)

PR #1820, opened 2016-11-14
Updated Errors to use this.reporter.lang

devoiddesign:
**Summary**
This edit was made in response to bug: "Update Errors with hard-coded strings to use reporter.lang #1387"

**Test plan/ Motivation**
Future-proofing yarn to make it translatable to different languages by assigning Errors a language file rather than a hard-coded string


bestander:
I don't think this is what @thejameskyle meant.
It is not about changing reporter.lang -> this.reporter.lang but actually finding hardcoded strings and switching them to reporter.lang.

I'll close the PR for now because it fails tests and linters but feel free to open a new one.
Good start!

PR #1844, opened 2016-11-15
Fix no choice for upgrade-interactive

torifat:
**Summary**

Currently when there's no outdated packages. `upgrade-interactive` shows the following without any options to choose from.

```
? Choose which packages to update. (Press <space> to select, <a> to toggle all, <i> to inverse selection)
```


bestander:
Thanks for the PR.
I just merged an alternative fix https://github.com/yarnpkg/yarn/pull/1846

PR #1858, opened 2016-11-15
Add support for --exact & --tilde in upgrade-interactive

torifat:
**Summary**

The way `upgrade-interactive` works, it already support both `--exact` & `--tilde`. So, I have just enabled them.

**Test Plan**
```
$ cat package.json
{
  "dependencies": {
    "react": "15.3.0"
  }
}
$ yarn upgrade-interactive --exact
```

**After**
```
$ cat package.json
{
  "dependencies": {
    "react": "15.3.2"
  }
}
```
---

- [ ] Add `test`

### DIFF
--- a/src/cli/commands/upgrade-interactive.js
+++ b/src/cli/commands/upgrade-interactive.js
@@ -16,8 +16,9 @@ const semver = require('semver');
 export const requireLockfile = true;
 
 export function setFlags(commander: Object) {
-  // TODO: support some flags that install command has
-  commander.usage('update');
+  commander.usage('upgrade-interactive');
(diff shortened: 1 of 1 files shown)


bestander:
How about a test?

torifat:
@bestander I have an interview, will do it after that 😄

torifat:
@bestander `upgrade-interactive` just pass these flags to `Add`. And, `Add` already has tests for these. So, can you point me toward what kind of tests you were thinking?

bestander:
Thanks, @torifat, you are right :)

Issue #1826, opened 2016-11-14
[0.17.0] yarn outdated doesn't report new version of a package that uses a tag


szimek:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
`npm outdated` returns:
```
Package                            Current         Wanted  Latest
webpack                      2.1.0-beta.25  2.1.0-beta.26  1.13.3
webpack-dev-server           2.1.0-beta.10  2.1.0-beta.11  1.16.2
```
`yarn outdated` returns nothing.

**If the current behavior is a bug, please provide the steps to reproduce.**
Add the following entries to `package.json` file:
```
"devDependencies": {
  "webpack": "beta",
  "webpack-dev-server": "beta"
}
```
and somehow make sure you don't have the latest beta versions installed ;)

**What is the expected behavior?**
`yarn outdated` should report both `webpack` and `webpack-dev-server` packages as outdated, similarly to npm.

**Please mention your node.js, yarn and operating system version.**
```
node.js 7.1.0
npm 3.10.9
yarn 0.17.0
macOS 10.12.1
```

wyze:
This is due to #1562 and #1696.

wyze:
Fixed via #1865.

PR #1865, opened 2016-11-15
Consolidate outdated logic and check for wanted > current

wyze:
**Summary**

This PR unifies the logic used within the `outdated` and `upgrade-interactive` commands so we will have consistent results. Also applied a npm compatibility fix that was outlined in #1826.

**Test plan**

All tests pass and `yarn outdated` and `yarn upgrade-interactive` show the same results.


torifat:
Ah, nice 👍 I wanted to do it 😄

bestander:
We need to add __tests__/fixtures/request-cache/GET/localhost/.bin to .gitignore.
@wyze, can you rebase?

wyze:
Added that file explicitly to `.gitignore` and rebased.

PR #1867, opened 2016-11-15
reverted Nested executables #1210 to fix #1823

### LINKED ISSUE #1823: [0.17.0] Error chmod-ing .bin files on 'yarn install'
bug

Running large install after updating to `0.17.0`, yarn erroring out at random times in the build running `chmod` on `.bin` files. This works on `0.16.1`

1st run

> [3/4] 🔗  Linking dependencies...
error An unexpected error occurred: "ENOENT: no such file or directory, chmod 'PATH/project/node_modules/.bin/esparse'".

2nd run

> [3/4] 🔗  Linking dependencies...
error An unexpected error occurred: "ENOENT: no such file or directory, chmod 'PATH/project/node_modules/.bin/rimraf'".

3rd run

> [3/4] 🔗  Linking dependencies...
error An unexpected error occurred: "ENOENT: no such file or directory, chmod 'PATH/project/node_modules/.bin/rimraf'".

Just running yarn on large project after cleaning out all the `node_modules`.

Not erroring out

v6.3.1
0.17.0
El Capitan

bestander:
Fixes #1823.
Looks like nested dependencies binaries can conflict with each other and I reverted #1210 for now.

In a fix we need to expect tests that prevent duplicate bins overwriting

### DIFF
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -521,7 +521,8 @@ test.concurrent(
   },
 );
 
-test.concurrent('install should hoist nested bin scripts', (): Promise<void> => {
+// disabled to resolve https://github.com/yarnpkg/yarn/pull/1210
+test.skip('install should hoist nested bin scripts', (): Promise<void> => {
   return runInstall({binLinks: true}, 'install-nested-bin', async (config) => {
     const binScripts = await fs.walk(path.join(config.cwd, 'node_modules', '.bin'));
     // need to double the amount as windows makes 2 entries for each dependency
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -187,7 +187,7 @@ export default class PackageLinker {
     if (this.config.binLinks) {
       const tickBin = this.reporter.progress(flatTree.length);
       await promise.queue(flatTree, async ([dest, {pkg}]) => {
-        const binLoc = path.join(this.config.cwd, this.config.getFolder(pkg));
+        const binLoc = path.join(dest, this.config.getFolder(pkg));
         await this.linkBinDependencies(pkg, binLoc);
         tickBin(dest);
       }, 4);


bestander:
Self merging

onemen:
With the patch in #1210 all dependencies bins are linked to node_modules/.bin
For example, when installing node-sass ```yarn add --dev node-sass```, the bin for mkdirp version 0.5.1 is linked three times, each time by a different dependency:
&nbsp;&nbsp;&nbsp;&nbsp;fstream@^1.0.0, fstream@^1.0.2 depend on mkdirp ">=0.5 0"
&nbsp;&nbsp;&nbsp;&nbsp;node-gyp@^3.3.1 depend on mkdirp ">=0.5 0"
&nbsp;&nbsp;&nbsp;&nbsp;node-sass@^3.13.0 depend on mkdirp "^0.5.1"
Yarn package-linker call cmdShim with the same 'src' and 'dest' arguments three times.
cmdShim is asynchronous and the calls from Yarn are parallel.
Sometimes a call to cmdShim with the same arguments can happens before the last call to the function completely resolves.

The cause of the error "ENOENT: no such file or directory, chmod 'PATH_TO_BIN'" in windows is the 'cmd-shim' module.
cmd-shim creates 2 file in node_modules/.bin for each bin, sh file and cmd file.
The process of creating these files divided in cmd-shim into four asynchronous steps:
&nbsp;&nbsp;&nbsp;&nbsp;1. fs.writeFile crate the *.cmd file from the bin
&nbsp;&nbsp;&nbsp;&nbsp;2. In parallel to 1, fs.writeFile crate the sh file from the bin
After 1 and 2 resolved without an error
&nbsp;&nbsp;&nbsp;&nbsp;3. fs.chmod set the *.cmd file mode to 0755
&nbsp;&nbsp;&nbsp;&nbsp;4. In parallel to 3, fs.chmod set the sh file mode to 0755

The error is throwing by step 3 or 4 when 2nd call to 'cmd-shim' with the same arguments is in step 1 or 2 while the 1st call is in step 2 or 4.

There are more than one why to fix this bug, here you can find two

Patch 1 – Queued the call to cmdShim
```
diff --git a/src/package-linker.js b/src/package-linker.js
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -48,15 +48,16 @@ export default class PackageLinker {
   async linkSelfDependencies(pkg: Manifest, pkgLoc: string, targetBinLoc: string): Promise<void> {
     targetBinLoc = await fs.realpath(targetBinLoc);
     pkgLoc = await fs.realpath(pkgLoc);
-    for (const [scriptName, scriptCmd] of entries(pkg.bin)) {
+    const promises = entries(pkg.bin).map(async ([scriptName, scriptCmd]): Promise<void> => {
       const dest = path.join(targetBinLoc, scriptName);
       const src = path.join(pkgLoc, scriptCmd);
       if (!await fs.exists(src)) {
         // TODO maybe throw an error
-        continue;
+        return null;
       }
-      await linkBin(src, dest);
-    }
+      return linkBin(src, dest);
+    });
+    await Promise.all(promises);
   }
 
   async linkBinDependencies(pkg: Manifest, dir: string): Promise<void> {
@@ -103,9 +104,9 @@ export default class PackageLinker {
     await fs.mkdirp(binLoc);
 
     // write the executables
-    for (const {dep, loc} of deps) {
-      await this.linkSelfDependencies(dep, loc, binLoc);
-    }
+    await promise.queue(deps, async ({dep, loc}): Promise<void> => {
+      this.linkSelfDependencies(dep, loc, binLoc);
+    });
   }
 
   getFlatHoistedTree(patterns: Array<string>): Promise<HoistManifestTuples> {
@@ -190,7 +191,7 @@ export default class PackageLinker {
         const binLoc = path.join(this.config.cwd, this.config.getFolder(pkg));
         await this.linkBinDependencies(pkg, binLoc);
         tickBin(dest);
-      }, 4);
+      });
     }
   }
``` 
 
Patch 2 – Call cmdShim with same 'dest' after the previous call resolved
```
diff --git a/src/package-linker.js b/src/package-linker.js
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -22,9 +22,13 @@ type DependencyPairs = Array<{
   loc: string
 }>;
 
+const allBins: Map<string, Promise<void>> = new Map();
+
 export async function linkBin(src: string, dest: string): Promise<void> {
   if (process.platform === 'win32') {
-    await cmdShim(src, dest);
+    const next = () => cmdShim(src, dest);
+    const promise = allBins.has(dest) ? allBins.get(dest).then(next) : next();
+    allBins.set(dest, promise);
   } else {
     await fs.mkdirp(path.dirname(dest));
     await fs.symlink(src, dest);
```

bestander:
Go ahead, send a new PR.
Tests for the thing that broke are vital

onemen:
Do you prefer patch 1 or 2?
What would be consider to be a good test for these - yarn add finished without an error?

bestander:
I think the algorithm should be a bit smarter - first priority for direct
bin dependencies, then all the others in any order but don't allow
rewriting existing links.

Should you depend in fs to sync whether file exists?

I would rather build a list of bins and resolve duplicates in memory and
then write them to disk.

The test should verify that proper bins get symlinked in .bin folder
(clashes resolved correctly)

On Wed, 16 Nov 2016 at 11:26, onemen <notifications@github.com> wrote:

Do you prefer patch 1 or 2?
What would be consider to be a good test for these - yarn add finished
without an error?

—
You are receiving this because you modified the open/close state.
<https://github.com/yarnpkg/yarn/pull/1867#issuecomment-260922518>, or mute
the thread

mateodelnorte:
Are we still intending a fix, here, for behavior reverted from #1210?

bestander:
I think it is up for grabs for the community memebers

Issue #1834, opened 2016-11-15
"No such file file or directory, lstat" for Yarn cache during package install on Windows


AntJanus:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

I'm on Windows 10, Node v6.9.1, and Yarn 0.17.0

I'm running `yarn` to install packages for a [personal project](https://github.com/AntJanus/omen/blob/master/package.json) and during installation, I get the error: 

```
error An unexpected error occurred: "ENOENT: no such file or directory, lstat 'C:\\Users\\[my username]\\AppData\\Local\\Yarn\\cache\\npm-gulp-4.0.0-alpha.2-58ae6d972ff84fa07ba9cc30cf7d13d9bc646f39'".
```

I'm guessing yarn is not a big fan of the Gulp version `"github:gulpjs/gulp#4.0"` which causes it to have the above issue. Running `yarn cache clean` and redoing the install doesn't help. 

**What is the expected behavior?**

Installation without errors. Running `npm install` with npm version 3.10.8 works without a hitch.

**Please mention your node.js, yarn and operating system version.**

- Windows 10 Anniversary
- Node v 6.9.1
- Yarn 0.17.0

adrai:
same on ubuntu

eltuza:
same on macOs Sierra 10.12.1

bestander:
I can confirm, this is a regression after 0.16.1

thomaschaaf:
I tried finding the reason and could nail it down to: https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L140 missing the hash attribute. This results in the package being created in the cache folder with a name without the hash (https://github.com/yarnpkg/yarn/blob/master/src/config.js#L265). After it is downloaded however the hash is set here: https://github.com/yarnpkg/yarn/blob/master/src/package-fetcher.js#L91 removing this assignment fixes the problem.

bestander:
@thomaschaaf, would you try sending a PR with a unit test?

thomaschaaf:
@bestander the better option would be to get the hash in https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L140 here. 
Removing the hash updating seems nasty. Or we could not add a hash if a hash was not set until that point.

I am still trying to navigate through the code but can't figure out how it was done before.

thomaschaaf:
[...]

PR #1871, opened 2016-11-15
Skip setting hash if hash was not set previously

thomaschaaf:
**Summary**

This should fix https://github.com/yarnpkg/yarn/issues/1834 where yarn had resolved the hosted-git-packages without a hash and downloaded them to the cache folder `<Packagename>-<Version>`. The tarball fetcher would then add the hash after the download completed resulting in the linker believing that the Package could be found in the cache folder `<Packagename>-<Version>-<Hash>`.

This https://github.com/yarnpkg/yarn/commit/ba5a20307992057c49663fce2569c01f59266701 commit seems to have added the logic behind tarball hashes.

**Test plan**

I added end to end tests and some jest tests
yarn add mulesoft/api-console.git#master would not work. Now works.


bestander:
I would appreciate a unit test here.
It should be a matter of adding a test similar to this https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/add.js#L554 that just checks that dependency gets installed correctly.

thomaschaaf:
I will add unit tests now :) Then we can hopefully merge it very soon

bestander:
Thanks, @thomaschaaf, I'll be ready to merge it at once.

src/package-fetcher.js
```diff
@@ -88,7 +88,11 @@ export default class PackageFetcher {
         newPkg = res.package;
 
         // update with new remote
-        ref.remote.hash = res.hash;
+        // but only if there was a hash previously as the tarball fetcher does not provide a hash.
+        if(ref.remote.hash) {
```
wyze: Just a minor nit to add a space here, `if(` -> `if (`.

bestander:
.tgz files got corrupted

thomaschaaf:
I don't know if this is the correct way to write jest tests as this makes a real http request to github and downloads the files but I did test the unit tests and when I reverted the change the unit tests failed.

end_to_end_tests/README.md
```diff
@@ -3,3 +3,4 @@ This directory contains end-to-end tests for Yarn. The tests perform the followi
 1. Spin up a fresh Docker container
 2. Install Yarn through the package repository
 3. Run `yarn add react` in a test directory
+4. Run `yarn add 'substack/node-mkdirp#master'` in a test directory
```
bestander: let's remove this now that it is covered with unit test.

end_to_end_tests/data/run-yarn-test.sh
```diff
@@ -17,3 +17,4 @@ mkdir yarntest
 cd yarntest
 echo {} > package.json
 yarn add react || fail_with_log
+yarn add 'substack/node-mkdirp#master' || fail_with_log
```
bestander: and this

wyze:
Should there be some assertions in the unit test to ensure the module installed in the `node_modules` folder?

bestander:
Yes, ideally there should be assertations.
Although install/add commands run check afterwards automatically, it would fail without the fix in place.

bestander:
Does the test fail without the fix, @thomaschaaf ?

thomaschaaf:
@bestander yes the tests fail if it is not fixed

bestander:
I'll do 0.17.3 in 30 min

wyze:
Thanks for the iterations @thomaschaaf!

emveeoh:
thanks for the quick fix everyone! :)

PR #1840, opened 2016-11-15
Fix `expected manifest` on global add

torifat:
**Summary**

Fixes #1824

PR #1878, opened 2016-11-16
Add very basic test for global add

torifat:
**Summary**

Add very basic test for `global add` which covers #1840

_I will try to cover the rest as soon as I can._

### DIFF
--- a/__tests__/commands/global.js
+++ b/__tests__/commands/global.js
@@ -5,6 +5,7 @@ import {run as global} from '../../src/cli/commands/global.js';
 import * as fs from '../../src/util/fs.js';
 import * as reporters from '../../src/reporters/index.js';
 import Config from '../../src/config.js';
+import assert from 'assert';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 
@@ -19,7 +20,7 @@ async function runGlobal(
   flags: Object,
   args: Array<string>,
   name: string,
-  checkGloabl?: ?(config: Config, reporter: Reporter, out: string) => ?Promise<void>,
+  checkGlobal?: ?(config: Config, reporter: Reporter, out: string) => ?Promise<void>,
 ): Promise<void> {
   const dir = path.join(fixturesLoc, name);
   const cwd = path.join(
@@ -61,8 +62,8 @@ async function runGlobal(
 
     await global(config, reporter, flags, [command, ...args]);
 
-    if (checkGloabl) {
-      await checkGloabl(config, reporter, out);
+    if (checkGlobal) {
+      await checkGlobal(config, reporter, out);
     }
 
   } catch (err) {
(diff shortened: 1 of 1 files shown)


torifat:
@wyze Can you review this please? I need this to write test for #1889

__tests__/commands/global.js
```diff
@@ -0,0 +1,74 @@
...
+    await config.init({
+      cwd,
+      globalFolder: path.join(cwd, '.yarn/.global'),
+      cacheFolder: path.join(cwd, '.yarn'),
+      linkFolder: path.join(cwd, '.yarn/.link'),
+    });
+
+    await global(config, reporter, flags, [command, ...args]);
+
+    if (checkGloabl) {
```
wyze: Fix misspelling of Global here and the other places.

__tests__/commands/global.js
```diff
@@ -0,0 +1,74 @@
...
+      await checkGloabl(config, reporter, out);
+    }
+
+  } catch (err) {
+    throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
+  }
+}
+test.concurrent('add without flag', (): Promise<void> => {
+  return runGlobal('add', {}, ['react-native-cli'], 'add-without-flag');
```
wyze: Should there be an assertion here to verify the module was added?

__tests__/commands/global.js
torifat: Ah, Thanks 👍

__tests__/commands/global.js
torifat: Sadly it has another issue so I skipped the assertion for the time being.

https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/global.js#L35

This resets the config obejct. So, It is difficult to test things after the command has executed. The returned `config` object no more contains the value we set for `globalFolder`.

wyze: Alright.

torifat: I can assert `out` though. Going for it now.

torifat: Never mind, I thought `out` is `Object` but it's string 😢

wyze: If you switch it to BufferReporter, you could do `reporter.getBuffer()` that will give you an array of output.

wyze: See the tests here as an example: https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/ls.js

torifat: Using `config.cwd`. Things for the idea though. I might use it for something else later 😄

torifat:
@wyze Travis is not working 😞

wyze:
Travis is having issues with the NodeSource Debian repository.

onemen:
The test add bins links to the user bin folder `yarn global bin` instead of the test folder

torifat:
@onemen can you please send a PR with the fix?

onemen:
Done #1918

PR #1889, opened 2016-11-16
Fix global add linking issue

### LINKED ISSUE #1886: [regression] [0.17.2] Yarn global add doesn't add the binary nor warn

bug

If I do `yarn global add angular-cli` as a user, the binary is nowhere to be found.

Either yarn should warn me that it cannot link the binary to /usr/bin or at least there should be, like before, a `.bin` directory inside `~/.config/yarn/global/node_modules`.

yarn 0.17.2

torifat:
**Summary**

`~/.config/yarn/global/node_modules/.bin` was not creating.

Fixes #1886

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -32,7 +32,10 @@ class GlobalAdd extends Add {
 const path = require('path');
 
 async function updateCwd(config: Config): Promise<void> {
-  await config.init({cwd: config.globalFolder});
+  await config.init({
+    cwd: config.globalFolder,
+    binLinks: true,
+  });
 }
 
 async function getBins(config: Config): Promise<Set<string>> {
--- a/__tests__/commands/global.js
+++ b/__tests__/commands/global.js
@@ -70,8 +70,10 @@ async function runGlobal(
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
   }
 }
+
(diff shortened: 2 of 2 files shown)


bestander:
I'll do a 0.17.3 today

src/cli/commands/global.js
```diff
@@ -32,7 +32,10 @@ class GlobalAdd extends Add {
 const path = require('path');
 
 async function updateCwd(config: Config): Promise<void> {
-  await config.init({cwd: config.globalFolder});
+  await config.init({
+    cwd: config.globalFolder,
+    binLinks: true,
```
torifat: > Shouldn't we take the value of config.binLinks instead of hard coding the value of binLinks to true.

That would be awesome but for some reason `config.binLinks` is false here. Since this is a major blocker for lots of people so I decided to fix it with `true` for the time being.

> What about all other option, config.init will set all other Boolean options to false, or to the default value from constants.

I tried what you are trying to do but it will break some other things. For example `prefix`. I have already asked @kittens about it. We might need some refactoring here.

onemen:
This is odd, for me `config.binLinks` is always `true`, unless i use `--no-bin-links` from the command line
I did not find the place in the code to use binLinks from the config file.

The configuration is confusing, it should take the configuration in this order
- command line
- config file
- default

PR #1891, opened 2016-11-16
Remove NodeSource repo from Travis build

Daniel15:
**Summary**
Removes the NodeSource repo from the TravisCI build - I don't think we need it any more

**Test plan**
Will check Travis on the PR


bestander:
1 test passes, let's merge

Issue #1895, opened 2016-11-16
Private npm registry behaviour differs compared to npm


KidkArolis:
This works:

```
npm login
npm install @something/private
```

This does not work:

```
npm login
yarn add @something/private
```

This is a bit tricky to reproduce if you don't have a private npm registry (or any private registry) access.

The config that `npm login` creates by default is:

```
registry=https://registry.qubit.com/
//registry.qubit.com/:_authToken=token
```

Looks like `npm` sends this token to the registry when installing private packages, but `yarn` doesn't. It's probably closely related to #1666 .

PR #1896, opened 2016-11-16
Fix installing from private registry.npmjs.org

KidkArolis:
First attempt at fixing #1895. Discussed in more detail in #1666.

The issue this tries to fix is that rewriting npmjs.org > yarnpkg.com will confuse any logic that relies on reading per registry configuration from the .npmrc file, such as reading authTokens.

I thinks this PR might affect how the lock files get generated, not sure if that's a problem.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -9,6 +9,10 @@ import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import envReplace from '../util/env-replace.js';
 import Registry from './base-registry.js';
 import {addSuffix, removePrefix} from '../util/misc';
+import {YARN_REGISTRY} from '../constants.js';
+
+const NPM_REGISTRY = /http[s]:\/\/registry.npmjs.org/g;
+const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 
 const defaults = require('defaults');
 const userHome = require('user-home');
@@ -16,7 +20,6 @@ const path = require('path');
 const url = require('url');
 const ini = require('ini');
 
-const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 
 function getGlobalPrefix(): string {
   if (process.env.PREFIX) {
@@ -66,7 +69,7 @@ export default class NpmRegistry extends Registry {
     }
 
     return this.requestManager.request({
-      url: requestUrl,
+      url: this.cleanRegistry(requestUrl),
       method: opts.method,
       body: opts.body,
       auth: opts.auth,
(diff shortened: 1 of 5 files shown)


bestander:
@KidkArolis, this change looks reasonable but it is hard to see the side effects.
What Test Scenario would you run for this PR?
Any unit test?

KidkArolis:
Hm, so I tried adding a test, but then realised I couldn't reproduce the issue discussed in #1666. I bought an npm subscription to investigate this further.

I think what happens is this:
1. Resolve registry for @kidkarolis/yarn-npm-issue to https://registry.npmjs.org/.
2. Construct a request URL, https://registry.npmjs.org/@kidkarolis%2fyarn-npm-issue.
3. Package is scoped and request URL matches registry URL, therefore we look for a token.
4. Find token in ~/.npmrc for //registry.npmjs.org/.
5. Send the token.
6. Now we need to download the tarball via the rewritten url https://registry.yarnpkg.com/@kidkarolis/yarn-npm-issue/-/yarn-npm-issue-1.0.1.tgz.
7. Resolve registry for this url to https://registry.yarnpkg.com this time.
8. Package is scoped and request URL matches registry URL (this time yarnpkg.com), therefore we look for a token.
9. Don't find a token for https://registry.yarnpkg.com.
10. Look for DEFAULT_REGISTRY (registry.npmjs.org) and find it.
11. Send the npm registry token to yarn's server.

So, yeah, it works though, so in a way, this isn't an issue.

Let me see if I can summon @SEAPUNK and ask him to reproduce.

KidkArolis:
@SEAPUNK , if you follow these instructions, does it work for you?

1. Download and build yarn's master.
2. `mv ~/.npmrc ~/.npmrc.bak`
3. `npm login`
4. `yarn add @xbpf/suppliers`

SEAPUNK:
@KidkArolis OH, didn't notice this until now. I'll try it out later tonight, if I remember to.

bestander:
ping, @KidkArolis, could you rebase?

jhen0409:
Any update on this? It work for me! 👍

bestander:
@KidkArolis could you rebase once more please?

src/registries/npm-registry.js
```diff
@@ -9,14 +9,17 @@ import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import envReplace from '../util/env-replace.js';
 import Registry from './base-registry.js';
 import {addSuffix, removePrefix} from '../util/misc';
+import {YARN_REGISTRY} from '../constants.js';
+
+const NPM_REGISTRY = /http[s]:\/\/registry.npmjs.org/g;
```
seansfkelley: I think this should be `https?`, not `http[s]`, right?

bestander:
Sorry for forgetting about this is it still relevant?

bestander:
As it is not in a mergeable state I am closing the PR.
Feel free to reopen a new one.

PR #1901, opened 2016-11-16
Gracefully quit when using unsupported node version

### LINKED ISSUE #1898: Node.js version check in yarn.js should be done earlier
Bug

`bin/yarn.js` is supposed to throw an error like "Node version x.xx is not supported, please use Node.js 4.0 or higher" when using an old Node.js version. However, when Node.js 0.10 it doesn't even reach that point, since it hits a syntax error beforehand:
```
yarn/lib-legacy/constants.js:7
const path = require('path');
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
```

I think we need to move the version check to the very top of `yarn.js` before any of our `require` statements are executed, so that it can exit as soon as possible when an outdated/legacy Node.js version is used.

hpurmann:
**Summary**
Fixes #1898

**Test plan**
I couldn't find a good way to test this. I doubt that it's worth the effort of installing an old, unsupported node version only to expect that it quits gracefully.

### DIFF
--- a/bin/yarn.js
+++ b/bin/yarn.js
@@ -4,16 +4,7 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
 'use strict';
 
-// init roadrunner
-var mkdirp = require('mkdirp');
-var constants = require('../lib-legacy/constants');
-mkdirp.sync(constants.GLOBAL_INSTALL_DIRECTORY);
-mkdirp.sync(constants.MODULE_CACHE_DIRECTORY);
-var roadrunner = require('roadrunner');
-roadrunner.load(constants.CACHE_FILENAME);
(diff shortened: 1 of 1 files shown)


Daniel15:
> I couldn't find a good way to test this

Do you use nvm? It's a utility that lets you have multiple different versions of Node.js installed, and switch between them. You could use it to switch to Node.js 0.10 and test.

hpurmann:
Oh yeah I did use `nvm` myself to test this on `0.10.48`. I meant automated tests on CI 😉

wyze:
@hpurmann, could you rebase this please?

hpurmann:
Done

Issue #1724, opened 2016-11-07
.roadrunner.json error


zeke:
:wave:

I just installed a nightly build of yarn to get all the latest goodies. Following the instructions at https://github.com/yarnpkg/yarn/issues/1474 and https://github.com/yarnpkg/website/pull/245:

```sh
wget https://yarnpkg.com/install.sh
chmod +x install.sh
install.sh --nightly
```

The install script told me to remove `~/.yarn` so I did, then re-ran it. Now I get this error when I run `yarn`:

```
❯ yarn -V
0.17.0-20161107.1501
/Users/zeke/.yarn/bin/yarn.js:47
      throw err;
      ^

Error: ENOENT: no such file or directory, open '/Users/zeke/Library/Caches/Yarn/.roadrunner.json'
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.writeFileSync (fs.js:1333:33)
    at /Users/zeke/.yarn/node_modules/roadrunner/index.js:25:6
    at /Users/zeke/.yarn/node_modules/roadrunner/index.js:12:12
    at emitOne (events.js:101:20)
    at process.emit (events.js:188:7)
    at processEmit [as emit] (/Users/zeke/.yarn/node_modules/signal-exit/index.js:140:35)
    at process.exit (internal/process.js:146:15)
    at Command.<anonymous> (/Users/zeke/.yarn/node_modules/commander/index.js:825:13)
```

Once I've actually used yarn to `add` a dep somewhere, this error goes away, pesumably because this directory is created as a side-effect of the installation process?

Daniel15:
I suspect this is due to #1638. @kittens

develar:
workaround: `mkdir /Users/develar/Library/Caches/Yarn`

Daniel15:
I just encountered this on Linux too (Windows Subsystem for Linux on Windows 10):
> Error: ENOENT: no such file or directory, open '/home/daniel/.cache/yarn/.roadrunner.json'

sapegin:
I have the same on Travis CI with a stable version (`npm install -g yarn`).

jbruni:
It just happened to me using Ubuntu inside Virtual Box (host is Windows 10). 
As suggested above, I installed any package using `yarn add` and the error goes away.

reederz:
Same here on Archlinux and yarn@v0.17.0

shaneog:
Same here when installed on CI via curl/bash:
```
[...]

PR #1902, opened 2016-11-17
Ensure cache diretory exists before using roadrunner

wyze:
**Summary**

This solves #1724. Ensures that we always have the cache directory created so roadrunner can work successfully.

**Test plan**

Not sure what kind of tests to add for this. @Daniel15, anything you can think of that could be added to e2e tests for this?

I was able to reduce the error with:

```sh
$ rm -rf ~/Library/Caches/Yarn
$ yarn --version
```

### DIFF
--- a/bin/yarn.js
+++ b/bin/yarn.js
@@ -8,6 +8,7 @@
 var mkdirp = require('mkdirp');
 var constants = require('../lib-legacy/constants');
 mkdirp.sync(constants.GLOBAL_INSTALL_DIRECTORY);
+mkdirp.sync(constants.MODULE_CACHE_DIRECTORY);
 var roadrunner = require('roadrunner');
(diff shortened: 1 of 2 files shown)


Daniel15:
Running `yarn --version` before doing something like `yarn add` was one way to replicate the bug, right? In that case, you could simply add a `yarn --version` call to the top of the e2e test, around here: https://github.com/yarnpkg/yarn/blob/master/end_to_end_tests/data/run-yarn-test.sh#L15

wyze:
Thanks for the direction. What I did will make the current e2e tests fail. If the fix applied in `./bin/yarn.js` is merged, then it should work again.

I'll see if I can investigate a way to get this into a unit test as well.

wellyshen:
The issue still happened, as I mentioned in #1724

PR #1918, opened 2016-11-17
Fix global test - Don't override symlink in the user global bin folder

onemen:
Global test save bin links in the user global bin

### DIFF
--- a/__tests__/commands/global.js
+++ b/__tests__/commands/global.js
@@ -15,6 +15,14 @@ const os = require('os');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'global');
 
+function getBinFolder(prefix: string): string {
+  if (process.platform === 'win32') {
+    return prefix;
+  } else {
+    return path.resolve(prefix, 'bin');
+  }
+}
+
 async function runGlobal(
   command: string,
   flags: Object,
@@ -53,9 +61,8 @@ async function runGlobal(
 
   // set global bin
   if (!flags.prefix) {
-    flags.prefix = path.join(cwd, '.yarn', '.global', '.bin');
+    flags.prefix = path.join(cwd, '.yarn/.tmpbin');
   }
-  await fs.mkdirp(flags.prefix);
 
   try {
     const config = new Config(reporter);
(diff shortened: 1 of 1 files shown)


torifat:
I have covered this in #1918

```
// set global bin
if (!flags.prefix) {
   flags.prefix = path.join(cwd, '.yarn/.tmpbin');
}
```

And, setting `flags.prefix` on false will never let us test the other cases. 

```
  if (flags.prefix) {
    return flags.prefix;
  } else if (config.getOption('prefix')) {
    return String(config.getOption('prefix'));
  } else if (process.env.PREFIX) {
    return process.env.PREFIX;
  } else if (process.platform === 'win32') {
    // c:\node\node.exe --> prefix=c:\node\
    return path.dirname(process.execPath);
  } else {
    // /usr/local/bin/node --> prefix=/usr/local
    let prefix = path.dirname(path.dirname(process.execPath));

    // destdir only is respected on Unix
    if (process.env.DESTDIR) {
      prefix = path.join(process.env.DESTDIR, prefix);
    }

    return prefix;
  }
```

Please let us know if you have any other thoughts.

onemen:
>
> I have covered this in #1919 <https://github.com/yarnpkg/yarn/pull/1919>
>
> // set global bin
> if (!flags.prefix) {
>    flags.prefix = path.join(cwd, '.yarn/.tmpbin');
> }
>
> You have one test that set flags.prefix, but the first test don't set
flags.prefix, so the bin files are added to the user global bin folder
instead of to the temp folder

> And, setting flags.prefix on false will never let us test the other cases.
>
>   if (flags.prefix) {
>     return flags.prefix;
>   } else if (config.getOption('prefix')) {
>     return String(config.getOption('prefix'));
>   } else if (process.env.PREFIX) {
>     return process.env.PREFIX;
>   } else if (process.platform === 'win32') {
>     // c:\node\node.exe --> prefix=c:\node\
>     return path.dirname(process.execPath);
>   } else {
>     // /usr/local/bin/node --> prefix=/usr/local
>     let prefix = path.dirname(path.dirname(process.execPath));
>
>     // destdir only is respected on Unix
>     if (process.env.DESTDIR) {
>       prefix = path.join(process.env.DESTDIR, prefix);
>     }
>
>     return prefix;
>   }
>
> I see your point,
then we have to test the return value of getGlobalPrefix and set the
flags.prefix accordingly.

New issue after #2007. <https://github.com/yarnpkg/yarn/issues/2007>
I'm working on some more global tests that uses fixtures files. the code in
the new _helpers.js run, copy the fixtures files to the temp folder. But
global tests expect the files to be in the .yarn-global folder, so we need
to do the extra step of moving the fixtures again.
<https://github.com/yarnpkg/yarn/issues/2007>

PR #1928, opened 2016-11-18
Add `--version` flag to install script

naganowl:
**Summary**

Adds a --version flag to the installation script, expanding on the nightly flag added in https://github.com/yarnpkg/yarn/pull/1554. When provided, the script will download the specified build of Yarn, rather than the stable released version. It grabs them from the downloads directory as specified in https://github.com/yarnpkg/yarn/issues/1724#issuecomment-260568698 and updates the version check to reflect if the user has installed the same version (regardless of flags provided) in the user's home directory. The speed of this script is ~1 minute faster compared to using `npm` and provides an alternative to switching versions to help with https://github.com/yarnpkg/yarn/issues/1724 as well.

**Test plan**

`./install-latest.sh` gets latest release version (`0.16.1`), `./install-latest.sh --nightly` gets latest nightly build (0.18.0-20161117.1624 right now), `./install-latest.sh --version 0.17.4` installs `0.17.4`.

### DIFF
--- a/scripts/install-latest.sh
+++ b/scripts/install-latest.sh
@@ -109,6 +109,7 @@ yarn_install() {
 
   if [ -d "$HOME/.yarn" ]; then
     if [ -n `which yarn` ]; then
+      local latest_url
       local specified_version
       local version_type
       if [ "$1" = '--nightly' ]; then
--- a/scripts/install-latest.sh
+++ b/scripts/install-latest.sh
@@ -11,6 +11,8 @@ yarn_get_tarball() {
   printf "$cyan> Downloading tarball...$reset\n"
   if [ "$1" = '--nightly' ]; then
     url=https://nightly.yarnpkg.com/latest.tar.gz
+  elif [ "$1" = '--version' ]; then
+    url="https://yarnpkg.com/downloads/$2/yarn-v$2.tar.gz"
   else
     url=https://yarnpkg.com/latest.tar.gz
   fi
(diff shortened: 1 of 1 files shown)


naganowl:
I'm open to renaming the install script (since it no longer installs the latest), but didn't want to make such a sweeping change yet.

scripts/install-latest.sh
```diff
@@ -107,16 +109,24 @@ yarn_install() {
 
   if [ -d "$HOME/.yarn" ]; then
     if [ -n `which yarn` ]; then
+      local SPECIFIED_VERSION
+      local VERSION_TYPE
```
Daniel15: Why do you use `local` for these but `latest_url` doesn't use it?

scripts/install-latest.sh
```diff
@@ -107,16 +109,24 @@ yarn_install() {
...
       if [ "$1" = '--nightly' ]; then
         latest_url=https://nightly.yarnpkg.com/latest-tar-version
+        SPECIFIED_VERSION=`curl $latest_url`
```
Daniel15: Please use lowercase for variables (I know, I messed this up in some files, I need to fix it 😛 ). Apparently uppercase is generally only used for environment variables.

Daniel15:
This is awesome! Thank you!

scripts/install-latest.sh
naganowl: Totally forgot about it, good catch!

PR #1936, opened 2016-11-18
Fix --no-progress flag.

### LINKED ISSUE #1927: `yarn install --no-progress` still prints a progress bar
Bug

Using `yarn install --no-progress` yields a progress bar, despite adding the flag to skip.

Execute `yarn install --no-progress` and behold the progress bar.

It installs the packages without displaying a progress bar.

 node 6.9.1, yarn 0.17.2, OSX 10.10.5

jvhoven:
Fixes #1927. 

**Summary**

Original implementation #1190 used `commander.noProgress` but as per the documentation of commanderjs the flag `--no-progress` sets the value of `commander.progress`, not of `commander.noProgress`. This caused `undefined` to be passed around which would always bypass the check to see if noProgress was set.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -185,7 +185,7 @@ if (commander.json) {
 }
 const reporter = new Reporter({
   emoji: commander.emoji && process.stdout.isTTY && process.platform === 'darwin',
-  noProgress: commander.noProgress,
+  noProgress: !commander.progress,
 });
 reporter.initPeakMemoryCounter();
 


jvhoven:
Appveyor seems to be choking :confused:.
```
The running command stopped because the preference variable "ErrorActionPreference" 
or common parameter is set to 
Stop: npm WARN deprecated node-uuid@1.4.7: use uuid module instead
```

hpurmann:
Is there a way to restart the build (maybe just rebase on current `master`)? Can't think of a reason why this particular change makes Appveyor fail like that.

cc @Daniel15

torifat:
I have rebuilt it. It passed :+1:

torifat:
@jvhoven Can you please add a test case for this?

PR #1942, opened 2016-11-18
Fix: warn instead of error when a shrinkwrap exists in the cwd

shahmirn:
**Summary**

Modified code so that a warning instead of an error is reported. This is so that in a CI environment, if it's set up to fail on any errors, this will fix the issue of a build failing on a false positive.

This is in regards to the comment @Turbo87 added on https://github.com/yarnpkg/yarn/pull/397

**Test plan**


kittens:
Thanks! This was an oversight on my behalf.

PR #1943, opened 2016-11-18
fix #1684 prefer --modules-folder

### LINKED ISSUE #1684: How to use --modules-folder flag?
Sorry if this was asked before but I couldn't find an example in the documentation/google search.

Does this flag mean that I could have a "vendor" directory instead of "node_modules"? How to use it?
I tried something like `yarn install --modules-folder ./vendor/node_modules` and it just installed the modules in `./node_modules`.

Windows 7
yarn 0.16.1

nifgraup:
**Summary**

This change fixes #1684: --modules-folder does not  not work if current folder already has node_modules

**Test plan**

These commands can be used in any test repo to compare results before and after (same commands as used to bisect #1684)

`rm -rf vendor node_modules/ && node ../yarn/lib-legacy/cli/index.js install && node ../yarn/lib-legacy/cli/index.js install --modules-folder vendor && ls vendor
`

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -590,19 +590,19 @@ export class Install {
     const possibleFolders = [];
     if (this.config.modulesFolder) {
       possibleFolders.push(this.config.modulesFolder);
-    }
+    } else {
+      // get a list of registry names to check existence in
+      let checkRegistryNames = this.resolver.usedRegistries;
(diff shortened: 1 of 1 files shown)


bestander:
Thanks, @nifgraup, can you add a test?

bestander:
Feel free to reopen this PR if it is still relevant

Issue #1944, opened 2016-11-18
Current repo commands failing on node version 5.xx.x


olingern:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
<img width="1277" alt="screen shot 2016-11-18 at 4 22 28 pm" src="https://cloud.githubusercontent.com/assets/1470297/20447292/158282a8-adac-11e6-9091-76746517719c.png">

**If the current behavior is a bug, please provide the steps to reproduce.**
pre-requisite: switch to node version `5.xx.x` using [n](https://github.com/tj/n) or [nvm](https://github.com/creationix/nvm)
1.) Clone latest
2.) Use `npm install` or existing globally installed `yarn install` to install deps
3.) `npm run watch`
4.) `bin/yarn ls`
**What is the expected behavior?**
Directory dependencies listed. 

**Please mention your node.js, yarn and operating system version.**
os: _OSX Yosemite 10.10.5_
node: _v5.11.1_ ... switching to v6.2.x fixes this, as yarn installs debug package version `2.3.2`

**Purposed fix:**
Bump Yarn package.json debug version to `2.3.2`. PR here: #1945

Daniel15:
Note that Node.js 5 is no longer supported, please refer to this:
![](https://cdn.rawgit.com/nodejs/LTS/master/schedule.png)
(source: https://github.com/nodejs/LTS)

So any bugs that are specific to Node.js 5.x will most likely not get fixed unless we receive pull requests to fix them. If simply bumping the version number of debug fixes it, then that's great! 😄

olingern:
@Daniel15 I see. This makes sense. I didn't know that Yarn doesn't currently target node version 5.xx.x, so I thought this was a small bug :) .... I believe node 5.xx.x support would be a bit more involved.

PR #1945, opened 2016-11-18
use latest version of debug pkg

olingern:
**Summary**

Refer to #1944

**Test plan**
Bump debug package version

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "cmd-shim": "^2.0.1",
     "commander": "^2.9.0",
     "death": "^1.0.0",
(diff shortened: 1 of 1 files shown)


olingern:
@wyze I did some more testing after @Daniel15 posted the node long term support. Unfortunately, to support node versions 5.xx.x, this PR would have to include a number of other additions. 

For now, I think closing this may be good as bumping the version doesn't solve the problem I originally thought it would :-\

Daniel15:
I think it's still valuable to bump dependencies to the latest version 😄

PR #1438, opened 2016-10-25
Ignore optional dependency if wrong platform

pierrefourgeaud:
https://github.com/yarnpkg/yarn/issues/1435

Expect to solve the issue where optional dependencies' dependencies are not ignored as they should be.

---

**What is the current behavior?**

I get an error while using yarn install instead of just ignoring the dependency.

**If the current behavior is a bug, please provide the steps to reproduce.**

Here is an excerpt of my `package.json`:

```
{
  // ...
  "dependencies": {
    // ...
  },
  "devDependencies": {
    // ...
  },
  "optionalDependencies": {
    "appdmg": "^0.4.5",
    // ...
  },
  // ...
}
```

Then : 
`yarn install`

Result:

```
yarn install v0.16.1
info No lockfile found.
warning airtame-packager@0.2.0: No license field
[1/4] Resolving packages...
warning electron-prebuilt > electron-download > nugget > progress-stream > through2 > xtend > object-keys@0.4.0: 
[2/4] Fetching packages...
warning appdmg@0.4.5: The platform "linux" is incompatible with this module.
info "appdmg@0.4.5" is an optional dependency and failed compatibility check. Excluding it from installation.
error macos-alias@0.2.11: The platform "linux" is incompatible with this module.
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

_As a comment, it is important to underline that **appdmg** has actually a dependency named **ds-store** that has one named **macos-alias**._

**What is the expected behavior?**

Yarn should completely ignore appdmg, therefore any dependency related to it.

**Please mention your node.js, yarn and operating system version.**
Node: 6.5.0
OS: Linux Ubuntu 16.04 (work on OSX 10.10.5 since appdmg is for Darwin).

PR #1954, opened 2016-11-19
Remove optional install error logging

### LINKED ISSUE #1059: Errors from optionalDependency installation scripts should be ignored
Bug, for compatibility with npm

If an optionalDependency install script fails, the whole 'yarn install' fails.
With npm client, these errors are ignored and the install succeeds.

The sinopia npm package has an optionalDependency on fs-ext, whose install script fails in Windows. This is a known issue, but since it is optional the errors get ignored with 'npm install'; but fails with 'yarn'.

In a Windows machine:
yarn add sinopia

This happens even if ignore-optional=true.
If ignore-scripts=true, error goes away.

'yarn install' prints out the errors,  but the installation succeeds.

node.js 6.7.0
yarn 0.15.1
Windows 10

mvestergaard:
**Summary**

When installing a project with optional dependencies, the install for those dependencies are run, and if they fail, the error message is quite overwhelming.

After some digging it looks to me like the error is harmless, but it can be very off-putting.
It also appears that the message output by this `this.reporter.info(this.reporter.lang('optionalModuleFail'));` is somehow swallowed. This would print the following to the user `This module is OPTIONAL, you can safely ignore this error` but it never happens.

The user is already informed earlier in the install process that the package is optional, and will be skipped.

```
warning fsevents@1.0.15: The platform "win32" is incompatible with this module.
info "fsevents@1.0.15" is an optional dependency and failed compatibility check. Excluding it from installation.
```

Fixes #1059

**Test plan**

Created a very simple package with an optional dependency of `fsevents`.

```json
{
  "name": "test",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "optionalDependencies": {
    "fsevents": "1.0.15"
  }
}
```

Before change:
error

After change:
no-error

### DIFF
--- a/src/package-install-scripts.js
+++ b/src/package-install-scripts.js
@@ -109,10 +109,7 @@ export default class PackageInstallScripts {
       const ref = pkg._reference;
       invariant(ref, 'expected reference');
 
-      if (ref.optional) {
-        this.reporter.error(this.reporter.lang('optionalModuleScriptFail', err.message));
-        this.reporter.info(this.reporter.lang('optionalModuleFail'));
-      } else {
+      if (!ref.optional) {
         throw err;
       }
     }


mvestergaard:
You may not agree with silencing the failure, which would be understandable. As far as I can tell the failure is irrelevant.

#1438 may be a better solution, in which case this at least gives you some idea of what the problem is.

mvestergaard:
Closing this again. After trying a bit more, it appears that the failure results in some kind of broken state that makes it retry the installation each time you add/remove packages. Quite annoying. #1438 is a better solution.

Issue #1619, opened 2016-11-02
"invalid tar file"


sibartlett:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

Install fails with this stack trace:

```
Trace: 
  Error: https://registry.yarnpkg.com/eslint-plugin-getsentry/-/eslint-plugin-getsentry-2.0.0.tgz: invalid tar file
      at Extract.Parse._startEntry (/usr/lib/node_modules/yarn/node_modules/tar/lib/parse.js:149:13)
      at Extract.Parse._process (/usr/lib/node_modules/yarn/node_modules/tar/lib/parse.js:131:12)
      at BlockStream.<anonymous> (/usr/lib/node_modules/yarn/node_modules/tar/lib/parse.js:47:8)
      at emitOne (events.js:96:13)
      at BlockStream.emit (events.js:188:7)
      at BlockStream._emitChunk (/usr/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:145:10)
      at BlockStream.flush (/usr/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:70:8)
      at BlockStream.end (/usr/lib/node_modules/yarn/node_modules/block-stream/block-stream.js:66:8)
      at Extract.Parse.end (/usr/lib/node_modules/yarn/node_modules/tar/lib/parse.js:86:23)
      at UnpackStream.onend (_stream_readable.js:511:10)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Installing dependencies using this command: `yarn`

**Please mention your node.js, yarn and operating system version.**

Node: 6.9.1
Yarn: 0.16.1
Dockerfile: https://github.com/sdelements/gitlab-ci-runner/blob/master/Dockerfile

kittens:
There's not enough information here for me to reproduce. What `package.json` are you using and what's the complete command?

swansontec:
I have experienced a similar error:

```
Trace: 
  Error: https://registry.yarnpkg.com/bootstrap-social/-/bootstrap-social-5.0.0.tgz: invalid tar file
      at Extract.Parse._startEntry (/usr/local/lib/node_modules/yarn/node_modules/tar/lib/parse.js:149:13)
      at Extract.Parse._process (/usr/local/lib/node_modules/yarn/node_modules/tar/lib/parse.js:131:12)
[...]

PR #1958, opened 2016-11-20
When tarball fetches fail, report the status code instead of "invalid…

KidkArolis:
Output before:

```
inrainbows:issue-1619 karolis$ ../../yarn/bin/yarn add appdynamics
yarn add v0.18.0-0
info No lockfile found.
warning appdynamics > node-uuid@1.4.7: use uuid module instead
warning appdynamics > set-immediate@0.1.1: Use `setimmediate` instead
error An unexpected error occurred: "http://packages.appdynamics.com/nodejs/4.2.9.0/appdynamics-zmq.tgz: invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/Users/karolis/Documents/workspace/yarn-issues/issue-1619/yarn-error.log".
```

Output after:

```
inrainbows:issue-1619 karolis$ ../../yarn/bin/yarn add appdynamics
yarn add v0.18.0-0
info No lockfile found.
warning appdynamics > node-uuid@1.4.7: use uuid module instead
warning appdynamics > set-immediate@0.1.1: Use `setimmediate` instead
error An unexpected error occurred: "http://packages.appdynamics.com/nodejs/4.2.9.0/appdynamics-zmq.tgz: Request failed \"403 Forbidden\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/karolis/Documents/workspace/yarn-issues/issue-1619/yarn-error.log".
```

Now instead of saying **"invalid tar file"** we report **Request failed "403 Forbidden"**. This could help triage the incoming issues better, because "invalid tar" could be hiding a variety of issues where requests simply fail. E.g. as seen in #1619.

Some pending questions:

- [ ] is this the best way to handle failed requests, `request` docs are a bit mysterious on this
- [ ] why are the quotes escaped in the terminal for `report.lang`s...
- [ ] should we stream in the body and in case it fails to untar, try to JSON.parse it to get more details on the error?

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import http from 'http';
 import {SecurityError, MessageError} from '../errors.js';
 import type {FetchedOverride} from '../types.js';
 import {UnpackStream} from '../util/stream.js';
@@ -157,6 +158,7 @@ export default class TarballFetcher extends BaseFetcher {
       },
       buffer: true,
       process: (req, resolve, reject) => {
+        const {reporter} = this.config;
         // should we save this to the offline cache?
         const mirrorPath = this.getMirrorPath();
         const tarballStorePath = path.join(this.dest, constants.TARBALL_FILENAME);
@@ -170,7 +172,17 @@ export default class TarballFetcher extends BaseFetcher {
           extractorStream,
         } = this.createExtractor(overwriteResolved, resolve, reject);
 
-        //
+        const handleRequestError = (res) => {
+          if (res.statusCode >= 400) {
+            // $FlowFixMe
+            const statusDescription = http.STATUS_CODES[res.statusCode];
+            reject(new Error(
+              reporter.lang('requestFailed', `${res.statusCode} ${statusDescription}`),
+            ));
+          }
+        };
+
+        req.on('response', handleRequestError);
         req.pipe(validateStream);
 
         validateStream
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -215,6 +215,7 @@ const messages = {
   cantRequestOffline: 'Can\'t make a request in offline mode',
   requestManagerNotSetupHAR: 'RequestManager was not setup to capture HAR files',
   requestError: 'Request $0 returned a $1',
+  requestFailed: 'Request failed $0',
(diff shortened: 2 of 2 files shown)


src/fetchers/tarball-fetcher.js
```diff
@@ -170,7 +172,17 @@ export default class TarballFetcher extends BaseFetcher {
           extractorStream,
         } = this.createExtractor(overwriteResolved, resolve, reject);
 
-        //
+        const handleRequestError = (res) => {
+          if (res.statusCode >= 400) {
+            // $FlowFixMe
+            const statusDescription = http.STATUS_CODES[res.statusCode];
```
bestander: what was the error?
I think that might be trivial

KidkArolis:
```
src/fetchers/tarball-fetcher.js:177
177:             const statusDescription = http.STATUS_CODES[res.statusCode];
                                                ^^^^^^^^^^^^ property `STATUS_CODES`. Property not found in
177:             const statusDescription = http.STATUS_CODES[res.statusCode];
                                           ^^^^ module `http`


Found 1 error
```

Daniel15: Good catch! I created an issue for it: https://github.com/facebook/flow/issues/2860

src/fetchers/tarball-fetcher.js
```diff
@@ -170,7 +172,17 @@ export default class TarballFetcher extends BaseFetcher {
...
+            reject(new Error(
+              reporter.lang('requestFailed', `${res.statusCode} ${statusDescription}`),
```
Daniel15: I wonder whether it's worth including the response text (or at least the first eg. 200 bytes, or something like that).

src/fetchers/tarball-fetcher.js
S-YOU: how about when server can't handle the requests temporarily and return 5xx errors?

KidkArolis: I think yarn's retry logic will handle that. This change is only meant to produce more useful error messages.

src/fetchers/tarball-fetcher.js
KidkArolis: It would be useful, but I don't think it's available in this callback. I'd have to get the body separately from the untar stream somehow.

Issue #1956, opened 2016-11-20
disable-progress config option


whitecolor:
Make it possible to permanently disable progress-bar so would be no need to use `--disable-progress` in each command.

olingern:
I can do this and make some minor improvements, too :)

thetrompf:
I believe this one has been fixed in these commits f5d1ded5cc2f6010c478aa9b92f98679eba930eb 2971d856da3f2d43256cdc45a9531d68ee91b97d 5cd11467ee409adf877e8de808c02f89b6d37832 ?

olingern:
The PR I'm working on would use the global yarn config to disable it for the user's machine. Currently, the flag is needed with each flag

thetrompf:
Cool :) maybe you should ping this issue from the PR or the other way around. It is much easier to see what is in progress, especially when this issue is marked as "good first bug" so others don't start implementing it as well? :)

kamilogorek:
This issue can be closed. It already works in `0.21.3`.

```
$ yarn config set no-progress
```

raould:
I set the options for no-progress and no-color but I still get crazy escape sequence characters in my emacs buffer whenever I run yarn command like "yarn add" or even "yarn config list". Am I missing the point / is this unrelated? thank you...

```
$ yarn config list
^[[1Gyarn config v0.23.2
^[[1Ginfo yarn config
^[[1G{ 'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  registry: 'https://registry.yarnpkg.com',
  'strict-ssl': true,
  'user-agent': 'yarn/0.23.2 npm/? node/v4.2.6 linux x64',
  lastUpdateCheck: 1492408117628,
  'no-color': true,
  'no-progress': true }
^[[1Ginfo npm config
^[[1G{}
^[[1GDone in 0.05s.
```

kamilogorek:
@raould can you check if it happens with you use `--disable-progress` directly?

olingern:
@raould This --no-progress is only for the progress bar. The output you're getting just seems to be the the steps yarn goes through during installation

PR #1962, opened 2016-11-21
Disable progress bar

olingern:
**Summary**
Per #1956, this enables users to specify a JSON key, **noProgress**, in their _package.json_ to disable the progress bar from showing. 

**Test plan**
Small refactor for testability, created `__tests__/cli/index.js` file and tests for addition.

**Using feature**
Adding the following the `package.json` will disable progress bar output:
`  "noProgress": true,`


torifat:
IMO, it's better to handle it in `yarn config` than `package,json`. Lets see what others think.

torifat:
@gihrig I wasn't talking about config file in project folder but global config.

gihrig:
@torifat Probly not. I have npm progress bar disabled in global config and never wanted otherwise. So I agree, on second thought, package json makes little sense.

olingern:
@torifat Will refactor this to use `yarn config` unless there are any objections from others :)

torifat:
LGTM cc @bestander

PR #1964, opened 2016-11-21
[WIP] Validating files during publish

### LINKED ISSUE #1930: Verify that files in bin, main, module, files exist on `yarn publish`
A couple times I published packages without the required files or with the wrong paths in these properties:
- `bin`
- `main`
- `module`
- `files`

It would be great to verify that the files specified in there exist before packing and uploading the module. For example:

``` json
{
    "bin": "bin/index.js"
}
```

Does the file `bin/index.js` exist? If not, fail the upload, this is likely an error. 

If that file is meant to be generated on `postinstall` or any time after the installation, then at least give a warning. This could be the default behavior.

Where globs are allowed (`files`) perhaps it could verify that the closest folder exists (e.g. the folder `includes` in `includes/**`) or that at least one file is matched. Alternatively globs could just be ignored.

---

_Originally posted on [npm](https://github.com/npm/npm/issues/13948) and [np](https://github.com/sindresorhus/np/issues/82)_

olingern:
**Summary**
Fixes #1930
This is a proof of concept of validating files upon publishing. My idea is to keep a list of common directory keys, check for their existence, and then iterate over the files pre-publish.

**Test plan**
Will add tests after 👍 

Example of an package.json with a non-existent file:

screen shot 2016-11-26 at 8 42 34 pm

And, the results of running `yarn publish` with a non-existent file:

screen shot 2016-11-20 at 9 54 58 pm

### DIFF
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -4,6 +4,8 @@ import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import NpmRegistry from '../../registries/npm-registry.js';
 import {ConcatStream} from '../../util/stream.js';
+import commonDirectories from '../../util/common-dir.js';
+
 import {MessageError} from '../../errors.js';
 import {setVersion, setFlags as versionSetFlags} from './version.js';
 import * as fs from '../../util/fs.js';
@@ -33,6 +35,20 @@ async function publish(
   if (access && access !== 'public' && access !== 'restricted') {
     throw new MessageError(config.reporter.lang('invalidAccess'));
   }
+  
+  commonDirectories.map((dir) => {
+    if(pkg[dir]){
+
+      for (const file in pkg[dir]) {
+          fs.access(config.cwd + '/' + pkg[dir][file].replace('./',''), ).then((res)=>{
+            // valid files
+          }).catch((err) => {
+            config.reporter.error(`file not found: ${err.path}`);
+            process.exit(1);
+          });
+      }
+    }
+  });
 
   // get tarball stream
   const stat = await fs.lstat(dir);
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -37,15 +37,14 @@ async function publish(
   }
   
   commonDirectories.map((dir) => {
-    if(pkg[dir]){
-
+    if (pkg[dir]) {
       for (const file in pkg[dir]) {
-          fs.access(config.cwd + '/' + pkg[dir][file].replace('./',''), ).then((res)=>{
-            // valid files
-          }).catch((err) => {
-            config.reporter.error(`file not found: ${err.path}`);
-            process.exit(1);
-          });
+        fs.access(config.cwd + '/' + pkg[dir][file].replace('./', '')).then((res) => {
+          // valid files
+        }).catch((err) => {
+          config.reporter.error(`file not found: ${err.path}`);
+          process.exit(1);
+        });
       }
     }
   });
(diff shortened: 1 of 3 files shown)


olingern:
cc @kittens : I looked through the publish's history, it looks like you'd know best.

src/cli/commands/publish.js
```diff
@@ -33,6 +35,19 @@ async function publish(
...
+  
+  commonDirectories.map((dir) => {
+    if (pkg[dir]) {
+      for (const file in pkg[dir]) {
+        fs.access(config.cwd + '/' + pkg[dir][file].replace('./', '')).then((res) => {
+          // valid files
+        }).catch((err) => {
+          config.reporter.error(`file not found: ${err.path}`);
+          process.exit(1);
+        });
```
bestander: A few concerns:
- better use async/awaut than promisses
- process.exit does not seem like the right way to do, a command should not terminate process, we have a top level function for that in cli/index.js
- I don't understand the logic here: it checks that user has access to folders bin/main/modules/files folders? What does it verify?

olingern: > **better use async/awaut than promisses**  

Good catch. Will change.

> **process.exit does not seem like the right way to do, a command should not terminate process, we have a top level function for that in cli/index.js**

Are you referring to this function inside of [cli/index.js](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/cli/index.js#L315) ? I was roughly following [this pattern](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/cli/index.js#L168), and verbose logging for a potentially simple user error (not compiled, typo, etc.) didn't seem necessary .

> **I don't understand the logic here: it checks that user has access to folders bin/main/modules/files folders? What does it verify?**

The feature mentioned in #1930 cites common files that exist in directories that are going to be published. 

> A couple times I published packages without the required files or with the wrong paths in these properties:

> bin
main
module
files

To attempt to solve the user error, I have this logic:  

1.) Keep a list of ["common" directories](https://github.com/olingern/yarn/blob/6a961258fb087a972079ae20927e2e7597f691d2/src/util/common-dir.js).
2.) Compare this list with the user's package.json file  
  *  If a match is found, attempt to access it:  
    ``` fs.access(config.cwd + '/' + pkg[dir][file].replace('./', '')) ```
  * If we come across an entry in package.json that cannot be accessed, then abort publishing and notify the user.  

    ```     
          }).catch((err) => {
            config.reporter.error(`file not found: ${err.path}`);
            process.exit(1);
          }); 
    ```

olingern:
> tests are needed

For sure. This PR is more an idea to attempt to solve a user error. I don't think it's a definitive answer, just a place to start with validating user files pre-publish.

The verification could be "smarter" and actual files / directories specific to that user's project could be collected at some point, pre-verification of files.

src/cli/commands/publish.js
```diff
@@ -4,6 +4,8 @@ import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import NpmRegistry from '../../registries/npm-registry.js';
 import {ConcatStream} from '../../util/stream.js';
+import commonDirectories from '../../util/common-dir.js';
```
olingern: Keys that are directories.... 
Never mind. This is poor naming. Maybe `common-pkg-file-keys.js` would be more fitting.

bestander:
Considering the requested changes I'll close this PR.
Feel free to reopen, looks like a good feature

Issue #1214, opened 2016-10-18
yarn install overwrites linked modules with cached version


lukebennett:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

Running `yarn install` for an app that has linked modules results in yarn copying its cached version of each module to the symlinked directory, overwriting the local copy in the process.

**If the current behavior is a bug, please provide the steps to reproduce.**

I struggled to reproduce this from scratch, it's possible that it's down to a specific combination of factors with my setup. This involves local copies of private/scoped npm modules `yarn link`ed from an application, and referenced in package.json via `file:../xxx`. The version of the modules yarn has in its cache appears to be from when I first installed and ran it a few days ago (there have been numerous changes to the modules since), so it's possible the problem lies with its caching strategy as much as its installation strategy.

One observation I've made is that the problem occurs even when using `yarn install --force` to supposedly bypass the cache.

Happy to answer any questions and/or try anything to help pinpoint the problem.

**What is the expected behavior?**

yarn install should skip linked modules like npm intstall does. Performing the same steps above using npm install does not result in this problem.

**Please mention your node.js, yarn and operating system version.**

node v6.2.2
yarn v0.16.0
Mac OS X v10.10.5

bmayen:
This isn't specific to cached packages. I have a published package `foo@1.0.0`. Locally I link an unpublished package `foo@2.0.0`. My project has a package.json dependency for `foo@2.0.0` and I link to that. When I run `yarn` after linking, it tries to install `foo@2.0.0` from the repo instead of using the linked version and it fails because `foo@2.0.0` hasn't been published yet.

jimdebeer:
really need this, think its a bug

mstdokumaci:
this is breaking

haggholm:
This applies even when you `install` *different* packages. E.g.

```bash
cd @my/local-dep
yarn link
cd ../../@my/main-project
yarn link @my/local-dep
yarn add lodash
```

Even though I didn’t even explicitly do anything with `@my/local-dep`, `yarn` still goes through the modules and installs anything not locally installed, even if it’s linked.

This happens even if I specify `--no-lockfile`.

It’s a particularly nasty problem because it overwrites my local *working copy* of `@my/local-dep`, so it will overwrite anything you haven’t committed yet, effectively erasing local work.

```bash
$ yarn --version
0.16.1
```

joefiorini:
[...]

PR #1970, opened 2016-11-21
Fix for #1214, linked scoped dependencies should not be overwritten

tgriesser:
**Summary**

#1214 identifies a problem with scoped linked dependencies being modified after various `yarn` commands. After digging into it a bit, it appears the problem is with the `package-linker.js`. The `possibleExtraneous` modules do not account for the fact that scoped deps are nested underneath their scope name, so the linked scoped deps are not properly added to the list of ignored packages. This leads yarn to replace the currently linked module with the cached version specified in the lockfile. 

In changing the behavior to properly add these modules, I noticed a test was breaking in `__tests__/commands/remove.js`, however I believe this test to be incorrect. It was testing the entire scope folder was removed rather than just the individual package.

**Test plan**

Added a test with the directory structure after a package is `yarn link`'ed, with a symlink existing in both `.yarn-cache`. The test checks that the linked / scoped `package.json` is kept intact and that the `index.js` is not copied from the cached module into the linked folder.


whitecolor:
Btw potentially yarn allows to have folder starting with `@` that is not a container fo scoped packages but a package itself, think this should be handled.

```
"@some-package": "file:../some-package"
```
so `node_modules/@some-package/package.json` exists.

whitecolor:
Also probably while the removal of scoped package it should remove `@scoped` container folder as well if there is nothing packages left inside.

tgriesser:
> Btw potentially yarn allows to have folder starting with @ that is not a container fo scoped packages but a package itself, think this should be handled.

Interesting. What do you think would be the best solution then? Testing for the existence of `/` in any package name prefixed with `@`?

> Also probably while the removal of scoped package it should remove @scoped container folder as well if there is nothing packages left inside.

Yeah, I wasn't sure where the most appropriate place to put this check was.

whitecolor:
@tgriesser 
I've made this fix for myself too, you may check it https://github.com/whitecolor/yarn/commit/94938a983e5660056e23b4936b0893daf3fe3e32

So it passes removal test without fixes.
I would aslo added test for `upgrade` cmd?

Simlinks should be carefully handled. Currenlty yarn doesn't remove it if you say remove package it it is simlinked (not sure what should be correct behaviour - should it remove the simlink?)

Also there is RFS about `link:` dependency if you didn't see.
https://github.com/yarnpkg/rfcs/issues/25

src/config.js
```diff
@@ -177,9 +177,25 @@ export default class Config {
...
     await fs.mkdirp(this.linkFolder);
-    this.linkedModules = await fs.readdir(this.linkFolder);
+
+    this.linkedModules = [];
+
+    const maybeLinkedModules = await fs.readdir(this.linkFolder);
+
+    for (const maybeLinked of maybeLinkedModules) {
+      const linkedPath = path.join(this.linkFolder, maybeLinked);
+      const isScopedName = maybeLinked.indexOf('@') === 0;
```
kittens: `maybeLinked[0] === '@'` is probably nicer to read

src/config.js
```diff
@@ -177,9 +177,25 @@ export default class Config {
...
     await fs.mkdirp(this.tempFolder);
-
     await fs.mkdirp(this.linkFolder);
-    this.linkedModules = await fs.readdir(this.linkFolder);
+
+    this.linkedModules = [];
+
+    const maybeLinkedModules = await fs.readdir(this.linkFolder);
+
+    for (const maybeLinked of maybeLinkedModules) {
```
kittens: `maybeLinked` isn't the best variable name here, how about `dir`?

src/config.js
```diff
@@ -177,9 +177,25 @@ export default class Config {
...
+      const linkedPath = path.join(this.linkFolder, maybeLinked);
+      const isScopedName = maybeLinked.indexOf('@') === 0;
+      const isScopedContainer = isScopedName && !await fs.exists(path.join(linkedPath, 'package.json'));
```
kittens: Hardcoding `package.json` isn't the best idea since internally we're really agnostic. I'd recommend leaving out this check since you can't have a package named `@foo` without having a slash in it anyway.

tgriesser:
Updated based on @kittens comments. Not sure what I need to fix the failing windows test though.

torifat:
@tgriesser I have figured out the problem.

In Windows `stat.isSymbolicLink()` return false for `yarn-install-dont-overwrite-linked-scoped-<random>/node_modules/@fakescope/fake-dependency`.

https://github.com/yarnpkg/yarn/blob/master/src/package-linker.js#L169

We can borrow the `isSymlink ` from `lerna`.

https://github.com/lerna/lerna/blob/40f8bd413d36eb1aa58d4de3e152c7af42f0526f/src/FileSystemUtilities.js#L82-L99

torifat:
Also, this part works differently in `windows`.

https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/_install.js#L112-L116

tgriesser:
I tried moving the `symlink` into the `beforeInstall` of the test helper, but now it's breaking on all CI's, though it's passing locally on my mac. Any ideas?

bestander:
@tgriesser, could you rebase first?
As for error, it seems to not able to find the path
```

(node:24521) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ENOENT: no such file or directory, symlink '../../dir-to-link' -> '/tmp/yarn-install-dont-overwrite-linked-scoped-0.563408474619264/.yarn-link/@fakescope/fake-dependency
```

torifat:
@bestander his initial PR was only failing in `Windows`. I tried to debug it and found it's a problem with symlink.

I tried this too: https://github.com/jprichardson/node-fs-extra#windows

tgriesser:
Alright, I believe I finally got this one. Needed to ensure the directories were created before the symlink in order for the tests to pass.

Also based on the changes made in https://github.com/yarnpkg/yarn/pull/2036 - in order to properly remove scoped directories when all modules have been removed, we need to maintain a list of these directories separately and check after the `copyBulk` step. Adding the parent (scoped) module to the `possiblyExtraneous` list directly will mess things up in the `copyBulk` step.

whitecolor:
I seem that it wan't published yet? BTW how is it possible to know if a PR was released in published version?

bestander:
@whitecolor, no, it is still on master branch, a release is coming next Monday the latest.
To see if commit is released click on it, e.g. https://github.com/yarnpkg/yarn/commit/6cca59fc84423e07e34191e8d8e3ceb8b348009d.
screen shot 2017-01-03 at 2 35 55 pm

And then see in what branches it is present.

screen shot 2017-01-03 at 2 36 26 pm

Currently it is only in master branch, later you'll see `0.19-stable` there

whitecolor:
@bestander thanks for explanation

Issue #1804, opened 2016-11-14
`yarn bin` path incorrect


shnhrrsn:
Running `yarn bin` in subdirs of the main project dir is reporting an incorrect path:

```bash
$ pwd
/apps/api/build/bin
$ yarn bin
/apps/api/build/bin/node_modules/.bin
$ npm bin
/apps/api/node_modules/.bin
```

I don’t have a `node_modules` dir in `/apps/api/build/bin`, the root `package.json` file and main `node_modules` dir is located in `/apps/api`.

`npm bin` correctly finds the root dir and outputs it, where as it appears `yarn bin` just takes the current dir and appends `node_modules/.bin`

hpurmann:
I'm working on this.

paul-soporan:
Closing as I *think* it's also been fixed in v1 but if not it's definitely fixed in v2 (https://yarnpkg.com/getting-started/migration).

PR #1980, opened 2016-11-22
Test bin command

hpurmann:
**Summary**
This PR is a response to https://github.com/yarnpkg/yarn/issues/1804

In the first step, I added a test for the existing behaviour. I was about to fix the bug @shnhrrsn reported. But then I noticed that none of yarns commands work in subdirectories. This is certainly a difference compared to `npm`. Is this intended?

We could add logic to find the current package root. I did so locally by traversing the folders up and searching for a `package.json`, then taking this as the `cwd`. It would add some complexity which could lead to other bugs. What do you think?

And a minor thing: Do we want to tell the user when he uses too many arguments (like we do in `yarn why`) or do we want be consistent with `npm bin` and just work, no matter how many arguments got passed to `bin`?


hpurmann:
@torifat, @kittens, any thoughts on being able to execute `yarn` in sub-directories?

src/cli/commands/bin.js
```diff
@@ -19,6 +19,6 @@ export function run(
     config.registries[RegistryYarn.registry].folder,
     '.bin',
   );
-  console.log(binFolder);
+  reporter.log(binFolder);
```
kittens: This will output an ansi code at the beginning of the line that resets the cursor to the start making it unsuitable for consumption via another tool.

__tests__/commands/bin.js
```diff
@@ -0,0 +1,45 @@
...
+import {run as bin} from '../../src/cli/commands/bin.js';
+import Config from '../../src/config.js';
+import assert from 'assert';
+import path from 'path';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 6000;
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'bin');
+
+async function runBin(
```
kittens: Can you use the helper method that I cleaned up in #2007?

src/cli/commands/bin.js
hpurmann: Oh ok, interesting. Is there a good way to test the console output of this command without using the `reporter`?

__tests__/commands/bin.js
hpurmann: Ok sure, had a pretty similar thought about these boilerplate code.

bestander:
@hpurmann, what is left to get this merged?
I think it would be a good start to get this feature right.

hpurmann:
Thanks for having a look on this.

Actually a few things are missing here:
* I need to use the helper method @kittens cleaned up in #2007
* I need to know how to test the output of the command if I'm not allowed to use the reporter ([see comments](https://github.com/yarnpkg/yarn/pull/1980#discussion_r89621763)) For example I could spy the console
* Do you want to only have the test or do we want to also change the behaviour (that `yarn bin` works in sub-folders)? Currently, most of `yarn`s commands do not work in sub-folders. This should maybe be first discussed as an RFC

bestander:
2. Let's spy on console for simplicity.
An alternative would be to use a reporter that does only console.

3. Let's start with a test, merge it and move on to the fix discussion

hpurmann:
1. I use the helper method now but got a problem with it (see below in 3). Maybe @kittens or you @bestander can help me with it?
1. I spy on console and use `sinon` to facilitate resetting the spy after each test case and cleaning it up in the end.
1. I still pushed a skipped test for the sub directory thing which makes clear that my current approach for these tests is not working. Ideally, I want to tell `yarn` to execute the `bin` command from a specified directory (the `fixtureLoc`). Then, in the test expectation, I want to check whether it told me the correct `./node_modules/.bin` path in _that_ directory. Unfortunately, the test does create a temporary directory by concatenating the random path and the specified `name` (e.g. "subdir"). So if I'm using this information generated by the test, I do not actually test anything.

package.json
```diff
@@ -71,6 +71,7 @@
     "gulp-watch": "^4.3.5",
     "jest": "^19.0.2",
     "mock-stdin": "^0.3.0",
+    "sinon": "^1.17.7",
```
bestander: I think jest can spy on its own without sinon

hpurmann: Ok, didn't know this. I'll change this. Can you give me a hint how to handle the problem described in 3.?

bestander: Re 3, maybe you should not test for Path but rather test that a script from bin can be executed?

bestander:
@hpurmann, are you blocked by something on this PR?

hpurmann:
No, currently just missing time. Will get back to it.

bestander:
Hey, @hpurmann, your contribution is very much welcome.
Once you have a chance to work on it open a new PR.
I'll close it for now to not keep unactionable PRs open.

Issue #1903, opened 2016-11-17
Dockerfile/Vagrantfile for developing on Yarn


Daniel15:
Inspired by (although somewhat tangential to) this Twitter thread: https://twitter.com/Daniel15/status/799044089035440141

We should add a Dockerfile or Vagrantfile for people that want to develop on Yarn itself, such that there's a consistent development environment and it's easy for people to get started.  There's already third-party Dockerfiles for using Yarn, but nothing for actually developing on Yarn itself.

Maybe Ubuntu or Debian based.

torifat:
I can work on it. I already use one for testing bug 😄

torifat:
@Daniel15 need your suggestion. Currently we have a chicken egg problem in the first run.

Daniel15:
> Currently we have a chicken egg problem in the first run

Is the issue that you need Yarn to install its dependencies? You can install a stable version of Yarn within the Docker container to deal with that, something like this:
```
RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
RUN echo "deb http://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list
RUN apt-get update -qq && apt-get install yarn -y
```

Then you'll have a stable version of Yarn installed, then you can do `yarn install` to install dependencies for the dev version of Yarn. 👍

torifat:
@Daniel15 yeah, I was thinking something similar 😄. Instead of installing it directly I wanted to extract it in `/opt/yarn-stable`. And, I'm linking the development folder to `/opt/yarn`.

viztastic:
Any luck with this? I'd love to help with testing etc.

Daniel15:
I'm going to be moving Yarn to CircleCI 2.0, which uses Docker. I'm going to try creating a Dockerfile for that purpose, and perhaps it could be reused for development as well. The CircleCI team have already created a Dockerfile for this purpose (https://raw.githubusercontent.com/ryanwohara/yarn/circleci-2.0/Dockerfile), I might reuse parts of their Docerfile.

Daniel15:
https://github.com/yarnpkg/yarn/pull/2628

Daniel15:
Pushed to Docker Hub: https://hub.docker.com/r/yarnpkg/dev/

PR #1983, opened 2016-11-22
Add public repository hint for issue creators

hpurmann:
Instead of copying snippets and creating files locally, I want to only clone a provided repository when looking into an issue.
This advice will hopefully be followed by some people and thus simplify reproducing issues.

@torifat: Can we work on a small base repository with a `Dockerfile` and a `package.json` so that people could simply fork it? I saw your comment in https://github.com/yarnpkg/yarn/issues/1903 Maybe we can chat on Discord?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,6 +6,7 @@
 **What is the current behavior?**
 
 **If the current behavior is a bug, please provide the steps to reproduce.**
+<!-- If you can, provide a link to a public repository which contains the files necessary to reproduce this. -->
 
 **What is the expected behavior?**
 


torifat:
Yes, I have one. I can publish it into docker.

> Instead of copying snippets and creating files locally

But, how it will solve this? You still need to run most of the commands.

I only use docker for testing `ubuntu` related stuff.

hpurmann:
I imagine a workflow where the issue reporter only has to

1. Fork base repository
1. Edit `package.json` and add scripts necessary
1. Push to his fork

Then anyone trying to reproduce could

1. Clone the fork
1. Docker build & docker run
1. Execute command inside the container (maybe even a `yarn test`)

Btw: What is your chicken and egg problem?

torifat:
>  What is your chicken and egg problem?

I was working on a `Dockerfile` for development environment. So, when you start the container it will run `gulp watch` but the problem is you need `yarn`/`npm` to install `gulp` 😄

hpurmann:
Ah ok. But this is not specific to a dev environment inside of Docker, right? If I do this on the host, I also have to execute `yarn install`/`npm install` before starting to work on `yarn`.

I don't think this is a big problem. The `yarn` binary you are using should be stable though.

When developing a compiler, it's pretty common to write the compiler _in the language it compiles_ (e.g. CoffeeScript, Rust, Scala). For more information, [see this thread on stackoverflow](http://stackoverflow.com/questions/193560/writing-a-compiler-in-its-own-language).

hpurmann:
Any comments on the actual PR? @torifat ?

torifat:
This would be too much to ask IMO. Please don't even want to follow issue template.

Instead I was thinking about something like `yarn report`, which will send `yarn-error.log`, `os`, `node`, & `yarn` version, & other useful `env` info.

hpurmann:
Oh yeah, `yarn report` sounds like a nice idea. So it would output this to `stdout`, ready to be copied?

torifat:
I was thinking about posting it to a `gist` and then give user a link.

hpurmann:
Ok but then you require a GitHub token, right?

bestander:
Great ideas expressed here, people!
I love the idea of automating the sharing experience of package.json or repro steps.

PR #1996, opened 2016-11-22
Log the artifact that caused the error when a malformed request is found

peterjanes:
**Summary**

When an object downloaded from the registry doesn't have a "dist-tags" property, an error is displayed. However, it doesn't include the name of the object that caused the error, which makes it difficult to debug.

**Test plan**

No additional tests; just a text change.

### DIFF
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -210,7 +210,7 @@ const messages = {
   publishing: 'Publishing',
 
   infoFail: 'Received invalid response from npm.',
-  malformedRegistryResponse: 'Received malformed response from registry. The registry may be down.',
+  malformedRegistryResponse: 'Received malformed response from registry for $0. The registry may be down.',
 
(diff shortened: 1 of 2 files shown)


wyze:
Can you give an example on where I can make this happen?

peterjanes:
I have it occur regularly [retrieving the "natives" package when using a proxy](https://github.com/isaacs/natives/issues/5#issuecomment-243162093). (According to Sonatype, "natives" was unpublished and republished a few days later, so the cached metadata may be corrupt.)

peterjanes:
This builds on #1356, by the way, which fixed #1348.

navels:
@peterjanes we are running into this same problem.  do you have a ticket number with Sonatype or a workaround or any other information?

peterjanes:
@navels Yes, invalidating the cache is the fix. The Sonatype ticket is [16374](http://support.sonatype.com/hc/requests/16374).

Issue #1435, opened 2016-10-25
Error when using yarn install on multi-platforms with optionalDependencies


pierrefourgeaud:
**Do you want to request a _feature_ or report a _bug_?**

Bug.

**What is the current behavior?**

I get an error while using `yarn install` instead of just ignoring the dependency.

**If the current behavior is a bug, please provide the steps to reproduce.**

Here is an excerpt of my package.json:

```
{
  // ...
  "dependencies": {
    // ...
  },
  "devDependencies": {
    // ...
  },
  "optionalDependencies": {
    "appdmg": "^0.4.5",
    // ...
  },
  // ...
}
```

Then : 
`yarn install`

Result: 

```
yarn install v0.16.1
info No lockfile found.
warning airtame-packager@0.2.0: No license field
[1/4] Resolving packages...
warning electron-prebuilt > electron-download > nugget > progress-stream > through2 > xtend > object-keys@0.4.0: 
[2/4] Fetching packages...
warning appdmg@0.4.5: The platform "linux" is incompatible with this module.
info "appdmg@0.4.5" is an optional dependency and failed compatibility check. Excluding it from installation.
error macos-alias@0.2.11: The platform "linux" is incompatible with this module.
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

_As a comment, it is important to underline that **appdmg** has actually a dependency named **ds-store** that has one named **macos-alias**._

**What is the expected behavior?**
Yarn should completely ignore appdmg, therefore any dependency related to it.

**Please mention your node.js, yarn and operating system version.**
Node: 6.5.0
OS: Linux Ubuntu 16.04 (work on OSX 10.10.5 since appdmg is for Darwin).

pierrefourgeaud:
I actually see two solutions to that issue. (I did two commit, preferring the second solution, I did a pull request)

In the file `src/package-request.js`, `find()`:

Solution 1: Simply mark all dependencies of an optionalDependency as optional.
https://github.com/pierrefourgeaud/yarn/commit/f6f159d63b5dfc1b854d5c479d07d6d08e2381b4

Solution 2: Check if the optional dependency works with the os, if it doesn't just pass it.
https://github.com/pierrefourgeaud/yarn/commit/6e6b613cb592d861e28fd07744252bc780536080
https://github.com/yarnpkg/yarn/pull/1438

pierrefourgeaud:
[...]

Issue #1934, opened 2016-11-18
yarn check "commonjs correctness" option


bestander:
**Do you want to request a *feature* or report a *bug*?**
*Feature*

**What is the current behavior?**

`check` command uses internal resolution mechanism to verify node_modules correctness.
If there is a bug in resolver then `install` will install node_modules incorrectly and `check` will fail to identify it.

**New behavior**

`yarn check --commonjs`

We need an option for "correctness" check that makes sure that node_modules is consistent with Node.js requirements independent from Yarn's resolver.

For example (https://github.com/yarnpkg/yarn/issues/1933), React Native node_modules installed by Yarn 0.16.1 is considered incorrect by `check` from Yarn 0.17.4.
However node_modules is correct from Node.js resolution point of view (is it? the tests did not catch it).
This will be a good feature for e2e "correctness" checks of Yarn

bestander:
Done https://github.com/yarnpkg/yarn/pull/2419

PR #1997, opened 2016-11-23
Properly skip installation of packages that do not need to be installed (incompatible optionalDependencies for example)

### LINKED ISSUE #1059: Errors from optionalDependency installation scripts should be ignored

mvestergaard:
**Summary**

This is a different approach to #1438 
It should make sure that packages incompatible with the current environment are not installed.
See #1954 for further motivation

With this change, any `optionalDependencies` that is incompatible with the environment will be skipped entirely in the install.

Installing a package.json like this on windows or linux:
```json
{
  "optionalDependencies": {
    "fsevents": "*"
  }
}
```
Will result in an empty node_modules folder, which is the same behavior as using `npm`.

**Test plan**

The current test 'root install with optional deps' should test the scenario on anything other than OSX. It uses `fsevents` which is only compatible with OSX.
To test on OSX you need to modify `__tests__/fixtures/install/root-install-with-optional-dependency/package.json` to use an optional package that is not compatible with OSX.
CI servers are running linux, so they should be fine.
After running install on a package with a single optional incompatible dependency, node_modules should be empty.

Fixes #1059 #1435


mvestergaard:
Hmm I'm not entirely sure what's causing the check-lockfile script to fail.
It's passing on my machine. Any pointers on what to look for?

src/cli/commands/check.js
```diff
@@ -40,7 +40,7 @@ export async function run(
   }
 
   // get patterns that are installed when running `yarn install`
-  const [, rawPatterns] = await install.hydrate();
+  const [, rawPatterns] = await install.hydrate(true);
```
bestander: Why is fetch true for a check operation?

mvestergaard: It's for this bit:
```js
await this.fetcher.init();
await this.compatibility.init();
```
It seems that if the fetcher is not run beforehand, the compatibility check does not have the os and cpu data for each package to properly do the check.

bestander: Would this work offline?
I am afraid that initing fetcher may cause a longer check

mvestergaard: That's a good question. I mean if the package is in the cache, it doesn't actually fetch anything, right?

mvestergaard: This happens if the fetch is not a part of the check:
```
yarn check v0.18.0-0
error chokidar#fsevents not installed
error fsevents#nan not installed
error fsevents#node-pre-gyp not installed
error node-pre-gyp#npmlog not installed
error node-pre-gyp#rc not installed
error node-pre-gyp#tar-pack not installed
error rc#deep-extend not installed
error tar-pack#fstream-ignore not installed
error rc#minimist not installed
error tar-pack#once not installed
error tar-pack#uid-number not installed
error Found 11 errors.
```

Because the check does not exclude those package due to incompatibility, it expects them to be there, which it shouldn't.

mvestergaard:
@bestander I'll look into adding some tests tonight. 
I suppose it kind of needs to be able to catch any issue on all OS?
Relying on `fsevents` alone will only catch issues on windows and linux, so something alternative has to be done to test on OSX.

src/cli/commands/check.js
bestander: Does check work with clean caches and offline for normal dependencies?

bestander:
> I suppose it kind of needs to be able to catch any issue on all OS?
Relying on fsevents alone will only catch issues on windows and linux, so something alternative has to be done to test on OSX.

You could depend on a fake optional dependency that does not work on any OS.
In install command we use quite a few and save them in the test case' offline morror

src/cli/commands/check.js
mvestergaard: I can imagine it doesn't.

What this change is mainly intended to do is to make sure that calls to `await install.linker.getFlatHoistedTree(rawPatterns);` will exclude any package that is incompatible, or any dependency of such a package.

To be able to determine if a package is incompatible, you would need to know the os and cpu requirement of packages while offline. In which case I guess it needs to be in the lock file or something?

bestander: Ah, it makes sense.
For cross platform modules - they are installed in node_modules and could be verified in place but the missing ones could be either "missing" or not installed because they are "optional".

Adding "optional" property to lockfile makes more sense but this needs a discussion involving more people, maybe an RFC?

bestander: As a matter of fact, I think there should be a third mode for check - https://github.com/yarnpkg/yarn/issues/1934.
I would rather rely on that and the integrity check than the default one

bestander: I wonder, if package is optional but it failed to install because you don't have a compiler, is it an error or is it ok?

mvestergaard: That should be an error.
In my opinion, if it is determined by the OS and CPU requirements of the package that your system is compatible, then any installation error that occurs, are out of the scope of yarn.

mvestergaard:
@bestander How does one go about correctly generating one of those .tgz files found in the `mirror-for-offline` fixture folders?

bestander:
> @bestander How does one go about correctly generating one of those .tgz files found in the mirror-for-offline fixture folders?

It is a bit involved.
Just run `npm pack` on a folder with package.json.
Then you'll need to manually edit yarn.lock file to add sha1 of the file, you could get it with some console.log in tarball-fetcher.js or use a sha1 util on Mac/linux.

mvestergaard:
@bestander I added two tests.
The `dep-incompatible` package has this in it's package.json `"os": [ "invalid" ]` so it is considered incompatible on any OS.
It also has a dependency of `dep-a`. The second test verifies that `dep-a` is also skipped.

bestander:
Great job, @mvestergaard, I'll have a look

mvestergaard:
@bestander Any ETA on next release?

bestander:
Saturday-Sunday this week

friosl:
Several years later, this worked for me, I was trying to do CI in aws and the problem was that package :D

PR #2007, opened 2016-11-23
DRY up command tests

kittens:
As we write more command tests the code kept on being copy and pasted and slightly altered. This was super gross, leaves a lot of redundant code and makes our test infrastructure more fragile. This PR changes all of those to use a single method.


onemen:
New issue after this merge 
I'm working on some more global tests that uses fixtures files. 
The code in the new _helpers.js run, copy the fixtures files to the temp folder. But global tests expect the files to be in the .yarn-global folder, so we need to do the extra step of moving the fixtures again.

onemen:
Forget about my previous post
I can just wrap all the file requires for the test in sub folder of the name `.yarn-global`

Issue #1991, opened 2016-11-22
Yarn shows the available bin commands twice


Haroenv:
**Do you want to request a *feature* or report a *bug*?**: bug

**What is the current behavior?**

Yarn shows the available commands (great feature btw) twice.

```
yarn run
yarn run v0.17.8
error No command specified.
info Commands available from binary scripts: coveralls, eslint, flow, jest, jsdoc, rollup, coveralls, eslint, flow, jest, jsdoc, rollup
info Project commands
   - coveralls
   - doc
   - flow
   - jest
   - lint
   - test
question Which command would you like to run?:
```

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

```
yarn run
yarn run v0.17.8
error No command specified.
info Commands available from binary scripts: coveralls, eslint, flow, jest, jsdoc, rollup
info Project commands
   - coveralls
   - doc
   - flow
   - jest
   - lint
   - test
question Which command would you like to run?:
```

**Please mention your node.js, yarn and operating system version.**

* node 7.1.0
* yarn 0.17.8
* macOS 10.12.1

hpurmann:
Oh, this bug is [even mentioned in the test](https://github.com/yarnpkg/yarn/blob/45a32e0bfe6596437a485fe65edbd6beeb0f1310/__tests__/commands/run.js#L74-L75). I think this can be a good first bug. Essentially, the problem arises because of the `npm` and `yarn` registry looking both in `node_modules`. An additional check on registry's binary paths is needed.

DenisGorbachev:
Hey, I've fixed this via https://github.com/yarnpkg/yarn/pull/2019 - could you please check it?

PR #2019, opened 2016-11-24
Fix duplication of available bin commands (#1991)

DenisGorbachev:
**Summary**

This PR solves https://github.com/yarnpkg/yarn/issues/1991.

**Test plan**

I've modified `__tests__/commands/run.js` to reflect the fixed behavior.

Also, I ran all test suites:

```
Test Suites: 1 skipped, 36 passed, 36 of 37 total
Tests:       18 skipped, 279 passed, 297 total
Snapshots:   50 passed, 50 total
Time:        28.826s
Ran all test suites.
```

### DIFF
--- a/__tests__/commands/run.js
+++ b/__tests__/commands/run.js
@@ -71,8 +71,7 @@ test('lists all available commands with no arguments', (): Promise<void> => {
   return runRun({}, [], 'no-args', (config, reporter): ?Promise<void> => {
     const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
     const scripts = ['build', 'prestart', 'start'];
-    // Notice `cat-names` is below twice as there is a bug with output duplication
-    const bins = ['cat-names', 'cat-names'];
+    const bins = ['cat-names'];
 
     // Emulate run output
     rprtr.error(rprtr.lang('commandNotSpecified'));
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -30,8 +30,8 @@ export async function run(
   const pkg = await config.readManifest(config.cwd);
   const scripts = map();
   const binCommands = [];
+  const visitedBinFolders = [];
   let pkgCommands = [];
-  let visitedBinFolders = [];
   for (const registry of Object.keys(registries)) {
     const binFolder = path.join(config.cwd, config.registries[registry].folder, '.bin');
     if (!visitedBinFolders.includes(binFolder)) {
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -31,13 +31,17 @@ export async function run(
   const scripts = map();
   const binCommands = [];
   let pkgCommands = [];
+  let visitedBinFolders = [];
   for (const registry of Object.keys(registries)) {
     const binFolder = path.join(config.cwd, config.registries[registry].folder, '.bin');
-    if (await fs.exists(binFolder)) {
-      for (const name of await fs.readdir(binFolder)) {
-        binCommands.push(name);
-        scripts[name] = `"${path.join(binFolder, name)}"`;
+    if (!visitedBinFolders.includes(binFolder)) {
+      if (await fs.exists(binFolder)) {
+        for (const name of await fs.readdir(binFolder)) {
+          binCommands.push(name);
+          scripts[name] = `"${path.join(binFolder, name)}"`;
+        }
       }
+      visitedBinFolders.push(binFolder);
     }
   }
   if (pkg.scripts) {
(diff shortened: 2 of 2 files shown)


DenisGorbachev:
I'd love to - however, there's no duplication of the paths inside each registry, so this check has to be done in a for-loop anyway. Or maybe I misunderstand the suggestion?

hpurmann:
Let me phrase it differently: I'd do a small loop to construct an array of paths which could be de-duplicated with something like [uniq](https://lodash.com/docs/4.17.2#uniq) before running the actual loop. This would be one way to solve it in a slightly more functional programming approach (which I think `yarn` is using way too few).

DenisGorbachev:
Like this? https://github.com/yarnpkg/yarn/pull/2019/commits/a968fd0701b93354d33c6c40f438ec0883042e05

hpurmann:
Yeah, I think this is better. You have to be careful though: `lodash` is not part of `yarn`s dependencies (so it currently only works coincidentally). I'd add the package [`lodash.uniq`](https://www.npmjs.com/package/lodash.uniq) separately to `yarn`.

I think moving the
```javascript
for (const registry of Object.keys(registries)) {
...
}
```
loop into a well-named function which takes `registries` and gives back unique `binFolders` makes it a bit easier to reason about. I'm not the only one though, other opinions might differ 😉

torifat:
@hpurmann I think previous approach was simple & clean.

DenisGorbachev:
@hpurmann @torifat how about applying the YAGNI principle and accepting the first approach (simple one)?

hpurmann:
Sure, I'm ok with that too. I just think that the overall codebase could profit from a more functional style. But yeah, consistency-wise, the first solution fits perfectly ;)

torifat:
@hpurmann How, using two loops instead of one is more functional? Or, I'm missing something here?

torifat:
I think here instead of array we could have just use `Set`.

DenisGorbachev:
So let's keep it simple :) I've reverted the code back to original version. Could you please accept the PR?

torifat:
@DenisGorbachev Sorry to bother you for one last time. Can you please just use `Set`. Like in https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/why.js#L113 It would be performant than checking with `includes`. Thanks for your patience ❤️

DenisGorbachev:
@torifat sure, changed to `Set` - let's make yarn faster! :rocket:

torifat:
Thanks for your contribution 😄

DenisGorbachev:
@torifat @hpurmann Thank you for feedback & merge :)

hpurmann:
@torifat Even though this is merged already and discussions about programming styles might lead nowhere, I still want to explain my comment and answer yours.

> How using two loops instead of one is more functional? Or, I'm missing something here?

Ok, functional programming might be a bit far-fetched. I want to keep concerns separated. I want to call well-named pure functions instead of mutating and accessing the same data structure on different lines. To give an example:

```javascript
let visited = []
for (const registry of registries) {
  let path = getPath(registry)
  // Do stuff here
  if (!visited.contains(path)) {
    // Do stuff here
    visited.push(entry)
  }
}
```
There are three lines where we access `visited`. We added these, and the addition is simple. But when reading it afterwards, I have to keep track of this collection and what it could mean to the execution.

In constrast, the example below uses a function to first collect and return the registry entries (which would probably be using a loop internally). Afterwards, we remove the duplicates.
```javascript
const availablePaths = getPathsfromRegistries(registries)
const uniquePaths = uniq(availablePaths)

uniquePaths.map((path) => {
  // Do stuff here
})
```

or even easier, by using [`flow`](https://lodash.com/docs/4.17.2#flow):
```javascript
const uniquePaths = flow(
  getPaths
  uniq
)(registries)
```

When looking at the actual example, there is no need to iterate over `registries` anyway. It's only used to construct the paths.

torifat:
@hpurmann I'm an advocate of this pattern. And, I use things like `ramda`, `ramda-fantasy`, `data.task` a lot. You can check this out: https://github.com/torifat/yarn-update/blob/master/index.js#L15-L34

> Yeah, I think this is better. You have to be careful though: lodash is not part of yarns dependencies (so it currently only works coincidentally). I'd add the package lodash.uniq separately to yarn.

The example you have given in your last comment and the change (https://github.com/yarnpkg/yarn/pull/2019/commits/a968fd0701b93354d33c6c40f438ec0883042e05) you said better are not same.

I also prefer using lots of small functions instead of nested loops and if/else. Check this: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade-interactive.js#L56-L116

I hope you get my point.

PR #2024, opened 2016-11-24
Fix yarn check --integrity behavior not aligning with yarn install's integrity checking

bouk:
**Summary**

There's a mismatch between what `yarn install` does for integrity checking and what `yarn check` does. `yarn check` [flattens the patterns](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/cli/commands/install.js#L342), while `yarn check` doesn't, as it just uses Install.hydrate (which doesn't return the patterns flattened)

We ran into some issues where `yarn install` would say everything was file while `yarn check --integrity` would say there's a mismatch. This patch fixes it

### DIFF
--- a/__tests__/commands/install.js
+++ b/__tests__/commands/install.js
@@ -213,6 +213,13 @@ test.concurrent('install should dedupe dependencies avoiding conflicts 1', (): P
   });
 });
 
+test.concurrent('check and install should verify integrity in the same way when flat', (): Promise<void> => {
+  return runInstall({flat: true}, 'install-should-dedupe-avoiding-conflicts-1', async (config, reporter) => {
+    // Will raise if check doesn't flatten the patterns
+    await check(config, reporter, {flat: true, integrity: true}, []);
+  });
+});
+
 test.concurrent('install should dedupe dependencies avoiding conflicts 2', (): Promise<void> => {
   // A@2 -> B@2 -> C@2
   //            -> D@1
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -41,9 +41,10 @@ export async function run(
 
   // get patterns that are installed when running `yarn install`
   const [, rawPatterns] = await install.hydrate(true);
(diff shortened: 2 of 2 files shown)


bestander:
@bouk, that is a great find, thanks!
Do you mind adding a test case so that it does not break in the future?

bouk:
There don't seem to be any tests for check currently, but I'll see what I can do

bestander:
You can use the install.js tests for the check function

bouk:
I have added a test that checks whether a yarn install --flat followed by a yarn check --integrity --flat succeeds

__tests__/commands/install.js
```diff
@@ -204,6 +204,15 @@ test.concurrent('install should dedupe dependencies avoiding conflicts 1', (): P
   });
 });
 
+test.concurrent('check and install should verify integrity in the same way when flat', (): Promise<void> => {
+  // A@2.0.1 -> B@2.0.0
+  // should result in B@2.0.0 flattened
+  return runInstall({flat: true, integrity: true}, 'install-should-dedupe-avoiding-conflicts-1', async (config) => {
```
bestander: runInstall has an implicit check to verify that all install operations are consistent.
It would be better to run an explicit check with integrity flag because we may refactor runInstall eventually and this test would be hard to keep consistent.

bouk:
@bestander done!

__tests__/commands/_helpers.js
```diff
@@ -23,7 +23,7 @@ export const runInstall = run.bind(
   async (args, flags, config, reporter, lockfile): Promise<Install> => {
     const install = new Install(flags, config, reporter, lockfile);
     await install.init();
-    await check(config, reporter, {}, []);
+    await check(config, reporter, flags, []);
```
bestander: I don't think we should pass flags for install and check via the same variable.
Can you please revert this change?

bestander:
@bouk. thanks, great job!

PR #2025, opened 2016-11-24
Add global add prefix test

torifat:
Test case for `yarn global add --prefix <path> <package>` & `PREFIX=<path> yarn global add <package>`


torifat:
This is weird 😞 

I have added this for testing. Running `yarn test` works, `jest global` works. But, `npm test` or `npm run` doesn't 😕 

```
"scripts": {
  "test": "jest global"
}
```

torifat:
When we run any script via `npm test`, it set config prefix to `/usr/local` (returned by `config.getOption('prefix')`). So, it bails out before reaching `} else if (process.env.PREFIX) {`.

```
  if (flags.prefix) {
    return flags.prefix;
  } else if (config.getOption('prefix')) {
    return String(config.getOption('prefix'));
  } else if (process.env.PREFIX) {
    return process.env.PREFIX;
  } else if (process.platform === 'win32') {
```

@bestander can you suggest anything here?

onemen:
@torifat I tried this and it work on my windows 10:
- Change the test 'add with PREFIX enviroment variable' from concordant to regular.
- Restore env before the assert. 

```
test('add with PREFIX enviroment variable', (): Promise<void> => {
  const tmpGlobalFolder = getTempGlobalFolder();
  const envPrefix = process.env.PREFIX;
  process.env.PREFIX = tmpGlobalFolder;
  return runGlobal(['add', 'react-native-cli'], {}, 'add-with-prefix-env', async (config) => {
    // restore env
    process.env.PREFIX = envPrefix;
    assert.ok(await fs.exists(getGlobalPath(tmpGlobalFolder, 'react-native')));
  });
});
```

The test will still fails  if user set prefix in its config file.
getGlobalPrefix function check `config.getOption('prefix')` before `process.env.PREFIX`

torifat:
@onemen thanks. I will try it.

> The test will still fails if user set prefix in its config file.

I can reset it with a `.yarnrc` in the fixture folder. Or, I could mock `user-home` module.

> Restore env before the assert

I'm not sure how it helped 😕

onemen:
It work for me also on linux
Add **.yarn-global**/.yarnrc in `__tests__/fixtures/global/add-with-prefix-env/.yarn-global/.yarnrc`

onemen:
I'v just update to the new master with https://github.com/yarnpkg/yarn/commit/ac90c23be2766229e46b177f9f7f80916ee3e663

add with PREFIX enviroment variable test failed on Windows and Linux, when the user profile includes **prefix**
- Move add-with-prefix-env/.yarnrc to add-with-prefix-env/.yarn-global/.yarnrc
- Fix typo enviroment > environment

PR #2040, opened 2016-11-25
Use commit hash for icon url

pascalberger:
**Summary**

Use commit hash for linking to the icon in the Chocolatey package. This solves the issue that the Chocolatey package still has a valid icon, even if the file is renamed or moved once (on the downside the hash needs to be updated if icon is changed).

**Test plan**

Ran `build-chocolatey.ps1` and tested on local Chocolatey feed.

### DIFF
--- a/resources/win-chocolatey/yarn.nuspec
+++ b/resources/win-chocolatey/yarn.nuspec
@@ -31,7 +31,7 @@ utilization.
 		<projectSourceUrl>https://github.com/yarnpkg/yarn</projectSourceUrl>
 		<licenseUrl>https://github.com/yarnpkg/yarn/raw/master/LICENSE</licenseUrl>
 		<requireLicenseAcceptance>false</requireLicenseAcceptance>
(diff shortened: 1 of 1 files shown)


torifat:
This seems unnecessary to me. We will update the path if we rename or move the file.

pascalberger:
@torifat Its up to you, but in this case existing packages already published still will be broken. See also eg. the (Chocolatey Core Packages guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#152-use-commit-hash)

torifat:
@pascalberger in that case we will have to automate this.

thetrompf:
@Daniel15 It seems like the link to yarn-kitten-circle.png is dead?

Daniel15:
> It seems like the link to yarn-kitten-circle.png is dead?

The file doesn't exist on the site yet. My suggestions was to copy the asset to the site, then use that as the URL.

bestander:
Let's close this PR then.
Would be great to get a PR to upload the image to website assets and then a new PR with a proper image URL here

PR #2041, opened 2016-11-25
Update build script so that it can be run from any directory

pascalberger:
**Summary**

Updates the `build-chocolatey.ps1` package so that it can be run from any directory and not only from the root directory.

**Test plan**

Ran `build-chocolatey.ps1` from root and scripts directory


Daniel15:
Nice, I didn't know about this :D

PR #2044, opened 2016-11-26
Remove devDependencies resolution in production mode

### LINKED ISSUE #761: `yarn install --production` doesn't install correct dependencies

torifat:
When `dependencies` & `devDependencies` have common dependency, it fails in `production` mode.

**TODO**
- [x] Add test

Fixes #761, #1014


torifat:
761 is a different issue. `entries-test` contains `node_modules` folder. 

https://registry.yarnpkg.com/entries-test/-/entries-test-1.0.1.tgz#1bf192e414ceadd0cf4b77b3969df32de2985d50

torifat:
Nested shared dependency doesn't work when `yarn --production`. Only `dep-a` & `dep-c` get installed.

screenshot 2016-11-27 19 51 51

kittens:
This isn't a correct fix, we need to consider dev dependencies when we're performing package hoisting so the same hoisting structure is retained in dev and prod. The `visibility` flag is what toggles this.

torifat:
@kittens ok, got it. I thought it wouldn't matter. If you want to work on it then you can use my test case.

torifat:
I can give it a try too. But, might need some time 😄

torifat:
https://github.com/yarnpkg/yarn/blob/master/src/package-resolver.js#L432-L434

Here, for the attached test case. `dep-b` sets `dep-d` as `ENVIRONMENT_IGNORE` and `this.fetchingPatterns[fetchKey]` to true. So, for `dep-c` it doesn't get updated.

kittens:
Superseded by #2116.

Issue #1316, opened 2016-10-20
possibility to specify cache-folder in yarnrc


ptrstpp950:
**Do you want to request a _feature_ or report a _bug_?**
A feature
**What is the current behavior?**
I don't know how to specify cache-folder in yarnrc. It is possible form CLI using --cache-folder option
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
An option in config
**Please mention your node.js, yarn and operating system version.**
yarn 16.1
node 6.5.0
windows 10

mmerickel:
I think this is a duplicate of #1211.

victornoel:
Would be a great feature!

valscion:
I suppose that yarn 0.18.0 released with PR #2046 makes this issue fixed, and this could be closed?

wyze:
Yup, thanks @valscion.

Issue #1211, opened 2016-10-18
Allow for ENVVAR specification of YARN cache directory location


ryanmt:
**Do you want to request a _feature_ or report a _bug_?**
Feature request

**What is the current behavior?**
The cache location is mapped relative to the user home directory across OSs. 

**If the current behavior is a bug, please provide the steps to reproduce.**
NA

**What is the expected behavior?**
To be able to specify by an environment variable or cli setting a yarn config path.  For many docker users, this would allow a mounted location to be used for the cache, or to specify a network path for a shared cache in automated build systems, etc.

**Please mention your node.js, yarn and operating system version.**
Ubuntu 16.04 (docker), node version 6.1.0, yarn version 0.16.0.

mmerickel:
I'm experiencing the same problem. I wanted to set the cache folder location via `yarn config` but looking at the `Config.init(...)` bootstrapping process, it's only possible to set it from the `--cache-folder` option on the cli at this time.

samccone:
Thanks for the issue @ryanmt and follow up @mmerickel 
This would make for a really great first PR into yarn.

Thanks

bengummer:
This would be really handy for running yarn inside AWS lambda. The user home directory on lambda is read-only which causes a lambda fs permission error to be thrown at https://github.com/yarnpkg/yarn/blob/553e55f146a465a4e773b53b0a93752163c4b039/bin/yarn.js#L10

rattrayalex:
My usecase is docker; during development, I want docker to read from (and write to) a project-specific cache which I can access by running `yarn add` within my project on the host machine (my laptop).

victornoel:
@ryanmt no PR? :)

valscion:
I suppose that yarn 0.18.0 released with PR #2046 makes this issue fixed, and this could be closed?

wyze:
This landed in v0.18.0.

nervo:
I'm sorry, but i don't get it :)
Is there somewhere a way to pass the yarn config dir as an environment variable ?
I don't see such a thing in https://github.com/yarnpkg/yarn/pull/2046

valscion:
> I'm sorry, but i don't get it :)
> Is there somewhere a way to pass the yarn config dir as an environment variable ?
> I don't see such a thing in #2046 

The original PR description had this text in it:

> To be able to specify by an environment variable or cli setting a yarn config path

So I'd say this requirement is fulfilled and thus this issue was closed as fixed.

[...]

PR #2046, opened 2016-11-27
add cache-folder config option for setting cache folder location

superxp1412:
**Summary**

Yarn need a elegant way to config cache folder's directory.   
This pull request is an answer to: [#1316](https://github.com/yarnpkg/yarn/issues/1316) & [#1211](https://github.com/yarnpkg/yarn/issues/1211)

With this feature we can define the location of yarn's cache folder:
e.g.
`yarn config set cache-folder ${path}` 

And the settings would be defined inside of .yarnrc file like:
`cache-folder "${path}"`

Besides, new configurations for cache-folder can also been verified by command:
`yarn cache dir`

@samccone


superxp1412:
Yes, I'll try to add test for this feature.

bestander:
@superxp1412, you should just use `install` tests to verify where the cache is coming from https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install.js#L726.
You can pass options and flags to the install command to configure where the cache is stored.

src/config.js
```diff
@@ -203,6 +200,12 @@ export default class Config {
       cert: String(opts.cert || this.getOption('cert') || ''),
       key: String(opts.key || this.getOption('key') || ''),
     });
+
+    //init & create cacheFolder, tempFolder
+    this.cacheFolder = String(opts.cacheFolder || this.getOption('cache-folder') || constants.MODULE_CACHE_DIRECTORY);
```
bestander: The only test needed is that all 3 options may setup cache folder location and they have different priority

superxp1412: I've already submitted the unit test for the priorities for the options. Please review it.
@bestander

bestander:
@superxp1412 any idea why tests failed?
Maybe Circle/Windows does not have access to the config folder.

superxp1412:
I think the test fails may caused by concurrently read/write .yarnrc file. There's a chance to reproduce that issue. 
Now, I've fixed it by changing the test running methods been run in sequence.
@bestander

mvestergaard:
This seems fairly broken atm.
This line in `src/config.js@205`: `this.cacheFolder = String(opts.cacheFolder || this.getOption('cache-folder') || constants.MODULE_CACHE_DIRECTORY);`

Sets `cacheFolder` to `set_config_folder_dir` because of `this.getOption('cache-folder')`. I'm not entirely sure where that value comes from, but it ends up creating that folder in your working directory, and using that folder as cache.

mvestergaard:
It also creates a `flag_config_folder_dir` inside your yarn working directory whenever you run tests.

superxp1412:
Hi @mvestergaard , @bestander 
Yes, the tests is not sandbox well. 
I promise to fix it by another pull request soon.

bestander:
@mvestergaard, could you raise an issue to track down your problem and ping me?
Can you send a PR with a fix for your situation?

mvestergaard:
@bestander sure i can add an issue if you'd like. I already created the PR #2144 with a fix

bestander:
Thanks, no need for issue if the PR is there :)

buschtoens:
I know now it's to late, but `folder` is a rather unconventional name. It really should have been `dir` or `directory`.

https://superuser.com/questions/169457/directory-vs-folder

bestander:
@buschtoens, we use word "folder" in 85 places, so this might be a big change.
If you feel strong about it and willing to do a codemod, please send an RFC, we can discuss the pros and cons.

buschtoens:
Thanks for taking note, @bestander! :smile: 

I primarily commented to vent for a bit, as I passed `--cache-dir` instead of `--cache-folder`. Sorry. :sweat_smile: 

Since this is already implemented, taught and in people's heads, changing it probably won't be worth the trouble.

bestander:
I totally understand your frustration.
How about all the possible ways to clear cache in a tool: --no-cache,
--clear-cache, clean, cache clean, --reset-cache?

PR #2052, opened 2016-11-28
Add --frozen-lockfile to fail if Lockfile needs update

lawliet89:
**Summary**

Adds the flag `--frozen-lockfile` to `yarn install` so that `yarn` will fail if it detects that the lockfile requires an update. This is similar to [bundler's `--frozen` flag](https://bundler.io/v1.13/man/bundle-install.1.html).

**Test plan**

Consider the `package.json` and `yarn.lock` in the fixture I have added in this PR. 

```
$ yarn --frozen-lockfile
yarn install v0.18.0-0
info No lockfile found.
[1/4] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`
Done in 6.28s.

```

But if you generate a lockfile and re-run `yarn`, it should install fine.


src/reporters/lang/en.js
```diff
@@ -88,6 +88,7 @@ const messages = {
   jsonError: 'Error parsing JSON at $0, $1.',
   noFilePermission: "We don't have permissions to touch the file $0.",
   allDependenciesUpToDate: 'All of your dependencies are up to date.',
+  frozenLockfileError: 'Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`',
```
kittens: run -> ran

kittens: Add a dot to the end of this as well since it's reporting an action.

lawliet89: [The past participle of run is run.](https://www.quora.com/Which-is-more-grammatically-correct-was-run-or-was-ran) I will add the period.

mmun:
@lawliet89 Thanks for implementing this feature! Looking forward to using it on our CI to catch when people forget commit their yarn.lock, or if they added a dependency with `npm install` instead of `yarn add`.

bestander:
@lawliet89, could you rebase please?

lawliet89:
@bestander I have rebased.

PR #2053, opened 2016-11-28
Nested executables fix

onemen:
Implementing 'Nested executables fix' as suggested by #1210
* Use cwd instead of package directory when linking bin scripts
* Update package-linker module to filter calls to linkBin function
(prevents #1823)
* Update global module to only link root packages to the global bin
folder
* Added tests


torifat:
@onemen I have re-ran you test. It's passing now 👍 

This is quite a big PR. Might take some time to review.

bestander:
@onemen, thanks for the effort to add all those tests! 👍 
Could you rebase please?

onemen:
@bestander,

I will do the rebase later this week.
I want to add test, to make sure that the bins files are actually usable.
That every module can find the bins of its dependency.

Where can i found simple example for this?

src/package-linker.js
```diff
@@ -91,18 +98,68 @@ export default class PackageLinker {
...
+          }
+        }
+
+        // When both package and dependency are in the same folder use the .bin
+        // in that folder, else use the .bin in the package.
+        const location = newDepLoc || loc;
+        const binLoc = path.dirname(location) === pkgLoc ? parentBinLoc : pkgBinLoc;
+        const binLinks = await getBinLinks(binLoc);
+        // Remove Duplicates
+        if (!binLinks.has(location)) {
```
bestander: I think top level dependencies should have a priority over nested ones

__tests__/commands/install/integration.js
```diff
@@ -452,15 +455,39 @@ test.concurrent(
...
+    fixturesLoc,
+    async (args, flags, config, reporter, lockfile): Promise<Install> => {
+      const install = new Install(flags, config, reporter, lockfile);
+      const linkBin = install.linker.linkBin;
+      install.linker.linkBin = mockLinkBin = jest.fn((...args) => {
+        return linkBin(...args);
+      });
+      await install.init();
+      await check(config, reporter, {}, []);
+      return install;
```
bestander: this test needs to verify that bin dependencies were flattened properly.
Especially for a case of conflicting bins

src/package-linker.js
onemen: I can fix the sort at the top of copyModules function, let me know if it ok with you.
After this sort, we populate binLinks with top level dependencies first.
```
    // sorted tree makes file creation and copying not to interfere with each other
+   // prioritize top level dependencies
    flatTree = flatTree.sort(function(dep1, dep2): number {
-     return dep1[0].localeCompare(dep2[0]);
+     const [len1, len2] = [dep1[0].split(path.sep).length, dep2[0].split(path.sep).length];
+     return len1 === len2 ? dep1[0].localeCompare(dep2[0]) : len1 - len2;
    });
```

bestander: If that works then sure.
I thought it would be easier though if you iterate over the top dependencies first to generate a set of bins and then go through deeper dependencies discarding bins that are already in the set.

skevy:
@onemen any updates on finishing up this PR?

onemen:
I hope to find the time later this month.

I need to re-base the code and create new branch, since my old branch was corrupted

thomasruiz:
@onemen any update?

bestander:
ping.
Would be great to get this in, otherwise let's close the PR.

bestander:
I'll close this PR for now.
Feel free to send a new one.

PR #1455, opened 2016-10-25
Add detection of NODE_ENV for `yarn install` command

wyze:
**Summary**

This fixes #1346. Now the install command will respect the NODE_ENV flag. The [docs](https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-production) mention:

> Using the --production flag, **or when the NODE_ENV environment variable is set to production**, Yarn will not install any package listed in devDependencies.

*Note: Feels kind of weird checking in the `.bin` files in request-cache, but I heard that is what we should do.*

**Test plan**

```sh
$ cat package.json
{
  "name": "yarn-example",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "is-negative-zero": "^2.0.0"
  },
  "dependencies": {
    "left-pad": "^1.1.3"
  }
}
```

##### Before

```sh
$ NODE_ENV=production yarn
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 0.48s.

$ ls node_modules
is-negative-zero left-pad
```

##### After

````sh
$ NODE_ENV=production yarn
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 0.62s.

$ ls node_modules
left-pad
```

PR #2057, opened 2016-11-28
Allow 'production' ENV to take precedence over NODE_ENV

### LINKED ISSUE #1975: devDependencies are not installed when both `NODE_ENV=production` and `NPM_CONFIG_PRODUCTION=false` are set
bug

`NODE_ENV` seems to be taking precedence over `NPM_CONFIG_PRODUCTION` (and `YARN_PRODUCTION`).

In npm, it's the opposite.

```bash
$ yarn init -y
$ yarn add Base64
$ yarn add --dev blob
$ rm -rf node_modules
$ NODE_ENV=production NPM_CONFIG_PRODUCTION=false yarn
```

This doesn't install `blob`.

It should install `devDependencies` like npm does:

```bash
$ yarn init -y
$ yarn add Base64
$ yarn add --dev blob
$ rm -rf node_modules
$ NODE_ENV=production NPM_CONFIG_PRODUCTION=false npm install
```

This *does* install `blob`.

```bash
$ node -v
v6.3.1

$ npm -v
3.10.3

$ yarn -V
0.17.8
```

Windows 10

joeyespo:
This fixes #1975.

The motivation is to fix the compatibility issue so that Yarn can be used in place of npm on Heroku ([related buildpack issue](https://github.com/heroku/heroku-buildpack-nodejs/issues/337)). In this case, when you need to install devDependencies, [as documented here](https://devcenter.heroku.com/articles/nodejs-support#devdependencies).

**Test plan**

A test is included.

(I also tested locally on my setup and observed that the devDependencies were indeed installed to `node_modules` when both `NODE_ENV=production` and `NPM_CONFIG_PRODUCTION=false` were set.)

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -403,6 +403,20 @@ test('install should respect NODE_ENV=production', (): Promise<void> => {
   });
 });
 
+// don't run this test in `concurrent`, it will affect other tests
+test('install should respect NPM_CONFIG_PRODUCTION=false over NODE_ENV=production', (): Promise<void> => {
+  const env = process.env.NODE_ENV;
+  const prod = process.env.NPM_CONFIG_PRODUCTION;
+  process.env.NODE_ENV = 'production';
+  process.env.NPM_CONFIG_PRODUCTION = 'false';
+  return runInstall({}, 'install-should-respect-npm_config_production', async (config) => {
+    expect(await fs.exists(path.join(config.cwd, 'node_modules/is-negative-zero/package.json'))).toBe(true);
+    // restore env
+    process.env.NODE_ENV = env;
+    process.env.NPM_CONFIG_PRODUCTION = prod;
+  });
+});
+
 test.concurrent('install should resolve circular dependencies 2', (): Promise<void> => {
   return runInstall({}, 'install-should-circumvent-circular-dependencies-2', async (config, reporter) => {
     assert.equal(
--- a/__tests__/fixtures/install/install-should-respect-npm_config_production/package.json
+++ b/__tests__/fixtures/install/install-should-respect-npm_config_production/package.json
@@ -1,0 +1,8 @@
+{
+  "dependencies": {
+    "left-pad": "1.0.0"
+  },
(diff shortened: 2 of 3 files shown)


joeyespo:
Huh. The test is passing locally. Digging in to see why it isn't working on CI.

joeyespo:
I'm not sure why my first approach wasn't working. It worked locally for some reason. (Anyone have any ideas?) But using the same method as #1455 worked.

This is all good to go.

bestander:
@joeyespo, do we have a support for `--no-production`?
Should it override ENV vars?
+ needs a rebase

joeyespo:
@bestander Rebased.

One test is failing due to `Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL`. Any ideas?

There's no `--no-production` to my knowledge, but this does work exactly like npm where you can set `NPM_CONFIG_PRODUCTION=false` in your config.

bestander:
@joeyespo, this is a known problem, we'll sort it out separately

joeyespo:
@bestander 

> and overriding with an arg can be a more obvious API

I agree with you on the usefulness of that. Yarn currently doesn't support that. (It appears npm does.)

This PR fixes real production compatibility issue though (https://github.com/heroku/heroku-buildpack-nodejs/issues/337). As of #1455, you can already set `NODE_ENV` to `production` without any way to override it. Heroku sets this, and [documents that the way to override it is with another ENV](https://devcenter.heroku.com/articles/nodejs-support#devdependencies). This PR fixes that specific case.

The `--production=false` case is another backwards incompatibility. It's independent of this fix though and may require some upfront discussion, So I think it might be better to address that in a new issue/PR instead of ultimately delaying this one.

chadly:
related #2167

joeyespo:
@chadly Perfect 👍 

@bestander Given that the test timeout is a known thing to be sorted out separately, this should be ready to go. Need anything else from me?

bestander:
Sounds reasonable, @joeyespo, thanks for the contribution!

PR #2068, opened 2016-11-29
Add --only-dev option for install (#998)

### LINKED ISSUE #998: Feature request: Only install devDependencies

Feature

`dependencies` are always installed and there is no command to prevent that.

Similar to `npm install --only=dev` a CLI option to only install devDependencies.

(My) usecase: Compile only environments where the resulting build will be used in a docker container (with a separate `yarn install`)

coreyc:
**Summary**
This closes #998. It adds a flag to only install devDependencies (`yarn install --only-dev`).

**Test plan**
Unit test added to check that only devDependencies are installed and not dependencies (from a given package.json).

Add --only-dev option for install


torifat:
This isn't a correct fix. You can read why here: https://github.com/yarnpkg/yarn/pull/2044#issuecomment-263275725

jonschlinkert:
just a though, but if this kind of flag is implemented it would make a lot more sense to have the "root" flag first, and the option last, e.g. `--dev-only`, rather than `--only-dev`. This way it would sort in documentation together and be easier to remember. just my 2c

coreyc:
@jonschlinkert npm uses `--only-dev` (actually it's `--only=dev` and `--only=prod`) which is why I kept it so.  Not opposed to making it `--dev-only`.  There is a --dev flag for add so I thought it might get confusing between `--dev-only` and `--dev`.

Looks like there was quite a bit of refactoring in #2116.  Think I'll need to wait for that to get merged and then make some modifications to my PR.

kittens:
I've merged #2116, the refactoring should make it easier. The relevant lines are [here](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L259-L261). `pushDeps` needs to be called for all three dependency types. Change the third `isUsed` argument of `pushDeps` to filter these from inclusion.

coreyc:
Thanks @kittens this does make it simpler.  I noticed you removed some of the imports (i.e. - Config, reporters) in the install integration test file, which actually brought up something I wanted to clarify.  In my original commit, the test I wrote creates a new instance of `Install` rather than the `runInstall()` helper function.  Using `runInstall()` was throwing errors such as `error {dependencies package name} not installed`.  I think this is because `runInstall()` is also performing a check, at which point it fails because the flags object is reset by then.  I had assumed this was ok because running `yarn check` after installing only devDependencies would (I assumed) rightly fail.  

I want to make sure that my original test is not giving a false positive - should `runInstall()` be the function used here?  If so, I'd likely need to add the onlyDev attribute to config as well, similar to the production attribute.

Please let me know if this is unclear

coreyc:
woops, had a merge issue here and accidentally messed this up.

moimael:
Any progress on this ?

coreyc:
@moimael have been very busy and not able to work on this, feel free to take it up. note my dec 5th comment. will close PR for now.

Issue #1893, opened 2016-11-16
switching from npm dep to a github dep requires a cache clean


tamird:
**What is the current behavior?**

Given the following `package.json`:
```
{
  "dependencies": {
    "jspm": "^0.16.33"
  }
}
```
running `yarn install` correctly installs `jspm` into `node_modules`. However, changing the `package.json` to:
```
{
  "dependencies": {
    "jspm": "tamird/jspm-cli#dont-rewrite-package-json"
  }
}
```
and re-running `yarn install` does not change the contents of `node_modules` (it does update `yarn.lock`. Running `yarn cache clear && yarn install` does correctly update the contents of `node_modules`.

```
$ node --version
v7.0.0
$ yarn --version
0.16.1
$ uname -a
Darwin Tamirs-MBP 16.1.0 Darwin Kernel Version 16.1.0: Thu Oct 13 21:26:57 PDT 2016; root:xnu-3789.21.3~60/RELEASE_X86_64 x86_64
```

markstos:
This may be a dupe of one of the other Github/caching issues already filed:

  * #1171
  * #1453
  * #1573

@tamird Could you look through other Github/caching issues including the above but possibly others (there are still more!) and confirm if you've really found a unique issue or if this is the same as one of the others? If it's a dupe, please close this copy of the issue, thanks.

tamird:
It's possible that those issues are related, but none of them are direct
duplicates of this one (i.e. the root cause may be the same but the symptom
is different).

bestander:
Should be fixed now #2074

PR #2074, opened 2016-11-29
Add hash to cache path for non-NPM packages

mourner:
**Summary**

Currently, only packages resolved to NPM registry have their hash appended to cache directory name. Other packages, such as those with GitHub URL, get cached by their `name` + `version` rather than actual content. This leads to a bug where even if you update the GitHub url in `package.json` (e.g. different GitHub commit hash), it is never updated on `yarn install` unless you clean the cached folders for the package along with `node_modules`. This is a huge pain for teams that depend on GitHub urls in dependencies.

Should address #1171, #1893, #1087, #1280, #1278, #1017 #1573 #2280. I'm not sure it fully fixes all the mentioned issues, so it's a work in progress in need of a discussion and thorough testing, but it should be a step in the right direction.

**Test plan**

This is my first dive into Yarn codebase and I haven't yet figured out a way to test this — currently `npm test` fails for me locally on `master` with "process exited unexpectedly error" in `commands/remove.js` without much detail.

cc @kittens @jfirebaugh @jordanburke @ronag @wmertens @tamird @mgutz @panrafal @omnidan

### DIFF
--- a/__tests__/package-resolver.js
+++ b/__tests__/package-resolver.js
@@ -12,6 +12,9 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
 
 const path = require('path');
 
+// regexp which verifies that cache path contains semver + hash
+const cachePathRe = /-\d+\.\d+\.\d+-[\dabcdef]{40}$/;
+
 function addTest(pattern, registry = 'npm') {
   // concurrently network requests tend to stall
   test(`resolve ${pattern}`, async () => {
@@ -30,6 +33,11 @@ function addTest(pattern, registry = 'npm') {
     });
     const resolver = new PackageResolver(config, lockfile);
     await resolver.init([{pattern, registry}]);
+
+    const ref = resolver.getPackageReferences()[0];
+    const cachePath = config.generateHardModulePath(ref, true);
+    expect(cachePath).toMatch(cachePathRe);
+
     await reporter.close();
   });
 }
--- a/src/config.js
+++ b/src/config.js
@@ -295,11 +295,14 @@ export default class Config {
     let uid = pkg.uid;
     if (pkg.registry) {
       name = `${pkg.registry}-${name}`;
-      uid = pkg.version || uid;
     }
 
     const {hash} = pkg.remote;
-    if (hash) {
+
+    if (pkg.version && (pkg.version !== pkg.uid)) {
+      uid = `${pkg.version}-${uid}`;
+
+    } else if (hash) {
       uid += `-${hash}`;
     }
 


kittens:
Where is `info.hash` coming from? This is the package manifest.

mourner:
@kittens it's saved [to the metadata file here](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/fetchers/base-fetcher.js#L58). The only occasion I found where `info.hash` is not present is when the file resolver is used, but that should be OK.

bestander:
@mourner, could you add a test for the new behavior and  the issues it addresses, please?

mourner:
Sure, I just wanted to get some early feedback before I start working on tests. Will do.

bestander:
It does make some sense, we have some confusion in the system where the hashes come from.
Let's see how it works with a few tests, looks like it does not break any existing ones

benmosher:
@mourner want any help with tests?

mourner:
@benmosher sure! I was going to take a look next week, but if you're willing to help out with the PR, you're very welcome!

mourner:
@bestander @kittens updated the PR with a simpler fix. It leaves `npm-resolver` cache paths the same while adding hashes to `tarball` and `hosted-git` (and derived) resolvers, fixing numerious issues with different commits of the same package version colliding in the cache folder.

Note that the test failures are the same as the ones in the current master, for which I'm not responsible.

I'm still not able to add tests for the change though because package resolver tests were disabled by @kittens in https://github.com/yarnpkg/yarn/commit/b59fd78d8580765fa6d7aef5d492d90383922727 and I don't have enough knowledge to reenable them and fix the failures (unrelated to the changes in this PR).

bestander:
@mourner, I've enabled most of the tests in https://github.com/yarnpkg/yarn/pull/2373.
Also master branch should be more stable now, could you please rebase?

mourner:
@bestander thanks! Rebased on top of master. I'll wait for #2373 to land before attempting to add any tests.

@benmosher are you still up for help on the tests?

mourner:
@bestander finally added some tests based on the reenabled resolver tests! Hopefully we can land this in a Yarn release soon.

Also, I used this Yarn fork locally and it worked perfectly when doing a difficult git bisect in a project with tons of git-hashed dependencies ([mapbox-gl-js](https://github.com/mapbox/mapbox-gl-js)), whereas it would be impossible in both stable Yarn (due to the bug) and NPM (due to unbearable install times).

mourner:
@jordanburke @ronag @wmertens @tamird @mgutz @panrafal @omnidan hey, could you please confirm that this PR fixes your corresponding opened issues (listed above) so that we could close them when this PR lands?

mourner:
Found another duplicate for the problem: #2280, cc @redgenie ^^^

bestander:
@mourner, great job!
Thanks for making an extra effort with the test!

mourner:
Yay, thanks for merging!

tamird:
I believe this has regressed; currently a github dependency is not updated on disk after a `yarn upgrade` without a cache clean.

mourner:
@tamird I think `yarn upgrade` behavior would be a different issue. This one is about linking wrong versions from cache (other than the ones specified in yarn.lock).

tamird:
Yes, `yarn upgrade` does the right thing, but the subsequent `yarn install` continues to use the old version (not the one which is now specified in `yarn.lock`.

mourner:
@tamird can you check if the relevant resolver tests still pass?

tamird:
What do you mean? Does yarn not have CI?

mourner:
@tamird it does but it's quite useless at the moment because all recent master commits have failed builds, so I don't know whether the resolver tests actually caught any regression or not. 😕

tamird:
Yes, it seems the resolver tests do pass.

SudoPlz:
I have the same problem on `yarn 0.17.10`

PR #2082, opened 2016-11-30
Fix scripts cmd slashes for windows

whitecolor:
**Summary**
This address the issue https://github.com/yarnpkg/yarn/issues/1729 which as about that on windows yarn fails to execute scripts that contain commands with unix-style slashes, for example this will fail on windows:
```json
"scripts": {
  "compile": "../node_modules/.bin/tsc"
}
```

So to run such script on windows machine successfully we need to replace slashes from `/` to `\`, 

```json
  "compile": "..\\node_modules\\.bin\\tsc"
```

For we need to find all the executable commands in the script line, so we assume that exec commands that need to be fixed:
- Are located in the beginning of the line
- Or located after symbols: `&&` or `&` or `||` or `|`
- `&&` or `&` or `||` or `|` symbols after which command go may not be inside the quotes. In opposite case this will mean that it is inside the nested command like:
	`docker run container bash -c "cat some.thing | some/bin/here"`
- Can be inside of single or double quotes (may contain spaces)
- Can be without quotes (no spaces)
- In the substrings that satisfies upper criteria all backslashes `/` are replaced with windows slashes `\`

**Test plan**

Added test that check correct transformation for different command cases.

So this need some triage and opinions.

### DIFF
--- a/__tests__/util/fix-cmd-win-slashes.js
+++ b/__tests__/util/fix-cmd-win-slashes.js
@@ -1,0 +1,60 @@
+/* @flow */
+import {fixCmdWinSlashes} from '../../src/util/fix-cmd-win-slashes.js';
+
+const cmdCases = [
+  [
+    'fixes just slashed command',
+    'some/command',
+    'some\\command',
+  ],
+  [
+    'fixes slashed command in double quotes',
+    '"./some/command/more"',
+    '".\\some\\command\\more"',
+  ],
+  [
+    'slashed command in single quotes',
+    '\'./some/command/more\'',
+    '\'.\\some\\command\\more\'',
+  ],
+  [
+    'fixes slashed command with slashed param',
+    './some/command/more slashed/param',
+    '.\\some\\command\\more slashed/param',
+  ],
+  [
+    'fixes slashed command in single quotes with two slashed params',
+    '\'./some/command/more\' justparam slashed/param --param slashed/param2',
+    '\'.\\some\\command\\more\' justparam slashed/param --param slashed/param2',
+  ],
+  [
+    'fixes two slashed commands (&&) with two slashed params',
+    '\'./some/command/more\' justparam slashed/param &&  ../another/command with/slashed/param',
+    '\'.\\some\\command\\more\' justparam slashed/param &&  ..\\another\\command with/slashed/param',
+  ],
+  [
+    'fixes three slashed commands (&&) with two slashed params',
+    '\'./some/command/more\' justparam slashed/param &&' + 
+      '  "../another/command" with/slashed/param | ./another/more/command with/slashed/param',
+    '\'.\\some\\command\\more\' justparam slashed/param &&' + 
+      '  "..\\another\\command" with/slashed/param | .\\another\\more\\command with/slashed/param',
+  ],
+  [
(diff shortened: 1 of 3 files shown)


Daniel15:
Internally, Windows actually supports both `\` and `/`, so the original path should be fine. For example, try running `c:/windows/system32/calc.exe` in either the run dialog or a command prompt - It'll work fine. Because of this, I'm wondering if the problem is actually elsewhere. I feel like we shouldn't need to do this.

Does it fail in npm too?

Daniel15:
Ah I see, it only fails when there's a dot at the start 
```
C:\Windows>./system32/calc.exe
'.' is not recognized as an internal or external command, operable program or batch file.
```

Absolute paths work fine. I wonder if we could resolve the path before running the command. Node.js should have a built-in path method for that.

whitecolor:
Yes this actually fails for relative paths, and as mostly exec paths in scripts are relative. Yes NPM has the [same the old issue](https://github.com/npm/npm/issues/13789). Windows users seem to have a big tolerance (which is not very good).

So I think it should be fixed in a proper way.  Windows people [are complaining in different repos](https://github.com/cyclejs/cyclejs/issues/481), but it is still not fixed in years.

Daniel15:
This diff is pretty complex, but the tests seem comprehensive. Let's ship it!

ctaggart:
I just tested on Windows with the latest nightly build and this is still not working for me:

```
yarn install v0.20.0-20170107.0149
info No lockfile found.
[1/4] Resolving packages...
warning mocha > glob > minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning mocha > to-iso-string@0.0.2: to-iso-string has been deprecated, use @segment/to-iso-string instead.
warning mocha > jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
[2/4] Fetching packages...
warning fsevents@1.0.17: The platform "win32" is incompatible with this module.
info "fsevents@1.0.17" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
success Saved lockfile.
$ ./node_modules/.bin/cpx ./node_modules/fixed-data-table/dist/*.css ../Assets/fixed-data-table
'.' is not recognized as an internal or external command,
```
It just needs to run the `cpx` command.
image

that is in `packages.json` like so:
```
  "scripts": {
    "postinstall": "./node_modules/.bin/cpx ./node_modules/fixed-data-table/dist/*.css ../Assets/fixed-data-table",
```

Daniel15:
@ctaggart  - You should be able to omit the `node_modules/.bin` from scripts. Try this:

```
"postinstall": "cpx ./node_modules/fixed-data-table/dist/*.css ../Assets/fixed-data-table"
```

ctaggart:
@Daniel15 Thank you! That worked great with nightly v0.20.0-20170107.0149 and the latest stable release 0.18.1.

Daniel15:
😃  By the way, npm has the same behaviour - It'll look in `node_modules/.bin` by default, so you don't need to explicitly include it in your command

whitecolor:
Btw I would recommend on windows 10 to completely migrate development process to "Ubuntu bash for windows".

PR #2092, opened 2016-11-30
Fixes race condition on yarn install mutex network

kentaromiura:
**Summary**

Fixes a race condition we observed in production, to easily reproduce it, I changed the code in `onServerEnd` to wait any amount of time before resolving:
```js
const onServerEnd = (): Promise<void> => {
  clients.forEach((client) => {
    client.write('closing. kthanx, bye.');
  });
  return new Promise(ok => {
    setTimeout( () => {
      server.close();
      ok();
    }, 20);
  }
};
```

This change makes it easier to spot the issue
**Test plan**
run:
```
./bin/yarn install --mutex network& ./bin/yarn install --mutex network&./bin/yarn install --mutex network&
```

IIRC The original code was written like that because I observed some issue in relying on the `close` event on my original implementation that was using a unix socket.
Since then I changed the strategy and we use the network.
I cannot recreate the problem observed with this implementation.
I believe using the correct events is also much cleaner :)

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -277,7 +277,8 @@ const runEventuallyWithNetwork = (mutexPort: ?string): Promise<void> => {
 
       socket
         .on('connect', () => {
-          socket.unref(); // let it die
+          // Allow the program to exit if this is the only active server in the event system.
+          socket.unref();
         })
         .on('close', (hadError?: boolean) => {
           // the `close` event gets always called after the `error` event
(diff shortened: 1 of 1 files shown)


src/cli/index.js
```diff
@@ -268,36 +268,34 @@ const runEventuallyWithNetwork = (mutexPort: ?string): Promise<void> => {
...
-      // another yarnn instance exists, let's connect to it to know when it dies.
+      // another Yarn instance exists, let's connect to it to know when it dies.
       reporter.warn(reporter.lang('waitingInstance'));
       const socket = net.createConnection(connectionOptions);
 
       socket
-        .on('data', () => {
-          // the server has informed us he's going to die soon™.
+        .on('connect', () => {
           socket.unref(); // let it die
```
bestander: Why do you socket.unref right after connecting?

Issue #346, opened 2016-09-16
"Bad hash" error when using kpm with tarball cache in Docker


skevy:
Note: I don't know what's causing this, and am in the middle of fixing a critical bug in our code so can't investigate deeper at the moment...will try to later.

I did a typical `kpm i` (on my mac) with a node_module cache that's stored in our monorepo. Saw node_modules get updated, etc. In my Docker image, I am using a docker volume to be able to see my node_module cache inside the image when I do a `kpm i`. Running `kpm i` on my mac works fine, but running `kpm i` inside the image (with the same tarball cache) returns:

```
error Bad hash. Expected 208df8fb8bbc74acb7d3e41568249eba0dab55f7 but got ed74b477a2ee9369f75efee2f08d5915e52a42e8
    at SecurityError (/root/fbkpm/lib/errors.js:17:1)
    at Extract.tar.Extract.on.on (/root/fbkpm/lib/fetchers/tarball-fetcher.js:123:16)
    at emitNone (events.js:91:20)
    at Extract.emit (events.js:185:7)
    at DirWriter.<anonymous> (/root/fbkpm/node_modules/tar/lib/extract.js:82:8)
    at emitNone (events.js:91:20)
    at DirWriter.emit (events.js:185:7)
    at end (/root/fbkpm/node_modules/fstream/lib/writer.js:324:14)
    at /root/fbkpm/node_modules/fstream/lib/writer.js:314:34
    at endUtimes (/root/fbkpm/node_modules/fstream/lib/writer.js:238:46)
```

I'm not sure why a SHA1 hash would be different between OS's, and fwiw this problem just started happening after I updated from 0.12.0 to 0.13.0.

Curious if anyone has thoughts here, or if I'm doing something stupid or missing something obvious.

alansouzati:
I'm still facing this issue with both 0.16.1 and 0.17.2. Any ideas what to do to fix it ?

alansouzati:
My problem was that I had manually copied the yarn.lock with my `COPY` command in docker. I added the `yarn.lock` to my .dockerignore and everything works fine now.

SimenB:
Hitting this now, using 0.17.8.

Why would I not copy over `yarn.lock`, I want to ensure I get the same deps?

SimenB:
Hmm, seems to have been a caching error, blowing away the cache on travis (containing `node_modules`) seems to have fixed it

PR #2095, opened 2016-11-30
Add try/catch when writing yarn-error.log

### LINKED ISSUE #2094: Yarn hangs forever when missing write access
Bug

If yarn need to write in a read-only directory, it hangs forever.

```bash
mkdir -m 500 bar && cd bar
yarn init -y
```
Then yarn says it successfully has written package.json, but just hangs forever.

For yarn to report the missing write access and exit.

node v6.9.1; yarn from source; official node docker image.

androa:
**Summary**
If yarn is lacking write access, the `fs.writeFileSync()` method throws an
uncaught exception that blocks further execution.

Simply catching it and allowing further exection provides desired behaviour by
giving the user feedback and having yarn terminate.

**Test plan**
```bash
mkdir -m 500 bar && cd bar
yarn init -y
```
now results in 
```bash
node@062ea1ba7c46:~/bar$ /yarn/bin/yarn init -y
yarn init v0.18.0-0
warning The yes flag has been set. This will automatically answer yes to all questions which may have security implications.
success Saved package.json
error An unexpected error occurred: "EACCES: permission denied, open '/home/node/bar/package.json'".
info If you think this is a bug, please open a bug report with the information provided in "/home/node/bar/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/init for documentation about this command.
```
Previously it just kept hanging forever.

Fixes #2094

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -339,7 +339,9 @@ function onUnexpectedError(err: Error) {
   log.push(`Trace: ${indent(err.stack)}`);
 
   const errorLoc = path.join(config.cwd, 'yarn-error.log');
-  fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
+  try {
+    fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
+  } catch (err) {}
 
   reporter.error(reporter.lang('unexpectedError', err.message));
   reporter.info(reporter.lang('bugReport', errorLoc));
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -341,10 +341,16 @@ function onUnexpectedError(err: Error) {
   const errorLoc = path.join(config.cwd, 'yarn-error.log');
   try {
     fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
-  } catch (err) {}
 
-  reporter.error(reporter.lang('unexpectedError', err.message));
-  reporter.info(reporter.lang('bugReport', errorLoc));
+    reporter.error(reporter.lang('unexpectedError', err.message));
+    reporter.info(reporter.lang('bugReport', errorLoc));
+  } catch (err) {
+    if (err.code === 'EACCES') {
+      reporter.error(reporter.lang('noFilePermission', err.path));
+    } else {
+      console.error(err);
+    }
+  }
 }
 
 //


src/cli/index.js
```diff
@@ -339,7 +339,9 @@ function onUnexpectedError(err: Error) {
   log.push(`Trace: ${indent(err.stack)}`);
 
   const errorLoc = path.join(config.cwd, 'yarn-error.log');
-  fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
+  try {
+    fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
+  } catch (err) {}
```
Daniel15: This should probably at least `console.error` or something rather than silently ignoring it.

Daniel15: Maybe something like `reporter.error('Could not write yarn-error.log: ' + err)` (with proper localized string)

androa: The thing is, [#346](https://github.com/yarnpkg/yarn/pull/2095/files/8ffa389e437d7a2f3aaedb7804765a131c53bed9#diff-867becf4a9c2c6c6d4e7c1278750724eR346) already handles it marvelously. It just needs to get there. I was thinking to rather piggyback on that one, rather than handling it explicitly. But I can of course change into doing so.

src/cli/index.js
```diff
@@ -341,10 +341,16 @@ function onUnexpectedError(err: Error) {
...
     fs.writeFileSync(errorLoc, log.join('\n\n') + '\n');
-  } catch (err) {}
 
-  reporter.error(reporter.lang('unexpectedError', err.message));
-  reporter.info(reporter.lang('bugReport', errorLoc));
+    reporter.error(reporter.lang('unexpectedError', err.message));
+    reporter.info(reporter.lang('bugReport', errorLoc));
+  } catch (err) {
+    if (err.code === 'EACCES') {
+      reporter.error(reporter.lang('noFilePermission', err.path));
```
androa: @Daniel15 About doing this instead then?

bestander: We don't use console.error anywhere, better to keep to reporter.error anyway.
How about you wrap only fs.writeFileSync into try/catch?

androa:
Seems to have been solved by #1592 and this can be closed.

PR #1837, opened 2016-11-15
Set UID and GID of untared files to zero if user is root

lawliet89:
**Summary**

Fixes #1750 

`TarballFetcher` uses [`tar`](https://github.com/npm/node-tar) to untar tarballs. `tar` in turn constructs a [`Writer`](https://github.com/npm/fstream/blob/24fabdec32e334dd3b130d77b38c010e3119b102/lib/writer.js#L26) object per entry in the tarball, preserving the original UID and GID in the tarball.

If `Writer` detects that the UID of the user is 0 (i.e. root user), it will attempt to [`chown`](https://github.com/npm/fstream/blob/24fabdec32e334dd3b130d77b38c010e3119b102/lib/writer.js#L195) the extracted file to the UID. 

Usually, this is not a problem. However, as in #1750 and other issues, if the _`root`_ user is not really root but inside a [user namespace](http://man7.org/linux/man-pages/man7/user_namespaces.7.html), and the UID is outside the [`subuid`](http://man7.org/linux/man-pages/man5/subuid.5.html) range of the mapped user, then it will fail. This is the same issue for GID.

This PR attempts to just set everything to UID 0 if the user is detected as root, regardless of whether the root is inside a namespace or not.

**Test plan**

First, configure Docker Daemon with [user namespace enabled](https://docs.docker.com/engine/reference/commandline/dockerd/#/daemon-user-namespace-options).

Consider the `Dockerfile`, `package.json`, and `yarn.lock` from [this gist](https://gist.github.com/lawliet89/891fb64540b94f4831409bf6905d8661).

```bash
$ docker build -t test .
Sending build context to Docker daemon 29.34 MB
Step 1 : FROM node:7.1
7.1: Pulling from library/node
386a066cd84a: Pull complete 
75ea84187083: Pull complete 
88b459c9f665: Pull complete 
1e3ee139a577: Pull complete 
bd2400e7d880: Pull complete 
b8c6812ba469: Pull complete 
1882bd83135c: Pull complete 
Digest: sha256:873c24bdb94e9b198bbec3c298fb7b2dfc64bce7b2dea71348b62f47814e16fb
Status: Downloaded newer image for node:7.1
 ---> f5eca816b45d
[...]

Issue #918, opened 2016-10-12
Installing in a Docker container on CircleCI fails with "error <package url>: EINVAL: invalid argument, chown '/root/.yarn-cache/<package>/package.json'"


ide:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
Running `yarn` in a Docker container (ex: from a Dockerfile) on CircleCI prints errors like this:

```
yarn install v0.15.1
[1/4] Resolving packages...
[2/4] Fetching packages...
error https://registry.yarnpkg.com/react/-/react-15.3.2.tgz: EINVAL: invalid argument, chown '/root/.yarn-cache/npm-react-15.3.2/package.json'
    at Error (native)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Make a circle.yml file that starts building a Docker image:

``` yml
machine:
  node:
    version: node
  services:
    - docker

dependencies:
  pre:
    - docker build .
```

In the Dockerfile, have some commands like:

```
RUN npm install -g yarn
COPY package.json package.json
COPY yarn.lock yarn.lock
RUN yarn
```

**What is the expected behavior?**
Yarn should successfully install the packages.

**Please mention your node.js, yarn and operating system version.**
Node 6.7.0 (on CircleCI) 
Yarn 0.15.1, 0.16.1
Ubuntu 14.04 (on CircleCI)

ide:
The bug is potentially related to this issue with the `tar` package that also affects jspm: https://github.com/npm/node-tar/issues/39. If that's the case, this problem isn't directly due to Yarn's own code.

artempartos:
Have this problem to

daveisfera:
When I run the same Dockerfile on my local machine, it builds without issue but fails when doing the build on CircleCI. My local machine is using docker 1.12.1, build 6f9534c but CircleCI is using a much older version (1.8, I think), so maybe that's part of the problem.
I tried running with [1.9.1](https://discuss.circleci.com/t/docker-1-9-1-is-available/1009) and [1.10.0](https://discuss.circleci.com/t/docker-1-10-0-is-available-beta/2100) in CircleCI and the error still happened.

jspiro:
Any ideas for workarounds?

artempartos:
@jspiro this is working workaround for this, but you need a non-root user
https://github.com/artempartos/test_yarn

guybedford:
[...]

PR #2103, opened 2016-12-01
Set UID and GID of untared files to zero if user is root (#1837)

daveisfera:
**Summary**
Backport of fix so yarn can be used inside of docker on CircleCI ( see #918 ).

**Test plan**
Tested using a nightly build of yarn and it worked.

### DIFF
--- a/__tests__/constants.js
+++ b/__tests__/constants.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import {getPathKey} from '../src/constants.js';
+import {getPathKey, isRootUser} from '../src/constants.js';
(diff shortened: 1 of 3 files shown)


dave-nm:
Currently, yarn can't be used inside of docker on CircleCI. This is a cherry pick of the change in master so a 0.17.x release can be made that works on CircleCI.

bestander:
I appreciate the effort, that kind of backports are very useful when branches diverge.
But we are going to release 0.18 today, let's just move forward and keep 0.17 behind.

Issue #2073, opened 2016-11-29
yarn install --production does not install dependencies that are transitive dependencies of a dependency and a dev dependency


dreid:
**What is the current behavior?**

When a package is depended on by both a dependency and a devDependency, the devDependency visibility takes precedence when installing into production, this results in shared dependencies not being installed in production.

**If the current behavior is a bug, please provide the steps to reproduce.**
https://github.com/dreid/yarn-dev-dep-failure

`yarn install --production` does not install function-bind

`npm install --production` does install function-bind

**What is the expected behavior?**

Production builds should install dependencies that are visible from production dependencies.

**Please mention your node.js, yarn and operating system version.**

```
❯ yarn --version
0.17.9
❯ node -v
v7.1.0
❯ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.1
BuildVersion:	16B2555
```

bestander:
Making it higher priority, thanks

itamararjuan:
@bestander are you fixing this yourself ?
was thinking of helping solve this as it's totally preventing me from moving some of my services to yarn.

bestander:
@itamararjuan, I would totally appreciate if you fix this.
It may take some time for me to get to this one.
Please don't forget to include a test

itamararjuan:
@bestander on it!

kittens:
Looks like this was fixed with #2116, it required significant refactor to how we do package visibility. Sorry @itamararjuan, didn't realise that the issue I was fixing was so pathalogical and extended to this case too.

itamararjuan:
hey sebastien
no worries thanks for doing so much for everybody. this is an awesome
project. if help is needed dont hesitate to ping and I'll do my best to
help with other issues :)

PR #2116, opened 2016-12-02
Refactor visibility to fix numerous environmental bugs

kittens:
**Summary**

This PR refactors package visibility, simplifies it heavily and fixes numerous bugs around production and optional dependencies.

 - Properly retrieve `production` flag from `config` rather than an `Install` instance.
 - Mark failed optional dependencies as ignored.
 - Filter out dev dependencies in production mode at a later stage rather than intermittently.

**Test plan**

`npm run test`


__tests__/commands/install.js
```diff
@@ -71,6 +71,18 @@ test.concurrent("removes extraneous files that aren't in module or artifacts", a
   }, create);
 });
 
+test.concurrent('--production flag ignores dev dependencies', () => {
```
bestander: we also need a test for https://github.com/yarnpkg/yarn/issues/2073 - when a dev dependency is a transitive dependency of a prod dependency.

src/config.js
```diff
@@ -203,6 +203,10 @@ export default class Config {
       cert: String(opts.cert || this.getOption('cert') || ''),
       key: String(opts.key || this.getOption('key') || ''),
     });
+
+    if (this.getOption('production') || process.env.NODE_ENV === 'production') {
```
bestander: We had a bug recently in React Native.
One of the scripts that calls `yarn install` had NODE_ENV = production and I could not override this with `--no-production`.
Should we open a separate issue for that?

bestander: There is a PR for that https://github.com/yarnpkg/yarn/pull/2057

src/cli/commands/install.js
```diff
@@ -232,7 +230,7 @@ export class Install {
       Object.assign(this.resolutions, json.resolutions);
       Object.assign(manifest, json);
 
-      const pushDeps = (depType, {hint, visibility, optional}) => {
+      const pushDeps = (depType, {hint, optional}, isUsed) => {
```
wyze: I don't know your stance/thoughts on default parameters, but you could default `isUsed = true` here and in the one instance below just pass in the evaluation of production flag.

jkrems:
Just tried this against one of our apps and I'm still seeing problems:

1. `yarn --prod` doesn't prune all non-prod dependencies. Might be connected to...
2. `yarn --prod` installs dependencies of packages it *doesn't* install.

Example for 2.:

```
> ll node_modules/testium/
total 0
drwxr-xr-x  5 jkrems  staff   170B Dec  2 09:13 node_modules/

> ll node_modules/testium/node_modules
total 0
drwxr-xr-x  30 jkrems  staff   1.0K Dec  2 09:13 lodash/
drwxr-xr-x   9 jkrems  staff   306B Dec  2 09:13 minimist/
drwxr-xr-x  10 jkrems  staff   340B Dec  2 09:13 mkdirp/
```

I suspect the problem is that those lodash/minimist/mkdirp are used elsewhere in the tree and the "is in use" logic doesn't handle the case where the same `pkg@version` might be sometimes in use and sometimes not.

kittens:
@jkrems Try the latest version of this PR, I've pushed a few times since creating it.

src/cli/commands/install.js
kittens: I've changed the callsites so there's just one use of `true` now. Eventually if we add a `--dev-only` flag we'd have to remove it anyway so I don't think there's much value.

jkrems:
Same result w/ 40fe07975586db4bb2627ba64d63c2262fe55554.

__tests__/commands/install/integration.js
```diff
@@ -71,36 +67,85 @@ test.concurrent("removes extraneous files that aren't in module or artifacts", a
...
-  const install = new Install({}, config, reporter, lockfile);
+    assert.ok(
+      !await fs.exists(path.join(config.cwd, 'node_modules', 'b')),
+    );
+  });
+});
 
-  const install2 = new Install({flat: true}, config, reporter, lockfile);
-  expect(install2.generateIntegrityHash('foo', [])).not.toEqual(install.generateIntegrityHash('foo', []));
+test.concurrent('hoisting should factor ignored dependencies', async () => {
```
bestander: nice!

bestander:
Looks good, do you know why it fails on Windows with this?
```
production mode with deduped dev dep shouldn't be removed
    SyntaxError: Unexpected token : in JSON at position 13
        at Object.parse (native)
     
```

kittens:
@jkrems Can you put together some small steps that I can use to repro? I'm not sure what issue you're describing so it's hard for me to reproduce or add a test for.

jkrems:
Sorry, I assume I forgot to run `npm run build` last time I tried (?). For future reference, this was my test case:

```
rm -rf package.json node_modules && npm init --yes && /path/to/yarn/bin/yarn.js add --dev testium && /path/to/yarn/bin/yarn.js add gofer && /path/to/yarn/bin/yarn.js --prod && ls node_modules/testium*; rm -rf node_modules && /path/to/yarn/bin/yarn.js --prod && ls node_modules/testium*; true
```

Looks good now! No more non-prod packages in `--prod` and pruning works as expected.

kittens:
Tests are passing across all CI.

Issue #2105, opened 2016-12-01
yarn lifecycle version


rtfpessoa:
Hi,

I just noticed that the yarn cli `version` command is different from npm version.
It does not seem to run the lifecycle "version" command, only runs the "preversion" and "postversion".

My use case is that I need to update some files before release, like compiling etc.
And ideally I would like to do this after the version number is updated in the package.json file.

Am I doing something wrong? How can I have the same behavior in yarn?

As an example: https://github.com/rtfpessoa/diff2html/blob/master/package.json#L48

rtfpessoa:
@Daniel15 just created a pr #2128

Like https://github.com/npm/npm/blob/latest/lib/version.js#L102 after package.json is changed and before the commit is created I added the lifecycle call here https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/version.js#L126

What do you think?

searls:
Safe to close since #2128 was merged?

rtfpessoa:
Yup, you are right.

PR #2128, opened 2016-12-02
Add version lifecycle step to the version cli command

rtfpessoa:
**Summary**

Like npm having the option to perform commands during version publishing is important.
This new lifecycle step `version` should add npm compatibility to the version cli command.

**Test plan**

I cannot find any tests to the version command. Is this a pr that should be tested? Can you give me some guidance on how to do some testing here?

Related to #2105

### DIFF
--- a/__tests__/commands/version.js
+++ b/__tests__/commands/version.js
@@ -1,0 +1,70 @@
+/* @flow */
+
+jest.mock('../../src/util/execute-lifecycle-script');
+jest.mock('../../src/util/child');
+
+import {run as buildRun} from './_helpers.js';
+import {BufferReporter} from '../../src/reporters/index.js';
+import {run} from '../../src/cli/commands/version.js';
+import * as fs from '../../src/util/fs.js';
+import * as reporters from '../../src/reporters/index.js';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
+
+const execCommand: $FlowFixMe = require('../../src/util/execute-lifecycle-script').execCommand;
+
+const path = require('path');
+
+beforeEach(() => execCommand.mockClear());
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'version');
+const runRun = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, config, reporter): Promise<void> => {
+  return run(config, reporter, flags, args);
+});
+
+const oldVersion = '1.0.0';
+const newVersion = '2.0.0';
+
+const gitTagVersion = true;
+
+test('run version with no arguments and --new-version flag', (): Promise<void> => {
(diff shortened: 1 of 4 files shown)


src/cli/commands/version.js
```diff
@@ -124,6 +124,8 @@ export async function setVersion(
       // add manifest
       await spawn('git', ['add', pkgLoc]);
 
+      await config.executeLifecycleScript('version');
```
Daniel15: This will only execute if the package is in a Git repo, as it's inside the `if (isGit) {` condition. You probably want to execute it when outside of a Git repo too, right?

rtfpessoa: You are right.

rtfpessoa: Just moved it out of the if. Still fulfills the same purpose.

rtfpessoa: I accidentally formatted the code inside Intellij IDEA and it started getting errors in ESLint.

Reverted the formatting. Everything seems fine now. @Daniel15

rtfpessoa:
@Daniel15 any news about this?

wyze:
Yes, please add a test for this. A good starting point would be to look at the `run` tests as they mock the execution of commands.

rtfpessoa:
@wyze what is the best way to format the code? Every time I use Intellij IDEA it changes lots of stuff. Any suggestions?

wyze:
I've never used Intellij, but there should be an option to turn off formatting? Even better if it could turn it off per project.

rtfpessoa:
@wyze ok, I will avoid formatting 😄 

Regarding the tests I need some more help.

I started with three tests but I could only make the first work.

Can you look at the tests I created and guide me in what can be wrong in the other two?

rtfpessoa:
@wyze @Daniel15 any ideas what I might be doing wrong?

rtfpessoa:
@wyze @Daniel15 after a huge struggle I was finally able to mock the calls to the lifecycle executor and git. Still, I had to make some changes for it to be used in a similar way to run.js.

What do you think?

__tests__/commands/version.js
```diff
@@ -0,0 +1,70 @@
...
+
+    const preversionLifecycle = ['preversion', config, pkg.scripts.preversion, config.cwd];
+    const versionLifecycle = ['version', config, pkg.scripts.version, config.cwd];
+    const postversionLifecycle = ['postversion', config, pkg.scripts.postversion, config.cwd];
+
+    expect(execCommand.mock.calls.length).toBe(3);
+
+    expect(execCommand.mock.calls[0]).toEqual(preversionLifecycle);
+    expect(execCommand.mock.calls[1]).toEqual(versionLifecycle);
+    expect(execCommand.mock.calls[2]).toEqual(postversionLifecycle);
```
bestander: Those are damn good tests.
Thanks for adding them!

Issue #1307, opened 2016-10-20
Fetching from GitHub raises a Connection reset by peer error


albertfdp:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

When running `yarn` on package.json that contains many repositories pointing to GitHub, this will cause yarn to fail on the fetch step.

**If the current behavior is a bug, please provide the steps to reproduce.**

I have created a example repository with a package.json that I can reproduce the error https://github.com/albertfdp/yarn-github-repos/blob/master/package.json.
1. Checkout https://github.com/albertfdp/yarn-github-repos
2. Run yarn

On the fetch step, it raises an error like this (note, it's not always the same dependency that fails):

```
yarn install v0.16.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads ssh://git@github.com/ReactTraining/react-router.git
Directory: /PATH/yarn-github-repos
Output:
ssh_exchange_identification: read: Connection reset by peer
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

(in a different run):

```
yarn install v0.16.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Command failed.
Exit code: 1
Command: git
Arguments: pull
Directory: /PATH/.yarn-cache/.tmp/00ce03dbdd7a458a1f39defd7ff9dcf5
Output:
ssh_exchange_identification: read: Connection reset by peer
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

The real use-case I have is for private GitHub repos, the repos in this example are public and I should use the npm registry, but it's only for helping to reproduce the bug.

**What is the expected behavior?**

That fetching packages from GitHub works just as with the `npm-cli`.

**Please mention your node.js, yarn and operating system version.**
- node v5.6.0
- yarn v0.16.1
- macOS 10.11.6

chase:
Works correctly with Node v6.6, yarn v0.16.1, and Linux 4.6.7-1.

albertfdp:
[...]

PR #2129, opened 2016-12-03
Add support to max network concurrent requests

albertfdp:
**Summary**

In my organization we are consuming many private dependencies as GitHub private repositories through ssh (i.e. `git+ssh://git@github.com/org/name.git`), which results that when the number of these dependencies exceeds some limit (around 10), the number of concurrent requests to GitHub from the same IP will result in a `Connection reset by peer error` which is a DoS protection from GitHub. 

Fixes https://github.com/yarnpkg/yarn/issues/1307

Since yarn will cache all successful fetches, it really does not matter that the first fetch of dependencies is a bit slow.

Let me know if `concurrency` is also the proper name, I'm thinking if I should use `network-concurrency` instead?

**Test plan**

Running `yarn install --concurrency 1` on https://github.com/albertfdp/yarn-github-repos should not raise a `Connection reset by peer` error.

I've been looking into the tests at https://github.com/yarnpkg/yarn/blob/master/__tests__/util/request-manager.js but I'm not sure how to add a test to verify the concurrency. I'd be very happy to add a unit test if I can get some guidance.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -34,6 +34,7 @@ export type ConfigOptions = {
   cafile?: ?string,
   production?: boolean,
   binLinks?: boolean,
+  networkConcurrency?: number,
 
   // Loosely compare semver for invalid cases like "0.01.0"
   looseSemver?: ?boolean,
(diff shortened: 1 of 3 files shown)


src/config.js
```diff
@@ -202,6 +203,7 @@ export default class Config {
       cafile: String(opts.cafile || this.getOption('cafile') || ''),
       cert: String(opts.cert || this.getOption('cert') || ''),
       key: String(opts.key || this.getOption('key') || ''),
+      concurrency: Number(opts.concurrency || this.getOption('concurrency') || constants.NETWORK_CONCURRENCY),
```
bestander: we also have child concurrency, I think this option should be more specific to avoid confusion what kind of concurrency it is

PR #2144, opened 2016-12-04
Fix an issue with cache config tests modifying actual config

mvestergaard:
**Summary**

The tests from #2046 modify the config in ~/.yarnrc which results in cache folders being created in weird places after having run the tests.

**Test plan**


superxp1412:
Hi @mvestergaard ,

This test is for cache-folder directory settings. That's the reason why it creates another cache folders during the tests.

superxp1412:
@bestander Please pick up this pull request to fully resolve the sandbox issue: #2157
Thanks to all you guys.

PR #2145, opened 2016-12-04
Duplicate upgrade-engines fix. Fixes #2139

### LINKED ISSUE #2139: yarn upgrade --help lists --ignore-engines twice.
```bash
$ yarn upgrade --help

  Usage: yarn upgrade [flags]

  Options:

    --verbose                   output verbose messages on internal operations
    --offline                   trigger an error if any required dependencies are not available in local cache
    --prefer-offline            use network only if dependencies are not available in local cache
    --strict-semver
    --json
    --ignore-scripts            don't run lifecycle scripts
    --har                       save HAR output of network traffic
    --ignore-platform           ignore platform checks
    --ignore-engines            ignore engines check
    --ignore-optional
    --force                     ignore all caches
    --no-bin-links              don't generate bin links when setting up packages
    --flat                      only allow one version of a package
    --prod, --production
    --no-lockfile               don't read or generate a lockfile
    --pure-lockfile             don't generate a lockfile
    --global-folder <path>
    --no-emoji                  disable emoji in output
    --proxy <host>
    --https-proxy <host>
    --no-progress               disable progress bar
    --ignore-engines            ignore engines check
```

`--ignore-engines` is listed twice above I'm on 0.17.10.

olingern:
**Summary**
Remove engine option from upgrade help as it's supported globally.

Fixes #2139

### DIFF
--- a/src/cli/commands/upgrade.js
+++ b/src/cli/commands/upgrade.js
@@ -8,8 +8,6 @@ import Lockfile from '../../lockfile/wrapper.js';
 export function setFlags(commander: Object) {
   // TODO: support some flags that install command has
   commander.usage('upgrade [flags]');
-
-  commander.option('--ignore-engines', 'ignore engines check');
 }
 
 export const requireLockfile = true;


wyze:
Thanks!

Issue #2158, opened 2016-12-05
yarn cache dir


philtay:
**Do you want to request a *feature* or report a *bug*?**
semi-bug / annoyance

**What is the current behavior?**
`yarn cache dir` prints:

```
yarn cache v0.18.0
/root/.cache/yarn
Done in 0.04s.
```

**What is the expected behavior?**
`yarn cache dir` command should print only the directory (in this case /root/.cache/yarn) without other useless info. Currently it can't be used in any script. e.g. forget about `export YARN_CACHE=$(yarn cache dir)` and similar. In this case `npm` does the right thing. `npm config get cache` prints just the directory.

vjpr:
`yarn cache dir --verbose` for existing behaviour perhaps.

philtay:
Another thing: this is somewhat related to https://github.com/yarnpkg/yarn/issues/986. If I want to remove the cache during a Docker build, a clean way would be `yarn install && rm -rf $(yarn cache dir)`.

paul-soporan:
Closing as in v2 `yarn config get cacheFolder` (the equivalent of `yarn cache dir`) only prints the path.

https://yarnpkg.com/getting-started/migration

PR #2159, opened 2016-12-05
Do not output wrapper for cache dir command (#2158)

iredchuk:
`yarn cache dir` command should not output header and footer, but only the cache directory path.

Resolves issue #2158.

<code>$ yarn cache dir</code>

Expected output should look like:
<code>/root/.cache/yarn</code>

### DIFF
--- a/src/cli/commands/cache.js
+++ b/src/cli/commands/cache.js
@@ -8,6 +8,10 @@ import {METADATA_FILENAME} from '../../constants';
 
 const path = require('path');
 
+export function hasWrapper(flags: Object, args: Array<string>): boolean {
+  return args[0] !== 'dir';
+}
+
 export const {run, setFlags} = buildSubCommands('cache', {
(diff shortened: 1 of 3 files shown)


src/cli/commands/cache.js
```diff
@@ -64,3 +64,7 @@ export const {run, setFlags} = buildSubCommands('cache', {
     }
   },
 });
+
```
torifat: Can you please move it to top. Like we have for other commands: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/licenses.js#L13

iredchuk: It's kind of mixed now: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/config.js#L77
But I'll move both.

torifat: Ah, cool. Thanks 👍

PR #2164, opened 2016-12-06
Fix an issue in package hoisting

### LINKED ISSUE #2142: v0.18.0 postinstall missing dependencies
Bug

I think some dependencies are missing at the time when `postinstall` scripts are ran.

I'm having this error:
```
yarn install v0.18.0
info No lockfile found.
warning test: No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
warning fsevents@1.0.15: The platform "linux" is incompatible with this module.
info "fsevents@1.0.15" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[1/3] ⠐ gifsicle
[2/3] ⠐ jpegtran-bin
[3/3] ⠐ optipng-bin
[-/3] ⠐ waiting...
error [my-project]/node_modules/jpegtran-bin: Command failed.
Exit code: 1
Command: sh
Arguments: -c node lib/install.js
Directory: [my-project]/node_modules/jpegtran-bin
Output:
module.js:472
    throw err;
    ^

Error: Cannot find module 'ini'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> ([my-project]/node_modules/rc/lib/utils.js:3:12)
    at Object.Module._extensions..js (module.js:580:10)
    at Function.Module._load (module.js:439:3)
```

In another project, after a `yarn upgrade` I also encountered this problem with other packages (`gifsicle`'s `postinstall` missing `deep-extend`). So it seems like a general problem.

Given this `package.json`:
```json
{
  "name": "test",
  "devDependencies": {
    "babel-cli": "^6.18.0",
    "gulp-imagemin": "^3.1.1"
  }
}
```
Run `yarn install`.

I noticed that if I try to install only one of these two, it will work. I guess that's an issue with shared dependencies between them.

Installation completes successful.

node.js: v7.2.0
yarn: v0.18.0
OS: Arch Linux

mvestergaard:
**Summary**

This is an attempt to fix #2142 

Using this package.json
```
{
  "name": "test",
  "devDependencies": {
    "babel-cli": "^6.18.0",
    "gulp-imagemin": "^3.1.1"
  }
}
```

Should create a scenario along the lines of this (on any platform except OSX):
```
babel-cli [installed]
└─ chokidar [installed]
   └─ fsevents [skipped, potentially the whole tree]
      └─ node-pre-gyp [skipped good]
         └─ rc [not skipped, good]
            ├─ deep-extend [skipped, bad!]
            ├─ ini [skipped, bad!]
            ├─ minimist [not skipped, good]
            └─ strip-json-comments [skipped, bad!]

gulp-imagemin [installed]
└─ imagemin-jpegtran [installed]
   └─ jpegtran-bin [installed]
      └─ bin-build [installed]
         └─ download [installed]
            └─ caw [installed]
               └─ get-proxy [installed]
                  └─ rc [installed]
                     ├─ deep-extend [missing]
                     ├─ ini [missing]
                     ├─ minimist [installed]
                     └─ strip-json-comments [missing]

[...]
{various-other-packages} [installed]
└─ minimist [installed]
```

My findings from debugging is that this happens because of the following:

It seems like what happens is that `rc` is ignored in the `fsevents` branch.

This causes `get-proxy#rc` to be removed from the tree in the method `hoist@283`: `this.tree.delete(key);` and when it looks at the dependency `deep-extend` in `_seed@141` it will skip it because `get-proxy#rc` was removed from the tree:
```javascript
      if (!this.tree.get(parent.key)) {
        return null;
      }
```

`get-proxy#rc` is removed from the tree due to:
```javascript
    if (duplicate) {
      info.addHistory(`Satisfied from above by ${newKey}`);
      this.declareRename(info, rawParts, parts);
      return;
    }
```

With this change, installed packages align with installiation through npm. `npm install` and `npm prune` yield no changes.

At this time, this seems to break the test `hoisting should factor ignored dependencies`.

**Test plan**

I have not been able to replicate the original scenario in a test.
I've made multiple attempts including https://github.com/mvestergaard/yarn/commit/9521157ace19bb4d42240532091d8b5402b05ef5


bestander:
Existing tests are failing

mvestergaard:
Yep I know. The test `hoisting should factor ignored dependencies` may need to be refactored, unless my fix is completely wrong.

bestander:
You probably should just add a new test with your example and test that all expected folders are in place.
```
  "babel-cli": "^6.18.0",
  "gulp-imagemin": "^3.1.1"
```
We cache network requests, so no need to worry about them.

mvestergaard:
@bestander It won't fail on OSX. So it's not very reliable.

bestander:
We have a few tests that are skipped on some platforms.
Clout CI run Windows and Linux, so it should be fine.
An isolated custom test is always better but real world examples are good
too

mvestergaard:
Fair enough. I'll add a test later.
Looking into whether the `hoisting should factor ignored dependencies` test should be refactored is not a task I feel I have enough knowledge of the internals to do however.

bestander:
It is ok to do it in steps.
If you add a test that fails properly we can merge it and then do a new PR
that fixes the issue based on the test

mvestergaard:
@bestander do I need to do anything special to do an actual install in tests?
I seem to be hitting some sort of different registry because i get the following:
```
    Error: Couldn't find any versions for "babel-register" that matches "^6.18.0". Possible versions: "6.1.4\n6.1.17\n6.1.18\n6.2.0\n6.2.4\n6.3.2\n6.3.13\n6.4.3\n6.5.0\n6.5.0-1\n6.5.1\n6.5.2\n6.6.0\n6.6.5\n6.7.2\n6.8.0\n6.9.0\n6.11.5\n6.11.6\n6.14.0\n6.16.0\n6.16.3"
```

bestander:
@mvestergaard, could you commit it here?
I'll have a look

mvestergaard:
@bestander Done

bestander:
For tests we cache http requests in __tests__/fixtures/request-cache so that our CI does not depend on network when running them.
Just remove the file for that package and rerun the test

mvestergaard:
Ok that helped on my machine. It fails on CI too though.

QWp6t:
I was having this issue as well. Just tested your fix and it worked for me on Windows 10. Optional dependency `fsevents` was being skipped, which was causing all of the imagemin plugins to fail due to `ini` and other dependencies of `fsevents` also being skipped.

I nuked all caches and my entire yarn installation before and after applying the patch to confirm that it was this patch that resolved my issue.

mvestergaard:
@bestander I've modified the `hoisting should factor ignored dependencies` to something I believe is more correct. Can you please take a look?
I've added comments that explain the reasoning behind the changes.

The request-cache has been giving me some issues behaving differently in CI compared to my own machine, hopefully my latest commit passes.

mvestergaard:
AppVoyer doesn't want to pick up `babel-register` from the request cache from some reason. Travis works though. Not sure if it makes sense to you.

mvestergaard:
@bestander Have you had a chance to look at this yet? Feedback would be appreciated.

bestander:
@mvestergaard will look right now, thanks a lot!

__tests__/commands/install/issue-2142-regression.js
```diff
@@ -0,0 +1,25 @@
...
+const assert = require('assert');
+const path = require('path');
+
+beforeEach(request.__resetAuthedRequests);
+// $FlowFixMe
+afterEach(request.__resetAuthedRequests);
+
+// problem occuring due to optional dependency incompatible with os, in this case fsevents
+// if issue exists, this will not fail on OSX
+test.concurrent('issue #2142 regression test',
```
bestander: I am not 100% sure when we want to split all tests into separate files.
How about moving it into integration.js for now?
Could you also give it a descriptive name so that people don't go to the web searching what issue 2142 was about?

bestander: I mean integration-deduping

mvestergaard:
@bestander Done. Having it in it's own file was mostly for my own convenience, so fair point.

__tests__/commands/install/integration.js
```diff
@@ -102,36 +103,43 @@ test.concurrent('hoisting should factor ignored dependencies', async () => {
...
+    // devDependency b should not be installed
     assert.ok(
       !await fs.exists(path.join(config.cwd, 'node_modules', 'b')),
     );
 
-    assert.ok(
-      !await fs.exists(path.join(config.cwd, 'node_modules', 'c')),
+    // c is only depended upon by d in version 2.0.0, so it should be hoisted
+    assert.equal(
+      (await fs.readJson(path.join(config.cwd, 'node_modules', 'c', 'package.json'))).version,
```
bestander: That is the idea of Yarn - it keeps packages hoisting deterministic between normal and production build.
The test checked that hoisting of C should be stable and I think it was correct

mvestergaard: In that case, my attempt at a fix probably won't work.

The root problem from what I could find is when you have two different packages with dependencies required by both. When one of the parent packages are ignored, either by `--production` flag or being incompatible, it ends up skipping the shared dependencies because the ignored package goes through the hoisting logic, and the dependency thereby end up being skipped due to being seen as a duplicate, even though it hasn't been installed at all.

bestander:
Let's undo your code change and merge your test as disabled in the meantime.
We'll come up with a fix in the next few days.

mvestergaard:
@bestander I reverted the changes, and set the new test to `.skip`

Issue #990, opened 2016-10-13
Need a big optimization for Windows


OshotOkill:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behavior?**

I've tested a lot about the installation speed between MacOS and Windows. According to the results it seems that yarn has far less optimizations for Windows. e.g. Here are the comparisons of installing `react-native` :

<br />

> Test Machines: 
> - ThinkPad X1 Carbon 4, 1TB PCI-E SSD, 16GB Memory
> - MacBook Air 2014, 256GB SSD, 4GB Memory
> 
> No cache & same network environment

<br />

**MacOS**

npm@3.10.9:  **1m 31s**

![2016-10-13 17 52 24](https://cloud.githubusercontent.com/assets/6079112/19346419/ca8abd58-9175-11e6-80af-0ec1a3baef7e.png)

yarn@0.15.1:  **39s**

![2016-10-13 17 54 53](https://cloud.githubusercontent.com/assets/6079112/19346420/ca8b7c98-9175-11e6-9776-b864db49d10f.png)

<br />

**Windows**

npm@3.10.9: **2m 24s**

![2](https://cloud.githubusercontent.com/assets/6079112/19346739/8b14e1ec-9177-11e6-9437-642b6d4e60c1.PNG)

yarn@0.15.1: **2m 19s**

![1](https://cloud.githubusercontent.com/assets/6079112/19346742/8ed1370e-9177-11e6-9135-56228dac77c9.png)

<br />

So, it seems yarn has no advantage over npm on Windows. Did anyone face with this appearance?

**Please mention your node.js, yarn and operating system version.**
nodejs:  6.8.0
yarn:      0.15.1
OS:        Windows 10 14393.321 & MacOS 10.12

atkawa7:
+1

Daniel15:
Hi @OshotOkill! Thanks for trying Yarn. Are you using Cygwin or WSL ("Bash on Ubuntu on Windows")? Both are known to have pretty bad disk IO performance.

Also, React Native has a *huge** number of files so copying them into `node_modules` is pretty slow, and disk IO for lots of little files on Windows is generally slower than Mac OS (which itself is slower than Linux ext4). We do have a task to experiment with hardlinks (#499) which should improve perf in this scenario. 

> No cache & same network environment

The main improvement with Yarn is when you have a warm cache (ie. after you've installed a package at least once).

OshotOkill:
[...]

PR #2176, opened 2016-12-07
[WIP] Skip already installed modules

voliva:
`yarn install` was very slow on Windows because it takes very long to copy files from the cache folder to `node_modules` folder, specially if it's not whitelisted in Windows Defender. See #990 for further details.
This made yarn very slow vs npm when adding new packages, specially if it had installed huge packages such as Webpack or React (or both), as all previous packages were re-coppied.

This PR solves this problem by skipping those packages that are already installed and have the exact same version that the ones that are about to be copied.

I'm not too sure about the correctness of this because I didn't go too deep into yarn's src architecture, so I can't be sure if I'm missing something... But for a normal use case it seems to work: Only the new packages or those that are needed to update are copied, while the previous ones are just ignored.


src/package-linker.js
```diff
@@ -137,6 +138,15 @@ export default class PackageLinker {
         phantomFiles.push(path.join(dest, file));
       }
 
+      try {
+        let obj = await fs.readJson(dest + "\\package.json");
```
wyze: The path shouldn't contain `\\` as that is only valid for Windows. This should use `path.join(dest, 'package.json')` instead to work for all OSes.

src/package-linker.js
```diff
@@ -147,6 +157,7 @@ export default class PackageLinker {
         },
       });
     }
+    this.reporter.info('skipped ' + skipped.length + ' modules because they are already installed');
```
wyze: This shouldn't be a string here, it should use `this.reporter.lang` like everywhere else.

voliva:
Fixed comments from @wyze and some lint warnings/errors. Will still need to check the tests (`Removes extraneous files` is failing)

bestander:
The good thing about yarn is that I can play around with node_modules - add console.log, edit code - and then clear everything up with `yarn install --force`.
I would not want to have a skip behavior by default.
Also, any ideas how we could whitelist the file operations in windows defender and file indexer?

voliva:
@bestander I think most of the times when someone runs yarn add / yarn install, node_modules folder can be assumed unchanged, so  it makes sense that the default behaviour is just to skip the installed folders.

What's a good point though is that when calling `yarn install --force` it should indeed replace all the installed modules, as that would be the expected behaviour (the user is asking yarn to install everything regardless of the current node_modules state).

bestander:
Since 0.17 we fixed integrity checking that runs during install, so it won't do installation if integrity files is in place and dependencies did not change.
So this optimization is only for cases when many dependencies changed since last install because yarn does not do actual copy if individual files did not change.

bestander:
Why I am a bit opposed to this is that this is another point of state in the system.
We already have quite complicated caches, copy that does not rewrite files and integrity checks that makes it easy for bugs to creep in.
Adding one more point where installation can be incorrect because of the state of the system is adding more risk.
I would rather us investigate all other possible ways to optimise first.

xantorres:
@bestander Yarn is terribly slow in Windows, even without windows defender activated and having a single dependency updated, at the point where it is totally impractical and unusable in a daily basis. We should either merge this PR or find another solution otherwise no Windows user will ever use this application

bestander:
There are some changes that are coming or landed in the latest version of Yarn:
- number of IO operations reduced with improved hoisting algorithm
- you can opt-in to use hard links with --link-duplicates if there are many modules installed more than once
- soon install command will do a structure check similar to the one in this PR before running fetch and link phases
- we improve linker and fetcher code to make less IO operations all the time, feel free to revise the code and send a PR to remove redundant disk IO calls

Besides that, read https://github.com/yarnpkg/yarn/issues/990 with tests regarding Windows IO operations compared to Linux on the same machine.
In general NTFS is just slower than Ext3 everywhere: npm, Yarn and basic file copy.

I'll close this PR for the reasons in a previous comment

PR #2178, opened 2016-12-07
Apply the network-concurrency option to the package fetchers

albertfdp:
**Summary**

This is a follow-up from https://github.com/yarnpkg/yarn/pull/2129 where the network concurrency was not being passed down to the package fetcher, so the problem was not fixed in that scenario.

**Test plan**

Running `yarn install --concurrency 1` on https://github.com/albertfdp/yarn-github-repos should not raise a Connection reset by peer error.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -81,7 +81,7 @@ commander.option(
   '--no-progress',
   'disable progress bar',
 );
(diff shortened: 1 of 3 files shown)


albertfdp:
@bestander Would it be ok to merge this as a fix for #2129, it seems that without this change the other merge is not effective. Thanks!

PR #2183, opened 2016-12-07
Support field paths in the info command

### LINKED ISSUE #2175: yarn info doesn't support field paths

Request a feature.

`yarn info` doesn't support the "dot notation" to access deeply nested fields. To see that, invoke:
```
yarn info angular-cli dist-tags.latest
```
and see that it returns undefined. Compare that to the `npm` output:
```
npm info angular-cli dist-tags.latest
```
which returns `1.0.0-beta.22-1`. Without using the dot notation, both look similar (although yarn introduces some authentication noise that erroneously goes to stdout instead of stderr).

It should behave as npm

Node 6.9.2, yarn 0.18.0, macOS 10.12.1 (16B2555)

iredchuk:
`yarn info` should support field paths for the second argument (fields separated by the dot), like `npm info` does.

`yarn info yarn dist-tags.latest`
Should output
`0.17.10`

Also 
`yarn info`
Should be equivalent to
`yarn info .`
i.e. output info on the local package.

* Also default to local package when second arg is not specified
* Fixes #2175


wyze:
Could you add a test please?

iredchuk:
@wyze Would be happy to add tests, but at the moment there's no tests for **info** command. So I will have to set them up, need to look into that. I would rather completely mock the response from registry, but it seems like other commands rather work with a real registry. What do you think?

wyze:
I think for now it is okay to hit the real registry. I know the tests for `outdated` hits the real registry.

iredchuk:
OK, I'll try + also test the existing cases. Test setups for commands look a bit complicated at the first look, so need to wrap my head around it.

iredchuk:
@wyze Added tests

wyze:
Thanks @iredchuk!

Issue #2180, opened 2016-12-07
very important xmas command is missing


blakek:
**Do you want to request a *feature* or report a *bug*?**
Requesting a *feature*.  This command is very important to our workflow and morale

**What is the current behavior?**

```bash
$ yarn xmas
```

outputs:

```
error Command "xmas" not found.
```

This makes our users very sad.

**What is the expected behavior?**

```bash
$ yarn xmas
```

should output something similarly festive:

```
﻿﻿﻿﻿
                     ★
                    ／＼
                   ／  ＼
                  ／⁂ ⸮&＼
                 ／⁂ ⸮｡ ⸛＼
                ／⁂ ｡&   ⸛＼
               ／   ｡      ＼
              ／        ⁂  i＼
             ／｡ ⸛  &     i  ＼
            ／   i @i ⸛ ｡  ｡ @＼
           ／⁂    ⸛ @ ⁂  ⁂i｡   ＼
          ／i ⸛   ｡@ ⸛⸮    @⁂ ⁂⸛＼
         ／@  ⸛@  @    ⸛@ ｡      ＼
        ／ ⸛   ii   @  && ⸛   @&  ＼
       ／   ⸛   &@  @    ⸮  ⸛⸮  i⸮ ＼
      ／⸛        ｡@ ⸛  i  @@   &    ＼
     ／ ⁂ ⸛       @&@  @       ⁂&    ＼
    ／ ⸮    ⸛   &  ⸮⸛i ⸮｡@ @   ⸛    & ＼
   ／&⸮        i&⁂    i⸛  @ @⸛ &｡  ⸮ ⸮ ＼
  ／ ⁂ ⸮  ⸮   ⸮⁂&i &  ⁂⁂@@ ⸛   ｡     ⁂⁂ ＼
 ／  i  i  ｡ ⸮   ⁂   i ⸮  @ ⸛@ ⸛i  &⸛    ＼
 ^^^^^^^^^^^^^^^^^^^|  |^^^^^^^^^^^^^^^^^^^
                    |  |                   

Happy Xmas from the yarn team
```

halvves:
@blakek ilu

Daniel15:
Pull requests implementing this extremely important feature would be highly appreciated!

ljharb:
Don't forget the xmas tree colors, and `npm substack`, and `npm visnup` :-)

blakek:
We like emojis, right?  How about something along these lines:
```
           🌟
           🎄
[...]

PR #2196, opened 2016-12-08
important xmas command

halvves:
**Summary**
This feature is in answer to #2180. Very important `xmas` command is missing. I have simply ported npm's existing `xmas` command, but we should eventually come up with our own original artwork.

**Test plan**

I'm not entirely sure if I've contributed correctly. I tried to follow the contributing guidelines to the best of my knowledge. If there is anything that need be changed/added, please let me know.

xmas


torifat:
Check https://github.com/yarnpkg/yarn/issues/2180#issuecomment-266555814

PR #2209, opened 2016-12-10
Show homepage of outdated dependencies

randycoulman:
**Summary**

When viewing outdated dependencies via `yarn outdated` or `yarn upgrade-interactive`, it is often desirable to look at what has changed in the dependent package.  While we don’t have easy access to each dependency’s CHANGELOG, we do have access to its homepage or repository URL.

This PR adds a new column to the output of `yarn outdated` and `yarn upgrade-interactive` to show a URL for each dependency.  Some terminal programs, such as iTerm, allow clicking on URLs to open them in a browser, making this very convenient.  Other terminals only allow copy and paste, but that’s still more convenient than manually attempting to find the package’s homepage.

This feature was inspired by a similar feature of [npm-check](https://github.com/dylang/npm-check).

**Test plan**

Run `yarn outdated` and `yarn upgrade-interactive` with both 0.17.10 (current release) and with this PR applied and check that the new column appears.

`yarn outdated`:

Before:
screen shot 2016-12-09 at 4 48 48 pm

After:
screen shot 2016-12-09 at 4 49 34 pm

`yarn upgrade-interactive`:

Before:
screen shot 2016-12-09 at 4 49 12 pm

After:
screen shot 2016-12-09 at 4 49 54 pm

**Implementation notes**

* For normal dependencies, the URL comes from `NpmRegistry.checkOutdated()`, which has access to the information in the dependency's `package.json` file.  Note that unlike `yarn licenses`, these commands prefer to use the `homepage` in preference to the `repository.url`, as that's generally a better place to go for release announcements and such.

* The URL gets passed up to `PackageRequest.getOutdatedPackages()`.  For "exotic" dependencies, we don't have `package.json` information available, so we use the normalized version pattern's range instead, as that will often be a git or file URL.  Perhaps more could be done here.

* `yarn upgrade-interactive` formats its own table by padding the columns.  I explicitly do not pad the url column as it is the last column and doesn't need it.  If there are packages with long names or long URLs, there will be some line wrapping going on, and adding extra spaces to the end of every URL just makes the problem worse.  The tradeoff is that if another column is later added after the URL, padding will need to be added at that time.

* I had to add URLs to one of the test cases, as it checks the entire output row.  I'm not thrilled with this extra coupling, but chose to keep the form of the current test.

### DIFF
--- a/src/cli/commands/upgrade-interactive.js
+++ b/src/cli/commands/upgrade-interactive.js
@@ -56,9 +56,9 @@ export async function run(
 
   const getNameFromHint = (hint) => hint ? `${hint}Dependencies` : 'dependencies';
 
-  const maxLengthArr = {name: 0, current: 0, latest: 0};
+  const maxLengthArr = {name: 0, current: 0, latest: 0, url: 0};
   deps.forEach((dep) =>
-    ['name', 'current', 'latest'].forEach((key) => {
+    ['name', 'current', 'latest', 'url'].forEach((key) => {
       maxLengthArr[key] = Math.max(maxLengthArr[key], dep[key].length);
     }),
   );
@@ -86,7 +86,8 @@ export async function run(
     const name = colorizeName(dep)(padding('name'));
     const current = reporter.format.blue(padding('current'));
     const latest = colorizeDiff(dep.current, padding('latest'));
-    return `${name}  ${current}  ❯  ${latest}`;
+    const url = reporter.format.cyan(dep.url);
+    return `${name}  ${current}  ❯  ${latest}  ${url}`;
   };
 
   const groupedDeps = deps.reduce((acc, dep) => {
--- a/src/cli/commands/upgrade-interactive.js
+++ b/src/cli/commands/upgrade-interactive.js
@@ -56,9 +56,9 @@ export async function run(
 
   const getNameFromHint = (hint) => hint ? `${hint}Dependencies` : 'dependencies';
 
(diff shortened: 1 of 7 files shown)


src/cli/commands/upgrade-interactive.js
```diff
@@ -56,9 +56,9 @@ export async function run(
 
   const getNameFromHint = (hint) => hint ? `${hint}Dependencies` : 'dependencies';
 
-  const maxLengthArr = {name: 0, current: 0, latest: 0};
+  const maxLengthArr = {name: 0, current: 0, latest: 0, url: 0};
   deps.forEach((dep) =>
-    ['name', 'current', 'latest'].forEach((key) => {
```
torifat: `maxLengthArr` for `url` is redundant since you are not using padding for it.

src/cli/commands/upgrade-interactive.js
```diff
@@ -56,9 +56,9 @@ export async function run(
 
   const getNameFromHint = (hint) => hint ? `${hint}Dependencies` : 'dependencies';
 
-  const maxLengthArr = {name: 0, current: 0, latest: 0};
+  const maxLengthArr = {name: 0, current: 0, latest: 0, url: 0};
```
torifat: `maxLengthArr` for `url` is redundant since you are not using padding for it.

randycoulman: Good point; fixed in dd82799

torifat:
Thanks 👍 

I forgot press the `Comment` button 😂

PR #2215, opened 2016-12-10
Add project commands `yarn run` could run

### LINKED ISSUE #2194: wish: `yarn run` should include the commands that project commands would run

feature.

```
$ yarn run
yarn run v0.17.8
[...]
info Project commands
   - start
   - test
```

Desired behavior is show the commands that the project commands would run. `npm run` does this and it's helpful:

```
$ yarn run
yarn run v0.17.8
error No command specified.
[...]
info Project commands
   - start
        node server.js 
   - test
       gulp test-backend --no-bail
```

N/A.

sotojuan:
**Summary**

Closes #2194.

This pull request improves `yarn run` by adding the project commands that would be run when used by `yarn run`.

Essentially, this makes `yarn run` with no arguments show the same thing as `npm run` with no arguments.

**Test plan**

Here's what I have now: http://i.imgur.com/6MKIPya.png

(Ignore the `vundefined` version—I fixed that after the screenshot).

**WIP**

I am vastly unfamiliar with Flow and types in general so this PR is far from finished. Here are things I could use help with:

* The linting error about `pkg.scripts` being undefined
* `run` uses `log` from `ConsoleReporter`, which takes an `Array<string>`. The new scripts array is an array of objects of `{script: command}`[1] structure. I'm not sure how to get both working at the same time without breaking other stuff... so I used `Array<any>` and did some gross checking before printing just to make sure the output looks fine.

I'll also add/fix tests once I fix my implementation.

I'm read the Flow docs and asking around, but any help would be appreciated :-)

[1] For example `{test: 'jest --coverage', lint: 'standard'}`.

### DIFF
--- a/__tests__/commands/run.js
+++ b/__tests__/commands/run.js
@@ -25,13 +25,18 @@ test('lists all available commands with no arguments', (): Promise<void> => {
   return runRun([], {}, 'no-args', (config, reporter): ?Promise<void> => {
     const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
     const scripts = ['build', 'prestart', 'start'];
+    const hints = {
+      build: "echo 'building'",
+      prestart: "echo 'prestart'",
+      start: 'node index.js',
+    };
     const bins = ['cat-names'];
 
     // Emulate run output
     rprtr.error(rprtr.lang('commandNotSpecified'));
     rprtr.info(`${rprtr.lang('binCommands')}${bins.join(', ')}`);
     rprtr.info(rprtr.lang('possibleCommands'));
-    rprtr.list('possibleCommands', scripts);
+    rprtr.list('possibleCommands', scripts, hints);
     rprtr.error(rprtr.lang('commandNotSpecified'));
 
     expect(reporter.getBuffer()).toEqual(rprtr.getBuffer());
--- a/src/reporters/json-reporter.js
+++ b/src/reporters/json-reporter.js
@@ -1,6 +1,10 @@
 /* @flow */
 
-import type {ReporterSpinnerSet, Trees, ReporterSpinner} from './types.js';
+import type {
+  ReporterSpinnerSet,
+  Trees,
+  ReporterSpinner,
+} from './types.js';
 import BaseReporter from './base-reporter.js';
 
 export default class JSONReporter extends BaseReporter {
@@ -26,8 +30,8 @@ export default class JSONReporter extends BaseReporter {
     this._dump('verbose', msg);
   }
 
-  list(type: string, items: Array<string>) {
-    this._dump('list', {type, items});
+  list(type: string, items: Array<string>, hints?: Object) {
+    this._dump('list', {type, items, hints});
   }
 
   tree(type: string, trees: Trees) {
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -44,9 +44,17 @@ export async function run(
       visitedBinFolders.add(binFolder);
     }
   }
-  if (pkg.scripts) {
+  const pkgScripts = pkg.scripts;
+  const cmdHints = {};
+  if (pkgScripts) {
     // inherit `scripts` from manifest
-    pkgCommands = Object.keys(pkg.scripts);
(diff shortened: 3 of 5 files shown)


sotojuan:
Other ideas I've been playing with:

* Make a `Command` type
* Make a `listCommands` function and leave the old `list` function alone
* Both of the above combined

sotojuan:
I had a bit of free time to mess around this morning and did the following:

* Added a `Command` type
* Added a `listCommand` function that deals with `Command`
* Fixed tests

Some of these seem overkill (especially the second bullet).

wyze:
Before your recent push, I was going to suggest maybe adding an optional 3rd parameter to `list` method of the reporters. Something like: `list(key: string, items: Array<string>, hints?: Object)` and then do a conditional based on if `hints` was passed in or not.

sotojuan:
> Before your recent push, I was going to suggest maybe adding an optional 3rd parameter to list method of the reporters

What would be in `hints`? The same as in `Command` type?

BTW, I'm mostly messing around/exploring, so no harm in reverting or changing before merging—I'll squash before hand :D

wyze:
Yes, same structure as `Command` type, but I left it as an Object to keep it more generic.

sotojuan:
I like your idea better; updated my changes! Wasn't 100% on what to do in JSON reporter though.

wyze:
In the JSONReporter, I would just remove the conditional and add hints to the output, because if it isn't passed, it won't be in the output anyways.

```js
{"type":"list","data":{"type":"possibleCommands","items":["build","prestart","start"]}}
```

Hints:

```js
{"type":"list","data":{"type":"possibleCommands","items":["build","prestart","start"],"hints":{"build":"echo 'building'","prestart":"echo 'prestart'","start":"node index.js"}}}
```

__tests__/commands/run.js
```diff
@@ -25,13 +25,18 @@ test('lists all available commands with no arguments', (): Promise<void> => {
   return runRun([], {}, 'no-args', (config, reporter): ?Promise<void> => {
     const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
     const scripts = ['build', 'prestart', 'start'];
+    const hints = {
+      build: 'echo \'building\'',
```
wyze: You can use double quotes here to avoid escaping inner single quotes.

src/reporters/json-reporter.js
```diff
@@ -26,8 +30,12 @@ export default class JSONReporter extends BaseReporter {
     this._dump('verbose', msg);
   }
 
-  list(type: string, items: Array<string>) {
-    this._dump('list', {type, items});
+  list(type: string, items: Array<string>, hints?: Object) {
+    if (hints) {
+      this._dump('list', {type, hints});
+    } else {
+      this._dump('list', {type, items});
```
wyze: I said in a comment, remove the conditional and just make it `this._dump('list', {type, items, hints});`.

sotojuan:
Done!

__tests__/commands/run.js
```diff
@@ -25,13 +25,20 @@ test('lists all available commands with no arguments', (): Promise<void> => {
   return runRun([], {}, 'no-args', (config, reporter): ?Promise<void> => {
     const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
     const scripts = ['build', 'prestart', 'start'];
+    /* eslint-disable quotes */
+    const hints = {
+      build: "echo 'building'",
+      prestart: "echo 'prestart'",
+      start: "node index.js",
```
wyze: This one should be single quotes since there is nothing to escape. That is why you were getting the ESLint error.

sotojuan:
Got it 👍

wyze:
Looks great, I know you mentioned above you would squash, so you can do that now, or I can in the UI.

sotojuan:
Squashed! Thanks for the help.

wyze:
Thanks for the great iteration and contribution, Juan!

Daniel15:
By the way @sotojuan - What font are you using in your screenshot?

sotojuan:
Operator Mono.

PR #2223, opened 2016-12-11
Allow production=false flag to override production ENV

### LINKED ISSUE #2167: Can't parse yarn install --production=false
bug

With the `NODE_ENV` set to _production_, running this command

```
yarn install --production=false
```

I expect it to install all dependencies (including `devDependencies`) the same way that npm does. See npm/npm#6803 ([this comment specifically](https://github.com/npm/npm/issues/6803#issuecomment-75621914)) for reference.

Instead yarn errors out because it can't parse the command:

```
yarn install v0.18.0
error `install` has been replaced with `add` to add new dependencies. Run "yarn add false" instead.
```

Node: v7.0.0
Yarn: 0.18.0
OS: Windows 10

alexbrazier:
**Summary**

Add ability to pass `production=false` when running `yarn install`, which will take precedence over the environment variable `NODE_ENV=production`.

This will allow you to have your NODE_ENV set as production while still installing dev dependencies, e.g. when running a script such `yarn install && yarn run build` to build your code in production mode while installing dev dependencies required for the build.

Closes #2167 

**Test plan**

A passing test is included.

I've also tested locally.

Currently it will accept `production=false`, `production=0`, `prod=false`, `prod=0`. Any other value will be considered true, including empty.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -232,6 +232,7 @@ export default class Config {
     await fs.mkdirp(this.tempFolder);
 
     if (this.getOption('production') || (
+        opts.production !== 'false' &&
         process.env.NODE_ENV === 'production' &&
         process.env.NPM_CONFIG_PRODUCTION !== 'false' &&
         process.env.YARN_PRODUCTION !== 'false')) {
@@ -256,7 +257,7 @@ export default class Config {
     this.modulesFolder = opts.modulesFolder;
     this.globalFolder = opts.globalFolder || constants.GLOBAL_MODULE_DIRECTORY;
     this.linkFolder = opts.linkFolder || constants.LINK_REGISTRY_DIRECTORY;
-    this.production = !!opts.production;
+    this.production = !!opts.production && opts.production !== 'false';
     this.offline = !!opts.offline;
     this.binLinks = !!opts.binLinks;
 
--- a/src/config.js
+++ b/src/config.js
@@ -231,12 +231,15 @@ export default class Config {
     await fs.mkdirp(this.cacheFolder);
     await fs.mkdirp(this.tempFolder);
 
-    if (this.getOption('production') || (
-        opts.production !== 'false' &&
+    if (opts.production === 'false') {
+      this.production = false;
+    } else if (this.getOption('production') ||
         process.env.NODE_ENV === 'production' &&
         process.env.NPM_CONFIG_PRODUCTION !== 'false' &&
-        process.env.YARN_PRODUCTION !== 'false')) {
+        process.env.YARN_PRODUCTION !== 'false') {
       this.production = true;
+    } else {
+      this.production = !!opts.production;
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/config.js
```diff
@@ -256,7 +259,8 @@ export default class Config {
     this.modulesFolder = opts.modulesFolder;
     this.globalFolder = opts.globalFolder || constants.GLOBAL_MODULE_DIRECTORY;
     this.linkFolder = opts.linkFolder || constants.LINK_REGISTRY_DIRECTORY;
-    this.production = !!opts.production;
+    this.noProduction = opts.production === 'false' || opts.production === '0';
```
bestander: That is a bit confusing, why do we need both noProduction and production in the class state?

bestander: Also do we want to support 0 and 1 values?

alexbrazier: @bestander Thanks for reviewing. I will make the change tonight to remove the `noProduction` variable as I realise it is not needed. I will also remove support for 0 and 1, and resolve conflicts.

alexbrazier:
@bestander I have made the changes you suggested, and fixed the merge conflict.

src/config.js
```diff
@@ -256,7 +257,7 @@ export default class Config {
     this.modulesFolder = opts.modulesFolder;
     this.globalFolder = opts.globalFolder || constants.GLOBAL_MODULE_DIRECTORY;
     this.linkFolder = opts.linkFolder || constants.LINK_REGISTRY_DIRECTORY;
-    this.production = !!opts.production;
+    this.production = !!opts.production && opts.production !== 'false';
```
bestander: Just a nit, is this this.production assignment needed?
Maybe it should set production to false?

bestander: I mean
```
this.production = false
```

And then all the logic with opts and ENV variables is done above.
Otherwise it is quite confusing where is the source of truth

alexbrazier: @bestander I'm not sure if I understand what you mean. Are you saying to set it to false here, then do everything in the other function?

alexbrazier: It still needs to be set to true when the flag is passed with no env variables set. The other part only sets to true when `NODE_ENV==='production'` and there aren't any overrides.

bestander: It is just weird that we assign same property twice in different ways.
Better to have all the production override logic in one place.

alexbrazier: Could do something like this in one line?
```javascript
    this.production = !!this.getOption('production') ||
      (opts.production !== 'false' &&
      process.env.NODE_ENV === 'production' &&
      process.env.NPM_CONFIG_PRODUCTION !== 'false' &&
      process.env.YARN_PRODUCTION !== 'false') ||
      !!opts.production && opts.production !== 'false';
```
Or
```javascript
    if (opts.production === 'false') {
      this.production = false;
    } else if (this.getOption('production') ||
        process.env.NODE_ENV === 'production' &&
        process.env.NPM_CONFIG_PRODUCTION !== 'false' &&
        process.env.YARN_PRODUCTION !== 'false') {
      this.production = true
    } else {
      this.production = !!opts.production;
    }
```

I prefer the latter. It also means that `production=false` will override the `getOption('production')`, which is probably the desired behaviour.

bestander:
Use your best judgement.
I was just nitpicking about having logic spread.
Thanks for following this up!

alexbrazier:
@bestander I've made those changes 😄

Issue #743, opened 2016-10-11
Yarn init fails on Windows under Git bash / MinGW / Cygwin


lewiswalsh:
Under Git Bash on Windows when entering 'yarn init' I get the following error:

 error Can't answer a question unless a user TTY
    at ConsoleReporter.question (C:\Program Files (x86)\Yarn\lib\reporters\console\console-reporter.js:189:29)
    at Object.<anonymous> (C:\Program Files (x86)\Yarn\lib\cli\commands\init.js:91:38)
    at next (native)
    at step (C:\Program Files (x86)\Yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)
    at C:\Program Files (x86)\Yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:28:20
    at run (C:\Program Files (x86)\Yarn\node_modules\core-js\library\modules\es6.promise.js:87:22)
    at C:\Program Files (x86)\Yarn\node_modules\core-js\library\modules\es6.promise.js:100:28
    at flush (C:\Program Files (x86)\Yarn\node_modules\core-js\library\modules_microtask.js:18:9)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)

Other yarn commands seem to work as expected.

Yarn v0.15.1
Node v6.0.0
Windows 10 Pro 1607

Daniel15:
Seems like this is an issue with Cygwin and MinGW in that they're not 'proper' TTYs: https://github.com/nodejs/node/issues/3006

oliversalzburg:
Well, one thing you could do would be to remove the check ;)

In all seriousness though. It is my understanding that the check is in place because you're trying to make sure that the I/O scenario that is about to happen will work. Does that check actually achieve this? Will the scenario actually fail in MinGW? If not, then a different check should be performed or it should be omitted completely IMHO.

Daniel15:
My understanding is that the check is to stop the process hanging indefinitely when not using a TTY (for example, if running as part of a build process). In theory, if there's no TTY, there's no way to accept input. May be the check should "if it's a TTY **or** if it's MinGW" to handle this case specifically?

jdalton:
FWIW `npm init` works in WSL but `yarn init` errors.

Daniel15:
Yeah, I just noticed that `npm init` is fine. Sounds like we should take a look at how npm is handling it 😄

[...]

PR #2230, opened 2016-12-12
Make yarn interactive commands run in Git Bash/MSYS environment

thetrompf:
The problem is caused by mintty, and it manifests weirdly because in most circumstances you can get interactive node programs to work because of a hack introduced via aliasing/wrapping commands in the bash profile:

```
bkc@DESKTOP-NNSLT74 ~/workspace/yarn-test $ type node
node is aliased to `winpty node.exe'
```

After a bit of digging I found the culprit in `/etc/profile.d/aliases.sh`

```bash
case "$TERM" in
xterm*)
        # The following programs are known to require a Win32 Console
        # for interactive usage, therefore let's launch them through winpty
        # when run inside `mintty`.
        for name in node ipython php php5 psql python2.7
        do
                case "$(type -p "$name".exe 2>/dev/null)" in
                ''|/usr/bin/*) continue;;
                esac
                alias $name="winpty $name.exe"
        done
        ;;
esac
```

Git bash wraps `node`, `ipython`, `php`, `php5`, `psql`, `python2.7` in a program called `winpty`
which means when yarn invokes node, this hack is not invoked as well.

Related to issue #743 

**Test plan**

I'm not sure how to setup an environment that could run an automated test for this case correctly
but I made the following before and after test:

```
bkc@DESKTOP-NNSLT74 ~/workspace/yarn-test $ ./bin/yarn init
yarn init v0.17.10
error An unexpected error occurred: "Can't answer a question unless a user TTY".
info If you think this is a bug, please open a bug report with the information p                                                                                                                                                    rovided in "C:\\Users\\bkc\\workspace\\yarn-test\\yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/init for documentation about this com                                                                                                                                                    mand.
```

After wrapping the process in winpty:

```
bkc@DESKTOP-NNSLT74 ~/workspace/yarn-test $ ./bin/yarn init
yarn init v0.17.10
question name (yarn-test):
```

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -1,8 +1,11 @@
 #!/bin/sh
 basedir=$(dirname "$(readlink "$0" || echo "$(echo "$0" | sed -e 's,\\,/,g')")")
 
+is_msys=0
+
 case `uname` in
   *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
+  *MSYS_NT*) is_msys=1;;
 esac
 
 command_exists() {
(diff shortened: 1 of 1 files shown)


Daniel15:
Awesome, thank you!

PR #2235, opened 2016-12-13
Add "Too many arguments" message to `info` and improve code coverage

skratchdot:
I'm working on some code, and rather than including all my commits in one PR, I'd like to separate a few of the smaller, non-related fixes into their own PR's (for easier review).

**Summary**

This is a very small change to get the code coverage of the `info` command up to 100%.  The currently included versions of `jest` and `babel-jest` still have slightly incorrect coverage reports, but updating them gives `100%` on this file.  I would like to upgrade all the modules in this repo, but it might make more sense for that to be in a different PR? (I didn't include `package.json` or `yarn.lock` updates in this PR)

**Test plan**

1. Run `yarn test`
2. Confirm `__tests__/commands/info.js` has no errors
3. `open ./coverage/lcov-report/index.html`
3. Confirm that the coverage for `src/commands/info.js` has improved

### DIFF
--- a/__tests__/commands/info.js
+++ b/__tests__/commands/info.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import * as reporters from '../../src/reporters/index.js';
 import {run as info} from '../../src/cli/commands/info.js';
 import {BufferReporter} from '../../src/reporters/index.js';
 import Config from '../../src/config.js';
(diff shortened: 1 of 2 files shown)


__tests__/commands/info.js
```diff
@@ -47,7 +48,7 @@ const expectedKeys = [
 ];
 
 test.concurrent('without arguments and in directory containing a valid package file', (): Promise<void> => {
-  return runInfo(['.'], {}, 'local',
+  return runInfo([], {}, 'local',
```
skratchdot: NOTE: this looks like I'm removing the `['.']` case, but there were actually 2 of the exact same test cases in this file.  All I'm doing here is making this test case match the description (and hereby improving the coverage report).

wyze:
Thanks!

PR #2236, opened 2016-12-13
Small aliases refactor.

skratchdot:
I'm working on some code, and rather than including all my commits in one PR, I'd like to separate a few of the smaller, non-related fixes into their own PR's (for easier review).

**Summary**

Locally, I've upgraded my eslint versions, and the current `aliases` file throws lint errors due to the `sort-keys` rule.  This commit fixes the errors by sorting all our keys.  I kept the `shorthands` vs `affordances` separation by using 2 variables and the object spread syntax in the exported value.

**Test plan**

1. I made my changes then ran `yarn build`
2. I ran `./bin/yarn ls` and a few other of the alias commands.
3. Confirmed that the behavior is the same before and after my changes.
4. Run `yarn test`
5. Confirm `__tests__/cli/aliases.js` has no errors
6. open `./coverage/lcov-report/index.html`
7. Confirm that the coverage for `src/cli/aliases.js` is still 100%

### DIFF
--- a/src/cli/aliases.js
+++ b/src/cli/aliases.js
@@ -1,34 +1,39 @@
 /* @flow */
 
-export default ({
-  // shorthands
-  'run-script': 'run',
+const shorthands: { [key: string]: string } = {
   c: 'config',
   i: 'install',
-  ls: 'list',
+  la: 'list',
+  ll: 'list',
   ln: 'link',
+  ls: 'list',
+  r: 'remove',
   rb: 'rebuild',
-  runScript: 'run',
+  rm: 'remove',
   t: 'test',
   tst: 'test',
   un: 'remove',
   up: 'update',
   v: 'version',
+};
 
-  // affordances
+const affordances: { [key: string]: string } = {
   'add-user': 'login',
-  'dist-tag': 'tag',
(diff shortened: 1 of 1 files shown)


skratchdot:
**NOTE:** I didn't remove any values, all I did was re-order them.  I also updated the flow syntax so the "string" values aren't optional.  I can't ever see us adding an alias to a non-string command.

bestander:
I don't think @kittens wanted to overload number of aliases for commands and I agree with him.
For example `yarn list` is short enough and if you use it a lot, it is easy to create your own shortcut on OS level.

skratchdot:
@bestander - it still doesn't let you run `yarn ls`, it'll throw the error message: `did you mean yarn list` per:
https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L126-L134

All I did was fix an eslint warning (which will happen if we ever upgrade our eslint versions).

Gusbenz:
@skratchdot I'm a fan of getting `yarn i` for install and was literally just thinking of adding that this morning. Anyway, thanks!

bestander:
Ah, thanks for clearing up, @skratchdot!
Thanks.

PR #2249, opened 2016-12-14
Adding `docs` (alias: `home`) command

skratchdot:
**Summary**

`npm home [package]` is one of my most used commands.  I really miss having that in `yarn`.  I've added `docs` as the main command (same as npm)- and I've also added the `yarn home` alias.  It works about the same as the npm command except I'm using [opn](https://www.npmjs.com/package/opn) instead of [opener](https://www.npmjs.com/package/opener), and I'm also using `--app` as the flag instead of `--browser`.  I can discuss more if you want.  Really hoping you can pull this in (as it's a command I really miss).  I'm not sure if you've had discussions about this in the past.  One other issue I had was that some of our dev dependencies are out of date.  I had issues w/ flow-types based on the version of jest we are using vs the flow type versions we are using.  Due to this fact, I had to use some `$FlowFixMe` comments.  Also, we are using an older version of jest, so I used `jest.clearAllMocks()`.  In newer versions of jest, I'd need to use `jest.resetAllMocks()`

**Test plan**

1. Run `yarn run test`
2. Confirm there are no lint or flow errors.
3. Confirm code coverage is 100% on all the `doc` releated files
4. Run `yarn build`
5. Do some smoke testing with the cli that was built:
  - `./bin/yarn docs --help`
  - `./bin/yarn docs`
  - `./bin/yarn home`
  - `./bin/yarn docs --app "Firefox" react flow redux`

**Commit Message**

- works similar to npm's `docs` command w/ the following caveats
  - using "opn" instead of "opener" library
  - using "--app" instead of "--browser"
- currently ignoring some jest flow-types. will update in another PR

**Quick Demo**

yarn-docs

### DIFF
--- a/__tests__/commands/__snapshots__/docs.js.snap
+++ b/__tests__/commands/__snapshots__/docs.js.snap
@@ -1,0 +1,240 @@
+exports[`test setFlags() modifies our commander object 1`] = `
+Object {
+  "calls": Array [
+    Array [
+      "docs [packages ...] [flags]",
+    ],
+  ],
+  "instances": Array [
+    [Function],
+  ],
+}
+`;
+
+exports[`test setFlags() modifies our commander object 2`] = `
+Object {
+  "calls": Array [
+    Array [
+      "-a, --app <app>",
+      "the app that is used to open websites. Defaults: macOS=\"open\", Windows=\"start\", Others=\"xdg-open\"",
+    ],
+  ],
+  "instances": Array [
+    [Function],
+  ],
+}
+`;
+
+exports[`test with 1 argument and a invalid request 1`] = `
+Object {
+  "calls": Array [],
+  "instances": Array [],
+}
+`;
+
+exports[`test with 1 argument and a invalid request 2`] = `
+Object {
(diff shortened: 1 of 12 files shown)


skratchdot:
**NOTE:** I added a new dependency to the package.json file, but didn't check in yarn.lock.  Not sure what the policy is here.  I didn't want to cause merge conflicts w/ existing PRs, and figured a build manager (or someone w/ commit access) would take care of updates like that.

skratchdot:
Also, the build should be working.  I `git push -f origin docs-command` to remove 1 extra piece of unused code, and I think it caused an issue (because the tests were passing in my initial commit).  I wanted to keep the commit log clean.

Is there anyway to kick off another CI build?

kittens:
From my comment in #778:

> We've deliberately decided not to add these sorts of commands to Yarn. There's a lot of them that don't actually see much use and the maintenance burden isn't really worth it. If the functionality of the commands is so simple (just opening a browser) then you can just do it yourself. Thanks for the report!

See also #914.

Issue #2261, opened 2016-12-15
Failing test due to platform ident bug


polizz:
**Bug**
/__tests__/commands/run.js (handles bin scripts) fails due to escaping issue.

**Reproduce**
1. Fork repo and run tests.

**What is the expected behavior?**
Mac (Darwin) should not receive Windows backslash path transformation.

**Platform**
Node 6.9.1, 7.2.1
Mac 10.12.1

Fix incoming.

Daniel15:
#2262 should fix this.

PR #2262, opened 2016-12-15
Increased specificity of Windows platform check

polizz:
**Summary**
Extended regex to match only Windows. Previous regex was matching Darwin in addition to Windows.

yarn run test > all tests pass.

#2261

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -131,3 +131,11 @@ export async function run<T, R>(
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
   }
 }
+export const setPlatform = (platform : string) : string => {
+  const previous = process.platform;
+  Object.defineProperty(process, 'platform', {
+    value: platform,
+  });
+
+  return previous;
+};
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -68,7 +68,7 @@ export async function run(
     for (const action of actions) {
       const cmd = scripts[action];
       if (cmd) {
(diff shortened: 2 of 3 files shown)


src/cli/commands/run.js
```diff
@@ -68,7 +68,7 @@ export async function run(
     for (const action of actions) {
       const cmd = scripts[action];
       if (cmd) {
-        const isWin = /win/.test(process.platform);
+        const isWin = /win32/.test(process.platform);
```
Daniel15: Could this just be `process.platform === 'win32'`? Looks like that's what Node.js itself does, so it should be safe: https://github.com/nodejs/node/blob/db1087c9757c31a82c50a1eba368d8cba95b57d0/lib/os.js#L6

polizz: Certainly. I noticed another section in execute-lifecycle that does just that.

__tests__/commands/_helpers.js
```diff
@@ -131,3 +131,11 @@ export async function run<T, R>(
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
   }
 }
+export const setPlatform = (platform : string) : string => {
+  const previous = process.platform;
+  Object.defineProperty(process, 'platform', {
```
Daniel15: I wish Node.js had a decent dependency injection framework, this is super hacky 😛

polizz: I'm going to redo that. It's not effective at all 😰 I originally was pondering just passing platform into **run** so the test could directly send it, but, to your point, that seems corny.

Issue #2112, opened 2016-12-02
can not add `lockfile` package


lili21:
<img width="566" alt="2016-12-02 2 34 53" src="https://cloud.githubusercontent.com/assets/4456380/20825015/3a0aafb0-b89c-11e6-9c4e-40e562075b46.png">

[lockfile](https://www.npmjs.com/package/lockfile).

nodejs: v6.9.1
os: macOS Sierra 10.12

tuures:
workaround: https://github.com/yarnpkg/yarn/issues/2220#issuecomment-266847731

iredchuk:
Caused by this (very old) bug from commander.js: https://github.com/tj/commander.js/issues/346

PR #2268, opened 2016-12-15
Fix adding package with name like one of the options

iredchuk:
Fixes issue when yarn cannot add/remove package which name is equal to one of the CLI options (e.g. **lockfile**, **progress**).

Fixes issue #2112.

The source of the error is this bug from commander.js: https://github.com/tj/commander.js/issues/346
when the first argument after the first two **process.agrv**s (i.e. **node yarn**) cannot have the same name as options defined with commander.

The solution is to parse args along with the command name and the remove command name from the args.

`$ yarn add lockfile`
should work correctly

I added basic integration tests, but IMO would probably be nicer to extract the whole arguments parsing to a separate module and unit test it more intensively, but I found it pretty hard with the current implementation. This might be done later on a separate improvement PR.

### DIFF
--- a/__tests__/fixtures/index/run-add-lockfile/package.json
+++ b/__tests__/fixtures/index/run-add-lockfile/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "test_add_lockfile",
+  "version": "1.0.0",
+  "license": "UNLICENSED"
+}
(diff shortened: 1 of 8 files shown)


torifat:
@iredchuk does it work global? Can you add test cases for `yarn global <command>`?

iredchuk:
@torifat Sure, [added](https://github.com/iredchuk/yarn/blob/fix-add-package-with-name-like-an-option/__tests__/index.js#L76)

torifat:
@iredchuk weird, test failing. Can you please rebase and push again?

iredchuk:
@torifat My bad: the test for **global** was conflicting with another one when running concurrently. Should be fixed now.

torifat:
@iredchuk Ah, Thanks 👍

PR #2269, opened 2016-12-16
Add use of reporter lang method

msaglietto:
**Summary**
Continuing with #1387 now most of the files are cover .. 
Some missing are:

https://github.com/yarnpkg/yarn/blob/master/src/lockfile/parse.js#L178 that dont have access to a reporter 

https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js it would be nice to translate the commander options .. but the definition of the options needs to be moved to after the reporter creation 

And if there is any other translation missing let me know 

**Test plan**

Just ran `yarn run test`  and hope for the best

### DIFF
--- a/__tests__/util/git.js
+++ b/__tests__/util/git.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import Git from '../../src/util/git.js';
+import {NoopReporter} from '../../src/reporters/index.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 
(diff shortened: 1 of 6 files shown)


msaglietto:
It seams the test, not only take to long to run, are not reliable  ... 

After a few runs locally it fails  on `should expose `npm_config_argv` environment variable to lifecycle scripts for back compatibility with npm (#684)` if I run `yarn run test`

But if I run `npm run test` locally it fails on `install should respect --no-bin-links flag` also 

And now the build fail in `install with arg that has install scripts` ... that it seems not related to my changes

bestander:
Thanks for the PR, let's merge it, could you rebase it please?

msaglietto:
@bestander Updated ...  5 test are failing locally and  1 appveyor fail .. but the others pass .. so not sure what to trust

bestander:
Some tests can depend on network so they might fail, if other CI are fine then it should be safe

bestander:
https://github.com/yarnpkg/yarn/issues/2439 caused by this PR.
Probably an older bug where # is not in ...vars caused escaping of the whole string

PR #2273, opened 2016-12-16
Set tgz files as binary in git

SimenB:
**Summary**

I kept on getting `warning: CRLF will be replaced by LF in __tests__/fixtures/add/install-deterministic/mirror-for-offline/mime-db-1.0.3.tgz.` after every single command in the repo.

**Test plan**
N/A


torifat:
Thanks

Issue #2293, opened 2016-12-19
Not all install scripts from duplicate nested dependencies run


onemen:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Not all install scripts from duplicate nested dependencies run, only the script from one copy is run.

**If the current behavior is a bug, please provide the steps to reproduce.**
Install package that include more than one instance of the dependency.
For example: 
  a1 > b1 > c1
  b2 > c1
  d1 > a1 > b1 > c1
  d1 > b1 > c1
  We get 4 hoisted packages: a1, b2, c1, d1, and 2 nested packages for b1: a1/b1, d1/b1

The current behavior that all the install scripts from dependency a1/b1 doesn't run!
 

I will post pull request with a test soon.

**What is the expected behavior?**
Existing Install scripts should run for all installed dependencies.

**Please mention your node.js, yarn and operating system version.**
node: v7.2.0
yarn: latest master
OS: Windows 10

PR #2294, opened 2016-12-19
Not all install scripts from duplicate nested dependencies run

onemen:
Not all install scripts from duplicate nested dependencies run, only the
script from one copy is run.

This test is failing with the current code see #2293


bestander:
@onemen thanks for sending this PR!
Could you please move it to integration.js?
And mark them as skipped while they are failing with a link to the open issue please.

onemen:
I hope to have more time later this year

bestander:
Cool, let's reopen this PR when it can be acted upon

PR #2295, opened 2016-12-19
Fix check of scoped packages' sub-dependencies

voxpelli:
**Summary**

When running `yarn check` on a freshly installed project with an extensive use of scoped packages yarn complains that some scoped packages uses the wrong version of another scoped package.

***What is the current behavior?***

With a module structure like following:

```
node_modules/
  @scope/
    subdependency@2.x/
    majordependency/
      node_modules/
        @scope/
          subdependency@1.x/
```

The following complaint is received when running `yarn check` (confirmed with latest `master`):

```
error @scope/majordependency##@scope/subdependency@^1.0.0 doesn't satisfy found match of @scope#subdependency@2.0.0
```

***What is the expected behavior?***

With the above mentioned module structure it would be expected that `yarn check` instead passes, as `@scope/majordependency` has the correct version of `@scope/subdependency` installed beneath it.

***Cause of bug***

`yarn check` when splitting the paths of a dependency expects `node_modules` to be inbetween every path part, but for scoped packages there should be no `node_modules` folder inbetween the scope and package name parts.

So when `yarn check` looks for `node_modules/@scope/node_modules/majordependency/node_modules/@scope/subdependency/package.json` it should instead be looking for that very same path but with `@scope/majordependency` instead of `@scope/node_modules/majordependency` in it.

**Test plan**

1. In a project, add a scoped module, lets call it `@scope/subdependency`
2. Run `yarn check`. Everything should be fine both with and without this PR.
3. Add another scoped dependency that also relies on the module from `1.`, but which requires an incompatible version of it. Lets call this `@scope/majordependency`. You should now have a similar module setup as outlined above.
4. Run `yarn check`. It should now complain without this PR, but pass with this PR.
5. (Manually remove `node_modules/@scope/majordependency/node_modules/@scope/subdependency/package.json` and run `yarn check` again. It should now correctly complain about the missing dependency.)

**Additional notes**

I'm seeing an unrelated failed test in the current master locally, both with and without this PR, but I've ignored it since it shouldn't be related to this.

I've not added any new tests because I couldn't really find the place where the tests for `yarn check` are. If someone could point me in the right direction then I could happily look at adding some tests for this.


bestander:
Thanks, @voxpelli!
The tests should be more stable now, could you rebase?

As for adding a new test, yes please, we would need it.
We usually check the `check` command right within install/add routines, e.g. https://github.com/yarnpkg/yarn/blob/38c7aa591b2352a2b31a94ef62e40bc4b4501920/__tests__/commands/add.js#L167-L173

voxpelli:
Thanks for the feedback @bestander, I've rebased now + created a test that I've verified that it failed before my fix and works after my fix.

Still not a 100% sure on whether the test is what is wanted, eg:

* Should this test really be in the install test when it's really testing the check method? Simplest for now at least
* Is it okay to just use a random scoped package that one has come across in the tests? It looked like the other tests included real live packages so I just took the first public scoped package I could think of where this error could arise in practice.

Happy to tweak the tests and of course to tweak the fix itself if there's any additional feedback on it.

bestander:
Thanks, @voxpelli.
You are right, ideally check should have a dedicated isolated test suite but for most cases having a quick e2e seems all right for me, we just pay a higher price running e2e tests instead of unit.

Yeah, it is ok to use a random small module from npm for a quick test.
There is a risk that they will break sometimes but our test suite has http caches committed so it does not affect stability all the time.

Let's get this in, just a small not

__tests__/commands/install/integration.js
```diff
@@ -737,6 +737,10 @@ test.concurrent('install will not overwrite files in symlinked scoped directorie
   });
 });
 
+test.concurrent('install of scoped package with subdependency conflict should pass check', (): Promise<void> => {
+  return runInstall({}, 'install-scoped-package-with-subdependency-conflict');
```
bestander: nit: could you add an explicit `check` call for this test?
Even though all install script have implicit check running it is better from maintainability point of view to have an explicit expectation from the test run, e.g.:
```
    let allCorrect = true;
    try {
      await check(config, reporter, {integrity: false}, []);
    } catch (err) {
      allCorrect = false;
    }
    expect(allCorrect).toBe(true);
```

voxpelli:
@bestander Added the explicit `check` + rebased on top of `0.19-stable`

bestander:
Thanks a lot @voxpelli.
If you want to get it into 0.19 do you want to cherry-pick and create a PR for that branch?

PR #2298, opened 2016-12-19
Reverted test due to inability to mock process.platform (#2262)

polizz:
**Summary**

This test is too complicated and does not thoroughly test the run command's handling of win32 vs. posix script command mangling.

It is not possible to test win32 and posix from one platform because the path module employed in the test alters its export based on process.platform in Node bootstrap, well before one can mock it. i.e. the test assertion will always be looking for the local platform's path delimiters etc.

One could hardcode path information, but it still feels risky to be changing the process.platform from within one test, for one test.

**Test plan**

The original test asserted that window cmd line formatting is properly applied or not applied, for only the target platform the tests are executed on - that will have to be good enough.

Tests pass.

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -131,11 +131,3 @@ export async function run<T, R>(
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
   }
 }
-export const setPlatform = (platform : string) : string => {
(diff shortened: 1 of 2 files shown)


Daniel15:
Thanks. I guess running the test on AppVeyor is probably sufficient for testing Windows, as it'll run in a real Windows environment.

PR #2305, opened 2016-12-20
Fix cache test timeout

wyze:
**Summary**

Fixes an issue with tests timing out due to a directory not being created properly for the `cache` command tests.

Also, this removes the `.only` in the run test that snuck in there in #2298.

**Test plan**

All tests should now pass.


wyze:
It appears this was only an issue locally on OSX.

bestander:
The cwd location issue is fixed now.
Could you rebase?

bestander:
Looks like it is not relevant anymore?
I'll close the PR

PR #2307, opened 2016-12-20
Upgrade to flow@0.37 / fix some typings

ryyppy:
* Adds inquirer libdefs for more exact typings
* Refine / Fix some flowtype errors (mostly in `__tests__`)
* Fix a small bug in `src/util/map.js` found by flow@0.37

**Summary**

This PR adds flow-typed definitions for `inquirer`. Nice thing is, I didn't even need to touch the code after adding the definitions (I played around with it and could produce errors when passing wrong parameters to `inquirer.prompt()`).

Also I invested some time to upgrade `flow` from `0.33` to the most recent version `0.37.3`, fixing some typing issues / bugs along the way.

**Test plan**

When I originally pulled the source from `master`, I realized that 11 tests were failing.
So I was doing two test runs, one testing the HEAD and one testing my modifications.

The same number of tests failed.
Also I made sure that `flow@0.37.3` type checks the source properly.


ryyppy:
@bestander What is happening with CI right now? Any open issue tackling this problem? Seems that several tests are failing due to some changes

bestander:
@ryyppy, I fixed this, could you rebase please?

ryyppy:
Hmm, not sure what's happening here:

image

package.json
```diff
@@ -16,6 +16,7 @@
     "defaults": "^1.0.3",
     "detect-indent": "^4.0.0",
     "diff": "^2.2.1",
+    "flow-bin": "^0.37.3",
```
bestander: That is the reason why Circle failed.
I'll send a quick fix

ryyppy:
Awesome, thanks 👍

Issue #2083, opened 2016-11-30
Error: Couldn't find the binary git


jdmunro:
I am using:

* yarn v0.17.9
* macOS Sierra 10.12.1
* Node v6.9.1

**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Trying to install packages using yarn. Outputs a strange error message:

```
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Couldn't find the binary git
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn --pure-lockfile
```

Outputs the above error message.

Git is installed:

```
Admins-Mac-mini:BWA-BETA203-FAST admin$ whereis git
/usr/bin/git
```

**What is the expected behavior?**

Yarn should install the packages without emitting an error about git. If I run this command using an older version of yarn, v0.16.1, everything is fine.

kittens:
What's the result of `which git`?

jdmunro:
```
Admins-Mac-mini:~ admin$ which git
/usr/local/bin/git
```

That's a different path to ```whereis git```.

emolchanov:
+1 - same problem macOS 10.12.1

```
yarn --version
0.17.10

node -v
v6.9.1

which git                           
/usr/local/bin/git

whereis git
/usr/bin/git

```

and on windows 7 and 10 with [git-scm](https://git-scm.com/)

Congelli501:
I have the same problem under ubuntu 16.04, dans yarn 18.1.

The reported error is bad:
The node spawn call return a "ENOENT" error (https://github.com/yarnpkg/yarn/blob/master/src/util/child.js#L41)
This is not caused by the git binary not found, but by the `cwd` given in `opts` witch is not found.


This is because of the `supportsArchive` boolean:
 - either archive is not supported, and everything is cloned locally.
 - or archive is used and nothing is cloned

However, supportArchive can change while nothing was cloned (https://github.com/yarnpkg/yarn/blob/master/src/util/git.js#L338).

vitch:
Same thing happened to me after a `yarn cache clean` (trying to get a git branch dependency to update).

Removing the `node_modules` folder and running `yarn` again seemed to work around the issue.

flocsy:
this happens here too after `yarn cache clean`

emmanuel-isaac:
I am experiencing this issue too. I started seeing it after running ```yarn cache clean && yarn clean```

[...]

PR #2320, opened 2016-12-22
fix 2083 supportsArchive can be changed with nothing beeing cloned

Congelli501:
Fix bug #2083
The git cloner could start using archive mode and switch back to local clone mode without cloning anything.
As a result, comands are spawned with a bad (non existent) cwd, leading to a ENOENT error.
This error is ms-interpreted a the lack of the git binary (https://github.com/yarnpkg/yarn/blob/master/src/util/child.js#L41)

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -338,7 +338,10 @@ export default class Git {
       }
 
       // `git archive` only accepts a treeish and we have no ref to this commit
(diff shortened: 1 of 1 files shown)


bestander:
Looks safe enough.
Could you rebase on master please and provide a Test/Repro plan if possible?

PR #2327, opened 2016-12-22
Use sync version of fs.symlink. Fixes #2326

### LINKED ISSUE #2326: Bug: Bad symlink creation for global bin installation

I have a global prefix set via the following `.npmrc`:

```
$ cat ~/.npmrc
prefix=${HOME}/.local
cache=${HOME}/.cache/npm
```

If I try to install any global package with a binary, I get an error:

```
$ yarn global add fastrealpath
yarn global v0.18.1
success Installed fastrealpath@0.1.1 with binaries:
      - realpath
error An unexpected error occurred: "ENOENT: no such file or directory, chmod '/home/fortes/.local/bin/realpath'".
info If you think this is a bug, please open a bug report with the information provided in "/home/fortes/.config/yarn/global/yarn-error.log".
```

It looks like an invalid symlink is created during the installation:

```
$ ls -l $(yarn global bin)
l?????????? ? ?      ?         ?            ? realpath
```

Install should complete successfully without an error. Symlink in `$(yarn global bin)` should be valid.

```
$ node --version
v4.6.1
$  yarn --version
0.18.1
$  uname -a
Linux localhost 3.14.0 #1 SMP PREEMPT Tue Nov 15 18:25:51 PST 2016 x86_64 GNU/Linux
```

fortes:
See Bug #2326 for a full explanation, but the issue is that for some reason `fs.symlink` is creating invalid symlinks during a global binary installation. This change switches to use `fs.symlinkSync`, which fixes the issue on my machine.

This is my first time poking around the yarn codebase, so I'm sure I did something wrong here. I

**Test plan**

I don't really know how to automate testing here. On my local machine global installation fails without this change. With this change it is successful:

```
yarn global v0.19.0-0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Installed fastrealpath@0.1.1 with binaries:
      - realpath
Done in 0.69s.
```

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -27,7 +27,8 @@ export async function linkBin(src: string, dest: string): Promise<void> {
     await cmdShim(src, dest);
   } else {
     await fs.mkdirp(path.dirname(dest));
-    await fs.symlink(src, dest);
+    // Async symlink creates invalid symlinks, see issue #2326
+    fs.symlinkSync(src, dest);
     await fs.chmod(dest, '755');
   }
 }
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -25,6 +25,7 @@ export const mkdirp: (path: string) => Promise<void> = promisify(require('mkdirp
 export const exists: (path: string) => Promise<boolean>  = promisify(fs.exists, true);
 export const lstat: (path: string) => Promise<fs.Stats> = promisify(fs.lstat);
 export const chmod: (path: string, mode: number | string) => Promise<void> = promisify(fs.chmod);
+export const symlinkSync: (target: string, path: string, type?: 'dir' | 'file' | 'junction') => void = fs.symlinkSync;
 
 const fsSymlink: (
   target: string,


fortes:
Test failure due to flow error, `fs.symlinkSync`'s final argument is not typed as optional. I do not know how to fix.

Daniel15:
If `fs.symlink` is creating an invalid symlink, that sounds like a Node.js bug? Maybe open a bug in the Node.js bug tracker?

fortes:
@Daniel15 I'm not actually sure it's a node bug, I just hacked around and tried to get something to work. Would love it if someone who actually understands this stuff took a look.

bestander:
This does not make much sense.
What could go wrong with async symlink is that event loop gets released while symlink is created and some other code may be executing that breaks the installation.
I don't think patching with Sync version is the correct way to address the issue

fortes:
> I don't think patching with Sync version is the correct way to address the issue

@bestander What is the correct way to address the issue?

bestander:
No idea at this moment, I would really appreciate if you could help us out finding out the reason

fortes:
Unfortunately, it's not a straightforward node bug, because the following works on my system:

```js
  fs.symlink(SOURCE, DEST, err => {
    if (err) {
      console.error("symlink error", err.message);
      return;
    }

    fs.chmod(DEST, "755", e => {
      if (e) {
        console.error(e.message);
      }
    });
  });
```

Will need to dig in more to see what's going on. @bestander does `yarn global add` work for you with a local prefix? I don't have a Mac, so I've only done this on Linux machines (repros on Debian Jessie & Stretch).

bestander:
@fortes, which command do you want me to check on windows?

fortes:
@bestander Are you able to `yarn global add fastpath` when using a custom prefix? I don't know how env variables work on windows, but does something like:

`PREFIX=~/.local yarn global add fastpath` 

Work?

bestander:
@fortes,

I tried this:
```
$ set PREFIX=C:\Users\besta\work\temp
$ node -e "console.log(process.env.PREFIX)"
C:\Users\besta\work\temp
$ yarn global add fastpath
yarn global v0.20.0-0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning undefined has no binaries
Done in 2.10s.
```
Looks a bit odd on master branch

bestander:
Tried with v23.0.0 on Mac, adding fastparh globally works
```
PREFIX=~/.local yarn global add fastpath
yarn global v0.23.0-0
warning No license field
[2/4] 🚚  Fetching packages...
warning "fastpath@2.1.0" has no binaries
warning No license field
✨  Done in 2.11s.
```

bestander:
There is not enough evidence that symlinkSync is more stable than async symlink.
I'll close the PR, feel free to add more comments and tests.

fortes:
Whatever bug was happening in the underlying system seems to have been fixed, I no longer repro this with latest versions of Node / Debian / Yarn.

PR #2338, opened 2016-12-25
updated tests caches

bestander:
Updated tests caches to get CI green again


bestander:
Selfmerging this, only caches were updated

PR #2362, opened 2016-12-31
Tests stability improvements

bestander:
**Summary**

I am working on Ubuntu on Windows and had to fix a few issues with tests that were broken.
 
- cache tests were slowed down because all of them were copying root folder to a temp location
- if node is installed globally with root then `global` tests were failing for non root users
- fixed a weird Ubuntu on Windows issue (going to report this over there)

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -12,7 +12,6 @@ import Config from '../../src/config.js';
 
 const stream = require('stream');
 const path = require('path');
-const os = require('os');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'install');
 
@@ -81,21 +80,12 @@ export async function run<T, R>(
 
   let cwd;
   if (fixturesLoc) {
-    // if fixture loc is not set create a tmp dir so that we don't copy CWD during test run
     const dir = path.join(fixturesLoc, name);
-    cwd = path.join(
-      os.tmpdir(),
-      `yarn-${path.basename(dir)}-${Math.random()}`,
-    );
-    await fs.unlink(cwd);
+    cwd = await fs.makeTempDir(path.basename(dir));
     await fs.copy(dir, cwd, reporter);
   } else {
-    cwd = path.join(
-      os.tmpdir(),
-      `yarn-${Math.random()}`,
-    );
-    await fs.unlink(cwd);
-    await fs.mkdirp(cwd);
+    // if fixture loc is not set then CWD is some empty temp dir    
+    cwd = await fs.makeTempDir();
   }
 
   for (const {basename, absolute} of await fs.walk(cwd)) {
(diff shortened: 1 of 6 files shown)


src/cli/commands/remove.js
```diff
@@ -41,13 +41,13 @@ export async function run(
       const registry = config.registries[registryName];
       const object = rootManifests[registryName].object;
 
-      for (const type of constants.DEPENDENCY_TYPES) {
```
bestander: This construct was crashing node on ubuntu on windows.
I know it is not a good idea to patch Yarn because of a very niche environment but I could not find a nicer workaround

bestander: Actually this only happens only on Node 7 on ubuntu on windows.
Probably needs to be reported with Node.

bestander: Yeah, I'll revert this change and wait for Node to be stable then

bestander: Most likely related to https://github.com/nodejs/node/issues/9419

__tests__/commands/_helpers.js
```diff
@@ -79,13 +79,24 @@ export async function run<T, R>(
...
+    cwd = path.join(
+      os.tmpdir(),
+      `yarn-${path.basename(dir)}-${Math.random()}`,
+    );
+    await fs.unlink(cwd);
+    await fs.copy(dir, cwd, reporter);
+  } else {
+    cwd = path.join(
+      os.tmpdir(),
+      `yarn-${Math.random()}`,
```
Daniel15: Does Node.js provide a way to call the system's native temporary directory creation function (such as [mkdtemp](http://man7.org/linux/man-pages/man3/mkdtemp.3.html) on GNU/Linux or [GetTempFileName](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364991(v=vs.85).aspx) on Windows)? The issue with creating temporary directories like this is that two different processes (for example, running different tests in parallel) could generate the same random number.

Failing that, a GUID would be better too.

Daniel15: Just found that it does support it natively in Nodejs 5.10+: https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback

bestander:
Yeah, sounds reasonable.
Could not find a polyfill, I'll implement a with rand it then

bestander:
Actually https://github.com/nodejs/node/issues/6142 makes me think that mdirtmp is quite useless for our case.
If I want to pass a prefix than I need to implement half of what fs.mkdtemp does and because I need to provide the polyfill anyway for Node 4 I'd better not to use Node's wrapper

bestander:
@Daniel15, how about now?

Daniel15:
This looks good to me, as `Date.now` should reduce the risk of collisions. It would be better to use the native functionality, but if Node.js doesn't expose that very well, it's fine to use some JS code for it. 

Another option is to use something like [node-tmp](https://github.com/raszi/node-tmp).

In any case, feel free to merge this whenever you like :)

PR #2366, opened 2017-01-01
Fixed global module functionality for Windows.

nimerix:
cmd-shim appends ".cmd" to the destination path when called. Yarn calls cmd-shim with the destination parameter `<package-name>.cmd`, resulting in an improperly formed shim: `<package-name>.cmd.cmd` referencing the wrong source: `<package-name>.cmd`. If the ".cmd" extension is removed from dest prior to calling cmd-shim, the result is `<package-name>.cmd` and `<package-name>`, respectively. This reliably fixes the behavior of global add and global remove in Windows.

**Summary**
Yarn does not create global modules properly for Windows users (see commit description).

The reason for the ".cmd.cmd" is because [cmd-shim](https://github.com/ForbesLindesay/cmd-shim) appends ".cmd" to the destination:
```js
// lines 157-158 for index.js
  fs.writeFile(to + ".cmd", cmd, "utf8", then)
  fs.writeFile(to, sh, "utf8", then)
```
My motivation for fixing this bug is simple: I like using Yarn. I use Windows at work. I can't use Yarn at work. I'd like to use Yarn at work.  

**Test plan**

Here are some before and after demonstrations. Notice the ".cmd.cmd" files being left behind when the module is removed.

Before

After


Daniel15:
Thanks! Could you please add a unit test too, if possible? If the test only works on win32, you can gate it in `process.platform === 'win32'`.

As an aside, it looks like we're still creating the .cmd files in the wrong directory - They should be created in %LocalAppData%\Yarn or similar, not the Node.js installation directory.

nimerix:
@Daniel15 Hey Daniel. 

In this last change, global add now uses %LOCALAPPDATA%\Yarn\.bin if it exists in the system path (If Yarn was installed via .msi or Chocolatey), and falls back to the node directory otherwise (e.g. `npm i -g yarn`). 

I'll have the unit tests shortly.

nimerix:
Rebased to current master branch, added tests and fixed a new bug introduced by fdade5d74c95f251a10ab8582eb0ccafbc3b4764.

Specifically: 
```js 
# src/cli/commands/global.js 
        if (process.platform === 'win32' && dest.indexOf('.cmd') !== -1) {
          await fs.rename(dest + '.cmd', dest);
        }
```
Since `dest + '.cmd'` is not checked for existence before calling `fs.rename`, the code only works If an invalid windows shim is present (`<package_name>.cmd.cmd`) and results in an install failure otherwise, since `fs.rename` is called on a non-existent path.

bestander:
This bug is fixed now.
But thanks for the PR.

Daniel15:
Is this actually fixed though? @nimerix  Can you please see whether you're still hitting this issue?

bestander:
I tested cowsay module on my windows device

PR #2369, opened 2017-01-03
[CI] flow is a dev dependency, removed it from prod dependencies, broken in #2307

bestander:
A quick fix to package building


bestander:
Self-merging to get trunk green

PR #2375, opened 2017-01-03
Remove failed to install optional modules from node_modules

### LINKED ISSUE #2274: optional dependencies that fail builds should be removed from node_modules

feature/bug

Optional Dependencies are first linked (files are copied from cache to node_modules) and then their install scripts are executed.
If install script fails this is considered OK but node_modules then has a half built folder.

I created test `optional dependency that fails build should still be installed` in integration.js that covers current behavior but I think it should be changed.

I think Yarn should remove failed to install optional modules from node_modules.

Yarn 0.18.0

lgeiger:
**Summary**

> Optional Dependencies are first linked (files are copied from cache to node_modules) and then their install scripts are executed.
If install script fails this is considered OK but `node_modules` then has a half built folder.

Yarn now removes failed to install optional modules from `node_modules`.

Addresses #2274

**Test plan**

I manually tested this PR and it works as expected.
~~Unfortunately the integration tests fail but I can't figure out why. Any help would be really appreciated.~~ The integration tests have been adjusted accordingly.


__tests__/commands/install/integration.js
```diff
@@ -764,18 +764,18 @@ if (process.platform !== 'darwin') {
...
 
+// Covers current behavior, issue opened whether this should be changed https://github.com/yarnpkg/yarn/issues/2274
 test.concurrent('a subdependency of an optional dependency that fails should be installed',
   (): Promise<void> => {
-    return runInstall({}, 'should-install-failing-optional-sub-deps', async (config) => {
-      assert.ok(await fs.exists(path.join(config.cwd, 'node_modules', 'optional-failing')));
-      assert.ok(await fs.exists(path.join(config.cwd, 'node_modules', 'sub-dep')));
+    return runInstallWithoutCheck({}, 'should-install-failing-optional-sub-deps', (config) => {
+      assert.equal(fsNode.existsSync(path.join(config.cwd, 'node_modules', 'optional-failing')), false);
+      assert.equal(fsNode.existsSync(path.join(config.cwd, 'node_modules', 'sub-dep')), true);
```
lgeiger: Sub-dependencies only required by the failing module still aren't deleted.

__tests__/commands/install/integration.js
```diff
@@ -764,18 +764,18 @@ if (process.platform !== 'darwin') {
...
 }
 
-// Covers current behavior, issue opened whether this should be changed https://github.com/yarnpkg/yarn/issues/2274
-test.concurrent('optional dependency that fails to build should still be installed',
+test.concurrent('optional dependency that fails to build should not be installed',
   (): Promise<void> => {
-    return runInstall({}, 'should-install-failing-optional-deps', async (config) => {
-      assert.ok(await fs.exists(path.join(config.cwd, 'node_modules', 'optional-failing')));
+    return runInstallWithoutCheck({}, 'should-not-install-failing-optional-deps', (config) => {
+      assert.equal(fsNode.existsSync(path.join(config.cwd, 'node_modules', 'optional-failing')), false);
```
bestander: Why did you switch to `existsSync` from await `fs.exists`?

__tests__/commands/_helpers.js
```diff
@@ -28,6 +28,18 @@ export const runInstall = run.bind(
   [],
 );
 
+export const runInstallWithoutCheck = run.bind(
```
bestander: Why would you need install without check?

__tests__/commands/install/integration.js
lgeiger: I did it primarily out of two reasons:
1. If I recall correctly `fs.exists` throws if the path is empty
2. `fs.exists` is deprecated while `fs.existsSync` isn't

__tests__/commands/_helpers.js
lgeiger: Since the `optional-failing` module will be removed from `node_modules`, `check` will fail.

bestander: That might be a problem, check should not fail after a correct install

lgeiger: Oh I see. So just removing the folder from `node_modules` isn't enough and will cause check to fail since the module isn't installed.

__tests__/commands/install/integration.js
lgeiger: The wrapped method uses Node's `fs.exists` underneath which is deprecated: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

That said I'm happy to change it for consistency 👍

bestander:
I suppose there needs to be a fix in Check that would ignore optional deps
if they are missing

__tests__/commands/install/integration.js
bestander: Thanks, I did not know that.
We'll probably do a one-off codemod to change to fs.access everywhere

rgbkrk:
What's reflected in the `yarn.lock` if the optional dependency is not kept?

bestander:
Lock file represents full dependency graph.
If you install with --production it will still contain dev dependencies.
Same thing for failed optional dependencies

rgbkrk:
Ok, that makes sense since many times the optional dependencies are platform specific modules (darwin, windows, linux, etc.). Thanks @bestander and @lgeiger!

PR #2404, opened 2017-01-06
[CI] supported non stable releases by default

bestander:
See https://github.com/yarnpkg/website/pull/338 for the improved workflow that allows people to get unstable releases.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -49,11 +49,11 @@ deployment:
     tag: /v[0-9]+(\.[0-9]+)*/
     owner: yarnpkg
     commands:
-      - ~/ghr/ghr --username yarnpkg --repository yarn --token $KPM_CIRCLE_RELEASE_TOKEN v$(dist/bin/yarn --version) artifacts/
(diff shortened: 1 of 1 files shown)


bestander:
Merging this one to make 0.19 release

Daniel15:
Thanks @bestander! This is a good start. We need to be a bit cleverer about whether to mark it as an RC. Currently it marks all releases as RC, but in reality I think we want to do something like this:
 - If it's a new minor release (x.x.0), mark it as RC
 - If the previous bugfix release was an RC, mark it as RC (for example, if 0.19.0 is an RC, 0.19.1 should be RC too)
 - If the previous bugfix release was **not** an RC, **do not** mark it as RC (for example, if you decide to mark 0.19.1 as stable, then 0.19.2 should be considered stable too)

One way we could handle this is by getting the latest stable version number (https://yarnpkg.com/latest-version) as part of the build and checking whether the minor version matches the version currently being released. If it matches, it's not an RC release. If it doesn't match, it's an RC release.

bestander:
Looks like a good idea, @Daniel15!
I'll try this for 0.19.1.

PR #2420, opened 2017-01-10
Add silent flag to allow raw output

rafaelrinaldi:
**Summary**

This PR implements a `--silent` flag (`-s` for short) to Yarn's CLI in order to allow raw script output. [This feature is available on npm](https://github.com/rafaelrinaldi/til/blob/master/npm/npm-run-silently.md) and is very useful for when you want to pipe the output of a script as an input of another script.

**Test plan**

Add a `foo` script to your `package.json` that prints a simple message:

```json
{
  "scripts": {
    "foo": "echo 'it works'"
  }
}
```

The default behavior will give you the script output followed by a header and a footer added by Yarn:

```sh
$ yarn run foo
yarn run v0.20.0-0
$ echo 'it works' 
it works
✨  Done in 0.14s.
```

However, if you want only the raw script output you can append the `--silent` flag:

```sh
$ yarn run foo --silent # or simply `-s`
it works
```

---

This is my first PR to Yarn so please let me know if I need to change anything in order to make it acceptable for review. Thanks!

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -75,6 +75,10 @@ commander.option(
   '--no-emoji',
   'disable emoji in output',
 );
+commander.option(
+  '-s, --silent',
+  'raw script output',
+);
 commander.option('--proxy <host>', '');
 commander.option('--https-proxy <host>', '');
 commander.option(
@@ -191,10 +195,11 @@ const reporter = new Reporter({
   emoji: commander.emoji && process.stdout.isTTY && process.platform === 'darwin',
   verbose: commander.verbose,
   noProgress: !commander.progress,
+  isSilent: commander.silent,
 });
+
 reporter.initPeakMemoryCounter();
 
-//
 const config = new Config(reporter);
 
 // print header
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -27,9 +27,10 @@ type Row = Array<string>;
 export default class ConsoleReporter extends BaseReporter {
   constructor(opts: Object) {
     super(opts);
-    this._lastCategorySize = 0;
 
+    this._lastCategorySize = 0;
     this.format = (chalk: any);
+    this.isSilent = !!opts.isSilent;
   }
 
   _lastCategorySize: number;
@@ -142,6 +143,7 @@ export default class ConsoleReporter extends BaseReporter {
   }
 
   _log(msg: string) {
+    if (this.isSilent) { return; }
     clearLine(this.stdout);
     this.stdout.write(`${msg}\n`);
   }
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -143,7 +143,9 @@ export default class ConsoleReporter extends BaseReporter {
   }
 
   _log(msg: string) {
-    if (this.isSilent) { return; }
+    if (this.isSilent) {
+      return;
+    }
     clearLine(this.stdout);
     this.stdout.write(`${msg}\n`);
   }
--- a/src/reporters/base-reporter.js
+++ b/src/reporters/base-reporter.js
@@ -29,6 +29,7 @@ export type ReporterOptions = {
   stdin?: Stdin,
   emoji?: boolean,
   noProgress?: boolean,
+  silent?: boolean,
 };
 
 export function stringifyLangArgs(args: Array<any>): Array<string> {
@@ -74,6 +75,7 @@ export default class BaseReporter {
   emoji: boolean;
   noProgress: boolean;
   isVerbose: boolean;
+  isSilent: boolean;
   format: Formatter;
 
   peakMemoryInterval: ?number;


src/reporters/console/console-reporter.js
```diff
@@ -142,6 +143,7 @@ export default class ConsoleReporter extends BaseReporter {
   }
 
   _log(msg: string) {
+    if (this.isSilent) { return; }
```
bestander: style nit: afaik we don't inline if blocks, could you expand it to 3 lines?

rafaelrinaldi: Sure thing. Here you go: 001aee04cec965ccd0cc302997c12db3056174d2

rafaelrinaldi:
@bestander Since the goal of this feature is to enable interoperability between scripts and CLI tools, it should only be used together with the `run` command. Shouldn't affect the output of any other commands.

bestander:
@rafaelrinaldi, the change touches the generic src/cli/index.js and base-reporter.js, would not it affect all the commands then?

rafaelrinaldi:
@bestander It would only affect the `run` command, other commands will simply ignore it. I could not find a different way to implement this feature, let me know if you have any suggestions.

bestander:
The change touches ConsoleReporter which is injected into config and then used in many places.
How come it is only used in run command and not others?

rafaelrinaldi:
@bestander As my PR says, this is a feature available on `npm` already and this is intended to be a implementation of the exact same feature. On `npm` it's a flag for the `run` command, it won't affect any other commands at all.

Let me know if either intent or implementation are not clear.

src/reporters/console/console-reporter.js
```diff
@@ -27,9 +27,10 @@ type Row = Array<string>;
 export default class ConsoleReporter extends BaseReporter {
   constructor(opts: Object) {
     super(opts);
-    this._lastCategorySize = 0;
 
+    this._lastCategorySize = 0;
     this.format = (chalk: any);
+    this.isSilent = !!opts.isSilent;
```
bestander: The intent is great, thanks for starting the PR.
The implementation confuses me because ConsoleReporter is used in other commands, not only in `run`.
In the 3 changed files I don't see this setting being used for only run command.

As a matter fact, I think silent could be as useful for other commands, too.
We just need to be aware of the effects.

rafaelrinaldi: @bestander Got it. What do you think I should do? The only file that made sense to me during implementation was `console-reporter.js`.

bestander: I think your solution makes total sense even for install and check commands.
Just add it to your test plan and check what happens if you run other commands with this flag.

rafaelrinaldi: @bestander Alright, I will give it a shot as soon as I can. Thanks for the feedback 👍

jonathandelgado:
This is a much needed feature (as demonstrated by #788)

@rafaelrinaldi let me know if you want any help on this.

rafaelrinaldi:
@jonathandelgado Thanks for volunteering. My plan is to submit my updates this weekend so @bestander can review.

mhipszki:
👍 although does it also remove the measured time of the task running? that might be a feature still useful in "silent" mode

rafaelrinaldi:
@mhipszki My idea is to reproduce exactly what npm offers. If we decide to change, we can do it separately.

mhipszki:
@rafaelrinaldi makes sense!

bestander:
So I tested this on install command, as expected the all the regular logs of install were hidden but progress bar and warnings were shown.

Actually I need this exact feature for a tool where Yarn's verbose install log is unnecessary, so I think we should get this feature in in some way..

However the flag should not be `--silent`, this feature mutes one of the reporters, console reporter.

@rafaelrinaldi, how about `--no-console-logs`?
We have another flag `--no-progress` that hides progress bars logs, I think it is sort of related.
Open to other suggestions.

rafaelrinaldi:
@bestander Okay, I'm good with switching the names. Will update my PR.

bestander:
@rafaelrinaldi ping, it would be great to get this in, do you have a chance to update the PR?

rafaelrinaldi:
@bestander Sorry, been working like crazy lately and didn't had a chance to update. Wait until Saturday if possible, I will have some free time and can work on this 👍

bestander:
Thanks :)

src/reporters/console/console-reporter.js
bitjson: @rafaelrinaldi – I'd love to see this merged, can I help in any way?

rafaelrinaldi:
@bestander I did most changes yesterday. I will submit a PR as soon as I get to my computer (probably tonight).

adamyonk:
Thanks for tackling this, @rafaelrinaldi! I run into this quite often!

bestander:
ping

rafaelrinaldi:
@bestander Sorry for taking a long time to tackle this, been very busy. I'm gonna have some free time at the end of the week and will be able to update this properly.

src/cli/index.js
```diff
@@ -75,6 +75,10 @@ commander.option(
   '--no-emoji',
   'disable emoji in output',
 );
+commander.option(
+  '-s, --silent',
+  'raw script output',
```
bestander: nit: `silence Yarn console logs` would be closer to correct.
Technically it is not doing anything to raw output.

bestander:
Thanks for the update, @rafaelrinaldi.
Just one nit that prevents me from merging, I don't want this option to be misleading.

src/cli/index.js
bestander: Nevermind, I just changed the message quickly.
Thanks for rebasing!

le93us:
And is it possible to make every `yarn $SCRIPT_NAME` inokation silent without extra flag passing?

rafaelrinaldi:
@le93us It is definitely possible but I don't believe this is the expected behavior.

PR #2440, opened 2017-01-12
Set boolean config values correctly

### LINKED ISSUE #2434: yarn config set does not allow to set boolean false
Bug

`yarn config set strict-ssl false` sets it to string 'false' instead of boolean false. In consequence, 'false' seems to be treated as truthy and has no effect.

see http://stackoverflow.com/questions/41439008/how-do-you-set-boolean-values-with-yarn-config

**Expected behaviour**
The config value should have been set to boolean false.

iredchuk:
* Fixes #2434

`yarn config set [option] [false|true]` 
should set the option value to false/true respectively. 
Currently it's set to string value "false"/"true".

*Test Plan*
* `yarn config set strict-ssl false`
* $HOME/.yarnrc should contain the line:
   strict-ssl false

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -19,8 +19,8 @@ export const {run, setFlags} = buildSubCommands('config', {
     if (args.length === 0 || args.length > 2) {
       return false;
     }
-
-    const [key, val = true] = args;
+    const key = args[0];
+    const val = args[1] || true;
     const yarnConfig = config.registries.yarn;
     await yarnConfig.saveHomeConfig({[key]: val});
     reporter.success(reporter.lang('configSet', key, val));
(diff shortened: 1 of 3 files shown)


src/cli/commands/config.js
```diff
@@ -19,8 +20,8 @@ export const {run, setFlags} = buildSubCommands('config', {
     if (args.length === 0 || args.length > 2) {
       return false;
     }
-
-    const [key, val = true] = args;
+    const key = args[0];
+    const val = BaseRegistry.normalizeConfigOption(args[1] || true);
```
torifat: Isn't it better to use `yarnConfig` without importing `BaseRegistry`?

iredchuk: Well, can't use it directly - it's static in BaseRegistry, or do mean something else?

iredchuk: But maybe it's better to ensure config is normalized in yarn-registry's `saveHomeConfig` from the dependency prospective?

OACDesigns:
this PR has introduced a regression (https://github.com/yarnpkg/yarn/issues/2856).
Setting a value to an empty string now results in defaulting to a boolean `true`

Issue #1941, opened 2016-11-18
Feature: Allow soft-linking of yarn shell script


jambonrose:
I am running macOS 10.11.6, Node 6.9, and yarn 1.17.3.

[I've just created a Portfile for yarn to be installed in Macports.](https://github.com/macports/macports-ports/pull/48)

In doing so, I changed the code in `bin/yarn` from the line of code in 1 into the code in item 2.

1. `basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")`
2. `basedir=$(dirname "$(readlink "$0" || echo "$(echo "$0" | sed -e 's,\\,/,g')")")`

The addition of the `readlink` command allows for the `bin/yarn` script to be soft linked to `/opt/local/bin/yarn` for installation, allowing for yarn to be easily uninstalled by the user if desired. 

I believe the code on line 2 allows the currently script to continue working in its original context, and that the new context should work in other shells on other systems, but I would like some feedback before I open a PR with this change.

Are people interested in this change?

Kineolyan:
Agreeing with @jambonrose. This also happens if you install yarn in /opt - as suggested by [doc for manual installation](https://yarnpkg.com/en/docs/install#alternatives-tab), but instead of add the /opt folder to `$PATH`, you add a symlink in one of the existing bin folders - /usr/bin, /usr/local/bin, ...

Cheers

PR #2441, opened 2017-01-13
support relative symlinks in yarn script

philtay:
**Summary**

https://github.com/yarnpkg/yarn/issues/1941 is great, but it doesn't support relative symlinks.
Given the following symlink `yarn -> ../share/yarn/bin/yarn` you get `Error: Cannot find module '/share/yarn/bin/yarn.js'`.

**Test plan**

Just link the `yarn` script with `ln -rs`. Then run it. With this patch it's going to work. I think this is 0.19 material. Basically a bugfix.

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -1,5 +1,5 @@
 #!/bin/sh
-basedir=$(dirname "$(readlink "$0" || echo "$(echo "$0" | sed -e 's,\\,/,g')")")
+basedir=$(dirname "$(readlink -f "$0" || echo "$(echo "$0" | sed -e 's,\\,/,g')")")
 
 use_winpty=0
 


philtay:
@bestander you can remove [this hack](https://github.com/yarnpkg/yarn/blob/master/scripts/build-deb.sh#L56) from the deb builder script. you don't need it anymore, now it works with symlinks

bestander:
Thanks.
Would you be willing to help me and send a PR? :)

bestander:
I get this message on my latest Mac OS X
```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

bestander:
@philtay, any idea for a quick patch or should I revert the 2 commits?

PR #2443, opened 2017-01-13
don't cache dependencies installed with file: protocol - fixes #2165

### LINKED ISSUE #2165: yarn probably shouldn't cache packages resolved with a file path

I guess a _bug_.

Say you have the following project structure:

```
component-foo/
└── package.json
└── index.js

yarn-test/
└── package.json
```

With the following files:

`component-foo/package.json`:
```
{
  "name": "component-foo",
  "version": "1.0.0",
  "private": true,
  "main": "index.js"
}
```

`component-foo/index.js`:
```
console.log('foo');
```

`yarn-test/package.json`:
```
{
  "name": "yarn-test",
  "version": "1.0.0",
  "private": true,
  "dependencies": {
    "component-foo": "file:../component-foo"
  }
}
```

Now you run `$ yarn install` inside `yarn-test/` and `yarn-test/node_modules/component-foo/index.js` is:
```
console.log('foo');
```

Now remove `yarn-test/node_modules/` and `yarn-test/yarn.lock` and change `component-foo/index.js` to this:
```
console.log('bar');
```

Now you run `$ yarn install` inside `yarn-test/` again and `yarn-test/node_modules/component-foo/index.js` will be:
```
console.log('foo');
```

It uses the cached version of `component-foo`, but `component-foo/index.js` has changed.

I'd expect that `yarn-test/node_modules/component-foo/index.js` should be this at the end:
```
console.log('bar');
```

Maybe packages installed with local paths like `file:../` shouldn't be cached at all, if we can't figure out, if they have changed.

(FYI: It looks like npm doesn't cache them.)

```
$ node -v
v6.9.1

$ yarn -V
0.18.0

macOS 10.12.1
```

donaldpipowitch:
**Summary**

I try to fix #2165. For now I just added a failing unit test. I've spoken with @bestander about this.

**Test plan**

An example is described here: #2165. And this PR contains a failing unit test. I try to install a dependency from a local file with content `foo\n`. Than I change the content to `bar\n` and install it again. It should now be `bar\n`, but it is still `foo\n`

---

@bestander I'm **not** sure this unit test works as expected. It looks like all tests assert only in the cache directory (?) given with `config.cwd`. In my fixtures folder no dependency is added which I could check directly. Could you review the test?
Is there any way to get a nicer test output when I only run one test? I do `$ npm run -s test-only -- -t 'without cache'`, but the output is quite verbose (e.g. contains a lot of `Unhandled promise rejection` errors which seems to result from skipping the other tests).


__tests__/commands/install/integration.js
```diff
@@ -212,6 +212,26 @@ test.concurrent('install file: protocol with relative paths', (): Promise<void>
   });
 });
 
+test.concurrent('install file: protocol without cache', async (): Promise<void> => {
+  const fixturesLoc = path.join(__dirname, '..', '..', 'fixtures', 'install');
+  const compLoc = path.join(fixturesLoc, 'install-file-without-cache', 'comp', 'index.js');
+
+  return fs.writeFile(compLoc,'foo\n').then(() =>
```
bestander: No reason to mix promises and async/await.
await fs.writeFile would work here, too

__tests__/commands/install/integration.js
```diff
@@ -212,6 +212,26 @@ test.concurrent('install file: protocol with relative paths', (): Promise<void>
...
+    runInstall({noLockfile: true}, 'install-file-without-cache', async (config) => {
+      assert.equal(
+        await fs.readFile(path.join(config.cwd, 'node_modules', 'comp', 'index.js')),
+        'foo\n',
+      );
+      await fs.writeFile(compLoc, 'bar\n');
+      await runInstall({noLockfile: true}, 'install-file-without-cache');
+      assert.equal(
+        await fs.readFile(path.join(config.cwd, 'node_modules', 'comp', 'index.js')),
```
bestander: I think here you still read from the first install.
config.cwd comes as argument to runInstall.
Instead make this check inside the runInstall callback and double check that we are not testing the same location.

__tests__/commands/install/integration.js
```diff
@@ -212,6 +212,26 @@ test.concurrent('install file: protocol with relative paths', (): Promise<void>
...
+        'bar\n',
```
bestander: We can't merge a failing test, once you make sure it works correctly, can you switch the last assert to the opposite so that this test described the current behavior and add a link to the issue? 
So that it would be easy to switch the assert when working on the fix.

__tests__/commands/install/integration.js
donaldpipowitch: Even better! I just followed the style of most tests returning the first promise (mostly `return runInstall`).

donaldpipowitch:
@bestander I changed the unit test... an now I'm surprised. It passes 😮  Was this fixed already?

bestander:
Ha ha, probably not.
The test passes because 2 runInstall calls are executed sandboxed.
The trick is to call runInstall once and for the second time call Install function which does not create a new sandbox.
See https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L306

donaldpipowitch:
@bestander I updated the unit test so it uses `Install`. I used `notEqual` in the last assertion and linked to this pull request to have a "working" unit test. Thank you very much for your help so far.

bestander:
Thanks, I'll have a look in the next couple of days

bestander:
Thanks, that will help to keep track of it.
Let me know if you want to collaborate on the fix next.
Feel free to cc me in a PR with some ideas.

donaldpipowitch:
Thank you a lot. Yeah, I'll try to fix the bug with another PR.

simonbuchan:
:( the PR title is a bit misleading...

donaldpipowitch:
Yes. I changed that. I'm sorry. In the beginning thought this would be an ongoing PR, but that idea was changed to two separate PRs: the first PR just adds a unit test, the second PR will fix it.

moimikey:
test for this was merged, but **what's the status on the actual fix**? its been mentioned in previous issues about the pain points of caching `file:` paths and the cache not updating to reflect new changes. the response has been to help contribute, but it's a time waster for another person to contribute if this is almost actually done.

donaldpipowitch:
As said here (https://github.com/yarnpkg/yarn/issues/2165#issuecomment-282716368) I couldn't start another PR and AFAIK no one else did.

PR #2444, opened 2017-01-13
Fix "file://" URI to accept absolute paths

Kronuz:
**Summary**

The check for absolute path should be done against the path without the `file://` prefix and in order to unambiguously check for absolute paths, the `file://` URI scheme should always start with double slash (as required by the specification), a third slash would make the path absolute.

**Example**
The following is a excerpt of a `package.json`
```json
{
  "dependencies": {
    "my-package": "file:///home/kronuz/packages/my_package"
  }
}
```


bestander:
This PR has been open for a while, but can't be merged this way.
Feel free to fix and open a new one

PR #2454, opened 2017-01-15
[FIXES #2096] fix relative symlinks of symlinks

### LINKED ISSUE #2096: Linking fails in .config is symlinked
Bug

`yarn link foo` fails if `~/.config` is a symlink.

```sh
$ ln -s ~/some/other/path/.config ~/.config
$ brew install yarn

$ cd foo && yarn link
yarn link v0.17.9
success Registered "foo".
info You can now run `yarn link "foo"` in the projects where you want to use this module and it will be used instead.
✨  Done in 0.07s.

$ cd ../bar && yarn link foo
yarn link v0.17.9
error No registered module found called "foo".
info Visit https://yarnpkg.com/en/docs/cli/link for documentation about this command.
```

The final command above should succeed in linking `foo` into `bar`. Or, if symlinking `.config` is not supported, it should warn on install or when the command runs.

`yarn 0.17.9`
`node 6.6.0`
`OS X 10.11.6`

----

The problem seems to stem from the fact that `utils/fs.js` [favors relative paths for symlinks when possible](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/util/fs.js). This means that the symlink created in `~/.config/yarn/link/foo` is a relative path, i.e. `../../../foo`.

When the OS resolves this relative path though, it does so based on the real location of the symlink file itself. So rather than resolving relative to `~/.config/yarn/link`, it resolves it relative to `~/some/other/path/.config` (which is where the `~/.config` symlink points to). This means the relative path is invalid. And Node's `fs.exists()` (used to [detect if the link exists when running `yarn link <pkg>`](https://github.com/yarnpkg/yarn/blob/099e76dc075763427605a6a63906127402370bb2/src/cli/commands/link.js#L34)) returns false if the file is a symlink whose destination does not exist.

Not sure if there are other benefits to keeping the symlink relative (the code I linked to above mentions something about maintaining links when things move). If the answer is simply that symlinked `.config` dirs are not supported, it would be nice if the install or link command at least warned that this will break yarn.

stefanpenner:
**Summary**

This makes `yarn link` work when `$HOME/.config/yarn/link/` is itself a symlink.

---

The following commit forces symlinks to be relative: https://github.com/yarnpkg/yarn/commit/36d73cc3bb0e3c4afcea4f25b06cc76b7b5531fa

This isn't ideal, although it enables some portability it breaks others. For example, it is not uncommon for `$HOME/.config` itself to be be a symlink: `$HOME/.config` -> `$HOME/src/stefanpenner/dotfiles`. 

Now when running `yarn link` inside `$HOME/src/ember-cli/ember-cli` we end up with:
```js
path.relative('/Users/spenner/.config/yarn/link/ember-cli', '/Users/spenner/src/ember-cli/ember-cli');
=> '../../../src/ember-cli/ember-cli'
```

Which results in the link located at: `/Users/spenner/.config/yarn/link/ember-cli` pointing to
`/Users/spenner/.config/src/ember-cli/ember-cli` rather then `/Users/spenner/src/ember-cli/ember-cli`

This is because `/Users/spenner/.config` is actually a symlink pointing to `/Users/spenner/src/stefanpenner/dotfiles/.config/`

--- 

This PR mitigates he issue raised in #2096, but reifying the paths via `realpathSync` before deriving the relative path. This ensures that at the time of linking, the paths are correct. This doesn't fix all issues, e.g. symlinks cannot always be changed after this reification.

**Test plan**

`src/lib/fs.js` appears currently untested... if this is an approach you all would like to take I can add some tests.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -419,7 +419,7 @@ export async function symlink(src: string, dest: string): Promise<void> {
       await fsSymlink(src, dest, 'junction');
     } else {
       // use relative paths otherwise which will be retained if the directory is moved
-      const relative = path.relative(path.dirname(dest), src);
+      const relative = path.relative(fs.realpathSync(path.dirname(dest)), fs.realpathSync(src));
       await fsSymlink(relative, dest);
     }
   } catch (err) {


arcanis:
I've just stumbled upon this PR, but I don't understand it.

> ```js
> path.relative('/Users/spenner/.config/yarn/link/ember-cli', '/Users/spenner/src/ember-cli/ember-cli');
> => '../../../src/ember-cli/ember-cli'
> ```
>
> Which results in the link located at: `/Users/spenner/.config/yarn/link/ember-cli` pointing to
`/Users/spenner/.config/src/ember-cli/ember-cli` rather then `/Users/spenner/src/ember-cli/ember-cli`

I don't understand this last sentence. If `/Users/spenner/.config/yarn/link/ember-cli` is a symlink to `../../../src/ember-cli/ember-cli`, then `realpath("/Users/spenner/.config/yarn/link/ember-cli")` will yield `/Users/spenner/src/ember-cli/ember-cli` as it should be, not `/Users/spenner/.config/src/ember-cli/ember-cli` as you report. The realpath doesn't change anything here.

@stefanpenner Do you recall more details? This `realpath` doesn't make sense to me and cause issues when one of the path component is invalid for whatever reason.

PR #2455, opened 2017-01-15
remove hack in deb builder script

philtay:
ref https://github.com/yarnpkg/yarn/pull/2441

ping @bestander


Daniel15:
I guess we don't need this any more, since the `yarn` script handles symlinks now?

PR #2471, opened 2017-01-16
Install a package with latest tag, it looks for the latest dist-tag

### LINKED ISSUE #1306: Install with 'latest' tag, installs as version string 'latest'

### LINKED ISSUE #2225: Provide a way to fallback to cached version when installing latest offline
Not sure if it's worth an RFC.
Yarn integration in Create React App is a bit disappointing, and there is a feature that could really help us out.

Create React App installs a package called `react-scripts` every time you run it. The idea is that this is the package with the “guts” of Create React App, and thus anytime you create a new project, you get the latest version of it. So people who don’t update the CLI still get new features.

However it doesn’t work offline. So if you’re in an airplane, you can’t create a React project. This is frustrating.

Ideally we’d like a way to tell Yarn that it should try installing latest version of a package **but** if that’s impossible, we’re fine falling back to the latest version that exists in the cache.

Note that I tried using `--offline` for this and it didn't work because it requested `latest` whereas we only have a specific version in cache (for example `0.8.4`). I wish there was a way to "agree" to use whatever local version we have if there is no network connection.

I think this was mentioned as last paragraph in https://github.com/yarnpkg/yarn/pull/1277#issue-184089258 so it’s not exactly a new idea.

voxsim:
**Summary**

Hi,

this should fix #1306 and I sincerely hope that in some way it resolves #2225.
Please let me know if you like it and I don't forget anything.

**Test plan**

I add a test case in `__tests__/commands/add.js`

cheers,
Simon

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -625,3 +625,12 @@ test.concurrent('add asks for correct package version if user passes an incorrec
     },
   );
 });
+
+test.concurrent('install with latest tag', (): Promise<void> => {
+  return runAdd(['left-pad@latest'], {}, 'latest-version-in-package', async (config) => {
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+    const version = await getPackageVersion(config, 'left-pad');
+
+    assert.deepEqual(pkg.dependencies, {'left-pad': `^${version}`});
+  });
+});
--- a/__tests__/fixtures/add/latest-version-in-package/package.json
+++ b/__tests__/fixtures/add/latest-version-in-package/package.json
@@ -1,0 +1,3 @@
+{
+  "dependencies": {}
+}
--- a/src/cli/commands/add.js
+++ b/src/cli/commands/add.js
@@ -65,7 +65,7 @@ export class Add extends Install {
       version = pattern;
     } else if (parts.hasVersion && parts.range) {
       // if the user specified a range then use it verbatim
-      version = parts.range;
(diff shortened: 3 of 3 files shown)


bestander:
Awesome!

voxsim:
:D

voxsim:
thanks @bestander !

voxsim:
Do you think this pull request resolves #2225?

voxsim:
Ah ok, yes, you closed the issue ;)

bestander:
@voxsim, #2225 was closed automatically.
Would you mind double checking if the scenario Dan described is fixed with this patch?

voxsim:
Yes, I will check ;)

voxsim:
@bestander yes it works!

@gaearon i will answer in the creare-react-app issue

billyjanitsch:
FYI (see #2582) this PR introduces an inconsistency between how yarn handles the "latest" tag vs. any other tag. For example, `yarn add webpack@latest` adds `"webpack": "^2.2.0"` to package.json's dependencies, while `yarn add webpack@beta` (incorrectly) adds `"webpack": "beta"`.

Was there a reason to special-case the "latest" tag?

Issue #2505, opened 2017-01-19
Yarn doesn't work with visual studio online package manager


lyweiwei:
Our team is using VSO [package manager](https://marketplace.visualstudio.com/items?itemName=ms.feed&targetId=486cbe5a-ee18-456f-bbb6-d786bd2fdfd0) to manage our internal packages. It works fine with NPM, but fails with Yarn.

You can use [this repository](https://github.com/lyweiwei/yarn-with-vso) to reproduce the bug
```sh
git clone https://github.com/lyweiwei/yarn-with-vso
cd yarn-with-vso

# this would work
npm install --save @shazhouke/dummy-package

# this would fail
yarn add @shazhouke/dummy-package
```

Issue #2151 caused part of the problem. But even I hacked the auth for the resolving stage, it would still fail at downloading the tarball. The problem is, with VSO package manager, 1 account may have multiple feeds, and each of them has 1 registry. But all the registries under the account share the same package store.

With the example I gave, the registry is
https://shazhouke.pkgs.visualstudio.com/_packaging/shazhouke-private-npm/npm/registry
But the tarball is located at
https://shazhouke.pkgs.visualstudio.com/_packaging/42aaa647-fd03-4edd-baa1-d2b3f6ac2b16/npm/registry/@shazhouke/dummy-package/-/dummy-package-1.0.0.tgz

Note, the tarball URL doesn't begin with the registry URL. This [condition](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L61) would always be false.
```js
if (this.token || (alwaysAuth && requestUrl.startsWith(registry))) {
  const authorization = this.getAuth(pathname);
  if (authorization) {
    headers.authorization = authorization;
  }
}
```

Seems we should assume the URL for resource of a given registry always begin with the registry URL.

I'm working on a patch to unblock our team and will create a PR when it's ready.

swese44:
any progress here? looks like it's been about a month since anything happened with https://github.com/yarnpkg/yarn/pull/2507

thanks!

PR #2507, opened 2017-01-19
Support VSO package manager registry

lyweiwei:
This is for a pre-review for the fix of #2505,
Will add test cases if the maintainer team agrees the basic idea.

### Root cause
As described in #2505, for VSO package manager registries the package tarball's URL is not prefixed with the registry URL.

### Fix
When downloading tarball, calculate the auth method with the original package nam instead of the tarball URL, if the URL is resolved from an NPM registry).


bestander:
Tests are broken.
Also very much appreciated would be new tests added

lyweiwei:
@bestander I just pushed a fix for the failed test cases. But it seems some of the cases timeout due to network issues. Is this a know issue?

> Summary of all failing tests
> FAIL  __tests__/commands/install/integration.js (205.516s)
>  ● install incompatible optional dependency should still install shared child dependencies
>
>    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
>      
>      at ontimeout (timers.js:365:14)
>      at tryOnTimeout (timers.js:237:5)
>      at Timer.listOnTimeout (timers.js:207:5)
>
> FAIL  __tests__/package-resolver.js (88.384s)
>  ● resolve react-native
>
>    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
>      
>      at ontimeout (timers.js:365:14)
>      at tryOnTimeout (timers.js:237:5)
>      at Timer.listOnTimeout (timers.js:207:5)

These test cases passed on Travis.CI

lyweiwei:
BTW, can we do a retry on the failed environments?

bestander:
@lyweiwei, yeah I think 3 retries could be reasonable.
Most of the network tests are actually mocked with checked in caches, so CI is not that unstable.
Probably your commit happened during some bad time.

bestander:
Yeah, I think the logic is reasonable.
Please go on with some tests and let's merge it once we are sure we don't break any existing installs

zumwald:
bump @bestander @lyweiwei what's the path forward for vsts support?

bestander:
@zumwald, this PR is quite good, it just needs a test and a rebase.
If you want to take it over, send a new PR with required changes and we'll merge it in.

lyweiwei:
I just pushed a new iteration
* merge the latest changes from master
* wait longer for the package-resolver test cases (6s->10s)

lyweiwei:
@bestander The PR passed the test this time :) sorry for the delay.

src/registries/is-request-to-registry.js
```diff
@@ -2,18 +2,25 @@
 
 import url from 'url';
 
+const SUFFIX_VISUALSTUDIO = '.pkgs.visualstudio.com';
```
bestander: I think I missed that detail earlier.
It looks hard to scale a maintain a solution where every registry has to be hardcoded in Yarn.
Can this setting be moved to .yarnrc for each registry to configure?

__tests__/package-resolver.js
```diff
@@ -10,7 +10,7 @@ import * as fs from '../src/util/fs.js';
 import * as constants from '../src/constants.js';
 import inquirer from 'inquirer';
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 100000;
```
bestander: Does it mean that requests get slower after this change?

bestander: Also can you add a test here for the supported case?

src/registries/npm-registry.js
```diff
@@ -50,16 +52,14 @@ export default class NpmRegistry extends Registry {
...
 
-  request(pathname: string, opts?: RegistryRequestOptions = {}): Promise<*> {
-    const registry = addSuffix(this.getRegistry(pathname), '/');
+  request(pathname: string, opts?: RegistryRequestOptions = {}, packageName: ?string): Promise<*> {
+    const registry = this.getRegistry(packageName || pathname);
     const requestUrl = url.resolve(registry, pathname);
-    const alwaysAuth = this.getScopedOption(registry.replace(/^https?:/, ''), 'always-auth')
-      || this.getOption('always-auth')
-      || removePrefix(requestUrl, registry)[0] === '@';
+    const alwaysAuth = this.getRegistryOrGlobalOption(registry, 'always-auth');
```
restrry: @bestander does it make sense to force adding auth key to resoled url?
I spent a day some months ago when tried to fix [similar problem](https://github.com/yarnpkg/yarn/issues/521#issuecomment-277500043) with our artifactory.

kenotron:
@lyweiwei Can you remove the VSTS specifics here and make Yarn work against VSO branch? Or did you find a different work around?

lumaxis:
@lyweiwei @bestander I tried picking continuing this in #3231. Would very much appreciate a review! 🙂

bestander:
This one got replaced with #3231

PR #2510, opened 2017-01-20
Fix `yarn version` which yields same output as `yarn --version` (#2491)

### LINKED ISSUE #2491: `yarn version` not working as stated in docs?
Bug.

`yarn version` returns the current version of yarn similar to the command `yarn --version`.

Run the command `yarn version` in your CLI.

Unless I'm misinformed, should it not be performing the behavior described in the docs? https://yarnpkg.com/en/docs/cli/version

Running `yarn version` should output something like the following:
```
info Current version: 1.0.1
question New version: 1.0.2
info New version: 1.0.2
✨  Done in 9.42s.
```

node: 7.4.0
yarn: 0.19.1
os: macOS 10.11.6

ConAntonakos:
There was a conflict when commander attempts to parse the incoming args between the command executed and the options since the name `version` was shared. In other words, executing the command `yarn version` would yield the same output as `yarn --version`. This fixes #2491. This also seems to relate to PR #2268.

Let me know what you think. And sorry for the issue reference spam! Thanks!

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -170,8 +170,13 @@ if (commandName === 'help' || args.indexOf('--help') >= 0 || args.indexOf('-h') 
   process.exit(1);
 }
 
+if (!command) {
+  args.unshift(commandName);
+  command = commands.run;
+}
+invariant(command, 'missing command');
+
 // parse flags
-args.unshift(commandName);
 commander.parse(startArgs.concat(args));
 commander.args = commander.args.concat(endArgs);
 
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -38,6 +38,12 @@ for (let i = 0; i < args.length; i++) {
   }
 }
 
+// NOTE: Pending resolution of https://github.com/tj/commander.js/issues/346
+// Remove this (and subsequent use in the logic below) after bug is resolved and issue is closed
+const ARGS_THAT_SHARE_NAMES_WITH_OPTIONS = [
+  'version',
+];
+
 // set global options
 commander.version(pkg.version);
 commander.usage('[command] [flags]');
@@ -170,13 +176,13 @@ if (commandName === 'help' || args.indexOf('--help') >= 0 || args.indexOf('-h') 
   process.exit(1);
 }
 
-if (!command) {
-  args.unshift(commandName);
-  command = commands.run;
-}
-invariant(command, 'missing command');
-
 // parse flags
+args.unshift(commandName);
+
+if (ARGS_THAT_SHARE_NAMES_WITH_OPTIONS.indexOf(commandName) >= 0 && args[0] === commandName) {
+  args.shift();
+}
+
 commander.parse(startArgs.concat(args));
 commander.args = commander.args.concat(endArgs);
 


bestander:
tests got broken

ConAntonakos:
Yeah; odd because running `yarn run test-only` did not yield these errors...

Is that the best way to ensure testing is smooth? Or am I missing something else?

bestander:
@ConAntonakos, the CI could fail because of network issues or was based on an unstable commit.
Can you please rebase on latest master?

ConAntonakos:
Sure! Thanks for letting me know.

bestander:
I think CI shows a genuine break in tests now

markstos:
Since `yarn version` vs `yarn --version` made it into a release, it seems there must have been a lack of test coverage that led to this bug being introduced. So it would be great if the fix included some additional test coverage that confirms that `yarn version` works expected (or at least that doesn't just print the version!).

ConAntonakos:
@markstos That's a good idea. I essentially reverted a previous change that caused the break. I'll keep at it and consider test coverage as well.

ConAntonakos:
Ok, I was able to get unit tests working again. It seems they weren't playing nicely with node v7, which breaks them. I'll get a screen capture of the error output next time.

I'm delving more into the source - and subsequently, the `commander` package with which I'm not familiar - and I noticed that similarly named commands and parsed flags are executing the same logic, i.e., `help` and `--help` (`version` and `--version`). I can't tell yet if that's a `commander` issue or a `yarn` source issue.

The major culprit is this line: https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L174

If the above line is removed, `yarn version` and `yarn --version` work perfectly, but it breaks the `https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/add.js` tests.

bestander:
To rebase properly I usually do
```
git rebase master
git push <my fork> <my branch> --force
```

Merging a master branch does not work well with github PRs

ConAntonakos:
Ugh! I apologize for that. Do you want me to redo that?

bestander:
No worries, yeah, better rebase, otherwise I don't know what happens when the PR is merged :)

ConAntonakos:
Looks like it's failing the following test type `TEST_TYPE="test-ci -- -- --maxWorkers 1"` and its timing out these subsequent tests:
* _--production flag ignores dev dependencies (90009ms)_
* _install transitive optional dependency from lockfile (90009ms)_

bestander:
@ConAntonakos, don't worry about the MacOS build on travis, it is a bit flaky.
Would you add a test in this PR?

bestander:
I'll merge the fix now.
@ConAntonakos a follow up PR with a test would be really great.

ConAntonakos:
@bestander Indeed! I want to digest the structure of the testing (and subsequently, the flow of logic) and figure out why they weren't catching this error.

ConAntonakos:
This is very odd. Instantiating `rprter` returns an `undefined` result: https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/version.js#L32

PR #2516, opened 2017-01-20
Remove `lockfileOutdated` description `$` to allow easier copy/paste

grant:
**Summary**
Remove `$` to allow easier copy/paste

**Test plan**
- Run `yarn outdated` with an outdated lockfile.


xaka:
@grant was about to send similar PR and found yours, good catch! i also find it too noisy to prepend commands with `$ `; there is another line (https://github.com/yarnpkg/yarn/blob/master/src/reporters/lang/en.js#L237) with same "issue", i believe it makes sense to clean everything up in single PR?

grant:
@xaka Thanks. Makes sense. LMK if there's anything else.

PR #2384, opened 2017-01-04
Reroll #1816 into a new PR + fix for test failures

chrisirhc:
**Summary**

Intended to solve #573 . This is a re-application of the changes in #1816 but with a fix for the tests in 3c55f6ba63aa87a87f7429f5e8e081004fc8664b. All work is credited to @f-sign . Only attempting to re-run tests and clean up the PR so it doesn't have duplicate commits from master.

**Test plan**

Tests pass. Left it up to author of #1816 .

**Background**

This originated from investigating https://github.com/yarnpkg/yarn/pull/1816#issuecomment-270288503 .

PR #2519, opened 2017-01-20
Fix private urls using colon separator

Apercu:
People were still experiencing issues with private urls using colons as separator even after the fix introduced in #2384.

This fix basically replace the initial pathname part of the parsed url in the git resolver, and reformat exploded parts onto the equivalent url, removing the colon culprit.

Unsure about why the require onto the `url` module is done the old way, I've decided to follow consistency, but can refactor if needed. Related issues referenced in commit.

:wave: @chrisirhc @lxe

### DIFF
--- a/__tests__/resolvers/exotics/git-resolver.js
+++ b/__tests__/resolvers/exotics/git-resolver.js
@@ -1,0 +1,45 @@
+/* @flow */
+
+import {parse} from 'url';
+
+import GitResolver from '../../../src/resolvers/exotics/git-resolver.js';
+
+test('GitResolver transformUrl method is defined', () => {
+  expect(GitResolver.transformUrl).toBeDefined();
+});
+
+test('GitResolver transformUrl does not affect normal urls', () => {
+
+  const urls = [
+    'git+https://github.com/npm-ml/ocaml.git#npm-4.02.3',
+    'git+ssh://git@gitlab.com/user/repo.git',
+    'git+ssh://git@github.com/user/repo.git',
+    'git+ssh://username@private.url/sub/right-pad',
+    'https://github.com/npm-ml/re',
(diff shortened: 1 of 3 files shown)


chrisirhc:
The might not work for some git urls since `:` doesn't always translate to `/`. The conversion needs to be through git/ssh client which understand in their own way what the `:` means based on the `.gitconfig` or `.ssh/config`. I've mentioned one way of solving this is to not attempt to rewrite the URL (since it'll just fail) and just send it straight to the git client for handling.

__tests__/resolvers/exotics/git-resolver.js
```diff
@@ -0,0 +1,45 @@
...
+    'package@git@bitbucket.org:team/repo.git',
+  ];
+
+  urls.forEach((url) => {
+    expect(GitResolver.transformUrl(parse(url))).toBe(url);
+  });
+
+});
+
+test('GitResolver transformUrl affect host colon separated urls', () => {
```
btford: Maybe call this `'GitResolver transformUrl replaces the colon on colon separated urls'` ?

src/resolvers/exotics/git-resolver.js
```diff
@@ -60,6 +61,14 @@ export default class GitResolver extends ExoticResolver {
     return false;
   }
 
+  // This transformUrl util is here to replace colon separators in the pathname
+  // from private urls. It takes the url parts retrieved using urlFormat and
+  // returns the associated url. Related to #573, introduced in #2519.
+  static transformUrl(parts) : string {
```
btford: I think we can probably come up with a more descriptive name than "transform."

bestander:
Nice work, thanks a lot for the tests!

PR #2521, opened 2017-01-21
make add command working with latest tag if offline or preferOffline is setted

voxsim:
**Summary**
The package constraint resolver had some issue with latest tag and if you try to specify --offline or --prefer-offline it throws exception because it does not retrieve the latest cached version :)

**Test plan**
I added two test cases and expanded the test for latest tag.

Let me know if it's ok!
cheers,
Simon


bestander:
Thanks for tests!

voxsim:
@bestander you are welcome.

I tried to clarify some tests better because runAdd has some magic, if i open a pull request could be interesting for you to discuss?

PR #2553, opened 2017-01-25
interpret a tilde at the start of config prefix path as the user's home directory

### LINKED ISSUE #2008: Installing global package will create directory with name '~' in current directory
When adding global package yarn will create directory with name '~' in current directory.

```
$ mkdir ~/temp && cd ~/temp
$ yarn global add nodemon
```
this will create directory "~" inside ~/temp

Node.js: v7.1.0
yarn: v0.17.8
OS: ArchLinux
kernel: Linux 4.8.8-2

chee:
This just interprets `~` as the user's `$HOME` when reading the prefix path in `~/.npmrc` as npm does

this fixes  #2008 

beforehand, running `yarn global add lite-server` with an npmrc containing `prefix = ~/.local/bin` would create a directory called literally `~` and put it in `bin` in there. after this change it puts it in `/usr/chee/.local/bin` for me

xo

### DIFF
--- a/__tests__/commands/config.js
+++ b/__tests__/commands/config.js
@@ -5,13 +5,16 @@ import * as reporters from '../../src/reporters/index.js';
 import * as configCmd from '../../src/cli/commands/config.js';
 import {run as buildRun} from './_helpers.js';
 import * as fs from '../../src/util/fs.js';
+import home from 'user-home';
 
 const path = require('path');
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'config');
 
 const runConfig = buildRun.bind(
   null,
   reporters.ConsoleReporter,
-  '',
+  fixturesLoc,
   (args, flags, config, reporter): CLIFunctionReturn => {
     config.registries.yarn.homeConfigLoc = path.join(config.cwd, '.yarnrc');
     return configCmd.run(config, reporter, flags, args);
@@ -47,5 +50,11 @@ test('set value "false" to an option', (): Promise<void> => {
 test('set value "true" to an option', (): Promise<void> => {
   return runConfig(['set', 'strict-ssl', 'true'], {}, '', (config) => {
     expect(config.registries.yarn.homeConfig['strict-ssl']).toBe(true);
+  });
+});
+
+test('get prefix interprets tilde', (): Promise<void> => {
+  return runConfig(['get', 'prefix'], {}, 'get-prefix-interprets-tilde', (config) => {
+    expect(config.registries.npm.config.prefix).toBe(`${home}/lol`);
   });
 });
--- a/__tests__/fixtures/config/get-prefix-interprets-tilde/.npmrc
+++ b/__tests__/fixtures/config/get-prefix-interprets-tilde/.npmrc
@@ -1,0 +1,1 @@
+prefix = ~/lol
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -126,6 +126,11 @@ export default class NpmRegistry extends Registry {
 
     for (const [, loc, file] of await this.getPossibleConfigLocations('.npmrc')) {
       const config = Registry.normalizeConfig(ini.parse(file));
+
+      if (config.prefix && userHome) {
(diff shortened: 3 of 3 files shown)


bestander:
Thanks for the fix, @chee.
I think we have a different interpretation of home dir in Yarn https://github.com/yarnpkg/yarn/blob/master/src/constants.js#L7.
Can you export it into a module and use it in both places?

Also could add a test?

chee:
oh yeah. well that one isn't about interpreting a user-given path, but seeing as yarn is already using this user-home module, it would make sense to use it instead of using os.homedir.

so i'll do that.

i was a little unclear on how to test getGlobalPrefix as it didn't seen to be being tested before.
do i need to export it here and then import it from `__tests__/commands/global.js` ?

bestander:
I think an integration test in  __tests__/commands/global.js should work fine for this case

chee:
that's it now using `user-home`

on one box:
```
$ ./bin/yarn config get prefix
/Users/chee/.local
```
 
on another box:
```
./bin/yarn config get prefix
/usr/chee/.local
```

will add the test now

chee:
i'm not actually sure how to write the test for this.
drowning.
`__tests__/commands/global.js` is, so far, only for testing the `global` command. as in `yarn global add` etc. whereas this is a test for `yarn config get` really. but it doesn't really belong in `__tests__/commands/config.js` either.

and i'm not sure how to write it or what to write or where to write it

chee:
i could do this:
```
test('get prefix interprets tilde', (): Promise<void> => {
  return runConfig(['get', 'prefix'], {}, '', async (config) => {
    expect(config.registries.npm.config.prefix).toMatch(RegExp(`^${home}`))
  })
})
```

if i could figure out how to feed it an npmrc to work from

chee:
okay, well i've got it reading the .npmrc from a fixture now, but the test's failing, because at that point it hasn't actually gone through the getGlobalPrefix function. it feels a little weird for this to be a special case when first parsing the .npmrc, but that might be the place to do it.

what are your thoughts

chee:
yes, this code is in the wrong place actually because then `yarn config get prefix` will return with a literal '~' where as uses of it will be user's $HOME.

chee:
okay, @bestander. ignore all that.

now it does it when it first loads the config, so now `yarn config get prefix` returns the same thing as `npm config get prefix` would, and there's a test for it.

x

chee:
let me just fix these failing lints

chee:
the test failed on ci ;-;

__tests__/commands/config.js
```diff
@@ -5,13 +5,16 @@ import * as reporters from '../../src/reporters/index.js';
...
 
 const path = require('path');
 
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'config');
+
 const runConfig = buildRun.bind(
   null,
   reporters.ConsoleReporter,
-  '',
+  fixturesLoc,
```
bestander: I think this runConfig is the same for every test in the suite.
Could they be affecting each other?

bestander: Nevermind, I see that you pass path only in one place

__tests__/commands/config.js
```diff
@@ -49,3 +52,9 @@ test('set value "true" to an option', (): Promise<void> => {
     expect(config.registries.yarn.homeConfig['strict-ssl']).toBe(true);
   });
 });
+
+test('get prefix interprets tilde', (): Promise<void> => {
```
bestander: Tests are running in parallel, in CI 3 suites at a time.

chee: i'm not sure what you're telling me here, i mean i'm not sure what to do with it D:

chee:
i can't seem to crack this and i never seem to have enough time or club mate to really get in and figure out what the issue is.

__tests__/commands/config.js
bestander: Lol, I think I was multitasking at reviewing PRs.
Nevermind :)

bestander:
Ok, let's close this one.
I think we really need interpreting ~ for configs.

chee:
the test passes if you run it on its own but fails as part of the suite.

i don't know where to begin with that.

good luck

Issue #2256, opened 2016-12-14
Error "Cannot read property 'split' of undefined" for interconnected file based dependencies:


mstijak:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
yarn fails on second run with message "Cannot read property 'split' of undefined".

**If the current behavior is a bug, please provide the steps to reproduce.**
I have extracted a simple repo.
https://github.com/mstijak/yarn-file-dep-repro

**What is the expected behavior?**
Install should work every time.

**Please mention your node.js, yarn and operating system version.**
Node 6.5. Yarn 0.17.10, Win10 x64

mstijak:
It seems that this is related to using `*` for dependency package version.

jjudd:
We're running into the same issue. I'm not sure it is related to using *, as we see the same behavior when  providing specific versions in the package.json dependencies section of our file dependencies. The same behavior can be observed by changing your example repo to have a depend on b@1.0.0 rather than b@*.

I'm not very familiar with yarn's lock file format, but it seems like it produces to entries for b, one that has a link to the file and one that does not. It seems to be blowing up on the one without a link:
```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


"a@file:./packages/a":
  version "1.0.2"
  dependencies:
    b "1.0.0"

b@*:
  version "1.0.0"

"b@file:./packages/b":
  version "1.0.0"
```

I wonder if this is related to the problem people were having in #1085 with `file:` dependencies. I'll take a closer look and see if I can figure out what is going on.

jjudd:
When I removed this lockfile entry, `yarn install` succeeds:

```
b@*:
  version "1.0.0"
```

The entry doesn't come back after running `yarn install`. I'm going to see if I can figure out why that lockfile entry is being generated and stop it from happening.

When I remove both the above lockfile entry and then either the entry for `b` or `a`, the `b@*` entry comes back and subsequent `yarn install` commands fail. As expected.

jjudd:
After digging into the code, I realized that my attempt to remove that lockfile entry was misguided.

I have a fix. I don't know if it is a good fix, but I will open a PR tomorrow morning.

jjudd:
@mstijak #2561 should fix this and has been merged. It hasn't been released yet, though. When you have a moment, would you be able to confirm and close this? Thanks!

PR #2561, opened 2017-01-26
Fix "Cannot read property 'split' of undefined" for interconnected file based dependencies

jjudd:
**Summary**
Currently, when you have interconnected file dependencies, yarn install fails after a lockfile has been generated. This failure occurs because yarn calls `split` on the package's resolved url, which is `undefined`. This fix makes sure we have a resolved url before trying to call `split` on it.

This is described in #2256. You can find a repo that reproduces this issue at https://github.com/mstijak/yarn-file-dep-repro

As described in #2256, running `yarn install` on the above repo produces the following lockfile:

```
# yarn lockfile v1


"a@file:./packages/a":
  version "1.0.2"
  dependencies:
    b "1.0.0"

b@*:
  version "1.0.0"

"b@file:./packages/b":
  version "1.0.0"
```

When you run `yarn install` a second, third, fourth, etc. time, yarn fails when trying to resolve the following package:

```
b@*:
  version "1.0.0"
```

It attempts to use a tarball resolver. In doing so, it calls split on a url that is undefined. My fix just checks makes sure the url is defined before trying to explode it. I'm not terribly familiar with yarn's code, so I'm not sure if this is a good fix or not.

**Test plan**
*To reproduce the bug*
1. Check out https://github.com/mstijak/yarn-file-dep-repro
2. Run `yarn install`. Notice that `yarn.lock` now exists.
3. Run `yarn install` again. This should fail with the following error
```
Trace: 
  TypeError: Cannot read property 'split' of undefined
      at Object.explodeHashedUrl (/home/james/yarn/lib/util/version.js:8:20)
      at PackageRequest.getLocked (/home/james/yarn/lib/package-request.js:128:59)
      at /home/james/yarn/lib/resolvers/registries/npm-resolver.js:202:37
      at next (native)
      at step (/home/james/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /home/james/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
      at Promise.F (/home/james/yarn/node_modules/core-js/library/modules/_export.js:35:28)
      at /home/james/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12
      at NpmResolver.resolve (/home/james/yarn/lib/resolvers/registries/npm-resolver.js:237:7)
      at /home/james/yarn/lib/package-request.js:183:23
```
*To verify the fix*
With the fix applied repeat the steps above. On step 3, `yarn install` should succeed without error.

**Gifs demonstrating the fix follow**
*Broken*
broken

*Fixed*
fixed


bestander:
Thanks for the fix, @jjudd.
Would you add a test please for this scenario?
Something similar to this https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L215

jjudd:
Sure thing. Working on it now.

jjudd:
I added the test @bestander. It fails without the fix and succeeds with it. Please let me know if you would like to do anything else.

__tests__/commands/install/integration.js
```diff
@@ -236,6 +236,13 @@ test.concurrent('install file: protocol without cache', async (): Promise<void>
   });
 });
 
+test.concurrent('install file: local packages with local dependencies', async (): Promise<void> => {
+  await runInstall({}, 'install-file-local-dependency', async (config, reporter) => {
+    const reinstall = new Install({}, config, reporter, await Lockfile.fromDirectory(config.cwd));
+    await reinstall.init();
```
bestander: Would be great to have an assert in the test that you have the expected state.

bestander:
Just a nit and ready to merge, thanks

__tests__/commands/install/integration.js
jjudd: What would the expected state be here? I'm fine with adding that, I just want to make sure I'm asserting the right thing. At the moment, I'm relying on the exception being thrown to fail the test. I could catch the exception and set a boolean and assert on that boolean or something similar.

bestander: It is either assert that exception not thrown or assert that module has been installed in node modules

jjudd: Makes sense. Thanks for the feedback. I updated the test to assert that the module is installed in node modules.

bestander:
Amazing PR, thanks @jjudd .
The tests, the description with the gif - great

jjudd:
Thanks for the kind words. Glad I could help out :)

PR #2568, opened 2017-01-27
Remove `babel-preset-es2015`.

wtgtybhertgeghgtwtg:
**Summary**
Removes `babel-preset-es2015` from `devDependencies`.  It looks like it stopped being used with https://github.com/yarnpkg/yarn/commit/862d20b9b3bdc87b7fb3fde22ccea8383bef4e06, which replaces it with `babel-preset-es2015-node4`, but was left as a devDependency.

**Test plan**
I don't think it changes much of anything.


Daniel15:
Seems legit to me. Thanks!

PR #2583, opened 2017-01-29
Use `string.prototype.repeat` over `repeating`.

wtgtybhertgeghgtwtg:
**Summary**
Drops the `repeating` package and uses `string.prototype.repeat`.

**Test plan**
Since it's an internal, no tests have been added that strictly test this.
A test has been added so that the affected lines in `upgrade-interactive` are covered.


wtgtybhertgeghgtwtg:
Tests failing on OSX, can't seem to find `/Users/travis/Library/Caches/Yarn/.roadrunner.json`.

Daniel15:
Just re-ran the build and the tests look fine. Thanks!

Issue #2214, opened 2016-12-10
yarn doesn't install packages properly, sometimes links to cache


mgol:
**Do you want to request a *feature* or report a *bug*?**

A bug.

**What is the current behavior?**
`yarn` sometimes doesn't install packages correctly:
1. Some packages have almost empty directories, containin just the `node_modules` folder and that's it.
2. Some binaries link to the yarn cache instead of locally installed packages.

**If the current behavior is a bug, please provide the steps to reproduce.**

With the following package.json:
```json
{
  "name": "test-package",
  "private": true,
  "dependencies": {
    "autoprefixer": "6.5.3",
    "babel-core": "6.20.0",
    "babel-loader": "6.2.9",
    "babel-plugin-lodash": "3.2.10",
    "babel-plugin-transform-runtime": "6.15.0",
    "babel-preset-es2015": "6.18.0",
    "babel-preset-es2016": "6.16.0",
    "babel-preset-es2017": "6.16.0",
    "chalk": "1.1.3",
    "css-loader": "0.26.1",
    "directory-tree": "1.1.1",
    "eslint": "3.12.0",
    "eslint-config-mgol": "0.0.34",
    "eslint-loader": "1.6.1",
    "extract-text-webpack-plugin": "1.0.1",
    "file-loader": "0.9.0",
    "html-webpack-harddisk-plugin": "0.0.2",
    "html-webpack-plugin": "2.24.1",
    "jquery": "3.1.1",
    "json-loader": "0.5.4",
    "koa": "1.2.4",
    "koa-compress": "1.0.9",
    "koa-etag": "2.1.1",
    "koa-fresh": "0.0.3",
    "koa-hbs": "0.9.0",
    "koa-router": "5.4.0",
    "koa-static-cache": "3.1.7",
    "lodash": "4.17.2",
    "node-sass": "3.13.1",
    "postcss-loader": "1.2.0",
    "sass-loader": "4.0.2",
    "semver": "5.3.0",
    "style-loader": "0.13.1",
    "supports-color": "3.1.2",
    "url-loader": "0.5.7",
    "webpack": "1.14.0"
  },
  "devDependencies": {
    "browser-sync": "2.18.2",
    "concurrently": "3.1.0",
    "nodemon": "1.11.0",
    "webpack-dev-middleware": "1.8.4",
    "webpack-hot-middleware": "2.13.2"
  }
}
```
[...]

PR #2584, opened 2017-01-29
Do not link bin scripts from dev dependencies with `--production`

bruderstein:
`bin` scripts for `devDependencies` when installing with `--production` should not be linked

**Summary**

When installing with `--production`, entries in `devDependencies` are not installed, but their `bin` scripts were linked to the `node_modules/.bin` directory. The symlinks were actually generated to the cache.

This change means that `bin` scripts are no longer added for devDependencies when in `--production` (actually for `ignore`d dependencies)

**Test plan**

In an empty directory
```
yarn init -y
yarn add eslint --dev
yarn add gulp
rm -rf node_modules
yarn install --production
ls -l node_modules/.bin
```

This shows the `eslint` binary linked to the cache. On my mac, that looks as follows:
```
$ ls -l node_modules/.bin
total 16
lrwxr-xr-x  1 dxxxxxxxxx  staff  107 Jan 29 21:24 eslint -> ../../../../../Library/Caches/Yarn/npm-eslint-3.14.1-8a62175f2255109494747a1b25128d97b8eb3d97/bin/eslint.js
lrwxr-xr-x  1 dxxxxxxxxx  staff   19 Jan 29 21:24 gulp -> ../gulp/bin/gulp.js
```

With the PR applied, only the `gulp` binary is linked in the `.bin` directory.

Added a test to test exactly this scenario.

This actually manifested itself as a problem in our CI environment, where the cache is owned by a different owner, so the `chmod` failed.


bruderstein:
Failing test runs seem to be timing?

bruderstein:
Fixes #2214

__tests__/fixtures/install/install-production-bin/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "devDependencies": {
```
bestander: This is an excessively large node_modules that gets installed when the test is running.
Could you find lighter ones for the test?

bruderstein:
@bestander fixed that - thanks for the tip! (Switched to `rimraf` and `touch`) The added test now takes 2ms \o/

The failure on travis is actually the other `--production` test timing out. Happy to switch those dependencies too, but that maybe shouldn't be part of this PR?

bestander:
Thanks, @bruderstein.
Would be great if you send a PR to fix the other test!

PR #2594, opened 2017-01-30
add Named Constructor for Config and create a real NoopReporter

voxsim:
These are the first steps to achieve some better understading of yarn internalities :)

First of all, I created a named constructor of the Config because I noticed that every time there was this pattern

```
const config = new Config(reporter);
await config.init({...});
```

now we have something like this

```
const config = await Config.create({...}, reporter); // if reporter is not NoopReporter
const config = await Config.create({...}, reporter); // otherwise we can do this
```

Pros:
- Removes many useless imports
- Better encapsulation
- Hides the temporal coupling between config constructor and init

Cons:
- I do not touch the initialization part in src/config.js because it hides too much complexity, maybe before it is useful add some unit testing ;)

Secondly, I created a real noop reporter

Pros:
- The time of tests on my machine drops from 91s to 64s
- Better clarification because if some tests really needs the BaseReporter, now is clear

Cons:
- Now NoopReporter extends BaseReporter, I preffered this way for now to be less invasive as possible

@bestander let me know if you like these tiny steps and after I will go with other tiny steps :)

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -114,15 +114,14 @@ export async function run<T, R>(
   }
 
   try {
-    const config = new Config(reporter);
-    await config.init({
+    const config = await Config.create({
       binLinks: !!flags.binLinks,
       cwd,
       globalFolder: path.join(cwd, '.yarn-global'),
       cacheFolder: flags.cacheFolder || path.join(cwd, '.yarn-cache'),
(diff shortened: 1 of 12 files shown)


bestander:
Wow, kudos for dropping tests time.
I think the little steps you've made are great, keep them coming, that is easy to review

PR #2604, opened 2017-02-01
CI: added retries to Travis runs

bestander:
Mac OS fails too often times, adding retries should reduce the noise

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
   include:
     - os: osx
       node_js: "6"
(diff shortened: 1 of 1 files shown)


wtgtybhertgeghgtwtg:
Why does it keep failing on OSX, anyway?

bestander:
OSX can only be a physical Apple device, usually it is a Mac mini and
companies can't buy enough of them.

So OSX CI are usually underpowered and maybe have worse internet connection
speed compared to the in cloud Linux containers.

That is why OSX timeouts more often I think.

Issue #2306, opened 2016-12-20
[feature] hint hoister algorithm


bestander:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the current behavior?**

Yarn hoists dependencies consistently but sometimes unpredictable.
A very common version of a package (core-js@2) can fail to hoist on top of node_modules tree if a less common package (core-js@1) got hoisted there first.

**What is the expected behavior?**

- Yarn can indicate that a dependency is getting hoisted too many times.
- Yarn can indicate how to fix package.json so that we get optimised hoisting
- package.json/.yarnrc can have hints for package-hoister to use as priorities

Daniel15:
Shouldn't Yarn be able to solve this by hoisting the "most popular" version of a package? How does it determine the version to hoist at the moment?

ljharb:
That's tricky tho, because what defines "most popular" - semver-latest? "latest" dist-tag on npm? most downloads reported on the registry? most packages requiring it in the current tree?

Daniel15:
> most packages requiring it in the current tree?

Sorry, I should have clarified. This is what I meant - relative popularity. If 10 dependencies/packages depend on core-js@2 and only one package depends on core-js@1, then core-js@2 should be the one that's hoisted.

ljharb:
Makes sense - what would be the deterministic tiebreaker?

Daniel15:
>  what would be the deterministic tiebreaker?

You mean if, for example, 5 packages depend on core-js@1 and 5 packages also depend on core-js@2? In that case, I think we should hoist the newest version.

gaearon:
Just a heads-up we plan to disable Create React App Yarn integration because of this.

benjaminapetersen:
+1 for this!  I just tried `yarn` for the first time, doing a simple 'hello world' install to play with `react`.  Side by side, `yarn` & `npm`:
- `131MB` install to `node_modules` via `nmp install`
- `556MB` install to `node_modules` via `yarn install`

Over half a gig worth of dependencies was quite a surprise. That said, it still looks like `yarn` is bringing some excellent features to the table!

hpurmann:
Isn't this fixed by https://github.com/yarnpkg/yarn/pull/2676?

farhan711:
Yes Isn't fixed by #2676

bestander:
#2676 gives an acceptable hoisting structure for most cases based on how often a dependency would be repeated if installed without hoisting.
[...]

PR #2620, opened 2017-02-03
--link-duplicates flag. Creates hardlinks to duplicate files in node_modules

bestander:
**Summary**
An attempt to mitigate inefficient hoisting for [create-react-app](https://github.com/facebookincubator/create-react-app/issues/1189) and https://github.com/yarnpkg/yarn/issues/2306.

Inspired by https://github.com/jeffbski/pkglink, instead of running copy for all repeated modules in node_modules this algorithm does only one copy and then hardlinks every file in the duplicate modules.

The win is in speed and size.

```
$ yarn install
yarn install v0.21.0-0
[2/4] 🚚  Fetching packages...
✨  Done in 112.94s.

$ du -sh node_modules/
530M	node_modules/

$ rm -rf node_modules/
$ yarn --link-duplicates
yarn install v0.21.0-0
[2/4] 🚚  Fetching packages...
✨  Done in 49.44s.

$ du -sh node_modules/
111M	node_modules/

$ npm -v
3.10.8
$ rm -rf node_modules/
$ time npm install
real	1m22.725s
$ du -sh node_modules/
126M	node_modules/

```
**Test plan**

- Added tests.
- Enabled by default and ran all integration tests
- tested on create-react-app


bestander:
Works on Mac and Windows 10 NTFS.
TODO: add test whether hardlinks are supported (should fail on fat FS)

wtgtybhertgeghgtwtg:
How does this relate to #2306?  Is the hoisting algorithm changed, as well?

bestander:
It does not implement #2306, we'll need to improve hoisting algorithm anyway.
This one just reduces the folder size

gaearon:
WOW.
Do you think we could enable this by default in CRA? Are there good reasons not to?

bestander:
If we don't find any bad side effects I think we could enable it by default everywhere.
Otherwise maybe CRA should have a .yarnrc file that would have this flag turned on.
Also we should fix the root cause (hoisting algorithm) anyway.

src/util/fs.js
```diff
@@ -251,6 +251,179 @@ async function buildActionsForCopy(
...
+  }
+  events.onStart(queue.length);
+
+  // start building actions
+  const actions: CopyActions = [];
+
+  // custom concurrency logic as we're always executing stacks of 4 queue items
+  // at a time due to the requirement to push items onto the queue
+  while (queue.length) {
+    const items = queue.splice(0, 4);
```
Daniel15: Maybe pull out the 4 as a constant (`const CONCURRENT_QUEUE_ITEMS = 4` or something like that, that explains what it means)

src/util/fs.js
```diff
@@ -251,6 +251,179 @@ async function buildActionsForCopy(
...
+          await access(dest, srcStat.mode);
+        } catch (err) {
+          // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
+          // us modes that aren't valid. investigate this, it's generally safe to proceed.
+          reporter.verbose(err);
+        }
+      }
+
+      // correct hardlink
+      if (bothFiles && (srcStat.ino === destStat.ino)) {
```
Daniel15: Does this work on Windows given Windows doesn't use inodes? Maybe this should check that `srcStat.ino` is not null, just in case?

src/util/fs.js
```diff
@@ -511,6 +734,22 @@ export async function writeFilePreservingEol(path: string, data: string) : Promi
   await promisify(fs.writeFile)(path, data);
 }
 
+export async function hardlinksWork(cwd: string): Promise<boolean> {
```
Daniel15: Maybe just call the param `dir`, as it could be called with any directory - It doesn't *have* to be the current directory.

__tests__/commands/install/integration-deduping.js
```diff
@@ -1,8 +1,10 @@
 /* @flow */
 
 import {getPackageVersion, runInstall} from '../_helpers.js';
+import * as fs from '../../../src/util/fs.js';
```
cpojer: why do you need to do `* as` here?

src/util/fs.js
bestander: Indeed there were some issues with ino on Windows back in Node 0.10 and 0.12 times https://github.com/nodejs/node-v0.x-archive/issues/2670.
The tests on Appveyor pass though.
I'll google a bit more to double check if we can rely on ino not being null

__tests__/commands/install/integration-deduping.js
bestander: You mean I could only `import stat` instead of `* as fs`?
Not sure if that would be better when the number of imports  from fs grow.
Everything grouped into one import makes sense rather than a lot functions that need sorting and formatting.
What should be the guideline here?

bestander:
Updated with fixes

bestander:
ping @Daniel15 @cpojer.
Good to be merged?

kittens:
I'm going to work on a prepass to optimise module hoisting today, this will go nicely in conjunction with it. Thank you so much @bestander!

STRML:
As this [tweet indicates](https://twitter.com/bestander_nz/status/833696608390836224), this may not make it to v21 final - anecdotally, this flag got our React app's install size down by 20%, whereas #2676 only reduced it by about 2%.

Unfortunately, `--link-duplicates` threw this error at the end:

>error An unexpected error occurred: "EEXIST: file already exists, link '/proj/frontend/node_modules/@kadira/storybook/node_modules/enhanced-resolve/node_modules/memory-fs/.gitattributes' -> '/proj/frontend/node_modules/babel-loader/node_modules/enhanced-resolve/node_modules/memory-fs/.gitattributes'".

Please indicate if you'd like me to move this to a separate issue.

bestander:
@STRML, if there this feature gives benefits to the community then we will keep it, thanks for speaking up.
Could you open a new issue and provide a way to repro?

STRML:
Sure, I'll try to get a minimal repo going then open the issue.

Issue #2192, opened 2016-12-08
When installing global modules, Yarn tries to create .cmd file in wrong directory


gaearon:
**What is the current behavior?**

Yarn fails to install `create-react-app`:

```
C:\Users\gaearon>yarn global add create-react-app
yarn global v0.17.10
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Installed create-react-app@1.0.0 with binaries:
      - create-react-app
error An unexpected error occurred: "EPERM: operation not permitted, open 'C:\\Program Files\\nodejs\\create-react-app.cmd'".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\gaearon\\AppData\\Local\\Yarn\\config\\global\\yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

I have no such problem with npm:

```
C:\Users\gaearon> npm i -g create-react-app
C:\Users\gaearon\AppData\Roaming\npm\create-react-app -> C:\Users\gaearon\AppData\Roaming\npm\node_modules\create-react-app\index.js
C:\Users\gaearon\AppData\Roaming\npm
`-- create-react-app@1.0.0
  ...
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn global add create-react-app
```

**What is the expected behavior?**

It gets installed.

**Please mention your node.js, yarn and operating system version.**

Node 6.9.2
Yarn 0.17.10
Windows 10 Enterprise 10.0.14393 Build 14393

**Yarn log**

https://gist.github.com/gaearon/e67437d7a4102fe696a34dec2c7fe825

gaearon:
If you work at FB you can repro this with a loaner Windows laptop.

Mikeysauce:
This is an issue with all global installs on windows in my experience. https://github.com/yarnpkg/yarn/issues/1324#issuecomment-256078658 is the effect and the solution. I noted that the path generated in the `example.cmd` file is incorrect whilst the path in the `example.cmd.cmd` is correct, hence deleting and renaming fixes this bug.

Daniel15:
Really bizarre, it's trying to write the file to C:\Program Files\Nodejs instead of in %LocalAppData% (C:\Users\gaearon\AppData\Local\ in your case). I remember seeing another issue about this.

juvasquezg:
The same with react-native 

```js
[...]

PR #2631, opened 2017-02-04
Fix typo in PATH added by Windows installer

neojp:
**Summary**

Global packages are currently not accessible through the command line because the PATH has been wrongly set through the Windows Installer c091ac4 & #1129

This gives the impression that global packages don't work on Windows at all and has been brought up in issue https://github.com/yarnpkg/yarn/issues/2192#issuecomment-277460434

**Test plan**

This has been tested to work on the original PR #1129. This should just fix a typo in the PATH.

### DIFF
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -39,7 +39,7 @@
 				<Environment
 					Id="BinPath"
 					Name="PATH"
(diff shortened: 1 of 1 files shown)


gaearon:
@neojp Are you interested in addressing the review comments? The issue still affects users so it would be great to get this fixed!

Daniel15:
This was fixed by #3233

PR #2635, opened 2017-02-05
Replace `user-home` with `os.homedir`.

### LINKED ISSUE #2596: Consider replacing `user-home` with `os.homedir`.
I don't think it counts as either. Reworking internals?

The `user-home` package is used to get the user home directory.

`os.homedir` is used to get the user home directory.

wtgtybhertgeghgtwtg:
**Summary**
Replaces `user-home` with `os.homedir`.  Proposed to close #2596.

**Test plan**
Refactors internals.  Nothing should change.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -12,7 +12,7 @@ import {addSuffix, removePrefix} from '../util/misc';
 import isRequestToRegistry from './is-request-to-registry.js';
 
 const defaults = require('defaults');
-const userHome = require('../util/user-home-dir');
+const userHome = require('../util/user-home-dir').default;
 const path = require('path');
 const url = require('url');
 const ini = require('ini');
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -12,7 +12,7 @@ import {addSuffix, removePrefix} from '../util/misc';
 import isRequestToRegistry from './is-request-to-registry.js';
 
 const defaults = require('defaults');
-const userHome = require('os').homedir();
+const userHome = require('../util/user-home-dir');
 const path = require('path');
 const url = require('url');
 const ini = require('ini');
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -12,7 +12,7 @@ import {addSuffix, removePrefix} from '../util/misc';
 import isRequestToRegistry from './is-request-to-registry.js';
 
 const defaults = require('defaults');
-const userHome = require('user-home');
+const userHome = require('os').homedir();
 const path = require('path');
 const url = require('url');
 const ini = require('ini');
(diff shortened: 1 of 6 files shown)


bestander:
I think user-home wraps os.homedir and has some important edge case handling.
@Daniel15 has some background on it IIRC

wtgtybhertgeghgtwtg:
`user-home` is just
```
'use strict';
module.exports = require('os-homedir')();
```
So the only edge case would be `node<4`.

bestander:
Lol, indeed it is https://github.com/sindresorhus/user-home/blob/master/index.js.

src/registries/npm-registry.js
```diff
@@ -12,7 +12,7 @@ import {addSuffix, removePrefix} from '../util/misc';
 import isRequestToRegistry from './is-request-to-registry.js';
 
 const defaults = require('defaults');
-const userHome = require('user-home');
+const userHome = require('os').homedir();
```
bestander: Considering that constants.js has custom edge case handling to homedir, can you extract that part into an internal module in `src/util`?
I think that edge case handling may cause some issues

wtgtybhertgeghgtwtg: Would you like me to use that internal module for all uses of `os.homedir`?  What would you like me to name that module?

bestander: Yeah, I think we should use one homedir definition everywhere.
Use your best judgement and lowercase letters, something with words user, home and dir maybe :)

wtgtybhertgeghgtwtg: Being up at 4 AM to fix PR's is the highlight of my day, so I wouldn't bank on my best judgement.
It needs access to `isRootUser` (which is exported) and `getUid` (which is not).  Would you like me to make `getUid` an export of `src/constants.js`, or move both to `src/util`?

wtgtybhertgeghgtwtg:
Tested failing because `babel-loader` [deleted the branch](https://github.com/babel/babel-loader/pull/353) [that was being checked](https://travis-ci.org/yarnpkg/yarn/jobs/199455023#L736).
Why does it check a branch of `babel-loader`, anyway?

bestander:
Thanks, @wtgtybhertgeghgtwtg.
I'll fix the test now.
We just needed a package to test that resolver works for this type of dependency.

bestander:
This should be safe to merge, all the other tests pass.

src/util/user-home-dir.js
```diff
@@ -0,0 +1,9 @@
+/* @flow */
+const path = require('path');
+const {ROOT_USER} = require('../constants');
```
connesc: `ROOT_USER` is undefined at this point due to the cyclic dependency.

Daniel15: Good catch, that explains #2807.

bestander: @connesc would you send a PR then?

connesc: Sure: #2855 :wink:

PR #2642, opened 2017-02-06
Added timestamps to verbose logs

bestander:
**Summary**

When debugging issues with verbose logs it is quite useful to know the timings of all operations.

**Test plan**
```
verbose 0.242 current time: 2017-02-08T02:47:26.486Z
...
verbose 0.445 Creating directory "/Users/bestander/work/temp/chek/node_modules/abbrev".
verbose 0.445 Creating directory "/Users/bestander/work/temp/chek/node_modules/buffer-equal".
verbose 0.445 Creating directory "/Users/bestander/work/temp/chek/node_modules/bunker".
verbose 0.445 Creating directory "/Users/bestander/work/temp/chek/node_modules/burrito".
```
Where 0.445 is in seconds since Yarn process was created.
I found full date/time stamps make the logs harder to read and not very useful repeated thousands of times.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -403,6 +403,8 @@ config.init({
   const exit = () => {
     process.exit(0);
   };
+  // verbose logs outputs process.uptime() with this line we can sync uptime to absolute time on the computer
+  reporter.verbose(`current time: ${new Date().toISOString()}`);
 
   const mutex: mixed = commander.mutex;
   if (mutex && typeof mutex === 'string') {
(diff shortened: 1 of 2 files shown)


cpojer:
Do you think it'd be more useful to show the timestamp since the last log message and uptime if there is no prior log message?

bestander:
That would complicate such a simple issue with state.
How about adding a verbose timestamp when console.reporter is instantiated so that whoever is tracking this could sync the time once

bestander:
Added timestamp to the beginning of the log.
Then people could match the uptime with the timestamp

Issue #2621, opened 2017-02-03
Feature Request: Allow Leaf Packages to Not Have package.json


jordwalke:
It's very convenient for packages to be able to depend on either git urls or http(s) urls (pointing to zipped bundles), where those endpoints do not have `package.json` files.

Why this helps: We often want to depend on a url/git-url, and right now we have to fork those bundles, or host our own versions of those zipped bundles, but with package.json added. With this simple feature, we would no longer have to.

I could imagine that if you encounter a leaf, you could just create a "virtual" package.json on the fly whose name matches the git url, and version is some hash of the URL. (I'm not saying this is best approach, but if the ease by which it can be implemented makes it more feasible, I'd be happy to see it implemented this way).

jordwalke:
Also, this would enable using Yarn to manage Vim plugins.

Issue #1011, opened 2016-10-13
Support for installing repos without a package.json


johans:
In npm there is this helper package napa (https://www.npmjs.com/package/napa) which enables one to install repos without a package.json file. Is this something you would consider adding support for in yarn?

bestander:
Will be implemented in #2651.

bestander:
duplicate #2621

joshxyzhimself:
Hi, why is #1011 and #2621 reopened after #2651 got merged?

Thank you.

joshxyzhimself:
Possible edge case.

This one below doesn't work, where `tradingview/charting_library` is a private repo, gets stuck on a continuous download (as seen in System Monitor, on Ubuntu Focal Fossa). Repo zipped is just `2.5 MB`, yet the download never ends.

```
yarn add ssh://git@github.com:tradingview/charting_library.git
```

---

However, this one below from a private fork works, after I added the `package.json` with the following contents; total download time is fast (as what is expected for a `2.5 MB` repo zipped)

```
{
  "name": "charting_library",
  "version": "0.0.0",
  "repository": "git@github.com:joshxyzhimself/charting_library.git"
}
```

```
yarn add ssh://git@github.com:tradingview/charting_library.git
```

floviolleau:
Hi,

Sorry to comment an old issue but it seems to not be possible to do it with yarn v3.

```
Manifest not found
    at Function.find (/home/.yarn/releases/yarn-3.3.1.cjs:390:29961)
    at async /home/.yarn/releases/yarn-3.3.1.cjs:593:12212
    at async Object.Gbe (/home/.yarn/releases/yarn-3.3.1.cjs:390:8896)
    at async fQ.resolve (/home/.yarn/releases/yarn-3.3.1.cjs:593:12173)
    at async zf.resolve (/home/.yarn/releases/yarn-3.3.1.cjs:391:1400)
    at async zf.resolve (/home/.yarn/releases/yarn-3.3.1.cjs:391:1400)
    at async /home/.yarn/releases/yarn-3.3.1.cjs:439:6346
    at async Ff (/home/.yarn/releases/yarn-3.3.1.cjs:390:8959)
    at async Z (/home/.yarn/releases/yarn-3.3.1.cjs:439:6328)
```

Any ideas?
Thanks

SKYBITDev3:
I updated yarn to v3.6.1 and got "Manifest not found" error. Please fix it ASAP.

PR #2651, opened 2017-02-07
[WIP] Feature: git dependencies can be without package json

bestander:
**Summary**

Implements #2621 #1011.
Adds feature from Bower where dependencies could be used without package.json/bower.json.

This is a dirty hack that breaks some assumptions that Yarn makes throughout the code.
Needs some cleanup.

**Test plan**

- added test
- yarn add <git url without package.json> gets it installed with generated package.json

### DIFF
--- a/src/resolvers/exotics/git-resolver.js
+++ b/src/resolvers/exotics/git-resolver.js
@@ -111,12 +111,11 @@ export default class GitResolver extends ExoticResolver {
       const {filename} = registries[registry];
 
       const file = await client.getFile(filename);
-      let json;
       if (!file) {
-        json = {};
-      } else {
-        json = await config.readJson(`${transformedUrl}/${filename}`, () => JSON.parse(file));
+        return null;
       }
+
+      const json = await config.readJson(`${transformedUrl}/${filename}`, () => JSON.parse(file));
 
       json._uid = commit;
       json._remote = {
@@ -144,7 +143,23 @@ export default class GitResolver extends ExoticResolver {
         return file;
       }
     }
-
-    throw new MessageError(this.reporter.lang('couldntFindManifestIn', transformedUrl));
+    // no registry files were found, in that case we fake a package.json thus allowing leaf dependencies without
+    // any registry file, same as bower
+    // TODO hack
+    const names = parts.path.split('/');
+    const registry = registries.npm;
+    const json = {
+      name: names[names.length - 1],
+      version: '0.0.0',
+      _uid: commit,
+      _remote: {
+        resolved: `${transformedUrl}#${commit}`,
+        type: 'git',
+        reference: transformedUrl,
+        hash: commit,
+        registry: 'npm',
+      }
+    };
+    return json;
   }
 }
--- a/src/resolvers/exotics/git-resolver.js
+++ b/src/resolvers/exotics/git-resolver.js
@@ -111,11 +111,13 @@ export default class GitResolver extends ExoticResolver {
       const {filename} = registries[registry];
 
       const file = await client.getFile(filename);
+      let json;
       if (!file) {
-        return null;
+        json = {};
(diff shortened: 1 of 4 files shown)


ljharb:
For what it's worth, I think this is a very bad idea. It's one of the worst parts of bower - that code I, the author, either did not intend, did not want, or am not prepared to support being installed by npm/yarn/etc, could be easily installed without my cooperation, means a future nightmare for authors (maintenance burden, bad PR when things don't work as intended), users (users will install a dep, and somewhere down the chain, there will be a bug, and since the author never intended the package to be used in this way, there's no support available to the user to fix the bug), and the whole ecosystem as a result.

andreypopp:
Probably that could be not the default behaviour but the one enabled by configuration?

src/resolvers/exotics/git-resolver.js
```diff
@@ -142,7 +142,23 @@ export default class GitResolver extends ExoticResolver {
...
     }
-
-    throw new MessageError(this.reporter.lang('couldntFindManifestIn', transformedUrl));
+    // no registry files were found, in that case we fake a package.json thus allowing leaf dependencies without
+    // any registry file, same as bower
+    // TODO hack
+    const names = parts.path.split('/');
+    const registry = registries.npm;
+    const json = {
+      name: names[names.length - 1],
```
andreypopp: That looks dangerous, what if `path` ends with `.git` suffix? The most logical way is to infer name from the key in `dependencies`:
```
"dependencies": {
  "desired-name": "some-repo-without-package-json"
}
```

Will be installed under `desired-name`.

As I understand that name is not available when request comes from CLI UI, right? That way maybe it's worth expanding CLI requests to `name=spec` syntax:

```
% yarn add desired-name=some-repo-without-package-json
```

andreypopp:
Also the same behaviour could be useful for tarball dependencies, not just git ones.

bestander:
@ljharb, thank you for sharing your view on this.

Back in the day when ComponentJS was a thing, it used the same filename for the package definition as Bower.

Some library authors did not want to deal with this fragmentation. And the community members were supporting those package managers in their forks.

I don't think that forcing people to fork if they want to consume a package is a better alternative to the down sides that you have described.
As long as lockfile will store the commit hash of the dependency I think we are quite protected from unexpected changes.

bestander:
@andreypopp, thanks for doing the review.

src/resolvers/exotics/git-resolver.js
bestander: Yeah, you are right.
The name parsing should be less naive

bestander:
> Also the same behaviour could be useful for tarball dependencies, not just git ones.

Interesting thought.
I'll look into if we can extract some common logic there

ljharb:
The concern is not about unexpected changes; it's about making it easy for people to use my software in a way that I don't support.

bestander:
@ljharb, noted.
If someone releases a library with a permissive license then people can consume those however they want.
If you don't support a package manager then don't, this feature is about package manager doing this work instead.

wtgtybhertgeghgtwtg:
Are there a lot of libraries that people want to use with `yarn` or `npm`, but do not support `yarn` or `npm`?

ljharb:
@bestander will yarn check the license of the git repo it's trying to install? If not, how can I programmatically prevent people from yarn installing my repo?

bestander:
I'll keep this PR open for some time but it needs being revamped

src/resolvers/exotics/git-resolver.js
sheerun: @andreypopp Name is available and the syntax is something like
`yarn add foobar@https://github.com/bestander/chrome-app-livereload.git`

bestander:
Implemented vi #3624

PR #2657, opened 2017-02-07
Optimize buildActionsForCopy routine

mourner:
**Summary**

Makes linking faster by reducing the number of file system calls in the `buildActionsForCopy` routine which synchronizes existing `node_modules` tree with the ideal one.

In my test runs (on `yarn` itself, and on [another big project](http://github.com/mapbox/mapbox-gl-js)), it saves about 400ms when running `yarn` without a lock file (or during upgrade).

- Remove redundant `fs.exists` (in favor of `lstat` which is called later anyway)
- Remove unnecessary `fs.access` (which does nothing — it's a placeholder in the code, so should be commented)
- Remove deeper directory traversal for `possibleExtraneous` since it should be populated with the same values later in the queue anyway.
- Avoid calling `mkdirp` if we know the destination folder already exists.

cc @bestander 

**Test plan**

Not necessary — it doesn't change the results of exported functions. Let's see if this PR passes the CI tests.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -146,9 +146,14 @@ async function buildActionsForCopy(
 
     let destStat;
     try {
+      // try accessing the destination
       destStat = await lstat(dest);
-    } catch (e) {}
+    } catch (e) {
+      // proceed if destination doesn't exist, otherwise error
+      if (e.code !== 'ENOENT') throw e;
+    }
 
+    // if destination exists
     if (destStat) {
       const bothSymlinks = srcStat.isSymbolicLink() && destStat.isSymbolicLink();
       const bothFolders = srcStat.isDirectory() && destStat.isDirectory();
@@ -157,11 +162,11 @@ async function buildActionsForCopy(
       // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
       // us modes that aren't valid. investigate this, it's generally safe to proceed.
 
-      // if (srcStat.mode !== destStat.mode) {
-      //   try {
-      //     await access(dest, srcStat.mode);
-      //   } catch (err) {}
-      // }
+      /* if (srcStat.mode !== destStat.mode) {
+        try {
+          await access(dest, srcStat.mode);
+        } catch (err) {}
+      } */
 
       if (bothFiles && srcStat.size === destStat.size && +srcStat.mtime === +destStat.mtime) {
         // we can safely assume this is the same file
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -144,21 +144,24 @@ async function buildActionsForCopy(
       srcFiles = await readdir(src);
     }
 
-    if (await exists(dest)) {
-      const destStat = await lstat(dest);
+    let destStat;
+    try {
+      destStat = await lstat(dest);
+    } catch (e) {}
 
+    if (destStat) {
       const bothSymlinks = srcStat.isSymbolicLink() && destStat.isSymbolicLink();
       const bothFolders = srcStat.isDirectory() && destStat.isDirectory();
       const bothFiles = srcStat.isFile() && destStat.isFile();
 
-      if (srcStat.mode !== destStat.mode) {
-        try {
-          await access(dest, srcStat.mode);
-        } catch (err) {
-          // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
-          // us modes that aren't valid. investigate this, it's generally safe to proceed.
-        }
-      }
+      // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
+      // us modes that aren't valid. investigate this, it's generally safe to proceed.
+
+      // if (srcStat.mode !== destStat.mode) {
+      //   try {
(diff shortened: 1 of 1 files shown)


mourner:
In the tests, there's only one failure (resolving `https://github.com/babel/babel-loader.git#greenkeeper/cross-env-3.1.4`) and it seems unrelated to the PR changes because the same failure happens in other recent PRs.

src/util/fs.js
```diff
@@ -144,21 +144,24 @@ async function buildActionsForCopy(
       srcFiles = await readdir(src);
     }
 
-    if (await exists(dest)) {
-      const destStat = await lstat(dest);
+    let destStat;
+    try {
+      destStat = await lstat(dest);
+    } catch (e) {}
```
bestander: there needs to be an explanation for a muted exception.
Could you add a one line comment?

src/util/fs.js
```diff
@@ -144,21 +144,24 @@ async function buildActionsForCopy(
...
-          await access(dest, srcStat.mode);
-        } catch (err) {
-          // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
-          // us modes that aren't valid. investigate this, it's generally safe to proceed.
-        }
-      }
+      // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
+      // us modes that aren't valid. investigate this, it's generally safe to proceed.
+
+      // if (srcStat.mode !== destStat.mode) {
```
bestander: Forgot to remove?

src/util/fs.js
mourner: Oh, sure. Basically, `fs.exists` is deprecated in Node, which recommends using `fs.stat` with error handling instead. In this case `destStat` will be undefined if error occurs during `lstat`.

src/util/fs.js
mourner: I thought it was left there as something to investigate in future, so I left it as a comment. Should I remove?

mourner:
@bestander addressed your first comment and rebased on master, so this PR should be good to go. Regarding the read mode leftover — it was there before my changes (as sort of a TODO to investigate), so I'm leaving it there as a comment. @kittens introduced it in https://github.com/yarnpkg/yarn/commit/4bbc653673bd65caf29084a30428cfa634166a66.

mourner:
The AppVeyor build failed with the following error, which seems unrelated to the PR changes:

> should expose `npm_config_argv` environment variable to lifecycle scripts for back compatibility with npm (#684)

src/util/fs.js
```diff
@@ -144,21 +144,29 @@ async function buildActionsForCopy(
...
+    if (destStat) {
       const bothSymlinks = srcStat.isSymbolicLink() && destStat.isSymbolicLink();
       const bothFolders = srcStat.isDirectory() && destStat.isDirectory();
       const bothFiles = srcStat.isFile() && destStat.isFile();
 
-      if (srcStat.mode !== destStat.mode) {
+      // EINVAL access errors sometimes happen which shouldn't because node shouldn't be giving
+      // us modes that aren't valid. investigate this, it's generally safe to proceed.
+
+      /* if (srcStat.mode !== destStat.mode) {
```
bestander: So what should we do with this code?
No reason to keep it commented out.

mourner: If no one from the Yarn team is willing to investigate cases where it's unsafe to proceed when the file read modes don't match, I'm OK removing the code. I'd leave the comment though, just in case.

bestander: This commit was added by @kittens in https://github.com/yarnpkg/yarn/commit/4bbc653673bd65caf29084a30428cfa634166a66.
I'll ping him to give some context

bestander: Ping me in 2 days if there is now answer, then let's just merge with this code removed.

TimvdLippe: @bestander Ping :wink:

bestander: Ok, let's remove the code and get back to it if we see errors reported.

mourner:
Sorry for the late response — I should probably wait before master starts passing CI tests more reliably before rebasing and suggesting to merge. Most recent commits in master fail on all servers.

bestander:
We have a few flaky tests but master got stable enough to get the signal if a PR is breaking something, so let's give this one one more try.

bestander:
@mourner, would you rebase?

mourner:
@bestander yes, will do!

mourner:
@bestander sorry for getting out of the loop. I see that Travis started having a failed test after this merge — do you think this is a consequence of this PR or an unrelated intermittent failure?

```
FAIL  __tests__/commands/add.js (174.708s)
  ● add with new dependency should be deterministic 3
```

arcanis:
It's probably related, previous tests weren't subject to this (it's usually timeout that kill our tests)

mourner:
OK, submitting a PR to revert a few lines that caused the failure until we figure out what's going on.

kittens:
There's a bunch of lint warnings on `master` introduced by this PR, could you open a PR to fix them?

PR #2663, opened 2017-02-08
Throw error on `--frozen-lockfile` failing check instead of simply bailing out

### LINKED ISSUE #2538: Error in yarn install --frozen-lockfile should result in non-zero exit code
I'd consider this a behavior change or bug.

When I run `yarn install --frozen-lockfile` with an out-of-date lockfile, I get an error message after the "Resolving packages" step. Then I get an exit code of 0, indicating success.

As a consequence of this, our CI system does not fail the build when the package.json file has been changed but the yarn.lock file has not been updated accordingly.

1. Check out the yarn repository at commit 93fa0f7c2857b5e9301d554b2b4c362d20988324
2. Remove all local files if necessary (i.e. `git clean -xdf`)
3. `yarn install`
4. Update the chalk dependency in package.json from "^1.1.1" to "^1.1.2"
5. `yarn install --frozen-lockfile`
6. `echo $?` to check that the exit code is 0

Whenever I run `yarn install --frozen-lockfile` and get an error message due to the lockfile being out-of-date, I get a non-zero error code.

Node: v6.2.2
Yarn: 0.19.1
Operating system: Mac OS X 10.11.6

lawliet89:
**Summary**

Fixes #2538.

If `--frozen-lockfile` is used while doing a `yarn install`, and it finds that the lockfile needs to be updated, it simply bails out with an error message. The return code is 0 in this case. We really should return a non-zero return code.

This switch was added for CI purposes. It's kind of silly it returns 0 on failure.

**Test plan**

Go to any directory with a `package.json` and `yarn.lock`, delete `yarn.lock`, and do:

```
$ yarn install --frozen-lockfile
yarn install v0.21.0-0
info No lockfile found.
[1/4] Resolving packages...
warning browser-sync > localtunnel > request > node-uuid@1.4.7: use uuid module instead
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
$ echo $?
1
```

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -303,8 +303,7 @@ export class Install {
     const haveLockfile = await fs.exists(path.join(this.config.cwd, constants.LOCKFILE_FILENAME));
 
     if (this.flags.frozenLockfile && !this.lockFileInSync(patterns)) {
-      this.reporter.error(this.reporter.lang('frozenLockfileError'));
-      return true;
+      throw new MessageError(this.reporter.lang('frozenLockfileError'));
     }
 
     if (!this.flags.skipIntegrity && !this.flags.force && match.matches && haveLockfile) {


mmun:
The `info` line looks a bit off there.

lawliet89:
I deleted the lock file for the test.

Issue #2673, opened 2017-02-09
yarn vs npm install node_modules bloat


zuidell:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
Yarn is bloating the node_modules directory, 5x file count vs npm and 2.5x file size

**If the current behavior is a bug, please provide the steps to reproduce.**

Using create-react-app as an example
1. create-react-app exampleApp
2. ~110,000 files in ./exampleApp, ~180 MB
3. rm -r ./exampleApp/node_modules
4. yarn (similar results)
5. rm -r ./exampleApp/node_modules
6. npm install    (instead of using yarn)
7. ~21,000 files in ./exampleApp, ~80 MB

**Folder Example:**
node_modules (after npm install, 24 KB)
-> babel-helpers 
-> -> lib
-> -> package.json
-> -> README.md

node_modules (after yarn install, 2 MB, everything is the same except sub node_modules)
-> babel-helpers
-> -> lib
**-> -> node_modules**
-> -> package.json
-> -> README.md

**What is the expected behavior?**
I wouldn't expect a yarn install to have 5x the files of an NPM install on the same package list.

**Please mention your node.js, yarn and operating system version.**
node - 7.5.0
yarn - 0.19.1
macOS Sierra Version 10.12.3 (verified also happening on Windows 7)

rally25rs:
For an application that I have here to test with:

NPM version: 3.10.3
Yarn version: 0.19.1

after `npm install`:
> 217,615,400 bytes (350.6 MB on disk) for 45,462 items

after `yarn install`:
> 224,005,232 bytes (391.5 MB on disk) for 54,394 items

I started trying to compare the 2 node_modules folders to see what is different, and it looks like yarn and npm have different ways of determining what to do when multiple versions of a module are needed. It looks like Yarn duplicates more under submodules node_modules, and npm either gets lucky, or determines which module version has more dependencies and puts that version at the root, resulting in less sub-module duplication.

In other words, if multiple versions of a module are needed, NPM and Yarn seem to put different versions at the root level, and then submodules that don't use that version at the root level get their own copy under the submodule's node_modules.

So conceptually I end up with something like:

If modules A, B and C all depend on module X.
A depends on X v1.
B and C depend on X v2.

npm:
```
A
 X v1
B
  (uses X v2 from root)
C
  (uses X v2 from root)
X v2
```

Yarn:
```
A
  (uses X v1 from root)
B
  X v2
C
[...]

Issue #570, opened 2016-10-10
make `--pure-lockfile` default for `install`


bestander:
**Do you want to request a _feature_ or report a _bug_?**

_feature_

**What is the current behavior?**

Not passing `--pure-lockfile` for `install` command confuses me because it modifies the lock file while installing node_modules.
We agreed on semantics that `add/upgrade/remove` are to change dependencies and `install` is to consistently rebuild node_modules from lockfile.

Consistency gets lost when lockfile is modified depending on environment (version of yarn currently installed).

**What is the expected behavior?**

Not write yarn.lock or package.json when doing `yarn install`.
To update yarn.lock use `yarn upgrade`

**Please mention your node.js, yarn and operating system version.**

yarn 0.14

esphen:
I agree. There should be a discussion on why `yarn install` writes a lockfile by default on the first place, as it seems to be at odds with the entire lockfile concept. Why have a lockfile if it is not locking versions by default?

There is a case for `yarn install` **creating** a lockfile if none is present, i.e. when someone is converting a project to use `yarn`, but the rationale for always writing it is not clear. I agree with @bestander's opinion that only mutating actions should update the lockfile, i.e. `add/upgrade/remove`.

ide:
Should there be a way to modify the lockfile without `add/remove/upgrade` ex: in the scenario when you upgrade Yarn and it uses a new lockfile version?

bestander:
I suppose the option could be inveresed
```
yarn install --save-lockfile
```

fingermark:
I am also confused by this.  What is the reasoning for the default behavior?

bestander:
Afaik, there was no strong reasons for the default behavior.
The idea, I suppose, is to keep people's lockfiles "evergreen".

bestander:
BTW PR is welcome

thejameskyle:
I think the reason for it was that yarn was originally designed with a single `install` command which was split into `install/add/upgrade`

Guuz:
So, as to check if i understand this correctly:

`yarn` installs all dependencies and also modifies the lockfile. On a CI server you should use `yarn install --pure-lockfile`?
Why does the lockfile get modified during an install? Since you are not upgrading anything... Yarn should just install the packages as described in the lockfile, right?

Thanks for the explanation!

kittens:
The problem is that if the lockfile is pure by default then people are going to forget to update it since it'd be a separate command.

idris:
[...]

PR #2676, opened 2017-02-09
Prioritise popular transitive dependencies when hoisting

kittens:
This PR prioritises popular transitive dependencies when we perform hoisting. For example, currently if a  dependency one level deep depends on `a@1.0.0` then it's immediately hoisted to the top. If then two levels deep there are multiple dependencies relying on `a@2.0.0` then the top position has already been taken.

Instead, with this patch, we first do a prepass over the dependency graph and hoist transitive dependencies that have multiple versions according to what one is depended on the most.

While testing this patch I used a new [create-react-app](https://github.com/facebookincubator/create-react-app) project to test, the following are before and after sizes of `node_modules`:

```
Before: 478MB
After: 118MB
```


kittens:
Will look into the failures tomorrow. Just looks like I need to update a test that was impacted by this prioritization.

paulmillr:
Great patch!

118MB is still a lot. Unacceptable for people with slow connections. For folks looking for a better, smaller solution than CRA take a look at [Brunch](http://brunch.io). 38MB for a basic react skeleton.

src/package-hoister.js
```diff
@@ -87,6 +87,8 @@ export default class PackageHoister {
    */
 
   seed(patterns: Array<string>) {
+    this.prepass(patterns);
+
     for (const pattern of this.resolver.dedupePatterns(patterns)) {
       this._seed(pattern);
```
bestander: Do you need to seed the patterns again here?

fingermark:
Will this help #570 at all?

kittens:
@fingermark It's not related. This is to do with transitive dependencies not being hoisted because of conflicts causing a lot of duplicated dependencies.

bestander:
The CI should probably be fixed with a rebase

kittens:
All green! Merging!

bestander:
Awesome, let's cherry-pick it into 0.21.1

rally25rs:
I think this fixes #2673 ???

bestander:
Yes it should

markstos:
Does this change imply that different `yarn.lock` file structures will be generated after this patch is landed?

bestander:
No, this step takes place after all dependencies are resolved and
downloaded.

There is a recent commit that fixed a bug though when several compatible
entries in lockfile were merging.
It was happening anyway but was not reflected in lockfile save

jesstelford:
Is there any downside with this change potentially causing unexpected dependency changes when relying on an implicit dependency?

For example;

At Domain.com.au, we have a centralized build tool which itself wraps other tools, named `fe-build`. We then use this tool as a `devDependency` for all (200+) of our React components (similarish to `react-scripts`)

`my-component/package.json`
```json
{
  "devDependencies": {
    "fe-build": "^10.0.0"
  },
  "scripts": {
    "lint": "eslint"
  }
}
```

One of those dependencies might be a specific version of `eslint`:

`fe-build/package.json`
```json
{
  "dependencies": {
    "eslint": "^2.0.0"
  }
}
```

We may also depend on a bunch of other things in `my-component`, each of which could conceivably bring in their own version of `eslint` (or whatever tool).

The previous algorithm would always grab one particular `eslint` to hoist (eg; the version depended on in `fe-build`: `"eslint": "^2.0.0"`). The developer could then confidently use `yarn run lint` to execute the expected version of `eslint`.

However, with this change, it could be the case that if other dependencies each specify `"eslint": "^3.0.0"`, the version hoisted would no longer match the expected version for the developer. Ie; Running `yarn run lint` would execute `eslint@3.x.x` instead of `eslint@2.x.x`.

I'm not sure if this is an issue in reality, but I'd be worried this may crop up in the future as our dependency trees get bigger and bigger, especially when the pressures to deliver products outweigh updating dependencies in our build tool `fe-build`.

bestander:
That should not be a problem.

If you want to run a specific version of eslint in your project then eslint needs to be a top level dependency, not an implicit one.
The hoisting algorithm will put most used versions of subdependencies on top but it will respect commonjs/node dependency resolution logic, so every package that depends on a specific version of eslint will always get the right one either in its direct node_modules or node_modules in one of the parents.

jesstelford:
To clarify; `fe-build` in this example is relying on the behaviour of the old hoisting algorithm as part of its API. `my-component` purposely does not have a dependency on `eslint`, expecting it to be hoisted from `node_modules/fe-build/node_modules/eslint` into `node_modules/eslint`.

I understand this is somewhat akin to doing a `require('foo/lib/some-internal-file.js')` (relying on internal APIs), but until this point, that API has been the same and using it has worked well for us. 

To be clear; I don't actually think this will come up, but if it did, it would be a hell of a thing to debug, and would cause a lot of head scratching for devs.

bestander:
Thank you for clarifying.

Yarn and npm don't have a feature to guarantee hoisting structure, this
particular limitation seems quite rare to me.

I would suggest to work around this at the level of fe-build or your
application so that if fe-build/eslint got hoisted it would not break.

From the top of my head, add eslint >2 to your application dependencies,
then fe-build/eslint would not be hoisted as long as it is 2.0.

Issue #605, opened 2016-10-11
Support local file dependencies without file protocol


ChristopherBiscardi:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
In a number of my projects I have a package.json with dependencies as such:

``` javascript
dependencies: {
    "babel-preset-something": "../ui-babel-preset-something",
}
```

`yarn` fails:

```
> yarn
yarn install v0.15.0
info No lockfile found.
[1/4] 🔍  Resolving packages...
error Couldn't find package "babel-preset-something" on the "npm" registry.
info Visit http://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

``` shell
> cd yarn-test
> ls
> mkdir a b
> cd a
> ~/.yarn/bin/yarn.js init
yarn init v0.15.0
question name (a):
question version (1.0.0):
question description:
question entry point (index.js):
question git repository:
question author:
question license (MIT):
success Saved package.json
✨  Done in 1.97s.
> cd ../b
> ~/.yarn/bin/yarn.js init
yarn init v0.15.0
question name (b):
question version (1.0.0):
question description:
question entry point (index.js):
question git repository:
question author:
question license (MIT):
success Saved package.json
✨  Done in 1.88s.
> ~/.yarn/bin/yarn.js add ../a
yarn add v0.15.0
info No lockfile found.
[1/4] 🔍  Resolving packages...
error Couldn't find package "../a" on the "npm" registry.
info Visit http://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

**What is the expected behavior?**

Expected behavior is that running `yarn` installs the package into node_modules.

**Please mention your node.js, yarn and operating system version.**
- node v6.2.0
- yarn 0.15.0
- OSX 10.11.3

zertosh:
Try setting the protocol as `file://`

kittens:
Change

```
dependencies: {
    "babel-preset-something": "../ui-babel-preset-something",
}
```

to:

```
dependencies: {
    "babel-preset-something": "file:../ui-babel-preset-something",
}
```

And this should work. Wasn't aware that npm had these semantics for bare paths.

montogeek:
It is possible to keep it compatible with npm?

kittens:
@montogeek Compatibility is absolutely a priority. I'll mark this issue as a bug so we can get this fixed for the next release.

[...]

PR #2684, opened 2017-02-12
install package with file: protocol as default if it exists in the filesystem

voxsim:
This pull request should resolve https://github.com/yarnpkg/yarn/issues/605.

Some info:
- I added existsSync function to not spread async annotation everywhere
- I created an instance method normalize that 'do the stuff', in detail: 
  everytime we normalizePattern in package-request and it is a request we also normalize the range  (if it is a file it will preprend file: and do other stuff, read the comment on the method).
- normalizePattern is a static function and it is used in other classes, I prefered to do not touch this method

Let me know if you like it!
cheers,
Simon


bestander:
The tests will pass when you rebase

voxsim:
@bestander I will try to do the same with async in another pull request and after that we will decide which is the best :)

bestander:
👍

voxsim:
@bestander I rebased with master and I opened another pull request https://github.com/yarnpkg/yarn/pull/2723 ;)

bestander:
#2723 is better :)

PR #2687, opened 2017-02-13
Use `Object.assign` over `defaults`.

wtgtybhertgeghgtwtg:
**Summary**
Drops the `defaults` package and uses `Object.assign`.  While
```
this.config = Object.assign({}, DEFAULTS, this.config);
```
is a little less obvious than
```
defaults(this.config, DEFAULTS);
```
I thought it might be worth making a PR.

**Test plan**
Since it's an internal, no tests have been added that strictly test this.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -11,7 +11,6 @@ import Registry from './base-registry.js';
 import {addSuffix, removePrefix} from '../util/misc';
 import isRequestToRegistry from './is-request-to-registry.js';
 
-const defaults = require('defaults');
 const userHome = require('../util/user-home-dir').default;
 const path = require('path');
 const url = require('url');
@@ -139,7 +138,7 @@ export default class NpmRegistry extends Registry {
         await fs.mkdirp(mirrorLoc);
       }
 
-      defaults(this.config, config);
+      this.config = Object.assign({}, config, this.config);
     }
   }
 
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "commander": "^2.9.0",
     "death": "^1.0.0",
     "debug": "^2.2.0",
-    "defaults": "^1.0.3",
     "detect-indent": "^5.0.0",
     "ini": "^1.3.4",
     "inquirer": "^3.0.1",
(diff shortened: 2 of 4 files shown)


wtgtybhertgeghgtwtg:
Travis failing because `ember-cli` [is not resolving](https://travis-ci.org/yarnpkg/yarn/jobs/201004921#L2258-L2265).

src/registries/npm-registry.js
```diff
@@ -139,7 +138,7 @@ export default class NpmRegistry extends Registry {
         await fs.mkdirp(mirrorLoc);
       }
 
-      defaults(this.config, config);
+      this.config = Object.assign({}, config, this.config);
```
bestander: I understand that defaults mutates the object instead of creating a copy.
Would it be a problem when passing a particular property across modules?

bestander:
Try rebasing to fix CI test

src/registries/npm-registry.js
wtgtybhertgeghgtwtg: It might if you're sharing a config object between registries and relying on outside mutations.

bestander: Could you check if this is the case for the options added here?

wtgtybhertgeghgtwtg: I don't think it is.  `this.config` is initialized [as an empty object](https://github.com/yarnpkg/yarn/blob/master/src/registries/base-registry.js#L29) when a registry is constructed, and its values are loaded from [a .ini](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L129) or [a lockfile](https://github.com/yarnpkg/yarn/blob/master/src/registries/yarn-registry.js#L81).
The only case I came come up with is if you have something like
```
const yarnReg = new YarnRegistry();
await yarnReg.loadConfig();
const config = yarnReg.config.
await yarnReg.loadConfig();
// config and yarnReg.config will be different and may have different values if the lockfile has been edited between `yarReg.loadConfig` calls.
```

bestander: Thanks for looking into this, @wtgtybhertgeghgtwtg

wtgtybhertgeghgtwtg: Thank you for merging.

PR #2689, opened 2017-02-13
Add global upgrade-interactive command (#2021)

### LINKED ISSUE #2021: yarn global upgrade-interactive
Hello guys,

I don't see any reason why `upgrade-interactive` shouldn't work for `yarn global` and it would be terrific. What do you think?

Feature

`yarn update-interactive` is awesome but it doesn't work for globally installed packages

`yarn global update-interactive` should do the same as `yarn update-interactive` for globally installed packages

Thanks,
Arthur

victornoel:
**Summary**

Introduce `upgrade-interactive` as a subcommand of `global`.

This closes #2021.

**Test plan**

This PR mimics the way the other subcommands of `global` works: the `upgrade-interactive` command is added as a subcommand of `global`. So it is very similar to how `upgrade` is implemented, nothing fancy.

`upgrade-interactive` already has tests, this commit is only about wiring stuffs together within `global`, I didn't see the need to introduce extra tests.

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -12,6 +12,7 @@ import {Install} from './install.js';
 import {Add} from './add.js';
 import {run as runRemove} from './remove.js';
 import {run as runUpgrade} from './upgrade.js';
+import {run as runUpgradeInteractive} from './upgrade-interactive.js';
 import {linkBin} from '../../package-linker.js';
 import * as fs from '../../util/fs.js';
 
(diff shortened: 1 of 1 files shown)


bestander:
The test plan is to double check that you have tested all the scenarios that this code affects.
Could you just write up (picture, gif, text) how this works now?
I agree, no need for new tests.

Could you rebase also please?

victornoel:
@bestander I did what I could for the test plan, I'm not totally sure it says what you expect it to say, don't hesitate to tell me if you need more details :)
I rebased, too bad I missed 0.21.0 :(

bestander:
Thanks, @victornoel.

Let's merge it.
Sorry for not being clear enough on what is a Test Plan in my understanding.
Usually I am expecting something like:
```
$ gulp --version
6.22.0
$ yarn global upgrade-interactive
? Choose which packages to update. (Press <space> to select, <a> to toggle all, <i> to inverse selection)    
 dependencies                                                                                                
❯◯ gulp                                6.22.0  ❯  6.23.0                                                                                       
# press space on a dependency
# press enter
$ gulp --version
6.23.0
```

The goal is to confirm that the feature was tested to be working, also it will be a good reference for people making changes after you as the functionality grows and becomes more complex

victornoel:
@bestander haaaaaaaaa, ok, sorry, I should have gone looking in other PR to see how it was, now I better understand! Next time I will better follow the PR format! Thanks :)

PR #2694, opened 2017-02-14
added prompt fix for add tests that fail occasionally because of wron…

bestander:
Trying to fix this type of failure https://circleci.com/gh/yarnpkg/yarn/2922


bestander:
Self-merging to fix CI builds

PR #2700, opened 2017-02-15
Fix tests when path location of yarn contains spaces

Keraito:
**Summary**

If the path location for Yarn contains spaces, the test suite would fail on a command. The corresponding upstream issue on Node is https://github.com/nodejs/node/issues/6803. This [specific comment](https://github.com/nodejs/node/issues/6803#issuecomment-219756302) noted that the path should be escaped with double quotes.

**Test plan**

1.  The path location to Yarn needs to contain spaces. 
2.  Run `yarn run test`.
3.  Observe a command fail with a corrupted path location.

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -22,7 +22,7 @@ Promise<Array<?string>> {
   }
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


bestander:
Much appreciated, thanks

Issue #1925, opened 2016-11-18
Cannot install packages from github when offline


onlywei:
**Do you want to request a *feature* or report a *bug*?**
Not sure

**What is the current behavior?**
Without an internet connection, I get:
```
[1/4] 🔍  Resolving packages...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
```
And then the install fails and tells me which github address it was trying to reach on the internet.

**If the current behavior is a bug, please provide the steps to reproduce.**
Just make a `package.json` with a github dependency, such as:
```
"nomnom": "douglasduteil/nomnom#next"
```

Install it via `yarn` while connected to the internet. 
Then unplug your internet, delete `node_modules` dir, and try to `yarn` again.

**What is the expected behavior?**
I expect it to install all my packages from cache even while offline.

**Please mention your node.js, yarn and operating system version.**
Node 6.9.1 LTS, yarn 0.17.4, Mac OS X 10.11.6

markstos:
There are conflicting interests here.

When you define that the version of a package you want is a Github branch, then there's no way to know if you've got the most recent version of the code from the branch without connecting to Github and checking.

Using a cached version could result an incorrect result.  It seems like the `--prefer-offline` flag exists for this issue. It's documentation says "'use network only if dependencies are not available in local cache'".

See if `yarn install --prefer-offline` works for you. Explicitly using this option usefully expresses that you are more interested in a cached result than one which might be inaccurate.

onlywei:
@markstos Thanks. Your response makes sense to me. I will see if pointing the packages to specific Commit Id SHAs works. 

To answer your question about the `--prefer-offline` flag, neither `--offline` nor `--prefer-offline` work for me. They both result in the same error:

```
error An unexpected error occurred: "https://github.com/someone/somerepo.git: getaddrinfo ENOTFOUND github.com github.com:443".
```

I also cannot find any documentation about these flags at https://yarnpkg.com/en/docs/cli/install.

Kitanga:
@gaearon also this is another issue that is related.

sce9sc:
Having the same issue. 
Windows 10, yarn version : 0.17.10
Followed the steps described https://yarnpkg.com/blog/2016/11/24/offline-mirror 
[...]

PR #2705, opened 2017-02-16
Fix a bug in the Hosted Git resolver, where it would make an HTTP request even if a mirrored tarball is present.

chrisgavin:
**Summary**
After mirroring a package locally, specifically [web-component-tester](https://www.npmjs.com/package/web-component-tester) and attempting to install the dependency offline from the mirror I discovered Yarn makes an HTTP request. This means the installation fails if you pass `--offline` to Yarn or don't have an internet connection.

If in offline mode the error "Can't make a request in offline mode" is displayed, otherwise an error about being unable to resolve github.com is shown.

It appears this problem occurs because the package in question references a shorthand GitHub dependency (`"test-fixture": "PolymerElements/test-fixture"`). A URL for this is created which gets passed through the hosted Git resolver. The hosted Git resolver then attempts to determine if this URL "hasHTTPCapability" by making a request to it. This fails in offline mode.

The fix I have made here is simply to assert that the URL "hasHTTPCapability" if a mirror tarball for it exists. Sorry if this is not the right fix, but I am not very familiar with the codebase so I tried to make the simplest change possible.

**Test plan**
```
git clone https://github.com/chrisgavin/fake-package
cd fake-package
yarn install
yarn install --offline
```

This fails before applying my patch, but succeeds afterwards.

### DIFF
--- a/src/resolvers/exotics/hosted-git-resolver.js
+++ b/src/resolvers/exotics/hosted-git-resolver.js
@@ -186,6 +186,10 @@ export default class HostedGitResolver extends ExoticResolver {
   }
 
   async hasHTTPCapability(url: string): Promise<boolean> {
+    if (this.request.getLocked('tarball')) {
+      return true;
+    }
+
     return (await this.config.requestManager.request({
       url,
       method: 'HEAD',


chrisgavin:
I might also hesitantly suggest this resolves #1925.

chrisgavin:
It seems I broke some tests. Closing this until I get a chance to fix them.

TimvdLippe:
@chrisgavin The test failure on AppVeyor is also happening with other PRs, so I think it is unrelated to your PR 😄

raymondwang:
@chrisgavin: This seems to work, and it's the singular feature preventing me from switching over to Yarn today. Can you reopen? I'd be happy to contribute unit tests if you don't have time as necessary.

PR #2720, opened 2017-02-18
Remove logic for loading self-update versions

Daniel15:
**Summary**
Updates `bin/yarn.js` to remove the logic that looks for an updated version of Yarn in `updates/current`. This will stop Yarn from loading older versions that were installed using `self-update`.

**Test plan**
Tried both `.\bin\yarn --version` and `node .\bin\yarn.js --version`, both still work as expected


bestander:
Thanks, Daniel

hpurmann:
Such a nice branch name 😜

Daniel15:
Haha, I only use the finest branch names :D

PR #2725, opened 2017-02-19
Fix user home config conflicting with the test suite

TimvdLippe:
**Summary**
The test suite did not properly isolate the home config from
influencing the behavior of the tests. Therefore if a user
used the global configuration and also cloned Yarn, the tests could fail.

A primary example of that was when a user set the prefix value.
If the user would set the prefix, the test suite would fail.

**Test plan**

1. Install `yarn` on your machine following https://yarnpkg.com/en/docs/install
2. `yarn config set prefix /usr/local` (or any other location)
3. Clone `yarn`
4. Execute the test suite of `yarn`

Without this fix, `react-native` would be installed in `/usr/local/bin/` and the test suite fails in `__tests__/commands/global.js`. Now the prefix is correctly set and `react-native` is installed in the proper location.

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -106,8 +106,6 @@ export async function run<T, R>(
 
   // remove the lockfile if we create one and it didn't exist before
   const lockfile = await createLockfile(cwd);
-  const homeFolderLocation = process.env.YARN_HOME_FOLDER;
-  process.env.YARN_HOME_FOLDER = path.join(cwd, '.yarn-home');
 
   // create directories
   await fs.mkdirp(path.join(cwd, '.yarn-global'));
@@ -131,14 +129,12 @@ export async function run<T, R>(
       production: flags.production,
     }, reporter);
 
-    const install = await factory(args, flags, config, reporter, lockfile, () => out);
+    const install = await factory(args, flags, config, reporter, lockfile, () => out, path.join(cwd, '.yarn-home'));
 
     if (checkInstalled) {
       await checkInstalled(config, reporter, install);
     }
   } catch (err) {
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
-  } finally {
-    process.env.YARN_HOME_FOLDER = homeFolderLocation;
   }
 }
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -106,12 +106,15 @@ export async function run<T, R>(
 
   // remove the lockfile if we create one and it didn't exist before
   const lockfile = await createLockfile(cwd);
+  const homeFolderLocation = process.env.YARN_HOME_FOLDER;
+  process.env.YARN_HOME_FOLDER = path.join(cwd, '.yarn-home');
 
   // create directories
   await fs.mkdirp(path.join(cwd, '.yarn-global'));
   await fs.mkdirp(path.join(cwd, '.yarn-link'));
   await fs.mkdirp(path.join(cwd, '.yarn-cache'));
   await fs.mkdirp(path.join(cwd, 'node_modules'));
+  await fs.mkdirp(path.join(cwd, '.yarn-home'));
 
   // make sure the cache folder been created in temp folder
   if (flags.cacheFolder) {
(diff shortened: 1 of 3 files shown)


src/registries/npm-registry.js
```diff
@@ -98,7 +98,7 @@ export default class NpmRegistry extends Registry {
   async getPossibleConfigLocations(filename: string): Promise<Array<[boolean, string, string]>> {
     const possibles = [
       [false, path.join(this.cwd, filename)],
-      [true, this.config.userconfig || path.join(userHome, filename)],
+      [true, this.config.userconfig || path.join(process.env.YARN_HOME_FOLDER || userHome, filename)],
```
TimvdLippe: Before coming up with this fix, I tried several others. Most notably I tried to pass on the config to override `this.config.userconfig` used on the same line. However, `this.config` is initialized as empty object in the `constructor` of the Registries. Passing on an `initialConfig` would work, but in [`cli/commands/global.js`](https://github.com/yarnpkg/yarn/blob/6656ee3717330fb93c616c0ce8d0d9c154585774/src/cli/commands/global.js#L39) this config is discarded and rewritten. Therefore that solution would require passing on config objects all over the place in order to properly inject the location.

TimvdLippe:
Rebased this PR to fix the import test failures, but now AppVeyor is still failing per https://github.com/yarnpkg/yarn/pull/2613#issuecomment-281302097

TimvdLippe:
I rebased this PR on master and all tests are passing now.

src/registries/npm-registry.js
bestander: This is a bit confusing because we have HOME defined in constants.js, npm-registry and yarn-registry.
Could you bring all of them into one place?

TimvdLippe: I have modified the `util-home-dir` to export a function instead. This allows for later extension if developers want to modify their home dir too. I hope that this was what you were looking for, else please let me know!

__tests__/commands/_helpers.js
```diff
@@ -9,6 +9,7 @@ import {run as check} from '../../src/cli/commands/check.js';
 import * as fs from '../../src/util/fs.js';
 import {Install} from '../../src/cli/commands/install.js';
 import Config from '../../src/config.js';
+import {userHomeDir, setUserHomeDir} from '../../src/util/user-home-dir';
```
bestander: I think for tests this can be achieved with a simple jest mock, https://github.com/yarnpkg/yarn/blob/ed6b9b99c4a83ec9ab8a113892ba900365316b5d/__tests__/commands/install/integration.js#L41

Then you won't need to do setter/getter and change the code

TimvdLippe:
@bestander thank you for your suggestion of using `jest.mock`, I was not aware of this functionality. I took the code of the integration test as example, but sadly I was unsuccesful. For some reason, reimporting `global` does not correctly use the mocked `user-home-dir`. I am unsure how I can make this work :cry:

Could you maybe point out what the issue is? Thank you in advance!

bestander:
@TimvdLippe, you can have a look at https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L38 how an internal js module can be mocked.
I think you are on the right path, what causes you the trouble?

TimvdLippe:
@bestander I have copied that to https://github.com/yarnpkg/yarn/pull/2725/files#diff-ae8f7de5bfa66fccd0d4a651ae01fc77R22

The problem right now is that when I run the test, the home directory does not seem to be mocked. I suspect that the problem is that `user-home-dir` is loaded by `constants` which in turn is loaded by various other modules. These modules are not reloaded after mocking, therefore not altering their usage of `user-home-dir`. Thus mocking for only `global` does not seem to work. However, I can't seem to figure out how to also reset these other usages. I tried to require all imports of the `global` testsuite, but that broke like everything :stuck_out_tongue:

bestander:
Yeah, you are right, the value from user-home-dir is imported into other projects.
Would it work if you mock `GLOBAL_MODULE_DIRECTORY` in `constants.js` instead?

TimvdLippe:
The latest commit tries to do so, but sadly still fails locally :sob: Maybe a Jest engineer could chime in to explain why it is not working?

PR #2731, opened 2017-02-20
fixed import version for not installed dependencies

bestander:
Import tests would grab latest version from npm for the lockfile.
This change should make the test pass.


bestander:
This fixes Appveyor, merging

Issue #2704, opened 2017-02-16
Incorrectly reports an unmet peer dependency upon peer upgrade


mjswensen:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Yarn incorrectly reports an unmet peer dependency when a satisfactory peer is upgraded.

**If the current behavior is a bug, please provide the steps to reproduce.**
Clone https://github.com/mjswensen/yarn-peer-dependency-bug. I've set up a small repo with two packages there that reproduce this issue with a single command (see [its README](https://github.com/mjswensen/yarn-peer-dependency-bug#readme)).

Here's this gist of what's going on:
* a main package depends on `foo@^2.0.0` and `bar`.
* bar has a peer dependency on `foo@^2.0.0`.
* within main, the user runs `yarn upgrade foo`, upgrading it from `2.0.0` to `2.1.0`.
* Even though `foo@2.1.0` still satisfies bar's peer dependency requirement (which was `foo@^2.0.0`), yarn reports `warning "bar@<version>" has unmet peer dependency "foo@^2.0.0".`

**What is the expected behavior?**
Yarn shouldn't warn when a peer-depended package gets upgraded if the new version still satisfies the peer dependency.

**Please mention your node.js, yarn and operating system version.**

    % yarn --version
    0.20.3

    % node --version
    v7.5.0

macOS Sierra 10.12.3 (16D32)

mjswensen:
I've got a fix for this and will send a pull request within the next few days.

hpurmann:
I guess this can be closed then, @bestander?

merriam:
Just an FYI, this can still be an issue.   

In may case, using yarn 0.21.3, node v7.7.1, I would expect no warning for `eslint-plugin-import` on this command:

> 
> yarn add eslint-config-airbnb eslint-plugin-import eslint-plugin-jsx-a11y eslint-plugin-prefer-object-spread eslint-plugin-react
> yarn add v0.21.3
> warning No license field
> [1/4] 🔍  Resolving packages...
> [2/4] 🚚  Fetching packages...
> [3/4] 🔗  Linking dependencies...
> warning "eslint-config-airbnb-base@11.1.1" has unmet peer dependency "eslint-plugin-import@^2.2.0".
> warning "eslint-config-airbnb@14.1.0" has unmet peer dependency "eslint-plugin-jsx-a11y@^3.0.2 || ^4.0.0".
[...]

PR #2733, opened 2017-02-20
Fix inaccurate peer dependency warning on upgrade (#2704)

mjswensen:
**Summary**

The motivation for this fix is outlined in #2704. (Yarn incorrectly warns about unmet peer dependencies upon upgrade of the dependency. This patch fixes that behavior.)

**Test plan**

I've created a repository where the issue can be easily reproduced: https://github.com/mjswensen/yarn-peer-dependency-bug. Trying the upgrade with the current version of yarn will incorrectly yield an `unmet peer dependency` warning, even though the package being upgraded still satisfies the peer dependency requirement of the other package. Use this fork to perform the upgrade and the warning will not appear, as expected.

Furthermore, with this fix in place, yarn will now more accurately report an `incorrect peer dependency` if the upgraded package falls outside the range required by the dependent package.

I have added automated tests to the suite for both of these scenarios.


mjswensen:
Looks like a similar AppVeyor failure is happening on master, and I don't believe it's likely to be caused by my changes. I'm happy to rebase after it gets fixed on master if necessary!

Issue #2714, opened 2017-02-17
Install regression between 0.19.1 and 0.20.0


SomeoneWeird:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

Package `skeletron` fails to install.

**If the current behavior is a bug, please provide the steps to reproduce.**

Installation works w/ `0.19.1`
```
➜ yarn --version
0.19.1
➜ yarn add skeletron
yarn add v0.19.1
warning No license field
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved 1 new dependency.
└─ skeletron@0.0.2
warning No license field
✨  Done in 0.72s.
```

Installation fails w/ `0.20.0` (also fails with latest stable `0.20.3` and latest prerelease `0.21.0`)
```
➜ yarn --version
0.20.0
➜ yarn add skeletron
yarn add v0.20.0
warning No license field
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/private/tmp/node_modules/findit/test/symlinks/dir1/does-not-exist'".
info If you think this is a bug, please open a bug report with the information provided in "/private/tmp/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

**What is the expected behavior?**

Package skeletron successfully installs.

**Please mention your node.js, yarn and operating system version.**

NodeJS 5/6/7

macOS 10.12.3 / Amazon Linux 2016.09

SomeoneWeird:
Looks like 65e89b5d02e06a15c65f13198c39271684c4f217 is the culprit

bestander:
@SomeoneWeird, thanks for finding the cause, I'll follow up with #2744 now

PR #2744, opened 2017-02-21
fix errors thrown if symlink source does not exist

canufeel:
**Summary**
#2454 introduced a regression as during install symlink sources wouldn't always exist at the time of symlink creation so as `fs.realpathSync` checks if the path actually exist the `ENOENT` error might be thrown here.

This PR makes sure we have more defensive approach by checking that symlink source exists before trying to derive absolute symlink path which might throw. So if source doesn't exist we would stick to more conservative relative symlinks otherwise where symlink source itself is not a requirement at least until the symlink is used.

Fixes https://github.com/yarnpkg/yarn/issues/2714

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -644,7 +644,12 @@ export async function symlink(src: string, dest: string): Promise<void> {
       await fsSymlink(src, dest, 'junction');
     } else {
       // use relative paths otherwise which will be retained if the directory is moved
-      const relative = path.relative(fs.realpathSync(path.dirname(dest)), fs.realpathSync(src));
+      let relative;
+      if (await exists(src)) {
+        relative = path.relative(fs.realpathSync(path.dirname(dest)), fs.realpathSync(src));
+      } else {
+        relative = path.relative(path.dirname(dest), src);
+      }
       await fsSymlink(relative, dest);
     }
   } catch (err) {


SomeoneWeird:
Someone able to look at this?

stefanpenner:
> So if source doesn't exist we would stick to more conservative relative symlinks otherwise where symlink source itself is not a requirement at least until the symlink is used.

Ya, I honestly I think yarn's custom symlink stuff shouldn't try to do anything funky here at all.  #2454 restores some functionality, but the eager reification causes the issue this PR is aiming to mitigate.

This PR does mitigate some additional scenarios, I do worry that it makes the behavior even less predictable. That trade-off may be fine though.

canufeel:
> This PR does mitigate some additional scenarios, I do worry that it makes the behavior even less predictable.

@stefanpenner This is indeed true. The proper solution seems to be using only one symlink type i.e. absolute path symlinks. Unfortunately i didn't have time to investigate in detail why the symlink problem happens now in the first place. It's either because of race-condition of some kind or because of the way `yarn` flattens the dependencies. In my experience this problem shows up with symlinks which are multiple level deep in `app/node_modules/some_dep/node_modules/...`.

If former is the case then probably the proper way to go would be to delay `symlink` creation till the point when all deps are already installed project-wise. If latter is true then some kind of more sophisticated symlink "correction" and resolve mechanism is required here.

Any ideas on that are welcome.

stefanpenner:
@canufeel  the current behavior is largely the result yarn trying to make symlinks relative: https://github.com/yarnpkg/yarn/commit/36d73cc3bb0e3c4afcea4f25b06cc76b7b5531fa

stefanpenner:
@canufeel if the yarn folks would prefer to keep the existing behavior (rather then restoring symlinks to work as they are expected to do), I do believe your PR improves the current state and should be considered.

bestander:
Thanks for @canufeel and @stefanpenner for the contribution.
The relative symlink behavior is needed to make caches work correctly, reverting to absolute symlinks could open other issues in this area.
I'll merge this and add a test.

bestander:
cc me if anyone finds a related issue in the future

bestander:
Also will patch 0.20 and 0.21

PR #2745, opened 2017-02-21
Unescapes ESC character, fixing #2439

### LINKED ISSUE #2439: reporter in `yarn check` broke for when rendering paths with #
*bug*

To repro  

```
$ git clone git@github.com:facebook/react-native.git
$ yarn install
$ rm -rf node_modules/chalk node_modules/symbol-tree

# install 0.18
$ yarn check
yarn check v0.18.1
error chalk not installed
error jsdom#symbol-tree not installed
error Found 2 errors.

# install 0.19
$ yarn check
yarn check v0.19.0
error "chalk" not installed
error "\u001b[2mjsdom#\u001b[22msymbol-tree" not installed
error Found 2 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.


```

arcanis:
Ping @bestander 

**Summary**

This PR fixes #2439 by automatically unescaping the ESC character (`\u001b`) from the stringified version of the arguments passed to `reporter.lang`.

I'm a bit concerned about potential security breaks that could occur should a third-party-providen string contain a literal ESC (let's say a package named "\u001bhello"), which is why this regex [is a bit more complex](https://regex101.com/r/mSWp0d/1) than it could be: it checks that the `\u001b` character isn't escaped before unescaping it. So:

```
'\\u001bhello' => '\u001bhello'
'\\\\u001bhello' => '\\\u001bhello'
```

This way, even if a package name (or anything else) contains "\u001b" somewhere in the argument, it won't be interpreted provided that the string has been escaped beforehand (ideally before adding the styles to it). That being said, I'm not sure this escaping is already done, so maybe this PR isn't the right place to be concerned about that.

WDYT? Would you prefer a simpler regex?

**Test plan**

screen shot 2017-02-21 at 5 43 41 pm

### DIFF
--- a/src/reporters/base-reporter.js
+++ b/src/reporters/base-reporter.js
@@ -37,7 +37,10 @@ export function stringifyLangArgs(args: Array<any>): Array<string> {
       return val.inspect();
     } else {
       try {
-        return JSON.stringify(val) || val + '';
+        const str = JSON.stringify(val) || val + '';
+	// should match all "u001b" that follow an odd number of backslashes and convert them to ESC
+	// we do this because the JSON.stringify process has escaped these characters
+        return str.replace(/((?:^|[^\\])(?:\\{2})*)\\u001[bB]/g, '$1\u001b');
       } catch (e) {
         return util.inspect(val);
       }
--- a/__tests__/reporters/base-reporter.js
+++ b/__tests__/reporters/base-reporter.js
@@ -113,3 +113,9 @@ test('BaseReporter.disableProgress', () => {
   reporter.disableProgress();
   expect(reporter.noProgress).toBeTruthy();
 });
+
+test('BaseReporter.termstrings', () => {
+  const reporter = new BaseReporter();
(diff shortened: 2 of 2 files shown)


src/reporters/base-reporter.js
```diff
@@ -37,7 +37,8 @@ export function stringifyLangArgs(args: Array<any>): Array<string> {
       return val.inspect();
     } else {
       try {
-        return JSON.stringify(val) || val + '';
+        let str = JSON.stringify(val) || val + '';
+        return str.replace(/((?:^|[^\\])(?:\\{2})*)\\u001[bB]/g, '$1\u001b');
```
bestander: simpler regex or an instruction how to read it would definitely be appreciated :)
Also a link to the issue would be great for people to have context why the change was made.

arcanis:
@bestander Done, I've edited my OP and rebased a new commit

bestander:
appveyor is broken for another reason.
Thanks! Merging.

Issue #2534, opened 2017-01-23
yarnpkg alias not working


beac0n:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
The alias for `yarnpkg` is not working:

Installing react and react-dom using yarnpkg...

events.js:160
throw er; // Unhandled 'error' event
^

Error: spawn yarnpkg ENOENT
at exports._errnoException (util.js:1022:11)
at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
at onErrorNT (internal/child_process.js:359:16)
at _combinedTickCallback (internal/process/next_tick.js:74:11)
at process._tickCallback (internal/process/next_tick.js:98:9)

**If the current behavior is a bug, please provide the steps to reproduce.**
OS: Arch Linux
Steps to reproduce:
install yarn, using the AUR: `yaourt -S yarn`
install `create-react-app`: `sudo yarn global add create-react-app`
create a new project with `create-react-app`: `create-react-app testProject`

**What is the expected behavior?**
The alias for `yarnpkg` should work.

**Please mention your node.js, yarn and operating system version.**
`> $ node --version                                                                                                                                                                                              
v7.4.0`

`> $ yarn --version                                                                                                                                                                                              
0.19.1`

Operating System: Arch Linux

Issue in `create-react-app` repo: 
https://github.com/facebookincubator/create-react-app/issues/1429

Daniel15:
Maybe it's possible the Arch Linux package doesn't come with a `yarnpkg` executable, just a `yarn` one? The Arch package is made by a third party rather than by the Yarn developers though, and I'm not quite sure how it's packaged.

AaronHarris:
I want to chime in, installing using the recommended Windows Installer method on Windows 10 Pro x64 does not seem to add a yarnpkg alias to the PATH. I assume the reason for this is there is only a yarn.cmd file, but no yarnpkg.cmd file in the .bin folder. Would it be sensible to make a windows symlink named yarnpkg.cmd and point it to yarn.cmd?

Daniel15:
@AaronHarris - Yeah, Windows only has a `yarn.cmd` file, not a `yarnpkg.cmd` file. I think it'd be easiest to add a `yarnpkg.cmd` file that simply calls `yarn.cmd`. Would you like to send a pull request?

AaronHarris:
Sure

PR #2746, opened 2017-02-21
Added yarnpkg.cmd batch shim for backwards-compatibility on Windows (#2534)

AaronHarris:
References: #2534, facebookincubator/create-react-app#1555, facebookincubator/create-react-app#1365

**Summary**

Calling yarnpkg on Windows previously did not call Yarn

```
'yarnpkg' is not recognized as an internal or external command,
```

`yarnpkg` command is used by create-react-app instead of `yarn` to prevent collision with other programs by that name in a user's PATH (See facebookincubator/create-react-app#1365). However on Windows, running `create-react-app` falls back to installation via npm because no `yarnpkg` command is found. This PR restores the expected functionality where Yarn is used to install packages for create-react-app if it is installed.

**Test plan**

Tested on Windows 10 x64. Words identically to existing `yarn.cmd`.

### DIFF
--- a/bin/yarnpkg.cmd
+++ b/bin/yarnpkg.cmd
@@ -1,0 +1,2 @@
+@echo off
+"%~dp0\yarn.cmd" %*


Daniel15:
Thanks!

PR #2747, opened 2017-02-21
add --prefer-online flag: use cache only if network is not available

voxsim:
**Summary**

After reviewing https://github.com/facebookincubator/create-react-app/issues/1235 we should have a --prefer-online flag that should use use cache only if network is not available and IMHO should be *default* option.

Right now we have:
1. the *default* one which it only downloads from the network; we can call it --online
2. --offline we use network only if dependencies are not available in local cache
3. --prefer-offline that fallbacks from offline to online if the package is not in the cache

**Test plan**

Right now I have no idea how to test it; npm-resolver doesn't have tests; and this is an idea, if someone liked it, i will go ahead and maybe I will rewrite everything.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -49,6 +49,7 @@ commander.usage('[command] [flags]');
 commander.option('--verbose', 'output verbose messages on internal operations');
 commander.option('--offline', 'trigger an error if any required dependencies are not available in local cache');
 commander.option('--prefer-offline', 'use network only if dependencies are not available in local cache');
+commander.option('--prefer-online', 'use cache only if network is not available');
 commander.option('--strict-semver');
 commander.option('--json', '');
 commander.option('--ignore-scripts', "don't run lifecycle scripts");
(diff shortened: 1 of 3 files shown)


bestander:
Yarn does not optimize for being on the most recent version for all subdependencies, on the contrary most of the effort goes towards hitting the network less.

Let's start from the end.
What is your real world test plan?

voxsim:
@bestander I think when you use yarn to develop something you usually want the last version from network and if you are offline for some reason, it's ok to download from your cache, and this is the case when you create a new react app.
On the contrary, when you use yarn for an infrastructure you will prefer to cache everything on your storage and download from it. What do you think about it?

Btw for both reasons, i will prefer to have --prefer-offline or --prefer-online to have as a default installation behaviour instead of --offline or --online.

bestander:
Ok, this may be useful for people who want the bleeding edge dependency chains, especially bigger library contributors.
Would an update work for you in that cse?
Like `yarn install --prefer-offline && yarn update`.
That is one command more but it would give you the desired behavior without extra complexity that needs tests and support.

voxsim:
Mmmh yes it could be a good trade-off ;) For now I will close this pull request and thanks @bestander :D

bestander:
Let me know if that does not work, we'll come up with something.

PR #2748, opened 2017-02-21
skip test in __tests__/index.js for win32 platform: it does some tric…

voxsim:
…ks in bash, it does not work with powershell

Thanks to @TimvdLippe we found that I broke the tests in AppVeyor with my pull request https://github.com/yarnpkg/yarn/pull/2613.

@bestander this pull request skip the test for win32 platform.


bestander:
👍

PR #2752, opened 2017-02-22
Personal access token for github over http - via env GITHUB_HTTP_AUTH

hmalphettes:
**Summary**

This PR supports importing `github:org/repo` over http even when the repository is private.
It is an alternative to using `git+ssh://git@github.com/org/repo` to force yarn to use git instead of http. When migrating from npm it is one less thing to refactor.

```
GITHUB_HTTP_AUTH=username:personal_access_token yarn install
```

To support importing dependencies hosted in a private github repo
over http, the user authentication needs to be passed:
https://developer.github.com/v3/auth/#basic-authentication

**Test plan**

This PR is a proof of concept: it does that with a shamelessly simple env
variable.
If this makes sense we can consider doing it cleanly.

For example: the current implementation will write the personal auth token in yarn.lock
Not an acceptable practice. Passing the authentication as headers is a good way to avoid that issue.

Let me know if this idea is worth working on.
Thanks!

### DIFF
--- a/src/resolvers/exotics/github-resolver.js
+++ b/src/resolvers/exotics/github-resolver.js
@@ -6,6 +6,7 @@ import HostedGitResolver from './hosted-git-resolver.js';
 export default class GitHubResolver extends HostedGitResolver {
   static protocol = 'github';
   static hostname = 'github.com';
+  static auth = process.env.GITHUB_HTTP_AUTH ? `${process.env.GITHUB_HTTP_AUTH}@` : '';
 
   static isVersion(pattern: string): boolean {
     // github proto
@@ -22,7 +23,7 @@ export default class GitHubResolver extends HostedGitResolver {
   }
 
   static getTarballUrl(parts: ExplodedFragment, hash: string): string {
-    return `https://codeload.${this.hostname}/${parts.user}/${parts.repo}/tar.gz/${hash}`;
+    return `https://${this.auth}codeload.${this.hostname}/${parts.user}/${parts.repo}/tar.gz/${hash}`;
   }
 
   static getGitSSHUrl(parts: ExplodedFragment): string {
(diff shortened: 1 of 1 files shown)


bestander:
I don't see a reason why not support this, so go ahead.
We need tests to make sure this does not get broken over time.

process.env sounds all right.
I would not want to write a token anywhere on disk for security reasons.

hmalphettes:
@bestander ok thanks for the review.

I had imagined that you would require using the secret in yarn's config a bit like yarn login does for npm. Very nice to stick to using an environment variable for passing those secrets at least for now.
I'll work on adding some tests and making sure the secrets are not showing up in the logs.

bestander:
I guess there is no reason to complicate things yet unless someone requests this feature.
If you add a test and maybe send a PR to the docs site then it should be good to go

bestander:
Ping @hmalphettes, do you have time to finish this up?
Otherwise let's close it and maybe someone else will pick it up

BYK:
@hmalphettes really appreciate the PR. Do you need any help from us for writing tests and getting this merged?

hmalphettes:
@BYK sorry for my silence, got really busy on some other tracks.
I wont be able to do much in the next couple of days either.
If this is something someone else want to pick up please do.

@BYK When I find some time to work on this, something helpful would be a pointer to a test for a close to what we need here. I think I can figure it out but if you have something on the top of your head, I appreciate it.

BYK:
@hmalphettes it's okay, this can wait :)

For tests you can look at `__tests__/resolvers/exotics/github-resolver.js` or `__tests__/commands/install` for end-to-end testing. I wouldn't merge this diff until it stops writing the credentials into the lock file though 😃

BYK:
@hmalphettes your time is almost up if you want to see this in Yarn 1.0! :)

CesarLanderos:
is this ever going to be merged?, i have this issue where i can't use yarn when compiling a docker images because i am using a private repo.

with npm i just set the credentials with `git config --global credential.helper store` and npm was able to clone the repos 😞

bestander:
@CesarLanderos, you are welcome to open a new PR based on this one and the feedback

bestander:
I'll close this one as it has not been acted for a long time

CesarLanderos:
@bestander i definitely will, thanks!

PR #2761, opened 2017-02-23
Add filename output to parse errors

dawsbot:
**Summary**
When running yarn with an invalid lockfile, the errors do not direct you to look at your lockfile. An example of one of these "vague" errors is `error An unexpected error occurred: "Invalid value type 2:0".`

In order to better point people towards the actual issue, this PR adds the filepath output after the error location so that people can find and fix the lockfile issue. After this PR, that same output would instead be `error An unexpected error occurred: "Invalid value type 2:0 in /Users/dawsonbotsford/.yarnrc".`

**Test plan**

This should only be a change in the error message. Are error messages included in the test coverage of yarn?

### DIFF
--- a/src/lockfile/parse.js
+++ b/src/lockfile/parse.js
@@ -156,9 +156,10 @@ export function* tokenise(input: string): Iterator<Token> {
 }
 
 export class Parser {
-  constructor(input: string) {
+  constructor(input: string, fileLoc: string) {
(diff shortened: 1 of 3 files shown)


bestander:
Good idea, just a few lint errors need to be fixed

bestander:
Great job!
Send more of those

Issue #2035, opened 2016-11-25
Yarn caches should be invalidated when switching versions


bestander:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the current behavior?**

We sometimes introduce changes to cache logic which make cache from previous version incompatible with the next one.

**What is the expected behavior?**

Yarn should ignore caches generated by another version.

torifat:
This has been the solution for most issues posted. `yarn cache clean`.

Daniel15:
Perhaps we should version the cache, and clear it if the version doesn't match (similar to how lockfiles are versioned). So, the current cache would have a version of `1`, then we'd bump it to `2` if any breaking changes are made.

bestander:
It would require some discipline but I agree with you, @Daniel15

bestander:
Bootcamped the issue

bestander:
Done!

PR #2768, opened 2017-02-23
Adds versioning to the cache path

arcanis:
cc @bestander 

**Summary**

Cf #2035 - sometimes the cache structure changes, and things break.

This commit adds an additional part in the cache path that allows to invalidate previous caches. These old caches will not get removed until after running `yarn cache clean`, which allows for some easy rollback, just in case.

It should not conflict with the current structure (old cache files will stay in the root cache directory until the next cache clean, but that shouldn't cause any other issue).

**Test plan**

The main effect is that the cache directory will change a bit:

```
mael-mbp:yarn mael$ ~/yarn/bin/yarn cache dir
/Users/mael/Library/Caches/Yarn/v1
```

Note that this variable will also be appended to custom paths:

```
mael-mbp:yarn mael$ ~/yarn/bin/yarn cache dir --cache-folder=/tmp/cache
/tmp/cache/v1
```


bestander:
What kind of tests could we implement here?

arcanis:
Not sure; testing it would require changing the constant value, which means that yarn would have to be rebuild while running the tests :/

bestander:
I think we could mock it with jest easily.
- install
- check contents of cache
- mock version to be 2
- install
- check the contents of cache

bestander:
Some more ideas:
- probably best to make it an integration test in __tests__/commands/install/integration.js
- any test case there has config with cache folder which you can access via config.cacheFolder, it would return something like `/var/folders/q1/mbtgtys10s18hb48c78_59wrw6jg9z/T/yarn-skip-integrity-check-1487865032287-0.49457392285559965/.yarn-cache/v1`
- it is possible to mock a module with Jest but then you need to make sure that config gets reloaded

arcanis:
I digged a bit

> any test case there has config with cache folder which you can access via config.cacheFolder, it would return something like /var/folders/q1/mbtgtys10s18hb48c78_59wrw6jg9z/T/yarn-skip-integrity-check-1487865032287-0.49457392285559965/.yarn-cache/v1

I can check that the path ends with the expected version number without issue.

> it is possible to mock a module with Jest but then you need to make sure that config gets reloaded

Now things start to get a bit hairy. Deleting the cache with `require.resolve` doesn't really work because jest hijacks the require process, and the actual cache is stored in `<jest runtime>._moduleRegistry` (which is not accessible from the tests afaik). I might be possible to use `jest.resetModules()`, but unfortunately 1/ it isn't recognized by Flow and 2/ it breaks two other tests (because they seem to depend on the `request` module internal state).

bestander:
It is ok to run the test in sequence (not .concurrent) so that it does not run with other tests in parallel.
However, would it work with ES6 proxy https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy?

arcanis:
> It is ok to run the test in sequence (not .concurrent) so that it does not run with other tests in parallel.

Yep, I'm already doing this - the issue is that the two affected tests somehow need the `request` module state to correctly work, otherwise the `request._getAuthedRequests` calls return the wrong results. I haven't yet looked *why*, but I suspect it's because the request used in the tests and the one used by the yarn code aren't the same (since we've reset them previously).

--- 

Thinking of it, I found a way to make it work. Running `request = require('request')` inside my `mockConstants` function fixes the reference we have, and further tests pass. However, I'm not completely satisfied, since it might introduce other bugs in the tests later that could be hard to debug for anyone not aware of this hack :/

arcanis:
I've just push my test - note that there's still a failure because Flow fails to find the `jest.resetModuleRegistry` function for some reason.

bestander:
1 test got broken.
Can you replace resetModuleRegistry with resetModules like in https://github.com/facebook/react-native/commit/4aabf4b6b384a695f09e21239d1d13d5e329e1ef#diff-c7958b7b0499b2e8e26ceea7b11c19d1?

BTW, jest got just updated to 19.0.0 in master, you may want to rebase

arcanis:
Just rebased. Unfortunately, neither `resetModuleRegistry` nor `resetModules` are recognized.

bestander:
You'll probably need to grab https://github.com/flowtype/flow-typed/blob/c17c09b83fd7e71ee4044beb73d7740575ec4975/definitions/npm/jest_v18.x.x/test_jest-v18.x.x.js and replace https://github.com/yarnpkg/yarn/blob/master/flow-typed/npm/jest_v14.0.x.js with it

arcanis:
Should be ok! 👍 I also had to change the file to manually add `jasmine.addMatchers` and `expect.doesContainPackage`, but now Flow seems happy.

arcanis:
Just noticed that I'm still using `opts.cacheFolder` and `--cache-folder` to setup the newly-renamed variable `config.cacheRootFolder` - maybe it should be changed too, but then it wouldn't be backward compatible :/

bestander:
Mayve cacheFilder should be meaning "root"everywhere but for the versioned
 you can come up with a new term

__tests__/commands/install/integration.js
```diff
@@ -37,6 +56,21 @@ test.concurrent('properly find and save build artifacts', async () => {
...
+  await runInstall({}, 'install-github', async (config): Promise<void> => {
+    const inOut = new stream.PassThrough();
+    const reporter = new reporters.JSONReporter({stdout: inOut});
+
+    await cache(config, reporter, {}, ['dir']);
+    assert.ok(!!(JSON.parse(String(inOut.read())) : any).data.match(/\/v1\/?$/));
+
+    await mockConstants({CACHE_VERSION: 42}, async (config): Promise<void> => {
+      await cache(config, reporter, {}, ['dir']);
+      assert.ok(!!(JSON.parse(String(inOut.read())) : any).data.match(/\/v42\/?$/));
```
bestander: I think you should add a more high level check that verifies connection between current CACHE_VERSION and what gets installed in node_modules.

E.g.
1. Install package-a
2. add am empty file MARKER to the cache folder of package-a
3. install package-a (with --skip-integrity-check flag) and verify that MARKER was copied
4. set CACHE_VERSION=42
5. install package-a (with --skip-integrity-check flag) and verify that MARKER was removed in node_modules
6. set CACHE_VERSION=1
7. install package-a (with --skip-integrity-check flag) and verify that MARKER was copied

bestander:
Awesome job!
A test needs to be expanded.
Also note Windows CI, the tests fails on Appveyor

src/config.js
```diff
@@ -232,10 +236,16 @@ export default class Config {
       networkConcurrency: this.networkConcurrency,
     });
 
+    this.cacheFolder = String(
```
bestander: After thinking about this I think this.cacheFolder should contain the folder with version.
Otherwise you will have to modify every command that reads cacheFolder from config

src/resolvers/registries/npm-resolver.js
```diff
@@ -89,8 +89,8 @@ export default class NpmResolver extends RegistryResolver {
     // find modules of this name
     const prefix = scope ? this.name.split(/\/|%2f/)[1] : `npm-${this.name}-`;
 
-    invariant(this.config.cacheFolder, 'expected packages root');
-    const cacheFolder = path.join(this.config.cacheFolder, scope ? 'npm-' + scope : '');
+    invariant(this.config.versionedCacheFolder, 'expected packages root');
+    const cacheFolder = path.join(this.config.versionedCacheFolder, scope ? 'npm-' + scope : '');
```
bestander: This will work for npm dependencies but not for other resolvers: git etc

__tests__/commands/install/integration.js
```diff
@@ -37,6 +57,48 @@ test.concurrent('properly find and save build artifacts', async () => {
...
+    await fs.writeFile(path.join(cachePath, 'yarn.test'), 'YARN TEST');
+    await fs.unlink(path.join(config.cwd, 'node_modules'));
+
+    const firstReinstall = new Install({}, config, reporter, lockfile);
+    await firstReinstall.init();
+
+    assert.ok(await fs.exists(path.join(config.cwd, 'node_modules', 'is-array', 'yarn.test')));
+
+    await mockConstants({CACHE_VERSION: 42}, async (config): Promise<void> => {
+      const firstReinstall = new Install({}, config, reporter, lockfile);
```
bestander: nit: it is secondReinstall then?

arcanis:
I've pushed a commit that changes `cacheFolder` to `_cacheRootFolder` and `versionedCacheFolder` to `cacheFolder`. However, there's still a disparity in the configuration parameters (`opts.cacheFolder` and `--cache-folder` are configuring `config.cacheRootFolder` instead of `config.cacheFolder`). Isn't it possible to delay merging this PR until the next major release, so that the options can be renamed as well?

On a similar note, I feel like `cacheRootFolder` should be public rather than private, since it is required to be able to fetch its value in order to derive a configuration from another (thing we do in the tests) or to just know the root cache path (thing we do in cache clean, but could be used in other contexts - for example if we were to list the cache directories for some reason).

bestander:
I am not sure when we have the next major release :)
- The goal is to not have breaking changes unless absolutely necessary, so I'd keep the argument `--cache-folder` for people to use.
- I am worried that if we keep --cache-root-folder public people will get confused which one they are supposed to use. I think --cache-root-folder should not be used by any class except config when it bootstraps

arcanis:
@bestander Should be fine now :) There's an error with Node 4, but I don't think it comes from this PR:

```
rm : Cannot find path 'C:\projects\yarn\dist\node_modules\core-js\client' because it does not exist.
```

Issue #2625, opened 2017-02-03
Allow yarn install to fail when a version doesn't exist


peterjanes:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**
`yarn install` will always prompt if a package version isn't found, and will never exit. This isn't desirable if it's run as a subprocess (e.g. `grunt-run`) or in an automated environment where an interactive shell isn't available.

**If the current behavior is a bug, please provide the steps to reproduce.**

Setup: my-package version 1.0.0 and 1.1.0 are available; my-package 1.2.0 is in local development.

```
$ yarn link my-package
yarn link v0.19.1
success Registered "my-package".
Done in 0.13s.
$ yarn install --modules-folder=my_node_modules --no-progress
yarn install v0.19.1
[1/4] Resolving packages...
Couldn't find any versions for "my-package" that matches "1.2.0"
? Please choose a version from this list: (Use arrow keys)
❯ 1.1.0 
  1.0.0 
```

**What is the expected behavior?**
`yarn` should have an option to not prompt and fail the install, e.g. `yarn install --non-interactive`

**Please mention your node.js, yarn and operating system version.**
Node 6.9.4, yarn 0.19.1, Fedora 25

PR #2770, opened 2017-02-23
Add —non-interactive flag for automated processes

matt-adam:
**Summary**

If yarn is used within an automated process (in a CI system, for example), interactive prompts are not desirable.  This adds a commandline flag to remove them.

This resolves https://github.com/yarnpkg/yarn/issues/2625.

**Test plan**

I intentionally misconfigured a package.json with an invalid version of async:

    "async": "^111.5.0",

1) Run 'yarn'.  Output is an interactive menu letting me select a valid version of async to install.
2) Run 'yarn --non-interactive'.  Output is:

yarn install v0.22.0-0
error Couldn't find any versions for "async" that matches "^111.5.0"

**Pending Issues**

* I wasn't sure how to build an effective test for this option.  If you think this warrants a test I'd appreciate some suggestions.
* I do not know how to update the documentation.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,12 +36,15 @@ export type ConfigOptions = {
   production?: boolean,
   binLinks?: boolean,
   networkConcurrency?: number,
+  nonInteractive?: boolean,
 
   // Loosely compare semver for invalid cases like "0.01.0"
   looseSemver?: ?boolean,
 
   httpProxy?: ?string,
   httpsProxy?: ?string,
+
+  nonInteractive?: ?boolean,
 
   commandName?: ?string,
 };
@@ -122,6 +125,8 @@ export default class Config {
   ignoreScripts: boolean;
 
   production: boolean;
+
+  nonInteractive: boolean;
 
   //
   cwd: string;
@@ -269,6 +274,8 @@ export default class Config {
 
     this.ignorePlatform = !!opts.ignorePlatform;
     this.ignoreScripts = !!opts.ignoreScripts;
+
+    this.nonInteractive = !!opts.nonInteractive;
 
     this.requestManager.setOptions({
       offline: !!opts.offline && !opts.preferOffline,
--- a/src/config.js
+++ b/src/config.js
@@ -44,8 +44,6 @@ export type ConfigOptions = {
   httpProxy?: ?string,
   httpsProxy?: ?string,
 
-  nonInteractive?: ?boolean,
-
   commandName?: ?string,
 };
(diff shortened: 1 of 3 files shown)


src/config.js
```diff
@@ -36,13 +36,16 @@ export type ConfigOptions = {
...
   networkConcurrency?: number,
+  nonInteractive?: boolean,
 
   // Loosely compare semver for invalid cases like "0.01.0"
   looseSemver?: ?boolean,
 
   httpProxy?: ?string,
   httpsProxy?: ?string,
 
+  nonInteractive?: ?boolean,
```
bestander: why 2 times?

src/resolvers/registries/npm-resolver.js
```diff
@@ -43,7 +43,7 @@ export default class NpmResolver extends RegistryResolver {
     const satisfied = await config.resolveConstraints(Object.keys(body.versions), range);
     if (satisfied) {
       return body.versions[satisfied];
-    } else if (request) {
+    } else if (request && !config.nonInteractive) {
```
bestander: Is it the only interactive place we have?

rkistner:
Can this also be enabled by default when using `--frozen-lockfile`? In that case it's not possible to persist the user's choice anyway.

bestander:
I wonder if --non-interactive is too a too broad name for this change.
As it only touches npm resolution.

src/config.js
matt-adam: Simple mistake, I will remove one of them.

matt-adam:
I'm happy to change the name for this option in whatever way you'd like.  I figured yarn may add additional interactive prompts in the future, and having a single option name was more predictable.

The only other place I saw a prompt was in upgrade-interactive, but I didn't think an explicitly-interactive command needed an option to disable interactivity.  I'm happy to add it to that command as well though if you'd prefer.

@rkistner I don't know anything about --frozen-lockfile, to be honest, so I'm not a great person to ask about this.  I am happy to enable non-interactive when frozen-lockfile is specified, if you are sure this is the correct behaviour.

bestander:
I think it is good to expand --non-interactive for other interactive commands in the future then.
Let's keep this implementation like this then.

bestander:
Thanks, @matt-adam

Ramblurr:
What about `yarn login`. I can't login programmatically:

```
error An unexpected error occurred: "Can't answer a question unless a user TTY".
```

Issue #2333, opened 2016-12-23
Lock file ignores `yarn-offline-mirror` config for git+ssh dependencies


albertfdp:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

~~When installing a dependency using `git+ssh`, the generated lock file doesn't reference the yarn offline mirror, but keeps pointing to the external dependency. The dependency however is correctly added to the offline mirror.~~

When installing a dependency using `git+ssh`, the first time the dependency is correctly added to the offline mirror and `yarn.lock` correctly references it. However, installing that same dependency again in a different project (so, resolved from yarn cache), it points to the external registry.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Add a yarnrc config to your project with the yarn-offline-mirror config
2. Run `yarn add webpack@git+ssh://git@github.com/webpack/webpack.git#v2.2.0-rc.2` or any private git dependency
3. Recreate the lock file, or install in a different project ~~so that dependencies points to the mirror~~
4. Now references of all npm packages point to the mirror, except the git+ssh webpack dependency

I have created a git repository showcasing the issue with the reproduction steps. See that the lock file has all dependencies pointing to the offline mirror except this one https://github.com/albertfdp/yarn-offline-cache-ssh/blob/master/yarn.lock#L1673

**What is the expected behavior?**

That the lock file points to the offline mirror tarball. This feature is probably used in combination with the `install --offline` flag, so access to the network makes it fail.

**Please mention your node.js, yarn and operating system version.**

* node 6.9.1
* yarn 0.18.1
* macOS 10.12

#### Note
23/02/2017 Edited to reflect a better description of the issue

staugaard:
Getting this fixed would be especially nice for private git dependencies. Then CI and deployments systems don't need to have git credentials.

pauloppenheim-gingerlabs:
I am investigating a workaround, by untar-ing the `yarn-offline-mirror` copy of the git ref into the cache myself. Notably, a normal cached version has a `.yarn-metadata.json` file. Here's how to workaround for now (you might need to change some paths):

```
yarn_cache_untar_hack() {
  # BAD BUG HACK temporarily manually fill the yarn cache with our git deps
  local YARNCACHEDIR=$1
  local project=$2
  local hashcode=$3
  local version=$4
  
  local cachedest="npm-$project-$version-$hashcode"
  local vendorsrc="$project.git-$hashcode"
[...]

PR #2775, opened 2017-02-24
Add failing test for issue #2333

albertfdp:
**Summary**

This PR tries to have a starting point to solve #2333. It reproduces the issue, so it can start a discussion on the solution.

screen shot 2017-02-24 at 13 43 08

**Test plan**

The test plan should be that this test passes.

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -314,6 +314,47 @@ test.concurrent('add with offline mirror', (): Promise<void> => {
   });
 });
 
+// broken https://github.com/yarnpkg/yarn/issues/2333
+test.skip('add-then-install git+ssh from offline mirror', () : Promise<void> => {
+  const mirrorPath = 'mirror-for-offline';
+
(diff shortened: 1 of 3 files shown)


albertfdp:
I'm very happy to work on a fix for it, but I might need some guidance.

__tests__/commands/add.js
```diff
@@ -314,6 +314,46 @@ test.concurrent('add with offline mirror', (): Promise<void> => {
   });
 });
 
+test.concurrent('add-then-install git+ssh from offline mirror', () : Promise<void> => {
```
bestander: can you make it `test.skip` now?
So that CI is green when we merge it

albertfdp: @bestander Sure, I just pushed it, thanks!

bestander:
All right, great start!
@albertfdp, would be great to get a fix next.
Ping me if you need some help.
The starting place would probably be git-fetcher.js https://github.com/yarnpkg/yarn/blob/master/src/fetchers/git-fetcher.js#L29 where it uses offline-mirror

PR #2778, opened 2017-02-24
Update jest to version 19

hpurmann:
Unfortunately this doesn't work in node7 but seems to be a bug in jest. It stops execution when testing remove, though returning a 0 exit code.


bestander:
nice!
Could you rebase?

hpurmann:
Done. The tests for `__tests__/commands/add.js` are hanging on my machine. I have to choose a matching version from the list to continue.

This is also the case with jest 16 though.

hpurmann:
@bestander can you rebuild circle? I think this failure is incorrect, flaky.

PR #2779, opened 2017-02-25
Add yarn version to lock file.

rasharab:
**Summary**

Adding the version of yarn used to generate the lock file into the lock file.

As a CI solution, we want to guarantee consistency. One of the ways is to know the explicit version of yarn used to generate the yarn.lock so that, when we come around to doing yarn on our build machines, we're using the exact version as the customer.

This is very similar to Cocoapods' Podfile.lock.

**Test plan**

We've altered tests to account for this, and have confirmed that the yarn.lock, when generated with the updated code, contains the version number.


bestander:
I appreciate the effort and understand where it is coming from.
But this change would cause yarn.lock changes every time it is running with new version of Yarn.
Sometimes different team members would have different versions of yarn and although compatible they would write to yarn.lock on every install.

We already have `yarn lockfile v1` in the lockfiles to track incompatible changes.
Developers introducing breaking changes to yarn structure are supposed to bump the version thus making Yarn aware of it.
Otherwise we should try to not introduce breaking changes at all.

reversefold:
It seems like this PR would only add a comment to the yarn.lock and not enforce the yarn version used to process the yarn.lock. Knowing that version of yarn generated the lockfile is useful for debugging build issues, *especially* when developers are using a different version than the build machines.

As someone who is in charge of a build system with hundreds of components I'd like this as well.

bestander:
How about enabling it behind a flag via .yarnrc?

PR #2780, opened 2017-02-25
`yarn upgrade` packages via protocol

juanca:
**Summary**

Whenever a dependency is not in a registry but is hosted elsewhere, upgrading the package throws an error. 

e.g. If a package is hosted on a private github repository, it can be installed via `yarn add git-url#sha`. The package is resolved correctly and set as a dependency (with a name). Attempting to `yarn upgrade <package name>` throws an error indicating a missing package in the registries. 

This also mirrors `npm upgrade <package name>` behavior.

**Test plan**

TDD. [See commit introducing the test](https://github.com/juanca/yarn/commit/6b8bcba397d2b421a6c751abf357441ba98e2f1e)


juanca:
Seems like my branch is outta date. I'm merging in `master` at the moment.

src/cli/commands/upgrade.js
```diff
@@ -19,6 +19,20 @@ export async function run(
...
-  const install = new Add(args, flags, config, reporter, lockfile);
+  let addArgs = [...args];
+
+  if (args.length) {
+    const [dependency] = args;
+    const manifest = await config.readRootManifest() || {};
+    const dependencies = manifest.dependencies || {};
+    const remoteSource = dependencies[dependency];
+
+    if (remoteSource && remoteSource.includes(':')) {
```
juanca: I'm a little concerned this is a bit too simple. Let me know if there are any preferred utility functions (assuming this codebase should have at least one).

juanca:
Looks like using `git+ssh://` does not work in AppVeyor -- switching to `https:`

src/cli/commands/upgrade.js
rally25rs: My first thought is that this check should probably be for the `remoteSource` matching any "exotic" resolver (git, bitbucket, file, tarball, etc...).

Might look at `package-request.js` which contains the function

```
static getExoticResolver(pattern: string): ?Function
```

juanca: Done -- I think it really add robustness to this.

__tests__/fixtures/upgrade/package-not-in-registry/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "yarn-test-git-repo": "https://github.com/juanca/yarn-test-git-repo.git#master"
+  }
```
bestander: Would it be possible to have a branch for this test in https://github.com/yarnpkg/e2e-test-repo?
Just to make Yarn CI more reliable

src/cli/commands/upgrade.js
```diff
@@ -19,6 +20,20 @@ export async function run(
   args: Array<string>,
 ): Promise<void> {
   const lockfile = args.length ? await Lockfile.fromDirectory(config.cwd, reporter) : new Lockfile();
-  const install = new Add(args, flags, config, reporter, lockfile);
+  let addArgs = [...args];
+
+  if (addArgs.length) {
+    const [dependency] = addArgs;
```
bestander: would this work with
```
yarn upgrade package-a package-b
```
?

juanca: This would not work with that.

I actually was curious whether that was a use case.

And also whether `args` / `addArgs` could be composed of other types of arguments (e.g. flags). 

If I assumed `args` consisted of only packages, I can easily switch this to a map-like function.

juanca: Just did some tests:

```
~/Projects/yarn (upgrade_git_packages *) $ yarn upgrade foo bar --ignore-index
yarn upgrade v0.17.4
yarn upgrade args: [ 'foo', 'bar' ]
```

I'll switch this out to iterate and appropriate build the new args.

__tests__/commands/upgrade.js
```diff
@@ -104,6 +104,24 @@ test.concurrent('upgrades from fixed version to latest', (): Promise<void> => {
...
+    const gitRemote = 'https://github.com/juanca/e2e-test-repo';
+
+    assert.equal(
+      lockfile[0],
+      `"yarn-test-git-repo@${gitRemote}#package-not-in-registry":`,
+      'Lockfile should point to the same branch.',
+    );
+
+    assert(
+      lockfile.indexOf(`${gitRemote}#d2027157d0c7188fc9ed6a6654325d1e3bf4db40`) === -1,
```
juanca: Welp, I'm really glad I'm adding multiple packages to this test. This particular assertion is actually not doing anything. As with any test asserting the non-existence of something, using the same assertion for the existence of something is required. This will shortly be fixed.

__tests__/commands/upgrade.js
```diff
@@ -104,6 +104,34 @@ test.concurrent('upgrades from fixed version to latest', (): Promise<void> => {
...
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const gitRemote = 'https://github.com/juanca/e2e-test-repo';
+
+    assert(
+      lockfile.includes(`"yarn-test-git-repo@${gitRemote}#package-not-in-registry":`),
+      'Lockfile should point to the same yarn-test-git-repo branch.',
+    );
+
+    assert(
+      lockfile.includes(`  resolved "${gitRemote}#5c57959f3c55a6cd6004e3855ca59ba98bfff56c"`),
```
juanca: This assertion is currently failing. Seems like there is a deeper issue with upgrading exotic packages. 

This [assertion was switched to be more correct](https://github.com/yarnpkg/yarn/pull/2780/commits/f9186622cde0686174abf5710a05a7f4c1e3e692):

> Replace assertion of non-existence with existence
> The assertion was incorrect and this particular assertion was testing nothing. The whole test itself still provides some coverage: packages not in registry does not blow up `yarn upgrade <package name>` command.

juanca: Fixed with `force: true` flag.

__tests__/fixtures/upgrade/package-not-in-registry/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "yarn-test-git-repo": "https://github.com/juanca/e2e-test-repo#package-not-in-registry",
```
juanca: These can be switched to `yarnpkg` once `package-not-in-registry` branch is created and [this PR is merged]().

bestander:
Thanks for following up with more tests!

PR #2787, opened 2017-02-27
[WIP] More helpful info on yarn upgrade

lukasoppermann:
Hey @bestander,

as suggested in https://github.com/yarnpkg/rfcs/pull/48, here is the empty PR. I have no idea where to start for this. 😞

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@
 * **More emojis.** 🐈
 
 ## Installing Yarn
-
 Read the [Installation Guide](https://yarnpkg.com/en/docs/install) on our website for detailed instructions on how to install Yarn.
 
 ## Using Yarn


bestander:
Cool, thanks for starting it.
I'll get back to you.

bestander:
@lukasoppermann, a few hints.

upgrade command is implemented in upgrade.js and all it does is calling init() method from add.js.
Add.js is a bit shallow too, it subclasses from install.js where all the installation is happening.
In add.js you have access to this.fetchRequestFromCwd() which returns patterns from you package.json file, that way you can filter patterns that are your direct dependencies from anything else.

lukasoppermann:
Hey, I looked into the files, I guess a change would be somewhere here: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/add.js#L133-L140 ? It is pretty confusing for me though, with all the files.

bestander:
If it is too intimidating feel free to close the PR and reopen it once you have something ready to review.

lukasoppermann:
Yeah, I don't think I have the understanding of yarn to build this feature. 😢

PR #2792, opened 2017-02-27
[WIP] Implement PJC spec

kittens:
This is a WIP tracking PR for my work to implement the [PackageJsonForCompilers](https://github.com/jordwalke/PackageJsonForCompilers) specification.

### DIFF
--- a/src/build/package-builder.js
+++ b/src/build/package-builder.js
@@ -1,22 +1,22 @@
 /* @flow */
 
-import type {Manifest} from './types.js';
-import type PackageResolver from './package-resolver.js';
-import type {Reporter} from './reporters/index.js';
-import type Config from './config.js';
-import type {ReporterSetSpinner} from './reporters/types.js';
-import executeLifecycleScript from './util/execute-lifecycle-script.js';
-import * as fs from './util/fs.js';
-import * as constants from './constants.js';
+import type {Manifest} from '../types.js';
+import type PackageResolver from '../package-resolver.js';
+import type {Reporter} from '../reporters/index.js';
+import type Config from '../config.js';
+import type {ReporterSpinner, ReporterSetSpinner} from '../reporters/types.js';
+import executeLifecycleScript from '../util/execute-lifecycle-script.js';
+import * as fs from '../util/fs.js';
+import * as constants from '../constants.js';
 
 const invariant = require('invariant');
 const path = require('path');
 
 const INSTALL_STAGES = ['preinstall', 'install', 'postinstall'];
 
-export default class PackageInstallScripts {
+export default class PackageBuilder {
   constructor(config: Config, resolver: PackageResolver, force: boolean) {
-    this.installed = 0;
+    this.installedCount = 0;
     this.resolver = resolver;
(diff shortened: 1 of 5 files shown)


src/build/package-builder.js
```diff
@@ -26,12 +26,18 @@ export default class PackageInstallScripts {
   needsPermission: boolean;
   resolver: PackageResolver;
   reporter: Reporter;
-  installed: number;
+  installedCount: number;
   config: Config;
   force: boolean;
 
+  getPJCConfig(pkg: Manifest): ?Object {
+    const obj: mixed = pkg.pjc || pkg.esy;
```
jordwalke: We have a bunch of packages already converted over, and we're just about ready to flip them over to the master branches, so that all the master branches will be ready for building on PJC spec. Currently, their `package.json` keys have `esy: {}`. We could change them to `pjc: {}` before flipping them over to the master branches, if that saves us one round of thrash, but maybe we can even convert them to their final form right now, if you have a good idea what this field should be called.

Some desirable characteristics of a `package.json` field name:

- Immediately clear that the field is intended to integrate with some other, specific and semi-formalized system. (As much as I love the field name `build: {}` it's not clear that the field conforms to some specification. `pjc` is clearly something specific, but not searchable.
-  `yarn:{}` might be a good candidate, but some might get the incorrect idea that we're not open to other package build systems (such as our own internal deployment build servers) also adopting the same spec. (It might still be worth creating it under `yarn: {}` though).

Importantly, I'd like to minimize thrash so if we can get it right from day one, all the better!

src/build/package-builder.js
```diff
@@ -94,15 +100,42 @@ export default class PackageInstallScripts {
...
     const ref = pkg._reference;
     invariant(ref, 'expected reference');
     const loc = this.config.generateHardModulePath(ref);
 
     try {
-      for (const [stage, cmd] of cmds) {
-        await executeLifecycleScript(stage, this.config, loc, cmd, spinner);
-      }
+      await this._installNPM(spinner, pkg, loc);
+      await this._installPJC(spinner, pkg, loc);
```
jordwalke: It looks like this will also respect the `--ignore-scripts` option, right?

jordwalke:
I wonder if there's a way to install a hook into yarn, that will allow running a custom build process *after* the entire typical postinstall process is completed. It could allow us to start using `yarn` master branch to install natively compiled packages, today - and then when a rewrite is complete, it would just continue to work, but better.

bestander:
This is not actionable right now but I'd like to keep it as a reference for the future attempts

jordwalke:
@andreypopp is straight up killing it if you're interested in checking out his progress: https://github.com/esy-ocaml/esy

He's now built two backends: One Makefile based and one JS based, so that you can either eject to run on hosts that have no node/opam/packagemanager installed and no network, or you can run on a host that have node available. I think there's some patches on top of yarn there which might also be helpful to have upstreamed. Thanks!

bestander:
So cool!
I'll get back to this to help out in the next half.

Issue #828, opened 2016-10-12
Garbage characters in Gulp.cmd after first install


dev-rowbot:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
I installed yarn in Windows as per the instructions and then ran yarn in a clean project directory - no existing node_modules. Yarn picked up the package.json file and pulled down all the correct modules. 

However, when I attempted to execute gulp using the command .\node_modules.bin\gulp it returned the error:
`s"  was unexpected at this time.`

I compared the gulp.cmd file with a version installed using npm3 and there are extra unexpected characters at the end of the file:

``` bat
@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe"  "%~dp0\..\gulp\bin\gulp.js" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "%~dp0\..\gulp\bin\gulp.js" %*
)s" %*
)
```

These characters should not be there: `s" %*)`

I removed the node_modules and reinstalled and there are no errors now

**If the current behavior is a bug, please provide the steps to reproduce.**
Steps as per above

**What is the expected behavior?**
The gulp.cmd should not have any bad characters after install

**Please mention your node.js, yarn and operating system version.**
- yarn - 0.15.1
- node - v4.4.5
- Windows 7

**Additional Info**

My package.json file is as follows:

``` json
{
  "name": "ci-scripts",
  "version": "1.0.0",
  "description": "CI Support Scripts",
  "main": "gulpfile.js",
  "dependencies": {
    "del": "^2.2.1",
    "gulp": "github:gulpjs/gulp#4.0",
    "gulp-cli": "^1.2.2",
    "jshint": "^2.9.2",
    "lodash": "^4.15.0",
    "node-etcd": "^5.0.3",
    "q": "^1.4.1",
    "request": "^2.74.0",
    "yamljs": "^0.2.8",
    "yargs": "^4.8.1"
  }
}
```

dev-rowbot:
For sanity's sake I also just check the bash script after deleting and reinstalling the node_modules and it is also broken

```bash
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node" ]; then
[...]

Issue #2356, opened 2016-12-28
npm run script broken on Windows


caseyWebb:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
```
C:\Code\UniversitySite [code-quality ≡ +0 ~1 -0 !]> npm run lint

> UniversitySite@2014.1.0 lint C:\Code\UniversitySite
> gulp lint

s" lint was unexpected at this time.
```

```
C:\Code\UniversitySite [code-quality ≡ +0 ~1 -0 !]> yarn run lint
yarn run v0.18.1
$ gulp lint
s" lint was unexpected at this time.
```

__note:__ it doesn't matter what the gulp task is.

**Please mention your node.js, yarn and operating system version.**

node v6.6.0
yarn v0.18.1
Windows 10

Everything works as expected on macOS and linux, Windows just doesn't want to behave (what's new...)

This may be a dupe, but I couldn't find anything OS specific in the issue tracker. If so, my apologies.

mike-morrison:
Try dropping the word run.  Just type "yarn lint".  It has been working for me on Window 10.  Installing global packages and publish are still broken in 18.1 on Windows 10 though.  At least for me.

OhBizzle:
I'm dealing with this issue as well, preventing use of yarn in production.

peterschmiz:
I have the same issue on Windows 10 :(

peterschmiz:
It looks like the generated `node_modules/.bin/gulp.cmd` contains an invalid part.

This is how its generated:

`@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe"  "%~dp0\..\gulp\bin\gulp.js" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "%~dp0\..\gulp\bin\gulp.js" %*
)s" %*
)`

and it should look like this (notice the missing **s" %*)**)

`@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe"  "%~dp0\..\gulp\bin\gulp.js" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "%~dp0\..\gulp\bin\gulp.js" %*
)`

@OhBizzle, @caseyWebb: changing this by hand fixed the issue for me!

narthollis:
[...]

PR #2795, opened 2017-02-27
Implemented mutex system to make bin links be generated consistently in Windows

DanReyLop:
Fixes https://github.com/yarnpkg/yarn/issues/828 and https://github.com/yarnpkg/yarn/issues/2356

**Summary**

`cmd-shim` (used to generate executables on the `.bin` dir on Windows) is not an atomic operation. So, a race condition can happen where 2 packages that produce the same CLI executable will produce an unusable `cmd` file.
Additional context: https://github.com/yarnpkg/yarn/issues/828#issuecomment-270035847
Examples of packages which produce the same CLI:
* `karma` and `karma-cli` produce the shortcut `karma`.
* `gulp` and `gulp-cli` produce the shortcut `gulp`.
* `grunt` and `grunt-cli` produce the shortcut `grunt`.

To solve this, I've implemented a Mutex (loosely based on [await-mutex](https://github.com/mgtitimoli/await-mutex)) around the usage of `cmd-shim`. It's not the most efficient way, but it's the most practical one, given how the linker is implemented.

A person has reported that this same bug happens in `npm`, but maybe it's less frequent because `yarn` is more aggressive with concurrency? Would it be better to implement this `Mutex` on `cmd-shim` itself? And if so, when can we expect Yarn to get the updated version of `cmd-shim`? Multiple packages having the same CLI are pretty common on build tools or test runners, and this bug basically makes Yarn unpredictable in Windows.

**Test plan**

Because this is a race condition, it's tricky to reproduce. I've managed to reproduce it with a complex project about 50% of the time.
* Go to a Windows machine
* Clone [woocommerce-services](https://github.com/Automattic/woocommerce-services), checkout the branch `try/yarn`
* Run `yarn install`
* Inspect the file `node_modules\.bin\karma.cmd`. If you see garbage characters at the end of the file, then you've managed to reproduce the bug (and if you are using this Yarn PR, it means I screwed up something).
* Alternatively, you can run `npm test` (it uses the `karma.cmd` program), if it gives an error before even starting, then that's the bug (ignore the Webpack tests, it's an old branch, if you get that far it means it's working).

### DIFF
--- a/src/util/mutex.js
+++ b/src/util/mutex.js
@@ -1,0 +1,18 @@
+/* @flow */
+
+const lockPromises: Map<any, Promise<*>> = new Map();
+
+/**
+ * Acquires a mutex lock over the given key. If the lock can't be acquired, it waits until it's available.
+ * @param key Key to get the lock for.
+ * @return {Promise.<Function>} A Promise that resolves when the lock is acquired, with the function that
+ * must be called to release the lock.
+ */
+export default (key: any): Promise<Function> => {
+  let unlockNext;
+  const willLock = new Promise((resolve) => unlockNext = resolve);
+  const lockPromise = lockPromises.get(key) || Promise.resolve();
+  const willUnlock = lockPromise.then(() => unlockNext);
+  lockPromises.set(key, lockPromise.then(() => willLock));
+  return willUnlock;
+};
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -11,6 +11,7 @@ import * as constants from './constants.js';
 import * as promise from './util/promise.js';
 import {entries} from './util/misc.js';
 import * as fs from './util/fs.js';
+import lockMutex from './util/mutex.js';
 
 const invariant = require('invariant');
 const cmdShim = promise.promisify(require('cmd-shim'));
@@ -24,7 +25,12 @@ type DependencyPairs = Array<{
 
 export async function linkBin(src: string, dest: string): Promise<void> {
   if (process.platform === 'win32') {
-    await cmdShim(src, dest);
+    const unlockMutex = await lockMutex(src);
+    try {
+      await cmdShim(src, dest);
+    } finally {
+      unlockMutex();
+    }
   } else {
     await fs.mkdirp(path.dirname(dest));
     await fs.symlink(src, dest);


src/util/mutex.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param key Key to get the lock for.
+ * @return {Promise.<Function>} A Promise that resolves when the lock is acquired, with the function that
+ * must be called to release the lock.
+ */
+export default (key: any): Promise<Function> => {
+  let unlockNext;
+  const willLock = new Promise((resolve) => unlockNext = resolve);
+  const lockPromise = lockPromises.get(key) || Promise.resolve();
+  const willUnlock = lockPromise.then(() => unlockNext);
+  lockPromises.set(key, lockPromise.then(() => willLock));
```
bestander: That looks a bit too fancy.
Is it necessary to go 3 promises to acquire a lock in single threaded JS?

DanReyLop: I've based it on `await-mutex` ([code here](https://github.com/mgtitimoli/await-mutex/blob/master/src/mutex.js)). I've honestly didn't try very hard to simplify it.

I'm actually pretty surprised that a fully-functioning mutex implementation can fit in so few lines of code.

bestander: Ok, looks reasonable, thanks!

wadim:
Awesome @DanReyLop! Would be great to get this released soon.

shawnmclean:
Is this released yet? I'm facing the same issue as #2356 on v0.23.2

DanReyLop:
Yes, this is already released. Are you sure your problem is the same?

shawnmclean:
Yes, it happened on the first run of switching from `npm install` to `yarn` on a windows server team city build server. It worked as normal on the second build. (I could paste the logs here if you want)

DanReyLop:
Hum... yes please. I don't think logs would tell the whole story, but they wouldn't hurt.

Maybe the `node_modules` folder was cached from one build to the next, and so this could be caused by the previous `npm install`? It's a long shot, but `npm` also has this problem (with lower probability). After this fix, I fail to see how the issue could possibly be happening.

bdukes:
We're also seeing this issue still in 0.23.2 (and saw it in 0.22.x, also).  This appears to be happening with fresh installs, not cached `node_modules` folders.  Is it possible this issue is in the yarn cache?  My understanding was that it would only happen during install, so clearing cache wouldn't help…

Vanuan:
#3097 proposes to reduce concurrency to 1. But maybe something is not right about this mutex implementation?

Let's look at a key. It's the source, not destination. So if there are multiple packages trying to install the same file in `.bin` it would still fail?

DanReyLop:
> Let's look at a key. It's the source, not destination. So if there are multiple packages trying to install the same file in .bin it would still fail?

You're totally right, the key should be `dest` instead of `src`, I mixed up the variables. So I guess this PR didn't fix anything at all. It appear to fix it for me, but since this is a race condition, you can never be sure.

I'll continue the discussion in #3097, thanks for bringing this up.

Issue #2132, opened 2016-12-03
Unmet peer dependency warning when the peer dependency is installed at the same time


fson:
**Bug**

**What is the current behavior?**

Installing a dependency and its peer dependency at the same time generates a warning "Unmet peer dependency" even though the peer dependency is installed.

Steps to reproduce:
```
> mkdir test
> cd test
> yarn init -y
> yarn add react react-dom
yarn add v0.17.10
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning Unmet peer dependency "react@^15.4.1".
```
On the other hand, first running `yarn add react` and then `yarn add react-dom` doesn't create any warnings.

**What is the expected behavior?**

No warnings about peer dependencies should be printed when other packages specified in the `add` command will satisfy the peer dependencies.

**Please mention your node.js, yarn and operating system version.**

Node.js v6.9.1
Yarn 0.17.10
macOS 10.12.1

andreiglingeanu:
Can confirm this. 

I've got while doing `yarn add babel-plugin-transform-vue-jsx`.

```
warning Unmet peer dependency "babel-helper-vue-jsx-merge-props@^2.0.0".
warning Unmet peer dependency "babel-plugin-syntax-jsx@^6.8.0".
```

connesc:
Same here with `yarn add request request-promise-native`:
```
warning Unmet peer dependency "request@^2.34".
warning Unmet peer dependency "request@^2.34".
```

idibidiart:
Same issue.

idibidiart:
low priority?

gaearon:
Would you like to send a PR to fix it?

idibidiart:
@gaearon 

do you have any idea what part(s) of the Yarn code the PR would have to touch (potentially)? :)

gaearon:
Not really :-). I'd just recommend searching for the error message string and then seeing where it's used.

andreiglingeanu:
@gaearon Yeah, right [here](https://github.com/yarnpkg/yarn/blob/3dc9cf43448da51d3b8890de339a0ed27b12ca62/src/package-linker.js#L320-L322) :)

kamilogorek:
After some debugging, and using `yarn add request request-promise-native` as a test case, I found that the main issue is this code:

[...]

PR #2796, opened 2017-02-27
fix: prevent from logging peer-dependencies that are met

ev1stensberg:
**Summary**

Solves the problem where peerDependencies are warning even though they are met. 

**Test plan**

Run `yarn add react react-dom` 

### Before: 

skjermbilde 2017-02-27 kl 23 56 20

### After: 
skjermbilde 2017-02-27 kl 23 57 29

I don't really know if the fix is legit. I removed a line of code, it may be doing something else than it was in my context, but the error is from `lib/package-resolver.js` L370, where it removes the peerDependencies we expect to have ( later adding the peerDep? ) . May need further investigation, but from the second check in `lib/package-linker` , it returns undefined for the package, since it has been removed from `this.patterns`. My suggestion, if this PR is somewhat volatile, is to check the packages you've got from the upper-tree against the ones you have, and then remove dupes. 

Related: 

- #2132 

If there's anything you need for me, feel free to assign me for other stuff. 

Even

### DIFF
--- a/src/package-resolver.js
+++ b/src/package-resolver.js
@@ -348,7 +348,6 @@ export default class PackageResolver {
     }
 
     byName.splice(byName.indexOf(pattern), 1);
-    delete this.patterns[pattern];
   }
 
   /**


bestander:
Looks like it is not that easy and broke some tests

ev1stensberg:
Yep. Assumed that! The bug might be more extensive than one might think. I ran yarn locally and it worked fine then, do you have an idea of what that might come from? The resolver is working as if should, but it may be that it gets the wrong information before resolving further layers upwards.

bestander:
Looks like it broke cases when packages get updated, i.e. patterns are replaced.
Your change seems logical so maybe there is some replace logic down the line that relies on this.patterns not being deleted yet.

ev1stensberg:
I'll have a look at it tomorrow. It's doable to update packages as it is doing in multiple if loops in resolve packages I think

PR #2806, opened 2017-02-28
[WIP] Fix git-fetcher offline mirror resolution

albertfdp:
**Summary**

This PR it's a follow up to https://github.com/yarnpkg/yarn/pull/2775 in an attempt to fix https://github.com/yarnpkg/yarn/issues/2333

**Test plan**

The failing test added in https://github.com/yarnpkg/yarn/pull/2775 should be green

### DIFF
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -24,6 +24,32 @@ export default class GitFetcher extends BaseFetcher {
     }
   }
 
+  async getResolvedFromCached(hash: string): Promise<any> {
+    const {reference: ref, config} = this;
+
+    const offlineMirrorPath = config.getOfflineMirrorPath();
+    if (offlineMirrorPath == null) {
+      // no mirror
+      return null;
+    }
+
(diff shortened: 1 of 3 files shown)


albertfdp:
Hi @bestander, I had a stab on this, and while I (think) found what the problem is, I might need some extra guidance.

From what I seen, the issue comes from the inconsistent usage of `hash` in git-fetcher, in some scenarios the `hash` is the commit while on others it's the tarball sha1. This is problematic, since the generic https://github.com/yarnpkg/yarn/pull/2806/files#diff-7246848d5d10d9c48ef4be28cb961582R38 will resolve different on the first request (using commit) than in the second (using sha1). That implies it's not actually read from the offline mirror.

I hope this was helpful.

bestander:
Good find, @albertfdp, this is a complex issue indeed.
We probably overloaded the `#hash` too much and I think fixing the issue may require a change to yarn.lock structure.

cc @arcanis who is working on the a related issue https://github.com/yarnpkg/rfcs/pull/51

bestander:
And thanks for diving into this @albertfdp, can I help you with something specific?

PR #2811, opened 2017-02-28
Fix #1435 (failing dependencies of optional dependencies)

### LINKED ISSUE #1435: Error when using yarn install on multi-platforms with optionalDependencies

bendemboski:
**Summary**

Mark all dependencies of optional dependencies as themselves optional, so that if they fail, the whole install process doesn't fail.

**Test plan**

I implemented integration tests for the install command for the cases of the install script failing, and platform incompatibility. I didn't find any similar tests for the add command, and since it appears that the code change is in a code path common to both, I though that would be fine.

I ran a number of manual test cases:

1. On linux, `yarn add electron-installer-dmg` (has optional dependency `appdmg`, which has normal dependency `macos-alias`, both of which are `"os": [ "darwin" ]`)
1. On MacOS, `yarn add electron-installer-dmg`, copy `package.json` and `yarn.lock` to linux and run `yarn install`
1. On MacOS, `yarn add appdmg`, copy `package.json` and `yarn.lock` to linux and run `yarn install`

All behaved as expected.

NOTE: This is a replacement for #1438


__tests__/fixtures/install/install-should-skip-incompatible-optional-sub-dep/.npmrc
```diff
@@ -0,0 +1 @@
+yarn-offline-mirror=./mirror-for-offline
```
bestander: Do you need offline mirror and .tgz files for the test?
I think you could be fine just using `file:` dependencies for tests

src/package-request.js
```diff
@@ -281,7 +281,8 @@ export default class PackageRequest {
       promises.push(this.resolver.find({
         pattern: depPattern,
         registry: remote.registry,
-        optional: false,
+        // dependencies of optional dependencies should themselves be optional
+        optional: this.optional,
```
bestander: What happens if a dependency is optional in one subtree and is not optional in another:
```
A -> (optional) B -> C
D -> C
```

Would it mark C as optional even though D requires it as a regular dependency?
Could you add a test for that to make sure we don't break it?

__tests__/fixtures/install/install-should-skip-incompatible-optional-sub-dep/.npmrc
bendemboski: No, I don't think it's needed. I was trying to make the changes fit with their surroundings and mimic [this](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L883) test which [uses a tgz file](https://github.com/yarnpkg/yarn/tree/master/__tests__/fixtures/install/install-should-skip-incompatible-optional-dep).

I'm happy to convert the new test to using `file:` dependencies. Should I also convert the one I modeled it after while I'm at it?

src/package-request.js
bendemboski: Pretty sure that would work fine because of [this code](https://github.com/yarnpkg/yarn/blob/6e7d396451ed7622a92f4a25d9370e5724c54635/src/package-reference.js#L112) but I'll add a unit test to prove it.

__tests__/fixtures/install/install-should-skip-incompatible-optional-sub-dep/.npmrc
bestander: Yeah, let's make it simpler in both places.
.tgz file are harder to review

src/package-request.js
bendemboski: Hmm...couldn't your scenario be simplified to

```
(optional) A -> C
B -> C
```

and still prove what you are concerned about? I'm not seeing why the first chain has to have a depth of 3 -- am I missing something?

bestander: Just a gut feeling, we had problems with deeper chains of dependency :)
Probably 2 layers should be good enough

bendemboski:
@bestander feedback addressed. Let me know if you'd like me to squash the commits.

bestander:
Thanks for the whole bunch of tests!

PR #2818, opened 2017-03-01
Refactored integrity generation and check

bestander:
**Summary**

I plan to improve integrity check code so that it could also verify node_modules structure.
This PR moves most of the integrity checking and saving code into a separate class, no actual changes to the logic.

**Test plan**

All tests should pass (integrity checks are covered with integration tests)

Pending things to do:
- a bit of cleanup

### DIFF
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -150,7 +150,7 @@ async function integrityHashCheck(
   const install = new Install(flags, config, reporter, lockfile);
 
   // get patterns that are installed when running `yarn install`
-  // TODO measure slowdown here
+  // TODO hydrate takes longer then install command bailout https://github.com/yarnpkg/yarn/issues/2514
   const {patterns: rawPatterns} = await install.hydrate(true);
   const patterns = await install.flatten(rawPatterns);
   const lockSource = lockStringify(lockfile.getLockfile(install.resolver.patterns));
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -157,10 +157,10 @@ async function integrityHashCheck(
 
   const match = await integrityChecker.check(patterns, lockfile, lockSource, flags);
   for (const pattern of match.missingPatterns) {
-    reportError('lockfileNotContainPatter', pattern);
+    reportError('lockfileNotContainPattern', pattern);
   }
   if (match.integrityFileMissing) {
-    reportError('noIntegirtyHashFile');
+    reportError('noIntegrityHashFile');
   }
   if (!match.integrityHashMatches) {
     reportError('integrityHashesDontMatch');
@@ -218,7 +218,7 @@ export async function run(
   // check if patterns exist in lockfile
   for (const pattern of patterns) {
     if (!lockfile.getLocked(pattern)) {
-      reportError('lockfileNotContainPatter', pattern);
+      reportError('lockfileNotContainPattern', pattern);
     }
   }
 
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -163,7 +163,7 @@ async function integrityHashCheck(
     reportError('noIntegirtyHashFile');
   }
   if (!match.integrityHashMatches) {
-    reportError('integrityHashesDontMatch', match.hashExpected, match.hashActual);
+    reportError('integrityHashesDontMatch');
   }
 
   if (errCount > 0) {
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -1,11 +1,12 @@
 /* @flow */
 
-import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import {MessageError} from '../../errors.js';
-import {Install} from './install.js';
+import {InstallationIntegrityChecker} from '../../integrity-checker.js';
 import Lockfile from '../../lockfile/wrapper.js';
+import type {Reporter} from '../../reporters/index.js';
 import * as fs from '../../util/fs.js';
+import {Install} from './install.js';
 
 const semver = require('semver');
 const path = require('path');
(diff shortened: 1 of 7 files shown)


src/cli/commands/check.js
```diff
@@ -142,30 +144,27 @@ async function integrityHashCheck(
     reporter.error(reporter.lang(msg, ...vars));
     errCount++;
   }
+  const integrityChecker = new InstallationIntegrityChecker(config);
 
   const lockfile = await Lockfile.fromDirectory(config.cwd);
   const install = new Install(flags, config, reporter, lockfile);
 
   // get patterns that are installed when running `yarn install`
+  // TODO measure slowdown here
```
bestander: This is a reminder for myself, I noticed check got slower in the last month.
Probably one of the following 2 commands to blame

src/integrity-checker.js
```diff
@@ -0,0 +1,133 @@
...
+    // ensure we only write to a registry folder that was used
+    for (const name of registryNames) {
+      const loc = path.join(this.config.cwd, this.config.registries[name].folder);
+      possibleFolders.push(loc);
+    }
+
+    // if we already have an integrity hash in one of these folders then use it's location otherwise use the
+    // first folder
+    const possibles = possibleFolders.map((folder): string => path.join(folder, constants.INTEGRITY_FILENAME));
+    let loc = possibles[0];
```
arcanis: Shouldn't it check to see if a directory exists before fallbacking to the first entry? For example, let's say Yarn supports Bower and that there's a `bower_modules` directory but no `node_modules` one. Shouldn't Yarn use the bower_directory rather than the non-existing node_modules? (on a similar note but possibly out of scope for this PR, shouldn't the integrity check file be different for each registry rather than global to all of them?)

src/cli/commands/check.js
```diff
@@ -142,30 +144,27 @@ async function integrityHashCheck(
...
+  const lockSource = lockStringify(lockfile.getLockfile(install.resolver.patterns));
 
-  // check if patterns exist in lockfile
-  for (const pattern of patterns) {
-    if (!lockfile.getLocked(pattern)) {
-      reportError('lockfileNotContainPatter', pattern);
-    }
+  const match = await integrityChecker.check(patterns, lockfile, lockSource, flags);
+  for (const pattern of match.missingPatterns) {
+    reportError('lockfileNotContainPatter', pattern);
```
arcanis: It seems there's a typo in the name of the error (but this typo also affects the lang file so it actually works)

src/cli/commands/check.js
```diff
@@ -142,30 +144,27 @@ async function integrityHashCheck(
...
-
-  const integrityLoc = await install.getIntegrityHashLocation();
-  if (integrityLoc && await fs.exists(integrityLoc)) {
-    const match = await install.matchesIntegrityHash(patterns);
-    if (match.matches === false) {
-      reportError('integrityHashesDontMatch', match.expected, match.actual);
-    }
-  } else {
+  if (match.integrityFileMissing) {
     reportError('noIntegirtyHashFile');
```
arcanis: Same here

bestander: Yeah, let's fix those here.
In the future it may be problematic when there are translations in the wild

src/integrity-checker.js
bestander: Check out the next few lines.
possibles[0] is taken by default but then it checks `fs.exists(possibleLoc)` and picks the first one that exists.

bestander: > on a similar note but possibly out of scope for this PR, shouldn't the integrity file be different for each registry rather than global to all of them?

We use one type of registry per project, unlikely there will be a bower_modules and node_modules in one folder, so it should be fine to have only one integrity there.

arcanis: It checks to see if the `path(folder, constants.INTEGRITY_FILENAME)` file exists, but not the `folder` itself. So for example, if there's a `bower_modules` without integrity file, the returned path will still be `node_modules/...`.

bestander: Still thinking it through.

I don't think we should choose location for .yarn-integrity based on a folder.

What if a project has checked-in bower_modules folder, then Yarn might pick it up as the location however user might want to install node_modules.

However, we should pick the location based on what type of registry are there in the project.
Right now we support only node_modules and yarn/npm registry so this any change I make here does not matter because we have a set of 1.

bestander:
@arcanis, I refactored integrity hash location a bit.
The idea is to pass the subset of used registries on save (and pick first if none have an integrity file).
As commented above I think we should not rely on a folder location

src/integrity-checker.js
```diff
@@ -33,30 +39,37 @@ export default class InstallationIntegrityChecker {
    * write a new one.
    */
 
-  async _getIntegrityHashLocation(): Promise<string> {
+  async _getIntegrityHashLocation(usedRegistries?: Set<RegistryNames>): Promise<IntegrityHashLocation> {
     // build up possible folders
+    let registries = registryNames;
+    if (usedRegistries && usedRegistries.size > 0) {
+      registries = usedRegistries;
+    }
```
arcanis: Even if it's currently the case here, you might want to put a comment to explain that passing falsy values for `usedRegistries` is only acceptable when one only needs to read the file but not write to it - otherwise multiple integrity files might get generated!

src/integrity-checker.js
```diff
@@ -33,30 +39,37 @@ export default class InstallationIntegrityChecker {
...
       if (await fs.exists(possibleLoc)) {
         loc = possibleLoc;
         break;
       }
     }
-    return loc;
+    return {
+      locationPath: loc || possibles[0],
+      exists: !!loc,
+    };
```
arcanis: Not fond of the location getter that also check if a file exists, but it's a private method, so no big deal.

src/integrity-checker.js
```diff
@@ -33,30 +39,37 @@ export default class InstallationIntegrityChecker {
...
     // first folder
     const possibles = possibleFolders.map((folder): string => path.join(folder, constants.INTEGRITY_FILENAME));
-    let loc = possibles[0];
+    let loc;
     for (const possibleLoc of possibles) {
       if (await fs.exists(possibleLoc)) {
         loc = possibleLoc;
         break;
       }
     }
```
arcanis: There's still the behaviour where the first entry will always be returned if there is no integrity file, even if a more "suitable" already directory exists, right? Do you want to keep it that way?

src/integrity-checker.js
bestander: I'll add a comment.
But it won't generate multiple integrity files. (see below)

src/integrity-checker.js
bestander: This script will return the first folder that has .yarn-integrity file.
If none found it will pick the first from the list of available.

Right now it is hard to decide whether existence of `bower_modules` is the right signal for Yarn to save .yarn-integrity there.
Let's get back to this subject when Yarn has more than one folder it can serve, currently it is node_modules only

src/integrity-checker.js
bestander: Removed my previous comment.
Actually it is fine to pass falsy usedRegistries when generating the integrity file, e.g. running `yarn install` for empty package.json.
Also Yarn won't generate multiple integrity files because this method checks whether one exists already and reuses that location

PR #2820, opened 2017-03-02
Improve discoverability of yarnpkg/rfcs

timoxley:
**Summary**

Currently the rfcs repo is not easily discoverable, the only place it appears to be mentioned is in the new issue template. Linking to the repo in more places improves discoverability of yarnpkg/rfcs.


timoxley:
This should also appear in the Contributing section in the Readme.

bestander:
Nice!

PR #2826, opened 2017-03-02
switched to tar-fs from node-tar to unpack tars

### LINKED ISSUE #961: Yarn should normalize permissions when untarring packages.
This is a bug report.

Yarn keeps the permissions of the archives is untars intact, while npm normalizes them. This means some npm packages cannot be installed using yarn.

Create a folder with the following package.json:

```
{
  "name": "test-case",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@types/history": "2.0.38"
  }
}
```

Run `yarn` using a version of yarn with support for scoped modules (e.g., `master`).

**Note:** This is **not** an issue with scoped modules in general. Other scoped modules work on `master`.

The operation should succeed (as it does with `npm install`), and the the permissions should be as follows:

```
total 40
-rw-r--r--   1 mcbr   611B Oct  3 08:23 README.md
-rw-r--r--   1 mcbr   5.7K Oct  3 08:23 index.d.ts
drwxr-xr-x  11 mcbr   374B Oct 12 22:07 lib
-rw-r--r--   1 mcbr   2.1K Oct 12 22:07 package.json
-rw-r--r--   1 mcbr   1.1K Oct  3 08:23 types-metadata.json

./node_modules/@types/history/lib:
total 72
-rw-r--r--  1 mcbr   624B Oct  3 08:23 DOMUtils.d.ts
-rw-r--r--  1 mcbr   143B Oct  3 08:23 actions.d.ts
-rw-r--r--  1 mcbr   133B Oct  3 08:23 createBrowserHistory.d.ts
-rw-r--r--  1 mcbr   130B Oct  3 08:23 createHashHistory.d.ts
-rw-r--r--  1 mcbr   203B Oct  3 08:23 createLocation.d.ts
-rw-r--r--  1 mcbr   134B Oct  3 08:23 createMemoryHistory.d.ts
-rw-r--r--  1 mcbr   135B Oct  3 08:23 useBasename.d.ts
-rw-r--r--  1 mcbr   182B Oct  3 08:23 useBeforeUnload.d.ts
-rw-r--r--  1 mcbr   165B Oct  3 08:23 useQueries.d.ts
```

I don't think it's required that the date for "lib" be set to the current date. `npm install` just does this for some reason.

Instead, the operation does not succeed:

```
yarn install v0.15.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
error https://registry.npmjs.org/@types/history/-/history-2.0.38.tgz: EACCES: permission denied, open '/Users/mcbr/.yarn-cache/npm-@types/history-2.0.38/lib/DOMUtils.d.ts'
    at Error (native)
```

Note: untarring `https://registry.npmjs.org/@types/history/-/history-2.0.38.tgz` results in:

```
total 40
-rw-r--r--   1 mcbr   611B Oct  3 08:23 README.md
-rw-r--r--   1 mcbr   5.7K Oct  3 08:23 index.d.ts
drw-r--r--  11 mcbr   374B Oct  3 08:23 lib
-rw-r--r--   1 mcbr   603B Oct  3 08:23 package.json
-rw-r--r--   1 mcbr   1.1K Oct  3 08:23 types-metadata.json

./history/lib:
```

Note that `lib` does not have executable ("cd") permissions. While we could ask DefinitelyTyped (`@types/history`) to fix their permissions, I'm reporting the issue here in case you with to improve compatibility with npm packages, especially in case there are other strange packages.

- Node v6.6.0
- Yarn `ee07644ed9080a22ca9742c9383100b8f03221a3` (`master`)
- macOS Sierra

bestander:
**Summary**

A mitigation for https://github.com/yarnpkg/yarn/issues/2629
Also fixes #961

Based on a few ad-hoc tests (//gitlab.com/linagora/petals-cockpit.git#075bac4c54fee466568c000c7ffe8025f593e212) tar-fs seems to make less IO thus is more stable and a bit faster.

I noticed that we never tested git repos that support archive.
Did they ever work?

----
Test data:

1. Using offline-mirror (no download), on my MBPro 13", clean cache and using `node-tar` to unzip files.
Concurrency 12 - fails
Concurrency 8 - 18 seconds
Concurrency 4 - 18 seconds
Concurrency 2 - 21 seconds

2. Using offline-mirror (no download), on my MBPro 13", clean cache and using `tar-fs` to unzip files.
Concurrency 12 - 15 seconds
Concurrency 8 - 15 seconds
Concurrency 4 - 17 seconds
Concurrency 2 - 18 seconds

3. Downloading packages from internet, on my MBPro 13", clean cache and using `tar-fs` to unzip files.
Concurrency 12 - failed once
Concurrency 8 - 21 seconds
Concurrency 4 - 23 seconds
Concurrency 2 - 34 seconds

**Test plan**
- All changes are covered with tests (verified every change)
- Could not properly test git archives (see https://github.com/yarnpkg/yarn/issues/2831) but did an ad-hoc test via `yarn add ssh://git@bitbucket.org/bestander/test-repo.git` and made sure that archive was accessed and installed.

### DIFF
--- a/__tests__/commands/pack.js
+++ b/__tests__/commands/pack.js
@@ -14,7 +14,7 @@ const os = require('os');
 const stream = require('stream');
 
 const zlib = require('zlib');
-const tar = require('tar');
+const tarFs = require('tar-fs');
 const fs2 = require('fs');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'pack');
@@ -83,8 +83,12 @@ export async function getFilesFromArchive(source, destination): Promise<Array<st
   const unzip = new Promise((resolve, reject) => {
     fs2.createReadStream(source)
       .pipe(new zlib.Gunzip())
-      .pipe(tar.Extract({path: destination, strip: 1}))
-      .on('end', resolve)
+      .pipe(tarFs.extract(destination, {
+        strip: 1,
+        dmode: parseInt(555, 8), // all dirs should be readable
+        fmode: parseInt(444, 8) // all files should be readable
+      }))
+      .on('finish', resolve)
       .on('error', reject);
   });
   await unzip;
--- a/__tests__/commands/pack.js
+++ b/__tests__/commands/pack.js
@@ -85,8 +85,8 @@ export async function getFilesFromArchive(source, destination): Promise<Array<st
       .pipe(new zlib.Gunzip())
       .pipe(tarFs.extract(destination, {
         strip: 1,
-        dmode: parseInt(555, 8), // all dirs should be readable
-        fmode: parseInt(444, 8), // all files should be readable
+        dmode: 0o555, // all dirs should be readable
+        fmode: 0o444, // all files should be readable
       }))
       .on('finish', resolve)
       .on('error', reject);
--- a/__tests__/commands/pack.js
+++ b/__tests__/commands/pack.js
@@ -86,7 +86,7 @@ export async function getFilesFromArchive(source, destination): Promise<Array<st
       .pipe(tarFs.extract(destination, {
         strip: 1,
         dmode: parseInt(555, 8), // all dirs should be readable
-        fmode: parseInt(444, 8) // all files should be readable
+        fmode: parseInt(444, 8), // all files should be readable
       }))
       .on('finish', resolve)
       .on('error', reject);
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -35,8 +35,8 @@ export default class GitFetcher extends BaseFetcher {
 
     return new Promise((resolve, reject) => {
       const untarStream = tarFs.extract(this.dest, {
-        dmode: parseInt(555, 8), // all dirs should be readable
-        fmode: parseInt(444, 8), // all files should be readable
+        dmode: 0o555, // all dirs should be readable
+        fmode: 0o444, // all files should be readable
       });
 
       const hashStream = new crypto.HashStream();
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -36,7 +36,7 @@ export default class GitFetcher extends BaseFetcher {
     return new Promise((resolve, reject) => {
       const untarStream = tarFs.extract(this.dest, {
         dmode: parseInt(555, 8), // all dirs should be readable
-        fmode: parseInt(444, 8) // all files should be readable
+        fmode: parseInt(444, 8), // all files should be readable
       });
 
       const hashStream = new crypto.HashStream();
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -7,7 +7,7 @@ import Git from '../util/git.js';
 import * as fsUtil from '../util/fs.js';
 import * as crypto from '../util/crypto.js';
 
-const tar = require('tar');
+const tarFs = require('tar-fs');
 const url = require('url');
 const path = require('path');
 const fs = require('fs');
@@ -34,7 +34,10 @@ export default class GitFetcher extends BaseFetcher {
     }
 
     return new Promise((resolve, reject) => {
-      const untarStream = tar.Extract({path: this.dest});
+      const untarStream = tarFs.extract(this.dest, {
+        dmode: parseInt(555, 8), // all dirs should be readable
+        fmode: parseInt(444, 8) // all files should be readable
+      });
 
       const hashStream = new crypto.HashStream();
 
@@ -42,7 +45,7 @@ export default class GitFetcher extends BaseFetcher {
       cachedStream
         .pipe(hashStream)
         .pipe(untarStream)
-        .on('end', () => {
+        .on('finish', () => {
           const expectHash = this.hash;
           const actualHash = hashStream.getHash();
           if (!expectHash || expectHash === actualHash) {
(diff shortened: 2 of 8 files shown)


bestander:
Don't merge, still need to remove node-tar from 2 more places.

bestander:
Also fixes #961

bestander:
@arcanis, could you review this please?

__tests__/commands/pack.js
```diff
@@ -83,8 +83,12 @@ export async function getFilesFromArchive(source, destination): Promise<Array<st
   const unzip = new Promise((resolve, reject) => {
     fs2.createReadStream(source)
       .pipe(new zlib.Gunzip())
-      .pipe(tar.Extract({path: destination, strip: 1}))
-      .on('end', resolve)
+      .pipe(tarFs.extract(destination, {
+        strip: 1,
+        dmode: parseInt(555, 8), // all dirs should be readable
+        fmode: parseInt(444, 8), // all files should be readable
```
arcanis: `0o555` and `0o444` ?

src/fetchers/git-fetcher.js
```diff
@@ -34,15 +34,18 @@ export default class GitFetcher extends BaseFetcher {
     }
 
     return new Promise((resolve, reject) => {
-      const untarStream = tar.Extract({path: this.dest});
+      const untarStream = tarFs.extract(this.dest, {
+        dmode: parseInt(555, 8), // all dirs should be readable
+        fmode: parseInt(444, 8), // all files should be readable
```
arcanis: Same

src/fetchers/tarball-fetcher.js
```diff
@@ -77,7 +77,11 @@ export default class TarballFetcher extends BaseFetcher {
   } {
     const validateStream = new crypto.HashStream();
     const extractorStream = new UnpackStream();
-    const untarStream = tar.Extract({path: this.dest, strip: 1});
+    const untarStream = tarFs.extract(this.dest, {
+      strip: 1,
+      dmode: parseInt(555, 8), // all dirs should be readable
+      fmode: parseInt(444, 8), // all files should be readable
```
arcanis: Same - shouldn't the unpack routine be splitted into an helper, since it used in a few places?

src/util/git.js
```diff
@@ -279,13 +287,21 @@ export default class Git {
...
-          parser.on('end', done);
-
-          parser.on('data', (entry: Buffer) => {
-            update(entry.toString());
+          parser.on('finish', done);
+
+          parser.on('entry', (header, stream, next) => {
+            let string = '';
+            stream.on('data', (buffer) => {
+              string += buffer.toString();
```
arcanis: iirc, this pattern might break when the buffer contains part of an unicode character but not the others. It might be better to concat the buffers together (possibly by storing them in a list and then concatenating everything at once with `Buffer.concat`) rather than concatenating strings.

src/util/git.js
```diff
@@ -205,9 +206,12 @@ export default class Git {
   async _cloneViaRemoteArchive(dest: string): Promise<void> {
     await child.spawn('git', ['archive', `--remote=${this.url}`, this.ref], {
       process(proc, update, reject, done) {
-        const extractor = tar.Extract({path: dest});
+        const extractor = tarFs.extract(dest, {
+          dmode: parseInt(555, 8), // all dirs should be readable
+          fmode: parseInt(444, 8), // all files should be readable
```
arcanis: Same

src/util/git.js
```diff
@@ -219,9 +223,13 @@ export default class Git {
     await child.spawn('git', ['archive', this.hash], {
       cwd: this.cwd,
       process(proc, resolve, reject, done) {
-        const extractor = tar.Extract({path: dest});
+        const extractor = tarFs.extract(dest, {
+          dmode: parseInt(555, 8), // all dirs should be readable
+          fmode: parseInt(444, 8), // all files should be readable
```
arcanis: [Ditto](http://cdn.bulbagarden.net/upload/3/36/132Ditto.png)

src/fetchers/tarball-fetcher.js
bestander: Sometimes we need to strip first folder, sometimes not.
And moving dmode,fbmode into a separate function looks like an overkill

src/util/git.js
bestander: Oh, smart!
I'll do that

__tests__/commands/pack.js
bestander: 👍

src/fetchers/tarball-fetcher.js
arcanis: 👍

bestander:
ping @arcanis, ready for another round

src/util/git.js
```diff
@@ -293,12 +295,15 @@ export default class Git {
...
+            let fileContent = '';
+
             stream.on('data', (buffer) => {
-              string += buffer.toString();
+              fileContent += decoder.write(buffer);
             });
             stream.on('end', () => {
-              update(string);
+              fileContent += decoder.end();
+              update(fileContent);
```
arcanis: Since it isn't really used anymore, maybe `fileContent` should be removed completely, using `update(decoded.end())` instead

__tests__/commands/pack.js
arcanis: Just noticed - you forgot this instance :)

src/util/git.js
bestander: decoded.end() won't return a full string.
It would just return remaining bytes if there were any.
So I needed to accumulate them in a string

arcanis: Oh I see, my bad ! I initially thought the decoder was holding the whole buffer 👍

victornoel:
@bestander it could make sense to also use `gunzip-maybe` (like `tar-fs`, it is by @mafintosh) instead of reimplementing it in `utils/stream`.

Issue #2109, opened 2016-12-02
When using an offline cache, yarn upgrade adds the new package to the cache but does not remove the old one


rmacklin:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

If you're using an in-repo offline cache (https://yarnpkg.com/blog/2016/11/24/offline-mirror), when you run `yarn upgrade <package>`, the new version of `<package>` will be added to the offline cache, but the old version won't be removed (even if it is not used by any other packages). It should be removed, so that manual deletion is not required to prevent the cache from growing to contain many unused packages over time.

**If the current behavior is a bug, please provide the steps to reproduce.**

Reproduction repository: https://github.com/rmacklin/yarn-issue-2109-repro

Steps to reproduce:

1. `git clone git@github.com:rmacklin/yarn-issue-2109-repro.git`
2. `cd yarn-issue-2109-repro`
3. `node yarn-0.17.10.js upgrade left-pad`
4. Observe:

This added a new, untracked tarball for the new version of `left-pad` in the
cache:

```
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   package.json
	modified:   yarn.lock

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	npm-packages-offline-cache/left-pad-1.1.3.tgz

no changes added to commit (use "git add" and/or "git commit -a")
```

However, the old `npm-packages-offline-cache/left-pad-1.1.2.tgz` tarball was _not_ removed even though it's no longer used. This means that it has to be manually deleted (an easy step to forget) in order to prevent the offline cache from growing to contain many unused packages over time.

**What is the expected behavior?**

The old, unused tarball should be removed, so that people don't have to remember to remove it from the repo. It would be fine if this behavior is behind an option or a .yarnrc setting.

**Please mention your node.js, yarn and operating system version.**

Node v6.2.0, yarn v0.17.10, Mac OSX 10.9.5

bestander:
This was intentional for a monorepo when the offline cache is shared between multiple package.json/yarn.lock files.
I understand your concern, yarn should probably have a command or setting that would clean up orphan files from monorepo.
PR and discussion is welcome.

In the meantime cleanup can be done with a very basic script that parses yarn.lock file.

talaviss:
[...]

PR #2836, opened 2017-03-05
Add optional offline mirror pruning

dguo:
**Summary**

This change is an implementation of yarnpkg/rfcs#49, which stems from #2109.

**Test plan**

I tested the feature locally with two scenarios where `.yarnrc` has `yarn-offline-mirror-pruning` set to `true`:

* Upgrading a package:
    * `$ yarn add is-positive-integer@1.0.0`
    * Check that the offline mirror has the `is-positive-integer` tarball, along with all its dependencies
    * `$ yarn upgrade is-positive-integer@1.1.1`
    * Check that the offline mirror only has the new version of `is-positive-integer`, which doesn't have any dependencies
* Removing a package:
    * `$ yarn add left-pad right-pad`
    * Check that the offline mirror has both the `left-pad` and `right-pad` tarballs
    * `$ yarn remove left-pad`
    * Check that the offline mirror only has `right-pad`

I also added two Jest test cases, one for upgrade and one for remove. They verify that sub-dependencies are also pruned.

For documentation, I'd be happy to submit a pull request for that as well, but I don't see any for the offline mirror feature on the main Yarn website. I've just been referencing [this blog post](https://yarnpkg.com/blog/2016/11/24/offline-mirror/).


src/cli/commands/install.js
```diff
@@ -572,14 +572,44 @@ export class Install {
...
+    const mirror = this.config.getOfflineMirrorPath();
+    if (!mirror) {
+      return;
+    }
+
+    const requiredTarballs = new Set();
+    for (const dependency in lockfile) {
+      const resolved = lockfile[dependency].resolved;
+      if (resolved) {
+        requiredTarballs.add(resolved.split('#')[0]);
```
dguo: I'm not sure if it is okay to depend on the format of `resolved` like this.

bestander: The tests would fail if there are breaking changes.
I suppose it should be fine.

bestander:
Looks nice, how about a test for add command as well?

bestander:
Tests seem to fail for some other reason

dguo:
Yep, I considered adding a test for add as well, but I couldn't come up with a situation where pruning would apply. Maybe there's some uncommon semver combination that I haven't thought of? If not, should I just copy over the upgrade test fixture and use that? So it would simulate:
`$ yarn add dep-a@1.0.0`
`$ yarn add dep-a@1.1.0`

The tests fail for me locally as well, but the last I tried, the same tests failed on the master branch. I attributed the failures to flakiness, but would you like me to dig into them?

bestander:
I think you are right, `add` that would cause package remove is only when a package already exists and gets updated.
RE tests, seems like it is flakiness

bestander:
@dguo, could you rebase please?
And also could you send a PR to the docs website?
Otherwise no one will ever know about this feature.

dguo:
Sure, I can rebase tonight when I'm at my computer again.

I'm happy to send a PR for docs, but I didn't see an existing section on
the website for the offline mirror feature. Could you point me to the right
place?

On Mar 7, 2017 10:16 AM, "Konstantin Raev" <notifications@github.com> wrote:

@dguo <https://github.com/dguo>, could you rebase please?
And also could you send a PR to the docs website?
Otherwise no one will ever know about this feature.

—

<https://github.com/yarnpkg/yarn/pull/2836#issuecomment-284750142>, or mute
the thread

bestander:
I think we only have a bog post
https://yarnpkg.com/blog/2016/11/24/offline-mirror/

Would you create a new section in the Docs please?
We'll convert the blog post into the doc after that.

dguo:
Ok, it's rebased, the tests passed, and I submitted yarnpkg/website#409 for the docs.

bestander:
Great job on pushing through the whole feature from RFC to great implementation with tests and docs, @dguo!

dguo:
Thank you!

rmacklin:
Wow, awesome! I'm participating in a hackathon today and this was actually the first thing I was gonna work on (I submitted the original issue), but I'm happy to see it's already done. Thanks @dguo!

Also, re: the tests and not having a test for `add`, I think there's an untested case we may want to cover. Suppose a team has been using yarn with an offline mirror for a while prior to the addition of this feature. Once they enable pruning, unless they had been manually pruning before, there will still be unused tarballs in the offline mirror. At that point, they should run `yarn` with no arguments (or equivalently `yarn install`) to prune all those unused tarballs, so that future add/remove/upgrade commands don't get mixed in with unrelated pruning.

So, I think it'd be good to add a test case for this scenario: start with a lockfile and a mirror that has more tarballs than what's in the lockfile, run `yarn` with no arguments (i.e. `yarn install`), verify that the unused tarballs were pruned. I've done that here: https://github.com/yarnpkg/yarn/pull/2880

We _could_ also add a similar test for running `add`, `remove`, and `upgrade` when the offline mirror already contains unused tarballs unrelated to the package that was added/removed/upgraded (we'd expect them to be pruned), though that may be overkill (if we don't think it's overkill, I can add those tests to the PR).

rmacklin:
Is there any chance this could make it into a (pre-)release soon?

bestander:
Yeah, we'll cut the branch mid next week

rmacklin:
Awesome, and thanks for the speedy reply!

PR #1000, opened 2016-10-13
avoid call to stdin on Windows

MiguelAlho:
**Summary**

As seen in other repos (namely https://github.com/evantahler/actionhero/pull/360) IISNode has no notion of stdin, so accessing ´stdin` in base-reporter throws an exception.

While developing an extension for Visual Studio's Task Runner Explorer component, that would allow devs to call yarn commands (and maybe run scripts) from this window. This is similar to existing extensions for other task-enabled engines like gulp, grunt, and npm. These tasks can also be binded to some of the IDEs events such as "project open". This is great for my current use case - when a yarn enabled project is open, run `yarn install` automatically.

Unfortunately, when executing yarn through the extension (which is really calling `cmd.exe /C yarn install`) the following error appears:

```
C:\repo> cmd.exe /c yarn install
C:\Users\AlhoM\AppData\Roaming\npm\node_modules\yarn\bin\yarn.js:48
      throw err;
      ^
Error: Implement me. Unknown stdin file type!
    at process.getStdin [as stdin] (internal/process/stdio.js:82:15)
    at ConsoleReporter.BaseReporter (C:\Users\AlhoM\AppData\Roaming\npm\node_modules\yarn\lib\reporters\base-reporter.js:60:37)
    at ConsoleReporter (C:\Users\AlhoM\AppData\Roaming\npm\node_modules\yarn\lib\reporters\console\console-reporter.js:58:5)
    at Object.<anonymous> (C:\Users\AlhoM\AppData\Roaming\npm\node_modules\yarn\lib\cli\index.js:192:18)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
Process terminated with code 1.
```

Hacking around a bit and using the info in the link above, the repose to the command returns the desired output:

```
 C:\repo> cmd.exe /c yarn install
yarn install v0.15.1
warning app@1.0.0: No license field
success Already up-to-date.
Done in 0.13s.
Process terminated with code 0..
```

**Test plan**

A clean clone did not produce a valid code run (I checked on discord and the info available was that some tests where not running.

With visual studio installed, the [extension I created](https://visualstudiogallery.msdn.microsoft.com/c9f7667c-ac1f-4be8-82c6-9a4e34fef47b) can be installed and used to test the change.

[...]

PR #2837, opened 2017-03-05
gracefully handle process.stdin exceptions

scottaddie:
**Summary**
Several users of Visual Studio 2015 & 2017's NPM Task Runner extension have reported problems on Windows 7 with regard to running Yarn commands inside the IDE's Task Runner Explorer window. Interestingly enough, this doesn't appear to be an issue on Windows 10.

A fix for this problem was attempted [here](https://github.com/yarnpkg/yarn/pull/1000). The Angular CLI project also encountered the same problem, and it was fixed [here](https://github.com/angular/angular-cli/pull/4871).

This PR takes a very similar approach as that taken by the Angular CLI team. The PR also doesn't specifically test for the "win32" platform, as who knows what other platforms are really impacted by this.

**Test plan**
I ran the following commands to ensure that my changes didn't introduce any problems:
`yarn run build`
`yarn run lint`
`yarn run test`

### DIFF
--- a/src/reporters/base-reporter.js
+++ b/src/reporters/base-reporter.js
@@ -18,6 +18,7 @@ import * as languages from './lang/index.js';
 import isCI from 'is-ci';
 
 const util = require('util');
+const EventEmitter = require('events').EventEmitter;
 
 type Language = $Keys<typeof languages>;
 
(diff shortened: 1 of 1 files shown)


src/reporters/base-reporter.js
```diff
@@ -113,6 +114,21 @@ export default class BaseReporter {
   _verbose(msg: string) {}
   _verboseInspect(val: any) {}
 
+  _getStandardInput(): Stdin {
+    let standardInput;
+
+    try {
+      standardInput = process.stdin;
+    } catch (e) {
+      delete process.stdin;
```
Daniel15: Perhaps this should `console.warn(e.message)`?

Daniel15:
Looks good to me, thanks! If this fix is okay for Angular-cli, it's probably fine for Yarn too 😛

PR #2838, opened 2017-03-06
Add console.warn to process.stdin catch block

scottaddie:
**Summary**
Per @Daniel15's suggestion, this PR adds a `console.warn` statement to the `catch` block introduced in PR https://github.com/yarnpkg/yarn/pull/2837.

**Test plan**
I ran the following commands to ensure that my changes didn't introduce any problems:
`yarn run build`
`yarn run lint`
`yarn run test`

### DIFF
--- a/src/reporters/base-reporter.js
+++ b/src/reporters/base-reporter.js
@@ -117,9 +117,11 @@ export default class BaseReporter {
   _getStandardInput(): Stdin {
     let standardInput;
 
+    // Accessing stdin in a win32 headless process (e.g., Visual Studio) may throw an exception.
     try {
       standardInput = process.stdin;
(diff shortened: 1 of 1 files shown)


src/reporters/base-reporter.js
```diff
@@ -120,6 +120,7 @@ export default class BaseReporter {
     try {
       standardInput = process.stdin;
```
bestander: could you please add a text comment why we do the trick?
Going forward this file may be refactored multiple times and we would loose the original reason for try/catch.

scottaddie: No problem. This is done.

Daniel15:
Thanks!

PR #2842, opened 2017-03-06
Esy better escaping [DO NOT MERGE YET]

jordwalke:
Summary:
    This requires bumping the esy version - and I've also created a
    convention of encoding the esy version in the store.

    Really, we want two different versions - the esy version, and then the
    store compatibility version. Bumping major esy versions needn't
    necessarily bump the store encoding version, but for simplicity sake,
    let's consider them the same version.

    Test Plan:


jordwalke:
Sent PR to wrong remote! Sorry for the trouble everyone.

PR #2844, opened 2017-03-06
Log the binFolder result from the `bin` command via the reporter.

GijsWeterings:
**Summary**

This PR redirects the output for the `bin` command through the configured reporter, instead of it being hardcoded to `console.log`. This way, in the future the bin command could be logged in different ways besides just printing.

The default behaviour does not change, as the default reporter is the `ConsoleReporter`, which again writes to `stdout`

**Test plan**

1. run `yarn bin`.
2. verify it outputs to console.
3. (optional) ensure output goes trough the `consoleReporter._log` function

### DIFF
--- a/src/cli/commands/bin.js
+++ b/src/cli/commands/bin.js
@@ -19,6 +19,6 @@ export function run(
     config.registries[RegistryYarn.registry].folder,
     '.bin',
   );
-  console.log(binFolder);
+  reporter.log(binFolder);
   return Promise.resolve();
 }


bestander:
There was a reason for this https://github.com/yarnpkg/yarn/pull/1980#discussion_r89621763

bestander:
I think we need to do something a bit smarter here, see reasons https://github.com/yarnpkg/yarn/pull/1980#discussion_r89621763.

I'll close for now, please reopen once there is handling for the linked case

PR #2870, opened 2017-03-08
fix the location for yarn-error.log and add a test case

### LINKED ISSUE #2719: Put error logs in cache folder

feature

Yarn will store `yarn-error.log` in the current directory.

In [npm v4.2.0](https://github.com/npm/npm/releases/tag/v4.2.0), `npm-debug.log` is no longer placed in the current directory, but in the `(npm config get cache)/_logs` directory. I suggest yarn do something similar with the error logs.

clanghout:
**Summary**
In order to change the folder where `yarn-error.log` is placed, we changed the path from `config.cwd` to `config.cacheFolder`. In the process of testing we also discovered a bug with specifying a cache folder from the command line.
The test executes a command that will error a `yarn-error.log` and verifies that this log is created in the specified cache folder.
Fixes #2719 

**Test plan**
Run `yarn test`

or by hand: run the command `yarn tag rm non-existing-pkg non-existing-tag` and press enter, then verify that the reporter put the error log in the cache folder.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -345,7 +345,7 @@ function onUnexpectedError(err: Error) {
 }
 
 function writeErrorReport(log) : ?string {
-  const errorReportLoc = path.join(config.cwd, 'yarn-error.log');
+  const errorReportLoc = path.join(config.cacheFolder, 'yarn-error.log');
 
   try {
     fs.writeFileSync(errorReportLoc, log.join('\n\n') + '\n');
@@ -362,7 +362,7 @@ config.init({
   binLinks: commander.binLinks,
   modulesFolder: commander.modulesFolder,
   globalFolder: commander.globalFolder,
-  cacheRootFolder: commander.cacheFolder,
+  cacheFolder: commander.cacheFolder,
   preferOffline: commander.preferOffline,
   captureHar: commander.har,
   ignorePlatform: commander.ignorePlatform,
--- a/__tests__/cli/index.js
+++ b/__tests__/cli/index.js
@@ -19,10 +19,8 @@ async function setupWorkingDir(fixture: string): Promise<string> {
 }
 
 function execCommand(workingDir: string, cacheDir: string): Promise<string> {
-  const relativeBin = path.relative(workingDir, yarnBin);
-
   return new Promise((resolve, reject) => {
-    exec(`node "${relativeBin}" tag rm non-existing-pkg non-existing-tag --cache-folder ${cacheDir} | cat`,
+    exec(`node "${yarnBin}" tag rm non-existing-pkg non-existing-tag --cache-folder ${cacheDir} | cat`,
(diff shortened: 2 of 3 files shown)


TimvdLippe:
It seems that CircleCI is timing out on `__tests__/commands/install/integration.js (357.669s)` and the same for AppVeyor: `__tests__\commands\install\integration.js (275.532s)`

On Travis all succeeded, but the 1 job on OS X is failing. That might be a mac related issue.

TimvdLippe:
On all CI's, the integration test is timing out. However, our own test case is not the culprit any longer. The last commit fixes the path issue on OS X.

src/cli/index.js
```diff
@@ -362,7 +362,7 @@ config.init({
   binLinks: commander.binLinks,
   modulesFolder: commander.modulesFolder,
   globalFolder: commander.globalFolder,
-  cacheRootFolder: commander.cacheFolder,
+  cacheFolder: commander.cacheFolder,
```
bestander: This is done on purpose afaik.
@arcanis we probably need to add a comment here.

clanghout: This probably does need a change to either `_cacheRootFolder` or `cacheFolder` since `cacheRootFolder` does not exist in the configOptions export type. The reason for this change was the cache folder location not changing when setting it with the `--cache-folder` command (see #2881).

clanghout:
Since the release of npm v4.2.0 the error logs are placed in the cache folder instead of the CWD:  
> This debug log would get cleared out as soon as you ran npm again. This was a bit annoying because 1) you would get a random file in your `git status` that you might accidentally commit, and 2) if you hit a hard-to-reproduce bug and instinctually tried again, you would no longer have access to the repro `npm-debug.log`.

> he cache will now hold a (configurable) number of `npm-debug.log` files, which you can access in the future. 

source: [npm release v4.2.0](https://github.com/npm/npm/releases/tag/v4.2.0). 
What differs however (and I have not implemented) is saving multiple log files inside the cache folder.

src/cli/index.js
arcanis: Yep the specific line where this is used is here: https://github.com/yarnpkg/yarn/blob/master/src/config.js#L241

In order to prevent BC break we decided to keep using `cacheFolder` in the public interface, but it seems I forgot one. Good catch @clanghout!

TimvdLippe:
Using `process.exec` does not generate any coverage. We might be able to directly `require` `cli/index.js` and then pass the arguments.

bestander:
Still it seems weird to have the error in the cache folder.
I understand the risk is people may commit it accidentally.

bestander:
I guess doing same as npm with `<cache>/_logs` and adding timestamp to the filename could be an option.
And also Yarn should log the error file location if it does not.

clanghout:
Different behavior of the cache folder is implemented in #3367 and the found bug is merged in #3033

PR #2880, opened 2017-03-09
Add test for pruning on install when mirror pruning has been enabled

rmacklin:
**Summary**

This PR extends the test cases for the offline mirror pruning that was added in https://github.com/yarnpkg/yarn/pull/2836.

The intention is to have test coverage for the following scenario:
> Suppose a team has been using yarn with an offline mirror for a while prior to the pruning feature. Once they enable pruning, unless they had been manually pruning before, there will still be unused tarballs in the offline mirror. At that point, they should run `yarn` with no arguments (or equivalently `yarn install`) to prune all those unused tarballs, so that future add/remove/upgrade commands don't get mixed in with unrelated pruning.

Essentially, this test verifies that running `yarn` in that scenario will indeed perform the proper pruning.

**Test plan**

I verified that this test passes as written, and (correctly) fails if we change the fixture by removing `yarn-offline-mirror-pruning true` from the `.yarnrc`. Since this PR is just adding a test, I think that's all that's needed.


bestander:
Nice, great job on keeping up the test coverage!

PR #2810, opened 2017-02-28
Use a class for help command and do some fixes to tests

voxsim:
This is my first tiny step towards a Base Command class to clash the number of ifs in src/cli/index.js. The way is quite long :D

**Summary**
For now I fixed some tests and write the help command as es6 class.

**Test plan**
I fixed some tests before touching some code; please see the commits.
Added tests for cache command.

@bestander I hope you like what I do and I have some questions for you:
1. Do you think is useful add an acceptance/feature test per command? It maybe be overkill but the code is too much entagled and if you agree we can remove in a second moment after the story ends if they are useless.
2. I think I will open an issue and maybe I will do a checklist because it is quite long and maybe it is useful some feedbacks/help, what do you think?

Right I am working to have all classes with three static functions (run, hasWrapper, setFlags), and when I will be ready everything will be switched to instance function. I will open more pull request to be careful ;)

PR #2887, opened 2017-03-10
Reduce complexity of src/cli/index.js

voxsim:
After working some time on https://github.com/yarnpkg/yarn/pull/2810 I thought that maybe I was trying to reduce the complexity of src/cli/index.js from the wrong point of view.

In this pull request I:
- Cherry picked some commits from https://github.com/yarnpkg/yarn/pull/2810
- Reduce a lot of if statements
- Move all the logic of help in help command

Test plan, add test:
- for not camelised command
- for not alphabetic command
- for not unknown command
- extract npm_config_argv_env_vars correctly if you use 'yarn run test' instead of 'yarn test'
- display the documentation link correctly if we use an alias
- display the documentation link correctly if we use a normal command
- do not display the documentation link correctly if we use have an unknown command

Everything else seems tested properly and this change is less invasive than what i do in https://github.com/yarnpkg/yarn/pull/2810.

@bestander let me know if you like and sorry if I spam you a lot :P 
Feedbacks from everyone are always welcome :) 

Questions:
- Why we camelcase the command name? And after digging in the function we return null if it has an uppercase letter, otherwise we camelcase the command name. I still don't get what it is the use case O.o

Simon

PS I left some other useful commits from https://github.com/yarnpkg/yarn/pull/2810, maybe I will add to this pull request or open others after, in detail: 
- https://github.com/yarnpkg/yarn/pull/2810/commits/13dd05eb0f7be2db668cdf5098d057c8ea06299a
- https://github.com/yarnpkg/yarn/pull/2810/commits/2ae055a9114bbbb9e55ded48984105565e60fb21

### DIFF
--- a/src/cli/commands/help.js
+++ b/src/cli/commands/help.js
@@ -4,7 +4,7 @@ import * as commands from './index.js';
 import * as constants from '../../constants.js';
 import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
-import {sortAlpha, hyphenate} from '../../util/misc.js';
+import {sortAlpha, hyphenate, camelCase} from '../../util/misc.js';
 const chalk = require('chalk');
 
 export function run(
@@ -17,24 +17,47 @@ export function run(
   const getDocsInfo = (name) => 'Visit ' + chalk.bold(getDocsLink(name)) + ' for documentation about this command.';
 
   if (args.length) {
-    const helpCommand = hyphenate(args[0]);
-    if (commands[helpCommand]) {
-      commander.on('--help', () => console.log('  ' + getDocsInfo(helpCommand) + '\n'));
-    }
-  } else {
-    commander.on('--help', () => {
-      console.log('  Commands:\n');
-      for (const name of Object.keys(commands).sort(sortAlpha)) {
-        if (commands[name].useless) {
-          continue;
+    const commandName = camelCase(args.shift());
+
+    if (commandName) {
+      const command = commands[commandName];
+
+      if (command) {
+        if (typeof command.setFlags === 'function') {
+          command.setFlags(commander);
         }
 
-        console.log(`    - ${hyphenate(name)}`);
+        const examples: Array<string> = (command && command.examples) || [];
+        if (examples.length) {
+          commander.on('--help', () => {
+            console.log('  Examples:\n');
+            for (const example of examples) {
+              console.log(`    $ yarn ${example}`);
+            }
+            console.log();
+          });
+        }
+        commander.on('--help', () => console.log('  ' + getDocsInfo(commandName) + '\n'));
+
+        commander.help();
+        return Promise.resolve();
       }
-      console.log('\n  Run `' + chalk.bold('yarn help COMMAND') + '` for more information on specific commands.');
-      console.log('  Visit ' + chalk.bold(getDocsLink()) + ' to learn more about Yarn.\n');
-    });
+    }
   }
+
+  commander.on('--help', () => {
+    console.log('  Commands:\n');
+    for (const name of Object.keys(commands).sort(sortAlpha)) {
+      if (commands[name].useless) {
+        continue;
+      }
+
+      console.log(`    - ${hyphenate(name)}`);
+    }
+    console.log('\n  Run `' + chalk.bold('yarn help COMMAND') + '` for more information on specific commands.');
+    console.log('  Visit ' + chalk.bold(getDocsLink()) + ' to learn more about Yarn.\n');
+  });
+
   commander.help();
   return Promise.resolve();
 }
--- a/src/cli/commands/help.js
+++ b/src/cli/commands/help.js
@@ -7,6 +7,10 @@ import type Config from '../../config.js';
 import {sortAlpha, hyphenate, camelCase} from '../../util/misc.js';
 const chalk = require('chalk');
 
+export function hasWrapper(): boolean {
+  return false;
+}
+
 export function run(
   config: Config,
   reporter: Reporter,
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -117,13 +117,11 @@ if (args.indexOf('--help') >= 0 || args.indexOf('-h') >= 0) {
 }
 
 // if no args or command name looks like a flag then default to `install`
-if (commandName && commandName[0] === '-') {
+if (commandName[0] === '-') {
   args.unshift(commandName);
   commandName = 'install';
 }
 
-//
-invariant(commandName, 'Missing command name');
 const camelised = camelCase(commandName);
 if (camelised) {
   command = commands[camelised];
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -24,19 +24,12 @@ const pkg = require('../../package.json');
 
 loudRejection();
 
-//
 const startArgs = process.argv.slice(0, 2);
-let args = process.argv.slice(2);
 
 // ignore all arguments after a --
-let endArgs = [];
-for (let i = 0; i < args.length; i++) {
-  const arg = args[i];
-  if (arg === '--') {
-    endArgs = args.slice(i + 1);
-    args = args.slice(0, i);
-  }
-}
+const doubleDashIndex = process.argv.findIndex((element) => element === '--');
+const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
+const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
 // NOTE: Pending resolution of https://github.com/tj/commander.js/issues/346
 // Remove this (and subsequent use in the logic below) after bug is resolved and issue is closed
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -40,12 +40,6 @@ const doubleDashIndex = process.argv.findIndex((element) => element === '--');
 const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
 const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
-// NOTE: Pending resolution of https://github.com/tj/commander.js/issues/346
-// Remove this (and subsequent use in the logic below) after bug is resolved and issue is closed
-const ARGS_THAT_SHARE_NAMES_WITH_OPTIONS = [
-  'version',
-];
-
 // set global options
 commander.version(pkg.version);
 commander.usage('[command] [flags]');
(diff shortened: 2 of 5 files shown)


bestander:
I'll get back to this in a couple of days, thanks for waiting

voxsim:
I added some comments in the code but i will wait your feedback before pushing anything else ;)

voxsim:
I opened other two pull requests for what can be saved from https://github.com/yarnpkg/yarn/pull/2810 ;)

voxsim:
@bestander  do you have any news on this? :p

bestander:
Sorry, Simon, it is definitely on my list.

I'll try to have a look ASAP.

voxsim:
Hey @bestander , don't worry ;) and thanks for your work!

voxsim:
yeah sure, I need some support from @arcanis for https://github.com/yarnpkg/yarn/commit/a5a5915965c94de116a30b64e1863b2ddd1e3768. Could you please write me an use case for `args = [commandName].concat(getRcArgs(commandName), args);`? So I can add a test and see that we won't have any regressions.

@bestander could you please answer about this queston?
`Why we camelcase the command name? And after digging in the function we return null if it has an uppercase letter, otherwise we camelcase the command name. I still don't get what it is the use case O.o`

I hope to mistype anything, but sorry I am writing from my phone :p

voxsim:
ok maybe I get it, sorry @arcanis for bother you ;)

bestander:
> Why we camelcase the command name? And after digging in the function we return null if it has an uppercase letter, otherwise we camelcase the command name. I still don't get what it is the use case O.o

Just a guess - so that commands could be referred a object properties when needed
like commands.upgradeInteractive.
Although I don't have much background on this issue

voxsim:
mmh if we do that right now It will not work because it has an uppercase letter. If we don't find any use case maybe we should delete that code and let's wait for  complains :p

arcanis:
@voxsim you may have already found it, but the code you quoted is related to [this PR](https://github.com/yarnpkg/yarn/pull/3033), and should be covered by [these tests](https://github.com/yarnpkg/yarn/blob/master/__tests__/lifecycle-scripts.js#L33-L57) 😃 

(disregard the "test disabled for now" comment, I've just enabled it but forgot to remove the comment)

arcanis:
Btw, I think there's two reasons the command name is camelCased:

- It works whether you write `yarn foo-bar` or `yarn fooBar` (isn't really useful, but still)
- All the commands are loaded via `import * as commands from './commands'`, which means that any command name *needs* to be valid a Javascript identifier, otherwise the symbol couldn't be exported.

voxsim:
@arcanis i tried both and they work as expected without camelCase. I always advocate "no code over code that i don't understand", but i don't want to be pushy, I'll let decide you, I understand that some time it's better to be conservative ;)

Btw I finished I should push in a bunch of minutes

voxsim:
Ok i finished to rebase everything ;) let me know! For what i see the tests failed for reasons not related to my pull request

voxsim:
I was wrong and after some digging we use the camelCase function only for 'generate-lock-entry' and 'upgrade-interactive'. Maybe we can achieve the same behaviour without camelCase every command. I will prepare another pull request on this part.

src/cli/index.js
```diff
@@ -98,90 +94,55 @@ commander.option('--network-concurrency <number>', 'maximum number of concurrent
...
 }
 
-//
-if (commandName === 'help' && args.length) {
-  commandName = camelCase(args.shift());
-  args.push('--help');
+let command;
+const camelised = camelCase(commandName);
+if (camelised) {
+  command = commands[camelised];
```
arcanis: Why not

```js
commandName = camelCase(commandName);

if (!Object.prototype.hasOwnProperty.call(commands, commandName)) {
    args.unshift(commandName);
    commandName = `run`;
}

command = commands[commandName];
```
?

src/cli/index.js
```diff
@@ -98,90 +94,55 @@ commander.option('--network-concurrency <number>', 'maximum number of concurrent
...
-if (ARGS_THAT_SHARE_NAMES_WITH_OPTIONS.indexOf(commandName) >= 0 && args[0] === commandName) {
-  args.shift();
-}
-
-commander.parse(startArgs.concat(args));
+commander.parse([
+  ...startArgs,
+  // we use this for https://github.com/tj/commander.js/issues/346, otherwise
+  // it will strip some args that match with any options
+  'cmd',
```
arcanis: Use something more descriptive than `cmd`, something like `this-arg-will-get-stripped-later`. It will help the next person to read this hack to understand how it works, and doesn't risk to conflict with any future command.

src/cli/index.js
```diff
@@ -98,90 +94,55 @@ commander.option('--network-concurrency <number>', 'maximum number of concurrent
...
 commander.args = commander.args.concat(endArgs);
 
-if (command) {
-  commander.args.shift();
-} else {
-  command = commands.run;
-}
-invariant(command, 'missing command');
+// we strip cmd
+commander.args.shift();
```
arcanis: We probably want to make a `console.assert` here to make sure that we're shifting the right argument, what do you think?

src/cli/index.js
```diff
@@ -382,7 +343,7 @@ config.init({
   httpsProxy: commander.httpsProxy,
   networkConcurrency: commander.networkConcurrency,
   nonInteractive: commander.nonInteractive,
-  commandName,
+  commandName: commandName === 'run' ? commander.args[0] : commandName,
```
arcanis: Not sure we can do this; `yarn version` and `yarn run version` would both end up with a `version` commandName, which is semantically wrong because they have different behaviors. If we want to support user-defined commands here, then it should be in a separate field.

src/cli/index.js
voxsim: I added a `|| ''` because `camelCase` function return null in some cases.

The final version:

```
commandName = camelCase(commandName) || '';

if (!Object.prototype.hasOwnProperty.call(commands, commandName)) {
    args.unshift(commandName);
    commandName = 'run';
}

command = commands[commandName];
```

src/cli/index.js
voxsim: Ok this was a pre-existent error, can we open another pull request on this?

src/cli/index.js
voxsim: Mmh could you please write an example to understand on how you want to use `console.assert` here?

src/cli/index.js
voxsim: @arcanis I am sorry but I can't do that :( because if you type `yarn custom-script` it does something orrible because it camel case `custom-script`.

voxsim:
`Maybe in a future PR it could be interesting to look if we can use anything out there that would discard the need to maintain all of this logic. What do you think?`

@arcanis I think you are totally right about that ;)

If you see some something weird about reviews, I am sorry but I forgot to submit many days ago O.o

src/cli/index.js
arcanis: Oooooh right :/ damn.

src/cli/index.js
arcanis:
```js
console.assert(commander.args.length >= 1);
console.assert(commander.args[0] === 'cmd');
commander.args.shift();
```

So that the code early crashes instead of doing something else than what the user expected, which might be hard to debug otherwise.

src/cli/index.js
arcanis: Yep, you're right. Let's keep the current behavior

src/cli/commands/help.js
```diff
@@ -17,24 +21,47 @@ export function run(
...
-    }
-  } else {
-    commander.on('--help', () => {
-      console.log('  Commands:\n');
-      for (const name of Object.keys(commands).sort(sortAlpha)) {
-        if (commands[name].useless) {
-          continue;
+    const commandName = camelCase(args.shift());
+
+    if (commandName) {
```
arcanis: Good idea! I'd probably put them on a single line, comma-separate, restricted to 80 columns. Otherwise they take a lot of vertical space.

voxsim: I have to check this because commandName can be null, sorry I lost this comment!

src/cli/index.js
voxsim: This eventing I have to remember to open at least one issue, otherwise I will forget everything :)

src/cli/commands/help.js
arcanis: Yeah ... it seems this `camelCase` function should rather be called `ifCamelCaseNullElseCamelCase` 😄

voxsim: YES T_T XD

voxsim: Do you mean the `commands` part?  I can do that if you confirm

arcanis: Yep!

src/cli/commands/help.js
```diff
@@ -17,24 +21,47 @@ export function run(
...
+      const command = commands[commandName];
```
voxsim: If we validate here maybe we should validate in src/cli/index.js. What do you think?

src/cli/index.js
voxsim: done!

src/cli/commands/help.js
arcanis: Yep! It will fix this:

```
❯ [mael-mbp?] yarn git:(0.23-stable) ❯ yarn constructor

Using globally installed version of Yarn
yarn constructor v0.23.0
error An unexpected error occurred: "command.run is not a function".
info If you think this is a bug, please open a bug report with the information provided in "/Users/mael/yarn/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/constructor for documentation about this command.
```

src/cli/commands/help.js
voxsim: I can't overwrite the commands part without doing other horrible things for what I understand because it is a feature of commander.js -.-. In detail the function help.

arcanis: Actually I think it's done userland, more precisely [here](https://github.com/yarnpkg/yarn/pull/2887/files#diff-776aa1c3229433c4c56fe8345fd89668R52). That being said, I've checked locally and it seems that this display (one command each line) was already printed this way in recent Yarn releases, so let's keep it this way. 👍

voxsim: You are right! As you wish :)

src/cli/commands/help.js
voxsim: This is a pre-existent error, but i prefer to fix here right now. If I am correctly after this part, we will finish this review.

arcanis: 👍

voxsim: @arcanis done!

src/rc.js
```diff
@@ -70,7 +70,7 @@ export function getRcArgs(command: string): Array<string> {
 
   let result = rcArgsCache['*'] || [];
 
-  if (command !== '*') {
+  if (command !== '*' && Object.prototype.hasOwnProperty.call(rcArgsCache, command)) {
```
arcanis: Ah! You got me :)

voxsim: :D

PR #2896, opened 2017-03-12
Revert reduce

jordwalke:
Reverting this because it would break a bunch of packages that haven't
    converted to the new form. We might go with the static substitution
    approach.


jordwalke:
Sent PR to wrong remote again, geez why does github populate this as the default? Sorry!

Issue #2304, opened 2016-12-20
Yarn install --production installs a devDependencies package


jucrouzet:
**What is the current behavior?**

`yarn install --production`  installs packages in `devDependencies` (npm install --production doesn't).

```
/data/api # npm install --production >& /dev/null 
/data/api # ls node_modules/typescript
ls: node_modules/typescript: No such file or directory
/data/api # rm -rf node_modules/
/data/api # yarn install --production >& /dev/null 
/data/api # ls node_modules/typescript
AUTHORS.md                CopyrightNotice.txt       LICENSE.txt               ThirdPartyNoticeText.txt  jenkins.sh                netci.groovy              test.config
CONTRIBUTING.md           Gulpfile.ts               README.md                 bin                       lib                       package.json
/data/api # yarn why typescript
yarn why v0.18.1
[1/4] Why do we have the module "typescript"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info Has been hoisted to "typescript"
info Reasons this module exists
   - "typings#typings-core" depends on it
   - Specified in "devDependencies"
info Disk size without dependencies: "21.05MB"
info Disk size with unique dependencies: "21.05MB"
info Disk size with transitive dependencies: "21.05MB"
info Amount of shared dependencies: 0
Done in 2.56s.
/data/api # yarn why typings
yarn why v0.18.1
[1/4] Why do we have the module "typings"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info Has been hoisted to "typings"
info This module exists because it's specified in "devDependencies".
Done in 1.88s.
```
**What is the expected behavior?**

Typescript should not be installed.

**Please mention your node.js, yarn and operating system version.**

(In a docker container)
```
/data/api # node -v
v6.9.2
/data/api # yarn -V
0.18.1
/data/api # uname -r
4.4.0-53-generic
```

jucrouzet:
Could this be because a yarn.lock is present in a dependency with typescript in it ?

Codelica:
I don't think so.  I'm noticing the same thing, basically "yarn install --production" installing my devDependencies.   It seems to act properly prior to Yarn 0.18.x from what I've seen.  But 0.18.x and 0.19.x seem broken for production installs.

daltones:
I'm having maybe the same problem with Yarn v0.20.3. I've noticed this:

If I have this `package.json`:
[...]

PR #2944, opened 2017-03-17
Avoid installing dev dependencies when `--production`

juanca:
**Summary**

I'm not sure if this is a substantial feature request. However, I believe the issue is severe and should be addressed immediately.

As of yarn v0.18.0, yarn will sometimes install development dependencies when using the production flag (`--production`). This PR introduces a failing spec from [an existing issue](https://github.com/yarnpkg/yarn/issues/2304#issuecomment-279888161). 

**Test plan**

Test driven development: See [Implement failing spec for accidentally hoisting dev dependencies](https://github.com/juanca/yarn/commit/4fbeeae147ffae738c2df510b04f0fdc2fa17da0)

**Halp**

I would appreciate some guidance on the files to look at for possible fixes. I've dived a bit and it looks like [hoisting](https://github.com/yarnpkg/yarn/pull/2116) might be the reason for installing dev dependencies: when resolving the dependency tree, dev dependencies are still considered -- which is not straightforward -- and used in the install process. 

My initial thought is to avoid any dev dependencies and use the lockfile's resolved list of packages (not in dev dependencies) when invoking `yarn install --production` -- and essentially avoid any type of hoisting.


src/cli/commands/install.js
```diff
@@ -257,7 +257,9 @@ export class Install {
       };
 
       pushDeps('dependencies', {hint: null, optional: false}, true);
-      pushDeps('devDependencies', {hint: 'dev', optional: false}, !this.config.production);
+      if (!this.config.production) {
```
juanca: This breaks 11 other tests... However, I question the validity of at least one of the breaking specs:

```
 FAIL  __tests__/commands/import.js
  ● import missing dev deps in production
```

__tests__/commands/install/integration.js
```diff
@@ -208,13 +208,17 @@ test.concurrent('hoisting should factor ignored dependencies', async () => {
 });
 
 test.concurrent('--production flag ignores dev dependencies', () => {
-  return runInstall({production: true}, 'install-production', async (config) => {
+  return runInstall({production: true}, 'install-production-without-dev', async (config) => {
```
bestander: I think it's better to add a new test rather than change existing one if the new one covers a different case

juanca: Sounds good. 

The first test was too basic. This new integration test covers a more common case with nested dependencies and dev dependencies.

src/cli/commands/install.js
bestander: Yeah, I've seen this import test to fail some times I think

__tests__/commands/install/integration.js
arcanis: Basic tests are good :) If they break, we don't have to debug which part of the test is failing

bestander:
@juanca, we have merged recently a new fix to hoisting algorithm.
Do you want to rebase and try again?

juanca:
Will do! Thanks for the update.

juanca:
Well, I seem to be breaking less tests with the updated code. 

I might just:

- [ ] revert my attempted fix
- [ ] keep basic test
- [ ] keep non-basic test

juanca:
Also, when you have a minute of spare time, do you mind linking the hoisting fix PR? I'm interested in knowing more about yarn's hoisting mechanisms.

bestander:
This PR fixed dev dependency hoisting bug https://github.com/yarnpkg/yarn/pull/2921

juanca:
I got an intermittent failure. :/ Any way to rebuild from here?

bestander:
You mean `Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.`?
It started happening on CI, as long as Travis is green it should be fine.

bestander:
The test passes, does it mean that your case now works?

__tests__/fixtures/install/install-production-without-dev/yarn.lock
```diff
@@ -0,0 +1,1219 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
bestander: That is quite a big dependency graph, it would increase the test run time.
Is it possible to get cover the case with a smaller example?

juanca: Sure. I'll try to come up with something else. Ensure that is fails pre-fix and ensure it is green post-fix.

juanca:
Yeah, the fix made my case work! 

We will be testing in staging (then production) servers once the fix is released. 

I'll try to come up with a smaller / optimized use cases for tests. Most likely going to make extra packages in e2e test repository.

bestander:
Awesome, thanks for investing in stability

bestander:
@juanca thanks for the PR.
I'll close it as it can't be merged this way yet.
A PR with a more isolated test will be very much welcome.

Issue #764, opened 2016-10-11
yarn install hangs during "Fetching packages..."


indieisaconcept:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

yarn install hangs at fetching packages and does not provide any further information as to the cause.

**If the current behavior is a bug, please provide the steps to reproduce.**

With the following [package.json](https://gist.github.com/indieisaconcept/875c30f41eb0683e17628c6b6f7b5ae1) run the below

```
> yarn cache clean & yarn install
```

**What is the expected behavior?**

Installation should succeed.

**Please mention your node.js, yarn and operating system version.**
- yarn@v0.15.1
- node@v4.4.6
- osx@10.10.5

sorgloomer:
I have the same issue on Windows 10 using nodejs v6.2.0 x64.

It hangs when fetching the last package:

```
C:\xxx>yarn
yarn install v0.15.1
info No lockfile found.
warning xxx@0.0.0: No license field
[1/4] Resolving packages...
warning wdio-mocha-framework > mocha > glob > minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning wdio-mocha-framework > mocha > to-iso-string@0.0.2: to-iso-string has been deprecated, use @segment/to-iso-string instead.
warning wdio-mocha-framework > mocha > jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
[2/4] Fetching packages...
█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 868/869
```

indieisaconcept:
@sorgloomer yep same behaviour, it is indeed hanging on last package.

kitsunde:
Same issue but with bower.json. In this case it's working on my local macOS. (with a complication https://github.com/yarnpkg/yarn/issues/846)

<detail>
```json
{
  "name": "jaguar",
  "version": "0.0.0",
  "private": true,
  "dependencies": {
    "bootstrap": "~3.3.5",
    "devicejs": "2ae5c775e35ccc837589e5af34e292c54936778c",
    "jquery": "2.1.3",
    "jquery-transform": "e195b9a7118558bb1141e50b80380ea5f31dffb8",
    "moment": "2.14.1",
    "moment-timezone": "0.5.5",
    "owl-carousel2": "2.0.0-beta.2.4",
[...]

Issue #2351, opened 2016-12-27
Yarn stuck at Fetching packages


vinz243:
**Do you want to request a *feature* or report a *bug*?**

I'd like to report a bug/I have an issue

**What is the current behavior?**

`yarn` gets stuck at Fetching packages.

Here is the actual output with `--verbose`
```
verbose Request "https://registry.yarnpkg.com/is-gzip" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/yauzl" finished with status code 200.
verbose Performing "GET" request to "https://registry.yarnpkg.com/fd-slicer".
verbose Performing "GET" request to "https://registry.yarnpkg.com/buffer-crc32".
verbose Request "https://registry.yarnpkg.com/minimalistic-assert" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/is-absolute" finished with status code 200.
verbose Performing "GET" request to "https://registry.yarnpkg.com/is-relative".
verbose Request "https://registry.yarnpkg.com/is-natural-number" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/sum-up" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/buffer-crc32" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/fd-slicer" finished with status code 200.
verbose Performing "GET" request to "https://registry.yarnpkg.com/pend".
verbose Request "https://registry.yarnpkg.com/is-relative" finished with status code 200.
verbose Request "https://registry.yarnpkg.com/pend" finished with status code 200.
[2/4] 🚚  Fetching packages...
verbose Performing "GET" request to "https://registry.yarnpkg.com/lodash/-/lodash-4.16.6.tgz".
verbose Performing "GET" request to "https://registry.yarnpkg.com/mediastic/-/mediastic-0.3.0.tgz".
████████████████████████████████████████████████████████████████████████████████████████████████████████████ 1336/1337
````

Please note that the last package doesn't matter. On other times it was stuck on `fsevents` or other packages. 

**What is the expected behavior?**

Finish the installation as expected.

**Please mention your node.js, yarn and operating system version.**

```
 ~/cassette master ●  screenfetch

                 -/+:.          vincent@Vincents-Mac
                :++++.          OS: 64bit Mac OS X 10.11.6 15G31
               /+++/.           Kernel: x86_64 Darwin 15.6.0
       .:-::- .+/:-``.::-       Uptime: 1d 14h 2m
[...]

Issue #1289, opened 2016-10-20
Fetching packages hung forever.


huangyingjie:
**Do you want to request a _feature_ or report a _bug_?**
Report a _bug_

**What is the current behavior?**
Fetching packages hung forever, waiting 30min+

``` js
$ yarn install
yarn install v0.16.0
info No lockfile found.
[1/4] 🔍  Resolving packages...
warning mapbox-gl-shaders > brfs > static-module > through2 > xtend > object-keys@0.4.0:
[2/4] 🚚  Fetching packages...
█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 976/977
```

**If the current behavior is a bug, please provide the steps to reproduce.**
$ git clone https://github.com/mapbox/mapbox-gl-js.git && cd mapbox-gl-js
$ yarn install
**What is the expected behavior?**
Install success, or show the detailed blocking info, like `npm install --verbose`.
**Please mention your node.js, yarn and operating system version.**
node v4.4.1
Mac OSX EI 10.11.6

rask:
Same issue here. A WordPress theme with various Gulp packages refuses to fetch the last package for some reason. Yarn version 0.16.0, Node version 6+, npm version 3+. Running on Ubuntu 14.04.

TekSiDoT:
I can confirm this behavior with the following setup:
node v6.9.1
Windows 10 1607 (build 14393.0)

hangs on:
```warning electron-prebuilt > electron-download > nugget > progress-stream > through2 > xtend > object-keys@0.4.0```

Might thus be related to object-keys@0.4.0

petermikitsh:
Also experiencing this :(

petermikitsh:
The issue for me was resolved by switching to npm semantic versioned releases in `package.json`  instead of using `git+https` urls and `https://github.com/user/repo#tag`.

strife25:
I encountered a similar issue w/ yarn hanging. Could this issue be related to SSH key passwords? https://github.com/yarnpkg/yarn/issues/2702

kirs:
I fixed the issue in #2950. There's been a missing timeout (literally one line fix).

TimvdLippe:
#2950 has been merged, but did not reference this issue properly. It can now be closed.

dpurrington:
This is still happening in version v0.23.2

PR #2950, opened 2017-03-18
Specify TCP timeout when downloading packages

kirs:
It's been an old and annoying issue: yarn doesn't specify `fetch` timeout when downloading packages, causing `yarn install` to hang forever on network issues.

Steps to reproduce:

1) Install toxiproxy (proxy to simulate network latency [[1]](http://madebymunsters.com/blog/posts/simulating-poor-network-conditions-with-toxiproxy/), [[2]](https://github.com/Shopify/toxiproxy))
2. Configure toxiproxy to simulate the latency:

```
~/.bin/toxiproxy-cli create yarnpkg --listen localhost:8888 -u registry.yarnpkg.com:443
~/.bin/toxiproxy-cli toxic add yarnpkg -t latency -a latency=9000 -a jitter=50
```

3. Point `yarn.lock` to use `localhost:8888` as a host for package registry

4. Run `yarn install` and see the process hanging

***

The issue has been reported multiple times and it's a very easy fix; I'm surprised that I'm the first one to look into it. As mentioned in the `fetch` [documentation](https://github.com/request/request#timeouts), "requests to external servers should have a timeout attached". 

Specifying timeout is a critical thing for distributed software and build tools. Earlier this week we found our build servers hanging on `yarn install` step for 40 minutes until it was hard killed. It took us a while to find that it was a network problem with connectivity between EC2 and `registry.yarnpkg.com`. Having the `fetch` timeout would help us to discover the issue and see the timeout error, as well as not causing build servers to hang until a hard kill.

fixes https://github.com/yarnpkg/yarn/issues/764, https://github.com/yarnpkg/yarn/issues/1289, https://github.com/yarnpkg/yarn/issues/2351

@bestander

### DIFF
--- a/__tests__/util/request-manager.js
+++ b/__tests__/util/request-manager.js
@@ -7,6 +7,7 @@ import * as fs from '../../src/util/fs.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
 
+const net = require('net');
 const https = require('https');
 const path = require('path');
 
@@ -94,6 +95,25 @@ test('RequestManager.request with mutual TLS', async () => {
     server.close();
   }
   expect(body).toBe('ok');
+});
+
+test('RequestManager.execute timeout Request', async () => {
+  jest.useRealTimers();
+
+  const server = net.createServer((c) => {
+    // emulate TCP server that never closes the connection
+  });
+
+  try {
+    server.listen(0);
+    const config = await Config.create({networkTimeout: 50});
+    const port = server.address().port;
+    await config.requestManager.request({url: `http://localhost:${port}/?nocache`});
+  } catch (err) {
+    expect(err.message).toContain('TIMEDOUT');
+  } finally {
+    server.close();
+  }
 });
 
 test('RequestManager.execute Request 403 error', async () => {
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -90,6 +90,7 @@ commander.option(
   'disable progress bar',
 );
 commander.option('--network-concurrency <number>', 'maximum number of concurrent network requests', parseInt);
+commander.option('--network-timeout <milliseconds>', 'TCP timeout for network requests', parseInt);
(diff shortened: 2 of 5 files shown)


bestander:
Nice.
Could you fix lint issues?
Also how about making it a configuration parameter?
I wonder if 10 seconds may be too little for some cases.

I think there is huge need to write a test here but what would be your Test Plan for the change?

kirs:
Thanks for feedback! I'll work on it.

I wish `RequestManager` had better test coverage to have more examples how I can test it...

bestander:
Start with a manual Test Plan and think if this could be automated in some way

bestander:
It could be an e2e test with http requests mocked

kirs:
@bestander fixed lint, added the config option and the unit test 🎉 

Random tests are failing on the CI but I can't reproduce them locally with the same version of node. Taking the fact that CI for master is also red, I assume the test suite is not 100% stable.

kirs:
@bestander please let me know if there's something else I can improve.

kirs:
CircleCI is now green 🎉

aliyazdani:
@kirs - thanks for this. Would love to see this merged in soon.

bestander:
Great work, @kirs.
What do you think of implementing retries?
Would it be hard to have this feature in?

kirs:
Happy to see it merged!

@bestander in my implementation timeout errors are considered the same as other networking errors and they will be retried:

https://github.com/yarnpkg/yarn/blob/b2c2c3be79088285418e88734841c7452aa3a02e/src/util/request-manager.js#L348
https://github.com/yarnpkg/yarn/blob/b2c2c3be79088285418e88734841c7452aa3a02e/src/util/request-manager.js#L281

kirs:
@casperisfine @stefanmb in case you've been wondering about Locutus builds stuck on `yarn install`... This fixes the issue.

PR #2954, opened 2017-03-20
yarnrc: respect XDG_CONFIG_HOME setting

### LINKED ISSUE #2334: XDG compliance
Hi :)

I noted that Yarn doesn't fit the [XDG basedir specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
It could be nice to support it.

Currently Yarn creates its own data / cache / config directories:

- ```$HOME/.yarn```
- ```$HOME/.yarn-cache```
- ```$HOME/.yarn-config```

Next the desired replacement (or fallback) of each directory.

- ```$HOME/.yarn```: ```${XDG_DATA_HOME:-$HOME/.local/share}/yarn```
-  ```$HOME/.yarn-cache```: ```${XDG_CACHE_HOME:-$HOME/.cache}/yarn```
-  ```$HOME/.yarn-config```: ```${XDG_CONFIG_HOME:-$HOME/.config}/yarn```

OS: Linux

kevinburke:
If a user has a XDG_CONFIG_HOME environment variable, read and write
the yarnrc file from XDG_CONFIG_HOME/.yarnrc, not from $HOME/.yarnrc.

The specification can be found here:
https://wiki.archlinux.org/index.php/XDG_Base_Directory_support

Fixes #2334.

### DIFF
--- a/src/registries/yarn-registry.js
+++ b/src/registries/yarn-registry.js
@@ -46,7 +46,11 @@ export default class YarnRegistry extends NpmRegistry {
   constructor(cwd: string, registries: ConfigRegistries, requestManager: RequestManager) {
     super(cwd, registries, requestManager);
 
-    this.homeConfigLoc = path.join(userHome, '.yarnrc');
+    if (process.env.XDG_CONFIG_HOME) {
+      this.homeConfigLoc = path.join(process.env.XDG_CONFIG_HOME, '.yarnrc');
+    } else {
+      this.homeConfigLoc = path.join(userHome, '.yarnrc');
+    }
     this.homeConfig = {};
   }
 


kevinburke:
One test failed when I ran them locally but it seems unrelated; I can add a separate issue for that if you like.

```
 FAIL  __tests__/commands/install/integration.js (224.941s)
  ● install should respect --no-bin-links flag

      at ontimeout (timers.js:380:14)
      at tryOnTimeout (timers.js:244:5)
```

kevinburke:
Another option would be to remove the leading period if a user explicitly specifies a directory; the "hidden" nature of dotfiles was apparently an accident, see https://plus.google.com/101960720994009339267/posts/R58WgWwN9jp for more information.

kevinburke:
I just rebased against master and I believe this is failing now because of an unrelated third-party dependency problem.

arcanis:
Thanks! Two things:

- Could you also update the `rc.js` file (it's a new file, so you might need to pull)?
- You probably also want to update [getPossibleConfigLocations](https://github.com/kevinburke/yarn/blob/105eca9d1ae4d21e94e5190db43d3ea8ba9faf72/src/registries/npm-registry.js#L97) (otherwise, unless I'm mistaken, you will write to XDG_CONFIG_HOME, but not read from it)

jasonkarns:
This PR, while an improvement, does not bring yarn into compliance with XDG. The XDG spec defines a default value for `XDG_CONFIG_HOME` as `~/.config`. This means XDG compliance would require `~/.config/yarn` be used when `XDG_CONFIG_HOME` is not set.

Also, I would recommend not using a hidden file within the XDG location. Part of the reason for leveraging XDG is to not require the files be hidden individually. I would suggest:

`$XDG_CONFIG_HOME/yarnrc` or more conventionally: `$XDG_CONFIG_HOME/yarn/config`

jasonkarns:
@kevinburke why closed? I very much want to see xdg support in yarn (raised my concerns above not to dissuade this PR but hopefully just tweak it a bit to be more spec compliant)

felixsanz:
@kevinburke friendly ping!

riri:
I'm a new yarn user, and I was happy to see the cache goes in $XDG_CACHE_HOME, but smiled when seeing the .yarnrc. @jasonkarns is right in his analysis, and the way it should go is rather simple finally:

- if $**XDG_CONFIG_HOME** is set, use (and write to) **$XDG_CONFIG_HOME/yarn/config**
- else if **$HOME/.config** exists, use **$HOME/.config/yarn/config**
- else as a fallback, use **$HOME/.yarnrc**

kevinburke:
I don't have time to work on this, and there are enough different ideas about how this should be implemented, feel free to add your own patch

PR #2957, opened 2017-03-20
Add test for bin command

voxsim:
**Summary**/**Test plan**

As described in the title this commit add tests for bin command.


arcanis:
Seems reasonable

PR #2969, opened 2017-03-22
Fix undefined error when package has no binaries

TimvdLippe:
**Summary**

Today we installed a package which apparently didn't have any binary.
The console message however was: `warning undefined has no binaries`.
This happens because [`packageHasNoBinaries`](https://github.com/yarnpkg/yarn/blob/6e9a9a6596ca8f177f68f6672a1ef4ff16705336/src/reporters/lang/en.js#L252) has 1 argument, which this call to the reporter does not serve.
Therefore the package name will be undefined.

**Test plan**

Given that the warning is not working properly (recursion, anyone? :smile: ), we suspect that the package was `generator-polymer-init-custom-build`.
Running `yarn global add generator-polymer-init-custom-build` shows the following output:
```
yarn global v0.21.3
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning undefined has no binaries
warning No license field
Done in 9.85s.
```

With this fix, the output is:
```
yarn global v0.23.0-0
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "generator-polymer-init-custom-build@2.0.1" has no binaries
warning No license field
Done in 17.47s.
```

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -165,7 +165,7 @@ function ls(manifest: Manifest, reporter: Reporter, saved: boolean) {
     }
     reporter.list(`bins-${manifest.name}`, bins);
   } else if (saved) {
-    reporter.warn(reporter.lang('packageHasNoBinaries'));
+    reporter.warn(reporter.lang('packageHasNoBinaries', human));
   }
 }
 


arcanis:
Good catch 👌

Issue #394, opened 2016-09-21
CI (internal) vs external lock files


bestander:
At FB we check in mirror folder with all .tar.gz files and yarn.lock that refers to them so that we could build node_modules offline (especially in CI).

We will want to commit yarn.lock with references to npm repo for the OSS users.

How can we make both mirror and external work together so that we could sync this file between internal and OSS code?

poeschko:
Maybe I'm missing something, but why does the lock file have to change at all when using a mirror? Can it not keep referencing the original source of a package (e.g. https://registry.yarnpkg.com) + checksum? If a mirror is set up in .yarnrc, the "primary key" into that mirror would still be package name + version + checksum; but the mirror would only be an (optional) cache, in addition to the original source.

bestander:
@poeschko, yeah, makes sense, related discussion for a more generic approach https://github.com/yarnpkg/yarn/issues/393.
But I like that your way makes it configuration-less

sjd78:
Has there been any action on this issue recently?  My team is hitting almost exactly the same issue trying to balance external pointing yarn.lock file in source control with internal pointing yarn.lock for CI.

bestander:
It is in the scope for the core team but no one is actively working on this.

It would be great if someone from the community volunteered to implement
this.
Feel free to ping me, we could chat and split the work.

gregsheremeta:
Just to reiterate, currently the lock file **does** need to change when reading from a mirror. The URLs need to be changed to just name + checksums. If they reference URLs,

```
$ yarn install --offline
yarn install v0.19.1
[1/4] Resolving packages...
[2/4] Fetching packages...
error Can't make a request in offline mode
```

But @poeschko 's idea is spot-on. Why should it have to?

I don't see why yarn shouldn't automatically stay offline and read the mirror when told to via --offline, but I'd also be ok with another option that did the same thing.

We currently work around this by postprocessing the lock file, which is a gross hack. But it works.

```
cp yarn.lock yarn.lock.bak
sed -i -e "s#resolved \"https.*/-/\(.*\)\"#resolved \\1#" yarn.lock

```

bestander:
Go ahead, send an RFC

gregsheremeta:
@bestander Still want to split the work with us? :)

bestander:
Looks too easy to split :)
On a serious note, I am here to provide help here or on discord https://discordapp.com/invite/yarnpkg or by email bestander@gmail.com

[...]

Issue #393, opened 2016-09-21
What's the intended workflow of the tarball cache?


wycats:
I understand how the feature works, but as I'm not using it, I'm not quite sure what the exact intended workflow is.

@bestander @skevy @kittens

glenjamin:
I'm unclear from my limited use of yarn how much of a role the `resolved` field plays when installing from lockfile on a second machine - but the npm behaviour of always phoning home for this was very frustrating. It led me to always remove this field from the generated npm-shrinkwrap.

The specific flow we wanted was that I would have my development machine point at the public registry, but CI would go via a proxy.

I think the way @wycats describes storing a reference to the package source separately from the package location would help enable this workflow.

As a strawman, something along the lines of this might work:

The checked-in lock file states the expected source and a hash
```
abab@^1.0.0:
  version "1.0.3"
  source "registry"
  hash "b81de5f7274ec4e756d797cd834f303642724e5d"
```

Those sources would have default locations, and then separate environment-specific not-checked in config could override source locations - possibly giving an ordered list?

mgcrea:
I've been testing custom yarn-cache folder for a few weeks, but I'm encountering a lot of `Tarball is not in network and can not be located in cache` errors (on gitlab-ci, etc.) that would only be solved with a `yarn --no-lockfile` install. By any chance, would you have encountered such errors?

```
"error \"bunyan-1.8.4.tgz\": Tarball is not in network and can not be located in cache (\"/srv/player/.yarn-cache/bunyan-1.8.4.tgz\")", "stdout": "yarn install v0.17.0\n[1/4] Resolving packages...\n[2/4] Fetching packages...\ninfo Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command."
```

UnrememberMe:
Somewhat related but maybe a little stray from the topic, is there any thoughts about dealing with node module installation scripts? There are plenty of node modules download additional codes during installation and thus the results of `yarn install` is still non-repeatable even with the offline mirror.

bestander:
The problem is that Node.js install scripts can execute any bash script, there is no way to reliably achieve offline mode without authors' cooperation.
The only thing that we can do is not use such modules or help the authors to think of offline mode.

UnrememberMe:
How about caching the post-install results instead of pre-install results?  I understand that will not work well with any codes with native platform dependencies, but that is not a problem any of our current solutions address either.

[...]

PR #2970, opened 2017-03-22
[breaking?] - When using offline mirror yarn.lock does not change anymore

arcanis:
**Summary**

Related rfc : https://github.com/yarnpkg/rfcs/pull/51 

When enabling the offline mirror, Yarn updates the lockfile by stripping the registry URL from its `resolved` fields. This feature aims to simplify this process by making such an update unneeded.
 
Related issues: https://github.com/yarnpkg/yarn/issues/393 / https://github.com/yarnpkg/yarn/issues/394
 
 
**Test plan**

All tests should pass fine. I had a rewrite a few because of the nature of this change, but the fixtures stay the same (backward compatibility is preserved).


bestander:
How does it handle backwards compatibility?
I see that you did not change existing yarn.lock files, so it should be seamless I suppose?

arcanis:
Yep. Basically, if we have an URL, the offline cache filename is derived from the basename of this URL. If we have a path, then it is the offline cache filename. That should work with both old & new lockfiles.

bestander:
You have lint errors and 3 tests are failing.
1 of them might be flaky though

package.json
```diff
@@ -15,6 +15,7 @@
     "death": "^1.0.0",
     "debug": "^2.2.0",
     "detect-indent": "^5.0.0",
+    "flow": "^0.2.3",
```
bestander: why do you need flow in deps?

src/fetchers/git-fetcher.js
```diff
@@ -15,22 +16,61 @@ const fs = require('fs');
...
 
 export default class GitFetcher extends BaseFetcher {
-  _fetch(): Promise<FetchedOverride> {
-    const {protocol, pathname} = url.parse(this.reference);
-    if (protocol === null && typeof pathname === 'string') {
-      return this.fetchFromLocal(pathname);
-    } else {
-      return this.fetchFromExternal();
+  async getLocalAvailabilityStatus(): Promise<bool> {
+    const tarballLegacyMirrorPath = this.getTarballMirrorPath({withCommit: false});
```
bestander: I think this justifies a comment explaining what is legacyMirrorPath and why we support it

package.json
arcanis: I tried last friday to bump Flow's version to check if my issue was fixed in the new versions, but forgot to rollback the change.

src/fetchers/git-fetcher.js
```diff
@@ -51,53 +91,48 @@ export default class GitFetcher extends BaseFetcher {
...
-          reject(new MessageError(reporter.lang('fetchErrorCorrupt', err.message, localTarball)));
+          reject(new MessageError(this.reporter.lang('fetchErrorCorrupt', err.message, tarballPath)));
         });
     });
   }
 
   async fetchFromExternal(): Promise<FetchedOverride> {
-    const commit = this.hash;
-    invariant(commit, 'Commit hash required');
+    const hash = this.hash; // if we don't store it in a temp variable, flow will not persist its invariant state
```
bestander: This is quite a common pattern in flow, you probably don't need to add a comment to describe it.
If someone decides to inline the variable flow won't let them anyway :)

Also I think we don't add comments on the same line with code (I might be wrong though)

arcanis:
👍

troyastorino:
Does this mean that the [description of the lockfile in the offline-mirror blog post](https://yarnpkg.com/blog/2016/11/24/offline-mirror/#initialize-the-new-lockfile) isn't accurate anymore?

If so, should probably either update the blog post, or port the content to [the docs page that links to the blog post](https://yarnpkg.com/lang/en/docs/offline-mirror/) and update the content there

bestander:
Good find.
Would you want to help the project and send a PR with blog post changes?

troyastorino:
Sure, happy to give it a go! I'm somewhat new to yarn, and I stumbled on this trying to setup the offline mirror cache for a sprawling multi-package repo (haven't gotten it working yet), so I want to make sure I understand what's going on:

* The lockfile should look the same whether or not the offline mirror is used?
  - It seems like this might be related to why [github dependencies aren't getting installed from the offline mirror cache in v0.23](https://github.com/yarnpkg/yarn/issues/3154#issuecomment-294306257) — the Resolved field looks like `https://codeload.github.com/stevemao/left-pad/tar.gz/aff6d744155a70b81f09effb8185a1564f348462` but the tarball in the offline mirror cache directory is just named `aff6d744155a70b81f09effb8185a1564f348462`. 
* A decent way to debug whether offline-mirror is working would be to run `yarn --offline` after you've cleared out your yarn cache?

Sunyang730:
Doesn't seems like the blog post is up to date. I will create a PR for that

bestander:
Thanks!

cc me please

griffinmichl:
Does this solve: https://github.com/yarnpkg/yarn/issues/3154?

bestander:
Git dependencies should still work but there can be an edge case where it broke, thanks for linking

griffinmichl:
Another potentially related ticket: #3194

ericgundrum:
This change has broken git+https dependencies as described in #3309. There is a mismatch between the calculated hash and the expected hash.

bestander:
Yeah, we are on it

steveax:
For anyone else wondering if the docs for offline mirror are up-to-date, [they were updated after this was merged](https://github.com/yarnpkg/website/commit/571273f0d14fa615e4cf20e77b75dbe692afe9a3#diff-220a2b6a12c106eb10cf45c2f00e184c)

Issue #2151, opened 2016-12-05
Wrong auth for scoped packages in private registries


donaldpipowitch:
**Do you want to request a *bug*?**
_bug_

**What is the current behavior?**
**If the current behavior is a bug, please provide the steps to reproduce.**

When a custom registry for scoped packages is used the authentification for the default registry is used. E.g. in my project I have a `.npmrc` like this:

```
strict-ssl=false
@some-scope:registry=https://npm-registry.our-company.lan/repository/npm-registry/
```

In my `~/.npmrc` I'm authenticated with the public npm registry:

```
//registry.npmjs.org/:_authToken=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

Now I call `$ yarn` to install deps and get `error Couldn't find package "@some-scope/some-package" on the "npm" registry.`. (This is actually a 401 error, not 404! See #2152 for this incorrect error message.)

**What is the expected behavior?**

Don't use the `_authToken` for `registry.npmjs.org` when I use the private registry for `@some-scope` so `@some-scope/some-package` can be installed correctly.

**Please mention your node.js, yarn and operating system version.**

```
$ node -v
v6.9.1

$ yarn -V
0.18.0

macOS 10.12.1
```

tyrsius:
I'm working around this right now with an alias that moves the `~/.npmrc` file to a backup, installs from yarn, then restores the `~/.npmrc` file.

FezVrasta:
Is anyone working on this? It's pretty annoying and prevents me from using Yarn 😩

danez:
This is fixed by #3231 in 0.26.0

PR #2975, opened 2017-03-22
Remove fallback npmjs.org auth data

### LINKED ISSUE #2953: npm auth token used on private repo for scoped packages

bug

```
$ cat .yarnrc
registry "http://example.com/nexus/repository/npm-repositories"
$ cat ~/.npmrc
//registry.npmjs.org/:_authToken=xxxxxxx
$ yarn --verbose
yarn install v0.21.3
verbose 0.232 current time: 2017-03-19T20:32:51.471Z
info No lockfile found.
verbose 0.468 Performing "GET" request to "http://example.com/nexus/repository/npm-repositories/@types%2fnode".
verbose 0.863 Request "http://example.com/nexus/repository/npm-repositories/@types%2fnode" finished with status code 401.
verbose 0.864 Error: Couldn't find package "@types/node" on the "npm" registry.
    at /usr/local/Cellar/yarn/0.21.3/libexec/lib/node_modules/yarn/lib/resolvers/registries/npm-resolver.js:209:15
    at step (/usr/local/Cellar/yarn/0.21.3/libexec/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /usr/local/Cellar/yarn/0.21.3/libexec/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
error Couldn't find package "@types/node" on the "npm" registry.
$ curl -v http://example.com/nexus/repository/npm-repositories/@types%2fnode > /dev/null 2>&1 | grep HTTP
> GET /nexus/repository/npm-repositories/@types%2fnode HTTP/1.1
< HTTP/1.1 200 OK
$ mv ~/.npmrc ~/.npmrc_original
$ yarn
[...]
✨  Done in 2.98s
```

(Host name and auth token replaced with dummy for privacy reasons.)

See above.

Do not use npm's auth token on a private repository for scoped packages.

```
$ node --version
v7.7.3
$ yarn --version
0.21.3
```

danez:
**Problem**

The current behaviour is that for scoped packages (`@<scope>/<pkgName>`) yarn always sends an authentication to the registry. This was introduced in #1146 in order to support private packages on npmjs.org
When a user is loggedin on the cli to npmjs.org and has configured a different pirvate registry, which in most cases is a proxy to npmjs.org, yarn still sends the npmjs.org auth data to the private registry, making the private-registry fail as it does not know anything about npmjs.org credentials.

**Summary**

The technical reason this happens is that because of the changes in #1146 the flag ‘always-auth’ is set to true for scoped packages (which is actually fine I think), but the method to retrieve the authToken/credentials has a fallback to npmjs.org auth data.
https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L175
The fix was simply to remove all fallbacks from getAuth().
Imho the Fallback to npmjs.org-credentials should never happen, as yarn should never ever send credentials from npmjs.org to any other API besides npmjs.org.
The fallback to `''` is most probably in there with old npm clients which were storing credentials without a registry-scope in the config. I couldn't find where `--registry` was added but it was already available in 1.4.

By looking at it I saw that it still supports insecure plaintext basic auth which is a feature from npm <= 1.4, so I took the risk and removed it. Neither npm >= 2 or yarn do save basic auth data anymore and npm 2 was released in 2014. (`_token` is also dropped in npm@5)

**Test plan**

I could add some unit-tests for npm-registry, though the critical part would more be to do integration tests.
But I wait for initial feedback before writing tests.

Fixes #2953 #2151

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -172,28 +172,12 @@ export default class NpmRegistry extends Registry {
       return this.token;
     }
 
-    for (let registry of [this.getRegistry(packageName), '', DEFAULT_REGISTRY]) {
-      registry = registry.replace(/^https?:/, '');
+    const registry = this.getRegistry(packageName).replace(/^https?:/, '');
 
-      // Check for bearer token.
-      let auth = this.getScopedOption(registry.replace(/\/?$/, '/'), '_authToken');
-      if (auth) {
-        return `Bearer ${String(auth)}`;
-      }
-
-      // Check for basic auth token.
-      auth = this.getScopedOption(registry, '_auth');
-      if (auth) {
-        return `Basic ${String(auth)}`;
-      }
-
-      // Check for basic username/password auth.
-      const username = this.getScopedOption(registry, 'username');
-      const password = this.getScopedOption(registry, '_password');
-      if (username && password) {
-        const pw = new Buffer(String(password), 'base64').toString();
-        return 'Basic ' + new Buffer(String(username) + ':' + pw).toString('base64');
-      }
+    // Check for bearer token.
+    const auth = this.getScopedOption(registry.replace(/\/?$/, '/'), '_authToken');
+    if (auth) {
+      return `Bearer ${String(auth)}`;
     }
 
     return '';


danez:
The tests are failing because of timeouts.

arcanis:
> When a user is loggedin on the cli to npmjs.org and has configured a different registry, which in most cases is a proxy to npmjs.org, yarn still sends the npmjs.org auth data to the registry, making the proxy-registry fail as it does not know anything about npmjs.org credentials.

That doesn't look like a wrong behavior to me. Shouldn't users logout instead if they don't wish to be authenticated to the registry? Wouldn't this patch prevent people from setting up authentication-protected registries?

danez:
This patch simply removes the fallback to npmjs.org credentials. Currently yarn would send my npmjs.org credentials (which i configured to be able to publish to npmjs.org) to the private repository that some of my work repositories have set.
But the private repositories (at least all I've seen) have their own user management (usually LDAP etc.). So the npmjs.org credentials will not work, and therefore the requests fail for scoped packages.

bestander:
That sounds reasonable although I don't understand the full impact of the change.
Some examples and e2e tests will help for sure

danez:
Fixed by #3231

PR #2982, opened 2017-03-24
Mock inquirer for all e2e tests

### LINKED ISSUE #2955: tests failing due to timeout errors
When running the test suite locally with the latest master, the following tests occasionally fail.

```
  ● install should respect --no-bin-links flag

      at ontimeout (timers.js:380:14)
      at tryOnTimeout (timers.js:244:5)

  ● should not loose dependencies when installing with --production

      at ontimeout (timers.js:380:14)
      at tryOnTimeout (timers.js:244:5)
```

I believe the second test should be "lose dependencies" as well.

bestander:
Fixes #2955, we don't want yarn to prompt during e2e installs

**Test plan**

CI should be green


arcanis:
Shouldn't that be an error by default if Yarn asks for multiple versions in tests?

bestander:
Appveyor tests are fixed with revert #2985

bestander:
> Shouldn't that be an error by default if Yarn asks for multiple versions in tests?

There is definitely an inconsistency in tests caching implementation, we should tackle it sooner.
The tests sometimes go to remote servers to get package metadata but the committed caches in `__tests__/fixtures/request-cache` don't match the online data and inquirer gets invoked.

I'd just like to get tests green sooner

arcanis:
👍

Issue #1796, opened 2016-11-12
ssh:// prefix is not removed from URLs passed to git CLI


pornel:
Bug in v0.17.0. Still in v0.20.3.

**What is the current behavior?**

URL in the format `git+ssh://host_alias:path.git` is given to git clone as `ssh://host_alias:path.git`. This is not compatible with npm, which clones `host_alias:path.git` (without the `ssh://` prefix). This changes the meaning of the URL in `git`.

(By `host_alias` I mean a name given in `Host` setting in `~/.ssh/config`. I'm using git URLs in such format, because it allows me to configure specific SSH deployment keys for private github repos.)

**What is the expected behavior?**

`ssh://` prefix should be stripped from URLs given to `git` CLI.

**Please mention your node.js, yarn and operating system version.**

yarn v0.17.0-20161111.1922. Tested again in v0.20.3.

#513 says git+ssh support is fixed, but this particular case still doesn't work for me in today's nightly.

Gastonite:
I've the same bug in yarn version v0.23.0-20170314.1919

Issue #573, opened 2016-10-10
Install from private github url 


thejameskyle:
**Do you want to request a _feature_ or report a _bug_?** Bug

**What is the current behavior?**

You can't install a package like `github:yarnpkg/yarn` because it is private even though if 

```
[2/4] 🚚  Fetching packages...
error git@github.com:yarnpkg/yarn.git: Tarball is not in network and can not be located in cache (/Users/thejameskyle/Projects/yarn-perf/git@github.com:yarnpkg/yarn.git)
```

**What is the expected behavior?**

It should be able to install the package like npm currently can.

doot0:
This is also the case for private URLs that require basic authentication, e.g. `git+https://gitusername:gitpassword@server/path/to/repo.git`

milosivanovic:
I've provided a simple workaround for authenticating via SSH in yarn at https://github.com/yarnpkg/yarn/issues/513#issuecomment-253008950 that you can use until this issue is fixed.

martynchamberlin:
Looks like @ramasilveyra closed https://github.com/yarnpkg/yarn/pull/971 in deference to https://github.com/yarnpkg/yarn/pull/1081, but that PR didn't actually fix the problem, at least for me. I'm manually grabbing a non-published bleeding-edge version of yarn so that it will have https://github.com/yarnpkg/yarn/pull/1081 in it, but I'm still getting this error: 

```
Tarball is not in network and can not be located in cache [...]
```

Any suggestions what to change to get this to work?

martynchamberlin:
Eh, ignore my comment. I had universally installed a stable release version of yarn and due to how my VM is setup, this was the one it was referencing instead of the bleeding-edge local version. So that's why I was getting that error. I've confirmed I no longer get that error on the latest version.

stiang:
I’ll just add that I initially had some problems getting this to work *after* it was reportedly fixed. It turned out that we had been using URLs like the following in package.json:
```
git+ssh://git@github.com:org/repo.git
```

Note the ":" between the hostname and the org name. When I replaced the ":" with a "/" it worked perfectly:
```
git+ssh://git@github.com/org/repo.git
```

I’m not sure it the first URL scheme is supposed to work, but I’m pretty sure we copied it from someone at some point, so we’re probably not alone in using it.

BryanCrotaz:
We are working on a PR that will handle this URL format as detailed above by @stiang

szimek:
[...]

PR #2990, opened 2017-03-25
Special parsing for git's not-really-urls containing scp host specifiers

pornel:
What git calls a URL is not strictly a URL, but a union of URLs and "scp host specifiers" (`host:path`). The syntax can't be treated as a regular URL, because hostname of scp hostname specifiers parses as a URL scheme.

This removes the previous workaround (`transformUrl`, #573), which was transforming scp host specifiers to ssh:// URLs. This wasn't compatible with npm, as it changed semantics of the URL and caused git to ignore SSH host aliases (#1796). 

The previous workaround also accidentally disabled `hasArchiveCapability` check — the regex in `hasArchiveCapability` expected `:` as a way to detect scp host specifiers. Since scp goes over SSH, I assume the check can be generalized to check for the SSH protocol, so I've made that explicit.

To make it clear that git "URLs" are not real URLs, I've created a new type for them. I also use the new type to store a couple of precomputed properties that used to be extracted from the URL (the `protocol` had to be stored explicitly anyway because of the git "URL" parsing ambiguity).


src/util/git.js
```diff
@@ -44,32 +50,49 @@ export default class Git {
...
+  gitUrl: GitUrl;
 
+  /**
+   * npm URLs contain a 'git+' scheme prefix, which is not understood by git.
+   * git "URLs" also allow an alternative scp-like syntax, so they're not standard URLs.
+   */
+  static npmUrlToGitUrl(npmUrl: string): GitUrl {
+    // Special case in npm, where ssh:// prefix is stripped to pass scp-like syntax
+    // which works as remote path only if there are no slashes before ':'
+    const match = npmUrl.match(/^git\+ssh:\/\/((?:[^@:\/]+@)?([^@:\/]+):.*)/);
```
arcanis: Can git URLs have a password, ie. `foo:bar@example.org`? This regex doesn't catch this case

pornel:
The scp urls are in the test. The format is `host:path` or `user@host:path`.

`git+ssh://foo:bar` should be passed to git as `git clone foo:bar`, and not as `ssh://foo/bar`.

src/util/git.js
pornel: This is intentional. In scp host specifiers there is no password.

Git manual (`man git-clone`) specifies it as `[user@]host.xz:/~[user]/path/to/repo.git/`, although relative paths are also supported (scp man page documents the syntax as `[[user@]host1:]file`)

pornel:
@arcanis Thanks for the review. Is there anything else that needs to be done?

arcanis:
Just one question: I see you're using `npmUrlToGit` before each call to a `Git` static method, except when calling the `Git` constructor itself. I understand that your methods require a specific type for their arguments to highlight that they aren't actually url - shouldn't the constructor do the same?

pornel:
I changed static methods only because it was necessary, but kept constructor as-is to make the change smaller. But it makes sense to have them all consistently take the same type of URL, so I've made that change.

arcanis:
Thanks for the rebase 👍

Issue #2774, opened 2017-02-24
Use shortand for gitlab dependencies


ramiel:
**What is the current behavior?**
`yarn add "gitlab:mygitlabuser/myproject"`

gives an error

**What is the expected behavior?**

Should behave the same as    
`npm install -S "gitlab:mygitlabuser/myproject"`

With npm this is a valid dependency also in package.json, while it doesn't work with yarn. This is actually a simple shorthand for a complete git url and work the same for bitbucket and gist

**Please mention your node.js, yarn and operating system version.**
node v7.6.0 on Kde Neon (linux 4.4.0-64-generic)

Daniel15:
Feel free to send a pull request adding this if you like.

ramiel:
If I have time it'll be a pleasure!

ramiel:
I investigated and discovered that yarn is totally capable to resolve that kind of dependency references (i.e. gitlab:mygitlabuser/myproject). The problem is, it is not able to download private packages. On the opposite npm is capable of doing it. The problem, I think, is that npm resolves the shorthand into the url
```
git+ssh://git@gitlab.com/mygitlabuser/myproject.git#<ref>
```
while yarn transforms it into
```
 https://gitlab.com/mygitlabuser/myproject.git
```
and so, it encounters authentication problems because it isn't using ssh.    
The yarn codebase is large and I need more time to understand why and where it differs like this from npm. If somebody can have a look at this or give me an hint, it should be appreciated.

ramiel:
Or maybe that url is only used to get back the ref from the repository. In any case npm does something different to retrieve it.

hx-markterry:
This line here tries to check to see if accessing a private repo would work over HTTP:

https://github.com/yarnpkg/yarn/blob/4fc2fe3ce4b19a90a3bbe2fd423bf543b3186f14/src/resolvers/exotics/hosted-git-resolver.js#L204

it always seems to pass, if you change it to return false, packages will be installed via git successfully.

I wonder if a "prefer git" CLI option could be used here, or have a better way of checking if a repo is available over HTTP.

ramiel:
Npm check http first,for what I can see in the logs,than fallback to git+ssh

SunnyTam:
suffering same problem

thedumbterminal:
Working on a PR to solve this issue: https://github.com/yarnpkg/yarn/pull/2992

hx-markterry:
This should be resolved in version 0.23.

[...]

PR #2992, opened 2017-03-26
Supports public and private git repos when using shorted repo syntax

thedumbterminal:
**Summary**

When using the shortened git syntax such as `github:thedumbterminal/grunt-gemnasium#vx`, yarn tests for HTTP access using a generated URL ending in `.git` this will always return a `301` status even if the user does not have permission to access the repo (private repos would require auth first).

This change now uses a full git URL which does not end in `.git` this means that private repos will fail with `404` so then SSH auth is attempted instead, as long as the correct SSH keys are setup in the users terminal access to the private repo will succeed.

Addresses existing issue: https://github.com/yarnpkg/yarn/issues/2774

**Test plan**

Added new unit tests for the new `getGitHTTPBaseUrl` method.

This can be manually tested in the following way:

1. Create a new nam project with: `mkdir /tmp/example; cd /tmp/example; npm init -y`
1. Update package.json to add a public repo to dependencies manually in the format of `"grunt-gemnasium": "github:thedumbterminal/grunt-gemnasium#v1.1.1"`
1. Update package.json to add a private repo to dependencies manually in the same format, (make sure you can only access it via SSH creds).
1. Run: `yarn --verbose`, both modules should be installed.

### DIFF
--- a/__tests__/resolvers/exotics/github-resolver.js
+++ b/__tests__/resolvers/exotics/github-resolver.js
@@ -27,15 +27,26 @@ test('getGitSSHUrl with no hash', () => {
   expect(gitSSHUrl).toContain('some-user');
 });
 
-test('getGitHTTPUrl should return the correct git github SSH url', () => {
+test('getGitHTTPBaseUrl should return the correct git github HTTP base url', () => {
   const fragment: ExplodedFragment = {
     user: 'foo',
     repo: 'bar',
     hash: '',
   };
 
-  const expected =  'git+ssh://git@github.com/' + fragment.user + '/' + fragment.repo + '.git';
-  expect(GitHubResolver.getGitSSHUrl(fragment)).toBe(expected);
+  const expected =  'https://github.com/' + fragment.user + '/' + fragment.repo;
+  expect(GitHubResolver.getGitHTTPBaseUrl(fragment)).toBe(expected);
+});
+
+test('getGitHTTPUrl should append ".git" to the HTTP base URL', () => {
+  const fragment: ExplodedFragment = {
+    user: 'foo',
+    repo: 'bar',
+    hash: '',
+  };
+
+  const expected =  GitHubResolver.getGitHTTPBaseUrl(fragment) + '.git';
+  expect(GitHubResolver.getGitHTTPUrl(fragment)).toBe(expected);
 });
 
 test('getGitSSHUrl should return URL containing protocol', () => {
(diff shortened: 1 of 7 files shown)


thedumbterminal:
@bestander any thoughts on the approach taken here?

bestander:
so far looks reasonable, go on :)

thedumbterminal:
Cheers will finish this up hopefully in the week.

__tests__/resolvers/exotics/github-resolver.js
```diff
@@ -27,15 +27,26 @@ test('getGitSSHUrl with no hash', () => {
...
-test('getGitHTTPUrl should return the correct git github SSH url', () => {
+test('getGitHTTPBaseUrl should return the correct git github HTTP base url', () => {
   const fragment: ExplodedFragment = {
     user: 'foo',
     repo: 'bar',
     hash: '',
   };
 
-  const expected =  'git+ssh://git@github.com/' + fragment.user + '/' + fragment.repo + '.git';
-  expect(GitHubResolver.getGitSSHUrl(fragment)).toBe(expected);
```
thedumbterminal: Previously the `getGitHTTPUrl` test was actaully running `getGitSSHUrl` instead!

__tests__/resolvers/exotics/gitlab-resolver.js
```diff
@@ -27,15 +27,26 @@ test('getGitSSHUrl with no hash', () => {
...
-test('getGitHTTPUrl should return the correct git gitlab SSH url', () => {
+test('getGitHTTPBaseUrl should return the correct git gitlab HTTP base url', () => {
   const fragment: ExplodedFragment = {
     user: 'foo',
     repo: 'bar',
     hash: '',
   };
 
-  const expected =  'git+ssh://git@gitlab.com/' + fragment.user + '/' + fragment.repo + '.git';
-  expect(GitLabResolver.getGitSSHUrl(fragment)).toBe(expected);
```
thedumbterminal: Previously the `getGitHTTPUrl` test was actaully running `getGitSSHUrl` instead!

thedumbterminal:
CI seems to be failing on random tests unrelated to my changes, this work is now ready for someone to review.

bestander:
Awesome work, @thedumbterminal!

gsklee:
Just tried it and it doesn't work (github repo shorthand with private repo); it prompted me to enter my ssh key's passphrase but I cannot actually enter anything at all.

thedumbterminal:
@gsklee You need to make sure you have your ssh key setup in your terminal first, if you are using linux or mac osx then you can run:

    eval `ssh-agent`
    ssh-add

Then try running the yarn command again.

Issue #2204, opened 2016-12-09
should output build logs in verbose mode


bestander:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the expected behavior?**

Sometimes something goes wrong when dependencies are installed and --verbose mode helps to identify problems.
Quite a few problems with Yarn are related to binary dependencies builds.

In verbose mode Yarn should output stdout of install lifecycle scripts instead of
```
[4/4] 📃  Building fresh packages...
[1/2] ⡀ dtrace-provider
[-/2] ⡀ waiting...
[-/2] ⡀ waiting...
[-/2] ⢀ waiting...
[-/2] ⢀ waiting...
```

xdissent:
Any guidance on what the output should look like? Something like this might be a quick and easy win:

```diff
--- a/src/package-install-scripts.js
+++ b/src/package-install-scripts.js
@@ -101,7 +101,8 @@ export default class PackageInstallScripts {
 
     try {
       for (const [stage, cmd] of cmds) {
-        await executeLifecycleScript(stage, this.config, loc, cmd, spinner);
+        const {stdout} = await executeLifecycleScript(stage, this.config, loc, cmd, spinner);
+        this.reporter.verbose(stdout);
       }
     } catch (err) {
       err.message = `${loc}: ${err.message}`;
```

But longer term I wonder if there's a format that would handle big builds better.

bestander:
Yeah, I think reporter.verbose could be a good solution.
There may be edge cases of course.
Send a PR and let's see it in action :)

manuq:
Shouldn't this be closed now that #2994 was merged?

bestander:
Thanks!

Myrmod:
As reference for Version 3 it's `yarn install --inline-builds`

PR #2994, opened 2017-03-26
Add verbose build log

xdissent:
**Summary**

Outputs build log when using `--verbose` flag as discussed in #2204.

**Test plan**

None! =) I couldn't find a single test for `PackageInstallScripts`, so this change is untested. If that's a deal breaker or an oversight on my part I'll update this PR.


arcanis:
Looks good, thanks! 👍

PR #2997, opened 2017-03-27
Lock to minor version 1.1.x of request-capture-har

### LINKED ISSUE #2996: request-capture-har is breaking yarn on node 4
The dep in package.json is ^1.1.4, so you all bring in 1.2.1, which is broken on node 4.

sdawood:
request-capture-har 1.2.1 introduced es6 syntax, which breaks yarn install on node 4.x, which is a wide deployment base especially for aws lambda

Closes #2996

**Summary**

**Test plan**

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "proper-lockfile": "^2.0.0",
     "read": "^1.0.7",
     "request": "^2.75.0",
-    "request-capture-har": "^1.1.4",
+    "request-capture-har": "~1.1.4",
     "rimraf": "^2.5.0",
     "roadrunner": "^1.1.0",
     "semver": "^5.1.0",


Daniel15:
You'll need to re-run `yarn install` to update the `yarn.lock` file. It still lists `request-capture-har@^1.1.4` rather than `request-capture-har@~1.1.4`

We should probably do this for all dependencies... The `^` is too broad of a range IMO.

TimvdLippe:
Per https://github.com/yarnpkg/yarn/issues/2996#issuecomment-289377483 this seems to be fixed again.

arcanis:
I'll close this one, since https://github.com/yarnpkg/yarn/pull/2995 has been merged into master 😃

PR #3000, opened 2017-03-27
Resolve Git dependencies offline if tarball is available.

### LINKED ISSUE #1925: Cannot install packages from github when offline

raymondwang:
**Summary**

When running yarn offline, Git dependencies get trapped in `hasHTTPCapability` even though `resolveOverHTTP` defaults to returning the tarball. This change allows us to bypass that step altogether if the tarball already exists.

Resolves #1925.

**Test plan**

Create a `package.json` with a Git dependency, like:
```
"nomnom": "douglasduteil/nomnom#next"
```

Then:

```
yarn
rm -r node_modules
yarn --offline
```

This should fail prior to this commit, but succeed after.

This change was first proposed in #2705, but @chrisgavin closed the PR and hasn't had a chance to revisit. I'm happy to add unit tests if/as needed, but may need some guidance on mocking offline functionality.

### DIFF
--- a/src/resolvers/exotics/hosted-git-resolver.js
+++ b/src/resolvers/exotics/hosted-git-resolver.js
@@ -15,7 +15,7 @@ export type ExplodedFragment = {
   hash: string,
 };
 
-export function explodeHostedGitFragment(fragment: string, reporter: Reporter): ExplodedFragment {  
+export function explodeHostedGitFragment(fragment: string, reporter: Reporter): ExplodedFragment {
 
   const preParts = fragment.split('@');
   if (preParts.length > 2) {
@@ -39,7 +39,7 @@ export function explodeHostedGitFragment(fragment: string, reporter: Reporter): 
     throw new MessageError(reporter.lang('invalidHostedGitFragment', fragment));
   }
 
-  const userParts = fragment.split('/');  
+  const userParts = fragment.split('/');
 
   if (userParts.length >= 2) {
 
@@ -140,11 +140,6 @@ export default class HostedGitResolver extends ExoticResolver {
   }
 
   async resolveOverHTTP(url: string): Promise<Manifest> {
-    const shrunk = this.request.getLocked('tarball');
-    if (shrunk) {
-      return shrunk;
-    }
-
     const commit = await this.getRefOverHTTP(url);
     const {config} = this;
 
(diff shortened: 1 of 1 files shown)


src/resolvers/exotics/hosted-git-resolver.js
```diff
@@ -201,7 +201,7 @@ export default class HostedGitResolver extends ExoticResolver {
     // If we can access the files over HTTP then we should as it's MUCH faster than git
     // archive and tarball unarchiving. The HTTP API is only available for public repos
     // though.
-    if (await this.hasHTTPCapability(httpUrl)) {
+    if (this.request.getLocked('tarball') || await this.hasHTTPCapability(httpUrl)) {
```
arcanis: It works, but I think it would make more sense semantically speaking to just move [those lines](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L143-L146) at the beginning of the `resolve()` function instead.

raymondwang: Yes, sounds good to me!

raymondwang:
@arcanis Done!

arcanis:
Thanks!

tp:
I am correct in assuming that this is not part of the just released yarn 0.23.4?

Because that one still fails for me with the "--offline" flag.

```
=> yarn --offline
yarn install v0.23.4
error Can't make a request in offline mode
```

If so, what release would this be included in then?

arcanis:
Yes. It will be released in the 0.24, which we want to publish later today.

PR #3009, opened 2017-03-28
Ability to opt-out of offline mirror

danharper:
`.yarnrc` files are already merged with `.yarnrc` files in parent directories, so currently if a `.yarnrc` in a parent directory enables the offline mirror, all projects in child directories would also use the offline mirror.

However there's currently no way to _disable_ the offline mirror in projects under a directory which enables is.

This commit allows a child project to opt-out of the offline mirror in their own .yarnrc with `yarn-offline-mirror false`.

Previously if you specified `false` you'd get an exception as false is parsed as boolean but it expects a string.

Also included tests verifying the existing .yarnrc merging behaviour (previously wasn't covered).

I've made an alteration to one of the test helpers to allow using a subdirectory as a test fixture.

The test fixtures layout:

- `offline-mirror-configuration/`
  - `.yarnrc` _specifies an offline mirror directory_
  - `enabled-from-parent/`
    - `package.json`
    - _this project will use the offline mirror as specified in the parent dir_
  - `enabled-from-parent-merge/`
    - `package.json`
    - `.yarnrc` _specifies misc options, testing for merge with parent config_
  - `disabled-locally/`
    - `package.json`
    - `.yarnrc` _overrides the parent offline mirror to disable it_

// cc @bestander


danharper:
CI failures just seem to be flakey tests.

bestander:
Yeah, we have some CI instability currently, working on it.

__tests__/commands/_helpers.js
```diff
@@ -81,9 +81,21 @@ export async function run<T, R>(
...
   let cwd;
   if (fixturesLoc) {
-    const dir = path.join(fixturesLoc, name);
+    let parentDir = name;
+    let childDir = '';
+
+    // if the name is a path, copy the parent dir, then use the child dir as the cwd
+    // there's some existing tests which reference fixtures in a parent dir (../), so excluding those
+    const nameSepIndex = name.indexOf('/');
+    if (nameSepIndex !== -1 && name.indexOf('../') === -1) {
```
bestander: This is hard to parse for me.
What if there are multiple separators in the name a/b/c?
Also it gets complicated with relative paths to the fixture.
Any idea to simplify this?
Maybe move the logic for changing CWD to the test itself?

__tests__/commands/install/integration.js
```diff
@@ -809,6 +810,36 @@ test.concurrent('lockfile should be created when missing even if integrity match
   });
 });
 
+test.concurrent('offline mirror can be enabled from parent dir', (): Promise<void> => {
+  return runInstall({}, 'offline-mirror-configuration/enabled-from-parent', async (config, reporter) => {
+    const rawLockfile = await fs.readFile(path.join(config.cwd, 'yarn.lock'));
+    const lockfile = parse(rawLockfile);
+    expect(lockfile['mime-types@2.1.14'].resolved).toEqual(
+      'mime-types-2.1.14.tgz#f7ef7d97583fcaf3b7d282b6f8b5679dab1e94ee',
```
bestander: Oh, https://github.com/yarnpkg/yarn/pull/2970 landed just yesterday, we don't strip `https://registry.yarnpkg.com/mime-types/-/` from the lockfile anymore if we are using offline mirror.

You probably can check that file was added to the offline mirror folder as an indicator.

bestander:
@danharper, this is great, love the tests!
Just a few fixes to the tests and it is ready to go

__tests__/commands/_helpers.js
danharper: Agreed it's a little weird. It works with multiple separators (copies the top-most dir, uses deepest dir as the cwd).

To avoid dual-purposing the name field, I could allow `name` to have the type: `string | { source: string, cwd: string }`, e.g. 

```js
return runInstall({}, { source: 'offline-mirror-configuration', cwd: 'enabled-from-parent' }, async (config, reporter) => {
  // ...
});
```

Or, could allow changing the cwd from within the test body, but it'd require another callback as the cwd is required for setup.

bestander: Yeah, having an object with `cwd` field like you proposed makes sense, let's give it a try

bestander:
ping, @danharper, needs a rebase

danharper:
Rebased @bestander

Issue #2897, opened 2017-03-12
Yarn doesn't fetch all remote refs from git repositories


a-rodin:
**Do you want to request a *feature* or report a *bug*?**

Incompatibility with `npm` (probably a bug)

**What is the current behavior?**

By default `yarn` doesn't fetch all remote refs from git repositories. When it takes a git URL with a commit hash it does something equivalent to

```
git clone git://github.com/LearnBoost/websocket.io
cd websocket.io
git checkout 5f0b3799f50944bc4d1eb58acdef61c38883d6cf # fails
```

The commit `5f0b3799f50944bc4d1eb58acdef61c38883d6cf` is from a pull request and by default it is not fetched by `git`, so this command would fail. 

On the other hand, `npm` does something equivalent to
```
git clone git://github.com/LearnBoost/websocket.io
cd websocket.io
git fetch --all
git checkout 5f0b3799f50944bc4d1eb58acdef61c38883d6cf # works
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
$ yarn global add "git://github.com/LearnBoost/websocket.io#5f0b3799f50944bc4d1eb58acdef61c38883d6cf"
yarn global v0.21.3
warning No license field
[1/4] Resolving packages...
error Couldn't find manifest in "git://github.com/LearnBoost/websocket.io".
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

Originally I found this by trying to do `yarn global add markserv` which uses this as a dependency.

**What is the expected behavior?**
```
$ npm install -g  "git://github.com/LearnBoost/websocket.io#5f0b3799f50944bc4d1eb58acdef61c38883d6cf"
/home/user/.nvm/versions/node/v6.10.0/lib
└── websocket.io@0.2.2  (git://github.com/LearnBoost/websocket.io.git#5f0b3799f50944bc4d1eb58acdef61c38883d6cf)
```
(installs without errors)

**Please mention your node.js, yarn and operating system version.**
```
$ yarn --version
0.21.3
$ node --version
v6.10.0
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"
```

XaserAcheron:
I managed to run into the same error through the following sequence of events:
[...]

PR #3012, opened 2017-03-29
Resolve git hashes that are in GitHub pull requests

josephfrazier:
**Summary**

This fixes the use case mentioned in https://github.com/yarnpkg/yarn/issues/2897#issuecomment-289969368:

    yarn add https://github.com/mozilla/webextension-polyfill#99faff94ffbf03094361875a10e1692acc4b8662

**Test plan**

I added a broken test in the second commit, then fixed it in the third. The first commit was to fix pre-existing lint issues.


__tests__/package-resolver.js
```diff
@@ -53,6 +53,7 @@ function addTest(pattern, registry = 'npm', init: ?(cacheFolder: string) => Prom
 // addTest('https://github.com/npm-ml/re'); // git url with no .git
 // addTest('git+https://github.com/npm-ml/ocaml.git#npm-4.02.3'); // git+hash
 // addTest('https://github.com/npm-ml/ocaml.git#npm-4.02.3'); // hash
+addTest('https://github.com/mozilla/webextension-polyfill#99faff94ffbf03094361875a10e1692acc4b8662'); // commit in pull request
```
josephfrazier: https://github.com/yarnpkg/e2e-test-repo may be a more appropriate location to create a pull request (and leave it open) for this test

src/util/git.js
```diff
@@ -349,7 +349,7 @@ export default class Git {
   }
 
   async setRefRemote(): Promise<string> {
-    const stdout = await child.spawn('git', ['ls-remote', '--tags', '--heads', this.url]);
+    const stdout = await child.spawn('git', ['ls-remote', this.url]);
```
josephfrazier: This ensures we get all the remote references, including pull requests.

src/util/git.js
```diff
@@ -360,14 +360,15 @@ export default class Git {
...
-    const {hash} = this;
+    const {hash, cwd} = this;
 
     const names = Object.keys(refs);
 
     if (Git.isCommitHash(hash)) {
       for (const name in refs) {
         if (refs[name] === hash) {
           this.ref = name;
+          await child.spawn('git', ['fetch', 'origin', name], {cwd});
```
josephfrazier: `git ls-remote` isn't enough to make the files in the pull request available, so we need to fetch the reference here.

src/util/git.js
```diff
@@ -406,7 +407,11 @@ export default class Git {
     for (const line of refLines) {
       // line example: 64b2c0cee9e829f73c5ad32b8cc8cb6f3bec65bb refs/tags/v4.2.2
       const [sha, id] = line.split(/\s+/g);
-      let name = id.split('/').slice(2).join('/');
+      let name = id.split('/').slice(1).join('/');
+
+      if (!name.startsWith('pull/')) {
+        name = name.split('/').slice(1).join('/');
+      }
```
josephfrazier: Pull request references need to preserve their 'pull/' prefix in order to be fetchable. See https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally

arcanis:
I wonder if that's really something we want to do - it could open a possible attack vector that would allow any user to distribute data on the behalf of another (for example, I'm not sure how git would react if someone opened a pull request with a hash that would conflict with another commit in the repository).

Does this PR enables things that weren't possible before? If you want to use a pull request in your dependencies, I would suggest adding a reference to the source repository rather than the target repository. ie:

```
{
    "websocket.io": "yjjnls/websocket.io#7888e3b11e9f00dfc4613a3b7e8b3393e12e6c45"
}
```

rather than

```
{
    "websocket.io": "LearnBoost/websocket.io#7888e3b11e9f00dfc4613a3b7e8b3393e12e6c45"
}
```

josephfrazier:
Thanks for taking a look, @arcanis. That's a good point about the risk of colliding commit hashes (however unlikely in practice it may be currently).

Using the PR's source repository instead is a good workaround, but I think my original intent was to make it more obvious that the "canonical" repository is still being used under the assumption that the pull request would eventually be merged in and the dependency could revert to being simply a package name. Looking back on it, I'm not sure if this is all that beneficial, and I'd be ok with closing this PR if the team thinks it's not worth the effort to support this use case.

arcanis:
I've been thinking about this a bit more, and I don't think that exposing pull request raw commit ids is something we should do. On top of the collision issue, it also obfuscates the actual source of the commit, which might be a problem during code / security reviews.

I'm going to close this PR, but I appreciate your contribution, thanks! :)

josephfrazier:
No problem, @arcanis. The code/security review implication of this is another good reason not to make this change. Thanks again for reviewing!

PR #3018, opened 2017-03-29
Fix bug when cannot set a config value to empty string

iredchuk:
* Fixes regression introduced in PR #2440
* Related to #2434

**Summary**

Currently `$ yarn config set <config_key> ""` sets the respective config value to true. It should be possible to set config values to an empty string.

Omitting config value should still default to **true**, e.g.
`$ yarn config set  strict-ssl`
should set **strict-ssl** config value to **true**.

**Test plan**

`$ yarn config set version-tag-prefix ""`
should set **version-tag-prefix** config value to an empty string.

### DIFF
--- a/__tests__/commands/config.js
+++ b/__tests__/commands/config.js
@@ -32,9 +32,15 @@ test('cache-folder flag has higher priorities than .yarnrc file', (): Promise<vo
     });
 });
 
-test('set true when option value is empty', (): Promise<void> => {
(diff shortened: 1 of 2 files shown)


arcanis:
Thanks !

coridrew:
This is still not working for me. I have yarn 1.3.2 installed.

iredchuk:
@coridrew What exactly is not working? Do you have steps to reproduce? For me it looks fine with yarn 1.3.2

PR #3025, opened 2017-03-31
`--check-files` for install/add/upgrade/check commands that verifies node_modules consistency

bestander:
**Summary**

This PR expands internal implementation of integrity check functionality.

1. `.yarn-integrity` file now contains all the properties as exploded JSON instead of sha-1 hashed.
This makes it easier to debug issues with integrity checks and installations

2. `.yarn-integrity` now contains list of files in node_modules if install/add command has `--check-files` flag passed.
When `check` command is executing it will have an option to test that files exist on disk.
This makes the command useful to identify when node_modules gets partially removed.

3. To use the new file checking feature:
```
yarn install --check-files
yarn check --integrity --check-files
```

This is an opt-in feature so people should not be affected buy it yet.
In the next iteration we may consider making --check-files the default option.
Also we may want to reevaluate other flags in `check` command and end up with more semantic/self explanatory options.

4. Exploded .yarn-integrity and especially `files` field allows us to integrate with third-party tools that may need list of files before `yarn install` is executed (e.g. BUCK).
Also it allows us to make a quick integrity check without loading much of the Yarn internals - only check that lock files entries are present, flags are the same and folders exist.

**TODO (coming in a follow up PR)**
- More tests for edge cases in check.js marked with TODO
- More refactoring to check/install UX

# Test plan

## New tests

See code

### Node 6.9.1, 7.8.0

| repo | .yarn-integrity size | (.yarn-integrity + files) size  | node_modules num files | T file check | T file generated |
| --- | --- | --- | --- | --- | --- |
| git@github.com:facebook/react-native.git | 4K | 752K | 14670 | 380ms | 700ms |
| https://github.com/angular/material2 | 220K | 2.2M | 39134 | 850ms | 1500ms |
| https://github.com/kumarharsh/yarn-bug | 248K | 3.8M | 61244 | 1350ms | 2200ms |

`T file check` - average time spent checking that all files exist in node_modules
`T file generated` - average time spent generating `files` field in .yarn-integrity during `install`

## Conclusions

The time added for checking files is growing linearly depending on the number of files.
About 250ms for every 10K files


bestander:
@arcanis for review (cc @davidaurelio)

bestander:
TODO: a few tests marked as TODO in check.js

src/reporters/lang/en.js
```diff
@@ -253,8 +249,13 @@ const messages = {
 
   couldBeDeduped: '$0 could be deduped from $1 to $2',
   lockfileNotContainPattern: 'Lockfile does not contain pattern: $0',
-  integrityHashesDontMatch: 'Integrity hashes don\'t match',
-  noIntegrityHashFile: 'Couldn\'t find an integrity hash file',
+  integrityCheckFailed: 'Integrity check failed',
+  noIntegrityFile: 'Couldn\'t find an integrity file',
+  integrityCheckLinkedModulesDontMatch: 'Integrity check: Linked modules don\'t match',
+  integrityPatternsDontMatch: 'Integrity check: Patterns don\'t match',
+  integrityFlagsDontMatch: 'Integrity check: flags don\'t match',
```
arcanis: Nit: `flags` should be uppercased to match the other sentences

src/integrity-checker.js
```diff
@@ -117,12 +199,38 @@ export default class InstallationIntegrityChecker {
...
+    try {
+      expected = JSON.parse(expectedRaw);
+    } catch (e) {
+      // ignore JSON parsing for legacy text integrity files compatibility
+    }
+    let integrityMatches;
+    if (expected) {
+      integrityMatches = this._compareIntegrityFiles(actual, expected);
+      if (flags.checkFiles && expected.files.length > 0) {
+        // TODO we may want to optimise this check by checking only for package.json files on very large trees
```
arcanis: Or maybe on random-but-reproductible selected files? For example one file every two entries?

Another idea: register files and directories in two separate lists, and map them with their mtime. During the integrity check, just compare the directories mtime on the filesystem with those they should have. If they don't match, it's that the directory has been tempered with, and you can check its content to see if a file has been deleted. This would drastically decrease the number of I/O checks for the most common case.

src/reporters/lang/en.js
bestander: fixed

src/integrity-checker.js
bestander: Yeah, I think we will evolve this solution once we understand how it is going to be used.
mtime makes sense especially that we already use it in fs.copy.
Maybe just a map {path: mtime} or {path: {size, mtime, or the whole stat}}.

I could not make a decision at the time of writing so decided to go there if we need it.

bestander:
@arcanis, fixed nit

PR #3027, opened 2017-03-31
fix uncalled matchers

danharper:
**Summary**

While on something else I noticed some matchers weren't being executed (i.e. just `expect(...)` with no assertion), I've fixed them.

**Test plan**

Existing tests pass.

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -794,14 +794,14 @@ test.concurrent('install should not continue if integrity check passes', (): Pro
     let reinstall = new Install({}, config, reporter, await Lockfile.fromDirectory(config.cwd));
     await reinstall.init();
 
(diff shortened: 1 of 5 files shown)


__tests__/commands/install/integration.js
```diff
@@ -794,14 +794,14 @@ test.concurrent('install should not continue if integrity check passes', (): Pro
     let reinstall = new Install({}, config, reporter, await Lockfile.fromDirectory(config.cwd));
     await reinstall.init();
 
-    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'yarn.test')));
+    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'yarn.test'))).toBe(true);
```
abouthiroppy: I think it would be better to use `toBeTruthy()`.

bestander: +1 although we use a lot of .toBe(true), we probably should do a one-off search and replace for all places like that

Issue #2266, opened 2016-12-15
node-gyp builds fail on bundled yarn


bestander:
**Do you want to request a *feature* or report a *bug*?**
**Bug**

**What is the current behavior?**

```
bestander-mbp:bignum bestander$ node ~/work/yarn/artifacts/yarn-0.17.10.js add base85
yarn add v0.17.10
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
⠁
⠁
⠁
⠁
error /Users/bestander/work/temp/bignum/node_modules/bignum: Command failed.
Exit code: 127
Command: sh
Arguments: -c node-gyp configure build
Directory: /Users/bestander/work/temp/bignum/node_modules/bignum
Output:
sh: node-gyp: command not found
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

**Please mention your node.js, yarn and operating system version.**

Node 6.9.1, Yarn 0.17.10

bestander:
Reason - we use `path.join(__dirname, '..', '..', 'bin', 'node-gyp-bin'))` in source.
When everything is bundled in one file __dirname is incorrect.

bestander:
+ needed tests to prevent from __dirname breaking bundled builds later

wyattjoh:
This is currently causing issues for our build, anything I can do to assist?

terribleplan:
Our build has been broken since docker started including this and we stopped running `npm install -g yarn`. I understand that it is not recommended, but at least it actually worked.

Paging @Daniel15 since he seems like a cool guy ;)

bestander:
Thanks the team is in the loop, we'll figure something out

terribleplan:
Cool, just wanted to make sure. We have a workaround in place, but it was pretty bad/interesting/weird/fun to have our build and dev workflow break when it wasn't our fault.

Daniel15:
Based on https://github.com/nodejs/docker-node/issues/346, it sounds like running `yarn install -g node-gyp` is a viable workaround for now.

> Paging @Daniel15 since he seems like a cool guy ;)

😊

Daniel15:
Based on https://github.com/nodejs/docker-node/issues/346, it sounds like running `yarn global add node-gyp` is a viable workaround for now.

> Paging @Daniel15 since he seems like a cool guy ;)

😊

[...]

PR #3030, opened 2017-04-01
Use standalone .js bundle in dist tarball rather than individual JS files

### LINKED ISSUE #2966: Consider distributing just the standalone .js file within the tarball
Feature

Our installation methods (tarball, Windows installer, Debian package) all include the raw JS files.

It might be beneficial to just include the standalone JS file along with all the other auxiliary files (`yarn.cmd`, `yarn` shell script, and `node-gyp` stuff). This would reduce the amount of space taken by a Yarn installation from ~13 MB to ~2.8 MB, mainly by avoiding all `node_modules` files that are not required by Yarn at runtime. It would also significantly improve installation time due to the much smaller number of files requiring extraction.

One of the reasons Node are using the standalone JS file in their Docker image (rather than the installation script) is due to the fact that it's smaller.

Daniel15:
**Summary**

Instead of including all the raw JS files in the dist tarball, just use the single Yarn JS file that's built as part of the build, along with a few other files that are required. This significantly reduces the number of files in the tarball:

```
C:\src\yarn\dist (bundle-as-dist) (yarn@0.23.0-0)
λ find .
.
./bin
./bin/node-gyp-bin
./bin/node-gyp-bin/node-gyp
./bin/node-gyp-bin/node-gyp.cmd
./bin/yarn
./bin/yarn.cmd
./bin/yarn.js
./bin/yarnpkg
./bin/yarnpkg.cmd
./lib
./lib/v8-compile-cache.js
./lib/yarn-cli.js
./LICENSE
./package.json
```

There are three .js files in the archive:
 - `lib/v8-compile-cache.js`:  Speeds up instantiation time by using the V8 code cache (https://www.npmjs.com/package/v8-compile-cache). This needs to be separate as it has to load **before** the bulk of the application code is loaded, so it can **not** be bundled
 - `lib/yarn-cli.js`: Contains all the bundled Yarn code
 - `bin/yarn.js`: Entry point to the app, just like today. Loads `v8-compile-cache` then loads `yarn-cli`

This change means that **only** the JavaScript files that are actually used are included, resulting in a nice file size reduction for the installation packages:

Differences are due to differing compression algorithms: Debian packages use xz or LZMA, RedHat uses gzip, Windows installer uses Cabinet

They're also slightly faster to extract:
image 3

Testing was performed on my desktop computer (Intel Core i5 6500, Samsung 850 Evo 1TB SSD, Windows 10), with testing for Linux stuff (like installing the Debian package) tested in a Docker container.

Raw data: https://docs.google.com/spreadsheets/d/1d8jdf3DU_GUFdotlPl08PkYa8SkzStK2tgnQ54ivsm0/edit?usp=sharing

Performance is very slightly faster when using `v8-compile-cache` along with the bundled file, but it's not extremely significant (`yarn --version` went from 0.19s to 0.14s on my BuyVM server). The difference might be bigger on servers with slower disks (HDD) or with more overloaded servers.

I also deleted the `build-dist.ps1` file because we _should_ be able to assume that Bash is available on Windows, particularly if Git is installed (as it comes with Git Bash). I need to verify that this works on AppVeyor.

**Test plan**
I tested the Debian package and Windows installer on my system. Both work as expected. I also manually inspected the tarball to ensure files were in the right place and `bin/yarn` still works as expected.

Wait for CI to pass before merging!

Closes #2966

cc @yarnpkg/core


Daniel15:
The Windows build is getting "stdout is not a tty" but it works fine on my computer... Will need to take a closer look.

bestander:
Oh My!

cpojer:
Wow, great work, man!

________________________________

bestander:
Love it so much!
Let me know when ready to merge.
What do you think of getting rid of roadrunner as well?
The extra cache layer for people who run from master branch does not make much impact.

Daniel15:
It looks like there's a build error in Node.js 4.x as build-webpack.js is using syntax it doesn't support: https://ci.appveyor.com/project/kittens/yarn/build/1942/job/33t035v195s3k19f

I'll update that, and then this should be ready to merge.

cdaringe:
does this affect:

- support requests?  if we use a bundle, the trace may be less helpful in the generated log file
- debug-ability?

w/ how many issues are open ATM, it would make sense to me that support take precedence over these perf improvements (which do look great, btw).

just sharing my opinion!

Daniel15:
> debug-ability?

People that are debugging Yarn itself would probably be using Yarn directly from this Git repo, which isn't affected by these changes

> support requests? if we use a bundle, the trace may be less helpful in the generated log file

This is one thing I was worried about, however all the function names remain intact (no minification is performed) so stack traces shouldn't be too bad.

We could ship two tarballs, Debian packages and Windows installers if needed - a "debug" one that's like the tarball today, and a "release" one that's the smaller combined source file.

bestander:
From my experience of debugging Yarn a lot, a single file build almost never was a problem

Daniel15:
I'm going to merge this so we can get it into the next RC. We can iterate as needed. The AppVeyor build is passing, and the CircleCI failure is due to unrelated test timeouts (the actual build itself passed).

scripts/build-dist.sh
```diff
@@ -1,23 +1,40 @@
...
+eval $system_yarn run build
+eval $system_yarn run build-bundle
+chmod +x artifacts/*.js
+
+cp package.json dist/
+cp LICENSE dist/
+cp artifacts/yarn-legacy-*.js dist/lib/yarn-cli.js
+cp bin/yarn-bundle-entry.js dist/bin/yarn.js
+cp bin/{yarn,yarnpkg,*.cmd} dist/bin/
+cp -r bin/node-gyp-bin dist/bin/
```
bestander: Do we want to copy node-gyp from node_modules, too?

Daniel15: Ahh, good catch! Do we need it? I tried `yarn install bcrypt` and it seemed to work, but perhaps I have a globally-installed node-gyp or something like that. I can test on a fresh VM and see if it works.

Daniel15: Actually I wonder if Yarn should install node-gyp when needed (ie. run `yarn global add node-gyp`) rather than bundling it? I think bundling node-gyp would also require us to bundle all of its dependencies, so we'll end up with a giant node_modules directory bundled with Yarn again :(

bestander: I agree.
That actually will solve the issue of single js file distributions.
Maybe Yarn could check if `node-gyp` binary is available and if not then install it.

bestander: The down side is that in some cases we Yarn might need sudo permissions?
What if Yarn would install node-gyp locally next to yarn.js?

teohhanhui: Does this change anything about #2266?

Daniel15: @teohhanhui  - We'll most likely update Yarn to automatically install node-gyp if needed (see #3114)

Issue #2429, opened 2017-01-11
[BUG] yarn runs node-gyp operations in parallel which can cause file lock error on ~/.node-gyp resources and fail


zumwald:
**Do you want to request a *feature* or report a *bug*?**
**bug**

**What is the current behavior?**
yarn installation with certain packages causes EBUSY error in node-gyp step. Perhaps due to Yarn trying to install all native modules simultaneously while `npm i` installs them synchronously.

**If the current behavior is a bug, please provide the steps to reproduce.**
run `yarn` in a directory with the following package.json:
```
{
  "name": "testing",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "diskusage": "0.1.5",
    "electron-spellchecker": "0.5.8"
  }
}
```

**What is the expected behavior?**
yarn installation of native packages is not impacted by how many/which other native packages are compiling at the same time.

**Please mention your node.js, yarn and operating system version.**
node: 6.2.2
OS: Windows x64 10.0.14393
Yarn: 0.18.1 (installed via `npm i -g yarn`)

**detailed output**
output from `yarn`:
```
error N:\repo\DesktopClient\app\node_modules\keyboard-layout: Command failed.
Exit code: 1
Command: C:\windows\system32\cmd.exe
Arguments: /d /s /c node-gyp rebuild
Directory: N:\repo\DesktopClient\app\node_modules\keyboard-layout
Output:
N:\repo\DesktopClient\app\node_modules\keyboard-layout>if not defined npm_config_node_gyp (node "C:\Users\username\AppD
ata\Roaming\npm\node_modules\yarn\bin\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node  rebuild )
gyp info it worked if it ends with ok
gyp info using node-gyp@3.4.0
gyp info using node@6.2.2 | win32 | x64
gyp info spawn C:\Users\username\.windows-build-tools\python27\python.exe
gyp info spawn args [ 'C:\\Users\\username\\AppData\\Roaming\\npm\\node_modules\\yarn\\node_modules\\node-gyp\\gyp\\gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'msvs',
gyp info spawn args   '-G',
gyp info spawn args   'msvs_version=2015',
gyp info spawn args   '-I',
[...]

PR #3032, opened 2017-04-02
Providing override for CHILD_CONCURRENCY via ENV

juturu:
**Summary**

Building native node modules parallely on windows is causing linker errors as iojs.lib/node.lib files could get locked as explained in #2429 

**Test plan**

Set CHILD_CONCURRENCY to 1 and ran yarn install with the package.json having multiple native node modules. Builds ran sequentially.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -38,6 +38,7 @@ export type ConfigOptions = {
   production?: boolean,
   binLinks?: boolean,
   networkConcurrency?: number,
+  childConcurrency?: number,
   networkTimeout?: number,
   nonInteractive?: boolean,
 
@@ -107,6 +108,8 @@ export default class Config {
   constraintResolver: ConstraintResolver;
 
   networkConcurrency: number;
+
+  childConcurrency: number;
 
   //
   networkTimeout: number;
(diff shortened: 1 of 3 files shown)


juturu:
Thanks @bestander for the review. I made the suggested changes. Can you please take a look?

src/config.js
```diff
@@ -226,6 +229,13 @@ export default class Config {
       constants.NETWORK_CONCURRENCY
     );
 
+    this.childConcurrency = (
+      opts.childConcurrency ||
+      Number(this.getOption('child-concurrency')) ||
+      Number(process.env.CHILD_CONCURRENCY) ||
```
bestander: Do we need an extra environment variable for this?
Would not CLI arg and .yarnrc setting be enough?

bestander:
Thanks, @juturu.
Could you rebase please?
Also I have concern about adding one more env variable configuration, is there a strong reason to have it?

juturu:
Did the rebase. The reason i left env configuration was i felt it was easier from gulp task to set the value before calling yarn, instead of changing the args/.yarnrc file.

bestander:
Thanks, @juturu.
I noticed we don't have any documentation for the environment variables that Yarn is using and the number is growing.

@juturu, would you send a PR for the website to add a new Documentation sub page for environment variables and add a description for CHILD_CONCURRENCY?
Similar to https://github.com/yarnpkg/website/commit/218fa02a0d4fac0bff50b47c0336a33a238dc945

juturu:
@bestander - checkout https://github.com/yarnpkg/website/pull/448

bestander:
Awesome, thanks

PR #3033, opened 2017-04-03
[feature] Ability to pass command CLI arguments via configuration in .yarnrc

arcanis:
**Summary**

This PR opens the way to share configuration between multiple projects via the use of the `.yarnrc`
 files. The current design it pretty simple: it simply adds the supported options from the yarnrc file at the beginning of the command line.

Some notes:

- The yarnrc is currently parsed synchronously. I believe this is not an issue, since it's quite literally the very first thing we need to do when booting the application (even before the command line can be parsed), so there's really nothing we could run concurrently anyway.

- The yarnrc is read in another file (https://github.com/yarnpkg/yarn/blob/master/src/registries/yarn-registry.js). I haven't changed how it works there, since I wanted my first draft to have as few changes as possible, but it's something that should probably be fixed so that the code only use one path to manage the yarnrc files.

**Test plan**

I still have to write a test to be sure that it works as expected, but the existing ones don't break.


src/rc.js
```diff
@@ -0,0 +1,101 @@
...
+    'flat',
+
+    'cache-folder',
+    'modules-folder',
+    'global-folder',
+
+    'mutex',
+
+    'no-emoji',
+    'har',
```
bestander: Do you think we should keep the list of synced args and .yarnrc properties?
I wonder if we can make it a dynamic setting that would be generic

arcanis: I've been wondering the same thing, but I'm a bit worried about the potential effects of letting the users put any argument into their rc files :/ For example, `--force` in the yarnrc seems like a bad idea.

bestander:
How about having a 'flags' section there or maybe prefixing all flags, e.g.
'argument-flag-force: true'?

arcanis:
It makes sense for most of the other parameters to be used as-is (for example, I'd prefer to write `cache-folder` rather than `argument-cache-folder`). But maybe we could only prefix a few parameters (on the top of my head, `--force`, `--verbose`, `--json`) and let the other ones be used without requiring such a prefix.

bestander:
Consider how this feature will be used in the long term and also how you
will handle collisions.
For example, there might be a setting that is named the same way as a flag
in one of the commands.

arcanis:
Shouldn't that be avoided? Even if arguments are stored in a separate "namespace", it sounds confusing to have two similarly named keys that unlock two different behaviors.

My initial approach was to consider that semantically speaking, nothing prevented the values stored in an rc file from being specified as a command line argument (in fact, the default behavior of the `rc` module is to return the result of applying the command line arguments on the content of the rc files - a behavior I've disabled so that we can have more control on how to do it). Do you think the yarnrc parameters should be completely dissociated from the command line arguments?

bestander:
Yeah, that actually makes sense.
Thanks for thinking it though.

I'll let you decide on that part as long as we don't have to hardcode every
new argument that needs a default value via rc file.

bestander:
A few use cases to consider:

```
yarn check --integrity --check-files # move  --integrity --check-files to .yarnrc
yarn install --mutex network:8084 # move --mutex network:8084 to .yarnrc
```

arcanis:
I was thinking about a syntax:

```
--production
--cache-folder /var/tmp/caches
--install.skip-integrity-check
```

Basically, all command line flags would be prefixed by `--` (so we wouldn't have to whitelist them, and their effects would be a bit clearer), and they could be restricted to a command by prefixing them with the command name (so `--install.foobar` would only add the `--foobar` option when calling the `install` command). It would make it easier to avoid adding arguments that don't make sense, and would prevent mistakes with flags that may have different meanings depending on the commands being run (`--force`).

bestander:
I like it!

arcanis:
Should be mostly fine, I'm waiting for the tests. The only issue I had is that the `rc` module I've used doesn't pass the yarnrc location to the custom parser function, which makes it impossible to resolve paths inside the yarnrc relative to this location. I'll try to see if they would merge a PR (https://github.com/dominictarr/rc/issues/61), but in the meantime the relative paths read from a yarnrc will just be ignored. WDYT?

src/rc.js
```diff
@@ -0,0 +1,87 @@
+/* @flow */
+
+import rc from 'rc';
```
bestander: I think we use require() for npm dependencies

bestander:
I think you don't need to bother about paths relative to .yarnrc.
Because .yarnrc files can be cascading it may be hard to keep track of the paths relative to all the files.
Instead the path should be relative to CWD IMHO.

felixfbecker:
> The yarnrc files are extracted starting from your current working directory - not the package.json directory.

This may be true if only yarn is run in the process, but not if yarn is used programmatically as part of an application. Blocking the whole app seems very dangerous in that scenario. I would appreciate it if you could keep this use case in mind for future improvements.

arcanis:
I'm not sure I see what you mean - do you have an example in mind?

felixfbecker:
You can install and `require()` locally and use it e.g. as part of a web server that does code intelligence. Of course you could use a child process, but that has the overhead of spinning up a V8 instance.

bestander:
I think we need to improve Yarn require-ability for programmatic use.
Anyone volunteers to start an RFC?

PR #3035, opened 2017-04-03
Add option to disable the self-update check

danharper:
**Summary**

Allow disabling of the self-update check with a `disable-self-update-check` option.

t16181086

**Test plan**

Forced the outdated prompt to appear, and verified that adding `disable-self-update-check true` to a `.yarnrc` file prevented it from appearing.

There's currently no tests for the self-update checks.

// cc @bestander

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -662,6 +662,11 @@ export class Install {
       return;
     }
 
+    // don't check if disabled
+    if (this.config.getOption('disable-self-update-check')) {
+      return;
+    }
+
     // only check for updates once a day
     const lastUpdateCheck = Number(this.config.getOption('lastUpdateCheck')) || 0;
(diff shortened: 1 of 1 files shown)


bestander:
Good job, @danharper!
What do you think would be the best way to let users know about this feature?
Maybe a new page on https://yarnpkg.com/en/docs?
Would you create a PR for a new section with this setting please?

danharper:
There doesn't seem to be a dedicated page for yarn-specific config options yet. Perhaps a separate page under https://yarnpkg.com/en/docs/configuration? (e.g. adding `.yarnrc` alongside `package.json` and `yarn.lock` sections)?

bestander:
Yeah, makes sense

centigrade-thomas-becker:
I arrived here from #3718. How do I use this option?

bestander:
@centigrade-thomas-becker,

```
yarn config set disable-self-update-check true
```

PR #3039, opened 2017-04-04
Optimized `check --integrity` and added tests

### LINKED ISSUE #2514: yarn check --integrity slowdown
UX

For current React Native repo:

```
$ yarn install
yarn install v0.18.1
[1/4] �  Resolving packages...
success Already up-to-date.
✨  Done in 0.52s.
```

```
$ yarn check --integrity
yarn check v0.18.1
success Folder in sync.
✨  Done in 1.26s.
```

Why does yarn check --integrity takes 3 times longer than yarn install with integrity check?

Node 6.9.1
Mac OS
Yarn >= 0.18.1 (0.17.10 was fine)

bestander:
**Summary**

Fixes #2514

Integrity check does not need resolver to be initialized and install hydrated.
It needs to read yarn.lock and pass patterns and entries to integrity checker.

This change reduces execution time of `check --integrity` from 2 seconds to 0.2 seconds.

Also I added a few more tests to integrity checking logic and moved some lockfile and integrity related tests into a separate test file

**Test plan**

New tests in CI


__tests__/commands/install/lockfiles.js
```diff
@@ -0,0 +1,290 @@
...
+
+test.concurrent('install uses OS line endings when lockfile doesn\'t exist', async (): Promise<void> => {
+  await runInstall({}, 'install-infers-line-endings-from-existing-lockfile',
+    async (config): Promise<void> => {
+      const lockfile = await promisify(fsNode.readFile)(path.join(config.cwd, 'yarn.lock'), 'utf8');
+      expect(lockfile.indexOf(os.EOL)).toBeGreaterThan(0);
+    });
+});
+
+test.concurrent('install should rewrite lockfile if patterns can be merged', (): Promise<void> => {
```
bestander: This is new test, all others are moved from integrity.js

src/integrity-checker.js
```diff
@@ -214,7 +213,16 @@ export default class InstallationIntegrityChecker {
     let integrityMatches;
     if (expected) {
       integrityMatches = this._compareIntegrityFiles(actual, expected);
-      if (flags.checkFiles && expected.files.length > 0) {
+      if (flags.checkFiles && expected.files.length === 0) {
+        // edge case handling - --check-fies is passed but .yarn-integrity does not contain any files
+        // check and fail if there are file in node_modules after all.
+        const actualFiles = [];
+        await this._getFilePaths(loc.locationFolder, actualFiles);
```
arcanis: shouldn't `_getFilePaths` return something rather than mutate its argument, C++-style?

bestander: Do you think it would be better this way?
_getFilePaths could have a return value, I actually considered making it a non recursive function but did not come up with a more readable code.

Do you want to suggest a non recursive _getFilePaths implementation?

PR #3040, opened 2017-04-04
fix --verify-tree test stability

bestander:
Fixed flakiness of `--verify-tree should check dev dependencies` test.

We had a few tests that verify that Yarn supports setting config.production via environment variable.
Jest does not allow us to execute a test in complete isolation, if they are in different test suites.
So those few tests that changed environment variable could affect any other test running in parallel with it.

The code that handles env variables is quite straight forward so I don't expect us to break it unintentionally.
And removing the tests that affect global state of Jest sounds like a better alternative to me vs running all tests in sequence by default.


arcanis:
Oh wow :o

PR #3062, opened 2017-04-07
CI stability: running all Jest workers single threaded

bestander:
**Summary**

We still have too many CI failures because of connection timeouts.
I'll rerun the test on Circle 10 times to verify if decreasing concurrency helps.


bestander:
Apparently single-threadedness did not help

bestander:
I'll open another  PR with other ideas

PR #3063, opened 2017-04-07
Remove the dependency on the "rc" module

arcanis:
**Summary**

We now ship our own logic to search for the Yarnrc file (initially, this change is prompted because the `rc` module doesn't support webpack, but it also allows us to resolve relative paths relatively to the yarnrc file in which they are found).

**Test plan**

Tests should all pass, a skipped test has been enabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "node-gyp": "^3.2.1",
     "object-path": "^0.11.2",
     "proper-lockfile": "^2.0.0",
-    "rc": "^1.2.1",
     "read": "^1.0.7",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -72,6 +71,7 @@
     "gulp-watch": "^4.3.5",
     "jest": "^19.0.2",
     "mock-stdin": "^0.3.0",
+    "shebang-loader": "^0.0.1",
```
bestander: do you needs a shebang loader now?

arcanis: Nope, I can remove it!

PR #3064, opened 2017-04-07
Checks that the webpack builds are working properly

arcanis:
**Summary**

This PR adds a small script to make sure that the generated artifacts (the two most recently modified files) are working properly after `build-webpack.js` has been called.


Daniel15:
Thanks! I'd prefer if it was something more explicit:

```
node ./artifacts/yarn-$version.js --version
node ./artifacts/yarn-legacy-$version.js --version
```

I'll roll that change into #3030

PR #3065, opened 2017-04-07
Fix missing subdeps in production install when those are present in d…

SimenB:
Cherry picked as per https://github.com/yarnpkg/yarn/pull/2921#issuecomment-292515854


SimenB:
Aww, it fails :(

SimenB:
@bestander master is red, so I won't look into this. Ok?

bestander:
Yeah, we have this instability in CI.
Looks like it is one of the flaky tests failing, should be ok

SimenB:
@bestander Sorry to nag (especially on a weekend), but when can we expect a patch release? 😄

bestander:
Oh, sorry @SimenB, I just got this message lost in my emails.
I'll send the patch now.

SimenB:
0.23.2 is stable so no issue for me 😄

PR #3067, opened 2017-04-07
Fixes integrity check for --production flag

bestander:
**Summary**

Fixes issue revealed in #2819

**Test plan**

Added test


SimenB:
Interesting that the repro from #2819 still logs a warning in 0.23.2. What is its purpose?

`yarn init -y && yarn add sequelize-cli && yarn add gulp --dev && yarn --prod && yarn check --verify-tree --prod`

I find the 2 last commands interesting

```
yarn install v0.23.2
warning Integrity check: Flags don't match
[2/4] 🚚  Fetching packages...
✨  Done in 2.09s.
yarn check v0.23.2
success Folder in sync.
✨  Done in 1.01s.
```

Why complain about "Flags don't match" when it's not wrong? What actions can I as a user do? And what does it really mean?

bestander:
Good point, @SimenB.

So let me explain:
```
yarn init -y && yarn add sequelize-cli
```
adds  `sequelize-cli` to dependencies in non production mode

```
yarn add gulp --dev
```
adds  `gulp` to dev dependencies in non production mode

```
yarn --prod 
```
installs dependencies in production mode.
Because previous install (add gulp) was done in non production mode the integrity check does not pass and Yarn proceeds with doing a full installation.

But I think you are right, there should not be a warning, I added it for information purposes when running `yarn check --integrity`.
I think integrity checker should not [report](https://github.com/yarnpkg/yarn/blob/master/src/integrity-checker.js#L169), check.js command should do that.

SimenB:
Ah, OK. Thanks for the explanation 😄

PR #3083, opened 2017-04-09
Automatically update Homebrew package when new versions are released

### LINKED ISSUE #2841: Add updating brew as part of the release process
(the template doesn't really apply, and rfc seems way overkill. I can do it if you really want one).

I would really appreciate it if a PR to the brew repo was included as part of the release process for stable versions.

It's very easy, just run a single command, and submit the PR. https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md#submit-a-version-upgrade-for-the-foo-formula. It'll come out like this: https://github.com/Homebrew/homebrew-core/pull/10432

Daniel15:
**Summary**
Adds a script to update Homebrew using their `brew-formula-pr` command. This runs fine on Linuxbrew. Also adds a Jenkins job for it.

**Test plan**
Manually tested the script with the origin set to a test repo: https://github.com/Daniel15/homebrew-core/pull/2

cc @SimenB 

Closes #2841

### DIFF
--- a/scripts/update-homebrew.sh
+++ b/scripts/update-homebrew.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Pushes the latest Yarn version to Homebrew
+
+set -ex
+
+version=`curl --fail https://yarnpkg.com/latest-version`
+
+# Ensure Linuxbrew is on the PATH
+PATH=$PATH:$HOME/.linuxbrew/bin/
+# Ensure homebrew-core is pointing to Homebrew rather than Linuxbrew
+pushd ~/.linuxbrew/Library/Taps/homebrew/homebrew-core
+#git remote set-url origin https://github.com/Daniel15/homebrew-core # for testing
+git remote set-url origin https://github.com/homebrew/homebrew-core
+# Remove any existing branch (eg. if the previous attempt failed)
+git branch -D yarn-$version || true
+popd
+
+# Grab latest Yarn release so we can hash it
+url=https://yarnpkg.com/downloads/$version/yarn-v$version.tar.gz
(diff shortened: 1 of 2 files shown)


SimenB:
Awesome that you could make the command work!

scripts/update-homebrew.sh
```diff
@@ -0,0 +1,26 @@
...
+version=`curl --fail https://yarnpkg.com/latest-version`
+
+# Ensure Linuxbrew is on the PATH
+PATH=$PATH:$HOME/.linuxbrew/bin/
+# Ensure homebrew-core is pointing to Homebrew rather than Linuxbrew
+pushd ~/.linuxbrew/Library/Taps/homebrew/homebrew-core
+#git remote set-url origin https://github.com/Daniel15/homebrew-core # for testing
+git remote set-url origin https://github.com/homebrew/homebrew-core
+# Remove any existing branch (eg. if the previous attempt failed)
+git branch -D yarn-$version || true
```
SimenB: this should switch to master branch (and pull) before deleting the branch, right?

EDIT: It automatically branches from origin/master (https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L299), but deleting the branch would fail if you are on it, so switching and at least doing `git fetch --all` might be a good idea 😄

Daniel15: Good point. It should always be on `master` when the script is ran, as `bump-formula-pr` switches to `master` after pushing the PR. I'll probably change it in a followup PR but I think this is okay for now.

Daniel15:
Merging this, we can iterate as needed.

PR #3086, opened 2017-04-10
Use ES6 import syntax everywhere

Daniel15:
**Summary**
Use ES6 module syntax everywhere, rather than a mix of CommonJS (`require()`) and ES6 modules. Currently there's no difference, but this will allow us to use more advanced optimizations such as tree shaking.

Most of these were find and replace (replace `const ([A-Z0-9_]+) = require\('([^']+)'\);` with `import $1 from '$2';`) with some manual tweaking.

**Test plan**
`yarn run build`, no build errors
`node scripts\build-webpack.js` then run `node "artifacts\yarn-0.24.0-0.js"`, no errors.

### DIFF
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -8,12 +8,13 @@ import * as network from './network.js';
 import map from '../util/map.js';
 
 import typeof * as RequestModuleT from 'request';
+import request from 'request';
 import type RequestT from 'request';
 
-const RequestCaptureHar = require('request-capture-har');
-const invariant = require('invariant');
-const url = require('url');
-const fs = require('fs');
+import RequestCaptureHar from 'request-capture-har';
+import invariant from 'invariant';
+import url from 'url';
+import fs from 'fs';
 
 const successHosts = map();
 const controlOffline = network.isOffline();
@@ -195,7 +196,6 @@ export default class RequestManager {
 
   _getRequestModule(): RequestModuleT {
     if (!this._requestModule) {
-      const request = require('request');
       if (this.captureHar) {
         this._requestCaptureHar = new RequestCaptureHar(request);
         this._requestModule = this._requestCaptureHar.request.bind(this._requestCaptureHar);
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -14,7 +14,7 @@ import {getPackageVersion, explodeLockfile, runInstall, createLockfile} from '..
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
 
-let request = require('request');
+const request = require('request');
 const semver = require('semver');
 const path = require('path');
 const stream = require('stream');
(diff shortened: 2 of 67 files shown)


src/util/request-manager.js
```diff
@@ -195,7 +196,6 @@ export default class RequestManager {
 
   _getRequestModule(): RequestModuleT {
     if (!this._requestModule) {
-      const request = require('request');
```
Daniel15: This doesn't need to be inlined as we're already inlining all imports via `babel-plugin-transform-inline-imports-commonjs`.

arcanis:
Thanks! Can you fix the missing semicolon in `request-handler.js`? The breaking tests don't look related to your PR, I'll investigate.

src/util/request-manager.js
arcanis: I think this require has been inlined not because of lazy import, but rather so that it can be mocked by Jest (if hoisted to the top of the file, each invocation will reuse the same `request` instance than the previous ones, ignoring any new mock registered through Jest).

I believe it's the root cause for your failing tests.

src/util/request-manager.js
```diff
@@ -8,12 +8,13 @@ import * as network from './network.js';
 import map from '../util/map.js';
 
 import typeof * as RequestModuleT from 'request';
+import request from 'request'
```
arcanis: nit: [semicolon](https://www.youtube.com/watch?v=M94ii6MVilw) (but don't fix it, cf below)

Daniel15:
I'll take a look later today. It's likely that we're just not enabling the
inline requires transform when running tests, which would fix it

bestander:
Finally!
@zertosh, didn't you say that our webpack build of yarn.js makes all ES6 imports automatically lazy?
So this should make more modules lazy init?

Daniel15:
>  our webpack build of yarn.js makes all ES6 imports automatically lazy?

Yeah `babel-plugin-transform-inline-imports-commonjs` inlines them.

I've been trying to get Yarn compiling with Rollup. So far the bundle is quite a bit smaller (since Rollup has more advanced "tree shaking" than Webpack), I just haven't gotten the resulting bundle to actually run properly yet 😛  Moving from `require()` to `import` was one of the things I needed to change.

bestander:
Love it!

Daniel15:
I think you're right that the test failures are due to not inlining the `request` module. However, `transform-inline-imports-commonjs` should be doing that 😕  I'll take a look later in the week when I get more free time. This PR isn't urgent so it can sit around for a while :)

Daniel15:
Hmm tests are failing in CI but they worked on my computer... will take a closer look.

bestander:
@Daniel15 rebase it or close it :)

Daniel15:
Yeah, I'll get back to this at some point 😛  Seems like there's a bunch of merge conflicts so I'll probably just do it again from scratch.

bestander:
Then let's close it to keep the dynamics of low number of open issues?

Daniel15:
Sure

Daniel15:
Added this to my personal todo list, I'll get back to it eventually. So many things I want to get done.

bestander:
I'll open an issue for you here :)

bestander:
https://github.com/yarnpkg/yarn/issues/3394

PR #3092, opened 2017-04-10
Fix `yarn pack` to always include the file in the "main" field

josephfrazier:
**Summary**

This is for compatibility with npm, which [specifies] that:

> Certain files are always included, regardless of settings:
> * package.json
> * README
> * CHANGES / CHANGELOG / HISTORY
> * LICENSE / LICENCE
> * NOTICE
> * The file in the "main" field

[specifies]: https://docs.npmjs.com/files/package.json#files

**Test plan**

I changed the `files-include-mandatory` test to exclude "index.js" from
`files`, and ensure that it still gets included, since it is the `main`
file.

### DIFF
--- a/__tests__/fixtures/pack/files-include-mandatory/package.json
+++ b/__tests__/fixtures/pack/files-include-mandatory/package.json
@@ -3,5 +3,5 @@
   "version": "1.0.0",
   "main": "index.js",
   "license": "MIT",
-  "files": ["index.js", "a.js", "b.js"]
+  "files": ["a.js", "b.js"]
 }
--- a/src/cli/commands/pack.js
+++ b/src/cli/commands/pack.js
@@ -68,13 +68,16 @@ function addEntry(packer: any, entry: Object, buffer?: ?Buffer): Promise<void> {
 
 export async function pack(config: Config, dir: string): Promise<stream$Duplex> {
   const pkg = await config.readRootManifest();
-  const {bundledDependencies, files: onlyFiles} = pkg;
+  const {bundledDependencies, main, files: onlyFiles} = pkg;
 
   // inlude required files
   let filters: Array<IgnoreFilter> = NEVER_IGNORE.slice();
   // include default filters unless `files` is used
   if (!onlyFiles) {
     filters = filters.concat(DEFAULT_IGNORE);
+  }
+  if (main) {
+    filters = filters.concat(ignoreLinesToRegex(['!/' + main]));
   }
 
   // include bundledDependencies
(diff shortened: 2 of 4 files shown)


__tests__/fixtures/pack/files-include-mandatory/package.json
```diff
@@ -3,5 +3,5 @@
   "version": "1.0.0",
   "main": "index.js",
   "license": "MIT",
-  "files": ["index.js", "a.js", "b.js"]
+  "files": ["a.js", "b.js"]
```
bestander: could you add index.js to main filed?

josephfrazier:
Hi @bestander, thanks for the quick review! Could you elaborate on your comments? I'm not sure I understand what needs to be changed, as index.js is already in the main field of the test fixture ([here](https://github.com/yarnpkg/yarn/pull/3092/files#diff-7694b9952a48da4deff6ab80f5149dacR4)), and it is tested to be present in the tarball [here](https://github.com/yarnpkg/yarn/pull/3092/files#diff-5ad8083fc2b0c87715111de283efecbcR133).

src/cli/commands/pack.js
```diff
@@ -68,14 +68,17 @@ function addEntry(packer: any, entry: Object, buffer?: ?Buffer): Promise<void> {
...
-  const {bundledDependencies, files: onlyFiles} = pkg;
+  const {bundledDependencies, main, files: onlyFiles} = pkg;
 
   // inlude required files
   let filters: Array<IgnoreFilter> = NEVER_IGNORE.slice();
   // include default filters unless `files` is used
   if (!onlyFiles) {
     filters = filters.concat(DEFAULT_IGNORE);
   }
+  if (main) {
```
bestander: sure, @josephfrazier.
You've added a new condition checking for `main` in package.json.
But have not added a test that covers this condition.
In `__tests__/fixtures/pack/files-include-mandatory/package.json` you probably wanted to add `main` field with `index.js` and not have it in `files`?

josephfrazier: Hmm, still not sure I understand. In `__tests__/fixtures/pack/files-include-mandatory/package.json`, the `main` field is already `index.js`. The point of this change is to make sure that `index.js` *doesn't* have to be in `files` in order for the packed tarball to contain `index.js`.

I split my changes into two commits - test first, fix second - which should help demonstrate how the new condition is covered by the test. If you run the following at each commit, you'll see how the tarball changes (this is also covered by `yarn test`, but I pushed the two commits too quickly, and the CI for the first commit seems to have run against the changes in the second commit):

```
yarn build && cd __tests__/fixtures/pack/files-include-mandatory && ../../../../bin/yarn.js pack && tar tvf *tgz | grep index.js; cd -
```

bestander:
Thanks, looks great

Issue #1961, opened 2016-11-21
EINVAL: invalid argument, symlink


atrauzzi:
So I'm getting this error, which I suspect I already know the reason for:

```
Arguments: 
  /usr/bin/nodejs /usr/bin/yarn add react-redux

PATH: 
  /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

Yarn version: 
  0.17.6

Node version: 
  6.9.1

Platform: 
  linux x64

Trace: 
  Error: EINVAL: invalid argument, symlink '../../../loose-envify/cli.js' -> '/MYPROJECT/node_modules/invariant/node_modules/.bin/loose-envify'
      at Error (native)
```

The reason this is happening is because I'm using Docker for Windows and the path it's looking to create a symlink on is actually backed by an NTFS filesystem.  This is going to be fairly common for anyone trying to use yarn through a Docker container on Windows.

I'm not sure what the best course of action would be here, but I don't mind even just having to tell yarn to skip doing symlinks.  You might however have an opportunity here to make the user experience a little nicer by detecting when this happens and either raising a warning, or improving the error feedback.

Anyway, this is a thing :)

nicotroia:
+1

meta-meta:
bump

adiachenko:
Interestingly, `npm install` works just fine on the latest Beta version of Docker for Windows. I am not a big fan of falling back to npm though.

meta-meta:
I'm currently working around this with `yarn install --no-bin-links` then `yarn global add` for things needed on the command line like `webpack` and `webpack-dev-server` since that puts the modules on the containers filesystem.

adiachenko:
@meta-meta Well, you could do that, but it requires changes to docker images that are tightly coupled with your codebase.

meta-meta:
@adiachenko I don't love it, but it works for now. Hoping this goes away soon.

mlgamble:
bump - Exact same situation as OP

cascornelissen:
I'm running into the same issue... The workaround provided by @meta-meta seems to work but that doesn't really work for my use case either...

kvahuja:
I am running into the same issue when installing create-react-app. any clue what to do to fix this?

FDiskas:
:spaghetti: same problem

novascreen:
I've created a small script that writes the bin links for all dependencies in `package.json`.
[...]

PR #3097, opened 2017-04-11
Remove concurrency in creating bin links

whitecolor:
Fix for #1961

This should probably not have any visible effect on the performance of the installation process.

As [stated in the comment](https://github.com/yarnpkg/yarn/issues/1961#issuecomment-293060324) *sometimes* `yarn ` produces the error while installing deps on docker volume, like:

```
EINVAL: invalid argument, symlink '../../../acorn/bin/acorn' -> '/app/node_modules/falafel/node_modules/.bin/acorn
```

This happens when there is need to create multiple bin links and cuncurrency in creating symlinks causes the error. **Probably the problem is specific for [docker for windows](https://docs.docker.com/docker-for-windows/install/)**.

Tried to test if it fails with making many symlinks in parallel inside this node_modules, but simple case seem doesn't show any problems, that means that something related to specific yarn's linking flow and need to be investigated further.

Steps to reproduce:

On windows 10:
1) Install [docker for windows](https://docs.docker.com/docker-for-windows/install/)
2) `> docker run --rm -v "D:\TEMP\yarn_docker_test:/app" -it mhart/alpine-node:7.8.0 sh`

Insider docker container:

3) `# npm i yarn -g` 
4) `# cd /app`
5) `# yarn init`
6) `# yarn add localforage` 

At least installing `localforage` causes error for me, may need some more packages with bin links if your machine manages to passes it.

To make sure fix works:

7) `# sed -i -e 's/})(), 4);/})(), 1);/g' /usr/lib/node_modules/yarn/lib/package-linker.js`
8) `# yarn add localforage`


bbrala:
Just to add to this point. This happens on all SMB mount from Linux to windows, even with `mfsymlink`. With this fix it works without a problem.

```
vagrant@dev:/home/localcopy/yarntest $ find -L ./ -xtype l

./npm/node_modules/.bin/gulp
./npm/node_modules/.bin/mkdirp
./npm/node_modules/.bin/semver
./npm/node_modules/.bin/strip-bom
./npm/node_modules/.bin/user-home
./npm/node_modules/.bin/which
./yarn/node_modules/.bin/gulp
./yarn/node_modules/global-prefix/node_modules/.bin/which
./yarn/node_modules/gulp/node_modules/.bin/semver
./yarn/node_modules/v8flags/node_modules/.bin/user-home
./yarn/node_modules/vinyl-fs/node_modules/.bin/mkdirp
./yarn/node_modules/vinyl-fs/node_modules/.bin/strip-bom
```

bestander:
Any idea why this fails in the first place?
Surely creating multiple links in parallel is not an issue, could it be trying to create the same link twice in parallel?
I am fine with mitigating an issue but I would not want to loose track of the real issue there.

src/package-linker.js
```diff
@@ -267,7 +267,7 @@ export default class PackageLinker {
         const binLoc = path.join(dest, this.config.getFolder(pkg));
         await this.linkBinDependencies(pkg, binLoc);
         tickBin(dest);
-      }, 4);
+      }, 1);
```
bestander: If we go with this, this line needs a comment why the concurrency is being reduced.

whitecolor:
@bestander I have narrowed down the case situation and created an issue in docker for win repo https://github.com/docker/for-win/issues/631.

bbrala:
I really think this is not really related to docker but to how a symlink is made on a smb share since vagrant using a smb share shows the same behaviour. 

@bestander has a point though, wonder what logging the calls would produce...

whitecolor:
> wonder what logging the calls would produce...

What calls? It tries to concurrntly create links in different directories:
```
fsSymlink ../../../esprima-fb/bin/esparse.js /app/node_modules/es3ify/node_modules/.bin/esparse
fsSymlink ../../../acorn/bin/acorn /app/node_modules/falafel/node_modules/.bin/acorn
fsSymlink ../../../esprima-fb/bin/esvalidate.js /app/node_modules/es3ify/node_modules/.bin/esvalidate
fsSymlink ../../../esprima-fb/bin/esparse.js /app/node_modules/jstransform/node_modules/.bin/esparse
fsSymlink ../esprima-fb/bin/esparse.js /app/node_modules/recast/node_modules/.bin/esparse
fsSymlink ../esprima-fb/bin/esvalidate.js /app/node_modules/recast/node_modules/.bin/esvalidate
fsSymlink ../../../esprima-fb/bin/esvalidate.js /app/node_modules/jstransform/node_modules/.bin/esvalidate
error An unexpected error occurred: "EINVAL: invalid argument, symlink '../../../esprima-fb/bin/esvalidate.js' -> '/app/node_modules/jstransform/node_modules/.bin/esvalidate'".
```
By logging calls I've narrowed the issue down and described how it can be reproduces without yarn.

bbrala:
It seems to break in trying to symlink the same file twice to a different location. In your example i see the same  esvalidate.js linked twice. When i just logged it in my VM it seemed to break also on linking the same file twice. Perhaps that is a clue 😊

bestander:
So it fails on Windows when symlinking the same file into different locations in parallel?
I guess we don't get much performance gain for symlinking in parallel but it would be better to find a more generic fix.

bestander:
Any chance this is related to #2795 and can be fixed the same way?

whitecolor:
@bestander I've checked and this really solves the issue, the problem is how to determine that container is actually running under windows with SMB. Currently it is applied only for `process.platform === 'win32'`

bbrala:
Personally I would say for compatibility reason i would just run it without concurrency. I think it's quite hard to find out if a folder is mounted as SMB, this would mean checking the current directory and anything higher against the output of the mounts on the system. That seems  a bit excessive.

whitecolor:
I would go without concurrency, It is kind of unlikely to make installations time any faster. Besides why current concurrency is set to 4? Why not to 100?

bestander:
Yeah, probably not really worth the effort to identify SMB specifically.
How about moving "4" to a constant in this class and set it to `process.platform === 'win32' ? 1 : 4`?

Also a comment why win32 is single threaded could be enough.

bestander:
As for 4 vs 10, I guess 4 was chosen empirically.
Going more threads usually does not get much performance gain but less can be slower.
I had run tests for concurrent .tgz downloads in TarballFetcher and the optimal number was 4-8.

bbrala:
The platform won't identify as win32 if it's a mount inside Linux i think. So that won't work to separate the two.

bestander:
Are you sure?
I thought that https://nodejs.org/api/process.html#process_process_platform is about the process executable, not file system

whitecolor:
@bestander of cource, container (or VM) is fully isolated envrionment from HOST system, the case is that you run linux container on windows HOST (it even runs actually on linux VM in case of docker, but volumes are attached from windows host via SMB).

bestander:
Oh snap, now I see.
Yeah,
That makes it harder but seems like a very rare edge case.
How about having this as a setting in .yarnrc for now?

whitecolor:
> How about having this as a setting in .yarnrc for now?

I think this would be ok.

bbrala:
This is not really an edge case imo. With hyperv getting bigger for isolated development env's on Windows this will only get worse imo.

I can't currently run some performance tests but if the performance hit is minimal i really don't see a reason not to change concurrency. When I get back home I could run some tests to see how this change impacts performance to be sure.

bestander:
This needs:
- rebasing
- moving 1 to constant and a comment why 1 is default so that people maintaining the code would have the context

Vanuan:
> like a very rare edge case

Not rarer than using yarn instead of npm ;)

Vanuan:
I think you can reproduce this in linux too. Just mount SMB folder with sources.

Vanuan:
Why mutex is not used on linux (`linkBin` function)?

DanReyLop:
So, @Vanuan discovered that #2795 didn't fix anything at all: https://github.com/yarnpkg/yarn/pull/2795#issuecomment-301623029

> Why mutex is not used in linux (linkBin function)? It doesn't look like it will hurt.

That's because `symlink` in Linux is supposed to be an atomic operation (as opposed to `cmd-shim` which deletes and then creates a file). But if it's buggy on Docker for Windows, then it doesn't hurt to use it always. It would be certainly more performant than removing concurrency.

So, I propose keeping the concurrency, and instead:
1. Changing `await lockMutex(src)` to `await lockMutex(dest)` in the function `linkBin`.
2. Use the mutex always, not only if `process.platform === 'win32'`.

How does that sound?

Vanuan:
But is _changing_ a symlink really an atomic operation?
Because that is what looks to be happening here:
http://blog.moertel.com/posts/2005-08-22-how-to-change-symlinks-atomically.html

Vanuan:
As a side note, the only atomic operation that works on both network shares and non-standard systems appears to be `mkdir`: https://www.npmjs.com/package/proper-lockfile

DanReyLop:
> But is changing a symlink really an atomic operation?
Because that is what looks to be happening here:

Oh well, my knowledge doesn't reach that layer haha. But it *is* a single NodeJS API call, and NodeJS is single-threaded, so in the context of a NodeJS process, it *should* be atomic. At least for the purposes of this issue.

Anyway, since it looks like the Docker for Windows symlink implementation is buggy, I would be OK using always the `mutex`, even if the underlying operation is "atomic enough" :)

Vanuan:
> But it is a single NodeJS API call, and NodeJS is single-threaded

The main thread (where user's code is executed) is single-threaded, but there are also IO threads. All async IO operations use threads. Otherwise there wouldn't be any performance gain from the asynchronous nature of NodeJS.

image

Do you see those worker threads?

bbrala:
I just want to state we should remember this is an SMB problem not a docker one :)

Vanuan:
Yeah, I wouldn't say that Docker for Windows symlink implementation (which is actually SMB implementation underneath) is buggy. It's just slower enough to experience a race condition.
You just don't experience that race condition in linux, since unlink/symlink calls are faster than time between `linkBin` calls. But if filesystem is slow enough (for example, on HDD) and exactly enough, you could reproduce without docker too.
That's my theory at least.

bestander:
Thanks, let's have this patch in.
If anyone has an idea for a more generic way to solve it, your help would be welcome.

Naninani:
Is there any update on this issue?
I still experience that problem with yarn 1.3.2

`error An unexpected error occurred: "EINVAL: invalid argument, symlink '../../../angucomplete-alt.js' -> '/node_modules/angucomplete-alt/example/js/libs/angucomplete-alt.js'".`

bestander:
It has been merged since 0.x versions, probably the one you see is related but different

Vanuan:
Yeah, this one is about bin links. The one you have is about source files.
I would grep for fs.symlink and look if there's symilar protection as implemented here:
https://github.com/yarnpkg/yarn/search?utf8=%E2%9C%93&q=fs+symlink&type=

PR #3101, opened 2017-04-11
Use an alias system instead of camelCase function

voxsim:
**Summary**
In this pull request I try to remove the camelCase function from src/cli/commands/index.js and
src/cli/commands/help.js and instead I use an alias system.
I move the existing aliases to unsupported-aliases.
We can't still remove the entire dependency `camelCase` because global command use it.
Finally I move every command-related stuff in src/cli/commands/index.js.

**Test plan**
- tests camelised comand
- tests command with hyphen
- tests every case above and some weird cases with --help option


voxsim:
@arcanis I rebased after the last commits ;)

src/cli/commands/index.js
```diff
@@ -1,49 +1,88 @@
...
+import * as remove from './remove.js'; commands['remove'] = remove;
+import * as run from './run.js'; commands['run'] = run;
+import * as tag from './tag.js'; commands['tag'] = tag;
+import * as team from './team.js'; commands['team'] = team;
+import * as unlink from './unlink.js'; commands['unlink'] = unlink;
+import * as upgrade from './upgrade.js'; commands['upgrade'] = upgrade;
+import * as version from './version.js'; commands['version'] = version;
+import * as versions from './versions.js'; commands['versions'] = versions;
+import * as why from './why.js'; commands['why'] = why;
+import * as upgradeInteractive from './upgrade-interactive.js'; commands['upgradeInteractive'] = upgradeInteractive;
```
arcanis: It might be clearer to do something like this:

```js
import * as bar from './bar';
import * as foo from './foo';
...

let commands = {
    bar,
    foo,
    ...
};
```

What do you think?

__tests__/index.js
```diff
@@ -211,20 +211,26 @@ test.concurrent('should not output JSON activity/progress if given --no-progress
...
 
-test.concurrent('should display correct documentation link for aliases', async () => {
+test.concurrent('should display correct documentation link for unsopported aliases', async () => {
   await expectAnInfoMessageAfterError(
     execCommand('i', [], 'run-add', true),
     'Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.',
   );
 });
 
+test.concurrent('should show help and ignore unsopported aliases', async () => {
```
arcanis: `unsopported`? :p

(same typo on 214 and 221 :)

voxsim: aaaah thanks and sorry XD

src/cli/commands/index.js
voxsim: I think is A LOT clearer; i will change it!

arcanis:
Nice work! Just a typo to fix and I merge this 👍 

I really feel like github should offer a way to edit PRs from the review interface, for cases like this ...

voxsim:
@arcanis done!

PR #3103, opened 2017-04-11
Do not test new version with pkg.version

binarykitchen:
See issue 3011 for details. Leave it to npm registry to validate the version number.

Motivation is to make it work with the way how I deploy my videomail-client library at
https://github.com/binarykitchen/videomail-client/blob/develop/bin/release.sh

There I am already bumping to the new version in package.json before publishing it with yarn to npm.

**Test plan**

1. Manually increment version in a random package.json
2. Run `yarn publish --new-version with that new version`
3. Yarn won't complain anymore that the new version is the same.
4. It will be up to the npm registry whether to accept that version or not.

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -90,7 +90,9 @@ export async function setVersion(
   invariant(newVersion, 'expected new version');
 
   if (newVersion === pkg.version) {
-    throw new MessageError(reporter.lang('publishSame'));
+    return function(): Promise<void> {
+      return Promise.resolve();
+    };
   }
 
   await runLifecycle('preversion');
--- a/__tests__/commands/version.js
+++ b/__tests__/commands/version.js
@@ -39,6 +39,14 @@ test('run version with no arguments and --new-version flag', (): Promise<void> =
   });
 });
 
+test('run version with no arguments, --new-version flag where version is same as pkg.version', (): Promise<void> => {
+  return runRun([], {newVersion, gitTagVersion}, 'no-args-same-version', async(config, reporter): ?Promise<void> => {
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+
+    expect(pkg.version).toEqual(newVersion);
+  });
+});
+
 test('run version and make sure all lifecycle steps are executed', (): Promise<void> => {
   return runRun([], {newVersion, gitTagVersion}, 'no-args', async(config): ?Promise<void> => {
     const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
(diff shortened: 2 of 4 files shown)


voxsim:
hey @binarykitchen thanks for your pr, could you please write two tests for both cases instead of leaving a comment?

binarykitchen:
not sure how this can be tested when it's up to the npm registry to validate version numbers ...

binarykitchen:
@voxsim ok, have code to cover the first test case

but there are no tests at all for the `yarn publish` command, so cannot add the second test case. maybe make an empty file for now with todo comments and raise this as a separate issue?

voxsim:
@binarykitchen maybe for now it is sufficient. Let @arcanis and @bestander decide if we need to do other stuff ;)

src/cli/commands/version.js
```diff
@@ -89,9 +89,11 @@ export async function setVersion(
...
   invariant(newVersion, 'expected new version');
 
-  if (newVersion === pkg.version) {
-    throw new MessageError(reporter.lang('publishSame'));
-  }
+  // we dont do a check whether the new version is the same as the one in package.json
+  // because of two reasons (see issue #3011):
+  // - npm registry will complain anway when it already exist (leave it to them to validate that)
+  // - some open source libraries/projects increment version just before deployment where
+  //   the new version already has been bumped in package.json
```
arcanis: Not sure this comment is needed.

However, shouldn't we just return early when the package version is the same as the version currently registered? Instead of running lifecycle hooks etc. Not doing this could cause issues if, for example, a `version` lifecycle script was registering a mercurial tag.

binarykitchen: true that @arcanis 

... just have updated PR. now it's returning early and comment is much shorter

arcanis:
👍 Would you mind also adding a `--no-version` option on the `publish` command to completely skip the version step?

binarykitchen:
@arcanis i think that would raise scope and should got in a separate issue. my commit here only touches the version command. or am i missing something here?

binarykitchen:
@arcanis ping?

arcanis:
I'm ok with merging this! Just a nit: can you remove the comments? Such a small change is relatively self-explanatory, so no need to write in english what's already written in js :)

binarykitchen:
@arcanis ok, all comments are removed - good now?

bestander:
Great job, everyone!

binarykitchen:
thanks - is there an eta when and in which version this will come out @bestander?

bestander:
0.24 next week

PR #3105, opened 2017-04-11
Ensure that every command has setFlags and hasWrapper functions

voxsim:
**Summary**
This pull request is a simplified version of https://github.com/yarnpkg/yarn/pull/2810 and maybe for some ways better because it prefers composition over inheritance and it is less invasive right now.
There are two commits that ensure that we always have setFlags and hasWrapper functions.
This is the first step to move some responsabilities from the index towards the command classes.

**Test plan**
For what I see every command is covered, but let me know if I have to cover something.


voxsim:
Mmh many commands are not covered by tests and maybe I forget some functions; I will open other pull request to raise the coverage.

arcanis:
👍

PR #3112, opened 2017-04-11
supported smaller package metadata API

bestander:
**Summary**

According to https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md we can now use smaller metadata API that does not have all the unnecessary descriptions etc.

**Test plan**

React Native:

| packages | metadata before change | metadata after change |
| --- | --- |--- |
| 26MB | 5.46MB | 3.31MB |

material2:

| packages | metadata before change | metadata after change |
| --- | --- |--- |
| 70MB| 8.89MB | 5.41MB |

Results:

This PR reduces metadata downloads by 40%.
Full metadata to packages is about 10-20% of total download.

There should be some minor improvements for Yarn memory now that it does not need to pass around the extra meta data.


bestander:
For review: avoid the last commit

cpojer:
Nice!

src/registries/npm-registry.js
```diff
@@ -57,7 +57,9 @@ export default class NpmRegistry extends Registry {
       || this.getOption('always-auth')
       || removePrefix(requestUrl, registry)[0] === '@';
 
-    const headers = {};
+    const headers = {
+      'Accept': 'application/vnd.npm.install-v1+json'
+    };
```
cpojer: Does that count as one line? :P

PR #3113, opened 2017-04-12
Correctly handle installationMethod when using bundled build

### LINKED ISSUE #3085: installationMethod not updated properly in master
Bug

Yarn's `package.json` contains an `installationMethod` key for determining the installation method that was used to install Yarn. This is used so we know which command to suggest to upgrade Yarn.

This used to work correctly as `package.json` was sitting alongside Yarn itself and it did `require('./package.json')`, but now that the standalone .js file is included in the tarball, `package.json` is bundled into the JS file and thus updating `package.json` doesn't work to update the installationMethod. Potentially Yarn should still load `package.json` (but be careful to not break running the standalone .js file separately.

Daniel15:
**Summary**
We do `require('package.json')` in a few places to get the version number and installation method of Yarn. In the Webpack build, this is embedded within the JS file itself. While this works fine for the version number (as that's constant), the installation methods can be different even with an identical build of Yarn. For example, Homebrew takes our tarball and then patches the `installationMethod` in `package.json` so we can tell that Yarn was installed via Homebrew. Using `require('package.json')` in this case won't reflect the update as it's only seeing package.json at the time of build (rather than at runtime).

Instead, we need to actually read the `package.json` file to check for the installationMethod at runtime. This looks in the parent directory (`../package.json`) due to the directory structure:
```
\package.json  <--- This is the file we want
\lib
  \yarn-cli.js <--- This is where the code is
\bin
  \yarn
  \yarn.cmd
```

**Test plan**
`.\bin\yarn --version`, ensure it says `0.24.0-0`
`.\bin\yarn install`. ensure heading says "Yarn install v0.24.0-0"
`.\bin\yarn versions`, ensure 
`.\scripts\build-webpack.js` then do all the same tests with `.\artifacts\yarn-0.24.0.js`

Closes #3085

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -8,6 +8,7 @@ import * as network from '../util/network.js';
 import {MessageError} from '../errors.js';
 import Config from '../config.js';
 import {getRcArgs} from '../rc.js';
+import {version} from '../util/yarn-version.js';
 
 const commander = require('commander');
 const fs = require('fs');
@@ -17,7 +18,6 @@ const loudRejection = require('loud-rejection');
 const net = require('net');
 const onDeath = require('death');
 const path = require('path');
-const pkg = require('../../package.json');
 
 loudRejection();
 
@@ -29,7 +29,7 @@ const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length 
 const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
 // set global options
-commander.version(pkg.version);
+commander.version(version);
 commander.usage('[command] [flags]');
 commander.option('--verbose', 'output verbose messages on internal operations');
 commander.option('--offline', 'trigger an error if any required dependencies are not available in local cache');
@@ -143,7 +143,7 @@ const config = new Config(reporter);
 const outputWrapper = !commander.json && command.hasWrapper(commander, commander.args);
 
 if (outputWrapper) {
-  reporter.header(commandName, pkg);
+  reporter.header(commandName, {name: 'Yarn', version});
 }
 
 if (command.noArguments && commander.args.length) {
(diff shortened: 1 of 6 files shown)


src/util/yarn-version.js
```diff
@@ -0,0 +1,28 @@
...
+
+  // If there's a package.json in the parent directory, it could have an
+  // override for the installation method, so we should prefer that over
+  // whatever was originally in Yarn's package.json. This is the case with
+  // systems such as Homebrew, which take the tarball and modify the
+  // installation method so we're aware of the fact that Yarn was installed via
+  // Homebrew (so things like update notifications can point out the correct
+  // command to upgrade).
+  const manifestPath = path.join(__dirname, '..', 'package.json');
+  if (fs.existsSync(manifestPath)) {
```
bestander: why sync?
We use async/await in all the other places, don't we?

src/cli/index.js
```diff
@@ -171,7 +171,7 @@ if (commander.json) {
   outputWrapper = false;
 }
 if (outputWrapper) {
-  reporter.header(commandName, pkg);
+  reporter.header(commandName, {name: 'Yarn', version});
```
bestander: I think this line broke a test

src/util/yarn-version.js
Daniel15: The async version of `existsSync` is deprecated.

cli/index.js also uses `existsSync` and `readFileSync`: https://github.com/yarnpkg/yarn/blob/45af6562c22652404993fb1d77616fadf3e5eff2/src/cli/index.js#L296-L297

Having said that, I can change reading the file to be async by creating a `getInstallationMethod` async function and using that where needed.

src/cli/index.js
Daniel15: Oops - I thought I ran the tests locally, but I might have only run the linter :/ I'll fix it.

bestander:
Hey @Daniel15, this needs a rebase

Daniel15:
Sure, I'll do that soon :)

Daniel15:
Rebased, I'll merge once I verify that the tests pass.

PR #3117, opened 2017-04-12
track missing request caches

bestander:
Improves CI stability by committing request caches, apparently we have not done it for a while.

Summary:
- Added a warning log when there is not request cache
- regenerated request cache on my mac

To Do:
- generate request caches from Linux and Windows machines (I'll send a new PR from my home computer)


bestander:
It gets generated automatically.
I just removed it from .gitignore.

The down side is that from time to time people will have new files in that cache after running tests and it confuses them.

Issue #960, opened 2016-10-13
Optional emoji support


daltones:
**Do you want to request a *feature* or report a *bug*?**
*Feature* (maybe a *bug*, once one of the features in README is “More emojis. 🐈” 😸)

**What is the current behavior?**
Emojis only for OSX

**What is the expected behavior?**
Emojis for other platforms

**Please mention your node.js, yarn and operating system version.**
node.js: v6.7.0
yarn: v0.15.1
OS: Arch Linux

This is sort of related to #405 and 9f58576d127b34aadbb0c5b89e506faf503b3e34.

As discussed in #405, emojis on Powershell are bad. But it could be good someplace else…

I tried to force enable emojis here (Gnome Terminal on Arch Linux) and worked very well (although limited to monochrome). That's because I have the properly font and configuration.

![captura de tela de 2016-10-13 02-16-05](https://cloud.githubusercontent.com/assets/5603632/19337371/0ef8e748-90eb-11e6-902a-1f671f37a779.png)

So it would be nicer if we could optionally enable emojis. It could be even an environment variable `YARN_EMOJIS=1`.

GRIM2D:
*Subjective opinion*

Emojis aren't meant to be used in professional development tools. Not all people like emojis. And I'm sure not only I am here, who dislikes image usages on CLI interface.

However, if this tool can handler emojis well, it should be optionally enabled (or disabled).

daltones:
@GRIM2D yeah, your “(or disabled)” is implicitly inside my proposal of optional emojis.

I was testing yarn today and I was curious why my outputs didn't have emojis (as the documentation did). Then checking the code I found this:
```js
emoji: process.stdout.isTTY && process.platform === 'darwin'
```

That's a weird imposition. The issue isn't just a opinion about emojis.

patrys:
@GRIM2D In this context emoji are used as equivalents of icons that your usual GUI would have. Just because the Unicode spec calls them emoji does not automatically make them unprofessional.

GRIM2D:
@patrys 

> emoji does not automatically make them unprofessional

But, emojis are widely used while chatting. Also I do. And I'm not saying that using emojis make tools unprofessional. I meant CLI is mostly text-based. And mostly uses English as primary language. Even being Russian native-speaker, seeing cyrillic chars on my terminal is not comfortable for me. Same goes emojis.

[...]

PR #3126, opened 2017-04-12
Allow emojis to be force enabled with `--emoji`.

elyobo:
The previous behaviour had a default of enabling emojis for Macs, and
allowed explicitly disabling them.

This PR keeps that default but allows `--emoji` to enable for other
platforms if the user wants to.  `--no-emoji` still works as before, but
due to the way that commander processes "--no-" flags (forcing the
default to true and overwriting the default value given in the `.option`
definition) it was necessary to switch the flag definition from
`--no-emoji` to `--emoji`.

**Summary**

Allows users to enable emojis on non-darwin platforms.

Emojis are a nice feature but are explicitly disabled on non-darwin platforms.  There appears to be some interest in having them on other plaforms (see #960) and this allows that by making `--emoji` work as you'd expect (`--no-emoji` and `--emoji` are already accepted, but only `--no-emoji` works).

**Test plan**

On linux

```
# Forcing emojis off
$ node ./lib/cli/index.js --no-emoji
yarn install v0.24.0-0
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.65s.

# Default behaviour; off on non-darwin
$ node ./lib/cli/index.js 
yarn install v0.24.0-0
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.68s.

# Forcing emojis on
$ node ./lib/cli/index.js --emoji
yarn install v0.24.0-0
success Already up-to-date.
✨  Done in 0.67s.
```

Note that there is a flow type error in this PR related to `process.stdout.isTTY`, however this PR didn't add that check; it seems like it's a flow error that it wasn't already failing (I'm not familiar with flow though, so happy to be corrected here).

Another tweak would be to shift the `process.stdout.isTTY` check into the default value for `commander.emoji`, which would allow users to force emoji output even if `stdout` isn't a TTY - I'm not sure if this is ever desirable, but it's not possible to override at the moment and the change is simple.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -65,8 +65,9 @@ commander.option(
   'use a mutex to ensure only one yarn instance is executing',
 );
 commander.option(
-  '--no-emoji',
-  'disable emoji in output',
+  '--emoji',
+  'enable emoji in output',
+  process.platform === 'darwin',
 );
 commander.option(
   '-s, --silent',
@@ -131,7 +132,7 @@ commander.args.shift();
 //
 const Reporter = commander.json ? JSONReporter : ConsoleReporter;
 const reporter = new Reporter({
-  emoji: commander.emoji && process.stdout.isTTY && process.platform === 'darwin',
+  emoji: process.stdout.isTTY && commander.emoji,
   verbose: commander.verbose,
   noProgress: !commander.progress,
   isSilent: commander.silent,


elyobo:
Test failures appear to be unrelated to the changes in this PR (tests are failing in the master from which I branched this).

elyobo:
Worked around the flow problems with `isTTY`, although it still checks it (possibly an error that it's _not_ a flow error?).

bestander:
I think this will make emoji off by default, we have them on by default on Mac.

elyobo:
@bestander, this PR keeps the default behaviour as is - enabled for Mac, disabled for the rest.

```
commander.option(
  '--emoji',
  'enable emoji in output',
  process.platform === 'darwin',
);
```

The third option to `commander.option` sets the default value for the `--emoji` flag to `true` if the platform is `darwin`.

bestander:
Sorry, did not notice :)

bestander:
Thanks for the PR.
Emojis FTW!

elyobo:
NP, thanks for the merge, now I will no longer have emoji envy :D :tada:

Issue #3138, opened 2017-04-13
Yarn 0.23.2 uses Array.include which is not present in node 4


MiguelMadero:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**

yarn tries to use Array.include which is not present in node 4. 

**If the current behavior is a bug, please provide the steps to reproduce.**

1. curl -o- -L https://yarnpkg.com/install.sh | bash
2. nvm use 4.8.2
3. yarn global add bower

**What is the expected behavior?**

It should install the dependencies and avoid depending on language features/apis not available on node 4. 

**Please mention your node.js, yarn and operating system version.**

```
node --version 
v4.8.1
y~/code-zenefits/z-buttons [mmadero/test-travis|✔] 
03:37 PM $ yarn --version 
0.23.2
```

**Full log**

```
03:50 PM $ vim /Users/mmadero/.config/yarn/global/yarn-error.log 



































  stringstream@~0.0.4:
    version "0.0.5"
    resolved "https://registry.yarnpkg.com/stringstream/-/stringstream-0.0.5.tgz#4e484cd4de5a0bbbee18e46307710a8a81621878"

  strip-ansi@^3.0.0:
    version "3.0.1"
    resolved "https://registry.yarnpkg.com/strip-ansi/-/strip-ansi-3.0.1.tgz#6a385fb8853d952d5ff05d0e8aaf94278dc63dcf"
    dependencies:
      ansi-regex "^2.0.0"

  supports-color@^2.0.0:
    version "2.0.0"
    resolved "https://registry.yarnpkg.com/supports-color/-/supports-color-2.0.0.tgz#535d045ce6b6363fa40117084629995e9df324c7"

  throttleit@^1.0.0:
    version "1.0.0"
    resolved "https://registry.yarnpkg.com/throttleit/-/throttleit-1.0.0.tgz#9e785836daf46743145a5984b6268d828528ac6c"

  tough-cookie@~2.3.0:
    version "2.3.2"
    resolved "https://registry.yarnpkg.com/tough-cookie/-/tough-cookie-2.3.2.tgz#f081f76e4c85720e6c37a5faced737150d84072a"
    dependencies:
      punycode "^1.4.1"

  tunnel-agent@~0.4.1:
    version "0.4.3"
    resolved "https://registry.yarnpkg.com/tunnel-agent/-/tunnel-agent-0.4.3.tgz#6373db76909fe570e08d73583365ed828a74eeeb"

[...]

PR #3139, opened 2017-04-13
Avoids using Array.prototype.includes since it is not avaialble in no…

MiguelMadero:
Fixes https://github.com/yarnpkg/yarn/issues/3138

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -179,7 +179,7 @@ const {run, setFlags: _setFlags} = buildSubCommands('global', {
     await updateCwd(config);
 
     const updateBins = await initUpdateBins(config, reporter, flags);
-    if (args.includes('yarn')) {
+    if (args.indexOf('yarn') !== -1) {
       reporter.warn(reporter.lang('packageContainsYarnAsGlobal'));
     }
 


MiguelMadero:
The test failures seem unrelated to my changes. I checked out master and it fails locally as well :( 
Are these tests flaky? This is the first time for me running it and sending a PR. Can someone help me decipher the issue here? I manually tested this version with node 6 and global installs still work fine.

arcanis:
Thanks! I think those tests are flaky, yes. We're working on it :(

Turbo87:
@arcanis do you think it would be possible to do a quick bugfix release including this fix? right now all Node 4 builds on Travis CI that use yarn are broken because Travis will try to install the most recent version.

MiguelMadero:
@Turbo87 while I think it's important to have the latest yarn and I would also like a release with this changes, the following installation step is working for us to pin yarn to 0.22.0

```
  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.22.0
```

Turbo87:
@MiguelMadero I know, but then you'd have to explicitly add the yarn installation instructions to your TravisCI config, but Travis already automatically detects the existance of a `yarn.lock` file and will install and use yarn then unless you overwrite those instructions. I'd personally rather keep using the defaults instead of having to touch the config in every one of my projects.

MiguelMadero:
Agreed. Just a different workaround than updating the same file to use node
6.

PR #3140, opened 2017-04-13
[ci] added request caches for linux

bestander:


Daniel15:
Is it possible to put these in a separate repo and pull it in as a Git submodule? I had to download 44 MB of stuff when pulling the repo today 😛

bestander:
Yeah, this thought crossed my mind.

Let's do it next week

voxsim:
@bestander maybe we can do:
- a yarn script that download everything we need for the cache (e.g. yarn test:offline)
- instead of contacting the npm registry, we can use something similar to https://github.com/node-nock/nock
what do you think? :P
I still don't get how request caches works, maybe my comment is out of the context.

bestander:
Request caches are binary files that __tests__/__mocks__/request.js uses to
serve instead of making an http request when running tests.
The good thing about it is simplicity, what we need is a folder with
.tgz.bin files.

Currently it grew to 60mb which is probably now a bit too much.
I would still want to keep it simple because any additional infrastructure
will increase the chance of failures.
Git submodule seems reasonable and simple enough

PR #3149, opened 2017-04-14
added check on response code on publish cli command to accomodate add…

mrichins:
…itional registries

This PR is based on the following closed PR https://github.com/yarnpkg/yarn/pull/1417
When working with Artifactory, the original implementation was dependent on the response body from the npm public registry. Following that original PR, checking the response code seemed to be a general enough approach to accomodate both.

Also I noticed that this exact response checking logic was found elsewhere in the codebase. If desired, I could migrate them over to this.

**Test plan**
I added a simple method to misc.js to check the response code of the response object, but when I was looking at the code, it looks like the publish.js cli command has very little test coverage, at least from what I was able to figure out.

If adding additional test coverage to the publish.js file is necessary, I would love to assist in that effort, as the projects im working with currently are trying to migrate completely to yarn, but this one defect is preventing us from completing that and extra test coverage would help us complete the migration with confidence.

### DIFF
--- a/__tests__/util/misc.js
+++ b/__tests__/util/misc.js
@@ -21,3 +21,8 @@ test('sortAlpha', () => {
     'foo@~6.8.x',
   ]);
 });
+
+test('has2xxResponse', () => {
+  const response = {responseCode: 200};
+  expect(misc.has2xxResponse(response)).toEqual(true);
+});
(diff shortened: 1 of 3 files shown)


bestander:
Sounds reasonable, thans

brimworks:
When I try to `yarn publish` a 200 response is percieved as an error. When I add debug code it appears that the responseCode field is not set. This change appears to have broken `yarn publish`, are others able to `yarn publish` with this code in place? Note that I'm using a custom npm repository.

Thanks,
-Brian

bestander:
@brimworks, it would be better if you open a new issue with repro steps.

Issue #3155, opened 2017-04-14
--no-progress Flag Does Not Silence Activity/Progress Output With --json Flag


alecmerdler:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**
The `--no-progress` flag does not silence activity/progress output with the `--json` flags on any command. 

**If the current behavior is a bug, please provide the steps to reproduce.**
Example: `yarn install --no-progress --json` 
Output: 
{"type":"step","data":{"message":"Resolving packages","current":1,"total":4}}
{"type":"activityStart","data":{"id":0}}
{"type":"activityTick","data":{"id":0,"name":"angular@1.6.2"}}
{"type":"activityTick","data":{"id":0,"name":"angular-animate@1.6.2"}}
{"type":"activityTick","data":{"id":0,"name":"angular-cookies@1.6.2"}}
{"type":"activityTick","data":{"id":0,"name":"angular-route@1.6.2"}}
{"type":"activityTick","data":{"id":0,"name":"angular-sanitize@1.6.2"}}
...
{"type":"progressTick","data":{"id":0,"current":715}}
{"type":"progressTick","data":{"id":0,"current":716}}
{"type":"progressTick","data":{"id":0,"current":717}}
{"type":"progressFinish","data":{"id":0}}

**What is the expected behavior?**
No progress output of any kind is produced.

**Please mention your node.js, yarn and operating system version.**
Node: **v7.2.0**
Yarn: **v0.21.3**
OS: **Manjaro Linux v17.0.1**

pweaver:
Shouldn't --json imply --no-progress?

sschuberth:
Looks like the issue has been fixed in Yarn 0.24.

PR #3158, opened 2017-04-15
Hide Activity/Progress Output in JSONReporter with --no-progress Flag

alecmerdler:
**Summary**
Addresses #3155 by applying the same logic used in the `ConsoleReporter` to hide activity and progress output for the `JSONReporter` when using `--no-progress`.

Originally wanted to be able to generate a list of licenses used in JSON format using `yarn licenses ls --json --no-progress`, but extra output from activity/progress was shown. Discovered that the `JSONReporter` does not contain the same logic to hide activity/progress output as other reporters.

**Test plan**
Automated: `yarn test`
Manual: `yarn [cmd] --json --no-progress`

**Before:**
screenshot_20170414_170617

**After:**
screenshot_20170414_170636

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ test/fixtures/**/.fbkpm
 /tmp/
 /__tests__/fixtures/**/_*
 /__tests__/fixtures/request-cache/GET/localhost/.bin
-
+.idea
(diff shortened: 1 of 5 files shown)


alecmerdler:
Snapshot tests are failing because [this line](https://github.com/yarnpkg/yarn/blob/master/src/reporters/base-reporter.js#L62) sets `noProgress` to `true` if in a CI environment.

alecmerdler:
I would appreciate some help with getting the AppVeyor build to succeed. One of the integration tests seems to be timing out.

arcanis:
Yeah, some tests are subject to timeouts, we're working on fixing this. Nice work, thanks!

Issue #3152, opened 2017-04-14
Feature: Add argument to expose json format of yarn.lock file


steven10172:
It would be nice to have a command that will print the output of the lock parser. 

Example Command: `yarn --lock-json`

Example code to achieve this output:
```
import Parse from 'yarn/src/lockfile/parse.js';
import * as fs from 'fs';

fs.readFile('yarn.lock', (err, data) => {
  if (err) throw err;
  console.log(Parse(data.toString()));
});
```

wmhilton:
Actually, I'd love it if someone made the yarn.lock file parser/writer a separate npm module. As it is, it is very hard for others to write package managers interoperable with yarn.

steven10172:
I would prefer it to be in this package as anyone who wants to can easily make a basic package that just imports yarn's src and uses the lines above.

wmhilton:
> anyone who wants to can easily make a basic package that just imports yarn's src and uses the lines above.

Wait, really? I hadn't thought of that. Looks like "src" isn't included in the package distribution though.

wmhilton:
> anyone who wants to can easily make a basic package that just imports yarn's src and uses the lines above.

Whadya know, you were right @steven10172! Actually those lines above don't work (unless you get the source from Github I guess) but the package has the transpiled versions are available in `yarn/lib`. https://npm.im/yarn-lockfile

PR #3159, opened 2017-04-15
Add command to output json format of lockfile

steven10172:
Original request: #3152 

**Summary**
This pr hooks into the existing lockfile parser and makes the data available through a new yarn argument (`lockfile-json`). This new argument will make it easier for users that want to be able to parse the lockfile and see what's being installed without having to write or rely on 3rd party lockfile parsers.

**Test plan**
Manual: `yarn lockfile-json`

Example:
screen shot 2017-04-15 at 3 34 17 am

### DIFF
--- a/src/cli/commands/index.js
+++ b/src/cli/commands/index.js
@@ -16,6 +16,7 @@ import * as init from './init.js'; export {init};
 import * as install from './install.js'; export {install};
 import * as licenses from './licenses.js'; export {licenses};
 import * as link from './link.js'; export {link};
+import * as lockfileJson from './lockfile-json'; export {lockfileJson};
 import * as login from './login.js'; export {login};
 import * as logout from './logout.js'; export {logout};
 import * as list from './list.js'; export {list};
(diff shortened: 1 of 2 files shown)


arcanis:
I'm not sure we want to do this. Maybe the lockfile file format should be more standardized (json, or yaml, maybe, since it seems quite close to what we currently have), but I'm not convinced that adding a new command is the best way to go. What do you think, @bestander?

steven10172:
I'm for using a standard format (JSON) in the lock file. It'll allow for the code in the parser to be removed, or greatly reduced. If the format does get changed it might be nice to have some additional metadata such as license so `yarn licenses ls` doesn't have to run a dry-run install.

If a standard format for the lock file isn't used it might still be useful to add the license data to the lock file and maybe making a `yarn-utils` package that publishes the full source code with proper entrances without requiring cli command context.

bestander:
I agree with @arcanis, we probably don't need this feature in Yarn.

This has been brought up before a few times and so far we all preferred readability over native JS parsing, that is why we are sticking to YAML-like format.

I understand that a JSON representation may be useful but I would recommend building a small standalone tool for your needs first.
That would keep Yarn smaller and easier to support.

bestander:
Feel free to add more examples where it may be useful, this is a community project after all.

PR #2598, opened 2017-01-31
Fix registry auth comparison for explicit port numbers

itaylor:
**Summary**
I use a private NPM registry hosted via Artifactory.  When using Yarn, I'm unable to install my modules when it gets to the "Fetching packages" phase, but it gets through the resolving packages phase using the private registry just fine.  I traced it back to the `npm-registry.js` doing a comparison to see if the request string `requestUrl.startsWith(registry)` this comparison fails on my server because the request url looks like 
`https://my.server:443/path/to/registry/-/something/somefile.tgz`
and the registry url looks like:
`https://my.server/path/to/registry/`

These are clearly requests to the same url repo, but a string.startsWith doesn't understand that.  As a result, the requests are made without auth info and fail with a Not authorized error.  Similar errors would occur with the string.startsWith on differing capitalization.

**Test plan**

I added tests for the function that does the registry comparisons in the `__tests__/registries` folder.   The tests cover both positive and negative cases for the new comparison function.

Run before change:
With `.npmrc` content in the folder:
```
registry = https://myserver.mydomain.com/artifactory/api/npm/npm-mkto
_auth = myAuthStringSomeCrazyThing=
always-auth = true
```

```
$ yarn install
yarn install v0.19.1
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "https://myserver.mydomain.com:443/artifactory/api/npm/npm-mkto/react-height/-/react-height-2.2.0.tgz: Request failed \"401 Unauthorized\"".
info If you think this is a bug, please open a bug report with the information provided in "/usr/src/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
Run in the same folder with this PR:
```
$ yarn install
yarn install v0.21.0-0
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
Done in 115.13s.

PR #3172, opened 2017-04-17
Fixies for running with artifactory 4.14.1

rogeriochaves:
**Summary**

This will help solving #521 

I use a private NPM registry hosted via Artifactory v4.14.1 and I was having some issues.

The first one is that apparently this version of Artifactory still does not support the `application/vnd.npm.install-v1+json` header implemented on #3112. To fix that, I noted by reading [npm docs](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md) that we can add a fallback to `application/json`.

This allowed yarn to correctly download metadata from artifactory. Then, I had another problem, with the `isRequestToRegistry` function implemented on #2598. The problem was that although I do specify `https://myartifactory` for my registy, some dependencies gets the `https` replaced with `http` by yarn, preventing it from sending the authentication header. I don't know why this happens, maybe its artifactory itself telling yarn to use http (I've noted it happens on public dependencies)

So, I've modified the function `isRequestToRegistry` to understand that `https://myartifactory` and `http://myartifactory` is still the same registry.

**Test plan**

I've added extra test cases to `isRequestToRegistry functional test`, maybe some snapshots testing needs to be changed for the `vnd.npm` header, but they gave no errors on my machine.

Cheers!

### DIFF
--- a/__tests__/registries/is-request-to-registry.js
+++ b/__tests__/registries/is-request-to-registry.js
@@ -20,14 +20,14 @@ test('isRequestToRegistry functional test', () => {
   )).toBe(true);
 
   expect(isRequestToRegistry(
-      'https://foo.bar/foo/bar/baz',
-      'https://foo.bar:443/foo/',
-    )).toBe(true);
+    'https://foo.bar/foo/bar/baz',
+    'https://foo.bar:443/foo/',
+  )).toBe(true);
 
   expect(isRequestToRegistry(
     'http://foo.bar:80/foo/bar/baz',
     'https://foo.bar/foo/',
-  )).toBe(false);
+  )).toBe(true);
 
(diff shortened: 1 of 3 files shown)


tuto:
we need this in our team :dancer: 
very useful and necesary

Thank you  very much Rogerio

rogeriochaves:
@Daniel15 @bestander @kittens could you review this PR? We really need this

joyarzun:
Wowwww finally, thanks @bestander and @rogeriochaves

Daniel15:
Thanks for the pull request! If you want to use this today (before a new stable/RC release comes out), you can grab a nightly build: https://yarnpkg.com/en/docs/nightly

PR #3170, opened 2017-04-17
Use tar-fs instead of tar-stream in `yarn pack` (and fix packed emojis)

josephfrazier:
**Summary**

This lets tar-fs do the [header construction] for us.

[header construction]: https://github.com/mafintosh/tar-fs/blob/b79d82a79c5e21f6187462d7daaba1fc03cdd1de/index.js#L101-L127

**Test plan**

I tested this by comparing the output of this command before and after
the change:

    ./bin/yarn.js pack >/dev/null && tar tvf yarn-v0.24.0-0.tgz | sort && wc -c < yarn-v0.24.0-0.tgz && rm *tgz

Here's the diff between the outputs:

```diff
diff --git a/before.txt b/after.txt
index 5e7f370e..5565a808 100644
--- a/before.txt
+++ b/after.txt
@@ -7,13 +7,13 @@
 -rw-r--r--  0 0      0         657 Mar  4 07:19 package/Dockerfile.dev
 -rw-r--r--  0 0      0        1346 Mar  4 07:19 package/LICENSE
 -rw-r--r--  0 0      0        1789 Apr 17 15:10 package/jenkins_jobs.groovy
--rw-r--r--  0 0      0        3061 Mar  4 07:19 package/README.md
--rw-r--r--  0 0      0        3438 Apr 17 16:18 package/package.json
+-rw-r--r--  0 0      0        3057 Mar  4 07:19 package/README.md
+-rw-r--r--  0 0      0        3430 Apr 17 16:18 package/package.json
 -rwxr-xr-x  0 0      0          42 Mar  4 07:19 package/bin/yarnpkg
 -rwxr-xr-x  0 0      0         172 Mar  4 07:19 package/bin/node-gyp-bin/node-gyp
 -rwxr-xr-x  0 0      0         906 Mar  4 07:19 package/bin/yarn
 -rwxr-xr-x  0 0      0         929 Apr 10 15:59 package/bin/yarn.js
 drwxr-xr-x  0 0      0           0 Apr 10 15:59 package/bin
 drwxr-xr-x  0 0      0           0 Apr 17 17:04 package
 drwxr-xr-x  0 0      0           0 Mar  4 07:19 package/bin/node-gyp-bin
-    6206
+    6177
```

I extracted the tarballs into `./package-master` and `./package-feature`,
then diffed them to find that this change has the side effect of
fixing emojis in the tarball. You can see examples of the broken emoji
here:

* https://unpkg.com/yarn@0.22.0/package.json
[...]

Issue #2942, opened 2017-03-17
Broken published packages caused by `yarn pack` failing to include entire directories


fvgs:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
`yarn pack`, and thus also `yarn publish`, fail to include the contents of directories specified in the `files` package.json field resulting in broken packages being published to the registry.
e.g.
```javascript
{
  ...

  "files": ["src/"]

  ...
}
```
The directory `src/` itself is included, but its contents are not included (except files named `index.js` as observed in the below example). This differs from the behavior of `npm pack` which includes all of the specified directory's contents. This also contradicts the [documentation of the files field](https://yarnpkg.com/en/docs/package-json#toc-files) which uses a similar example to the one above.

**If the current behavior is a bug, please provide the steps to reproduce.**
```sh
# The `files` field in the express package is appropriate for testing this
git clone https://github.com/expressjs/express
cd express

# Observe the files missing from the lib/ directory
# (Adjust the name of the tarball as appropriate)
yarn pack && tar -xvf express-v4.15.2.tgz && rm -r package

# Observe the correct package contents
# (Also note that npm does not package the examples/ directory while yarn does.
#  But that is not the main issue here)
npm pack && tar -xvf express-v4.15.2.tgz && rm -r package
```

**What is the expected behavior?**
All contents of `src/` should be included in the resulting tarball, matching npm behavior and yarn's own documentation.
Additionally, the `examples/` directory should probably not be included in the tarball, thereby matching npm behavior. But the primary issue is the first one.

**Please mention your node.js, yarn and operating system version.**
node -> 7.7.3
yarn -> 0.21.3
os -> macOS Sierra 10.12.3

For others encountering this issue, I have found the use of `src/**` instead of `src/` in the `files` field to achieve the desired results (i.e. include all contents of the `src/` directory).

joshburgess:
I just experienced this also. A user reported that one of my npm packages had completely empty folders after they installed, causing all imports to break. I quickly republished using npm and the problem was fixed. yarn publish is currently broken...

satya164:
I've faced this issue too. Seems like `yarn publish` is unusable right now.

Issue #2851, opened 2017-03-07
yarn pack: directories listed in "files" property are empty


simonvizzini:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Directories specified in the "files" property of the package.json are empty inside the package produced by ``yarn pack``

**If the current behavior is a bug, please provide the steps to reproduce.**

1. git clone https://github.com/simonvizzini/yarn_pack_bug.git
2. yarn pack

The produced ``yarn_pack-v1.0.0.tgz`` contains an empty "src" folder. ``npm pack`` on the other hand works fine.

**What is the expected behavior?**

Directories specified in "files" should not be empty.

**Please mention your node.js, yarn and operating system version.**

node 7.3.0
yarn 0.21.3
Windows 10 x64

alexkrolick:
Same on Mac OS Sierra, node 7.5.0, yarn 0.23.1

Issue #2498, opened 2017-01-19
Files inside directories aren't included when running pack


squiddy:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Package built with `yarn pack` doesn't include files inside directories.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
$ yarn init
$ mkdir a
$ touch a/b
```

Edit `package.json` and add `files` setting:

```js
{
  "name": "tmp_package",
  "version": "1.0.0",
  "main": "index.js",
  "repository": {},
  "license": "MIT",
  "files": [
	  "a/"
  ]
}
```

```
$ yarn pack
$ tar --list -zf tmp_package-v1.0.0.tgz
package
package/a
package/package.json
```

**What is the expected behavior?**

The package includes file `a/b`.
```
$ tar --list -zf tmp_package-v1.0.0.tgz
package
package/a/b
package/package.json
```

When running `npm pack` the file is included.

Perhaps this is merely an documentation issue:

>  If you name a folder in the array, then it will also include the files inside that folder. (Unless they would be ignored by another rule.)

https://docs.npmjs.com/files/package.json#files

> You can specify single files, whole directories or [...]
https://yarnpkg.com/en/docs/package-json#toc-files

**Please mention your node.js, yarn and operating system version.**

```
node: 6.6.0
yarn: 0.19.1
os: Linux 4.8.13-1-ARCH #1 SMP PREEMPT Fri Dec 9 07:24:34 CET 2016 x86_64 GNU/Linux
```

squiddy:
On a related note, packages built with `yarn pack` that include directories in the `files` setting aren't installable.

Trying to install the package I build in the steps above results in this error:
```
$ yarn add file:/tmp/tmp_package/tmp_package-v1.0.0.tgz                                                                                                                                                                                [11:57]
yarn add v0.19.1
[1/4] Resolving packages...
error An unexpected error occurred: "invalid tar file".
info If you think this is a bug, please open a bug report with the information provided in "/tmp/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

Same error as in #1253

majgis:
We are seeing this issue too.  Replacing "foldername/" with "foldername/**/*" appears to be a workaround.

jamiewinder:
I've just encountered this too, but the workaround @majgis posted doesn't seem to work either:
```json
  "files": [
    "dist/",
[...]

PR #3175, opened 2017-04-18
`pack`: include contents of directories in `files` field

josephfrazier:
**Summary**

This makes it so that you don't have to put '/**' after a directory in the
`files` field of package.json to ensure that the contents of the directory
will be published.

Fixes https://github.com/yarnpkg/yarn/issues/2498
Fixes https://github.com/yarnpkg/yarn/issues/2942
Fixes https://github.com/yarnpkg/yarn/issues/2851

Includes and closes https://github.com/yarnpkg/yarn/pull/3170

**Test plan**

I modified one of the pre-existing pack tests to also include and check
for a file inside a directory listed in package.json.


TrySound:
Fixes https://github.com/yarnpkg/yarn/issues/2986
Fixes https://github.com/yarnpkg/yarn/pull/3166

arcanis:
A [test](https://ci.appveyor.com/project/kittens/yarn/build/2113/job/7pxs5e44eq109adb#L3046) seems to be failing on Windows, can you check?

arcanis:
The Windows test still breaks.

Btw, do you support this edge case: `{ "files": [ "test/*" ] }`? Since it doesn't evaluate to a valid stat call it will get added as-is, so it probably won't pack any file beneath.

Maybe a better way to handle this would be to transform each `[ "file-name" ]` into `[ "file-name", "file-name/**" ]`, whether it's a file or a folder. Glob will filter them accordingly anyway.

josephfrazier:
Yeah.... I don't have a Windows machine readily available to develop on, so I've been relying on AppVeyor as a bit of debugging tool (hence https://github.com/yarnpkg/yarn/pull/3175/commits/517d7df3cbd0642931eda84ba56f5eded4752eb6) :P

I just pushed a commit that tries the `[ "file-name" ]` into `[ "file-name", "file-name/**" ]` approach, and it looks like that gets a little closer to passing CI (only [one missing file](https://ci.appveyor.com/project/kittens/yarn/build/2137/job/ayfe6ncn81qq84us#L5135), rather than [two](https://ci.appveyor.com/project/kittens/yarn/build/2130/job/4hjfo6ke6dcsh2rw#L5160)). Thanks for the suggestion!

As for the `{ "files": [ "test/*" ] }` edge case, I tried changing the `dir/` into `dir/*` [here](https://github.com/yarnpkg/yarn/pull/3175/commits/6626c5e37ad66eec58143bf5bb481812258efbf2#diff-76aeddc6ab4277c9c808f7bdbe1684eeR6), and the tests still pass locally, so hopefully that won't be an issue, especially after switching to your recommended way of handling this.

I'm going to be mostly unavailable until next week, but I should be able to take another look at this then. In the mean time, feel free to push more commits onto this branch if you have some insight on how to fix the Windows tests.

bestander:
Thanks a lot for sending the fix, @josephfrazier.
Looking forward to see the Windows tests pass.

EnoahNetzach:
The main problem with Windows is the path separator: in places like [this one](https://github.com/josephfrazier/yarn/blob/cbdfee66af0b4f16d999af31eebb66a486a463e7/src/util/filter.js#L96) it is assumed it's `/` (in the regexs).

src/util/filter.js
```diff
@@ -99,6 +99,7 @@ export function matchesFilter(filter: IgnoreFilter, basename: string, loc: strin
   }
   return filter.regex.test(loc) ||
          filter.regex.test(`/${loc}`) ||
+         filter.regex.test(`\\${loc}`) ||
```
EnoahNetzach: why not using `path.sep`?

EnoahNetzach: The problem I see with keeping both `/` and `\\` is that the backslash has a meaning in unix systems, so false-positives could occur.

josephfrazier: The argument to `test` is a string rather than a RegExp pattern, so it shouldn't be interpreting backslashes as the escape character in this context. Note that the following expression (on Windows) should produce a string whose first character is a backslash:

```js
`${path.sep}${loc}`
```

That said, I'm not sure why https://github.com/yarnpkg/yarn/pull/3175/commits/c2df043343092dcc408f8792ad16eb86cad6ba3a seemed to make things worse, test-wise (an additional test [failed](https://ci.appveyor.com/project/kittens/yarn/build/2195/job/q5u26f85qlroy533#L3011))

EnoahNetzach:
Printing all the regexs that are tested, I see a lot of unix separators:

```regex
(1) /^(?:\/(?=.)package\.json)$/i
(2) /^(?:\/(?=.)readme[^\/]*?)$/i
(3) /^(?:\/(?!\.)(?=.)(?:license|licence)+[^\/]*?)$/i 
(4) /^(?:\/(?!\.)(?=.)(?:changes|changelog|history)+[^\/]*?)$/i
(5) /^(?:(?=.)index\.js)$/i
(6) /^(?:(?=.)a\.js)$/i
(7) /^(?:(?=.)b\.js)$/i
(8) /^(?:(?=.)dir\/)$/i
(9) /^(?:(?=.)index\.js\/(?:(?!(?:\/|^)\.).)*?)$/i
(10) /^(?:(?=.)a\.js\/(?:(?!(?:\/|^)\.).)*?)$/i
(11) /^(?:(?=.)b\.js\/(?:(?!(?:\/|^)\.).)*?)$/i
(12) /^(?:(?=.)dir\/(?:(?!(?:\/|^)\.).)*?)$/i
(13) /^(?:(?!\.)(?=.)[^\/]*?)$/i
```

2, 3, 4, 8, 9, 10, 11, 12, and 13 all have a `\/`, whereas probably there should be a `\${path.sep}`, again..

Or am I overcomplicating things?

josephfrazier:
The regexes are being generated in [ignoreLinesToRegex](https://github.com/josephfrazier/yarn/blob/c2df043343092dcc408f8792ad16eb86cad6ba3a/src/util/filter.js#L105), which [uses minimatch](https://github.com/josephfrazier/yarn/blob/c2df043343092dcc408f8792ad16eb86cad6ba3a/src/util/filter.js#L127) to build the regex, which normalizes paths to POSIX-style:

* https://github.com/isaacs/minimatch/blob/eed89491bd4a4e6bc463aac0dfb5c29ef0d1dc13/minimatch.js#L122-L125
* https://github.com/isaacs/minimatch/blob/eed89491bd4a4e6bc463aac0dfb5c29ef0d1dc13/minimatch.js#L711-L714

Accordingly, maybe we should change the logic in [matchesFilter](https://github.com/josephfrazier/yarn/blob/c2df043343092dcc408f8792ad16eb86cad6ba3a/src/util/filter.js#L96) to use [Minimatch#match](https://github.com/isaacs/minimatch/blob/eed89491bd4a4e6bc463aac0dfb5c29ef0d1dc13/minimatch.js#L699) instead of manually operating on the regexes.

josephfrazier:
@bestander, @arcanis: The (relevant, non-flaky) tests are passing on all platforms now. Can you have another look when you get a chance? For reference, here are the changes since last week: https://github.com/yarnpkg/yarn/pull/3175/files/cbdfee66af0b4f16d999af31eebb66a486a463e7..37fc30aaf6f006a0a5f7a6fb731cca20c3d197a8

arcanis:
Merged! Thanks! :)

josephfrazier:
Thanks! I think this also fixes https://github.com/yarnpkg/yarn/pull/3166 (thanks for [noticing](https://github.com/yarnpkg/yarn/pull/3175#issuecomment-294710697), @TrySound!)

TrySound:
@josephfrazier Thanks for PR :)

PR #3182, opened 2017-04-18
Use stdout.isTTY instead of tty module

kaatt:
Just a small patch to make it more consistent. I noticed `isTTY` is used in some other files but not in the edited two files.


arcanis:
Good catch! Unfortunately, I'm not sure we'll be able to merge this. It seems the reason we're using the two syntaxes is because we need the `instanceof` check to instruct Flow that `process.stdout` possess two extra properties: `rows` and `columns`. When using `isTTY`, Flow cannot deduce that those two property exists, and reports an error (as you can see in the lint report of this PR). Sorry :(

PR #3190, opened 2017-04-19
feat(Upgrade): add scope flag to the upgrade command

Joge97:
The `upgrade` command now supports the `-S, --scope <scope>` flag. If the flag is set and a scope is provided yarn upgrade only the packages, which belong to the specified scope.

It is very usefull when you have packages under the `@angular` scope, so you can upgrade all of them at the same time without touching the other packages.

I have tested the command in multiple projects and I have written some test code.

```
yarn upgrade --scope @angular
```

I hope you understand the use case and integrate the pull request.


src/cli/commands/upgrade.js
```diff
@@ -27,16 +28,37 @@ export async function run(
     peerDependencies,
   } = await config.readRootManifest() || {};
   const allDependencies = Object.assign({}, peerDependencies, optionalDependencies, devDependencies, dependencies);
+  let addArgs = [];
 
-  const addArgs = args.map((dependency) => {
-    const remoteSource = allDependencies[dependency];
+  console.log(allDependencies);
```
arcanis: Hum 😛

src/cli/commands/upgrade.js
```diff
@@ -27,16 +28,37 @@ export async function run(
...
 
-    if (remoteSource && PackageRequest.getExoticResolver(remoteSource)) {
-      return remoteSource;
+  if (flags.scope) {
+    const searchPattern = new RegExp(`^${flags.scope}`);
```
arcanis: 1/ We probably want to run a check to make sure that `flags.scope` is a valid scope, and abort otherwise
2/ The regex above will make `-S @angular` upgrade `@angular`, but also `@angular-community` etc.

src/cli/commands/upgrade.js
```diff
@@ -27,16 +28,37 @@ export async function run(
...
+        }
+
+        addArgs.push(dependency);
+      }
     }
+  } else {
+    addArgs = args.map((dependency) => {
+      const remoteSource = allDependencies[dependency];
+
+      console.log(remoteSource);
```
arcanis: Bis :)

src/cli/commands/upgrade.js
```diff
@@ -27,16 +28,37 @@ export async function run(
...
+
+    for (const dependency of Object.keys(allDependencies)) {
+      if (searchPattern.test(dependency)) {
+        const remoteSource = allDependencies[dependency];
+
+        if (remoteSource && PackageRequest.getExoticResolver(remoteSource)) {
+          addArgs.push(remoteSource);
+        }
+
+        addArgs.push(dependency);
```
arcanis: Shouldn't that be in an `else` statement?

__tests__/commands/upgrade.js
```diff
@@ -292,3 +292,16 @@ test.concurrent('can prune the offline mirror', (): Promise<void> => {
...
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+
+    expect(lockfile.indexOf('@angular/core@^2.4.9:')).toBeGreaterThanOrEqual(0);
+    expect(lockfile.indexOf('left-pad@^1.0.0:')).toBeGreaterThanOrEqual(0);
+
+    expect(pkg.dependencies['@angular/core']).not.toEqual('^2.4.9');
+    expect(pkg.dependencies['left-pad']).toEqual('^1.0.0');
+  });
+});
```
arcanis: Please also add a test to make sure that `-S @angular` doesn't upgrade `@angular-whatever`

src/cli/commands/upgrade.js
```diff
@@ -27,16 +28,37 @@ export async function run(
     peerDependencies,
   } = await config.readRootManifest() || {};
   const allDependencies = Object.assign({}, peerDependencies, optionalDependencies, devDependencies, dependencies);
+  let addArgs = [];
```
arcanis: Regarding the global workflow of the code, I'd prefer using ES5 array functions:

```js
let addArgs = Object.keys(allDependencies);

if (flags.scope) {
    addArgs = addArgs.filter(arg => {
        return ...;
    });
}

addArgs = addArgs.map(arg => {
    return ...;
});
```

arcanis:
Please also note that your test isn't passing.

Joge97:
I will fix the issues tomorrow.

src/cli/commands/upgrade.js
Joge97: Yes, I will do that.

src/cli/commands/upgrade.js
Joge97: Sorry, my mistake...

src/cli/commands/upgrade.js
Joge97: I will implement a check.

src/cli/commands/upgrade.js
Joge97: Yes, definitely!

__tests__/commands/upgrade.js
Joge97: I will fix the test issue and add this test case.

Joge97:
I have fixed te issues (@arcanis).

src/cli/commands/upgrade.js
```diff
@@ -27,19 +29,43 @@ export async function run(
...
-  const addArgs = args.map((dependency) => {
-    const remoteSource = allDependencies[dependency];
+  if (flags.scope) {
+    flags.scope = flags.scope.replace(/\/?$/g, '/');
+    flags.scope = flags.scope.replace(/^@?/g, '@');
 
-    if (remoteSource && PackageRequest.getExoticResolver(remoteSource)) {
-      return remoteSource;
-    }
+    if (/^@\S*\/$/g.test(flags.scope)) {
```
arcanis: Use `[a-zA-Z0-9-][a-zA-Z0-9_.-]*` instead of `\S*`, this way we ensure that the scope is a valid scope according to npm rules (and we won't risk any schenanigans with extra `/`).

src/cli/commands/upgrade.js
```diff
@@ -27,19 +29,43 @@ export async function run(
...
+      const searchPattern = new RegExp(`^${flags.scope}`);
```
arcanis: It's better to check directly with `String.prototype.startsWith` (otherwise you'll have issues with the `.` character)

arcanis:
The Travis test is failing (apparently, `@angular-mdl/core` get upgraded as well).

src/cli/commands/upgrade.js
Joge97: Okay, I will change it.

src/cli/commands/upgrade.js
Joge97: Can you make an example, because I don't understand the issue?

Joge97:
I have fixed te issues (@arcanis).

Joge97:
I fixed the issues (@arcanis).

arcanis:
Thanks ! :+1:

PR #3205, opened 2017-04-20
Add hasWrapper function in clean and remove command

voxsim:
@arcanis sorry I totally forgot to inform you that clean and remove command miss a hasWrapper function.
This should fix both commands after https://github.com/yarnpkg/yarn/pull/3105

Let mw know,
cheers,
Simon

### DIFF
--- a/src/cli/commands/clean.js
+++ b/src/cli/commands/clean.js
@@ -129,3 +129,7 @@ export async function run(
 }
 
 export function setFlags() {}
+
+export function hasWrapper(): boolean {
+  return true;
+}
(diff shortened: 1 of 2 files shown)


arcanis:
Oh, thanks. Curious Travis didn't catch that.

voxsim:
@arcanis the main problem is that every command has tests that use directly the run function; instead this problem is in `src/cli/index.js`; they are kinda like integration tests. We can think to add a test for every command in `__tests__/index.js` as acceptance test that covers the whole path; What do you think? I think we do that to have fastest tests :P

PR #3207, opened 2017-04-20
Move inquirer code to console reporter

torifat:
This is a long time due.

When I first wrote it I had no idea about the rest of the yarn codebase. Later I realized it would be a better idea to move this inside the reporter. Then we can easily use this in other parts of our codebase.

**Future Wok**
* Update Inquirer flow type.
* ~~Fix *press double `ctrl+c` to get out of interactive mode*~~ (done)

### DIFF
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -397,10 +397,37 @@ export default class ConsoleReporter extends BaseReporter {
       pageSize = process.stdout.rows - 2;
     }
 
+    const rl = readline.createInterface({
+      input: this.stdin,
+      output: this.stdout,
+      terminal: true,
+    });
+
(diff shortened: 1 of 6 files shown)


src/reporters/console/console-reporter.js
```diff
@@ -380,4 +384,51 @@ export default class ConsoleReporter extends BaseReporter {
...
+      terminal: true,
+    });
+
+    // $FlowFixMe: Need to update the type of Inquirer
+    const prompt = inquirer.createPromptModule({
+      input: this.stdin,
+      output: this.stdout,
+    });
+
+    let rejectRef = () => {};
```
torifat: I didn't like doing this at all. Please feel free to suggest any better alternative.

bestander:
Cool stuff!

PR #3216, opened 2017-04-21
Fix bin path on Windows, fixes #2565

### LINKED ISSUE #2565: Wrong bin path on Windows 10
Bug

On path variable, the path to global bin executables installed by yarn is set to `C:\Users\username\AppData\Local\Yarn\.bin` when it should be `C:\Users\username\AppData\Local\Yarn\config\global\node_modules\.bin`

> 

Install yarn, try installing a package globally with `yarn global install bower`, then check the path variable that is set to wrong path.

The correct path where yarn bins are installed is `C:\Users\devBogdan\AppData\Local\Yarn\config\global\node_modules\.bin`
Node 6.9.4, Yarn 0.19.1, Windows 10

SaschaNaz:
**Summary**

`yarn global add (package name)` installs a package on `%LOCALAPPDATA%\Yarn\config\global\node_modules\.bin` but the installer adds wrong path. This PR fixes it.

**Test plan**

No test, can you help me if a test for this is required?

### DIFF
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -39,7 +39,7 @@
 				<Environment
 					Id="BinPath"
 					Name="PATH"
-					Value="[LocalAppDataFolder]Yarn\.bin"
+					Value="[LocalAppDataFolder]Yarn\config\global\node_modules\.bin"
 					Permanent="no"
 					Part="last"
 					Action="set"


SaschaNaz:
It seems that master already fails..

Daniel15:
I think #3233 will actually fix this.

SaschaNaz:
@Daniel15 Great :D

PR #3218, opened 2017-04-21
Add .yarnrc env config option

kittens:
**Summary**

This PR adds an option called `env` to `.yarnrc`. It allows you to specify environment variables that will be exposed to lifecycle scripts.

For example:

**.yarnrc**

```
env:
  FOO "BAR"
```

**package.json**

```
{
  "scripts": "node install.js"
}
```

**install.js**

```
console.log(process.env.FOO === "bar");
```

**Test plan**

Added a test.


torifat:
LGTM 👍

PR #3221, opened 2017-04-21
Fix tarball file

arcanis:
**Summary**

The offline mirror feature accidentally disabled the codepath that was used for installing packages from `file:` protocol. The tests didn't caught this error because they're using a wrapper class, `LocalTarballFetcher`, that wasn't affected by this issue.


arcanis:
Wait, the revert shouldn't be there, I'll strip it

arcanis:
👍

bestander:
Looks like a few tests failed

arcanis:
Should be fine now

__tests__/commands/install/lockfiles.js
```diff
@@ -10,11 +10,16 @@ import {getPackageVersion, runInstall} from '../_helpers.js';
 import {promisify} from '../../../src/util/promise';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
-
+test.concurrent = test.skip;
```
bestander: oops :)

bestander:
@arcanis, you forgot to reenable tests

__tests__/commands/install/lockfiles.js
arcanis: :scream:

bestander:
This PR fixes relative paths but not absolute ones both on Windows and Unix.
Is it safe to add a root path to the regex, @arcanis?

arcanis:
I could change the regex to: `/^(?!\.\.\/|\/|[a-z]:[\\\/])/`, what do you think?

arcanis:
It seems to catch everything: https://regex101.com/r/zS5bmW/1

bestander:
Looks good to me, can you send a PR when you have a chance?

PR #3222, opened 2017-04-21
Partially revert #2657 to fix add command test

mourner:
Adds back a few lines that were removed in #2657 that apparently caused the `command add` test to fail. cc @bestander @arcanis


mourner:
Yep, Travis and Circle pass again.

PR #3231, opened 2017-04-22
[Replace #2507] Support VSTS and other custom registries with unusual feeds

### LINKED ISSUE #2505: Yarn doesn't work with visual studio online package manager

lumaxis:
This is a continuation of #2507 where the progress has recently stalled. Fixes #2505.

## Quick Summary

[Visual Studio Team Services](https://www.visualstudio.com/team-services/) (VSTS) unfortunately produces package feed URLs that look like `https://account.pkgs.visualstudio.com/_packaging/feed-name/npm/registry` but the URLs to fetch the actual tars then point to `https://account.pkgs.visualstudio.com/_packaging/a917cf10-dfe4-4781-b0d1-105df772da02/npm/registry/@scope/package-name/-/package-name-0.0.1.tgz`.

This Pull Request aims to fix this by picking up a suggestion from #2507 by @bestander and introduces a new `.yarnrc` option called `custom-host-suffix`. That options allows to skip a check in the form of `requestPath.startsWith(registryPath)` if the request's URL contains the value of `custom-host-suffix`.

## Tests

I tried to add tests as best as I could but had trouble coming up with a good way of integration testing this change as I couldn't find an easy way to set up a package feed on VSTS that is publicly accessible. Is there a precedence for using shared credentials in the test suite? Maybe that would be an option.
I'll appreciate any advice for better tests 👍

### DIFF
--- a/__tests__/fixtures/install/install-from-authed-private-registry-no-slash/.npmrc
+++ b/__tests__/fixtures/install/install-from-authed-private-registry-no-slash/.npmrc
@@ -1,2 +1,3 @@
 //registry.yarnpkg.com/:_authToken=abc123
+//registry.yarnpkg.com/:always-auth=true
 @types:registry=https://registry.yarnpkg.com
--- a/__tests__/fixtures/install/install-from-authed-private-registry/.npmrc
+++ b/__tests__/fixtures/install/install-from-authed-private-registry/.npmrc
@@ -1,2 +1,3 @@
 //registry.yarnpkg.com/:_authToken=abc123
+//registry.yarnpkg.com/:always-auth=true
 @types:registry=https://registry.yarnpkg.com/
--- a/__tests__/registries/is-request-to-registry.js
+++ b/__tests__/registries/is-request-to-registry.js
@@ -3,49 +3,70 @@
 
 import isRequestToRegistry from '../../src/registries/is-request-to-registry.js';
 
-test('isRequestToRegistry functional test', () => {
-  expect(isRequestToRegistry(
-    'http://foo.bar:80/foo/bar/baz',
-    'http://foo.bar/foo/',
-  )).toBe(true);
+describe('isRequestToRegistry functional test', () => {
+  test('request to registry url matching', () => {
+    expect(isRequestToRegistry(
+      'http://foo.bar:80/foo/bar/baz',
+      'http://foo.bar/foo/',
+    )).toBe(true);
 
-  expect(isRequestToRegistry(
-    'http://foo.bar/foo/bar/baz',
-    'http://foo.bar/foo/',
-  )).toBe(true);
+    expect(isRequestToRegistry(
+      'http://foo.bar/foo/bar/baz',
+      'http://foo.bar/foo/',
+    )).toBe(true);
 
-  expect(isRequestToRegistry(
-    'https://foo.bar:443/foo/bar/baz',
-    'https://foo.bar/foo/',
-  )).toBe(true);
+    expect(isRequestToRegistry(
+      'http://foo.bar/foo/00000000-1111-4444-8888-000000000000/baz',
+      'http://foo.bar/foo/',
+    )).toBe(true);
 
-  expect(isRequestToRegistry(
-    'https://foo.bar/foo/bar/baz',
-    'https://foo.bar:443/foo/',
-  )).toBe(true);
+    expect(isRequestToRegistry(
+      'https://foo.bar:443/foo/bar/baz',
+      'https://foo.bar/foo/',
+    )).toBe(true);
 
-  expect(isRequestToRegistry(
-    'http://foo.bar:80/foo/bar/baz',
-    'https://foo.bar/foo/',
-  )).toBe(true);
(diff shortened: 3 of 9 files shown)


lumaxis:
Would be great to get some feedback on this! 🙂

bestander:
Great job, thanks for patience, @lumaxis.

bestander:
@lumaxis, now that we have this feature in Yarn it would be great for people to know about it.
Would you write a post https://github.com/yarnpkg/website/tree/master/_posts how it can be used?

xnnkmd:
Has this feature been released in a public downloadable yarn ? I so, which version?

bestander:
Not yet, will be released in 0.26

lumaxis:
@bestander I think I should be able to put something together within the week 🙂 Anything special I should look out for?

bestander:
Cool!
Nothing special, just spread the word about the cool feature :)

lumaxis:
@bestander There you go 🙂

bestander:
@lumaxis, this change might have caused https://github.com/yarnpkg/yarn/issues/3765.
We have a tentative quick fix baking here https://github.com/yarnpkg/yarn/pull/3774.

Apparently we don't have tests covering neither private repos nor vsts repo and I am concerned that change in one might break the other.
Can you chime in the review of #3774?
And let's think more about setting up e2e tests, I think the best way would be to create an account with read access to one private package and store the credential on the CI.
Would you help us out with that?

lumaxis:
@bestander Oh dear, that sounds terrible! Very sorry to maybe have caused that 😞 
I was kind of expecting that private repos were covered by either unit or integration tests already, so didn't check that explicitly ...
I'll take a look at #3774 and also check if and how we could set up an E2E test.

bestander:
Thanks, Lukas.
Very much appreciated, it could be not your change after all but it seems
to touch related code.

It is our (mine) laziness to blame though, we should have set up e2e tests
for private packages long ago.

Mardoxx:
We had an issue where we would get 401 with VSTS npm feed, yielded the following error for `yarn install`:

```
[1/4] Resolving packages...
verbose 1.275 Performing "GET" request to "https://companyname.pkgs.visualstudio.com/_packaging/FeedName/npm/body-parser".
verbose 1.465 Request "https://companyname.pkgs.visualstudio.com/_packaging/FeedName/npm/body-parser" finished with status code 401.
verbose 1.466 Error: Couldn't find package "body-parser" on the "npm" registry.
    at C:\Program Files (x86)\Yarn\lib\cli.js:48169:15
    at step (C:\Program Files (x86)\Yarn\lib\cli.js:92:30)
    at C:\Program Files (x86)\Yarn\lib\cli.js:103:13
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:169:7)
error Couldn't find package "body-parser" on the "npm" registry.

```

.npmrc was like this
```
registry=https://companyname.pkgs.visualstudio.com/_packaging/FeedName/npm/registry
always-auth=true
```

setting the URL to `https://companyname.pkgs.visualstudio.com/_packaging/FeedName/npm/registry/` i.e. with the trailing slash made it work fine.

VSTS have updated the URL generated in the instructions so it now produces them with the trailing slash. So this should not be a problem in the future, but this tripped us up this morning. Not sure why this would be an issue. Worth noting that it works fine with, and without, trailing slash with `npm install`.

PR #3233, opened 2017-04-23
Fix global module executable path on Windows

### LINKED ISSUE #3055: Yarn global packages installed in incorrect directory on Windows

bug (regression?)

yarn add global fails with `EPERM: operation not permitted`

1. install yarn using MSI on windows 10 x64
2. do `yarn init` and then `yarn global add codeceptjs`

same as `npm -g install codeceptjs` i.e. install the package globally without errors

nodejs v7.8
yarn v0.22.0
windows 10 pro x64

**Stacktrace:**
```
yarn init v0.22.0
question name (shaadi-try):
question version (1.0.0):
question description:
question entry point (index.js):
question repository url:
question author:
question license (MIT):
success Saved package.json
Done in 4.13s.

C:\my-home\AutomationProjects\CodeceptProjects\shaadi-try>yarn global add codeceptjs
yarn global v0.22.0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Installed "codeceptjs@0.5.1" with binaries:
      - codeceptjs
error An unexpected error occurred: "EPERM: operation not permitted, open 'C:\\Program Files\\nodejs\\codeceptjs.cmd'".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\Dheeraj\\AppData\\Local\\Yarn\\config\\global\\yarn-error.log".
```

**yarn error log:**
[https://gist.github.com/dheerajbhaskar/07cdc55e1d80282c4782f948638fa7c4](https://gist.github.com/dheerajbhaskar/07cdc55e1d80282c4782f948638fa7c4)

Daniel15:
**Summary**
Node.js is almost always installed in `C:\Program Files\Yarn`, which regular users can't write to. However, Yarn defaults to installing the executables for global modules in the Node directory. This means that the out-of-the-box experience for people installing global modules on Windows is totally broken.

It should instead place them in a directory that the user can write to. I'm going with `%LocalAppData\Yarn\bin` (eg. `C:\Users\Daniel\AppData\Local\Yarn\bin`) as it makes the most sense given all the config and cache is already in that directory. I also updated the path in the installer, as the path the installer was using didn't actually exist (`.bin` instead of `bin`).

**Test plan**

Tested `bin\yarn global add webpack`, it placed the files in the right directory:

Closes #3055

### DIFF
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -39,7 +39,7 @@
 				<Environment
 					Id="BinPath"
 					Name="PATH"
-					Value="[LocalAppDataFolder]Yarn\.bin"
+					Value="[LocalAppDataFolder]Yarn\bin"
 					Permanent="no"
 					Part="last"
 					Action="set"
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -77,6 +77,9 @@ function getGlobalPrefix(config: Config, flags: Object): string {
   } else if (process.env.PREFIX) {
     return process.env.PREFIX;
   } else if (process.platform === 'win32') {
+    if (process.env.LOCALAPPDATA) {
+      return path.join(process.env.LOCALAPPDATA, 'Yarn', 'bin');
+    }
     // c:\node\node.exe --> prefix=c:\node\
     return path.dirname(process.execPath);
   } else {


dustinkingen:
Can confirm the new if statement fixes the issue. Not sure whether changing `.bin` -> `bin` is a breaking change. I noticed that the `.bin` folder is not created after installing yarn. I had to manually add the folder to `AppData\Local\Yarn`. Yarn will need to ensure `.bin` or `bin` exists so that scripts installed with `yarn global` will be available on the path.

Daniel15:
>  Not sure whether changing .bin -> bin is a breaking change

`.bin` never actually existed, so that should be fine.

Daniel15:
Just realised this is broken for non-root users on Linux too... Need to fix that as well.

```
test@a5cf448820fd:/tmp/lol$ /yarn-src/bin/yarn global add create-react-app
yarn global v0.24.0-0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Installed "create-react-app@1.3.0" with binaries:
      - create-react-app
error We don't have permissions to touch the file "/usr/bin/create-react-app".
```

gaearon:
Thanks for the fix.

wtchangdm:
Sorry for asking question here, is the PATH environment variable added only when yarn was installed by Windows Installer and not by npm?

Daniel15:
Yeah, that's right. We could probably also do it as a post install script,
but it was much easier to do in the installer so we just have it there for
now.

Issue #3208, opened 2017-04-20
Environment variable for yarn cache directory location


colinodell:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Cache location can only be configured via the `cache-folder` config option.

**What is the expected behavior?**
Being able to configure the cache location using an environment variable.

Other prominent package managers offer this functionality:

 - Composer: `COMPOSER_CACHE_DIR`
 - npm: `NPM_CONFIG_CACHE`
 - RubyGems: `GEM_SPEC_CACHE`
 - bower: `bower_storage__packages` (unconfirmed)
 - pip: `PIP_DOWNLOAD_CACHE` (unconfirmed)

This is extremely useful for continuous integration - you may want jobs to use a shared cache, but each individual job may not know (or perhaps shouldn't know) where that global cache exists.  Additionally, if that global cache directory changes, it's much easier to update a single environment variable in the CI system than updating the `yarn` commands in every individual project.

I therefore propose using an envvar like `YARN_CONFIG_CACHE_FOLDER`.  If defined, this would take precedence over the default cache directory.  (However, it would not take precedence over any `cache-folder` option used on the command line.)

zurawiki:
Sounds like a good first bug @colinodell 

Can I take take a look and try to send in a PR?

colinodell:
Yeah absolutely!

zurawiki:
This feature already seems to exist. The env variable name is `YARN_CACHE_FOLDER`

```sh
$ YARN_CACHE_FOLDER="foo" ./bin/yarn cache dir
foo/v1
```

bestander:
Yes, actually Yarn converts all environment variables into 
[YarnRegistry.config](https://github.com/yarnpkg/yarn/blob/master/src/registries/yarn-registry.js#L60) if they have `yarn_` prefix.

That is not a widely known fact, do you want to write start a doc on the [website](https://yarnpkg.com/en/docs/configuration) for that, @zurawiki ?

glensc:
@bestander never link to branches! now, what line you had in mind?


🚧 Never link to branches, share a permalink 🚧
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
  and https://docs.gitlab.com/ee/user/shortcuts.html#project-files
- hit `y`
- and perhaps for https://github.com/josephfrazier/octopermalinker (GitHub)

PR #3237, opened 2017-04-23
Add environment variable for yarn cache folder

zurawiki:
**Summary**

Following up from issues #3208 and #1211, I have added configurability for the cache folder via environment variable.

The precedence for config is:

1. Yarn Config
2. Environment Variable `YARN_CACHE_FOLDER` I am open to other naming suggestions. I thought this was the clearest name that reflected its intended usage.
3. Constant settings as determined by `src/constants.js`

**Test plan**

  - Added a unit test to ensure correct behavior
    - This involved making some changes to the test harness. please let me know if a better testing method is preferred.

Below are some examples as run from a Mac OS X laptop:

```sh
./bin/yarn cache dir
/Users/roger/Library/Caches/Yarn/v1
```

```sh
YARN_CACHE_FOLDER="foo" ./bin/yarn cache dir
foo/v1
```
```sh
$ YARN_CACHE_FOLDER="foo" ./bin/yarn cache dir --cache-folder bar
bar/v1
```

@colinodell @nervo

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -10,8 +10,9 @@ import * as fs from '../../src/util/fs.js';
 import {Install} from '../../src/cli/commands/install.js';
 import Config from '../../src/config.js';
 
-const stream = require('stream');
+const os = require('os');
 const path = require('path');
+const stream = require('stream');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'install');
 
@@ -49,6 +50,10 @@ export async function getPackageVersion(config: Config, packagePath: string): Pr
   const loc = path.join(config.cwd, `node_modules/${packagePath.replace(/\//g, '/node_modules/')}/package.json`);
   const json = JSON.parse(await fs.readFile(loc));
   return json.version;
+}
+
+export function getTempGlobalFolder(): string {
+  return path.join(os.tmpdir(), `yarn-global-${Math.random()}`);
 }
 
 export async function run<T, R>(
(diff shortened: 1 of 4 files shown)


__tests__/commands/_helpers.js
```diff
@@ -51,6 +52,10 @@ export async function getPackageVersion(config: Config, packagePath: string): Pr
   return json.version;
 }
 
+export function getTempGlobalFolder(): string {
+  return path.join(os.tmpdir(), `yarn-global-${Math.random()}`);
+}
```
zurawiki: moved out from `global.js` for reuse in `cache.js`

voxsim: @zurawiki I suggest to you to use `import mkdir from './_temp.js';`. One example is in https://github.com/yarnpkg/yarn/blob/eac1836cea4a2a8e89d140fa22c204c1c416f14a/__tests__/fetchers.js#L10

zurawiki:
Upon further investigation, this PR is redundant. Yarn already has this feature implemented by the configuration option code (the `this.getOption('cache-folder')` reads from YARN_CACHE_FOLDER it seems.

I noticed this behavior when realizing that my tests still pass without the change to the source code.

This env parsing code works for all options like, for example, `YARN_STRICT_SSL`

```sh
$ yarn config get strict-ssl
true
$ YARN_STRICT_SSL=false yarn config get strict-ssl
false
```

bestander:
IIRC `this.getOption('cache-folder')` reads data from .yarnrc file only.
I am not sure we want to increase number of environment variables that affect Yarn, why it can't be configured via .yarnrc?

bestander:
Yeah, basically any setting that uses `this.getOption` in config.js can be set via .yarnrc or an environment variable with `YARN_` prefix

PR #3238, opened 2017-04-23
Add tests for global command

voxsim:
This pull request adds:
- tests for every global command excluding upgradeInteractive and everyone is side-effect free (I hope)
- move the pre-existent tests under isCI branch because they cause some changes in the prefix or global folder; but If I am wrong please let me know O.o
- Fix https://github.com/yarnpkg/yarn/issues/3162, https://github.com/yarnpkg/yarn/issues/3142, https://github.com/yarnpkg/yarn/issues/3089 (?)
- It closes https://github.com/yarnpkg/yarn/pull/2725 ;)

cc @bestander @arcanis

PR #3239, opened 2017-04-23
Use skipIntegrityCheck instead of skipIntegrity in licenses command

voxsim:
In https://github.com/yarnpkg/yarn/pull/3238 I noticed that we changed the skipIntegrity flag in skipIntegrityCheck (this is the commit https://github.com/yarnpkg/yarn/commit/bd8e326668aa1deac07b53bb9b9d93daf172c070)

**Test plan**
I have no idea how to test it; right now I don't have time to see this part. Let's park here for now.

### DIFF
--- a/src/cli/commands/licenses.js
+++ b/src/cli/commands/licenses.js
@@ -16,7 +16,7 @@ export function hasWrapper(flags: Object, args: Array<string>): boolean {
 
 async function getManifests(config: Config, flags: Object): Promise<Array<Manifest>> {
   const lockfile = await Lockfile.fromDirectory(config.cwd);
-  const install = new Install({skipIntegrity: true, ...flags}, config, new NoopReporter(), lockfile);
+  const install = new Install({skipIntegrityCheck: true, ...flags}, config, new NoopReporter(), lockfile);
   await install.hydrate(true, true);
 
   let manifests = install.resolver.getManifests();


bestander:
No need for a unit test, I think this should have been caught by Flow.

voxsim:
I thought the same, we should thank `git bisect` to found the commit :D 

I think I will spend some time to understand package-* and integrity-checker because IMO they require some love

bestander:
Yeah.
I am planning to do a bit more changes to integrity-checker in the next couple of weeks.
- remove reporter.warn from it
- split check operation from reading artifacts data from .yarn-integrity
- move integrity check into an earlier phase of install command

voxsim:
That sounds great :D

voxsim:
I think to add some unit/integration testing do understand what effectively does

voxsim:
Ah before I forget, If you want to delegate something, I am happy to collaborato too :D

bestander:
Would you want to extract the reporter out of integrity-checker?
I think I won't be able to get down to this task before next week.

voxsim:
Sure, I will open a pull request soon. By 'extract the reporter out of integrity-checker' do you mean remove every report.warn from integrity-checker and add a generic `this.reporter.warn(this.reporter.lang('integrityFailed'));` or something else?
One other guess could be extract _compareIntegrityFiles out from integrity-checker O.o but IMO does not seem a great idea

bestander:
I think integrity-checker should not have side effects and reporting warning is one.
Separation of concerns.
We actually care about outputting warnings about what exactly failed in integrity check in `check` command.

Would be great if those reporter.warn could be moved to check.js command and integrity-checker would just output whether integrity matches and if not in what aspect it failed.

PR #3240, opened 2017-04-23
Use npm's node-gyp if available, otherwise automatically install node-gyp

### LINKED ISSUE #2266: node-gyp builds fail on bundled yarn

### LINKED ISSUE #3114: Yarn should automatically install node-gyp if needed
Bug

The bundled .js version of Yarn (which is now used in all the other installers) doesn't include node-gyp, resulting in errors like this:

```
error C:\temp\license-test\node_modules\libalac: Command failed.
Exit code: 1
Command: C:\WINDOWS\system32\cmd.exe
Arguments: /d /s /c node-gyp rebuild
Directory: C:\temp\license-test\node_modules\libalac
Output:
'node-gyp' is not recognized as an internal or external command,
```

This was already an issue with the standalone .js build, but it'll be an issue with other builds now that we use the standalone .js file everywhere.

Rather than bundling node-gyp with Yarn (like we did before), we should probably run `yarn global add node-gyp` to install it when needed.

Repro:
```
yarn init
yarn add nodetunes
```

Daniel15:
**Summary**
Some Node.js packages depend on `node-gyp` but do **not** declare the dependency in their `package.json`, instead relying on node-gyp to already be globally installed. For packages like this, we used to bundle node-gyp with Yarn itself. However, this does not work for the standalone JS build (which is used by the tarball now too)

The ideal fix is for packages to explicitly specify that they require node-gyp, as relying on the global version is dangerous (eg. if there's breaking changes between releases).

The changes I've made are as follows:
1. Include node-gyp version from npm in the PATH for lifecycle scripts. This means that Yarn should behave the same way as npm, as long as the user has npm installed (not every Node.js user uses npm - Many distributions of Node.js require you to install npm separately).
2. If no node-gyp is available (not even from a local npm installation), attempt to automatically install it by running `yarn global add node-gyp`.

**Test plan**

Tested manually on Windows and on Ubuntu 16.04 by using `yarn add nodetunes`

I'm not sure how I could unit test this... I guess I could mock out the `fs` methods, but that might mess up other tests 😕 

Fixes #2266
Fixes #3114

cc @bestander

### DIFF
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -5,8 +5,13 @@ import type Config from '../config.js';
 import {MessageError, SpawnError} from '../errors.js';
 import * as constants from '../constants.js';
 import * as child from './child.js';
+import {exists} from './fs.js';
 import {registries} from '../resolvers/index.js';
 import {fixCmdWinSlashes} from './fix-cmd-win-slashes.js';
+import {
+  run as globalRun,
+  getBinFolder as getGlobalBinFolder,
+} from '../cli/commands/global.js';
 
 const path = require('path');
 
@@ -116,8 +121,9 @@ export async function executeLifecycleScript(
   // split up the path
   const pathParts = (env[constants.ENV_PATH_KEY] || '').split(path.delimiter);
 
-  // add node-gyp
-  pathParts.unshift(path.join(__dirname, '..', '..', 'bin', 'node-gyp-bin'));
+  // Add global bin folder, as some packages depend on a globally-installed
+  // version of node-gyp.
+  pathParts.unshift(getGlobalBinFolder(config, {}));
 
   // add .bin folders to PATH
   for (const registry of Object.keys(registries)) {
@@ -125,6 +131,8 @@ export async function executeLifecycleScript(
     pathParts.unshift(path.join(config.linkFolder, binFolder));
     pathParts.unshift(path.join(cwd, binFolder));
   }
+
+  await checkForGypIfNeeded(config, cmd, pathParts);
 
   // join path back together
   env[constants.ENV_PATH_KEY] = pathParts.join(path.delimiter);
@@ -167,6 +175,57 @@ export async function executeLifecycleScript(
 }
 
 export default executeLifecycleScript;
+
+let checkGypPromise: ?Promise<void> = null;
+/**
+ * Special case: Some packages depend on node-gyp, but don't specify this in
+ * their package.json dependencies. They assume that node-gyp is available
+ * globally. We need to detect this case and show an error message.
+ */
+function checkForGypIfNeeded(
+  config: Config,
+  cmd: string,
+  paths: Array<string>,
+): Promise<void> {
+  if (cmd.substr(0, cmd.indexOf(' ')) !== 'node-gyp') {
+    return Promise.resolve();
+  }
+
+  // Ensure this only runs once, rather than multiple times in parallel.
+  if (!checkGypPromise) {
+    checkGypPromise = _checkForGyp(config, paths);
+  }
+  return checkGypPromise;
+}
+
+async function _checkForGyp(
+  config: Config,
+  paths: Array<string>,
+): Promise<void> {
+  const {reporter} = config;
+
+  // Check every directory in the PATH
+  const allChecks = await Promise.all(
+    paths.map((dir) => exists(path.join(dir, 'node-gyp'))),
+  );
+  if (allChecks.some(Boolean)) {
+    // node-gyp is available somewhere
+    return;
+  }
+
+  reporter.info(reporter.lang('packageRequiresNodeGyp'));
+
+  try {
+    await globalRun(
+      config,
+      reporter,
+      {},
+      ['add', 'node-gyp'],
+    );
+  } catch (e) {
+    throw new MessageError(reporter.lang('nodeGypAutoInstallFailed', e.message));
+  }
+}
 
 export async function execFromManifest(config: Config, commandName: string, cwd: string): Promise<void> {
   const pkg = await config.maybeReadManifest(cwd);
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -125,13 +125,13 @@ export async function executeLifecycleScript(
   // if available.
   pathParts.unshift(
     path.join(
-      path.dirname(process.execPath), 'node_modules', 'npm', 'bin', 'node-gyp-bin'
-    )
+      path.dirname(process.execPath), 'node_modules', 'npm', 'bin', 'node-gyp-bin',
+    ),
   );
   pathParts.unshift(
     path.join(
-      path.dirname(process.execPath), '..', 'lib', 'node_modules', 'npm', 'bin', 'node-gyp-bin'
-    )
+      path.dirname(process.execPath), '..', 'lib', 'node_modules', 'npm', 'bin', 'node-gyp-bin',
+    ),
   );
 
   // Add global bin folder, as some packages depend on a globally-installed
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -121,6 +121,19 @@ export async function executeLifecycleScript(
   // split up the path
   const pathParts = (env[constants.ENV_PATH_KEY] || '').split(path.delimiter);
 
+  // Include node-gyp version that was bundled with the current Node.js version,
+  // if available.
+  pathParts.unshift(
+    path.join(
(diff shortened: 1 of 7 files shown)


bestander:
Love the solution, thanks for considering so many edge cases

SimenB:
Can this get a patch release asap zulu honolulu? 😄

src/util/execute-lifecycle-script.js
```diff
@@ -116,8 +121,22 @@ export async function executeLifecycleScript(
...
-  pathParts.unshift(path.join(__dirname, '..', '..', 'bin', 'node-gyp-bin'));
+  // Include node-gyp version that was bundled with the current Node.js version,
+  // if available.
+  pathParts.unshift(
+    path.join(
+      path.dirname(process.execPath), 'node_modules', 'npm', 'bin', 'node-gyp-bin',
+    ),
+  );
+  pathParts.unshift(
+    path.join(
```
SimenB: this stuff can use `path.resolve` which gets rid of the `..` from the resulting string, and you can pass a single arg with `/`. Yes, that works on windows.

```sh-session
$ node -pe "require('path').win32.resolve('/root', 'some/path/../slashes')"
\root\some\slashes
```

SimenB: so this can be `path.resolve(path.dirname(process.execPath), '../lib/node_modules/npm/bin/node-gyp-bin')`

Daniel15: `path.join` gets rid of the `..` too 😃 

I used `path.join` with all the pieces rather than hard-coding a string with `/` since it seemed cleaner. I guess either way is fine.

SimenB: >path.join gets rid of the .. too 

Didn't know! Nice

```sh-session
$ node -pe "require('path').win32.join('/root', 'cool', 'some/path/../slashes', '..')"
\root\cool\some
```

Would you look at that, it also fixes the slashes! You learn something every day

bcomnes:
I thin this might have complicated rebuilding native dependencies for electron: https://github.com/electron-userland/electron-builder/issues/1542

Still trying to track down the source of the problem, but I can't rebuild sqlite3 for electron now when installing with 0.24.0 and greater.

Daniel15:
@bcomnes -  I think there's some issue specific to `node-pre-gyp`. Unfortunately I only tested with regular `node-gyp`.

bcomnes:
@Daniel15 Do you think the fix should live there?  I'm not terribly familiar with the internals of pre-gyp.

bcomnes:
@springmeyer do you have any idea why these changes might confuse or break the way node-pre-gyp works?

springmeyer:
@bcomnes the error you refer to in the electron ticket: `Failed to execute 'node-gyp clean' (Error: spawn node-gyp ENOENT)` indicates that `node-gyp` is not being found by `node-pre-gyp`. So it must be that `node-gyp` is not being detected by any of the checks at https://github.com/mapbox/node-pre-gyp/blob/aecb0dad52e6c185f4380664cd8019293fbf0322/lib/util/compile.js#L20-L54. So, can you figure out where `node-gyp` is on your system and figure out why none of the checks in node-pre-gyp are finding it? Perhaps the way that this PR searches for `node-gyp` might need to be added to node-pre-gyp?

bcomnes:
I'll dig into it.  From the sound of the changes in yarn, the only location of node-gyp would be in npm's node_modules.

springmeyer:
@bcomnes sounds good. A quick glance at this PR shows `path.dirname(process.execPath), 'node_modules', 'npm', 'bin', 'node-gyp-bin'` which might need to be added to node-pre-gyp since I see that https://github.com/mapbox/node-pre-gyp/blob/aecb0dad52e6c185f4380664cd8019293fbf0322/lib/util/compile.js#L48-L51 is finding `node-gyp` at a slightly different (less likely?) location inside the `npm` install.

bcomnes:
Sounds right.  Want a PR?  I might be able to get one out today or tomorrow.

springmeyer:
Another idea: maybe someone here would be interested in writing a separate module that:

  - Is specifically focused on finding the `node-gyp` binary in likely locations
  - Has great tests
  - Both yarn and node-pre-gyp could depend on

springmeyer:
> Sounds right. Want a PR? I might be able to get one out today or tomorrow.

Happy to review a PR over at node-pre-gyp. Thanks again for the ping here.

Daniel15:
The paths that Yarn uses to look for node-gyp are based on my experience with it:
 - `node_modules/npm/bin/node-gyp-bin` on Windows (eg. Node is at `C:\Program Files\nodejs\node.exe`, node-gyp is in `C:\Program Files\nodejs\node_modules\npm\bin\node-gyp-bin\`
 - `../lib/node_modules/npm/bin/node-gyp-bin` on other platforms
   - On Debian: `/usr/bin/nodejs` for Node, `/usr/lib/node_modules/npm/bin/node-gyp-bin/` for node-gyp
   - In the Linux binary tarball from nodejs.org: `bin/node` for Node, `lib/node_modules/npm/bin/node-gyp-bin/` for node-gyp

Yarn adds these paths to the `PATH`, so any command being executed should find it. @springmeyer  - Does `node-pre-gyp` not look on the `PATH` for node-gyp?

The only special case is when you do **not** have npm installed, in which case Yarn needs to install its own version of `node-gyp`. The logic in Yarn only checks if the command contains `node-gyp` when running that logic. It probably needs to check for `node-pre-gyp` too. However, that logic only applies when you do not have npm installed.

Looking at the code that @springmeyer linked to, it sounds like you could fix this by setting the `npm_config_node_gyp` environment variable.

bcomnes:
I have to head home now and I haven't finished testing a fix yet.  Perhaps tomorrow.

bcomnes:
I just tried searching the path for `node-gyp` when called from electron-builder's `install-app-deps` launched from yarn, and it doesn't work.  Its not getting the path generated by yarn for some reason.

The next solution I will try is resolving the node-gyp path from npm in the exact same way as yarn, as @springmeyer  originally suggested. I'm confident this will fix the current problem.

But...

This brings me to a question for @Daniel15: Can you help me understand what problem this change solves by not depending on `node-gyp` like `npm` does?  It seems like this change causes some problems not mentioned yet:

- You are correct that packages that need node-gyp should depend directly on node-gyp, in theory.
- Many packages do not though because npm includes it.  I believe this was one of npm's original sins that they just decided to live with due to breakage it would cause to try to remove it.
- Yarn adding an optional and silent peer dependency on npm is more of a sin than just shipping `node-gyp` with yarn.
- Promoting additional global dependencies is more of a sin than yarn shipping `node-gyp` as a dependency. 
- Shipping `node-gyp` is not an expensive dependency for yarn in any way that I can see.

I'm cool either way, but those are just my thoughts now that I understand the situation a little deeper.

bcomnes:
I just PR'd a possible fix for node-pre-gyp: https://github.com/mapbox/node-pre-gyp/pull/292/files

That should fix the issue, or if you like the argument I made above, reverting back to yarn shipping node-gyp would also put us back into a working state.

Daniel15:
> This brings me to a question for @Daniel15: Can you help me understand what problem this change solves by not depending on node-gyp like npm does?

The standalone .js build of Yarn nas **never** bundled node-gyp. This meant that node-gyp stuff never worked with it. The primary purpose of this pull request was to make node-gyp work the same way on all versions of Yarn - Both the bundled .js build as well as the tarball/installer version.

Additionally, the Yarn tarball now contains the bundled .js file rather than all the raw Yarn JS + `node_modules`. This significantly decreased the installation footprint of Yarn, allowing Yarn to be used in places where the developers are space conscious (this is how we were able to have Yarn added to the official Node.js Docker image, for example):


However, it means we can't easily bundle node-gyp. It would need to be a separate JS file in the bundle, including **all** its dependencies, defeating one of the main gains of the combined .js file.

It's worth noting that Yarn will **automatically** install node-gyp if it's not available on your system. Essentially it runs `yarn global add node-gyp` if `node-gyp` is required and not on the path.

> Many packages do not though because npm includes it. 

This is a fixable problem 😃

bcomnes:
Thanks for the info, I think I understand now.

AquiGorka:
This solves it: `npm i -g node-gyp`

PR #3244, opened 2017-04-24
Adds yarn create

arcanis:
**Summary**

This PR implements https://github.com/yarnpkg/rfcs/pull/55

**Test plan**

```
$> yarn create react-app
```

### DIFF
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -17,7 +17,7 @@ export async function run(
   flags: Object,
   args: Array<string>,
 ): Promise<void> {
-  const [builderName, rest] = args;
+  const [builderName, ... rest] = args;
 
   if (!builderName) {
     throw new MessageError(reporter.lang('invalidPackageName'));
@@ -27,6 +27,6 @@ export async function run(
     throw new MessageError(reporter.lang('createUnsupportedScope'));
   }
 
-  await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${args[0]}` ]);
-  await spawn(`${getGlobalBinFolder(config, flags)}/yarn-create-${args[0]}`, [], { stdio: `inherit` });
+  await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${builderName}` ]);
+  await spawn(`${getGlobalBinFolder(config, flags)}/yarn-create-${builderName}`, rest, { stdio: `inherit` });
 }
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -4,12 +4,11 @@ import {resolve} from 'path';
 
 import type Config from '../../config.js';
 import {MessageError} from '../../errors.js';
-import {registries, registryNames} from '../../registries/index.js';
+import {registryNames} from '../../registries/index.js';
 import type {Reporter} from '../../reporters/index.js';
 import * as child from '../../util/child.js';
 import * as fs from '../../util/fs.js';
-import {run as runGlobal, getBinFolder as getGlobalBinFolder} from './global.js';
-import {run as runRun} from './run.js';
+import {run as runGlobal} from './global.js';
 
 export function setFlags() {}
 
@@ -32,7 +31,7 @@ export async function run(
   const packageName = builderName.replace(/^(@[^\/]+\/)?/, '$1yarn-create-');
   const commandName = packageName.replace(/^@[^\/]+\//, '');
 
-  await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${packageName}` ]);
+  await runGlobal(config, reporter, {}, ['add', `yarn-create-${packageName}`]);
 
   for (const registry of registryNames) {
     const packagePath = `${config.globalFolder}/${config.registries[registry].folder}/${packageName}`;
@@ -57,7 +56,7 @@ export async function run(
       throw new MessageError(reporter.lang('createInvalidBin', packageName));
     }
 
-    await child.spawn(binPath, rest, { stdio: `inherit` });
+    await child.spawn(binPath, rest, {stdio: `inherit`});
     return;
   }
 
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -28,5 +28,5 @@ export async function run(
   }
 
   await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${args[0]}` ]);
-  await spawn(`${getGlobalBinFolder(config, flags)}/yarn-create-${args[0]}`, []);
+  await spawn(`${getGlobalBinFolder(config, flags)}/yarn-create-${args[0]}`, [], { stdio: `inherit` });
 }
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -1,0 +1,32 @@
+import type Config from '../../config.js';
+import {MessageError} from '../../errors.js';
+import type {Reporter} from '../../reporters/index.js';
+import {spawn} from '../../util/child.js';
+import {run as runGlobal, getBinFolder as getGlobalBinFolder} from './global.js';
+import {run as runRun} from './run.js';
+
+export function setFlags() {}
+
+export function hasWrapper(): boolean {
+  return true;
+}
+
+export async function run(
+  config: Config,
+  reporter: Reporter,
+  flags: Object,
+  args: Array<string>,
+): Promise<void> {
+  const [builderName, rest] = args;
+
+  if (!builderName) {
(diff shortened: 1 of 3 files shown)


arcanis:
Some notes:

- I made it so `yarn create react-app` install `create-react-app` (ie. the `create` is automatically added to the specified package name, which means that only `create-*` packages will be installable through this command).

- It currently doesn't work with `create-react-app`, because the tool currently doesn't support running on an initialized path (and in the current schema we need to initialize it, because we need to add the `create-react-app` package to the devDependencies).

arcanis:
Updated the pull request:

- It now install the package globally.
- It prefixes all package name with `yarn-create-`.

So `yarn create <pkg-name>` effectively becomes an alias for `yarn global add <pkg-name> && $(yarn global bin)/yarn-create-<pkg-name>`.

I've also wrote and published a test package: `yarn-create-testcase`, than can be invoked via `yarn create testcase`. When run, it automatically generates into the current directory a Yarn testcase that can be run using `yarn start`. The idea is to make it easier for our users to report bugs via reproducible testcases. It obviously won't work on Windows (yet), but it might still be useful :)

src/cli/commands/create.js
```diff
@@ -0,0 +1,30 @@
...
+  flags: Object,
+  args: Array<string>,
+): Promise<void> {
+  const [ builderName, rest ] = args;
+
+  if (!builderName)
+    throw new MessageError(reporter.lang('invalidPackageName'));
+
+  if (builderName.match(/^@/))
+    throw new MessageError(reporter.lang('createUnsupportedScope'));
```
cpojer: add curly braces around if statements.

src/cli/commands/create.js
```diff
@@ -0,0 +1,30 @@
...
+  return true;
+}
+
+export async function run(
+  config: Config,
+  reporter: Reporter,
+  flags: Object,
+  args: Array<string>,
+): Promise<void> {
+  const [ builderName, rest ] = args;
```
cpojer: nit: no spaces after `[` and before `]`.

src/cli/commands/create.js
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!builderName)
+    throw new MessageError(reporter.lang('invalidPackageName'));
+
+  if (builderName.match(/^@/))
+    throw new MessageError(reporter.lang('createUnsupportedScope'));
+
+  await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${args[0]}` ]);
```
cpojer: what does this do if the module is already installed? Will it upgrade the package?

src/reporters/lang/en.js
```diff
@@ -132,6 +132,8 @@ const messages = {
   linkUnregistered: 'Unregistered $0.',
   linkUsing: 'Using linked module for $0.',
 
+  createUnsupportedScope: 'Scoped packages cannot be installed via yarn create for now - read <github issue> to know why.',
```
cpojer: Instead of adding an error here, I'd probably just support it out of the box. We discussed 

```
yarn create @foo/react-app
```

would translate to: `@foo/yarn-create-react-app`. Can we implement it this way?

src/cli/commands/create.js
```diff
@@ -0,0 +1,30 @@
+import type Config                                            from '../../config.js';
+import {MessageError}                                         from '../../errors.js';
+import type {Reporter}                                        from '../../reporters/index.js';
+import {spawn}                                                from '../../util/child.js';
+import {run as runGlobal, getBinFolder as getGlobalBinFolder} from './global.js';
+import {run as runRun}                                        from './run.js';
```
arcanis: Nope! Just an old habit, I'm gonna fix this.

src/config.js
```diff
@@ -579,6 +580,9 @@ export default class Config {
 
       await fs.writeFilePreservingEol(loc, JSON.stringify(object, null, indent || constants.DEFAULT_INDENT) + '\n');
     }
+
+    // clear the cache, otherwise the new manifests can't be loaded through this config
+    this.cache = map();
```
cpojer: I'm not sure why this is needed here? Is this related to yarn create?

src/cli/commands/create.js
arcanis: Yep

src/reporters/lang/en.js
arcanis: Because all binaries are merged together inside the `.bin` directory, executing a scoped binary might not call the right one. I'll fix this before merging, but I figured it might be better to have a prototype without this feature so we can agree on the feature's most basic behavior.

src/config.js
arcanis: Not anymore, I'll remove it (basically, the old behavior was to execute a script. However, in order to get this script, I had to clear the package.json cache, since it had stored before during the then-local `add`).

src/reporters/lang/en.js
cpojer: Yeah, maybe we can shell out to the binary specified in package.json of the package as we discussed in the team sync today :)

src/cli/commands/create.js
```diff
@@ -0,0 +1,32 @@
+import type Config from '../../config.js';
```
cpojer: This needs `/* @flow */` otherwise the type imports won't do anything.

cpojer:
Lovely, this feature is so nice and simple to use, I'm really looking forward to shipping it.

It doesn't seem like this works right now:

```
yarn create react-app my-app
```

should create an app in `process.cwd() + path.sep + 'my-app'` just like `create-react-app my-app` would do. Can we support that and pass options forward?

Other than that and the scoped packages issue, I think we are good here.

arcanis:
I've added support for forwarding parameters, but unfortunately it breaks `yarn create react-app` (with no args), because the `create-react-app` package require a path. The only way I can fix this would be to parse the command line arguments to see if there's one, but that would require to keep a list of all supported options, which I'd like to avoid.

A better solution imo would be for CRA to either support invoking the binary without argument or, if they don't want to toggle this behavior by default, to add a `--use-cwd` option that would enable it if there's no path specified on the command line.

arcanis:
Worth noting that currently, `create-react-app` installs its dependencies via NPM, even when installed via Yarn. I've looked on their issues and apparently Yarn is supported, but for some reason it isn't correctly picked up.

cpojer:
@arcanis can we forward the param through `yarn create` to the `yarn-create-react-app` wrapper and then in the wrapper decide what to do and make `create-react-app` work? That way we pass arguments to yarn-create plugins and we can make create-react-app work.

src/cli/commands/create.js
```diff
@@ -0,0 +1,32 @@
...
+
+  if (!builderName) {
+    throw new MessageError(reporter.lang('invalidPackageName'));
+  }
+
+  if (builderName.match(/^@/)) {
+    throw new MessageError(reporter.lang('createUnsupportedScope'));
+  }
+
+  await runGlobal(config, reporter, {}, [ 'add', `yarn-create-${builderName}` ]);
```
cpojer: I just wanna make sure, `yarn global add <package>` will always either install the latest version of update to the latest version of that package and all its dependencies, right?

arcanis:
>  can we forward the param through yarn create to the yarn-create-react-app wrapper and then in the wrapper decide what to do and make create-react-app work? 

But then the wrapper needs to have a list of the option supported by `create-react-app`. Consider the following command line:

```
--foobar /tmp/a/b
```

You can't know if there is a path or not unless you know that the `foobar` option accepts (or not) an argument.

arcanis:
I just pushed a commit that should implement the scope support. With this commit, the feature should be ok. Some refinements will need to be made, especially to add unit tests, but the basics are there.

Current creators (mostly useful for testing purposes):

- `yarn-create-testcase`
- `yarn-create-react-app`
- `@arcanis/yarn-create-hello`

src/reporters/lang/en.js
```diff
@@ -132,6 +132,9 @@ const messages = {
   linkUnregistered: 'Unregistered $0.',
   linkUsing: 'Using linked module for $0.',
 
+  createInvalidBin: 'Invalid bin entry found in package $0.',
+  createMissingPackage: 'Package not found - this is probably an internal error, and should be reported on Github.',
```
cpojer: should this include a link to some specific issue tracker? "On GitHub" could mean anything.

cpojer:
This is awesome @arcanis, looking forward to shipping this!

arcanis:
Seems good, let's merge it. I'm gonna work on `yarn-create-next-app` & `yarn-create-react-native-app`

PR #3248, opened 2017-04-24
extract the reporter out of integrity checker

voxsim:
**Summary**
Integrity-checker should not have side effects and we should reporting warning only in the check command.

**Todo list**
- [x] add some testing for the check command
- [x] Right now I report an error with the reason why integrity matches failed in the check command, this is incorrect, I should use a warning

**Test plan**
See above.

### DIFF
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -3,7 +3,6 @@
 import type Config from './config.js';
 import type {LockManifest} from './lockfile/wrapper.js';
 import type {RegistryNames} from './registries/index.js';
-import type {Reporter} from './reporters/index.js';
 import * as constants from './constants.js';
 import {registryNames} from './registries/index.js';
 import * as fs from './util/fs.js';
@@ -47,15 +46,11 @@ type IntegrityFlags = {
 export default class InstallationIntegrityChecker {
   constructor(
     config: Config,
-    reporter: Reporter,
   ) {
     this.config = config;
-    this.reporter = reporter;
   }
 
   config: Config;
-  reporter: Reporter;
-
 
   /**
    * Get the location of an existing integrity hash. If none exists then return the location where we should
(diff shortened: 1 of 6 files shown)


voxsim:
I think I am finished, @bestander let me know if you like the pr and if I understand correctly what you mean :P

bestander:
Thanks, Simon!

src/integrity-checker.js
```diff
@@ -167,32 +162,58 @@ export default class InstallationIntegrityChecker {
...
+    return null;
+  }
+
+  async _compareIntegrityFiles(
+    actual: IntegrityFile,
+    expected: ?IntegrityFile,
+    checkFiles: boolean,
+    locationFolder: string): Promise<string> {
+    if (!expected) {
+      return Promise.resolve('EXPECTED_IS_NOT_A_JSON');
```
bestander: nit: I think flow enums would be a better fit here

src/integrity-checker.js
```diff
@@ -167,32 +162,58 @@ export default class InstallationIntegrityChecker {
...
+    }
+    if (!compareSortedArrays(actual.linkedModules, expected.linkedModules)) {
+      return Promise.resolve('LINKED_MODULES_DONT_MATCH');
     }
     if (!compareSortedArrays(actual.flags, expected.flags)) {
-      this.reporter.warn(this.reporter.lang('integrityFlagsDontMatch'));
-      return false;
+      return Promise.resolve('FLAGS_DONT_MATCH');
```
bestander: Should all these errors be a reject rather than resolve?

voxsim:
@bestander I am sorry I really didn't know that with babel and async/await you can return something and this is 'automagically' wrapped by Promise.resolve O.o thanks for let me work on this and learn something new. I am reading how flow works, I will push something soon ;)

voxsim:
done :D

bestander:
Nice work!

PR #3250, opened 2017-04-25
Move integrity artifacts tracking to separate method - fixes #3247

### LINKED ISSUE #3247: .yarn-integrity artifacts gets lost after add
bug

```
yarn add node-sass@^3.13.0
tail node_modules/.yarn-integrity

  "artifacts": {
    "node-sass@3.13.1": [
      "vendor",
      "vendor/darwin-x64-48",
      "vendor/darwin-x64-48/binding.node"
    ]
  }

yarn add left-pad

tail node_modules/.yarn-integrity

"artifacts": {}

# !however

yarn remove left-pad

tail node_modules/.yarn-integrity

"artifacts": {
    "node-sass@3.13.1": [
      "vendor",
      "vendor/darwin-x64-48",
      "vendor/darwin-x64-48/binding.node"
    ]
  }
```

artifacts should remain after add operation

yarn master 0.24.0-0

kittens:
**Summary**

Moves the artifact retrieval to a separate method that is always called. Has the downside that we're reading the integrity file twice, we could cache it but since the integrity checker is going to be refactored it's probably not worth it. I've added a regression test to ensure `artifacts` is retained.

**Test plan**

Added a test


bestander:
Awesome work on this, @kittens, thanks

PR #3268, opened 2017-04-27
support pattern matching with list command

### LINKED ISSUE #2098: Add pattern matching to `ls` lookup

Feature

Yarn returns exact matches.

Ideally, an expression could be passed to list all matching values.

```bash
# current
yarn ls gulp
└── gulp@3.9.1
✨  Done

yarn ls gulp*
✨  Done

# improvement
yarn ls gulp*
└── gulp@3.9.1
└── gulp-rev@7.1.2
└── gulp-xo@0.14.0
✨  Done
```

Continued from #1599

helloandre:
**Summary**

fixes #2098

adds the ability to `list` using patterns:

```
yarn list "gulp*"
```

this brings `yarn list` more in line with the documentation that it "mimics the expected Unix behavior of listing" by supporting globs. should the documentation be updated to explicitly call this out?

**Test plan**

added tests for maintaining the current behavior of exact matching as well as patterned matching.

previous behavior (run against this repo):

```
> yarn list gulp
yarn list v0.23.2
└─ gulp@3.9.1
✨  Done in 0.75s.
```

with this change:

```
> bin/yarn list "gulp*"
yarn list v0.24.0-0
├─ gulp-babel@6.1.2
├─ gulp-if@2.0.2
├─ gulp-match@1.0.3
├─ gulp-newer@1.3.0
├─ gulp-plumber@1.1.0
├─ gulp-sourcemaps@2.4.1
├─ gulp-util@3.0.8
├─ gulp-watch@4.3.11
├─ gulp@3.9.1
└─ gulplog@1.0.0
✨  Done in 0.91s.
```


arcanis:
That's an interesting idea. Three remarks:

- How does it handle the case where a sub-dependency is called gulp-toto? Does it print gulp-toto as being at the root of the project, or does it print the parent package of gulp-toto, and gulp-toto underneath?

- The pattern should probably only operate on the name part of the package (not its scope), since we've recently introduced a `--scope` option to restrict a package selection based on its scope (only on the `upgrade` command, tho, it might be a good occasion to implement it there as well).

- I feel like this should be an option, like `--pattern 'foo*'`. What do you think? Plus, as mentioned above, it would match the `--scope` option :)

helloandre:
to answer your questions:

 - this does not change the behavior of ignoring "dim" packages (see the `notDim` var). I.E. only packages that appear as a root node in a `yarn list` call.

 - similarly this does not change the behavior and only searches on `tree.name.slice(0, tree.name.lastIndexOf('@'));` I.E. the name of the package.

 - I could go either way with this. it seems that across other established cli tools, some do it "inline" and some do it with a flag. For this particular implementation, because the non-glob behavior is unchanged I think it is more intuitive to do it "inline" instead of with a separate flag, but as I'm the interloper here I understand if I'm overruled.

bestander:
I think we have just switched to micromatched from minimatch, may need a rebase

bestander:
@helloandre, that is a cool feature.
Feel free to reopen the PR using micromatch, minimatch is not a dependency anymore.

PR #3269, opened 2017-04-27
improved error message

bestander:
**Summary**

Something like "Bad hash. Expected X but got Y" is not really useful when debugging.
This PR changes the error message to include the file name.

** Test **

```
<break a sha1 in one of the dependencies>
yarn install v0.24.0-0
[2/4] 🚚  Fetching packages...
error Hashes don't match when extracting file "JSONStream-1.2.1.tgz". Expected "12aa5790e799481083b49b4b7fa94e23bae69bf9" but got "32aa5790e799481083b49b4b7fa94e23bae69bf9"
```

The snapshot test for some reason has `<url>:` error message format but in real application the `<url>:` gets dropped


szarouski:
While this PR is not merged, is there a way to find out what is triggering Bad hash error? Thank you.

bestander:
Find the hash in yarn.lock

szarouski:
In my case yarn.lock is not even created. I'll keep searching, thanks for your help.

bestander:
@szarouski then run `yarn install --verbose`, it will print all the GET request

PR #3284, opened 2017-04-29
Add Jenkins job for bumping version on website

Daniel15:
**Summary**
- Adds a build job to call the script that was added in https://github.com/yarnpkg/website/pull/472. This job will be executed as part of the build infra whenever a new version is released
- Updates the Chocolatey and Homebrew build jobs to run when this version bump job runs, rather than polling every 15 minutes

**Test plan**
Tested with a fork of the website repo

References https://github.com/yarnpkg/website/issues/471

### DIFF
--- a/jenkins_jobs.groovy
+++ b/jenkins_jobs.groovy
@@ -1,18 +1,39 @@
 // Jenkins build jobs for Yarn
 // https://build.dan.cx/view/Yarn/
 
-/**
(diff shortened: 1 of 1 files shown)


Daniel15:
Just realised that I can't rely on `yarn-version` running to kick off the Chocolatey and Homebrew build jobs, as there could be a delay between the website update being pushed and the website actually being deployed (depending on Netlify). I might reinstate the polling for now.

Issue #3003, opened 2017-03-28
Yarn global only takes ls but not list


szhongren:
I think this might be a small bug.

`yarn global list` returns `error Invalid subcommand. Try "add, bin, ls, remove, upgrade"`
`yarn global ls` works as expected.

Using yarn 0.21.3 with node 7.6.0 on Windows 10 Pro build 14393.953. I'd be happy to make a PR if someone can point me to instructions on running it from source with a debugger.

sapegin:
Yeah, `yarn global ls` and `yarn list` are inconsistent.

bestander:
I agree, send a PR.
We'll need to support ls for a while though with a warning that it goes away

szhongren:
Are the command line args passed to commander? I'm having trouble tracing it in a debugger.

PR #3308, opened 2017-05-03
Added a list sub-command for the yarn global command. Added unit test.

gonzofish:
This involved moving the list function out of the buildSubCommands so it could be called from ls & list.

Calling global ls will display a warning that global ls is deprecated and to use global list.

**Summary**
Changed the `yarn global ls` command to `yarn global list` to be inline with the non-global command, as per #3003.

**Test plan**

Tests were updated to ensure that the output of `yarn global ls` and `yarn global list` produced the same results. Running `yarn run test` verified the new test passed.

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -69,6 +69,8 @@ async function getBins(config: Config): Promise<Set<string>> {
   return paths;
 }
 
+
+
 function getGlobalPrefix(config: Config, flags: Object): string {
   if (flags.prefix) {
     return flags.prefix;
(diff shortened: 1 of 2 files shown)


gonzofish:
There was a timeout issue in each of the failed tests, but they were different tests that failed and they were not related to my change (as far as I can tell)...what should I do to remedy?

src/cli/commands/global.js
```diff
@@ -210,18 +227,17 @@ const {run, setFlags: _setFlags} = buildSubCommands('global', {
     flags: Object,
     args: Array<string>,
   ): Promise<void> {
-    await updateCwd(config);
-
-    // install so we get hard file paths
-    const lockfile = await Lockfile.fromDirectory(config.cwd);
-    const install = new Install({skipIntegrityCheck: true}, config, new NoopReporter(), lockfile);
-    const patterns = await install.init();
+    reporter.warn(`\`yarn global ls\` is deprecated. Please use \`yarn global list\`.`);
```
bestander: nice

PR #3311, opened 2017-05-03
Fix offline resolution

arcanis:
**Summary**

Fix #3221


bestander:
A test with absolute unix path would be awesome ;)

arcanis:
@bestander How would you do it? Adding a marker like `%%PATH%%` somewhere in the `package.json` file from the fixture, and replacing it at runtime from within the test?

bestander:
This is the example from create-react-app:
yarn add /path/create-react-app/packages/react-scripts/react-scripts-0.9.4.tgz

bestander:
Also replacing path in yarn.lock during test could work

PR #3319, opened 2017-05-04
Fixes offline resolution for absolute Windows paths

arcanis:
`c:/` is a protocol, apparently 😞


arcanis:
Fifth time's the charm.

bestander:
looks like it is fixed now, appveyor node 6 break is random

PR #3338, opened 2017-05-06
Use /bin/bash in shebang

### LINKED ISSUE #3321: Error "/usr/bin/yarn: [[: not found" with 0.24.0
Using the latest yarn version `0.24.0`
```
user@hostname:/tmp$ yarn --version
/usr/bin/yarn: 15: /usr/bin/yarn: [[: not found
0.24.0
```

Reporting a bug

When I use the yarn command, it works but I *always* get this :
```
/usr/bin/yarn: 15: /usr/bin/yarn: [[: not found
```
Then my command is executed normally.

- Using Debian (testing) with `4.9.0-2` kernel.
- Install the deb package for Yarn `0.24.0`. Not currently available on repository, find it [here](https://github.com/yarnpkg/yarn/releases/download/v0.24.0/yarn_0.24.0_all.deb)
- Use yarn command (eg. `yarn --version`)

It seems there is a problem with the Yarn "entrypoint" script. Did not investigate on it, I just report this before version `0.24.0` is pushed into APT repository.

I'm just sure, but the .deb package seems really light (580KB) comparing to previous versions (1.95MB). Maybe it is not generated well ?

Daniel15:
**Summary**
Use `/bin/bash` as the shebang, as the script may use Bash-specific features.

**Test plan**
Run `yarn` with `/bin/sh` symlinked to `/bin/dash` and ensure it still works.

Closes #3321

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 argv0=$(echo "$0" | sed -e 's,\\,/,g')
 basedir=$(dirname "$(readlink "$0" || echo "$argv0")")
 


myfreeweb:
This is a terrible idea, `/bin/bash` is only available on GNU/Linux. This would break yarn on all of my machines, since they run FreeBSD. The real solution is to use `[` instead of `[[`.

Daniel15:
@bestander Do you know why this was changed from `[` to `[[`?

bestander:
It started complaining on my mac after I merged this https://github.com/yarnpkg/yarn/pull/3245
No idea what I was doing there

Daniel15:
@myfreeweb Do you know how to fix #3260 without using `[[`? I'm not very familiar with shell scripting.

myfreeweb:
@Daniel15 use `=` instead of `-eq`. `=` is for strings, `-eq` is for numbers

Daniel15:
Sounds good. Let's fix it the proper way.

Issue #3336, opened 2017-05-05
yarn pack does not honor globs in "files" option of package.json


jrop:
**Do you want to request a *feature* or report a *bug*?** bug

**What is the current behavior?**

My package.json "files" option specifies globs as entries:

```txt
  "files": [
    "config/**/*.js",
    "lib/**/*.js"
  ]
```

When I run `yarn pack` it produces a `.tgz` file with the following entries:

```txt
package
package/.gitignore
package/config/.gitignore
package/files/.gitignore
package/package.json
```

However, when I run `npm pack`, it packs the following files instead:

```txt
package/package.json
package/lib/file1.js
package/lib/file2.js
package/lib/index.js
package/config/index.js
package/config/config1.js
package/config/config2.js
```

**If the current behavior is a bug, please provide the steps to reproduce.**
See above

**What is the expected behavior?**
See the `npm pack` output

**Please mention your node.js, yarn and operating system version.**
node: v7.7.3
yarn: v0.23.4
OS: macOS Sierra 10.12.2

jrop:
Keeping notes for myself:

yarn pack eventually calls: [this line](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/pack.js#L119) with the following filters:

```js
[ { base: '.',
    isNegation: true,
    regex: /^(?:\/(?=.)package\.json)$/i },
  { base: '.',
    isNegation: true,
    regex: /^(?:\/(?=.)readme[^\/]*?)$/i },
  { base: '.',
    isNegation: true,
    regex: /^(?:\/(?!\.)(?=.)(?:license|licence)+[^\/]*?)$/i },
  { base: '.',
    isNegation: true,
    regex: /^(?:\/(?!\.)(?=.)(?:changes|changelog|history)+[^\/]*?)$/i },
  { base: '.',
    isNegation: false,
    regex: /^(?:(?!\.)(?=.)[^\/]*?)$/i },
  { base: '.', isNegation: false, regex: /^(?:(?=.)\.[^\/]*?)$/i },
  { base: '.',
    isNegation: true,
    regex: /^(?:(?=.)config\/(?:(?!(?:\/|^)\.).)*?\/(?!\.)(?=.)[^\/]*?\.js)$/i },
  { base: '.',
    isNegation: true,
    regex: /^(?:(?=.)lib\/(?:(?!(?:\/|^)\.).)*?\/(?!\.)(?=.)[^\/]*?\.js)$/i },
  { base: '.',
[...]

PR #3339, opened 2017-05-06
Switch to micromatch: fixes yarnpkg/yarn#3336

jrop:
**Summary**

The micromatch library provides an inconsistent implementation of `.makeRe(...)` with the rest of it's interface.  It is known that `makeRe` does not produce Regular Expressions that can be trusted:

* isaacs/minimatch#67
* babel/babel#3914

This PR switches from minimatch => micromatch.

**Motivation**: #3336 😄 

**Test plan**

TODO:
* [x] Update tests and make them pass (before merge)

Tested so far:
```sh
$ alias my_yarn='node /github/yarn/lib/cli/index.js'
$ cd /my/other/project
$ my_yarn pack
$ tar -tf my-package-v1.0.0.tgz
#
# Output is as expected
#
```


jrop:
Unpredictable tests on CircleCI fail with "timeout" errors:

```sh
      
      at ontimeout (timers.js:386:14)
      at tryOnTimeout (timers.js:250:5)
```

This failure does not seem to be related to this PR, am I correct?

bestander:
Nice, thanks for a thorough explanation, tests and fix

jrop:
@bestander 😰  I just realized that the way I added the extra test causes it to fail on windows in AppVeyor, because I hard coded the path separator `lib/a.js` for example.  On windows the test fails because the path will be `lib\a.js`.  I pushed one more commit that changes the test to use `path.join('lib', 'a.js')`.  Does it need another PR?

bestander:
Yeah, could you send a new PR?

bestander:
@jrop when to expect the PR?
I'd like to cut 0.25 release

jrop:
@bestander Just submitted.  Sorry about that ☹️

bestander:
no worries :)

PR #3351, opened 2017-05-08
modify git ref parsing to not error on non-matching lines

### LINKED ISSUE #3325: yarn installation of self hosted git+ssh repos fail when there's a message returned from gitlab
yarn 0.23.4
macOs 10.12.4

When working with a "self-hosted" gitlab repo (such as talked about in #513), there's a case where gitlab is setup to output a warning message before returning the contents.

In certain settings gitlab server could possibly return a header prior to every interaction like:

```
*******************************************
Matt's Gitlab server, proceed at your own risk
*******************************************
abcde1234 refs/heads/blah
bdbadf4321 refs/heads/blah2
```

In these cases the following error is seen:

     TypeError: Cannot read property 'split' of undefined

When using a git+ssh://git@something/... etc. url

Stack trace:

```
at Function.parseRefs (...node/v6.10.3/lib/node_modules/yarn/lib/util/git.js:519:20)
at ...node/v6.10.3/lib/node_modules/yarn/lib/util/git.js:450:24
at next (native)
at step (node/v6.10.3/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30
at node/v6.10.3/lib/node_modules/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
```

rally25rs:
**Summary**

Fixes #3325 

Modified the parsing of Git ref lines to check it against a regexp before assuming the input line is a reference.
Since I was now using a regexp, I also use capture groups to extract the tag name and sha hash, insted of hte old implementation that was using string splits.

**Test plan**

Added unit tests for `Git.parseRefs()` to `__tests__/util/git.js`

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -31,6 +31,11 @@ const supportsArchiveCache: { [key: string]: boolean } = map({
   'github.com': false, // not support, doubt they will ever support it
 });
 
+// This regex is designed to match output from git of the style:
+//   ebeb6eafceb61dd08441ffe086c77eb472842494  refs/tags/v0.21.0
+// and extract the hash and tag name as capture groups
+const gitRefLineRegex = /^([a-fA-F0-9]+)\s+(?:[^/]+\/){2}(.*)$/;
+
 export default class Git {
   constructor(config: Config, gitUrl: GitUrl, hash: string) {
     this.supportsArchive = false;
@@ -423,7 +428,7 @@ export default class Git {
   }
 
   /**
-   * TODO description
+   * Parse Git ref lines into hash of tag names to SHA hashes
    */
 
   static parseRefs(stdout: string): GitRefs {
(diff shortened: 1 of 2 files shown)


src/util/git.js
```diff
@@ -434,14 +439,21 @@ export default class Git {
...
+      if (match) {
+        const [, sha, tagName] = match;
 
-      refs[name] = sha;
+        // As documented in gitrevisions:
+        //   https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html#_specifying_revisions
+        // "A suffix ^ followed by an empty brace pair means the object could be a tag,
+        //   and dereference the tag recursively until a non-tag object is found."
+        // In other words, the hash without ^{} is the hash of the tag,
+        //   and the hash with ^{} is the hash of the commit at which the tag was made.
```
TimvdLippe: TIL, great description!

rally25rs: haha thanks, I figured it was better than the old code which was "// TODO: find out why this is necessary. idk it makes it work..." 😆

bestander:
Thanks, @rally25rs!

mmucklo:
Seems to have fixed the issue.

PR #3359, opened 2017-05-09
feat(type): add support for new link: dependency type

mgcrea:
**Summary**

Following the [accepted 0000-link-dependency-type.md RFC](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-link-dependency-type.md), and as discussed in https://github.com/yarnpkg/rfcs/pull/34, here is a rebase of my previous work implementing both:

- Support for a new `link:` specifier
- CLI flag `--linkFileDependencies` to override `file:` specifier handling.

**Test plan**

Did not port previous unit test back for now as the idea is to first discuss the implementation with @arcanis and @bestander. Code is very likely to change. Will happily add unit tests once the implementation looks OK. 

After rebasing the PR, tests are currently failing (master is failing too so probably ok).


bestander:
Awesome!
Thanks

arcanis:
Great PR! Seems good to merge, except for the way you're accessing the command line parameter -
 you should use the `config` object to store it, and retrieve it from there later (unless there's a reason why you can't?). Otherwise, it won't work properly with the `.yarnrc` command line arguments :)

bestander:
ping @mgcrea, we'd love to have it in 0.25 that we plan to cut tomorrow

benlangfeld:
@arcanis / @bestander Is there precedent for that storage in config that we could follow?

arcanis:
Sure, you can take a look at [this commit](https://github.com/yarnpkg/yarn/commit/98095116f5d156f3e647e7b22b2ea4a042a35bfa#diff-9d9a6cd82f41984872a66a3ab0d440c4L30), for example. You just have to use `getOption` in the Config's `init()` method, and then you can access from the `config` object (usually available via `this.config`).

benlangfeld:
@arcanis Is [this](https://github.com/mgcrea/yarn/pull/1) what you had in mind?

arcanis:
@benlangfeld Not exactly, but not far! Look at the commit I just pushed if you're interested; basically, instead of adding a `linkFileDependencies` argument to the resolver, we just use the `config` object as source of truth.

@mgcrea I've added a test to make sure that everything works, and unfortunately it seems there's still a few issues (I quickly checked, and I think one of them is because the install command [is trying to read the manifest from the location](https://github.com/yarnpkg/yarn/blob/20e5779ec324b3103ef9cca14ce9f96d29fd93c6/src/cli/commands/install.js#L713). The other one is inside the check command, which fails because it is trying to make sure that the package path exists, even tho it might not have been created yet). Can you take a look at this?

bazyli-brzoska:
I've tested this branch and the package name is not inferred from the path:

```shell
$ yarn add link:../i18n-import-webpack-plugin --dev
yarn add v0.25.0-0
error Package "@0.0.0" doesn't have a "name".
```

I have to run the following instead:

```shell
yarn add i18n-import-webpack-plugin@link:../i18n-import-webpack-plugin --dev
```

The first command works with `file:` specifier.

arcanis:
Ping @mgcrea ? 🙂

mgcrea:
@arcanis I'll have a look today. Thanks for the patches.

mgcrea:
@arcanis any chance you could commit your `install-link` fixture? Looks like it's missing on the PR, would avoid unnecessary conflicts. Thanks!

mgcrea:
Did finish a first squash+rebase+update, had to add code since the functionality was broken by a few recent commits on master (mostly from the `cli/commands/check.js`), I've kept the commit separated for now to ease the [required patches](https://github.com/yarnpkg/yarn/pull/3359/commits/064ecbae66a9b7f1cae6223d9402bdfa1ec5fe03) since initial commit. Some tests seems a bit hacky, but tried to keep the refactoring to a minimum.

Looks like my existing workflow is still broken on master, so I'll have to dig a bit deeper to make sure I have the relevant unit tests.

bestander:
Thanks, @mgcrea

src/package-fetcher.js
```diff
@@ -37,6 +37,13 @@ export default class PackageFetcher {
     const dest = this.config.generateHardModulePath(ref);
 
     const remote = ref.remote;
+
+    // Mock metedata for linked dependencies
+    if (remote.type === 'link') {
```
bestander: Would it be a bit cleaner if LinkFetcher existed instead?

mgcrea: @bestander iirc at the time I started to work on the feature, I had to diverge earlier in the stack so there were no use for a LinkFetcher. Might be worth revisiting though.

mgcrea:
I think we're getting close to something "ready". I've fixed and added a unit-test for the issue encountered by @bazyli-brzoska. Test are OK, rebase done.

@bestander, regarding the `package-fetcher.js` patch, I tried again to use a `LinkFetcher` but it would have still required a lot of branching in `PackageFetcher` as we need to opt out cache entirely. So I decided against it (would lead to more LOCs & increased complexity).

Let me know if there is something else I can do.

bestander:
@mgcrea, I agree about the fetcher, makes sense to have an if block rather than more complexity.

bestander:
Amazing work, @mgcrea!
You've lead the whole feature from design to implementation.
I am sure quite a lot of people would want to use it if they know about it, do you want to post in the [blog](https://yarnpkg.com/blog/) about it?

bestander:
It is ready to merge although windows tests fail:

```
__tests__\commands\install\integration.js (120.573s)
  ● creates a symlink to a non-existing directory when using the link: protocol
    
    Expected value to equal:
      "../baz"
    Received:
      "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\yarn-install-link-1495898245160-0.05036491088610817\\baz\\"
```

mgcrea:
@bestander can fix the tests (add a platform ternary) if you want.

Blogpost is a great idea, I'll try to come up with something.

bestander:
If you could fix it, please go ahead

src/cli/commands/check.js
```diff
@@ -49,21 +49,33 @@ export async function verifyTreeCheck(
   const dependenciesToCheckVersion: PackageToVerify[] = [];
   if (rootManifest.dependencies) {
     for (const name in rootManifest.dependencies) {
+      const version = rootManifest.dependencies[name];
+      // skip linked dependencies
+      const isLinkedDepencency = /^link:/i.test(version) || (/^file:/i.test(version) && config.linkFileDependencies);
```
mgcrea: @bestander This is the last thing I don't really like, as regex testing feels a bit hackish, but I would need to change the parent method API to pass the proper type. So was unsure if it was worth it.

mgcrea:
I've fixed the tests for windows, (looks like the test are still failing there but for unrelated reasons).

bestander:
Thanks, yeah one test got broken earlier, investigating

zertosh:
https://github.com/zertosh/yarn-link-test is an example of how `link` breaks down due to Node's realpath module resolution (https://github.com/nodejs/node/issues/3402)

benlangfeld:
Much ❤️ to everyone involved here, particularly @mgcrea. You guys are awesome and this will make my team very happy. We owe you at least some drinks!

benlangfeld:
@bestander I guess this will now go in 0.26.0? By the recent cadence of minor releases, that would likely be cut within the next two weeks. Is that a fair estimate?

bestander:
We plan to cut on Monday.
There are a couple of hi pri issues we want to fix before cutting 0.26.

PR #3361, opened 2017-05-10
Allow prerelease versions when checking peer deps

### LINKED ISSUE #2760: [enhancement] Include prereleases (-beta.1) in peerDep semver ranges
This is a half-feature request, half-bug report.

Currently yarn (and npm) does not include prerelease versions of packages to satisfy semver ranges, unless those ranges explicitly ask for a prerelease version. For example, `react@16.0.0-alpha.2` does not satisfy the range `react >= 15` even though 16-dot-anything is greater than 15.

Install `react@16.0.0-alpha.2` and another package that has a peerDep on `react >= 15` (`react-helmet` for example). Run `yarn` which prints:
```
warning "react-helmet@4.0.0" has incorrect peer dependency "react@>=15.0.0".
```

With this requested enhancement, when doing peerDep checks, Yarn would consider `X.Y.Z-anything.N` to be a semver string that is greater than `X.Y.{Z-1}` and less than `X.Y.{Z+1}` (and `X.{Y+1}.0` and `{X+1}.0.0`). 

I believe this also addresses an issue that Facebook has been having internally between react and other packages that have a peerDep on react.

The `semver` package doesn't implement this behavior so we'd likely want to write a small module that does a more inclusive version check for peerDeps.

Node 7.6.0
Yarn 0.21.2
macOS 10.12

ide:
**Summary**
Uses lower-level APIs in node-semver to allow prerelease versions when checking peer deps. This honors the intent of library authors e.g. when they write a React component that asks for `"react": ">=15.0.0"` they're saying OK to `react@16.0.0` (partly because the React team specifically has said that if you don't have any warnings in version N, then version N+1 should work for you). By extension and modulo bugs, `react@16.0.0-alpha` works too if the user explicitly installs it.

Fixes #2760

**Test plan**
Added unit tests for the new semver utility.

Tested in a real project that uses React 16.0.0-alpha.11 and verified I didn't get any peer dep warnings that usually appear from libraries that ask for `"react": ">=15.0.0"`. Downgraded React to 0.14.0 (too low) and saw peer dep warnings come back as expected.

### DIFF
--- a/__tests__/util/semver.js
+++ b/__tests__/util/semver.js
@@ -1,0 +1,53 @@
+/* @flow */
+
+import {satisfiesWithPreleases} from '../../src/util/semver.js';
+
+const semver = require('semver');
+
+describe('satisfiesWithPreleases', () => {
+  it('matches the behavior of node-semver for non-prerelease versions', () => {
+    // true
+    expect(satisfiesWithPreleases('2.0.0', '>=1.0.0'))
+      .toBe(semver.satisfies('2.0.0', '>=1.0.0'));
+    expect(satisfiesWithPreleases('0.1.1', '^0.1.0'))
+      .toBe(semver.satisfies('0.1.1', '^0.1.0'));
+    expect(satisfiesWithPreleases('1.0.1', '~1.0.0'))
+      .toBe(semver.satisfies('1.0.1', '~1.0.0'));
+
+    // false
+    expect(satisfiesWithPreleases('0.2.0', '^0.1.0'))
+      .toBe(semver.satisfies('0.2.0', '^0.1.0'));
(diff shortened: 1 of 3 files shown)


bestander:
Thanks, @ide.
Would you fix the lint error please?

SimenB:
Can this be made to work with `engine` as well?

```sh-session
$ node --version
v8.2.0-rc.1
$ yarn
yarn install v0.24.6
[2/4] 🚚  Fetching packages...
error ansi-styles@3.0.0: The engine "node" is incompatible with this module. Expected version ">=4".
```

bestander:
@SimenB I think it is a good idea, would you send a PR?

PR #3365, opened 2017-05-10
Workspaces phase 2: executing commands

arcanis:
The current implementation allows to run a command inside the context of a workspace, regardless where the user is inside the project:

```
$> yarn workspace babel-runtime add babel-core
```


arcanis:
There might be flow / lint issues, I'll fix them before merging.

bestander:
Awesome!
Will look at it later today

bestander:
How about some tests?
See new test file in #3229

arcanis:
Will git see that these are the same files even if on different branches?

bestander:
no, you'll probably need to rebase and resolve conflicts after my branch is merged

bestander:
but you can probably rebase on top of my branch

arcanis:
If I do this it'll cause issues if you rebase yours later :/ I'll try to add a commit with your fixtures, and remove it before merging if git doesn't like it.

src/cli/commands/workspace.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const [workspaceName, ... rest] = args;
+
+  if (!Object.prototype.hasOwnProperty.call(workspaces, workspaceName)) {
+    throw new MessageError(reporter.lang('projectUnknownWorkspace', workspaceName));
+  }
+
+  try {
+    await child.spawn(process.argv[0], [process.argv[1], ... rest], { stdio: 'inherit' });
```
bestander: Smart move :)
But I don't think we want to necessarily shell out to start another node VM.
I would prefer us doing it via requires.

arcanis: There's quite a bit of logic that we currently can't run in the same process (most notably, everything related to the command line parsing, which is run in [src/cli/index.js](https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js)). That could be refactored, but I felt it could be too heavy for prototyping the command.

arcanis:
@bestander I've just rebased on master (ie on your merged PR), and added a commit that implements a project detection logic inside `install` (which is then used by `add` & `remove`). Could you review the implementation, except on the following points?

- [ ] I need to avoid computing the project path multiple times. It's really bad for performances. Ideally, it should only be computed once. Two solutions: either it is computed in the `Config.init()` function, or I can add a cache to `findProject`. I like the first solution better.

- [ ] I need to write some tests to make sure everything's acting correctly, but the manual tests I've made seem to indicate that everything looks right.

src/cli/commands/add.js
```diff
@@ -204,7 +204,7 @@ export async function run(
     throw new MessageError(reporter.lang('missingAddDependencies'));
   }
 
-  const lockfile = await Lockfile.fromDirectory(config.cwd, reporter);
+  const lockfile = await Lockfile.fromDirectory(config, config.cwd, reporter);
```
bestander: config.cwd looks redundant here, do we ever need it not from config?

src/lockfile/wrapper.js
```diff
@@ -86,9 +87,12 @@ export default class Lockfile {
     [key: string]: LockManifest
   };
 
-  static async fromDirectory(dir: string, reporter?: Reporter): Promise<Lockfile> {
+  static async fromDirectory(config: Config, dir: string, reporter?: Reporter): Promise<Lockfile> {
```
bestander: nit: this couples lockfile wrapper to config, not sure if it is a concern but maybe we'll want to decouple it back in the future

src/cli/commands/workspace.js
bestander: Got it, I think workspace command won't interfere with regular workflows, looks fine then

src/cli/commands/workspace.js
```diff
@@ -0,0 +1,48 @@
...
+  } catch (err) {
```
bestander: I think you don't want to gobble up the error here?

arcanis: The issue I had was that regular errors are already handled by the spawned child. Discarding the error here was to good way to prevent the error to be printed twice:

screen shot 2017-05-17 at 11 08 09 am

However, it also swallows the exit code ... what would you think about manually calling `process.exit` with the exit code returned by the child?

bestander: I think double error is ok-ish as long as it does not show the same error message twice.
I would vote against calling process.exit in multiple places in the code, better leave  only one exit code in index.js, this makes code hard to reuse

src/lockfile/wrapper.js
arcanis: Yep, it should be fairly easy. I think most tools currently stored inside the `config` object (such as `tryManifest`, etc) could be safely moved somewhere else (for example in miniyarn, I had a `util/yarn` file to store all such functions), so that the config would only contain the configuration data and nothing else.

src/cli/commands/add.js
arcanis: Not really, but it makes sense semantically to specify the path, otherwise it should be renamed to `fromConfig` - and as you mentioned in the comment below, it would probably be for the best to keep this function not too tightly coupled with the config object.

arcanis:
I fixed the issue I raised above, rebased, and implemented the `exec` command. However, I'm not sure I'll be able to write the tests right now:

- The current implementation forwards stdio to the spawn process (which is useful, because it allows the spawn processes to interact with the user, which would break otherwise). Unfortunately, it means that it bypasses the reporters, so I can't inspect to make sure that the output is the one I expect. I tried to workaround this by inspecting the state of the filesystem, but:

- The current implementation uses `process.argv[0]` and `process.argv[1]` in order to spawn a new instance of Yarn. Unfortunately, doing this cannot work when running Jest, since it will run Jest instead of Yarn.

src/config.js
```diff
@@ -521,6 +525,59 @@ export default class Config {
...
+
+    return null;
+  }
+
+  async resolveWorkspaces(root: string, patterns: Array<string>): Promise<{ [string]: { loc: string, manifest: Manifest } }> {
+    const compiledPatterns = filter.ignoreLinesToRegex(patterns);
+
+    // We need the ignoreFiles, not the keepFiles, because the patterns are exclusion patterns rather than the opposite
+    const ignoreBasenames = new Set(this.registryFolders);
+    const files = filter.sortFilter(await fs.walk(root, undefined, ignoreBasenames), compiledPatterns).ignoreFiles;
```
bestander: Let's merge this code with the one used in install command https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L238.

I think using glob is better than ad-hoc fs.walk because the JS community is quite familiar with glob notation.

arcanis: The code above is also using the glob syntax, because of `sortFilters`. I tried to use the same implementation from what we use for `pack`, but maybe `glob` is more efficient, yeah.

arcanis: Should be fine now 👍 Ok to merge?

src/cli/commands/add.js
bestander: In that case would it be better to pass no the full config to the fromDirectory function?
Technically you need only the `config.worktreeFolder || config.cwd` and is `dir` needed then?

src/config.js
```diff
@@ -521,6 +524,75 @@ export default class Config {
...
+      if (manifest) {
+        return manifest;
+      }
+    }
+
+    return null;
+  }
+
+  async findWorktree(initial: string): Promise<?string> {
+    let previous = null;
```
bestander: install command is guarded by `projectManifestJson.workspaces && this.config.workspacesExperimental`.
How about extracting it in a setting and checking it here?

src/cli/commands/install.js
```diff
@@ -660,7 +660,7 @@ export class Install {
     }
 
     // build lockfile location
-    const loc = path.join(this.config.cwd, constants.LOCKFILE_FILENAME);
+    const loc = path.join(this.config.worktreeFolder || this.config.cwd, constants.LOCKFILE_FILENAME);
```
bestander: The `this.config.worktreeFolder || this.config.cwd` expression is used in multiple places.
How about moving it to a function that would be self explaining?
Should we make `this.config.cwd` private?

bestander:
Great progress, @arcanis.
We need more tests ;)

bestander:
I am working on phase 3, there is a need to preserve the workspaces structure during whole installation from fetch phase till linking.
I think we should Manifest could be the right structure to hold workspaces, probably could be useful for this PR as well as workspaces cache.

bestander:
I'll link my PR here

src/cli/commands/install.js
arcanis: Hm, I think right now it's fine, but maybe that's something that could be improved later. I was thinking that maybe we could add another variable, `workspaceFolder`, that would contain the path to the top-most workspace, and use it instead of `cwd` about everywhere (the difference with `cwd` would be that if you're in a subdirectory, then the `workspaceFolder` would go up until the `package.json` file). This way, people wouldn't have the issue of adding dependencies to their subfolders instead of their project, if they're in the wrong directory.

But that would be a BC, so we should probably discuss this in an other PR/RFC

src/lockfile/wrapper.js
```diff
@@ -86,9 +87,10 @@ export default class Lockfile {
     [key: string]: LockManifest,
   };
 
-  static async fromDirectory(dir: string, reporter?: Reporter): Promise<Lockfile> {
+  static async fromDirectory(config: Config, dir: string, reporter?: Reporter): Promise<Lockfile> {
     // read the manifest in this directory
-    const lockfileLoc = path.join(dir, constants.LOCKFILE_FILENAME);
+    const lockfileLoc = path.join(config.worktreeFolder || config.cwd, constants.LOCKFILE_FILENAME);
```
arcanis: Good catch! This code became boggus after I refactored it a few times. I've now restored the old code, since we don't use `config` in this function anymore 👍

__tests__/commands/workspace.js
```diff
@@ -0,0 +1,34 @@
...
+    expect(await fs.exists(`${config.cwd}/packages/package-a/yarn.lock`)).toEqual(false);
+
+    await add(await makeConfigFromDirectory(`${config.cwd}/packages/package-b`, reporter), reporter, {}, ['right-pad']);
+
+    expect(await fs.exists(`${config.cwd}/node_modules/right-pad`)).toEqual(true);
+    expect(await fs.exists(`${config.cwd}/packages/package-b/node_modules/right-pad`)).toEqual(false);
+
+    expect(await fs.exists(`${config.cwd}/yarn.lock`)).toEqual(true);
+    expect(await fs.exists(`${config.cwd}/packages/package-b/yarn.lock`)).toEqual(false);
+  });
```
bestander: You also need to check in which package.json left-pad/right-pad entries were added.
And run `yarn add` from the root.

I think https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/workspaces-install.js is a better location for this test as you are not testing `workspace` command here

src/cli/commands/install.js
bestander: The question is - someone adds a new command, say `amazing-command`, do they need to pass `this.config.worktreeFolder || this.config.cwd` when reading lockfile?

It looks like we leak workspaces implementation detail into every command, seems like a red flag and besides, we have `config.worktreeFolder ||` repeated 13 times already. It definitely asks for encapsulation

src/config.js
bestander: Ping about the question above.

I think it is related to my previous comment regarding `config.worktreeFolder || config.cwd`. 
Should resolving workspaces and root folders be hidden behind the configuration flag and whether workspaces is enabled in the root manifest?

arcanis: Not sure what you mean with `projectManifestJson.workspaces` - if we're not in a workspace (ie. if there is no manifest with a `workspaces` key), then `findWorktree` will return `null`. Do you want it to do something different?

I'll add a check on the experimental flag, tho.

src/cli/commands/install.js
arcanis: Here are the options (I assume in these option that there is a `workspaceFolder` variable, but it work the same way if we don't):

1. Keep things as they are (eh), with the worktree definition being "the closest location where lies the `workspaces` field", and explicitely specify the locations from which we want to read the lockfile

2. Redefine `worktreeFolder` as being "the closest location where lies the lockfile", which would effectively `worktreeFolder || workspaceFolder || cwd` - the reason I didn't do this is that it means that a project with no workspaces would *internally* work just like a project with a single workspace (we wouldn't be able to distinguish a project that doesn't use workspaces like we currently can with `worktreeFolder === null`). That's not a problem strictly speaking, but we haven't specified this in our RFCs so we have to agree on this before I can do this.

3. Add a new variable (let's say `lockfileFolder`) that would be defined as `worktreeFolder || workspaceFolder || cwd`. The only issue I see with this is that it would mean yet another folder variable.

What do you think? My order of preference is 3 > 1 > 2

src/config.js
bestander: I guess I got my thoughts mixed up a bit :)
I think resolveWorkspaces should check for experimental flag and private field.
Then this check would propagate to be implicit everywhere workspaces are used.

src/cli/commands/install.js
bestander: `lockfileFolder` sounds quite descriptive, I like it, let's try 3

src/config.js
arcanis: Hm I was thinking of implementing this check inside `Config#init`, would that be fine with you? The problem with `resolveWorkspaces` is that it doesn't have to be executed in every case (for example, when adding a dependency we don't use it, since we only care about the current workspace and the base worktree).

src/cli/commands/install.js
arcanis: 👍

src/config.js
bestander: Yeah, definitely, you are correct

arcanis:
@bestander Everything should be fixed 👍

bestander:
Hey @arcanis, I think we lost this feature.
Running `yarn workspace jest-matcher-utils add isarray` does not run the add from workspace anymore.

JKillian:
Ran into this same bug today, file a short ticket over at https://github.com/yarnpkg/yarn/issues/4021

PR #3366, opened 2017-05-10
upgrade babel-core

voxsim:
I just upgrade babel-core.
I think we should adopt something like https://greenkeeper.io/ to update our dependencies automatically. I saw zeit using it for their projects: the bot just opens a pull request every time there is a possible upgrade.

what do you think? @bestander @arcanis @Daniel15


bestander:
Yeah, totally, give greenkeeper a try.

PR #3367, opened 2017-05-10
Implements the enable-meta-folder

arcanis:
**Summary**

The following PR adds an `enable-meta-folder` option. When active, the `.yarn-integrity` and `yarn-error.log` files will be written inside this folder rather than respectively inside `node_modules` and the cwd.

**Test plan**

I added a test to make sure the integrity file was correctly created.


voxsim:
@arcanis why we don't just add a 'meta-folder' option instead of enabling it? And in case we do not specify we fallback to default option like we do in the global-folder

voxsim:
I am asking because maybe there are some reasons, I am curious :)

arcanis:
You mean allowing people to set the path to whatever they want? The meta folder is supposed to store only things related to the internals of Yarn, so I'm not convinced that we should give users the ability to choose where it's stored, the same way we don't let them choose how should be named the yarn.lock file.

voxsim:
Mmh but if someone does a link of the meta folder, isn't it the same? 
If we don't want to allow people to set the path, why now do we give them a choice?

bestander:
I think this is a good first step.
If it is going to be a part of someone's workflow and making the location configurable would make sense then let's just add that feature then.
Thanks for bringing it up, @voxsim.
Feel free to drive a discussion forward.

PR #3374, opened 2017-05-10
Use yarn instead of npm everywhere is possible ...

voxsim:
… aka Eating your own dog food

**Summary**
This should fix https://github.com/yarnpkg/yarn/issues/153 but the real purpose of this pull request is understand if there are parts/commands that maybe don't fully work and in some way it forces developers to use npm instead of yarn.

**Test Plan**
There are a lot of tests, but I am not totally confident on this PR, please double check it before breaking things maybe infrastructurally.

### DIFF
--- a/scripts/build-dist-debug.sh
+++ b/scripts/build-dist-debug.sh
@@ -6,8 +6,8 @@ set -ex
 # potentially be useful for debugging purposes, but it's more bloated than the
 # regular distribution.
 
-npm run build
-npm pack
+yarn build
+yarn pack
 rm -rf dist-debug
 mkdir dist-debug
 mkdir -p artifacts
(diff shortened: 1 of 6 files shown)


bestander:
Awesome work!
Thanks a lot

scripts/build-dist-debug.sh
```diff
@@ -18,7 +18,7 @@ umask 0022 # Ensure permissions are correct (0755 for dirs, 0644 for files)
 tar -xzf pack.tgz --strip 1
 rm -rf pack.tgz
 # Change this to "yarn install --production" once #1115 is fixed
```
Daniel15: This comment doesn't really make sense any more, right? It should probably be removed 😛

voxsim: yeah you are right !

PR #3382, opened 2017-05-11
[PoC] Use worker-farm during linking phase.

cpojer:
**Summary**

I'm hacking and I figured I'll try to see if I can make the linking phase faster. With worker-farm, I'm getting a quite consistent ~10% win in the linking phase when installing react-native. However, it also comes with a bunch of downsides – it makes it harder to use the single file build as well as on CI systems this kind of stuff usually leads to trouble (from experience with Jest).

**Test plan**

#yolo

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "tar-stream": "^1.5.2",
     "uuid": "^3.0.1",
     "v8-compile-cache": "^1.1.0",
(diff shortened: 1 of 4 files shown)


bestander:
Let us know when ready :)

cpojer:
Unlikely that I will be able to finish this, unfortunately :(

PR #3389, opened 2017-05-12
Make yarn-bundle-entry.js executable

Daniel15:
**Summary**
This ends up as `bin/yarn.js` in the tarball. It should be executable

**Test plan**
N/A

References https://github.com/Homebrew/homebrew-core/pull/13516


ilovezfs:
:heart:

PR #3390, opened 2017-05-12
update eslint, eslint-config-fb-strict and other dependencies

voxsim:
**Summary**
I updated eslint, babel-eslint, eslint-config-fb-strict. This is preopedeutic for prettier.
For now I disabled sort-keys in the eslintrc otherwise we have to change *every* file drastically, maybe it is better to do incrementally.

Some info:
- Now arrow functions with one paramenter don't have parens
- I fixed some tests not catched by the old version of eslint

**Test plan**
Tests already in place


bestander:
lockfile is not updated

bestander:
Otherwise, great job!
Thank you

voxsim:
Some news:
- downgraded babel-eslint
- removed eslint-config-fbjs because the latest is incompatible with the latest of eslint-config-fb-strict (eslint-config-fb-strict picks an alpha version of eslint-config-fbjs 😱 )
- updated eslint-plugin-flow and eslint-plugin-react

now it seems to work and check the lockfile correctly O.o

@bestander when you have time, let me know if everything is correct

voxsim:
Now prettier works as expected, let me know if you prefer to do dirrent merges or I can push it here ;)

bestander:
Thanks, @voxsim, great work!

Issue #610, opened 2016-10-11
Better `yarn publish`


cpojer:
There was no time to polish `yarn publish` pre-release. I think we can do so much more and provide a fantastic experience (on top of the existing publish functionality).

Let's discuss what the ideal version of `yarn publish` may look like.

Inspiration: https://github.com/sindresorhus/np cc @sindresorhus

suitupalex:
Really love all the checks and balances `np` runs, but I don't like how it and `yarn` currently bump the version right before publish time.

Just to put out my personal preference on bumping a package, my typical workflow is as follows:

- Create a branch `release/vX.x.x` where the last commit is a version bump commit in `package.json`.
- CI does its thing and reviewers approve the release.
- A rebase merge is performed onto `master`.
- CI does its thing on `master` just in case.
- A `git` tag is applied onto the version bump commit.
- CI/CD does its thing on the new tag.
- `npm publish` is finally run on the new master.

I feel like publishing before the version bump is in source control is a little backwards. Maybe we can have  a separate `yarn bump`? It'd also be useful for private projects that still keep track of their versioning in `package.json` but don't need to publish to a registry.

grempe:
I also agree that bumping version in the working dir without committing that change is not the right flow. Yarn publish should require a clean working dir if in git so it doesn't pick up WIP.

Important side note, for me `yarn publish` didn't work at all.  It asked me what version I wanted to bump to and then hung until ctrl-c (npm publish worked immediately after this):

```
$ yarn publish
yarn publish v0.16.0
[1/4] Bumping version...
info Current version: 2.0.2
question New version: 2.0.3
info New version: 2.0.3
[2/4] Logging in...
[3/4] Publishing...
^C
```

MikeRalphson:
>Important side note, for me yarn publish didn't work at all. It asked me what version I wanted to bump to and then hung until ctrl-c

+1 for this, Windows 10, Node 6.9.0, Yarn 0.16.0 - except I don't have `npm` installed... :(

JulianLeviston:
It'd be really sweet to have some better logging for this process as it does what it does... so maybe if some step is taking longer than a second, it could log what it's been doing / what it will do, perhaps? Or at least output some form of logging or have a trace mode so we can debug where things are going a bit awry?

Waize:
Hi, I've investigated this a bit and found that there is a mistake in reading the stream of the tgz-file or directory, depending on what you gave as arguments. 

[...]

PR #3391, opened 2017-05-12
yarn publish not ready for CI

bestander:
Until #610 is fixed we have to use npm publish on CI


bestander:
self merging, trivial

Issue #3329, opened 2017-05-05
Creating yarn.lock file after running "yarn -v"


g-plane:
**Do you want to request a *feature* or report a *bug*?**

Feature request

**What is the current behavior?**
After I ran `yarn -v`, it will generate `yarn.lock` in a non-yarn project directory.

**If the current behavior is a bug, please provide the steps to reproduce.**

Run `yarn -v` in a non-yarn project directory which does not include `yarn.lock` file.

**What is the expected behavior?**
Just show the version of yarn in CLI without creating `yarn.lock` file if I run `yarn -v`.

**Please mention your node.js, yarn and operating system version.**
node.js: v7.9.0
OS: Windows 10 x64

rally25rs:
Verified on Yarn 0.23.4

I suspect that this happens because the "default" `yarn` command does an `install`. A `yarn install -v` would create a lock file. Probably need to add a check for no explicit command, and the only option is `-v`.

rally25rs:
@bestander triage: Duplicate of #2710

g-plane:
In fact, to display version of Yarn we must run `yarn -V` or `yarn --version`. But in most of the time, we are used to using `-v` to do that.

bestander:
I agree, this is annoying behavior

bestander:
Send a PR for Yarn `install` to not create a yarn.lock if there is no package.json in the folder.
Also maybe send a PR to make -v -and -V to be the same, I get confused all the time. But that one may be debatable because of some unix CMD standarfds

g-plane:
Personally, I prefer to making `-v` and `-V` to be the same. May I send a PR?

bestander:
Please do

Issue #2710, opened 2017-02-17
Unexpected side effect of `yarn -v`: it runs yarn / yarn install


thomasklein:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Running `yarn -v`will also run `yarn` / `yarn install` when I only wanted to see the version.

**If the current behavior is a bug, please provide the steps to reproduce.**
Run `yarn -v` anywhere. It becomes most apparent in a folder with a package.json file.

**What is the expected behavior?**
Only the version number is shown.

**Please mention your node.js, yarn and operating system version.**
node v6.9.4
yarn 0.20.3
Mac OSX 10.12.3

chrisbujok:
I think it duplicates #2600

redstrike:
I think the command line interface is inconsistent here.
**node -v** => version
**npm -v** => version
**yarn -v** => install ??? and why **yarn -V** ???

PR #3395, opened 2017-05-13
Use "-v" to display version number instead of "-V"

g-plane:
**Summary**

Before, we run `yarn -V` or `yarn --version` to display version number of Yarn. But we are used to using `-v` to display it, like running `node -v` and `npm -v`.
Maybe caused by [commander.js](https://github.com/tj/commander.js), flags of command for displaying version number cannot be `'-v, -V, --version'`. In other word, we cannot run `yarn -V` to display version number any more.

**Test plan**

Just run `yarn -v`. Note that do **NOT** use `yarn -V` any more.
```
$ yarn -v
0.25.0-0
$ yarn --version
0.25.0-0
```

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -502,7 +502,12 @@ export class Install {
     }
 
     // fin!
-    await this.saveLockfileAndIntegrity(topLevelPatterns, workspaceLayout);
+    // The second condition is to make sure lockfile can be updated when running `remove` command.
+    if (topLevelPatterns.length || (await fs.exists(path.join(this.config.cwd, constants.LOCKFILE_FILENAME)))) {
+      await this.saveLockfileAndIntegrity(topLevelPatterns, workspaceLayout);
+    } else {
+      this.reporter.info(this.reporter.lang('notSavedLockfileNoDependencies'));
+    }
     this.maybeOutputUpdate();
     this.config.requestManager.clearCache();
     return flattenedTopLevelPatterns;
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -29,7 +29,7 @@ const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length 
 const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
 // set global options
-commander.version(version);
+commander.version(version, '--version');
 commander.usage('[command] [flags]');
 commander.option('--verbose', 'output verbose messages on internal operations');
 commander.option('--offline', 'trigger an error if any required dependencies are not available in local cache');
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -182,6 +182,8 @@ const messages = {
   foundWarnings: 'Found $0 warnings.',
   foundErrors: 'Found $0 errors.',
 
+  notSavedLockfileNoDependencies: 'Lockfile not saved, no dependencies.',
+
   savedLockfile: 'Saved lockfile.',
   noRequiredLockfile: 'No lockfile in this directory. Run `yarn install` to generate one.',
   noLockfileFound: 'No lockfile found.',
--- a/__tests__/commands/install/lockfiles.js
+++ b/__tests__/commands/install/lockfiles.js
@@ -44,15 +44,13 @@ test.concurrent("writes new lockfile if existing one isn't satisfied", async ():
   });
 });
 
-test.concurrent('writes a lockfile even when there are no dependencies', (): Promise<void> => {
-  // https://github.com/yarnpkg/yarn/issues/679
+test.concurrent("doesn't write a lockfile when there are no dependencies", (): Promise<void> => {
   return runInstall({}, 'install-without-dependencies', async config => {
     const lockfileExists = await fs.exists(path.join(config.cwd, 'yarn.lock'));
     const installedDepFiles = await fs.walk(path.join(config.cwd, 'node_modules'));
 
-    expect(lockfileExists).toEqual(true);
-    // 1 for integrity file (located in node_modules)
-    expect(installedDepFiles).toHaveLength(1);
+    expect(lockfileExists).toEqual(false);
+    expect(installedDepFiles).toHaveLength(0);
   });
 });
 
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -198,7 +198,7 @@ test.concurrent('should install if no args', async () => {
 
 test.concurrent('should install if first arg looks like a flag', async () => {
   const stdout = await execCommand('--json', [], 'run-add', true);
-  expect(stdout[stdout.length - 1]).toEqual('{"type":"success","data":"Saved lockfile."}');
+  expect(stdout[stdout.length - 1]).toEqual('{"type":"info","data":"Lockfile not saved, no dependencies."}');
 });
 
 test.concurrent('should not output JSON activity/progress if given --no-progress option', async () => {


bestander:
Any idea how to make it non breaking?
Would be great to support -V

g-plane:
sp20170514_084947

arcanis:
I don't really agree; `-v` is either `--version` or `--verbose`. Since the version can already be extracted via `-V` and/or `--version`, I would prefer to keep `-v` to be the verbose option.

g-plane:
But it seems that at current version running with `-v`, it will run `install` actually, not `--verbose`. For most people, after installed Yarn or other tools with CLI, they may run `-v` to test if the tool has been installed successfully, not just to get the version number.

bestander:
Ok, considering that opinions are split regarding -v/-V I propose to get rid of both and use only long version for arguments.

I agree that it is annoying when running `yarn -v` it executes `yarn install`, even worse it saves an empty lockfile.
Let's discuss fixing that effect then?

g-plane:
Do you mean that do not execute `install` if current directory doesn't have lockfile?

bestander:
Maybe if there is not package.json.
Missing yarn.lock can happen

arcanis:
Or we could just make `yarn` print Yarn's usage. By doing this, we could actually support both `-v` as `--version` and `--verbose`:

--- 

```
$> yarn
Usage: yarn [command] [... options]
Commands: ...
$> yarn -v
0.24.3
$> yarn -V
0.24.3
$> yarn install -v
[info] Fetching http://...
```

bestander:
I quite like that `install` is the default command of yarn, I use it a lot and it would be a noticeable breaking change IMHO

g-plane:
I like `install` as a default command too.

g-plane:
@arcanis I think the example you showed is a good idea, except the first usage.

bestander:
Any thoughts on how to proceed?

g-plane:
Do you prefer to making Yarn not execute `install` if missing `package.json`? If so, I will try doing.

bestander:
This is nuanced.
Running `yarn add` in a folder without package.json is fine.
Running `yarn install` is not.
However we work on workspaces where a project can have multiple package.json files, in that case Yarn should find the top level folder and run install from there.

bestander:
Here we have 2 issues, I's like us to get somewhere about -v/-V.
@arcanis, @g-plane, how about we phase out ambiguous -v/-V  completely?
And print usage message, e.g. `Use --version to print version and --verbose to install verbously`

g-plane:
I think using -v is convenient to check if yarn has been installed successfully, not only display version number. Maybe using -v to check that for any tools is a habit, I think.

bestander:
Well -v could print a version and show help message :)

g-plane:
How about making running `yarn -v` only it will display version and running `yarn install -v` it will actually execute `yarn install --verbose`?

g-plane:
How should I add tests?

bestander:
@g-plane, sorry for a delay, I was in a bug crunch and then a time off.
Are you still available to do more?

1. I think there is not need to add automated tests.
Showing screenshots of all possible scenarios in the PR description should be enough.

2. > How about making running yarn -v only it will display version and running yarn install -v it will actually execute yarn install --verbose?

Sounds non consistent, people are used to have `install` command as default, e.g. `yarn --force`, so `-v` behaving differently may be more confusing.

On that note, when you do `yarn -v` it outputs
```
yarn -v
Using globally installed version of Yarn
yarn install v0.24.3
```
Which shows version of Yarn.
The only annoying side effect is that it adds an empty package.json file.
Can we fix just that?

g-plane:
@bestander Well, I have done something at the last commit, which make yarn won't save lock file. Here is the code: https://github.com/yarnpkg/yarn/pull/3395/commits/ff00684909143b1f68119afa53de3c33b3326eb2

The first line `if (Object.getPrototypeOf(this) === Install.prototype)` is to check the current command. If it is `install`, it won't save lock file. But if the command is `add`, it will do what `yarn add` should do and save `yarn.lock`.

Anything should be changed? I will do.

src/cli/index.js
```diff
@@ -29,7 +29,7 @@ const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length
 const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
 // set global options
-commander.version(version);
+commander.version(version, '-v, -V, --version');
```
bestander: let's not alias -v/-V to --version yet because we have not agreed whether people will confuse it with --verbose.

g-plane: Only use `--version` ?

src/cli/commands/install.js
```diff
@@ -431,6 +431,25 @@ export class Install {
...
 
+    if (Object.getPrototypeOf(this) === Install.prototype) {
+      const registryNames = Object.keys(registries);
+      for (let i = 0; i < registryNames.length; i++) {
+        const {filename} = registries[registryNames[i]];
+        const loc = path.join(this.config.cwd, filename);
+
+        // `fs.exists()` is deprecated since nodejs v8
+        try {
+          await fs.access(loc);
```
bestander: looks a bit radical, I think this may be done coming from another angle.
If package.json does not exist then patterns from fetchRequestFromCwd call would be empty.

src/cli/index.js
bestander: Yeah, we avoid ambiguous shorthands in Yarn.

src/cli/commands/install.js
g-plane: Okay, how should I change?

g-plane: Do you mean I should call the function `fetchRequestFromCwd`?

bestander: No, it is called on line 421.
depRequests contains all the patterns from a package.json.
If it is empty and it is not an Add command then installation should stop without saving a lockfile.

g-plane:
@bestander I have do a force push for the latest commit. You can review the code. If something should be changed, I will do.

Screenshot:

sp20170617_162838

g-plane:
It seems there are some conflicts with unit test.

bestander:
Yeah, a few tests were relying on previous behavior

g-plane:
So should I change the test?

src/cli/commands/install.js
```diff
@@ -431,6 +431,10 @@ export class Install {
       this.scripts.setArtifacts(artifacts);
     }
 
+    if (Object.getPrototypeOf(this) === Install.prototype && depRequests.length === 0) {
```
bestander: Look at code on line 367, it saves explicitly when there is nothing to install.
Maybe that logic should be there?

src/cli/index.js
```diff
@@ -29,7 +29,7 @@ const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length
 const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
 
 // set global options
-commander.version(version);
+commander.version(version, '--version');
```
bestander: what this command changes?

bestander:
> So should I change the test?

Yes, as long as the change is reasonable.

src/cli/index.js
g-plane: Remove -V
https://github.com/tj/commander.js/blob/master/index.js#L834

src/cli/commands/install.js
g-plane: Maybe I should move the code to there?

bestander: I think you can just enhance the code there, you probably don't need to check for Install.prototype.
The logic should be "don't save lockfile if there are no dependencies"

g-plane: Ok. I will try.

bestander:
Ok, this looks good now, thanks @g-plane.
Could you rebase please?
And could you confirm what message is written when there is no package.json and user types `yarn -v`?

g-plane:
How about: `No package.json found.` ?

bestander:
Maybe not needed.
What does it output now?

g-plane:
image

bestander:
Yeah, then no package.json found sounds good.

g-plane:
Sorry for delay.

bestander:
Can you run `yarn prettier` to make code comply with the styleguide?

src/cli/commands/install.js
```diff
@@ -494,7 +494,9 @@ export class Install {
     }
 
     // fin!
-    await this.saveLockfileAndIntegrity(topLevelPatterns, workspaceLayout);
+    if (topLevelPatterns.length || await fs.exists(path.join(this.config.cwd, constants.LOCKFILE_FILENAME))) {
+      await this.saveLockfileAndIntegrity(topLevelPatterns, workspaceLayout);
+    }
```
bestander: how about 
```
} else {
  reporter.info(reporter.lang('noPackageJsonFound'));
}
```

Then the the block at line 843-846 can be removed

src/reporters/lang/en.js
```diff
@@ -181,6 +181,8 @@ const messages = {
   foundWarnings: 'Found $0 warnings.',
   foundErrors: 'Found $0 errors.',
 
+  noPackageJsonFound: 'No package.json found.',
```
bestander: How about
```
notSavedLockfileNoDependencies: 'Lockfile not saved, no dependencies.',
```

We are not focusing on a package.json existence but on the packages that were installed

src/cli/commands/install.js
bestander: I think I missed this one, do you need to check `await fs.exists(path.join(this.config.cwd, constants.LOCKFILE_FILENAME)`?

bestander:
Good job, @g-plane, just a few nits.

src/cli/commands/install.js
g-plane: Check what? I am sorry that I do not quite understand what you mean.

bestander: Why are you doing this
```
if (topLevelPatterns.length || await fs.exists(path.join(this.config.cwd, constants.LOCKFILE_FILENAME))) {
```

instead of this?

```
if (topLevelPatterns.length) {

}
```

What does the second condition mean?

g-plane: When running `remove ` command, yarn will update lock file by running `install`.
https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/remove.js#L74

bestander: I see, good thinking.
Yes, makes sense in that case.
Would you add a comment for that?

g-plane: Ok, I will.

g-plane: One more thing, `fs.exists()` is deprecated since node 8. Should I make a change?
https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_exists_path_callback

bestander: no, we use our own wrapper of fs.exists, so it is fine

g-plane:
I have run `yarn prettier` and `yarn lint` before committing, but the tests still failed. I don't know why.

bestander:
Try rebasing to master, `yarn install && yarn prettier` and see if there are any file changes

g-plane:
At the last commit, I have run `yarn` (to install dependencies) , `yarn prettier` and `yarn lint`, and it exited without any errors on my computer. But CI told me:
```
yarn lint
yarn lint v0.21.3
$ yarn run lint-prettier && eslint . && flow check 
yarn run v0.21.3
$ node scripts/prettier.js lint 

  This project uses prettier to format all JavaScript code.
    Please run yarn prettier and add changes to files listed above to your commit.

yarn lint returned exit code 1
```
It said "add changes to files listed above". but on the log above, I cannot find any files listed. Seems confusing.

g-plane:
Is that OK?

bestander:
Great job, @g-plane!
Looking forward for more PRs.

g-plane:
Thanks!

g-plane:
Should #2710 and #3329 be closed?

PR #3397, opened 2017-05-13
feat-fix(pkg): restore partial export of internal APIs

cdaringe:
**Summary**

# problem statement

- yarn@24 bundles the CLI, removing entry points & usability from the tool.
  - for instance, i wrote [parse-yarn-lock](https://www.npmjs.com/package/parse-yarn-lock), which (used to) use `yarn`'s very own [lockfile parser](https://github.com/yarnpkg/yarn/blob/master/src/lockfile/parse.js), to be used for a variety of reasons.  in version 24, the parser is now bundled, and cut off from external access.

In my described use case, it is logical to use the _very same parser_ rather than an independent parser.  There's no value in duplicate tooling, and integrity/quality of the operation can suffer if using an alternative.  Some may challenge that exposing such functionality is out of scope for this project.  However, I would argue that exposing internals is a big enabler and comes with a very low cost + cost-of-ownership.

# solution

- incrementally expose some internal APIs so `yarn` can be used programmatically.  `YARN_AS_LIB` tells the entry point to expose the library by default, vs executing the CLI.

**Test plan**

- an integration test could be done to call into the library from a shell, although im not seeing such test code in `__tests__` ATM, and i figured core maintainers would have a preference on how to execute such a test.

# review

**review with `?w=1` in the URL such that github more cleanly displays the very few changes.  https://github.com/yarnpkg/yarn/pull/3397/files?w=1**

yarn-as-lib mov

### DIFF
--- a/bin/yarn.js
+++ b/bin/yarn.js
@@ -30,4 +30,4 @@ var mkdirp = require('mkdirp');
 var constants = require('../lib-legacy/constants');
 mkdirp.sync(constants.MODULE_CACHE_DIRECTORY);
 
-module.exports = require(path);
+module.exports = process.env.YARN_AS_LIB ? require(path).library : require(path).default();
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -9,6 +9,7 @@ import {MessageError} from '../errors.js';
 import Config from '../config.js';
 import {getRcArgs} from '../rc.js';
 import {version} from '../util/yarn-version.js';
+import * as yarn from '../index.js';
 
 const commander = require('commander');
 const fs = require('fs');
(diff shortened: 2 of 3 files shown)


Daniel15:
I wonder whether we should split the lockfile parser out into a separate module (like `yarn-lockfile`) to more easily facilitate reuse like this.

cdaringe:
hey @Daniel15, ya, that would make me a happy camper :).  i do wonder how many other tasks would be great to drive programmatically.  this PR is admittedly a cheap fix.  the result of this PR is simple enablement, albeit w/ debatably awkward ergonomics.

Daniel15:
This seems okay to me, until we have the lockfile parser in its own separate package (cc @cpojer). 

Please rebase as it looks like there's a conflict  :)

cdaringe:
rebased!

cdaringe:
ah man, conflicts again?  bummer!  @Daniel15, if I rebase again, can we merge pronto?  the failing CI i dont believe the Travis build was failing due to change, but rather just some flakiness

cdaringe:
friendly and coordial bump!  @Daniel15, will you take a peek at my last comment?  thx

Daniel15:
Sorry I've been so busy recently, and haven't been able to review this PR :( 

It seems reasonable, someone else on the team (like @cpojer or @bestander) might get a chance to merge it before I do.

bestander:
Thanks for making an effort and updating PR, @cdaringe.
And sorry for making it hang for so long time.

IMHO the change in cli/index.js becomes fragile this way because this file now has dual usage.
And using "yarn as a lib" seems like a temporary hack.
I may be wrong here but maybe we could achieve this another way?

From the top of my head we were discussing splitting Yarn source code into workspaces, i.e. packages and then the lockfile parser could be a standalone package.

cdaringe:
hey @bestander, no sweat man.  thx for getting back.

> IMHO the change in cli/index.js becomes fragile this way because this file now has dual usage.

yup.

> And using "yarn as a lib" seems like a temporary hack.

yup.  total hack

> I may be wrong here but maybe we could achieve this another way?

we could have webpack dump 2 builds, one for `yarn` and one for `yarn/lib`, but then we're double distributing, and people surely eventually (if not immediately) won't like that either.

> From the top of my head we were discussing splitting Yarn source code into workspaces, i.e. packages and then the lockfile parser could be a standalone package.

definitely ideal.  yarn24 just really cut us off hard w/ that parser, especially when dependency reports are so key to legal processes for those of us who aggressively make sure we are always shipping compliant/approved things.

would you be OK letting it in with a big disclaimer, perhaps `console.warn` when that env var is set, denoting its soon deprecation, but at least re-enabling the functionality in the interim.  im thinking a big evil banner at the top of the file too :)

bestander:
@cdaringe, IMHO it is better to keep this feature outside of core Yarn for now.
What prevents you to create a wrapper project that would clone Yarn and have a yarn.lock parser compiled with gulp?
We could have some code that would make it easier in Yarn but I don't like putting a large wrapper for the full cli/index file.

bestander:
I'll close for now, please let us know how we can help without introducing much complexity.

Issue #3318, opened 2017-05-04
Use Prettier


cpojer:
This repo must start using Prettier ;)

voxsim:
Hey @cpojer I agree with you; do you have any suggestions on how we can use it? I mean we can format the yarn code on pre-commit or before merging a pull request or maybe before a release, etc.. O.o maybe you can share on how you use it in other projects!

cpojer:
See https://github.com/facebook/jest

PR #3401, opened 2017-05-14
add prettier and prettying everything

voxsim:
**Summary**

This fix https://github.com/yarnpkg/yarn/issues/3318.

**Test plan**

Tests already in place and green :D


scripts/prettier.js
```diff
@@ -0,0 +1,60 @@
...
+const shouldWrite = process.argv[2] === 'write';
+const isWindows = process.platform === 'win32';
+const prettier = isWindows ? 'prettier.cmd' : 'prettier';
+const prettierCmd = path.resolve(__dirname, '../node_modules/.bin/' + prettier);
+
+const config = {
+  ignore: ['**/node_modules/**'],
+  options: {
+    'bracket-spacing': 'false',
+    'print-width': 120,
```
cpojer: What happens when we change this to 80 (or 100)? Prettiers output actually looks worse the longer the line-length is because it tries to cram everything into one line.

voxsim: ah ok I can reduce the number and run `yarn prettier` if you prefer

voxsim: in the jest project there is 80 for print-width

voxsim:
Ah ok I really don't know that plugin, I can switch to eslint-plugin-prettify if we prefer, I see what the jest project does and copy it (infact if you see the file is the same apart some modifications).

scripts/prettier.js
```diff
@@ -0,0 +1,60 @@
+/**
```
bestander: People say you can use thus plugin https://github.com/zertosh/eslint-plugin-prettify instead of this file now

voxsim:
@kittens @bestander @cpojer now should works as expected :)

voxsim:
I am sorry but this not works as expected for some reasons:
- ESLint's --fix flag will not apply fixes anymore. This is a known issue of processors. However, the fix information is still outputted to formatters. So if you're using ESLint through something like Arcanist, the fixes will work. (https://github.com/zertosh/eslint-plugin-prettify#caveats)
- I used the string "fb" to set "Facebook defaults", but I have to add a docblock with /** @format */ in every page and it seems not to work if I already have /* @flow */ O.o

I will dig further and maybe i will open another pull request.

bestander:
Thanks, should we revert?

bestander:
Or a patch would be fine?

voxsim:
Maybe now it is better to revert everything, I really don't know how much it could take.
If it will be problematic I should re-open a pr with the custom scripts like jest does.

voxsim:
@bestander after thinking a while, if i open a patch and I delete eslint-plugin-prettify and reintroduce the custom scripts, would it be fine? I mean without reverting everything

bestander:
Yeah, let's have a temporary patch

RomanovRoman:
@voxsim
it's about https://github.com/yarnpkg/yarn/pull/3401#issuecomment-301886565
Link https://github.com/zertosh/eslint-plugin-prettify#caveats is broken because eslint-plugin-prettify merged into https://github.com/prettier/eslint-plugin-prettier.
I can't find any "caveats" in docs.
Have you some progress?

PR #3412, opened 2017-05-15
Atomic node modules

Xion:
**Summary**
This PR adds the `--atomic` flag.

When invoked with the `--atomic` flag, Yarn will temporarily rename the node_modules directory and install all packages within this new destination, only to rename it back at the end.

In a sense, this makes the `yarn install` operation "atomic", because anything that watches node_modules for changes will not see its intermediate state. This is useful to prevent processes like Watchman from taking action based on incomplete installation state.

**Test plan**
Tests included in the PR.

### DIFF
--- a/__tests__/util/module-storage.js
+++ b/__tests__/util/module-storage.js
@@ -1,0 +1,199 @@
+/* @flow */
+
+import Config from '../../src/config.js';
+import type BaseRegistry from '../../src/registries/base-registry.js';
+import {registryNames} from '../../src/registries/index.js';
+import ModuleStorage from '../../src/util/module-storage.js';
+import {Reporter, NoopReporter} from '../../src/reporters/index.js';
+import makeTemp from '../_temp.js';
+import * as fs from '../../src/util/fs.js';
+
+describe('ModuleStorage', () => {
+  let config: Config;
+  let reporter: Reporter;
+  let moduleStorage: ModuleStorage;
+
+  beforeEach(async () => {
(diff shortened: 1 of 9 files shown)


rally25rs:
Seems like a good idea. Well thought out. Good tests.
Needs some merge conflicts resolved, but otherwise 👍

BYK:
@Xion  - Still working on this?

@bestander, @arcanis any thoughts?

bestander:
I think the PR is great, thanks again for the contribution @Xion and really sorry for leaving it unattended for a while.

I'll leave it to @BYK to make do the review and follow up with the rebase.

BYK:
@Xion - up for rebasing?

Xion:
Yep, I'll rebase & resolve the conflicts.

Xion:
The CI uses an ancient version of Node which doesn't have `Object.values`. I'll shim it.

Xion:
Looks like it's good to go!

BYK:
>The CI uses an ancient version of Node which doesn't have Object.values. I'll shim it.

The CI uses Node 4 and 6 now which are the versions we commit to support :) Node 8 will be there soon. Thanks for fixing it!

BYK:
I am going to go ahead and close this PR. Although it breaks my heart, I do not think it is worth having the additional code and complexity around this feature for the potential benefits of it.

If anyone disagrees with me, please feel free to keep the discussion open and ask for reopening this PR.

PR #3417, opened 2017-05-16
Plumb "--network-timeout" cli flag through to the request manager

walkerburgin:
**Summary**
+ Noticed tonight while debugging a timeout issue that `yarn` doesn't actually respect the `--network-timeout` flag (`v0.24.5` at least)
    - You can repro this by running `yarn add <package> --network-timeout 10` and observing that the request doesn't actually time out
+ This change just plumbs the `--network-timeout` command line flag through to the request manager

**Test plan**
+ I confirmed (locally) that after this change yarn will respect a network timeout specified on the command line


destroyerofbuilds:
Recently, while attempting to work around network timeouts because of a slow private npm registry, I ran into the `--network-timeout` CLI option not working.

So, thank you for submitting a fix for this!

:+1:

bestander:
Could you rebase please?

Issue #2918, opened 2017-03-14
Make resolver stateless


bestander:
**Do you want to request a *feature* or report a *bug*?**
*refactoring*

**What is the current behavior?**

PackageResolver is a member of all the other classes: PackageFetcher, Install etc.
This leads to complex stateful logic that is hard to make sense of e.g. https://github.com/yarnpkg/yarn/blob/master/src/package-fetcher.js#L93

Would be great to make all the resolvers, fetchers and linkers stateless with a clear ownership of install data.

**Great opportunity to contribute**

PR #3422, opened 2017-05-16
Make package-compatibility and package-fetcher stateless

voxsim:
**Summary**
Make package-compatibility and package-fetcher stateless and indipendent from package-resolver.
I made some ugly choices, but imho the road towards https://github.com/yarnpkg/yarn/issues/2918 is open :D (I hope this merge it will not close that issue).
I think it is important to be incremental as possible because I am touching the core O.o
Big Pro: now we can time every part and maybe introduce some unit tests :)

**Test plan**
Tests are green, before made every change I carefully check if every line were covered indirectly from tests.

### DIFF
--- a/src/package-fetcher.js
+++ b/src/package-fetcher.js
@@ -18,7 +18,8 @@ async function fetchCache(dest: string, fetcher: Fetchers, config: Config): Prom
     dest,
     cached: true,
   };
-} 
+}
+
 async function fetchOne(ref: PackageReference, config: Config): Promise<FetchedMetadata> {
   const dest = config.generateHardModulePath(ref);
 
@@ -37,7 +38,7 @@ async function fetchOne(ref: PackageReference, config: Config): Promise<FetchedM
   await fs.unlink(dest);
 
   try {
-    return await fetcher.fetch();
+    return fetcher.fetch();
   } catch (err) {
     try {
       await fs.unlink(dest);
--- a/src/package-fetcher.js
+++ b/src/package-fetcher.js
@@ -81,37 +81,41 @@ export async function fetch(pkgs: Array<Manifest>, config: Config): Promise<Arra
   });
   const tick = config.reporter.progress(pkgs.length);
 
-  return await promise.queue(pkgs, async pkg => {
-    const ref = pkg._reference;
-    if (!ref) {
-      return pkg;
-    }
-
-    const res = await maybeFetchOne(ref, config);
-    let newPkg;
-
-    if (res) {
-      newPkg = res.package;
-
-      // update with new remote
-      // but only if there was a hash previously as the tarball fetcher does not provide a hash.
-      if (ref.remote.hash) {
-        ref.remote.hash = res.hash;
+  return await promise.queue(
+    pkgs,
+    async pkg => {
+      const ref = pkg._reference;
+      if (!ref) {
+        return pkg;
       }
-    }
 
-    if (tick) {
-      tick(ref.name);
-    }
+      const res = await maybeFetchOne(ref, config);
+      let newPkg;
 
-    if (newPkg) {
-      newPkg._reference = ref;
-      newPkg._remote = ref.remote;
-      newPkg.name = pkg.name;
-      newPkg.fresh = pkg.fresh;
-      return newPkg;
-    }
+      if (res) {
+        newPkg = res.package;
 
-    return pkg;
-  }, config.networkConcurrency);
+        // update with new remote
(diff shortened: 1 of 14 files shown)


src/package-resolver.js
```diff
@@ -300,25 +291,25 @@ export default class PackageResolver {
...
-      }
-    }
+  // add pattern to the manifest we're collapsing to
+  for (const pattern of refPatterns) {
+    collapseToReference.addPattern(pattern, collapseToManifest);
+  }
+}
 
-    return collapseToPattern;
+return collapseToPattern;
```
arcanis: Indentation's broken 😃

arcanis:
That's really nice! What would you think about moving `Config` inside the function parameters as well? So, instead of:

```js
export class {

    constructor(config) {
        this.config = config;
        this.fetcher = new PackageFetcher(this.config)
    }

    doSomething() {
        return this.fetcher.doSomething();
    }

}
```

We could have:

```js
export class {

    doSomething({ config }) {
        return PackageManager.doSomething(..., { config });
    }

}
```

src/package-fetcher.js
```diff
@@ -75,10 +72,13 @@ export default class PackageFetcher {
     }
   }
 
-  async init(): Promise<void> {
-    let pkgs = this.resolver.getPackageReferences();
+  async init(pkgs: Array<Manifest>): Promise<Array<Manifest>> {
```
voxsim: I think the right name is `fetch` :) what do you think?

voxsim:
mmm @arcanis please see my changes: I prefer to export multiple functions instead of one class.
For now we need *always* the config as parameter because we use the report, etc.. it is too complicated to make it as an option imho.

arcanis:
As you wish. I think a class would be better because it would be easier to type it, and it would make it possible to easily pass it around (for example, at some point we can imagine the resolver would need to have access to the fetcher in order to extract the correct information from a git repository), but that's something we can change later if we need to.

This PR looks fine by me, the only thing that worries me a bit is that it will introduce a delta between parts of the architecture (most notably the resolver, which will still use the `new PackageResolver()` architecture while the fetchers will be quite different). Do you plan to also work on the other tools as followups?

voxsim:
@arcanis yep i am trying to working on other tools incrementally, but they are very difficult to refactor..
One other option is that I can continue to work on this pr, but it will be difficult to merge later maybe O.o
Let me know what you suggest to do :)

voxsim:
And if you prefer I can reintroduce classes, I am open to revert my last experiment in case :P

src/package-fetcher.js
arcanis: 👍

bestander:
Ping @arcanis, good to merge?

voxsim:
I added some commits to reduce some confusions: i removed seedPatterns and newPatterns. Now in the manifest there is a flag called `fresh` that tell you if the pattern is new. I am starting to understand the internal flow of package-resolver :)

src/types.js
```diff
@@ -125,6 +125,8 @@ export type Manifest = {
   main?: string,
 
   workspaces?: Array<string>,
+
+  fresh?: boolean
```
arcanis: Could you add a small comment here to explain what means `fresh` in this context, and a short summary detailing how it is used inside the Yarn internals?

voxsim:
Mmh when I program I usually type `yarn build && yarn test` to check if everything is ok after rebasing. There was an issue in the package.json that every time we run `yarn lint` we run `yarn run prettier` instead of `yarn run lint-prettier`, the main issue is that I prettify some random files from last commits O.o and maybe we don't notice because we don't lint correctly and stop the pipeline in case.

voxsim:
Btw I committed all the prettified files in one commit (the last one). If there are any problems I can revert it without problems.

arcanis:
I'm not familiar with prettier - what's the different? If I understand correctly, `yarn prettier` was updating the file, whereas `yarn lint-prettier` will only be reporting warnings, letting us manually run prettier, is that it?

voxsim:
Yep the issue is that if we run `yarn run prettier` in the lint phase we updated files in place and we never commit :(

arcanis:
Ok 👍 I'll wait for CI checks to end, and then I'll merge everything

voxsim:
Mhh all failed tests seem flaky O.o @arcanis if you catch something please let me know

bestander:
Needs a rebase anyway :)

voxsim:
Done, I hope everything will be alright :P

PR #3425, opened 2017-05-16
Fix support for private git repos when using shorted repo syntax and lock file exists

ramasilveyra:
> Description based on https://github.com/yarnpkg/yarn/issues/573#issuecomment-301674068

https://github.com/yarnpkg/yarn/pull/2992 added support for private git repos when using shorted repo syntax (this landed on v0.24.0).

The generation for the first time of a `yarn.lock` file works well, but then with a yarn.lock file if you try `rm -rf node_modules && yarn` fails with this error:
```sh
error An unexpected error occurred: "Invalid protocol: git+ssh:".
```
That error comes from https://github.com/request/request/blob/b12a6245d9acdb1e13c6486d427801e123fdafae/request.js#L455  and I think it's because yarn tries to use the tarball fetcher instead of the git fetcher.

The offline feature/fix added recently makes the git resolver defaults to the tarball to avoid http calls https://github.com/yarnpkg/yarn/pull/3000/files
Returning the git shrunk instead of the tarball if is possible seems to fix the issue.

### DIFF
--- a/__tests__/package-request.js
+++ b/__tests__/package-request.js
@@ -1,0 +1,50 @@
+/* @flow */
+
+import PackageRequest from '../src/package-request.js';
+import * as reporters from '../src/reporters/index.js';
+import PackageResolver from '../src/package-resolver.js';
+import Lockfile from '../src/lockfile/wrapper.js';
+import Config from '../src/config.js';
+
+async function prepareRequest(pattern, version, resolved): Object {
+  const privateDepCache = {[pattern]: {version, resolved}};
+  const lockfile = new Lockfile(privateDepCache);
+  const reporter = new reporters.NoopReporter({});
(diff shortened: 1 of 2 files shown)


bestander:
@arcanis is the best to review this.
@ramasilveyra any idea of a unit/e2e test that will cover it?

ramasilveyra:
@bestander this night (Argentina time) I can add the tests (maybe I will need to move the get shrunk logic to a new method `getShrunk()` to make the unit test). In the meantime feedback if this solution it's fine will be great @arcanis 😄

ramasilveyra:
**Unit tests added!** Solution change: issue fixed directly on the method `this.request.getLocked()`.

bestander:
Thanks, @ramasilveyra

Issue #3441, opened 2017-05-17
URL column is empty in `yarn outdated`


devoto13:
**Do you want to request a *feature* or report a *bug*?**

Bug (regression)

**What is the current behavior?**

Note missing values in URL column.

```
$ yarn outdated
yarn outdated v0.24.4
Package                           Current Wanted Latest Package Type    URL
webpack                           2.2.1   2.2.1  2.5.1  devDependencies
webpack-dev-server                2.4.1   2.4.1  2.4.5  devDependencies
webpack-merge                     4.0.0   4.0.0  4.1.0  devDependencies
```

**If the current behavior is a bug, please provide the steps to reproduce.**

To reproduce run `yarn outdated` with yarn >=0.24.0 with any outdated dependencies.

This is a regression after [this commit](https://github.com/yarnpkg/yarn/commit/c92411d8540001f6f4043b17d7d38c58301e9640). It switched to smaller package metadata, which doesn't return `homepage` property in the response. When I remove [this header](https://github.com/yarnpkg/yarn/commit/c92411d8540001f6f4043b17d7d38c58301e9640#diff-b053bee294c216269844e5874039b6caR61) it starts working again.

Obviously smaller metadata is important, but having packages homepage URLs was very convenient feature to quickly go to packages' changelogs. That's why I'm unsure what is the right way to fix this. But if somebody have pointers I can probably prepare a PR with the fix. For now I have opened PR #3442 with the failing test case.

**What is the expected behavior?**

```
$ yarn outdated
yarn outdated v0.23.4
Package                           Current Wanted Latest Package Type    URL
webpack                           2.2.1   2.2.1  2.5.1  devDependencies https://github.com/webpack/webpack
webpack-dev-server                2.4.1   2.4.1  2.4.5  devDependencies http://github.com/webpack/webpack-dev-server
webpack-merge                     4.0.0   4.0.0  4.1.0  devDependencies https://github.com/survivejs/webpack-merge
```

**Please mention your node.js, yarn and operating system version.**

```
$ yarn --version
0.24.4
```

PR #3442, opened 2017-05-17
Fix `yarn outdated` missing values in URL column

devoto13:
**Summary**

Adding failing test for a regression. See #3441 for more details.

**Test plan**

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -86,7 +86,9 @@ export default class NpmRegistry extends Registry {
   }
 
   async checkOutdated(config: Config, name: string, range: string): CheckOutdatedReturn {
-    const req = await this.request(NpmRegistry.escapeName(name));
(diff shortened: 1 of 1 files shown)


bestander:
Thanks @devoto13.
I don't think a test is necessary, this is a straight forward integration issue.

Let's just fix is instead.
The fix for info command was in passing the right request header https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/info.js#L63.

For outdated command the header needs to be passed via a couple of functions into checkOutdated https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L89

devoto13:
@bestander Apparently `checkOutdated` is used only by `yarn outdated` and `yarn upgrade-interactive` commands. Both commands need full response from registry to show `URL` column, so it was enough to just modify this method.

I've added correct header and removed test. PTAL

bestander:
Great work, @devoto13~

PR #3445, opened 2017-05-17
Added support for more git URL formats

thedumbterminal:
**Summary**

After recent work on fixing the private shortened git URL formats I've noticed there are still some that do not work.

For example this URL format works:

    github:thedumbterminal/test-private-node-module

But this format does not:

    github:thedumbterminal/test-private-node-module.git

This work fixes this.

**Test plan**

Create a new directory with the following `package.json`:

```json
{
  "name": "1",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
   "grunt-gemnasium": "github:thedumbterminal/grunt-gemnasium.git"
  }
}
```

Without this fix you will get the following error:

```
ssh: Could not resolve hostname github: nodename nor servname provided, or not known

and the repository exists.
```
(Notice the error: `Could not resolve hostname github` which is incorrect)

With this fix you will able to install these modules as it will switch to ssh auth.

### DIFF
--- a/__tests__/util/git.js
+++ b/__tests__/util/git.js
@@ -41,6 +41,26 @@ test('npmUrlToGitUrl', () => {
     hostname: 'scp-host-nickname',
     repository: 'user@scp-host-nickname:npm-opam/ocamlfind.git',
   });
+  expect(Git.npmUrlToGitUrl('github:npm-opam/ocamlfind.git#v1.2.3')).toEqual({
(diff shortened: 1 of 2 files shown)


bestander:
Thanks for adding tests!

Issue #2064, opened 2016-11-29
Global packages no longer work after node.js update with homebrew


publicarray:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

When node and yarn are installed with homebrew `yarn global bin` returns a path that is tied to the a specific version of node. If node is updated with brew the symlinks are no longer valid, i.e global packages with binaries need to be reinstalled.

```bash
~ ❯ yarn global bin
/usr/local/Cellar/node/7.2.0/bin
```

Related issue: https://github.com/Homebrew/homebrew-core/issues/7283

**If the current behavior is a bug, please provide the steps to reproduce.**

I've used brew reinstall to emulate an update, I'm not sure how different it is to `brew upgrade node`.

```bash
~ ❯ brew install node
~ ❯ node --version
v7.2.0
~ ❯ brew install yarn
~ ❯ yarn global bin
/usr/local/Cellar/node/7.2.0/bin
~ ❯ yarn global add bower
~ ❯ bower --version
1.8.0
~ ❯ brew reinstall node --HEAD
~ ❯ node --version
v8.0.0-head
~ ❯ yarn global bin
/usr/local/Cellar/node/HEAD-11b0872/bin
~ ❯ bower --version
command not found: bower
```

**What is the expected behavior?**

```bash
~ ❯ bower --version
1.8.0
```

**Please mention your node.js, yarn and operating system version.**

```bash
~ ❯ yarn --version
0.17.9
~ ❯ node --version
v7.2.0
~ ❯ sw_vers
ProductName: Mac OS X
ProductVersion: 10.12.1
BuildVersion: 16B2555
```

nicgordon:
The same issue exists for node installed via nvm:
```
~ $ which node
/Users/nicgordon/.nvm/versions/node/v7.2.0/bin/node
~ $ yarn global bin
/Users/nicgordon/.nvm/versions/node/v7.2.0/bin
~ $ yarn global add gatsby
~ $ which gatsby
gatsby not found
```

However adding this prefix flag symlinks the installed module into `/usr/local/bin` and fixes the issue for me
```
~ $ yarn global add gatsby --prefix /usr/local
~ $ which gatsby
/usr/local/bin/gatsby
```

Checking the symlink shows me that the module was actually installed here:
```
/Users/nicgordon/.config/yarn/global/node_modules/.bin/gatsby
```

Found this solution here https://github.com/yarnpkg/yarn/issues/1877

[...]

PR #3458, opened 2017-05-19
Change global prefix directory to `/usr/local` on OS X

### LINKED ISSUE #1194: Location of global bin doesn't match the doc

Bug

```
$ brew install yarn
==> Downloading https://homebrew.bintray.com/bottles/yarn-0.15.1.el_capitan.bott
Already downloaded: /Users/work/Library/Caches/Homebrew/yarn-0.15.1.el_capitan.bottle.tar.gz
==> Pouring yarn-0.15.1.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/yarn/0.15.1: 5,239 files, 12.6M
$ yarn global bin
/usr/local/Cellar/node/6.8.1/bin
```

Shouldn't `yarn global bin` output `~/.yarn/bin` as per [the doc](https://yarnpkg.com/en/docs/install)?

Node v6.8.1
yarn v0.15.1
Os X 10.11.6

KishanBagaria:
When node is installed from Homebrew, node exists at `/usr/local/Cellar/node/7.10.0/bin/node`

So yarn's global bin directory (`yarn global bin`) becomes: `/usr/local/Cellar/node/7.10.0/bin`

```
$ node -e 'console.log(process.execPath)'
/usr/local/Cellar/node/7.10.0/bin/node
$ yarn global bin
/usr/local/Cellar/node/7.10.0/bin
```

This creates two problems:

1. `/usr/local/Cellar/node/(version)/bin` is not present in `PATH` by default. Users have to modify their profile to extend `PATH` with `yarn global bin`.

2. When node is upgraded, all linked global binaries become inaccessible since `yarn global bin` automatically gets changed to `/usr/local/Cellar/node/(latest-version)` from `/usr/local/Cellar/node/(old-version)`. The global bin directory should remain constant on upgrades.

The fix is to simply use `/usr/local/bin` as the default global bin directory in OS X installations. NPM (`npm --global prefix`) also uses that directory (at least when installed from Homebrew).

This should also fix #1194.

**Alternatives**

1. Instead of all OS X installations, this change can be applied only to installations where node is installed from Homebrew, by checking if `process.execPath` begins with `/usr/local/Cellar`. Although, this issue would still be present with nvm or some other node-installer.

2. Instead of modifying yarn, a patch could be made in the [Yarn formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/yarn.rb) to automatically run `yarn config set prefix /usr/local`. (EDIT: It's not possible atm: https://github.com/Homebrew/homebrew-core/issues/7283#issuecomment-289795136) Other node-installers like nvm will also require similar changes.

3. In the `getGlobalPrefix` function, instead of
```
let prefix = path.dirname(path.dirname(process.execPath)); //=> /usr/local/Cellar/node/7.10.0
```
this could be used:
```
const whichNode = execa.sync('which', ['node']).stdout; //=> /usr/local/bin/node
let prefix = path.dirname(path.dirname(whichNode)); //=> /usr/local
```

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -4,6 +4,7 @@ import type {Reporter} from '../../reporters/index.js';
 import type {Manifest} from '../../types.js';
 import type Config from '../../config.js';
 import {MessageError} from '../../errors.js';
+import {OSX_GLOBAL_PREFIX_DIRECTORY} from '../../constants.js';
 import {registries} from '../../registries/index.js';
 import NoopReporter from '../../reporters/base-reporter.js';
 import buildSubCommands from './_build-sub-commands.js';
@@ -76,6 +77,8 @@ function getGlobalPrefix(config: Config, flags: Object): string {
     return String(config.getOption('prefix'));
   } else if (process.env.PREFIX) {
     return process.env.PREFIX;
+  } else if (process.platform === 'darwin') {
+    return OSX_GLOBAL_PREFIX_DIRECTORY;
   } else if (process.platform === 'win32') {
     if (process.env.LOCALAPPDATA) {
       return path.join(process.env.LOCALAPPDATA, 'Yarn', 'bin');
--- a/src/constants.js
+++ b/src/constants.js
@@ -58,6 +58,7 @@ export const MODULE_CACHE_DIRECTORY = getCacheDirectory();
 export const CONFIG_DIRECTORY = getDirectory('config');
 export const LINK_REGISTRY_DIRECTORY = path.join(CONFIG_DIRECTORY, 'link');
 export const GLOBAL_MODULE_DIRECTORY = path.join(CONFIG_DIRECTORY, 'global');
+export const OSX_GLOBAL_PREFIX_DIRECTORY = '/usr/local';
 
 export const META_FOLDER = '.yarn-meta';
 export const INTEGRITY_FILENAME = '.yarn-integrity';


rally25rs:
I believe this PR would also fix: #2064

KishanBagaria:
Didn't see that before, thanks.

It looks like some change related to this is required. The current behavior isn't preferred by everyone.

Do you know if the participants in that issue got notified of this PR when you mentioned it?

bestander:
Looping in @Daniel15  as I completely lost faith in trying to understand all those prefixes needed for every OS

Daniel15:
I'm not very familiar with Mac OS, but the issue on Linux would be that regular users normally don't have write access to `/usr/local`. My preferred fix here would be to use `/usr/local` if it's writable by the current user, otherwise use `~/.yarn/bin` or something similar.  

On Windows we always use a user directory: `%LOCALAPPDATA%\Yarn\bin` (eg. `C:\Users\Daniel\AppData\Local\Yarn\bin`)

bestander:
To me it seems like a safe enough change for Mac OS, let's merge it.
Speak up if you want to revert it.

@KishanBagaria, you have expressed a few good ideas about paths in Linux.
Do you want to give it a try and help Yarn to provide a great experience with all those PATH problems?

paulirish:
On Google corporate OS X machines `/usr/local` is owned by root, rather than the user. As a result, most Googlers with homebrew install to another location rather than only using homebrew via sudo (which is terrrible).

image

FWIW I've been manually adding my homebrew'd `node` path to `PATH` in order to make sure my yarn global installs work. [And I rev this each time I upgrade node. It's kinda gross. :/](https://github.com/paulirish/dotfiles/blob/6fbe3da9b16cc48511eceab38be407daca0f499b/fish/path.fish#L26-L43)

For me, it'd be perfect to install the global links into `~.yarn/bin`  (as I've expected them to be there multiple times).  @Daniel15's proposal of falling back to there if `/usr/local` isn't writable is fine with me.

bestander:
Thanks for feedback @paulirish.
I'll revert this now.

@KishanBagaria would you give it another try with @Daniel15's feedback?

KishanBagaria:
@paulirish That must needlessly require sudo for a lot of install scripts. Maybe you should drop the suggestion of making `/usr/local` user-accessible in their issue tracker.

Until the next Yarn release, you can simply run `yarn config set prefix ~/.yarn` and get rid of the hacks in your dotfile.

PR #3461, opened 2017-05-20
Optimize async functions

KishanBagaria:
* Add lint rule `no-return-await`

* Replace redundant `return await` calls with `return`

* Convert async functions which no longer have `await` calls to normal functions

Similar to #352


voxsim:
Hey, thanks for the pr! @bestander I think we should merge it.

bestander:
Thanks, @KishanBagaria!

Issue #3435, opened 2017-05-17
yarn-offline-mirror-pruning also removes .git file


pmorch:
**Do you want to request a *feature* or report a *bug*?**
An improvement suggestion or a bug. You decide.
This is a minor issue for us, I just thought I'd report it. See workaround below.

**What is the current behavior?**
The current behavior is that yarn-offline-mirror-pruning (apparently) removes every single file in the folder that is not a current tarball.

**If the current behavior is a bug, please provide the steps to reproduce.**
See [Pruning an Offline Mirror | Yarn](https://yarnpkg.com/en/docs/prune-offline-mirror) but create e.g. a `.git` file in the folder first before pruning.

**What is the expected behavior?**
My expectation is that yarn pruning should remove all non-current _tarballs_ (i.e. `*.tgz` files?) that aren't current, but leave my `.git` file alone.

My expectation is fuelled by [Pruning an Offline Mirror | Yarn](https://yarnpkg.com/en/docs/prune-offline-mirror) that describes that:
> However, it does not automatically remove tarballs that are no longer referenced in yarn.lock. For example, if you $ yarn remove a dependency, the tarball will remain in the mirror, even if no other dependencies have it as a sub-dependency. This behavior can be desirable in a setting where many projects share the same mirror, but when that is not the case, you may want to have Yarn remove unnecessary tarballs.

So this isn't life and death. The workaround for us is to put all the tarballs in a `cache/` sub-directory, and let yarn prune there instead. But this extra directory is a little ugly.

Thanks for giving us yarn-offline-mirror-pruning to begin with!

**Please mention your node.js, yarn and operating system version.**
```
> node --version
v7.10.0
> yarn --version
0.23.4
> lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.8 (jessie)
Release:	8.8
Codename:	jessie
```

pichouk:
Ping @bestander. I guess this issue was fixed by #3467 and can be closed.

bestander:
thanks

pmorch:
Wonderful! Thanks!

PR #3467, opened 2017-05-22
Add explicit check for tarball files in mirror files directory

kaylieEB:
**Summary**

This PR is a bug fix from https://github.com/yarnpkg/yarn/issues/3435
The issue was that files that weren't tarballs (.git, for example) were also removed since they weren't present in the lockfile.

**Test plan**
I've stepped through the STRs from the original bug report
1) Set `yarn config set yarn-offline-mirror ./npm-packages-offline-cache` and `yarn config set yarn-offline-mirror-pruning true`
2) Create a git repo inside of packages-cache
3) Run `yarn remove` on an installed dependency
4) Check that the git repo still exists, unlike previous behavior

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -718,7 +718,7 @@ test.concurrent('should install if symlink source does not exist', async (): Pro
   await runInstall({}, 'relative-symlinks-work', () => {});
 });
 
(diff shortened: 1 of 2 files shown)


voxsim:
Hey @kaylieEB, could you please write a test that covers the case? If you need help, please let me know ;)

voxsim:
Thanks :) @bestander I think we should merge it

PR #3476, opened 2017-05-22
Fix npm command

battaile:
'update' instead of 'upgrade'

**Summary**

If your yarn version is out of date, you get prompted to remedy this by running 'npm upgrade...' which results in an error.  Using 'npm update...' instead works as expected.

**Test plan**

Run on a machine with an older version of yarn.  Should be able to follow on-screen instructions and update to latest version without errors.

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -85,7 +85,7 @@ function getUpdateCommand(installationMethod: InstallationMethod): ?string {
   }
 
   if (installationMethod === 'npm') {
-    return 'npm upgrade --global yarn';
+    return 'npm update --global yarn';
   }
 
   if (installationMethod === 'chocolatey') {


bestander:
upgrade should be an alias https://docs.npmjs.com/cli/update

bestander:
I see one advantage of keeping it as `npm upgrade` - it is the same as `yarn upgrade`...
@battaile, what error did you see?

battaile:
Thanks @bestander .  It works fine on my Mac, but I get the npm command usage help screen when I try on Windows.  I I must have something in my npm install configured incorrectly there.

arcanis:
Personally, whether it's `upgrade` or `update`, I always have to check if it means that the command will bump the package dependencies to the latest possible versions and then update the package.json, or just regenerate the lockfile with the latest possible versions :(

bestander:
Considering that this is a bit of a grey area and opinionated I tend to not change anything yet.
Thanks for the PR anyway, @battaile, hope to see more!

PR #3478, opened 2017-05-23
Bugfixes

arcanis:
- Fix an issue where the scope was incorrect, which broke the error message reported
- Fix an issue where the `check-files` option was incorrectly behaving when used with `enable-meta-folder`

### DIFF
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -61,6 +61,36 @@ export default class InstallationIntegrityChecker {
 
   config: Config;
 
+  async _getModuleLocation(usedRegistries?: Set<RegistryNames>): Promise<string> {
+    // build up possible folders
+    let registries = registryNames;
+    if (usedRegistries && usedRegistries.size > 0) {
+      registries = usedRegistries;
+    }
+    const possibleFolders = [];
+    if (this.config.modulesFolder) {
+      possibleFolders.push(this.config.modulesFolder);
+    }
+
+    // ensure we only write to a registry folder that was used
+    for (const name of registries) {
+      const loc = path.join(this.config.cwd, this.config.registries[name].folder);
+      possibleFolders.push(loc);
+    }
+
+    // if we already have an integrity hash in one of these folders then use it's location otherwise use the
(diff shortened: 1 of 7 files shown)


bestander:
A description would be nice :)

bestander:
Test would be superb but we can skip them in this time of urgency

src/integrity-checker.js
```diff
@@ -253,7 +259,7 @@ export default class InstallationIntegrityChecker {
       lockfile,
       patterns,
       Object.assign({}, flags, {checkFiles: false}), // don't generate files when checking, we check the files below
-      loc.locationFolder,
+      await this._getModuleFolder(),
```
bestander: Where is _getModuleFolder coming from?

arcanis: https://github.com/yarnpkg/yarn/pull/3478/files#diff-9aeebf9c5a183aebcbff42975e733805R64

It's the old behavior to find the module folder, extracted into its own function. It's required, since until now the `files` entry was populated with the file entries starting from the `.yarn-integrity` location. Now that they are separated, the logic needs to be separated as well.

arcanis: Ah, I think I see what you mean - I'm cleaning those lint issues :)

bestander:
For a test you could probably ad a mix of https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/check.js#L180 and https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/check.js#L280

arcanis:
👍

bestander:
Seems legit, let's wait for tests to pass.
Would you mind adding some description?

arcanis:
Tests are passing locally, but failing on CircleCI ... I'm investigating 😕

bestander:
It could be a random issue

bestander:
However it is odd that check commands started timing out.
Maybe increase default timeout in check.js?

bestander:
jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;

PR #3480, opened 2017-05-23
Replace `no-async-without-await` plugin with in-built rule

KishanBagaria:
[`no-async-without-await` plugin by @zertosh](https://github.com/zertosh/eslint-plugin-no-async-without-await) is no longer needed since a rule with the same functionality was merged into ESLint core (https://github.com/eslint/eslint/pull/7435) a month after his plugin was published.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,12 +8,12 @@
     "yarn-internal"
   ],
   "rules": {
-    "yarn-internal/warn-language": 2,
-    "max-len": [2, 120],
-    "prefer-arrow-callback": 0,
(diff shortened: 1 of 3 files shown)


arcanis:
Does it work with the newly merged `no-return-await`? The doc is not clear on this.

KishanBagaria:
Yes, both rules are part of the core and work well together.

arcanis:
Nope, it doesn't work :/

KishanBagaria:
@arcanis They worked fine when I tried it. Repro steps?

PR #3482, opened 2017-05-23
Refactor and optimize `bin/yarn.js`

KishanBagaria:
* Running `bin/yarn.js` failed if `lib-legacy` wasn't built. Most developers run node v4+ and don't need/want to build the legacy files while developing for yarn. This PR also adds a gulp command `watch-modern`.

* Requiring files from the `lib-legacy` directory in node v4+ environments resulted in loading some modules from disk twice, once from `lib-legacy` and once from `lib`.
For example, previously, `constants.js` was loaded from the `lib-legacy` directory when `require('../lib-legacy/constants')` was called and later again from the `lib` directory when `require(path)` was called. Any module that was required by `constants.js` directly or indirectly, were all loaded twice.

### DIFF
--- a/bin/yarn.js
+++ b/bin/yarn.js
@@ -9,12 +9,12 @@ var semver = require('semver');
 var ver = process.versions.node;
 ver = ver.split('-')[0]; // explode and truncate tag from version #511
 
-var path = null;
+var dirPath = null;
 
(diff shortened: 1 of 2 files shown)


bestander:
Awesome, @KishanBagaria!

PR #3484, opened 2017-05-23
Exit process with code 1 when outdated dependencies are found

### LINKED ISSUE #3483: Make "yarn outdated" exit with a non-zero exit code when there are outdated dependencies

Feature (probably).

With outdated packages present, `yarn outdated` exits with `0`

The `npm outdated ` in the same situation will exit with `1` (as of NPM `4.2.0`).

With outdated packages present, `yarn outdated` should exit with non-zero exit code.

OS: Debian Jessie inside the Docker, [`library/node` image](https://hub.docker.com/_/node/), `7.10` tag.
Node: `v7.10.0`
Yarn: `0.24.4`

**P.S.**
Bundlers `bundle outdated` also returns non-zero exit code if outdated gems are present, just for the reference.
We use that part of behavior of both Bundler and NPM in automated checks on CI.

KishanBagaria:
**Summary**
Fixes #3483 

**Test plan**

### DIFF
--- a/src/cli/commands/outdated.js
+++ b/src/cli/commands/outdated.js
@@ -16,7 +16,7 @@ export function hasWrapper(): boolean {
   return true;
 }
 
-export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
+export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<number> {
   const lockfile = await Lockfile.fromDirectory(config.cwd);
   const install = new Install(flags, config, reporter, lockfile);
   let deps = await PackageRequest.getOutdatedPackages(lockfile, install, config, reporter);
@@ -44,7 +44,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     });
 
     reporter.table(['Package', 'Current', 'Wanted', 'Latest', 'Package Type', 'URL'], body);
+    return 1;
   }
-
-  return Promise.resolve();
+  return 0;
 }
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -155,11 +155,12 @@ if (command.requireLockfile && !fs.existsSync(path.join(config.cwd, constants.LO
 //
 const run = (): Promise<void> => {
   invariant(command, 'missing command');
-  return command.run(config, reporter, commander, commander.args).then(() => {
+  return command.run(config, reporter, commander, commander.args).then(exitCode => {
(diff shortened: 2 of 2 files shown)


src/cli/commands/outdated.js
```diff
@@ -44,6 +44,9 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     });
 
     reporter.table(['Package', 'Current', 'Wanted', 'Latest', 'Package Type', 'URL'], body);
+    process.on('exit', () =>
+      process.exitCode = 1
```
bestander: That looks like a hack.
We prefer to have a single exit code in cli/index.js that will be set code 1 if the command returned a rejected promise

KishanBagaria:
Good idea. What do you think of the update?

This is an alternative way of setting the exit code:
```
process.on('exit', () =>
  process.exit(EXIT_CODE)
);
```

[NPM does it with `process.exitCode = EXIT_CODE`](https://github.com/npm/npm/blob/latest/lib/outdated.js#L115) but we can't do that because of commander (see first comment above).

src/cli/index.js
```diff
@@ -155,11 +155,16 @@ if (command.requireLockfile && !fs.existsSync(path.join(config.cwd, constants.LO
 //
 const run = (): Promise<void> => {
   invariant(command, 'missing command');
-  return command.run(config, reporter, commander, commander.args).then(() => {
+  return command.run(config, reporter, commander, commander.args).then(exitCode => {
     reporter.close();
     if (outputWrapper) {
       reporter.footer(false);
     }
```
bestander: That would not work well, commands may resolve with non numeric values.
Also at line 337 in this file a rejection of run() promise causes process.exit(1) already.

Can you just throw an error in the outdated command?

KishanBagaria: Throwing an error causes the `yarn-error.log` file to be written and error message stuff to be printed. Am I missing something here?

I did a find all and none of the commands return anything. They all have the signature of `Promise<void>`. I've updated the PR again.

KishanBagaria:
It wasn't commander that was preventing `process.exitCode = EXIT_CODE` from working. It was yarn's `exit` method.

Another alternative way to do this:
The `exit` method in `cli/index.js` can be modified to not pass `0` to `process.exit`, and `outdated.js` can mutate `process.exitCode = 1`.

bestander:
I think this is better than modifying process, thanks @KishanBagaria!

PR #3497, opened 2017-05-25
Fix --frozen-lockfile not preventing lockfile changes

### LINKED ISSUE #3313: The --frozen-lockfile command-line option does not work
The `--frozen-lockfile` command-line option does not work as advertised. It behaves the same way as when the option is not supplied, modifying the yarn.lock.

Here's an example:

```
$ yarn --version
0.23.4
$ git status -s
$ yarn --frozen-lockfile
yarn install v0.23.4
[2/4] 🚚  Fetching packages...
success Saved lockfile.
✨  Done in 6.99s.
$ git status -s
 M yarn.lock
```

bug

The yarn.lock can be overwritten when `--frozen-lockfile` is provided on the command line.

Have a package.json that includes transitive dependencies with different versions. Sometimes, yarn will opportunistically collapse multiple dependencies on the same package into a single entry in the lock file if they can be satisfied by a single version. I'm not sure what triggers this. #579 contains a more thorough discussion.

Here is the link to a Travis CI build that exhibits the behavior: https://travis-ci.org/atomist/travis-rugs/builds/228523519
Here is the call to yarn that ended up with the yarn.lock being changed: https://github.com/atomist/travis-rugs/blob/c8b2dddfd129ae836a30864ec410ca9f01ff535e/.atomist/build/travis-build.bash#L68

yarn will make due with what is in the yarn.lock if it satisfies everything in package.json or fail if it can't.

```
$ node --version
v7.9.0
$ yarn --version
0.23.4
$ uname -a
Darwin mbp.local 15.6.0 Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64 x86_64
```

gpoole:
**Summary**

Fixes an issue where Yarn would rewrite yarn.lock instead of failing when `--frozen-lockfile` was used.

Fixes #3313.

**Test plan**

Added a test which checks the expected behaviour.


src/cli/commands/install.js
```diff
@@ -337,7 +337,7 @@ export class Install {
       return false;
     }
     const match = await this.integrityChecker.check(patterns, lockfileCache, this.flags);
-    if (this.flags.frozenLockfile && match.missingPatterns.length > 0) {
+    if (this.flags.frozenLockfile && (match.missingPatterns.length > 0 || match.integrityError)) {
```
bestander: Unfortunately that is not the fix.
integrityError can just mean that node_modules was corrupted.

bestander:
Thanks for the attempt, @gpoole.
The issue is deeper, since 0.23 Yarn started optimizing the lockfile during installs and this causes the changes.
We need to make sure lockfile does not change when install runs if it satisfies the package.json.
The issue is somewhere in the resolver.

gpoole:
That makes sense @bestander, it did seem like that's what was happening for me. I am seeing a lockfile error (`LOCKFILE_DONT_MATCH`) in when I run it against real projects that have this problem, but looks like the test is failing here so that's obviously not it. Is the integrity checker not comparing the newly generated structure with the original one read from the file?

bestander:
Integrity checker checks the integrity of previous installation in
node_modules folder and compares it to current lockfile and package.json
and CLI arguments.
It is not directly related to the issue with lockfiles changing but there
is a relation.

PR #3501, opened 2017-05-25
Fetcher overhaul

arcanis:
**Don't merge this PR**

You can skip the first three commits, they just put miniyarn into the source tree and make the PR bigger than it actually is. The cache and mirror paths are currently hardcoded, but they shouldn't be hard to resolve (the cache is easy, the main issue is the mirror, that is currently read from multiple registries - what I was talking this afternoon).

- [ ] Add flow types
- [ ] Make the resolver return a package tree, so that I can plug the linker
- [ ] Fix the link step, it currently breaks when reading a meta file that I don't generate
- [ ] Remove the cache / mirror from the rest of the code
- [ ] Use either always Immutable objects, or always native objects
- [ ] Streamline the utils (currently split between src/miniyarn/utils and src/util)

--- 

Some notes I made while working on integrating the fetchers inside Yarn. I'll update them as I read more (might be wrong on them due to me not understanding how they actually work!):

- The resolver is building a "flat" list of the dependency required in the top-level, and a map of each package that might get used by the linker. I feel like it should probably return a tree instead of this flat list so that when then it returns, we know for sure what will the package tree look like, without having to traverse everything again (currently, some work needs to be done by the linker). Doing this would allow us to separate the resolution process from everything else, which would make it much easier to refactor it / maintain it / swap it out at runtime (plugin).

- Related to the point discussed above, the current package linker seems to also be doing some part of the resolution process ([see here](https://github.com/yarnpkg/yarn/blob/8ede3c41cfb389ee408ae1094503649dc11b11cc/src/package-linker.js#L315-L350)) - I feel like this should be moved to the resolver, so that the linker would only be tasked with linking the project. The benefits would be the same: it would instantly become much easier to implement custom linkers (for example by hardlinking instead of copying). We could experiment various link strategies.


arcanis:
The last commit adds a stub `.yarn-metadata.json` file inside the directories returned by the fetchers. Yarn is now able to install registry packages just fine, but many tests are failing because I haven't implemented the directory fetching yet. That's my next step :)

arcanis:
This commit refactors a few things and implements a `LegacyMirrorFetcher`, which can be used to load (and save) tarballs generated by the current Yarn codebase (in this PR, the `CacheFetcher` and `MirrorFetcher` fetchers are completely responsible for managing both the cache and the mirror, and the rest of the code is completely unaware of the underlying caching/mirroring logic; it makes the code easier to think about, but it also requires to change a bit the archive format - hence this `LegacyMirrorFetcher`, which makes the code backward-compatible).

There's still a few issues (`"Cannot convert object to primitive value"` inside package-request), but I'll put this PR on hold until later next week, the time to work on some high-priority bugs :)

arcanis:
`__tests__/commands/install/integration.js` : 10 tests were failing this morning, 5 remaining

arcanis:
One test remaining in `integration`. The problem is that `generateHardModulePath` now returns a path that doesn't map anymore to the cache path (which follows a convention very similar to the mirror). I have an idea (basically, I'll just export the CacheFetcher to access a locator path)

arcanis:
All tests inside `__test__/commands/install/` are passing o/ Next step: fix the few breaks in `add.js` (from the look of it, it seems like most of these failures are simply due to the mirror change - should be easily fixable).

arcanis:
`__tests__/commands/add.js` is now passing. However, running the full testsuite seems to trigger a lot of ENFILE errors, I'm not sure how to workaround that :/

arcanis:
Found the culprit! By default, the `tmp` module manually opens a file descriptor to the files it creates.

bestander:
Great, progress, @arcanis!
I'll start reviewing tomorrow.

arcanis:
I'm going to close this one, it can't be merged in its current state. I keep the idea in mind, tho :)

Issue #3384, opened 2017-05-12
yarn upgrade should have a force flag


OmgImAlexis:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
`yarn upgrade -f` ignores the `-f` flag

**What is the expected behavior?**
`yarn upgrade -f` should update the packages to the latest version ignoring the version range specified in the `package.json`.

rally25rs:
As a workaround, you could use `yarn upgrade-interactive` which already ignores the version ranges.

rally25rs:
I'm thinking about adding a `--latest` flag instead of `-f` to be clearer in it's meaning.
It would ignore the package.json defined semver range when a package name with no explicit version is passed.

Examples:

Given a package.json

```
"left-pad": "<=1.1.1"
```

And the `left-pad` package has these versions defined in the registry:

```
0.0.9
1.0.0
1.1.0
1.1.1
1.1.2
1.1.3
2.0.0
```

# Existing Behavior:

> yarn upgrade

Upgrades to 1.1.1 because it matches the package.json semver range (<=1.1.1)

> yarn upgrade left-pad

Upgrades to 2.0.0 because a package was passed with no explicit version

> yarn upgrade left-pad@~1.1.2

Upgrades to 1.1.3 because it is the latest match for the passed in semver range (~1.1.2)

# New Behavior:

> yarn upgrade --latest

Upgrades to 2.0.0 because it would ignore the package.json semver range and use the latest as defined in the registry

>yarn upgrade left-pad --latest

Upgrades to 2.0.0 because that is the latest. This is what the command already does, so it is equivalent to ignoring the flag.

> yarn upgrade left-pad@~1.1.2 --latest

Upgrades to 1.1.3 because it is the latest match for the passed in semver range (~1.1.2).  This is what the command already does, so it is equivalent to ignoring the flag. (i.e. explicit version range as a parameter takes precedence over the --latest flag)

-----

In addition, I'd like to add the opposite as well, some way to specify

> yarn upgrade left-pad --SOMETHING

and have it respect the package.json defined semver.

I do this all the time where I have dozens of deps but just want to upgrade to the latest on 1 or 2 libraries in a certain range. For example, for a while my project was stuck on PhantomJS v1, but could not use any v2 release. I'd liked to have done a `upgrade phantomjs` and have it get the latest v1 release as defined in package.json, without having to specify it on the command line `upgrade phantomjs@^1.0.0`

[...]

Issue #3455, opened 2017-05-19
Figure out why unit tests occasionally time out


rally25rs:
CI builds often fail due to tests timing out. It also happens (very infrequently) when running tests locally.

Example output (this was from a local test run on a Macbook Pro, OSX Sierra, Node v6.3.0):

```
Summary of all failing tests
 FAIL  __tests__/commands/check.js (21.465s)
  ● --verify-tree should report wrong version

    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

      at tryOnTimeout (timers.js:224:11)
      at Timer.listOnTimeout (timers.js:198:5)

 FAIL  __tests__/commands/install/integration.js (184.171s)
  ● root install with optional deps

    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

      at tryOnTimeout (timers.js:224:11)
      at Timer.listOnTimeout (timers.js:198:5)
```

I'll try to dig into this as best I can, but I would appreciate some help... 🆘

rally25rs:
From discussion in 3510:

> > rally25rs: So all these tests are actually resolving real metadata? that seems fragile...
>
> arcanis: Think so. That's why our tests practically never completely pass: we always end up timeouting at least one of them 😞 Probably should fix this at some point! I thought maybe we could set up some sort of fake registry with only the few endpoints required, that would extract its data from a repository local folder.

If the tests making web requests and hitting live metadata repos, this could be a cause for fragile tests in the future.

It would be good to make a plan to make the tests run as "offline" as possible.

Since Yarn already supports a fully offline install through use of the cache and "offline mirror" features, would it be feasible to load all packages and metadata used for tests into an offline mirror, and then always run tests in an "offline" mode?

rally25rs:
Alternatively, could we just load the metadata package.json file from the repository into `__tests__/fixtures/request-cache/GET/...`?

rally25rs:
@arcanis It looks like some of the tests don't hit real HTTP requests.

There is a mock requester in `__tests__/__mocks__/request.js` that first checks for a file in `__tests__/fixtures/request-cache/GET/...`

If the file exists, it just returns it.

If the file does not exist, then it does the actual HTTP request, and writes the response to the request-cache directory (where a dev could then check it in to Git)

[...]

PR #3510, opened 2017-05-26
Add "latest" flag to "yarn upgrade"

rally25rs:
Implements requested feature in #3384 

Adds a `yarn upgrade [--latest/-L]` flag.

This flag instructs Yarn to upgrade all dependencies to the latest version specified by it's registry, ignoring the versions specified in `package.json`.

This is functionally equivalent to running `yarn upgrade <package>` and passing all dependency names (`yarn upgrade foo bar baz ...etc`).

**Test plan**

In test file `__tests__/commands/upgrade.js` :
* Added additional tests for the existing `yarn upgrade` behaviors.
* Added tests for behavior specific to `latest` flag.

**Documentation**

PR for the website documentation repo should be merged at the same time:
https://github.com/yarnpkg/website/pull/513


arcanis:
> This is functionally equivalent to running yarn upgrade <package> and passing all dependency names (yarn upgrade foo bar baz ...etc).

Shouldn't this flag be called `--all`, then?

rally25rs:
I've always found this to be the most confusing part of Yarn;

I always expect `yarn upgrade` to be the same as `yarn upgrade <all packages>` but it isn't.

* If you specify a package name, it ignores the version range in package.json.
* If you don't specify a package name, then it uses the version ranges in package.json.

I think `--latest` is more indicative of what it's doing:

`yarn upgrade` - upgrades all respecting version range in package.json
`yarn upgrade --latest` - upgrades all to latest, ignoring package.json

In other words, `yarn upgrade` itself is already a "--all" in that it upgrades all packages. This change is really about what version it upgrades all packages to.

... hope that makes sense. It's early and I'm on my 1st cup ☕️

cpojer:
I like `--latest`, I think the explanation makes sense.

src/cli/commands/upgrade.js
```diff
@@ -50,6 +52,11 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     } else {
       throw new MessageError(reporter.lang('scopeNotValid'));
     }
+  } else if (flags.latest && args.length === 0) {
+    addArgs = Object.keys(allDependencies)
+      .map(dependency => {
+        return getDependency(allDependencies, dependency);
+      });
```
cpojer: nit: you don't need the curly braces and the return here.

src/cli/commands/upgrade.js
```diff
@@ -11,6 +11,7 @@ export function setFlags(commander: Object) {
   // TODO: support some flags that install command has
   commander.usage('upgrade [flags]');
   commander.option('-S, --scope <scope>', 'upgrade packages under the specified scope');
+  commander.option('-L, --latest', 'upgrade packages to the latest version, ignoring version ranges in package.json');
```
cpojer: do we really need the `-L` shorthand? Let's just go with `--latest`.

cpojer:
Would you mind rebasing this diff and making it pass on CI? :)

src/cli/commands/upgrade.js
rally25rs: OK, was just following the convention of the other flags. I'll remove that `-L`.

rally25rs:
oops, I made the classic mistake of `git commit -a` and forgetting to `git add` new files 😢 
Merged latest master into the branch and fixed tests. Should be good to go now @cpojer

__tests__/commands/upgrade.js
```diff
@@ -102,6 +102,70 @@ test.concurrent('upgrades from fixed version to latest', (): Promise<void> => {
...
+test.concurrent('--latest upgrades to latest ignoring package.json when no package name passed', (): Promise<void> => {
+  return runUpgrade([], {latest: true}, 'range-to-latest', async (config): ?Promise<void> => {
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+    const lockEntryIndex = lockfile.indexOf('left-pad@^1.1.3:');
+
+    expect(lockEntryIndex).toEqual(0);
+    expect(lockfile.indexOf('left-pad@<=1.1.1:')).toEqual(-1);
+    expect(lockfile[lockEntryIndex + 1]).toContain('1.1.3');
+    expect(pkg.dependencies).toEqual({'left-pad': '^1.1.3'});
```
arcanis: You shouldn't use `left-pad` for this test, otherwise it will break if they release a new version. I think you can use `yarn-test` (cc @kittens ?)

__tests__/commands/upgrade.js
```diff
@@ -102,6 +102,70 @@ test.concurrent('upgrades from fixed version to latest', (): Promise<void> => {
...
+
+test.concurrent('upgrades to latest matching semver when package name passed with version', (): Promise<void> => {
+  return runUpgrade(['left-pad@~1.1.2'], {}, 'range-to-latest', async (config): ?Promise<void> => {
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+    const lockEntryIndex = lockfile.indexOf('left-pad@~1.1.2:');
+
+    expect(lockEntryIndex).toEqual(0);
+    expect(lockfile[lockEntryIndex + 1]).toContain('1.1.3');
+    expect(pkg.dependencies).toEqual({'left-pad': '~1.1.2'});
```
arcanis: You can use `left-pad` here, but then use `~1.0.1` (should resolve to `1.0.2`) instead of `~1.1.2`, since we know there won't be any more release to this branch, which is not the case for `1.1.x`.

__tests__/commands/upgrade.js
```diff
@@ -102,6 +102,70 @@ test.concurrent('upgrades from fixed version to latest', (): Promise<void> => {
...
+  return runUpgrade(['left-pad'], {}, 'range-to-latest', async (config): ?Promise<void> => {
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+    const lockEntryIndex = lockfile.indexOf('left-pad@^1.1.3:');
+
+    expect(lockEntryIndex).toEqual(0);
+    expect(lockfile.indexOf('left-pad@<=1.1.1:')).toEqual(-1);
+    expect(lockfile[lockEntryIndex + 1]).toContain('1.1.3');
+    expect(pkg.dependencies).toEqual({'left-pad': '^1.1.3'});
+  });
```
arcanis: This doesn't feel right ... shouldn't we have to pass the `--latest` flag to have this behavior? It contradicts the "no argument means all packages" behavior ("upgrades to latest matching package.json semver when no package name passed").

cpojer:
Oops, that was within the same 15 seconds! @rally25rs would you mind sending a follow-up PR to fix the issues that @arcanis pointed out?

__tests__/commands/upgrade.js
rally25rs: This test actually passes in current Yarn without my modifications (in other words, this is how Yarn has worked for a while). This is testing the command `yarn upgrade left-pad`, which is roughly equivalent to `yarn add left-pad`. The `add` command resolves to the passed in version range for `left-pad` which is unspecified, so it resolves to latest. 

There is some discussion around that confusing behavior in 3384 3266 3204
I've always sort of wondered if it's an unintentional behavior.

__tests__/commands/upgrade.js
rally25rs: Hmm, I actually thought the packages used during tests were all loaded from the 'offline mirror' directory of `__tests__/fixtures/request-cache/...` but it does like like just the left-pad tgz files are in there, not the metadata. So all these tests are actually resolving real metadata? that seems fragile...

I needed a package that had enough diverse version numbers to cover all these upgrade cases, and left-pad was the first one I found that worked.

Is there someplace I can see what version of `yarn-test` exist?

(side-note, it might be useful to document how these tests actually work and should be constructed, for new developers who want to contribute. It took me a while the first time I needed to add one.)

__tests__/commands/upgrade.js
rally25rs: I just re-read my comment and decided it's confusing... In clearer terms, the way Yarn 0.24 - 0.26 all work currently is that if your package.json is:

```
"dependencies": {
  "left-pad": "^1.0.0"
```

Then `yarn upgrade` will upgrade left-pad to the greatest match for `^1.0.0`.
However `yarn upgrade left-pad` will upgrade to `latest`.

I didn't change that behavior.
This PR just adds the option to:

`yarn upgrade --latest` which will upgrade left-pad to `latest` (ignoring what was in package.json)

There is currently, and with this PR, still no way to `yarn upgrade <package_name>` and have it use the range specified in package.json, unless you manually copy-paste the range yourself and do `yarn upgrade left-pad@^1.0.0`

arcanis: Yep, you're right. I feel like this should be considered as a bug ... otherwise there's no way to upgrade a specific package inside its allowed boundaries :/ can you open a separate PR where we could discuss this?

__tests__/commands/upgrade.js
arcanis: > So all these tests are actually resolving real metadata? that seems fragile...

Think so. That's why our tests practically never completely pass: we always end up timeouting at least one of them 😞 Probably should fix this at some point! I thought maybe we could set up some sort of fake registry with only the few endpoints required, that would extract its data from a repository local folder.

__tests__/commands/upgrade.js
rally25rs: I spawned #3603 as a discussion issue and added you as an assignee.

__tests__/commands/upgrade.js
rally25rs: We can discuss this in #3455

develar:
One problem with this flag — range specifier is not preserved:

 `"electron-osx-sign": "0.4.6"` 

changed to 

`"electron-osx-sign": "^0.4.6"`

i.e. package not upgraded, but range specifier changed.

Range specifier must be preserved.

arcanis:
@develar good remark, can you open an issue?

develar:
Filed https://github.com/yarnpkg/yarn/issues/3609 .

__tests__/commands/upgrade.js
rally25rs: The metadata for `left-pad` is taken from the `__tests__/fixtures/request-cache/` directory, so a new version pushed to the live registry would not break the test. These tests actually pass with no internet connection.

PR #3512, opened 2017-05-26
Allow tildes (~) in any path as a placeholder for the home directory

mjesun:
**Summary**

Tildes (`~`) were not correctly processed by Yarn when setting up the cache directory. Instead of expanding it to the home folder, it was creating a folder called `~` on the current directory where the command was executed.

The fix included proxying all `fs` calls through our custom module (`src/util/fs.js`), and expand each and every path we get through it. In places where this was not possible (e.g. external modules which accept paths, like `tar-fs`, an extra method called `fs.expandPath` is used just before calling them. Path expansion does not happen on Windows platforms.

Overridding the native `fs` module seemed way too risky and could lead to confusion.

**Test plan**

* Change your cache folder to point to something that includes the tilde; for example `~/foo`. Be sure to quote this, otherwise, your shell will automatically expand it.
* Open your `.yarnrc` to make sure path was correctly changed.
* Install any package, or perform any `yarn install` over an existing package.
* You should not see any folder called `~` on your current directory. Instead, you should have a new `foo` folder on your home directory, which contains all cached artifacts.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -6,6 +6,7 @@ import * as promise from './promise.js';
 import {promisify} from './promise.js';
 import map from './map.js';
 
+const userHome = require('./user-home-dir').default;
 const fs = require('fs');
 const globModule = require('glob');
 const os = require('os');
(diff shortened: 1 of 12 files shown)


bestander:
Thanks, @mjesun, from a glance looks solid.
I'll have a better look next week

KishanBagaria:
Read this comment on GH

src/util/fs.js
```diff
@@ -6,34 +6,47 @@ import * as promise from './promise.js';
...
-export const glob: (path: string) => Promise<Array<string>> = promisify(globModule);
+export const createWriteStream = (path: string, options?: Object): fs.WriteStream =>
+  fs.createWriteStream(expandPath(path), options);
+export const createReadStream = (path: string, options?: Object): fs.ReadStream =>
+  fs.createReadStream(expandPath(path), options);
+export const writeFileSync = (path: string, data: Buffer | string, options?: Object | string): void =>
+  fs.writeFileSync(expandPath(path), data, options);
+export const readFileSync = (path: string, options?: Object | string): any =>
+  fs.readFileSync(expandPath(path), options);
+export const existsSync = (path: string): boolean => fs.existsSync(expandPath(path));
```
bestander: I think we should not use util/fs.js as a proxy to all methods to Node/fs module.

I would add more code to support.
Also I would not want *Sync methods to be easily exposed because people sometimes are tempted to use them instead of async/await.

src/util/fs.js
```diff
@@ -808,3 +821,23 @@ export async function makeTempDir(prefix?: string): Promise<string> {
   await mkdirp(dir);
   return dir;
 }
+
+export function expandPath(path: string): string {
```
bestander: Is it an fs method?
From node perspective it would rather be in `path` modules rather than `fs`.

src/util/fs.js
```diff
@@ -6,34 +6,47 @@ import * as promise from './promise.js';
...
-export const stat: (path: string) => Promise<fs.Stats> = promisify(fs.stat);
-export const unlink: (path: string) => Promise<void> = promisify(require('rimraf'));
-export const mkdirp: (path: string) => Promise<void> = promisify(require('mkdirp'));
-export const exists: (path: string) => Promise<boolean> = promisify(fs.exists, true);
-export const lstat: (path: string) => Promise<fs.Stats> = promisify(fs.lstat);
-export const chmod: (path: string, mode: number | string) => Promise<void> = promisify(fs.chmod);
-export const link: (path: string) => Promise<fs.Stats> = promisify(fs.link);
-export const glob: (path: string) => Promise<Array<string>> = promisify(globModule);
+export const createWriteStream = (path: string, options?: Object): fs.WriteStream =>
+  fs.createWriteStream(expandPath(path), options);
```
bestander: Do we need to expandPath in every fs method?
I would add it only in the source of config.getOption method instead.
I don't think `~` needs to be expanded in every fs operation

bestander:
Thanks, @mjesun, good start.

mjesun:
@KishanBagaria @bestander Related to why all calls are proxied via `fs`, and not expanded when read + collapsed when writting, it is because it can lead to risky behaviors. Once you have expanded a path, both the unexpanded and expanded become indistinguishable between them. For instance:

```
~/cache
/home/mjesun/cache
```

They both become the same after expansion. If you want to write them back later or operate with them, you have lost their initial value, and you don't know if the `$HOME` was hardcoded or not. In other words, replacing `$HOME` later by a `~` felt like a too risky assumption on the user desire.

Because of that behavior, I thought that the common layer for transforming all paths should be as deep as possible, so the whole app moves paths with and without the tilde all along while being safe. That layer turns out to be the one that directly talks to the SO for managing files; i.e. the `fs`module.

bestander:
I was thinking of putting the resolution code into config.getOption then Yarn could contract to resolve ~ to HOME everywhere where it is used.
Putting it into FS seems a bit intrusive.
But thanks for going an extra mile and thinking about the side effects

src/util/fs.js
mjesun: Nice one! I created a `path` file equivalent and put the code there.

mjesun:
@bestander Nice! I will keep that one for future reference, in case we want to re-visit it. I created a new PR in #3537 🙂

PR #3513, opened 2017-05-26
Make lockfile Yarn and Node versions opt-in instead of opt-out

raido:
**Summary**

Based on https://github.com/yarnpkg/yarn/pull/3265#issuecomment-304353153 - this PR reverts the logic for lockfile Yarn and Node.js versions to be opt-in feature instead of opt-out.

**Test plan**

Includes unit test for generating file with versions included if config enabled in .yarnrc

If this gets accepted, it should be included in v0.26.0 as this would be breaking change for current v0.25.x release candidate.


SimenB:
Huge +1 for this, but I also vote for doing it in 0.25. It's just an RC, unless 0.26 stable is expected right after 0.25 is stable, it's gonna be quite a big churn

raido:
Yes, you are right, it would be nice land in v0.25 instead.

raido:
Was there an RFC for the initial feature implementation? I could not find it or maybe i missed it.

bestander:
A PR to 0.25 branch is welcome, we can release a patch

Issue #3324, opened 2017-05-04
[Feature request] Clean named package from local cache


kslimani:
> **Do you want to request a *feature* or report a *bug*?**

It is a request feature. 

> **What is the current behavior?**

Current behaviour is `yarn cache clean` will clean all cached package from local cache.

> **If the current behavior is a bug, please provide the steps to reproduce.**

Current behaviour is not a bug.

> **What is the expected behavior?**

I think i could be a nice addition to have a feature to remove a "named package" from local cache. *(an alternative from removing all by default)*

For example :

```shell
$ yarn cache clean <PACKAGE_NAME>
```

*This syntax is just a proposition, there may be a better syntax to achieve this.*

What do you guys think about this ?

bestander:
Nice feature, send a PR

kslimani:
I close the issue, as #3523 is merged.

PR #3523, opened 2017-05-29
Allow to clean named package from cache

kslimani:
**Summary**

This minor feature is described in #3324 . It introduce an optional argument to existing `yarn cache clean` command.

The usage become : `yarn cache clean [package_name]`

For example, `yarn cache clean yargs` will remove all versions of `yargs` from cache (_if exists in cache folder, otherwise will display a warning message_)

**Motivations**

Default behaviour is to clean all the cache. But for some reasons, you may need to only clean a single package from cache.

**Test plan**

I added a test in `__tests__/commands/cache.js`.

### DIFF
--- a/__tests__/commands/cache.js
+++ b/__tests__/commands/cache.js
@@ -70,3 +70,23 @@ test('clean', async (): Promise<void> => {
     expect(files.length).toEqual(0);
   });
 });
+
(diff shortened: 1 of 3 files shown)


kslimani:
BTW, does this require me to write a PR in RFC repository ?

bestander:
nice, @kslimani!

PR #3527, opened 2017-05-30
Fix GitFetcher.fetchFromLocal using wrong hash

Ngo-The-Trung:
Found on 0.25.3

Following up #3449

expectHash is used beforehand to generate the cache destination, but actualHash is returned. This causes future cache destination path generation to use the wrong hash, resulting in EOENT errors.

@bestander 

Full story:

PackageFetcher generates `dest` parameter using Config.generateHardModulePath before fetching using expectHash i.e. the commit hash

GitFetcher returns actualHash (i.e. tarball hash) as result of fetch

Some time later updateManifest is called with the fetch result hash (actualHash)

This pollutes the resolved patterns cache (PackageResolver.patterns) with the wrong hash (actualHash)

Future Config.generateHardModulePath calls getResolvedPattern and gets actualHash, (as opposed to expectHash. This causes it to look for the cached package in the wrong directory.

### DIFF
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -92,13 +92,15 @@ export default class GitFetcher extends BaseFetcher {
         .pipe(untarStream)
         .on('finish', () => {
           const expectHash = this.hash;
+          invariant(expectHash, 'Commit hash required');
+
           const actualHash = hashStream.getHash();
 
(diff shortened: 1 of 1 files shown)


bestander:
@arcanis, this is for you to review

Ngo-The-Trung:
@arcanis fixed

arcanis:
Thanks 👍

PR #3528, opened 2017-05-30
PATCH for using .yarnrc registry url instead of default https://regis…

Arnaud-Nauwynck:
…try.yarnpkg.com/

**Summary**

**Test plan**

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -55,7 +55,10 @@ export default class NpmRegistry extends Registry {
 
   request(pathname: string, opts?: RegistryRequestOptions = {}, packageName: ?string): Promise<*> {
     const registry = this.getRegistry(packageName || pathname);
-    const requestUrl = url.resolve(registry, pathname);
+    const requestBaseUrl = url.resolve(registry, pathname); // does not work...
+    const requestUrl = requestBaseUrl.replace('https://registry.yarnpkg.com/', registry);
+    // console.log('PATCH npm-registry.request(' + pathname + '..)  => requestUrl:' + requestUrl + ' instead of ' + requestBaseUrl);
+    
     const alwaysAuth = this.getRegistryOrGlobalOption(registry, 'always-auth');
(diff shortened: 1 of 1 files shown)


src/registries/npm-registry.js
```diff
@@ -55,7 +55,10 @@ export default class NpmRegistry extends Registry {
 
   request(pathname: string, opts?: RegistryRequestOptions = {}, packageName: ?string): Promise<*> {
     const registry = this.getRegistry(packageName || pathname);
-    const requestUrl = url.resolve(registry, pathname);
+    const requestBaseUrl = url.resolve(registry, pathname); // does not work...
```
bestander: what does not work?

src/registries/npm-registry.js
```diff
@@ -55,7 +55,10 @@ export default class NpmRegistry extends Registry {
...
+    const requestUrl = requestBaseUrl.replace('https://registry.yarnpkg.com/', registry);
```
bestander: I think you are on a right path.
Just make the domains configurable via .yarnrc.
Those can be read in config via getOption, e.g. https://github.com/yarnpkg/yarn/blob/master/src/config.js#L267

src/registries/npm-registry.js
```diff
@@ -55,7 +55,10 @@ export default class NpmRegistry extends Registry {
...
+    // console.log('PATCH npm-registry.request(' + pathname + '..)  => requestUrl:' + requestUrl + ' instead of ' + requestBaseUrl);
```
bestander: comment unnecessary

bestander:
Thanks, @Arnaud-Nauwynck. 
This is almost good enough but needs some configuration and a test.
Let me know if you are willing to address those.
Otherwise we could wait for someone else to pick it up

bestander:
ping, @Arnaud-Nauwynck

bestander:
I'll close as it can't be merge this way, please feel free to reopen after rebasing and addressing the issue.

PR #3537, opened 2017-05-30
Allow tilde to be used in config paths, expanding it to $HOME

mjesun:
**Summary**

Tildes (`~`) were not correctly processed by Yarn when setting up the cache directory. Instead of expanding it to the home folder, it was creating a folder called `~` on the current directory where the command was executed.

The fix included proxying all `fs` calls through our custom module (`src/util/fs.js`), and expand each and every path we get through it. In places where this was not possible (e.g. external modules which accept paths, like `tar-fs`, an extra method called `fs.expandPath` is used just before calling them. Path expansion does not happen on Windows platforms.

Overridding the native `fs` module seemed way too risky and could lead to confusion.

**Test plan**

* Change your cache folder to point to something that includes the tilde; for example `~/foo`. Be sure to quote this, otherwise, your shell will automatically expand it.
* Open your `.yarnrc` to make sure path was correctly changed.
* Install any package, or perform any `yarn install` over an existing package.
* You should not see any folder called `~` on your current directory. Instead, you should have a new `foo` folder on your home directory, which contains all cached artifacts.

### DIFF
--- a/__tests__/util/path.js
+++ b/__tests__/util/path.js
@@ -1,0 +1,45 @@
+/* @flow */
+
+jest.mock('../../src/util/user-home-dir.js', () => ({
+  default: '/home/foo',
+}));
+
+import {expandPath} from '../../src/util/path.js';
+
+describe('fileDatesEqual', () => {
+  const realPlatform = process.platform;
+
+  describe('!win32', () => {
+    beforeAll(() => {
+      process.platform = 'notWin32';
+    });
+
+    afterAll(() => {
+      process.platform = realPlatform;
+    });
+
+    test('Paths get expanded', () => {
+      expect(expandPath('~/bar/baz/q')).toEqual('/home/foo/bar/baz/q');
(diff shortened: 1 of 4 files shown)


__tests__/util/path.js
```diff
@@ -0,0 +1,45 @@
...
+}));
+
+import {expandPath} from '../../src/util/path.js';
+
+describe('fileDatesEqual', () => {
+  const realPlatform = process.platform;
+
+  describe('!win32', () => {
+    beforeAll(() => {
+      process.platform = 'notWin32';
```
bestander: There is a trick.
Jest runs tests on CI concurrently, even if in one file none of the tests are `test.concurrent`, CI will run tests suites in parallel.
And currently there is no elegant way to make sure that this test is running in isolation.

Double checking with @cpojer  that I am right.

How about a just running this test on non windows CI?
Yarn runs on OSX, linux and Windows, we'll notice if the test breaks

bestander: Nevermind, @cpojer says that process is a separate object per test

Issue #685, opened 2016-10-11
Yarn pack is ignoring .npmignore files


psachs21:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

If I specify a .npmignore with a bunch of files that should NOT be published along with the package, a bunch of files seem to be included no matter what and some aren't included at all.

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Take existing project. My project has a lib directory. Create a file `.npmignore'
2. Add a few directories that you want to exclude from the package.
   eg

```
src
node_modules
```
1. Run `yarn pack`
2. Run `tar -ztvf <package>`

Notice that the package contains README.md files that were specified in node_modules AND in src. It also does not include the directory lib.

**What is the expected behavior?**

node_modules and src should not exist in the archive. lib should exist in the archive.

**Please mention your node.js, yarn and operating system version.**

OS X: 10.11.6
node: 5.5.0
yarn: 0.15.1

kittens:
Thanks for the report! Can you post the `package.json` you have? (Feel free to censor sensitive parts) We have some magic to handle the `files` and `bundledDependencies` fields which may be effecting this.

psachs21:
Sure.

```
{
  "name": "patternity",
  "version": "2.0.0",
  "description": "Patternity is the pattern library and style guide for all Influitive apps",
  "main": "index.js",
  "scripts": {
    "test": "tape-babel-css-modules \"src/**/*/*.test.js\" | faucet && karma start",
    "test-circleci": "tape-babel-css-modules \"src/**/*/*.test.js\" | faucet && karma start --single-run --log-level error",
    "karma-test": "karma start",
    "tape-watch": "watch 'clear && tape-babel-css-modules \"src/**/*/*.test.js\" | faucet && exit 0' src -d -u",
    "coverage": "babel-node node_modules/.bin/isparta cover src/**/*/*.test.js | tap-spec",
    "babel": "babel {src,infl-components-src} --out-dir lib --ignore *.test.js && gulp copy-lib-styles",
    "babel-watch": "babel {src,infl-components-src} -w --out-dir lib & gulp watch-scss",
    "testbed": "node ./webpacktestbed/devServer.js",
    "postinstall": "npm run babel",
    "create": "gulp create-component",
    "lint": "eslint ./src; exit 0",
    "lint:check": "eslint ./src",
[...]

PR #3538, opened 2017-05-30
Give higher priority to .npmignore over .gitignore

GAumala:
Here is a solution for #685. I found that the source of the problem is that yarn gives the same priority to `.npmignore` and `.gitignore` while npm does not. According to [npm's developer guide](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package):

> Use a .npmignore file to keep stuff out of your package. If there's no .npmignore file, but there is a .gitignore file, then npm will ignore the stuff matched by the .gitignore file. If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it. Like git, npm looks for .npmignore and .gitignore files in all subdirectories of your package, not only the root directory.

In this patch, If `.npmignore` or `.yarnignore` are present in a directory that also contains a `.gitignore`, don't take into account the `.gitignore` when running the `pack` command.

I created a new function `filterOverridenGitignores` in the the `util/filter.js` module that takes a `WalkFiles` object with all the potential package files and returns another `WalkFiles` object only with the relevant dot ignore files (`.npmignore`, `.yarnignore` or `.gitignore`), taking into account the priority rules. This new collection is the used by the `pack` command to create the ignore regexes. 

I have also added a unit test for this function. For a more thorough test just run the `pack` command on any project and compare the contents of the tarball.

### DIFF
--- a/__tests__/util/filter.js
+++ b/__tests__/util/filter.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import {ignoreLinesToRegex} from '../../src/util/filter.js';
+import {ignoreLinesToRegex, filterOverridenGitignores} from '../../src/util/filter.js';
 
 test('ignoreLinesToRegex', () => {
   expect(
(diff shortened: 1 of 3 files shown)


bestander:
Thanks, @GAumala.

bestander:
cc @BYK, this fixes some of the publish errors we have seen

BYK:
Nice!

I feel like we should not support merging multiple ignore files at all tho. In my mind it goes like `.yarnignore` -> `.npmignore` -> `.gitignore` and only one file gets everything. So if I have both `.yarnignore` and `.npmignore` files, we should only honor `.yarnignore`. Am I wrong?

bestander:
Seems right considering that we don't merge gitignore

PR #3544, opened 2017-05-31
Auto detect and merge lockfile conflicts

kittens:
**Summary**

Git merge conflicts are common with `yarn.lock`. This PR allows you to run `yarn` to automatically merge the two versions of `yarn.lock` which will run Yarn and regenerate a new lockfile.

**Test plan**

Added tests.

### DIFF
--- a/__tests__/lockfile.js
+++ b/__tests__/lockfile.js
@@ -11,19 +11,19 @@ const objs = [{foo: 'bar'}, {foo: {}}, {foo: 'foo', bar: 'bar'}, {foo: 5}];
 let i = 0;
 for (const obj of objs) {
   test(`parse/stringify ${++i}`, () => {
-    expect(parse(stringify(obj))).toEqual(nullify(obj));
+    expect(parse(stringify(obj)).object).toEqual(nullify(obj));
   });
 }
 
 test('parse', () => {
-  expect(parse('foo "bar"')).toEqual(nullify({foo: 'bar'}));
-  expect(parse('"foo" "bar"')).toEqual(nullify({foo: 'bar'}));
-  expect(parse('foo "bar"')).toEqual(nullify({foo: 'bar'}));
+  expect(parse('foo "bar"').object).toEqual(nullify({foo: 'bar'}));
+  expect(parse('"foo" "bar"').object).toEqual(nullify({foo: 'bar'}));
+  expect(parse('foo "bar"').object).toEqual(nullify({foo: 'bar'}));
 
-  expect(parse(`foo:\n  bar "bar"`)).toEqual(nullify({foo: {bar: 'bar'}}));
-  expect(parse(`foo:\n  bar:\n  foo "bar"`)).toEqual(nullify({foo: {bar: {}, foo: 'bar'}}));
-  expect(parse(`foo:\n  bar:\n    foo "bar"`)).toEqual(nullify({foo: {bar: {foo: 'bar'}}}));
-  expect(parse('foo:\n  bar:\n    yes no\nbar:\n  yes no')).toEqual(
+  expect(parse(`foo:\n  bar "bar"`).object).toEqual(nullify({foo: {bar: 'bar'}}));
+  expect(parse(`foo:\n  bar:\n  foo "bar"`).object).toEqual(nullify({foo: {bar: {}, foo: 'bar'}}));
+  expect(parse(`foo:\n  bar:\n    foo "bar"`).object).toEqual(nullify({foo: {bar: {foo: 'bar'}}}));
+  expect(parse('foo:\n  bar:\n    yes no\nbar:\n  yes no').object).toEqual(
     nullify({
       foo: {
         bar: {
@@ -192,4 +192,45 @@ test('Lockfile.getLockfile (sorting)', () => {
   };
 
   expect(actual).toEqual(expected);
+});
+
+test('parse merge conflicts', () => {
+  const file = `
+a:
+  no "yes"
+
+<<<<<<< HEAD
+b:
+  foo "bar"
+=======
+c:
+  bar "foo"
+>>>>>>> branch-a
+
+d:
+  yes "no"
+`;
+
+  const {type, object} = parse(file);
+  expect(type).toEqual('merge');
+  expect(object.a.no).toEqual('yes');
+  expect(object.b.foo).toEqual('bar');
+  expect(object.c.bar).toEqual('foo');
+  expect(object.d.yes).toEqual('no');
+});
+
+test('parse merge conflict fail', () => {
+  const file = `
+<<<<<<< HEAD
+b:
+  foo: "bar"
+=======
+c:
+  bar "foo"
+>>>>>>> branch-a
+`;
+
+  const {type, object} = parse(file);
+  expect(type).toEqual('conflict');
+  expect(Object.keys(object).length).toEqual(0);
 });
--- a/__tests__/lockfile.js
+++ b/__tests__/lockfile.js
@@ -194,7 +194,7 @@ test('Lockfile.getLockfile (sorting)', () => {
   expect(actual).toEqual(expected);
 });
 
-test('parse merge conflicts', () => {
+test('parse single merge conflict', () => {
   const file = `
 a:
   no "yes"
@@ -219,6 +219,41 @@ d:
   expect(object.d.yes).toEqual('no');
 });
 
+test('parse multiple merge conflicts', () => {
+  const file = `
+a:
+  no "yes"
+
+<<<<<<< HEAD
+b:
+  foo "bar"
+=======
+c:
+  bar "foo"
+>>>>>>> branch-a
+
+d:
+  yes "no"
+
+<<<<<<< HEAD
+e:
+  foo "bar"
+=======
+f:
+  bar "foo"
+>>>>>>> branch-b
+`;
+
+  const {type, object} = parse(file);
+  expect(type).toEqual('merge');
+  expect(object.a.no).toEqual('yes');
+  expect(object.b.foo).toEqual('bar');
+  expect(object.c.bar).toEqual('foo');
+  expect(object.d.yes).toEqual('no');
+  expect(object.e.foo).toEqual('bar');
+  expect(object.f.bar).toEqual('foo');
+});
+
 test('parse merge conflict fail', () => {
   const file = `
 <<<<<<< HEAD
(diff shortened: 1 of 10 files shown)


src/lockfile/parse.js
```diff
@@ -314,9 +314,80 @@ export class Parser {
   }
 }
 
-export default function(str: string, fileLoc: string = 'lockfile'): Object {
-  str = stripBOM(str);
+const GIT_MERGE_CONFLICT_START = '<<<<<<<';
```
cpojer: Drop the "GIT_", other tools will generate similar conflict markers.

src/lockfile/parse.js
```diff
@@ -314,9 +314,105 @@ export class Parser {
...
+          variantOne.push(line);
+        }
+      }
+      variants[0] = variants[0].concat(variantOne);
+
+      // get the second variant
+      let variantTwo = [];
+      while (lines.length) {
+        const line = lines.shift();
+        if (line.startsWith(MERGE_CONFLCIT_END)) {
```
arcanis: typo

src/lockfile/parse.js
```diff
@@ -314,9 +314,105 @@ export class Parser {
...
+    const line = lines.shift();
+    if (line.startsWith(MERGE_CONFLICT_START)) {
+      // get the first variant
+      let variantOne = [];
+      while (lines.length) {
+        const line = lines.shift();
+        if (line === MERGE_CONFLICT_SEP) {
+          break;
+        } else {
+          variantOne.push(line);
```
arcanis: can't you just push directly into `variants[0]`? or at least use `let variantOne = variants[0]` so that the concat becomes unneeded

src/lockfile/parse.js
```diff
@@ -314,9 +314,105 @@ export class Parser {
...
+        }
+      }
+      variants[0] = variants[0].concat(variantOne);
+
+      // get the second variant
+      let variantTwo = [];
```
arcanis: same as above

kittens:
Addressed nits, thanks @arcanis!

arcanis:
@kittens The tests seem to break [here](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/check.js#L176), and in a few other place (which all seem related to the resolver .. which is weird since your PR doesn't alter it ôo)

src/lockfile/parse.js
```diff
@@ -314,9 +314,101 @@ export class Parser {
...
+  const variants: Array<Array<string>> = [[], []];
+  const lines = str.split(/\n/g);
+
+  while (lines.length) {
+    const line = lines.shift();
+    if (line.startsWith(MERGE_CONFLICT_START)) {
+      // get the first variant
+      while (lines.length) {
+        const line = lines.shift();
+        if (line === MERGE_CONFLICT_SEP) {
```
dustinspecker: This feature looks awesome.

Curious if you all are wanting to support git's `diff3` merge conflict style (https://git-scm.com/docs/git-merge#_how_conflicts_are_presented). I think this code will break for users using `diff3` because the first variant would include both the real variant and the common parent ancestor of both conflicts. I wouldn't mind making a PR in the future for supporting `diff3`.

Sorry if I'm overlooking something.

src/lockfile/parse.js
```diff
@@ -314,9 +314,101 @@ export class Parser {
...
+  type: 'merge' | 'none' | 'conflict',
+  object: Object,
+};
+
+/**
+ * Parse and merge the two variants in a conflicted lockfile.
+ */
+
+function parseWithConflict(str: string, fileLoc: string): ParseResult {
+  const variants = extractConflictVariants(str);
```
sejoker: We assume here that only 1 merge conflict exists. We could handle 2+ conflicts if call `parseWithConflict` recursively.

__tests__/lockfile.js
```diff
@@ -193,3 +193,44 @@ test('Lockfile.getLockfile (sorting)', () => {
 
   expect(actual).toEqual(expected);
 });
+
+test('parse merge conflicts', () => {
```
sejoker: better name -> 'parse single merge conflict'

src/lockfile/parse.js
arcanis: It supports multiple conflicts in the same file, but during a same conflict pass. I think it should be good enough for a first iteration :)

sejoker: @arcanis cool, would be nice to add an extra test for in that case.

caesarsol:
@kittens neat! Could you please add git diff3 conflictstyle support?
http://psung.blogspot.it/2011/02/reducing-merge-headaches-git-meets.html?m=1

cpojer:
Rebased, fixed up the tests, made flow more strict. Added support for diff3 by discarding the common ancestor from the parsed result. It seems to me like the information shown is purely for humans to make better assumptions about what to do in case of a conflict, which doesn't apply here. Let me know if that is wrong.

Anything else needed here to get this merged?

src/lockfile/parse.js
```diff
@@ -314,9 +319,87 @@ export class Parser {
...
+  const variants = [[], []];
+  const lines = str.split(/\n/g);
+  let skip = false;
+
+  while (lines.length) {
+    const line = lines.shift();
+    if (line.startsWith(MERGE_CONFLICT_START)) {
+      // get the first variant
+      while (lines.length) {
+        const line = lines.shift();
```
BYK: Better to not mask the outer variable?

src/lockfile/parse.js
```diff
@@ -314,9 +319,87 @@ export class Parser {
...
+  }
+
+  return [variants[0].join('\n'), variants[1].join('\n')];
+}
+
+/**
+ * Check if a lockfile has merge conflicts.
+ */
+function hasMergeConflicts(str: string): boolean {
+  return str.includes(MERGE_CONFLICT_START);
```
BYK: Wouldn't it be better to check if all 3 markers exist?

__tests__/lockfile.js
```diff
@@ -193,3 +193,101 @@ test('Lockfile.getLockfile (sorting)', () => {
 
   expect(actual).toEqual(expected);
 });
+
+test('parse single merge conflict', () => {
+  const file = `
```
cpojer: It's much easier to manage it like this, and we care more about the text input than we care about where it is coming from for the purpose of this test.

__tests__/lockfile.js
```diff
@@ -193,3 +193,101 @@ test('Lockfile.getLockfile (sorting)', () => {
...
+  expect(object.e.foo).toEqual('bar');
+  expect(object.f.bar).toEqual('foo');
+});
+
+test('parse merge conflict fail', () => {
+  const file = `
+<<<<<<< HEAD
+b:
+  foo: "bar"
+=======
```
cpojer: Ah, it might not be obvious but in this example, the format is actually wrong because yarn lockfiles don't support colons `:`, which is why it results in a conflict.

cpojer:
I just tried this on Jest https://github.com/facebook/jest/pull/3906 and it worked flawlessly, this feature is so awesome. @kittens thanks for the idea and getting us 99% there!

screen shot 2017-07-07 at 2 06 16 pm

wodin:
@cpojer, it's a bad habit to do this:
$ rm -rf packages/*/yarn.lock
Sorry. Pet peeve :) There's no recursion happening there so why use -r? Also it's unlikely you'd need the -f.
But I know of someone at a small ISP who was trying to delete a mailbox and typed "rm -rf /var/spool/mail/" and then pasted the username and pressed enter. When it took longer than expected he suddenly realised there was a space before the username. A lot of mail was lost. Would not have been a problem if he had not used "-rf".
I suspect someone (possibly your distribution) encouraged that habit by aliasing "rm" to "rm -i", but the fix for that is to remove the stupid alias! :) Not to use "-rf" as a matter of course.
Anyway, that's enough of a rant from me. Sorry for the off-topic comment.

raix:
Would be nice to update the `# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.` in yarn.lock adding a hint to guide the user?
*(sorry about the late note)*

Vanuan:
What are implications of using `yarn install --pure-lock-file` wrt resolving conflicts?

BYK:
@Vanuan I expect that to not change the file at all but use the merged version in-memory.

Vanuan:
So how do I resolve conflicts without upgrading packages? Only by hand?

BYK:
@Vanuan I don't really follow that reasoning. `yarn install` without `--pure-lockfile` wouldn't (or at least shoudn't) upgrade packages.

Vanuan:
Hm. Somehow I was under impression that it would. I've seen 3 situations where yarn.lock ends up changed:

1. edit a version in package.json
2. run yarn add/remove
3. remove yarn.lock to resolve merge conflicts and run yarn.install

The 3rd situation is mitigated by this PR. But it isn't clear whether conflicted packages would be resolved again potentially upgrading the version.
The 2nd situation might upgrade versions if there are requirements of installed/removed packages changed (and it doesn't conflict with package.json)
But the 1st situation I'm not sure about. Might be the same as 2nd.

Well, it looks like `--pure-lock-file` only helps if `package.json` constraints don't match `yarn.lock`. I.e. when `yarn install` would change `yarn.lock`. Probably that situation should be catched by CI:

1) run `yarn install`
2) check if `git diff --exit-code` fails and faild the build with message "yarn.lock isn't up to date"

Vanuan:
The "upgrade" i'm referring to is when `yarn install` fetches new version that satisfies the version in `package.json`

1. yarn install with package@^1.0.0 generates lock file with package@1.0.1
2. package 1.1.0 is released
3. yarn install with package@^1.0.0 generates lock file with package@1.1.0

So the question is whether package version gets reevaluated in case of merge conflicts.

BYK:
@Vanuan it is quite hard to keep track of what needs to be done under a closed PR. Would you mind summarizing this in a new issue as a feature request or a bug so it is clear to anyone outside of this PR what is asked to be done. This way it would be asier to reason about and work on it.

Vanuan:
Thanks for you patience. This probably doesn't belong here on GitHub.

cpojer:
I'm locking this. The conversation that evolved isn't related to the original issue, if you have concerns about this issue, please send a pull request to fix it. Thanks!

Issue #3312, opened 2017-05-03
yarn install --prod installs many more packages than npm install --prod


coreyfarrell:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
`yarn install --prod` installs many packages that are not requested by package.json.

**If the current behavior is a bug, please provide the steps to reproduce.**
* Download minimist - https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz
* Extract and enter folder.
* Run `npm install --prod`
* Note that no modules are installed (minimist has no production deps).
* Run `yarn install --prod` (no yarn.lock exists before this command)
* Not that many modules are installed.

**What is the expected behavior?**
Extra modules should not be installed.

**Please mention your node.js, yarn and operating system version.**
node.js: v7.9.0
yarn: 0.23.4
CentOS 7
npm: 4.2.0

bestander:
Yeah, I suppose there is no need to download and unzip the tarballs that are dev only dependencies.
Send a PR

pichouk:
@bestander said in #3577 that it was fix in master, so I guess it can be closed.

PR #3546, opened 2017-05-31
Ignore unused patterns to avoid dev dependencies download when installing with --production flag (#3312)

pichouk:
**Summary**
As described on #3312, `yarn` do not consider the `--production` flag when installing packages (with `yarn install --production`) and try to download dev dependencies too. It can be really annoying since, sometimes, production environment does not have access to private packages used in development. 
I try to set a boolean to `true` which, I guess, will disable this behavior and only download real dependencies.

**Test plan**
I build my modified version of `yarn` and test some install to verify it have the correct behavior.

On an empty folder with the following `package.json` (notice the not existing package in `devDependencies`) :
```
{
  "name": "testproject",
  "version": "0.5.6",
  "license": "UNLICENSED",
  "description": "Fake project to test yarn install --production",
  "dependencies": {
    "lodash": "^4.17.4"
  },
  "devDependencies": {
    "notExistingProjeCt": "^1.17.65"
  }
}
```

Install with version `0.24.5` :
```
kyane@kyane-pc:/tmp/test-yarn$ yarn --version
0.24.5
kyane@kyane-pc:/tmp/test-yarn$ yarn install --production
yarn install v0.24.5
[1/4] Resolving packages...
error Couldn't find package "notExistingProjeCt" on the "npm" registry.
```

Install with my modified version :
```
kyane@kyane-pc:/tmp/test-yarn$ ~/code/yarn/dist/bin/yarn --version
0.26.0-0
kyane@kyane-pc:/tmp/test-yarn$ ~/code/yarn/dist/bin/yarn install --production
yarn install v0.26.0-0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
Done in 0.42s.
```

Can see that `yarn` does not try to resolve dev dependencies anymore :)

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -191,7 +191,7 @@ export class Install {
 
   async fetchRequestFromCwd(
     excludePatterns?: Array<string> = [],
-    ignoreUnusedPatterns?: boolean = true,
+    ignoreUnusedPatterns?: boolean = false,
   ): Promise<InstallCwdRequest> {
     const patterns = [];
     const deps = [];
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -191,7 +191,7 @@ export class Install {
 
   async fetchRequestFromCwd(
     excludePatterns?: Array<string> = [],
-    ignoreUnusedPatterns?: boolean = false,
+    ignoreUnusedPatterns?: boolean = true,
   ): Promise<InstallCwdRequest> {
     const patterns = [];
     const deps = [];


pichouk:
Ok tests failed, seems more complicated that I was thinking...

bestander:
Yeah, it is more complex.
We still need to resolve all dev dependencies but we need to not fetch them in package-fetcher.

bestander:
Thanks for giving it a try, @pichouk!
Much appreciated.

pichouk:
@bestander I searched a little tonight and I cannot found have to solve this.
It seem that the `markIgnored` function in `src/cli/commands/install.js` *need* to resolve package to mark it as ignored. And the resolving step raise an error when the dependency is not accessible (eg. a private dev dependency). I don't know how to do it, I'll see later but maybe it is too difficult for a "Yarn newbie" :p

pichouk:
Closing since I have not real solution for this issue.
I guess Yarn assume that all dependencies (even dev dependencies) should be resolved even when using `--production` flag. I think it is really annoying since you cannot install a project using only public dependencies on production if there is some private dev dependencies...

Issue #3154, opened 2017-04-14
yarn no longer saves git/github packages in offline mirror


agibralter:
**Do you want to request a *feature* or report a *bug*?**

*bug*

**What is the current behavior?**

Given our `./package.json` like this:

```json
{
  "dependencies": {
    "@kadira/react-storybook-addon-info": "^3.3.0",
    "@kadira/storybook-ui": "^3.11.0",
    "internal-pkg": "git+ssh://git@github.com/me/internal-pkg.git#abcdef0",
    ...
  }
}
```

And `./.yarnrc`:

```
yarn-offline-mirror "./vendor/npm_cache"
yarn-offline-mirror-pruning true
```

Running `yarn` no longer seems to save the dependencies from github/git in our `npm_cache`.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Running `ls` on the `./vendor/npm_cache` should include non-npm-hosted packages.

**Please mention your node.js, yarn and operating system version.**

macOS 10.12.4:
```
$ which node
/Users/me/.nvm/versions/node/v6.9.4/bin/node

$ node -v
v6.9.4

$ which yarn
/usr/local/bin/yarn

$ yarn --version
0.23.2
```

(Node is installed via `nvm` and `yarn` via `brew`)

tribou:
I'm getting the same issue when yarn tries to resolve `console-polyfill` inside the `rollbar-browser` package:

```
yarn install v0.23.2
warning You don't appear to have an internet connection. Try the --offline flag to use the cache for registry queries.
[1/4] 🔍  Resolving packages...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://github.com/rollbar/console-polyfill.git: getaddrinfo ENOTFOUND github.com github.com:443".
info If you think this is a bug, please open a bug report with the information provided in "/Users/tribou/dev/react-template/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

```
> cat yarn.lock | grep -A 3 console-polyfill
console-polyfill@rollbar/console-polyfill#eab6ff9d2b7597fc2f259baa18100556bdb94dbe:
  version "0.2.3"
[...]

PR #3548, opened 2017-05-31
Calls setupMirrorFromCache when fetching through the git-fetcher

arcanis:
Otherwise, the mirror will not be populated with data imported from the cache. Cf #3154


bestander:
Any test you could come up with?

arcanis:
Do we have tests with git, outside of the resolver?

bestander:
I think there are a couple e2e tests with git, we have a GitHub repo for
basic e2e tests.

If setting one up is too much a hassle we could merge it without it but
better with it

__tests__/commands/install/integration.js
```diff
@@ -61,6 +61,24 @@ test.concurrent('properly find and save build artifacts', async () => {
...
+  await runInstall({}, 'install-git', async (config, reporter): Promise<void> => {
+    console.log('aaaaa');
+    const lockfile = await Lockfile.fromDirectory(config.cwd);
+
+    expect(await fs.glob('example-yarn-package.git-*', {cwd: `${config.cwd}/offline-mirror`})).toHaveLength(1);
+
+    await fs.unlink(path.join(config.cwd, 'offline-mirror'));
+    await fs.unlink(path.join(config.cwd, 'node_modules'));
+
+    console.log('bbbbb');
```
bestander: nice console.log! :)

PR #3553, opened 2017-06-01
Run install and pack when fetching git dependencies with a prepare script

### LINKED ISSUE #2875: Installing package from github URL doesn't trigger prepublish script
Bug/feature

Installing package from Github URL doesn't run the prepublish script

Yarn should run the `prepublish` script when installing from a Github URL just like it should when installing a local package.

Ref: https://github.com/npm/npm/issues/3055
More specifically: https://github.com/npm/npm/issues/3055#issuecomment-222897703

Volune:
Implement a new behaviour in `GitFetcher` to fix #2875 and match the new behaviour of npm 5 regarding git dependencies:
- If the git repository's package.json does not contain a _prepare_ script, no change of behaviour
- If the git repository's package.json contains a _prepare_ script:
  - clone the repository in a new temporary folder
  - run `install` command in that folder, excluding the _prepublish_ script
  - run `pack` command and use the content of the generated tarball as result of GitFetcher
  - remove the temporary folder

(@zkat please let me know if you think it differs from npm)

----
At the moment, I open the PR mostly for review and comments. The PR still lacks documentation of the new behaviour.

Some implementation details:
- There is a new boolean `config.gitDependency` to disable the _prepublish_ script
- The tarball generated if there is a prepare script uses the `pack` code. That means all the files are in a `package/` folder. If generated from `git archive`, there are no `package/` folder. To differentiate during the tarball extraction, when I generate a tarball from the `pack` command, I add a parameter in the header of the first file.
- The `tarballMirrorPath` filename does not have extension. The `tarballCachePath` filename has extension `tgz`, but the content does not seem gzipped. I did not change these points.
- The `git archive` may be run multiple times for one fetch. I replicated this behaviour, running the `pack` command multiple times. I did so to ensure a better error handling with minimal work. I think both cases can be optimised.

I added only one test. I'm thinking of improving it to ensure the _prepublish_ script is not run.
I'm not sure what other tests are worth adding, while staying in the scope of this PR. (I don't want to fix all git-fetcher issues here).
By the way, if yarn supports a "git+file://" protocol, that would be useful for testing.

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -823,6 +823,13 @@ export function setFlags(commander: Object) {
   commander.option('-T, --save-tilde', 'DEPRECATED');
 }
 
+export async function install(config: Config, reporter: Reporter, flags: Object, lockfile: Lockfile): Promise<void> {
+  await wrapLifecycle(config, flags, async () => {
+    const install = new Install(flags, config, reporter, lockfile);
+    await install.init();
+  });
+}
+
 export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   let lockfile;
   if (flags.lockfile === false) {
@@ -855,10 +862,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     throw new MessageError(reporter.lang('installCommandRenamed', `yarn ${command} ${exampleArgs.join(' ')}`));
   }
 
-  await wrapLifecycle(config, flags, async () => {
-    const install = new Install(flags, config, reporter, lockfile);
-    await install.init();
-  });
+  await install(config, reporter, flags, lockfile);
 }
 
 export async function wrapLifecycle(config: Config, flags: Object, factory: () => Promise<void>): Promise<void> {
@@ -871,7 +875,9 @@ export async function wrapLifecycle(config: Config, flags: Object, factory: () =
   await config.executeLifecycleScript('postinstall');
 
   if (!config.production) {
-    await config.executeLifecycleScript('prepublish');
+    if (!config.disablePrepublish) {
+      await config.executeLifecycleScript('prepublish');
+    }
     await config.executeLifecycleScript('prepare');
   }
 }
--- a/__tests__/fetchers.js
+++ b/__tests__/fetchers.js
@@ -76,6 +76,32 @@ test('GitFetcher.fetch', async () => {
   expect(name).toBe('beeper');
 });
 
+test('GitFetcher.fetch with prepare script', async () => {
+  const dir = await mkdir('git-fetcher-with-prepare');
+  const fetcher = new GitFetcher(
+    dir,
+    {
+      type: 'git',
+      reference: 'https://github.com/Volune/test-js-git-repo',
+      hash: '96e838bcc908ed424666b4b04efe802fd4c8bccd',
+      registry: 'npm',
+    },
+    (await Config.create()),
+  );
+  await fetcher.fetch();
+  const name = (await fs.readJson(path.join(dir, 'package.json'))).name;
+  expect(name).toBe('test-js-git-repo');
+  const dependencyName = (await fs.readJson(path.join(dir, 'dependency-package.json'))).name;
+  expect(dependencyName).toBe('beeper');
+  // The file "prepare.js" is not in "files" list
+  expect(await fs.exists(path.join(dir, 'prepare.js'))).toBe(false);
+  // Check executed lifecycle scripts
+  expect(await fs.exists(path.join(dir, 'generated', 'preinstall'))).toBe(true);
+  expect(await fs.exists(path.join(dir, 'generated', 'install'))).toBe(true);
+  expect(await fs.exists(path.join(dir, 'generated', 'postinstall'))).toBe(true);
+  expect(await fs.exists(path.join(dir, 'generated', 'prepublish'))).toBe(false);
+});
+
 test('TarballFetcher.fetch', async () => {
   const dir = await mkdir('tarball-fetcher');
   const fetcher = new TarballFetcher(
--- a/src/cli/commands/pack.js
+++ b/src/cli/commands/pack.js
@@ -50,7 +50,10 @@ const NEVER_IGNORE = ignoreLinesToRegex([
   '!/+(changes|changelog|history)*',
 ]);
 
-export async function pack(config: Config, dir: string): Promise<stream$Duplex> {
+export async function packTarball(
+  config: Config,
+  {mapHeader}: {mapHeader?: Object => Object} = {},
+): Promise<stream$Duplex> {
   const pkg = await config.readRootManifest();
   const {bundledDependencies, main, files: onlyFiles} = pkg;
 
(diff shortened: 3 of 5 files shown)


zkat:
The algorithm specified seems to be right. Just make sure all the relevant lifecycle scripts are being run in the same order and such.

Volune:
Tests updated, and I think they pass. (The master branch builds are failing, so I don't know how confident I should be about the tests)
Can someone point me to the documentation I need to update?

package.json
```diff
@@ -37,6 +37,7 @@
     "strip-bom": "^3.0.0",
     "tar-fs": "^1.15.1",
     "tar-stream": "^1.5.2",
+    "test-js-git-repo": "https://github.com/Volune/test-js-git-repo",
```
bestander: not on npm?

bestander: Is this dependency needed?

src/cli/commands/install.js
```diff
@@ -842,7 +846,9 @@ export async function wrapLifecycle(config: Config, flags: Object, factory: () =
   await config.executeLifecycleScript('postinstall');
 
   if (!config.production) {
-    await config.executeLifecycleScript('prepublish');
+    if (!config.gitDependency) {
```
bestander: I think it is better to name it more generic `disablePrepublish`

bestander:
Thanks, @Volune, this looks legit and well thought through.

Volune:
I cleaned the code according to comments, and rebased, so tests are passing now. Let me know if anything is missing.

bestander:
Great work, @Volune.
I think this feature is under documented now and many people might want to use it.
Do you think it will be possible for you to write a short blog post here https://github.com/yarnpkg/website/tree/master/_posts?

Volune:
I'll look into it. It will take some time, I'm not used to writing.

sammarks:
@bestander @Volune just checked out the new feature, and I'm wondering what your recommendation is for the following setup given the new functionality:

- I have `package-a` that depends on `package-b` (through a git URL as `package-b` is not published to a registry).
- `package-a` and `package-b` both use Babel (listed in `devDependencies` in both).
- `package-b`'s `prepare` script is: `babel --optional runtime src -d dist`

When installing the dependencies of `package-a`, I get an error message: `babel: command not found`

When I change to the temporary directory where the build is supposed to be happening before packing, it appears the `node_modules/.bin` directory is not created (but the dependencies are installed), which causes the reference to `babel` to be not found (since we can't expect babel to be installed globally).

Here's the output just for reference:

```
[1/4] Resolving packages...
[2/4] Fetching packages...
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[1/2] ⠄ ngrok
[2/2] ⠄ fsevents
[-/2] ⠄ waiting...
[-/2] ⠄ waiting...
warning Error running install script for optional dependency: "/Users/sammarks/Library/Caches/Yarn/v1/.tmp/a339a22b5c259ad43ee07022e7413f6b.9abc356b9eaa1ddf2c84e3bec8f94758bb49eaf4.prepare/node_modules/fsevents: Command failed.\nExit code: 1\nCommand: sh\nArguments: -c node install\nDirectory: /Users/sammarks/Library/Caches/Yarn/v1/.tmp/a339a22b5c259ad43ee07022e7413f6b.9abc356b9eaa1ddf2c84e3bec8f94758bb49eaf4.prepare/node_modules/fsevents\nOutput:\nevents.js:160\n      throw er; // Unhandled 'error' event\n      ^\n\nError: spawn node-pre-gyp ENOENT\n    at exports._errnoException (util.js:1022:11)\n    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)\n    $ babel --optional runtime src -d dist
sh: babel: command not found
info Visit https://yarnpkg.com/en/docs/cli/upgrade for documentation about this command.
```

Could this be related to https://github.com/yarnpkg/yarn/pull/2053?

I thought best to continue the discussion on this PR since it's recent and relevant, but I can create a new issue to house the discussion if you'd like.

Volune:
@sammarks I can reproduce.
An option `binLinks: true` is missing [when installing dependencies](https://github.com/yarnpkg/yarn/blob/5839c36d3f9fd737adbfd7c6e54b2c02840b2a41/src/fetchers/git-fetcher.js#L164)

I will prepare a PR with a fix and improved tests.

@bestander Do you or another contributor with good knowledge of the `binLinks` option know if it can have undesired side effects?

ramasilveyra:
Hello! I just published a tool to create git tags for packages with build steps and monorepos packages https://github.com/ramasilveyra/gitpkg, my intention is to create very lightweight git tags that don't delay the install command and also be able to publish the gitpkg packages to the same repo.

Please if you want let me know what do you think about this approach.

Thanks!

sammarks:
This is pretty cool @ramasilveyra! Looks like a nice alternative until we can nail down this issue.

bestander:
@ramasilveyra that looks exciting.
What can we do to make it work better with Yarn and Lerna?

ramasilveyra:
@sammarks Sadly they are not the same 🙁 the approach choose by npm and yarn is to run the build scripts when you install the git dep, this could increase a lot the installation time. And it doesn't work with lerna git tags.

`gitpkg publish` at this moment does:

1. Read and validate package.json.
2. Run prepublish scripts.
3. Prepare package: npm pack and untar tarball to temp dir.
4. Upload package: create git tag from temp dir and push to resolved gitpkg registry.
5. Run postpublish scripts.

The end result is a git tag with only the files needed, look at this https://github.com/ramasilveyra/public-registry/tree/gitpkg%401.0.0-beta.0-gitpkg

----------------------

@bestander I believe that there is a lot of wins if the functionality of gitpkg gets integrated in yarn, npm and lerna/workspaces.

- On the yarn and npm side:

  - **`publishConfig.registry`:**  [`"publishConfig":{"registry":"http://my-internal-registry.local"}`](https://docs.npmjs.com/misc/registry#i-dont-want-my-package-published-in-the-official-registry-its-private) maybe it would be cool to check if that entry is a git url and if it is the publish command will create the git tag and push it to that repo. 
  - **semver ranges support:** for example that the dep `ramasilveyra/public-registry#gitpkg:^1.0.0-beta.0` (note the `:^` chars) resolves to the git tag `gitpkg@1.0.0-beta.0-gitpkg` on `git@github.com:ramasilveyra/public-registry.git`.

- On the lerna and workspaces:
  - That the bootstrap command creates the symlinks for the git tags deps too.

Is this something for a RFC? Theres is a common place to discuss this ideas to get implemented on npm and yarn?

bestander:
@ramasilveyra, yes an RFC would be the right way to go.
Please consider a multistep approach, there are a lot of things that are not the best in publish experience and we are going to work on revamping it a lot, so big changes may be hard to consider against multiple other changes.

masaeedu:
@Volune Doing `yarn add masaeedu/ava` doesn't seem to result in the `prepare` script being executed with `yarn 1.1.0`. Is this feature perhaps part of a later milestone?

rwjblue:
@masaeedu - I was just looking into a similar issue, in my case it turned out that the logic added in this PR only applies to `git:` references and not `github:` references.  Once I changed my `package.json` to reference it via `git:` the `prepare` was properly being invoked.

Volune:
@masaeedu  @rwjblue 
Yarn has a different behavior for "hosted git repositories"
I started a discussion a while ago regarding this issue on yarnpkg/rfcs#79, but it then happened that I didn't have time to take care of it. The idea is to use the same algorithm to install "hosted git repositories" and "git repositories"
I can't promise I'll have time to work on it soon.

PR #3578, opened 2017-06-05
Fix Appveyor tests

arcanis:


arcanis:
That should do the trick - the problem was that `getPossibleLocation` was splitting the path depending on path.sep, despite the OS accepting both backslashes and forward slashes as valid path (and the new test was using forward slashs).

I really think we should eventually try to normalize our implementation so that Yarn **always** use forward slashes inside the application code, leaving the utils implementation decide on what to do with these paths. There's some early work on this inside the fetchers PR.

src/registries/npm-registry.js
```diff
@@ -110,7 +110,7 @@ export default class NpmRegistry extends Registry {
       [false, path.join(getGlobalPrefix(), filename)],
     ];
 
-    const foldersFromRootToCwd = this.cwd.split(path.sep);
+    const foldersFromRootToCwd = this.cwd.split(/[\\/]/g);
```
BYK: How about doing
```
this.cwd = path.normalize(this.cwd);
const foldersFromRootToCwd = this.cwd.split(path.sep);
```

instead? See https://nodejs.org/api/path.html#path_path_normalize_path

BYK:
Btw. thanks a lot for doing this! :)

src/registries/npm-registry.js
cpojer: There is also the "slashes" module on npm.

arcanis: Normalize would work when using posix path on windows, but not the opposite (on posix, backslashes are valid filename characters, so path.posix doesn't convert them) :(

I'm gonna use `slash`, but I don't understand why it does an ASCII check ... filenames can definitely contain special characters, especially when dealing with user folders (oh how many times I've been burned when Windows put my home folder under `C:\Users\Maël` :().

PR #3584, opened 2017-06-05
Fix: Send auth token only when forced or mutating (fixes #1952)

### LINKED ISSUE #1952: Error: Could not authenticate deployment
A bug

Getting an error when trying to run `yarn install` on an Angular 2 project that works fine with `npm install`

```
yarn install v0.17.6
info No lockfile found.
error An unexpected error occurred: "https://registry.yarnpkg.com/@angular%2fcommon: Could not authenticate deployment".
info If you think this is a bug, please open a bug report with the information provided in "/Users/user/Project/yarn-error.log".
```

- Create a project using angular-cli
- Delete `node_modules`
- Run `yarn install`

To yarn to run successfully and to have all the dependencies installed in `node_modules`

node 7.1.0, installed with nvm
MacOS 10.11.6
yarn 0.17.6

BYK:
**Summary**

Fixes #1952 

Currently, when `this.token` is set, the auth token is always sent,
even with read-only GET and HEAD requests. This seems to be causing
yarn registry to error out. This patch changes the behavior so the
token is only sent with mutating HTTP requests (such as `PUT`) or
when it is forced via `always-auth`.

**Test plan**

TBD. Currently there are no tests for registries and I'm not sure how to start testing them. Any input and direction regarding this would be much appreciated.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -65,7 +65,11 @@ export default class NpmRegistry extends Registry {
       },
       opts.headers,
     );
-    if (this.token || (alwaysAuth && isRequestToRegistry(requestUrl, registry, customHostSuffix))) {
+
+    const useAuth =
+      (opts.method !== 'GET' && opts.method !== 'HEAD' && this.token) ||
+      (alwaysAuth && isRequestToRegistry(requestUrl, registry, customHostSuffix));
+    if (useAuth) {
       const authorization = this.getAuth(packageName || pathname);
       if (authorization) {
         headers.authorization = authorization;


BYK:
Travis and AppVeyor failures seem unrelated.

arcanis:
That's a weird bug ... have you find out the code where npm set the auth header or not? I would have thought that sending the token was required, since otherwise it would mean that private packages are available without token, which I assume is incorrect.

BYK:
>That's a weird bug ... have you find out the code where npm decides to set the auth header or not? I would have thought that sending the token was required, since otherwise it would mean that private packages are available without token, which I assume is incorrect.

Ah, good point. I looked at the docs a lot and couldn't figure it out. That said they might be trying anon first and then using auth on a 400 failure. I'll check their source and share what I find. Even in that case, this PR would need updating.

BYK:
@arcanis - From this test, I think what I assumed was true: https://github.com/npm/npm-registry-client/blob/debec76884db8092c2c7a21ab5b4ed083f8ce2c9/test/fetch-not-authed.js

Seems like unless `alwaysAuth` is set to `true`, they don't send authorization headers for `GET` even they are present.

arcanis:
Ok then. As for the tests, I haven't been able to reproduce this issue so I'm not sure what would be required to implement a test - do you have a repro script? I've tried logging in both inside npm & yarn, but everything worked still fine after that.

BYK:
>Ok then. As for the tests, I haven't been able to reproduce this issue so I'm not sure what would be required to implement a test - do you have a repro script? I've tried logging in both inside npm & yarn, but everything worked still fine after that.

I don't have a repro script yet and I don't think we should merge this without making sure it solves our problem consistently (and hopefully with tests). I'll dig more. Do you have the `auth` line in your `npmrc` file? I think newer versions of NPM stopped using and generating that so that may explain why you were not able to repro.

BYK:
Okay, this PR is unnecessary because the issue is already resolved on master and with v0.26.0

PR #3594, opened 2017-06-07
Use `.every` instead of `.filter().length === 0` pattern

BYK:
**Summary**

When checking lockfile patterns and resolved patterns, we were using `.filter(!expr).length === 0` pattern, I guess due to Node < 4 support in the past. Since we have `.every(expr)` now, this patch uses that and comes with a slight performance boost due to the short-circuiting nature of `every()` on top of readability.

**Test plan**

Existing tests.

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -662,13 +662,11 @@ export class Install {
       this.scripts.getArtifacts(),
     );
 
-    const lockFileHasAllPatterns = patterns.filter(p => !this.lockfile.getLocked(p)).length === 0;
-    const resolverPatternsAreSameAsInLockfile =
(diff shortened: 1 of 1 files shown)


BYK:
@kaylieEB I thought `.every()` is also fail fast: once something returns `false`, it quits and returns `false`. Isn't that the case? 
[
MDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) the following:
>The every method executes the provided callback function once for each element present in the array until it finds one where callback returns a falsy value. If such an element is found, the every method immediately returns false.

kaylieEB:
oh cool, actually didn't know `.every` also short-circuits. Looks good to me!

cpojer:
Nice cleanup!

Issue #3566, opened 2017-06-02
Yarn does not support `yarn run env` as npm does


jasonkarns:
**Do you want to request a *feature* or report a *bug*?** inconsistency with npm

**What is the current behavior?**
One can run `npm run env` in any package and it will print the environment under which package scripts would run. It is in essence as if the script `"env": "printenv"` existed in the package.

Yarn, however, does not do this.

```
$ yarn run env
yarn run v0.24.5
error Command "env" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

$ npm run env

> testdouble@2.1.2 env /Users/jasonkarns/Projects/td/testdouble.js
> env

[... env var dump ...]
```

**If the current behavior is a bug, please provide the steps to reproduce.**
`yarn run env` from any package that doesn't already have an "env" script.

**What is the expected behavior?**
If it is desired to match npm-cli behavior, the expected output would be a `printenv` call from within the context of a package script.

**Please mention your node.js, yarn and operating system version.**
- node 8.0
- npm 4.2 (same with 5.0)
- yarn 0.24.5
- macOS Sierra 10.12.5

arcanis:
TIL. It would be nice to have this feature, I myself often typed `"env": "env"` in my scripts to have the same behavior. Would you be willing to make a PR?

jasonkarns:
@arcanis I don't have the time to PR this. Though as an aside, I would recommend you use `printenv` as the command for your "env" script, because `env` does more than just print the environment.

bestander:
Done

jasonkarns:
This seems buggy. Just tested on 0.27.5 and the `yarn run env` command doesn't include any of the `npm_package_*` env vars that are present from a yarn/npm script.

```
$ npm run env | ag npm_package | wc -l
      88
$ yarn run env | ag npm_package | wc -l
       0
```

Since this has been released, should I open this bug as a separate issue?

PR #3596, opened 2017-06-07
Add `yarn run env` script

kaylieEB:
**Summary**
Per feature request in https://github.com/yarnpkg/yarn/issues/3566, add a script command that prints env variables. The current implementation can also be extended to other shell commands.

**Test plan**
Added unit tests to `run` file

screen shot 2017-06-06 at 11 48 49 pm

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -11,6 +11,10 @@ import map from '../../util/map.js';
 const leven = require('leven');
 const path = require('path');
 
+const shellCommands = {
+  env: 'printenv',
+};
+
 function sanitizedArgs(args: Array<string>): Array<string> {
   const newArgs = [];
   for (let arg of args) {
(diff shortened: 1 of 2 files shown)


src/cli/commands/run.js
```diff
@@ -58,7 +62,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       cmdHints[cmd] = pkgScripts[cmd] || '';
     }
 
-    Object.assign(scripts, pkg.scripts);
+    Object.assign(scripts, pkg.scripts, shellCommands);
```
arcanis: `shellCommands` should probably be put before `pkg.scripts` so that user scripts can override the default mapping

kaylieEB:
@arcanis that's a really good point! Updated based on your feedback, thanks :)

arcanis:
Looks good - it is not compatible with NPM but using a JSON output makes a lot of sense, so let's use this and see if people complain 👍 Thanks !

jasonkarns:
I like the idea of the json output, though I quite prefer npm's convention of printing json output when `--json` flag is present.

However, in practice, my use of this command is almost exclusively paired with a pipe-to-grep, so text or json makes no difference to me, as long as it is one variable per line. :)

Thanks @kaylieEB !

PR #3599, opened 2017-06-07
Fixed failure to set proxy during redirects

atti187:
**Summary**
I have a scenario where a request is being made to an internal company server (no proxy) which redirects to an external server (proxy needed). The consecutive request is also being done without a proxy.

An example is how Microsoft TFS handles npm-feeds, see https://www.visualstudio.com/en-us/docs/package/npm/upstream-sources#caching (Internet requirements). Normally, TFS is installed on-prem (no proxy needed) but will redirect to npmjs.org (proxy needed).

Also - see https://github.com/request/request/issues/2683.

**Test plan**
Connect to a TFS repo behind a corporate firewall, which requires a client proxy to get internet access. Request a package that has never been cached in TFS. Before this change, the request is failing. After the change (testing with a new package that has never been cached before), the request goes through.

### DIFF
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -427,7 +427,7 @@ export default class RequestManager {
 
     req.on('error', onError);
 
-    // make sure request is re-evaluating proxy based on redirect uri
+    // make sure request is re-evaluating proxy based on redirect uri (#2683)
     req.on('redirect', () => {
       delete req.proxy;
     });
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -427,6 +427,11 @@ export default class RequestManager {
 
     req.on('error', onError);
 
+    // make sure request is re-evaluating proxy based on redirect uri
+    req.on('redirect', () => {
+      delete req.proxy;
+    });
+
     const queue = params.queue;
     if (queue) {
       req.on('data', queue.stillActive.bind(queue));


arcanis:
@atti187 Just to get a better understanding of the fix, what exactly happens when deleting `req.proxy`? Won't that cause the redirect request to not use any proxy, since we've just removed this configuration?

atti187:
@arcanis No, on https://github.com/request/request/blob/master/lib/redirect.js#L149 init is being called after the emit of 'redirect' and on https://github.com/request/request/blob/master/request.js#L278 a new proxy is being assigned, based on https://github.com/request/request/blob/master/lib/getProxyFromURI.js#L40.

I guess this is a quick fix for https://github.com/request/request/issues/2683... Not sure if it's by design, or if I'll be able to pull it through.

BYK:
@atti187 - can you add a test case for this?

Also a comment in the code would be great for innocent souls reading it and trying to make sense of it (a link to the request issue with a `// TODO:` comment would be even better). I'd also prefer to get his fixed in upstream request if possible. Do you think you can get some attention on that side since your bug report seems unattended on request repo.

BYK:
>Also a comment in the code would be great for innocent souls reading it and trying to make sense of it (a link to the request issue with a // TODO: comment would be even better).

I just saw the comment above the event handler. I think just making it a `// TODO: Remove this when https://git.io/vQ2Np is resolved` would be good.

BYK:
Hey @atti187 - need any help from us? Would love to get this merged.

atti187:
Yes, please - I haven't had the time to look into any tests. It seems a bit complex to set up proxies etc...

BYK:
>Yes, please - I haven't had the time to look into any tests. It seems a bit complex to set up proxies etc...

Well, you don't need to set up proxies I think. We should be able to simulate this behavior. I'll try to look into this too.

arcanis:

PR #3607, opened 2017-06-08
Color fix

### LINKED ISSUE #2724: Yarn help changes text color in powershell
bug in yarn 0.20.3 (installed through chocolatey)

when running [yarn help] the help is displayed but the textcolor is changed and left to a dark color, so nothing can be read afterwards.

Just run yarn help in powershell (I also have poshgit installed in case that makes a difference)

The text color should be reset to whatever it was before yarn help was called

Node 7.5.0
Yarn 0.20.3

As a work around I can run `[Console]::ResetColor()` in powershell to reset the text

ViceIce:
**Summary**
This PR resets the console style after using bold, because the new windows console does not support the ansi escape codes ^[[21m and ^[[22m. This would fix  #2724 

**Test plan**

Only one test case had to be fixed. :)

### DIFF
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -27,6 +27,11 @@ const tty = require('tty');
 type Row = Array<string>;
 type InquirerResponses<K, T> = {[key: K]: Array<T>};
 
+// fixes bold on windows
+if (process.platform === 'win32' && process.env.TERM && !/^xterm/i.test(process.env.TERM)) {
+  chalk.styles.bold.close += '\u001b[m';
+}
+
 export default class ConsoleReporter extends BaseReporter {
   constructor(opts: Object) {
     super(opts);
--- a/__tests__/reporters/console-reporter.js
+++ b/__tests__/reporters/console-reporter.js
@@ -13,6 +13,7 @@ const stream = require('stream');
 require('chalk').enabled = true;
 require('chalk').supportsColor = true;
 require('chalk').styles.blue.open = '\u001b[34m';
+require('chalk').styles.bold.close = '\u001b[22m';
 
 test('ConsoleReporter.step', async () => {
   expect(


ViceIce:
Test timeouts 😞

rally25rs:
I was going to merge this PR, but was a little unclear on the "Fix prettier on windows." commit.
Seems like a large diff. Is this change actually related to fixing the color output on Win?

ViceIce:
It's not relate to the Bug, but the old prettier script was not working on windows. Command too long error. I found this solution on Facebook flow project.

https://github.com/facebook/react/issues/9500

ViceIce:
Removed prettier commit and opened another PR #3635

src/reporters/console/console-reporter.js
```diff
@@ -27,6 +27,12 @@ const tty = require('tty');
 type Row = Array<string>;
 type InquirerResponses<K, T> = {[key: K]: Array<T>};
 
+// fixes bold on windows
+// $FlowFixMe TERM is not a string?
+if (process.platform === 'win32' && !/^xterm/i.test(process.env.TERM)) {
```
torifat: Does `flow` still complain if you do
```js
if (process.platform === 'win32' && process.env.TERM && !/^xterm/i.test(process.env.TERM)) {
```

And, how about using `startswith` instead of the regex?

ViceIce: I can test it, this was copied from /chalk/chalk/blob/v1.1.1/index.js#L8

startsWith is not case insensitive.

torifat: Yeah, I missed the `i`.

It's better if we can get rid of `$FlowFixMe`.

ViceIce: Done 😄

src/reporters/console/console-reporter.js
```diff
@@ -27,6 +27,11 @@ const tty = require('tty');
 type Row = Array<string>;
 type InquirerResponses<K, T> = {[key: K]: Array<T>};
 
+// fixes bold on windows
+if (process.platform === 'win32' && process.env.TERM && !/^xterm/i.test(process.env.TERM)) {
+  chalk.styles.bold.close += '\u001b[m';
```
Daniel15: I wonder if this should be submitted upstream to the `chalk` project?

ViceIce: See chalk/chalk#145

Daniel15: Cool, sounds like we can remove this hack in the future, then.

PR #3629, opened 2017-06-13
Enhancement: slightly optimize how we update status from running scripts

BYK:
**Summary**

This patch does two minor optimizations:

1. Do not even have a handler for data updates if we don't have a
   spinner. This avoids noop calls to a handler that just returns
   if we don't have a spinner.
2. Optimize getting the last available line from the buffer.
   Instead of using `.split()` and `.pop()` it uses `.lastIndexOf()`
   and `.substr()` to locate and extract the last line.

**Test plan**

Existing tests, behavior should not change.

### DIFF
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -173,20 +173,26 @@ export async function executeLifecycleScript(
     conf.windowsVerbatimArguments = true;
   }
 
-  const stdout = await child.spawn(sh, [shFlag, cmd], {cwd, env, stdio, ...conf}, data => {
-    if (spinner) {
-      const line = data
+  let updateProgress;
+  if (spinner) {
+    const ticker = spinner.tick.bind(spinner);
+    updateProgress = data => {
+      const dataStr = data
         .toString() // turn buffer into string
-        .trim() // trim whitespace
-        .split('\n') // split into lines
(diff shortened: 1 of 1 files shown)


src/util/execute-lifecycle-script.js
```diff
@@ -173,20 +173,26 @@ export async function executeLifecycleScript(
     conf.windowsVerbatimArguments = true;
   }
 
-  const stdout = await child.spawn(sh, [shFlag, cmd], {cwd, env, stdio, ...conf}, data => {
-    if (spinner) {
-      const line = data
+  let updateProgress;
+  if (spinner) {
+    const ticker = spinner.tick.bind(spinner);
```
BYK: I know this looks weird but `flow` demands it :(

arcanis: Why use `bind`?

arcanis: eh, nevermind, I just saw your comment 😄

BYK: May be I should put this comment in the code?

arcanis: That being said, I think you could just use an invariant inside the `updateProgress` callback to tell Flow that `spinner` *does* exist.

PR #3633, opened 2017-06-14
Fix: prevent Git from showing a password prompt

### LINKED ISSUE #764: yarn install hangs during "Fetching packages..."

BYK:
**Summary**

Fixes #764.

Turns out when fetching packages from remote sources via Git,
it may ask for a username or password and get stuck there since
we run it in the background. This patch passes the following env
variables to disable any prompts:

1. `$GIT_ASKPASS=""`
   This is the command Git runs to get username and pass. Setting
   it to an empty string disables it and git exits with an error.
2. `$GIT_TERMINAL_PROMPT=0`
    This is new in Git 2.3 and it prevents git from even trying to
    ask for a password. Surpasses `$GIT_ASKPASS` above.
3. `$GIT_SSH_COMMAND="ssh -oBatchMode=yes"`
    This is also new in Git 2.3 and it makes git make the call to
    ssh via this command. `BatchMode=yes` option tells ssh to not
    do anything interactive (because batch/script mode) like
    password prompts in case publickey auth fails etc.

**Test plan**

Do `yarn add git+https://github.com/Napsty/privrepo.git#1.0.2`. `yarn`
hangs before this patch and fails with a "Refusing to download" error
after the patch.

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -31,6 +31,9 @@ const supportsArchiveCache: {[key: string]: boolean} = map({
   'github.com': false, // not support, doubt they will ever support it
 });
 
+// Suppress any password prompts since we run these in the background
+const env = {GIT_ASKPASS: '', GIT_TERMINAL_PROMPT: 0, GIT_SSH_COMMAND: 'ssh -oBatchMode=yes'};
+
 // This regex is designed to match output from git of the style:
 //   ebeb6eafceb61dd08441ffe086c77eb472842494  refs/tags/v0.21.0
 // and extract the hash and tag name as capture groups
@@ -87,6 +90,10 @@ export default class Git {
     };
   }
 
+  static git(args: Array<string>, opts?: child_process$spawnOpts = {}): Promise<string> {
+    return child.spawn('git', args, {...opts, env});
+  }
+
   /**
    * Check if the host specified in the input `gitUrl` has archive capability.
    */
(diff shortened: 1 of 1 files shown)


BYK:
Three notes about this PR:

1. Not sure if this covers the entire #764 but I think it covers a significant portion
2. The error messages are not really descriptive, almost deceptive. Planning to address that in a follow-up but I can change them here if requested.
3. I'm not very happy about the manual test plan. Any directions to automate this without `git` and network dependency are welcome.

src/util/git.js
```diff
@@ -87,6 +90,10 @@ export default class Git {
     };
   }
 
+  static git(args: Array<string>, opts?: child_process$spawnOpts = {}): Promise<string> {
```
BYK: May be I should rename this to `spawn`?

cpojer: Yes please. "Git.git" is weird :D

kirs:
❤️

kisamoto:
If I understand this correctly this disables the use of passphrase protected SSH keys as now there is no opportunity to enter the passphrase for a private SSH key or password for HTTPS authentication?

BYK:
Sounds about right. Both of those scenarios can be addressed via SSH agent and Git Credentials Manager respectively though.

bestander:
@BYK, this seems to have broken ability to use git dependencies behind a proxy, I raised an internal task with details t19705486

BYK:
@bestander created #3742 for public tracker.

PR #3635, opened 2017-06-14
Fix prettier on windows.

ViceIce:
**Summary**

This fix will allow to run prettier on windows. Otherwise it will fail with a commandline to long error.

**Test plan**

Try to run prettier on windows with and without this fix. 😄


arcanis:
Thanks ! :)

PR #3637, opened 2017-06-14
Make `yarn run` only add global path to env if it is not present already

### LINKED ISSUE #3636: `yarn run` changes `PATH` and breaks `rbenv`
bug

`yarn run` changes the `PATH` variable. In my case, in addition to adding some `node_module`-paths it also prepends `/usr/bin` to the existing PATH even though it is already present.

For example, my normal PATH is `/usr/local/rbenv/shims:/usr/local/rbenv/bin:/usr/local/bin:/usr/bin:/bin`. Having `/usr/local/rbenv/shims` in front of `/usr/bin` makes sure that the ruby executable in the first path is executed.

When doing a `yarn run` the process' PATH is changed to:

> /path/releases/20170613205524/node_modules/.bin:/path/.config/yarn/link/node_modules/.bin:/path/releases/20170613205524/node_modules/.bin:/path/.config/yarn/link/node_modules/.bin:**/usr/bin**:/usr/lib/node_modules/npm/bin/node-gyp-bin:/usr/bin/node_modules/npm/bin/node-gyp-bin:**/usr/local/rbenv/versions/2.4.1/bin:/usr/local/rbenv/libexec:/usr/local/rbenv/plugins/ruby-build/bin:/usr/local/rbenv/plugins/rbenv-vars/bin:/usr/local/rbenv/shims**:/usr/local/rbenv/bin:/usr/local/bin:/usr/bin:/bin

There are some duplicates added etc., but the culprit is that the normal PATH is also prefixed with `/usr/bin`. Now `/usr/bin/` comes before the PATH set up by rbenv. Calling `ruby` from a yarn bin now calls the wrong ruby version, because the PATH changes made by rbenv come later.

I traced it down to https://github.com/yarnpkg/yarn/blob/8ede3c41cfb389ee408ae1094503649dc11b11cc/src/util/execute-lifecycle-script.js#L142 which always adds the "global bin" folder. Commenting out this line fixes my problem.

Maybe this should be made conditional to only add it if it is not present already, so if it is already there, the order does not change?

For reference, I stumbled upon this while working with rails/webpacker, see the issue there: https://github.com/rails/webpacker/issues/499

It should not change order of entries already present in PATH (or not add such global path at all, but I assume it is there for a reason ...).

node v6.11.0
yarn 0.24.6
ubuntu 16.04

doits:
**Summary**

If the global path is present already, prepending it should not be necessary
and might change the order of path entries. This can break
applications that rely on the correct order of path entries, like
rbenv.

fixes #3636

**Test plan**

Should not be needed here, really small code change.

### DIFF
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -138,9 +138,12 @@ export async function executeLifecycleScript(
     path.join(path.dirname(process.execPath), '..', 'lib', 'node_modules', 'npm', 'bin', 'node-gyp-bin'),
   );
 
-  // Add global bin folder, as some packages depend on a globally-installed
-  // version of node-gyp.
-  pathParts.unshift(getGlobalBinFolder(config, {}));
+  // Add global bin folder if it is not present already, as some packages depend
+  // on a globally-installed version of node-gyp.
+  const globalBin = getGlobalBinFolder(config, {});
+  if (!pathParts.includes(globalBin)) {
+    pathParts.unshift(globalBin);
+  }
 
   // add .bin folders to PATH
   for (const registry of Object.keys(registries)) {


bestander:
Thanks, @doits

PR #3640, opened 2017-06-15
Add missing await in npm-resolver.js

pweaver:
**Summary**
In the process for figuring out why --prefer-offline is causing a `Couldn't find package "expose-loader" on the "npm" registry.` for us in some cases I noticed a missing await which is probably causing the null check below to be useless.

**Test plan**


arcanis:
Good spot - without the await, the condition that follows will always be true 👍

PR #3646, opened 2017-06-16
Fix: Use a dedicated sink for tar reading instead of a transform stream

### LINKED ISSUE #3011: Publish without version prompt
minor feature

`yarn publish` is prompting for a new version number

For a release and deployment bash script, I bump the new version internally and would like to pass on the new version to yarn with an argument like `yarn publish --version new.version.number`

And better, `yarn publish auto`, would parse the new version number in package.json. Sometimes I bump it there too.

Think that's irrelevant?

BYK:
**Summary**

Potentially fixes #3011 (specifically https://git.io/vHAzW).

Turns out our `ConcatStream` implementation was only used to read
from `tar-fs` when creating a package to upload and put that stream
into a memory buffer. Since `ConcatStream` was implemented as a
`stream.Transform` and nothing was reading back from it, it had the
potential to just hang there until something reads from it. This
patch replaces that with [a small script][1].

[1]: http://www.geekpeak.de/images/produkte/i22/22-go-away-or-i-will-replace-you-de.jpg

**Test plan**

Removed existing tests for the old module. Rely on other existing
tests for the replacement code.


tomsonpl:
@BYK hey, this issue is still happening. Do you have any plans to fix this? 

i am receiving 9.154668655 Request "https://registry.npmjs.org/xxxxxx" finished with status code 200. 

And gets stuck. 

yarn v 1.22.10

PR #3648, opened 2017-06-16
Don't use .includes for compatibility with Node 4

Volune:
**Summary**

Several tests fail on Node 4 because of the use of `Array.prototype.includes`

**Test plan**

Existing tests on Node 4

### DIFF
--- a/src/cli/commands/check.js
+++ b/src/cli/commands/check.js
@@ -203,7 +203,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     const normalized = path.normalize(relative).split(path.sep);
     return normalized.filter(p => p !== 'node_modules').reduce((result, part) => {
       const length = result.length;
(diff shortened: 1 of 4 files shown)


bestander:
So did we stop transpiling ES6 correctly ?
I think we should fix the build tool to include proper shim rather than limit people in language.

Volune:
I would agree, but that's not the solution that was chosen in #3592 and #3654
Should we add a dependency to `babel-polyfill` ?

bestander:
I think those PRs we just agreed on a quick fix.
People keep using ES6 so it breaks all the time.

Does babel-polyfil fix the problem?

bestander:
Opened an issue https://github.com/yarnpkg/yarn/issues/3693 to track the problem.

bestander:
Thanks for raising this, @Volune.

PR #3652, opened 2017-06-16
[WIP] update jest and flow

voxsim:
**Summary**
In this pull request I try to update jest and flow. Now they are failing 30 tests, but flow check passes (I fixed almost 100 failing checks).

Pain points:
- From version v0.46.0 Flow does a strict check of Function Call Arity (https://flow.org/blog/2017/05/07/Strict-Function-Call-Arity/)
- Flow complains when I add a custom matcher in Jest, now I fixed adding the function in the flow definition of jest (I opened an issue https://github.com/flowtype/flow-typed/issues/948)
- the semver definition in flow-typed miss some definitions (https://github.com/flowtype/flow-typed/blob/master/cli/flow-typed/npm/semver_v5.1.x.js). Maybe I will open a pull-request in that repo too.

**Test plan**
The same tests in place, maybe I will add other tests for some edge cases.

**Questions**
- Is there any way to upgrade one minor at time with `yarn upgrade-interactive`? I think It could be an interesting feature..

Feedbacks are welcome :)

EDIT: I tried to update flow because when I try to refactor `package-request` I found that there are some flow errors in integrity-checker, I think the old version of flow has some problems (obviously :D).


arcanis:
:O I'm curious, what makes those tests failing (at least those which fail not because of Flow)?

voxsim:
I write some bash commands to fix *fastly* some ripetitve errors, I think in some parts I put some bugs D: I think I will dig further in the next week, in the meantime if someone find something erronous please let me know!

voxsim:
@arcanis @bestander it seems to work. The only BIG issue is that:

```
****.includes is not a function
```

It seems that upgrading jest, something related to babel is upgraded as well.. and the polyfill `Array.includes` stops to work or maybe it is not included correctly.
Now I am doing some experiments to understand why it doesn't work. My last chance is to revert jest and I hope everything will work like before.
I also noticed that we are using many deprecated babel packages, maybe someone who knows Babel better can help me in another pull request O.o

voxsim:
ok I finally found why includes does not work with jest and babel-jest https://github.com/facebook/jest/pull/2755.

Also https://github.com/avajs/ava/issues/263 was a very interesting reading :)

Now I have the big picture, tomorrow I will open three pull requests to update flow, jest and babel separately :D

voxsim:
ah TL; DR we should not use includes function in our tests O.o

PR #3655, opened 2017-06-16
Update: Save a call when checking for links

BYK:
**Summary**

`fs.readlink` already makes an internal call to `fs.lstat` so rely
on the `EINVAL` error it throws instead of checking if the path is
a link or not via `lstat`.

**Test plan**

Existing tests.


arcanis:
Isn't the previous code more sound, semantically speaking? I feel like this change might make the code harder to understand (highlighted by the need to add a comment to explain why we might end up with an exception thrown).

BYK:
@arcanis - Yeah I agree. I also don't see much value in this "fix" so I'm okay with closing if you agree with that.

arcanis:
👍 let's close it then

PR #3656, opened 2017-06-17
Fixes 3598

### LINKED ISSUE #3598: Workspaces creates self-references
I tried workspaces on Jest and the symlinks it created referred to itself:

screen shot 2017-06-07 at 1 52 22 pm

I could repo this two out of three times.

bestander:
**Summary**

Fixes #3598 

Fixed a very naive condition that matched workspaces `jest-cli` as a subpath of workspace `jest`.

**Test plan**

Added test + tested that workspaces install correctly for jest.

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -179,7 +179,7 @@ export default class PackageLinker {
 
       // fs copy can't copy through a symlink, so we replace those with real paths to workpsaces
       for (const [symlink, realpath] of symlinkPaths.entries()) {
-        if (dest !== symlink && dest.indexOf(symlink) === 0) {
+        if (dest.indexOf(symlink + path.sep) === 0) {
           dest = dest.replace(symlink, realpath);
         }
       }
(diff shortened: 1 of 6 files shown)


src/package-linker.js
```diff
@@ -179,7 +179,7 @@ export default class PackageLinker {
 
       // fs copy can't copy through a symlink, so we replace those with real paths to workpsaces
       for (const [symlink, realpath] of symlinkPaths.entries()) {
-        if (dest !== symlink && dest.indexOf(symlink) === 0) {
+        if (dest.indexOf(symlink + path.sep) === 0) {
```
BYK: Why drop the symlink check?

bestander: I think the code was misleading, it was not a symlink check.

What I am trying to do here is to replace paths
`node_modules/jest/node_modules/left-pad` with `packages/jest/node_modules/left-pad` because the former is a symlink and we need a real path.

With clause `dest !== symlink` I just made sure I don't change the path of symlink itself.

BYK: I think we can us a comment inside the code to explain the aim of this operation for future generations :)

bestander: I have a comment a line above but I'll suggest a better one

Issue #3506, opened 2017-05-25
Yarn install --pure-lockfile deleting artifact folders for dependencies that already existed and were not reinstalled


collinsauve:
**What is the current behavior?**

I have discovered some cases where some dependency artifact folders are deleted when running `yarn install --pure-lockfile`.  In particular this affects [node-sass](https://www.npmjs.com/package/node-sass) for my day-job project if I have deleted some *other* `node_modules/package-name` folder.  I have created a small repo that reproduces this problem.

**If the current behavior is a bug, please provide the steps to reproduce.**

Using this repo and the `test.sh` in there to reproduce: https://github.com/collinsauve/yarn-test-consumer.

```
git clone git@github.com:collinsauve/yarn-test-consumer.git
cd yarn-test-consumer
./test.sh
```

This uses a dependency ([yarn-test-package](https://github.com/collinsauve/yarn-test-package))that creates `preinstall`, `install`, and `postinstall` folders when those steps are run.  It also has a dependency on `promisfy`, just as a sample other dependency so that we can delete it from node_modules.

The tail of the output will be the following:

```
After initial install, these folders exist:
./node_modules/yarn-test-package/ ./node_modules/yarn-test-package/install ./node_modules/yarn-test-package/postinstall ./node_modules/yarn-test-package/preinstall
After second install, these folders exist:
./node_modules/yarn-test-package/
```

**What is the expected behavior?**

Tail of the output *should* be the following:

```
After initial install, these folders exist:
./node_modules/yarn-test-package/ ./node_modules/yarn-test-package/install ./node_modules/yarn-test-package/postinstall ./node_modules/yarn-test-package/preinstall
After second install, these folders exist:
./node_modules/yarn-test-package/ ./node_modules/yarn-test-package/install ./node_modules/yarn-test-package/postinstall ./node_modules/yarn-test-package/preinstall
```

**Please mention your node.js, yarn and operating system version.**

Node: v7.7.4
Yarn: v0.24.6
OS: Windows 10 v1607 (OS Build 14393.447)
GNU bash, version 4.3.46(2)-release (x86_64-pc-msys)

bestander:
Thanks for a good repro step

mridgway:
I have found that just running `yarn install --pure-lockfile` twice will remove artifacts:

```
yarn init -y
yarn add grpc
rm -rf node_modules/
yarn install --pure-lockfile
ls node_modules/grpc/src/node/
yarn install --pure-lockfile
[...]

PR #3666, opened 2017-06-19
Fix for #3506 - Use setFresh method for determining fresh install 

kaylieEB:
**Summary**
This is a fix for https://github.com/yarnpkg/yarn/issues/3506
The root cause of the issue was that after an initial installation method that didn't generate a lockfile, e.g.`--pure-lockfile` or `--no-lockfile`, the second install skipped packages w/pre and post install scripts because [this line](https://github.com/yarnpkg/yarn/blob/master/src/package-install-scripts.js#L137-L139) returned false, even though fresh was being correctly passed [here](https://github.com/yarnpkg/yarn/blob/master/src/package-resolver.js#L455). It turns out `fresh` attribute wasn't being hydrated correctly, so this diff fixes that.

**Test plan**
Existing tests, and tested that the pre and post-install scripts get executed correctly after `yarn --pure-lockfile` command. Not sure how else to test, since it's just setting a property on the object.

screen shot 2017-06-18 at 6 54 59 pm

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -293,6 +293,7 @@ export default class PackageRequest {
     const ref = new PackageReference(this, info, remote);
     ref.addPattern(this.pattern, info);
     ref.addOptional(this.optional);
+    ref.setFresh(fresh);
     info._reference = ref;
     info._remote = remote;
 


arcanis:
👍

PR #3667, opened 2017-06-19
Update: Make tests great again on Travis CI macOS

BYK:
**Summary**

macOS tests on Travis CI keep timing out and they run very slowly. This patch attempts to solve this by doing the following:

- Use ramfs for `__tests__` folder and for `$TMPDIR` for faster file system
- Remove unnecessary `brew update` and `brew install yarn` commands
- Remove unnecessary `--max-workers` argument since all CI VMs have only 2 cores
- Clean up tmp folders per test after they are done
- Fix `link/unlink` test's race condition and previous test case reliance
- Fix `request-manager` timeout tests to be more reliable and finish in under our normal timeout
- Use real timers by default and add necessary `useFakeTimers` calls
- Increase heap size for Linux and OS X to 4 GB since OS X builds were crashing due to limited heap space
- Removes retries from Travis CI

**Test plan**

All tests on all platforms should pass and pass in about 20 minutes max.

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -142,5 +142,7 @@ export async function run<T, R>(
     }
   } catch (err) {
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
+  } finally {
+    await fs.unlink(cwd);
   }
 }
--- a/__tests__/commands/unlink.js
+++ b/__tests__/commands/unlink.js
@@ -8,6 +8,8 @@ import type {CLIFunctionReturn} from '../../src/types.js';
 import mkdir from './../_temp.js';
 import * as fs from '../../src/util/fs.js';
 
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
+
 const path = require('path');
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'link');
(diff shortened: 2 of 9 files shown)


BYK:
>All checks have passed
>3 successful checks

Success.

BYK:
This PR is ready for review.

__tests__/commands/_helpers.js
```diff
@@ -142,5 +142,7 @@ export async function run<T, R>(
     }
   } catch (err) {
     throw new Error(`${err && err.stack} \nConsole output:\n ${out}`);
+  } finally {
+    await fs.unlink(cwd);
```
bestander: Oh, that is good!
Tests were eating my hard drive on breakfast

__tests__/commands/unlink.js
```diff
@@ -25,7 +27,7 @@ const runUnlink = buildRun.bind(
   ConsoleReporter,
   fixturesLoc,
   (args, flags, config, reporter): CLIFunctionReturn => {
-    return unlink(config, reporter, flags, args);
+    return link(config, reporter, flags, args).then(unlink.bind(null, config, reporter, flags, args));
```
bestander: .then ? :)
what about await?

bestander:
Great job, @BYK, mega impact!

__tests__/commands/unlink.js
BYK: Right, I should fix this :)

BYK: https://github.com/yarnpkg/yarn/pull/3763

Issue #3490, opened 2017-05-24
package-resolver should not optimize yarn.lock by default


bestander:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**

If yarn.lock has this entry

```
babel-runtime@^6.0, babel-runtime@^6.0.0, babel-runtime@^6.11.6, babel-runtime@^6.9.0, babel-runtime@^6.9.1, babel-runtime@^6.9.2:
  version "6.18.0"
  resolved "https://registry.yarnpkg.com/babel-runtime/-/babel-runtime-6.18.0.tgz#0f4177ffd98492ef13b9f823e9994a02584c9078"
  dependencies:
    core-js "^2.4.0"
    regenerator-runtime "^0.9.5"

babel-runtime@^6.22.0:
  version "6.23.0"
  resolved "https://registry.yarnpkg.com/babel-runtime/-/babel-runtime-6.23.0.tgz#0a9489f144de70efb3ce4300accdb329e2fc543b"
  dependencies:
    core-js "^2.4.0"
    regenerator-runtime "^0.10.0"
```

Yarn would try to optimize it and use the highest compatible resolution everywhere, specifically after yarn install the lockfile will be optimized into

```
babel-runtime@^6.0, babel-runtime@^6.0.0, babel-runtime@^6.11.6, babel-runtime@^6.9.0, babel-runtime@^6.9.1, babel-runtime@^6.9.2, babel-runtime@^6.22.0:
  version "6.23.0"
  resolved "https://registry.yarnpkg.com/babel-runtime/-/babel-runtime-6.23.0.tgz#0a9489f144de70efb3ce4300accdb329e2fc543b"
  dependencies:
    core-js "^2.4.0"
    regenerator-runtime "^0.10.0"
```

And the packages will be installed 6.23 everywhere.

**What is the expected behavior?**

Yarn should respect yarn.lock that was generated earlier to achieve installation determinism.
Otherwise people will get unexpected changes to installation tree.

**Please mention your node.js, yarn and operating system version.**
Yarn 0.26

Issue #3489, opened 2017-05-24
package resolver is sensitive to order


bestander:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

From https://github.com/yarnpkg/yarn/pull/3477#issuecomment-303537660:

With 2 patterns, a@1.0.1 (A) and a@^1.0.0 (B), current version 1.0.2, the following can happen:
Order 1

A resolves, adds version 1.0.1
B resolves, finds matching version 1.0.1, delays
final resolve for B, to version 1.0.1
Outcome: a@1.0.1 -> 1.0.1, a@^1.0.0 -> 1.0.1
Order 2

B resolves, adds version 1.0.2
A resolves, adds version 1.0.1
Outcome: a@1.0.1 -> 1.0.1, a@^1.0.0 -> 1.0.2
The committed code will still mitigate problems, but it won't completely solve them...

I'm afraid a better algorithm is needed to get fully deterministic resolves.

**Please mention your node.js, yarn and operating system version.**

Yarn 0.26

bestander:
@blexrob, you've done a great fix for pattern resolver in https://github.com/yarnpkg/yarn/pull/3477, would you have time to solve this one?

bestander:
On top of it, I think we should disable this lockfile optimization if `--pure-lockfile` is passed otherwise does it mean that we may have different trees resolved with the same lockfile?

Another idea to discuss, should we make this dependency optimization an opt-in feature?

bestander:
Related to this https://github.com/yarnpkg/yarn/issues/3313

bestander:
Raised this one https://github.com/yarnpkg/yarn/issues/3490

hmarr:
I think I've hit this one too - I've found a case where upgrading a package then running a `yarn install` results in a changed lockfile (even with yarn 0.26).

I've put a script together that reproduces the issue here https://github.com/hmarr/yarn-lock-inconsistency

bestander:
Thanks for repro, @hmarr.

blexrob:
This is a first try for a deterministic algorithm. No early outs based on earlier registry resolves,
so resolving order won't influence the final outcome.

```
  bestversion = map of pattern->version
  resolvedversion = map of pattern->version

  # determine the best matching version for pattern, recurse through dependencies of those versions
  # note - no early out if a matching older version was already found! This will cause more lookups...
  recurse, start with root patterns
[...]

Issue #3364, opened 2017-05-10
Bug: how to update package.json on yarn add


aseem2625:
Yarn version v0.21.3

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
`yarn add <packagename> -D` isn't adding the package to package.json

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Please mention your node.js, yarn and operating system version.**
node version v7.7.3
OSX 10.12.1

aseem2625:
Not able to reproduce. Closing

PR #3563, opened 2017-06-02
[WIP] Fix 3023

bestander:
Fixes #3023 

**Summary**

The idea is to do resolutions optimization after the resolution tree is built.
Currently resolution process is done in parallel and depending on a run we may end up with a different resolution tree.

PR #3673, opened 2017-06-19
Fixes resolution concurrency problem

### LINKED ISSUE #3023: get different version dependencies between before there is a yarn.lock file and after.
bug

I try to migrate to yarn from npm. So I test the yarn lock with such a package.json file.
``` javascript
{
  "name": "yarn0test",
  "version": "1.0.0",
  "description": "yarn test",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "dependencies": {
    "uglify-js": "2.5.0",
    "webpack": "1.14.0",
    "webpack-parallel-uglify-plugin": "0.2.0"
  },
  "author": "h",
  "license": "MIT"
}

``` 
- when I first run `yarn` to install dependencies, it generates a `yarn.lock` file. 
For uglify-js part:
``` 
uglify-js@2.5.0:
  version "2.5.0"
  resolved "http://r.cnpmjs.org/uglify-js/download/uglify-js-2.5.0.tgz#4ab5d65a4730ecb7a4fb62d3f499e2054d98fba1"
  dependencies:
    async "~0.2.6"
    source-map "~0.5.1"
    uglify-to-browserify "~1.0.0"
    yargs "~3.5.4"

uglify-js@^2.6.2:
  version "2.8.18"
  resolved "http://r.cnpmjs.org/uglify-js/download/uglify-js-2.8.18.tgz#925d14bae48ab62d1883b41afe6e2261662adb8e"
  dependencies:
    source-map "~0.5.1"
    yargs "~3.10.0"
  optionalDependencies:
    uglify-to-browserify "~1.0.0"

uglify-js@~2.7.3:
  version "2.7.5"
  resolved "http://r.cnpmjs.org/uglify-js/download/uglify-js-2.7.5.tgz#4612c0c7baaee2ba7c487de4904ae122079f2ca8"
  dependencies:
    async "~0.2.6"
    source-map "~0.5.1"
    uglify-to-browserify "~1.0.0"
    yargs "~3.10.0"
```
and the node_modules is : 
`/node_modules/uglify-js`  version: 2.5.0
`/node_modules/webpack/node_modules/uglify-js` version 2.7.5
`/node_modules/webpack-parallel-uglify-plugin/node_modules/uglify-js` version **2.8.18**

- then I remove the node_modules directory and rerun `yarn` with the yarn.lock file.
I get the node_modules like this:
`/node_modules/uglify-js` version 2.5.0
`/node_modules/webpack/node_modules/uglify-js` version 2.7.5
`/node_modules/webpack-parallel-uglify-plugin/node_modules/uglify-js` version **2.7.5**

As you see , it get different uglify-js version for webpack-parallel-uglify-plugin.

1, run yarn the the package.json above
2, remove node_modules dir
3, run yarn

It should get the same version before there is a yarn.lock file and after.

node version:  4.4.7
windows 10
yarn version: 0.21.3

arcanis:
Should fix #3023 for sure, possibly #3364, #3490, and #3489 as well.

Two notes:

- I thought this would make perfs worse by a large factor, but actually it barely had any impact at all, **even when running without lockfile**. On a project requiring react-native, react, react-dom and webpack, over a hundred tries, it went from Avg:11s to Avg:12s overall.

- I tried to write an integration test, but unfortunately couldn't automatically reproduce from inside the test environment #3023 - for some reason I always ended up getting 2.7.5 instead of the bogus 2.8.0 :( However, I was able to systematically reproduce it manually, so I'm confident that this PR should fix this issue.

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -344,7 +344,10 @@ export default class PackageRequest {
       }
     }
 
-    await Promise.all(promises);
+    for (const promise of promises) {
+      await promise;
+    }
+
     ref.addDependencies(deps);
 
     // Now that we have all dependencies, it's safe to propagate optional
--- a/src/package-resolver.js
+++ b/src/package-resolver.js
@@ -471,7 +471,10 @@ export default class PackageResolver {
     this.frozen = Boolean(isFrozen);
     this.workspaceLayout = workspaceLayout;
     const activity = (this.activity = this.reporter.activity());
-    await Promise.all(deps.map((req): Promise<void> => this.find(req)));
+
+    for (const req of deps) {
+      await this.find(req);
+    }
 
     // all required package versions have been discovered, so now packages that
     // resolved to existing versions can be resolved to their best available version


bestander:
That is bold :)

arcanis:
That makes me think - maybe it could be interesting to have "perf regression tests", to make sure that we don't unknowingly introduce big regressions.

bestander:
Definitely a good idea to have some performance regression tracking.

@trueadm, do you track performance regressions on OSS React with some framework?

arcanis:
On Yoga we just run a [script with some stats](https://travis-ci.org/facebook/yoga/jobs/244505833#L1294). Maybe we could do this kind of thing with a simple bash script?

bestander:
@arcanis, this change removes parallelism from top level resolutions, but parallelism still exists for second and deeper layer of dependencies https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L348.

You'll probably want to make them in sequence as well.
In that case you'll probably notice the difference especially for large projects with 100s of dependencies to resolve but it may depend on your network latency.

Can you add data to Test Plan even for the manual tests you perform?
For a true test you'd want to add an artificial network latency to your system.

bestander:
I've had this test for #3023 https://github.com/yarnpkg/yarn/pull/3563/files#diff-c2dc3bd2d0bdc7b8d87c36ed3039c2d4R300

arcanis:
> You'll probably want to make them in sequence as well.

Right!

> In that case I expect to see an increased resolution time especially for large projects with 100s of dependencies to resolve but it may depend on your network latency.

I'd expect it too, but apparently there's little change. It even seems *faster*, which I don't completely understand:

```
❯ [mael-mbp?] /tmp/foolol ❯ python -m timeit -v -n 1 -r 10 -s 'import os' 'os.system("(export PATH=~/yarn/bin:$PATH; rm -rf node_modules yarn.lock && yarn) >& /dev/null")'

raw times: 56.8 27.3 24.6 26.1 24.7 26.6 25.5 25.4 57.7 26.2
1 loops, best of 10: 24.6 sec per loop
```

```
❯ [mael-mbp?] /tmp/foolol ❯ python -m timeit -v -n 1 -r 10 -s 'import os' 'os.system("(export PATH=~/not-yarn/bin:$PATH; rm -rf node_modules yarn.lock && yarn) >& /dev/null")'

raw times: 28.9 30.5 29.8 32.9 33.9 30.4 32.6 27.5 30.8 30.5
1 loops, best of 10: 27.5 sec per loop
```

bestander:
That looks all right when you are in corporate network, what if you are on a 3G grade network with significant latency, 3G for example?
You can test it with Network Link Conditioner http://nshipster.com/network-link-conditioner/ on mac.

I think we should do it a bit smarter and go sequential only of there is a lockfile already with all patterns resolved.
Otherwise we can allow some randomness on first-come-first-serve basis although with smarter post processing we could eliminate it too.

arcanis:
The following numbers are from a custom build with fetching / linking disabled. We can see there's some slowdown on very slow networks (+- 4s in average, +17%), but it looks quite small, and probably insignificant compared to the time required to run the fetching + linking steps.

I'd like to avoid complexifying the code by adding another parameter "is the lockfile new?", that would only be a workaround that would eventually become legacy. Ideally, I think the tree should be resolved asynchronously without any regard for any kind of context-sensitive optimizations (optimization passes being then ran on this tree), and I plan to address that in a followup PR. It should fix your concerns. WDYT?

**3G sync:**

```
raw times: 27.2 27.5 27.6 27.3 27.6 28.3 28 28.3 27.5 27
1 loops, best of 10: 27 sec per loop
```

**3G async:**

```
raw times: 23.4 24.1 23.6 23.5 23.9 23.5 23.3 23.7 23.5 48.6
1 loops, best of 10: 23.3 sec per loop
```

trueadm:
@bestander in regards to how we track performance regressions on React, we don't have a formal process in place yet, more an ad-hoc one. We've implemented some benchmarking tools that allow us to do this using Lighthouse: https://github.com/facebook/react/tree/master/scripts/bench

arcanis:
I'll merge this PR to fix the concurrency issues that make the install unstable, and try to figure out a more secure way to improve this next week.

bestander:
@arcanis, you need to add a test too otherwise we won't notice regressions later
Can you copy https://github.com/yarnpkg/yarn/pull/3563/files#diff-c2dc3bd2d0bdc7b8d87c36ed3039c2d4R300?

PR #3674, opened 2017-06-19
feat(config): use XDG config specification

michael-zapata:
As per [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), prioritise the environment variable `$XDG_CONFIG_PATH` when possible, and use `$HOME/.config/` as a fallback.

Backward-compatibility is handled to avoid unnecessary breakage, this is a first step towards full compliance with the spec.

Related to #2334 (But only solves part of it)

### DIFF
--- a/src/util/rc.js
+++ b/src/util/rc.js
@@ -2,6 +2,7 @@
 
 import {readFileSync} from 'fs';
 import {basename, dirname, join} from 'path';
+import {CONFIG_DIRECTORY} from '../constants.js';
 
 const etc = '/etc';
 const isWin = process.platform === 'win32';
@@ -49,6 +50,7 @@ export function findRc(name: string, parser: Function): Object {
   }
 
   if (home) {
+    addConfigPath(CONFIG_DIRECTORY, `${name}rc`);
     addConfigPath(home, '.config', name, 'config');
     addConfigPath(home, '.config', name);
     addConfigPath(home, `.${name}`, 'config');
--- a/end_to_end_tests/data/run-yarn-test.sh
+++ b/end_to_end_tests/data/run-yarn-test.sh
@@ -12,6 +12,11 @@ fail_with_log() {
   exit $exitcode
 }
 
+check_config() {
+  XDG_CONFIG_HOME=~/config_path yarn config get $1 > config_output
+  echo $2 | diff config_output - || fail_with_log
+}
+
 cd /tmp
 mkdir yarntest
 cd yarntest
@@ -23,3 +28,12 @@ yarn --version || fail_with_log
 rm -rf ~/.cache/yarn
 
 yarn add react || fail_with_log
+
+# Ensure that we follow the xdg spec
+mkdir ~/config_path
+echo "foo bar" > ~/config_path
+check_config foo bar
+
+# Ensure that compatibility with the old config format is maintained
+echo "bar baz" >> ~/.yarnrc
+check_config bar baz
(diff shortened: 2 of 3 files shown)


bestander:
The only concern is the extra clutter, can we have it in one place?

michael-zapata:
Do you mean between the two registry files?

I tried to put everything in `YarnRegistry` but it felt hackish to repeat parts of `getPossibleConfigLocations` to keep it contained there. And using a `yarn` directory inside the `NpmRegistry` file felt even more wrong. I can certainly modify the constructor to include the name, but that feels strange to say the least. I could use an internal `name` field in both classes instead. That'd still require some changes in both classes, but would keep things cleaner than the empty string hack. What do you have in mind?

src/util/rc.js
```diff
@@ -49,6 +49,7 @@ export function findRc(name: string, parser: Function): Object {
   }
 
   if (home) {
+    addConfigPath(process.env.XDG_CONFIG_HOME || join(home, '.config', name, `${name}rc`));
```
bestander: not clear why process.env.XDG_CONFIG_HOME OR home/.config is used now

src/registries/yarn-registry.js
```diff
@@ -41,7 +41,7 @@ export default class YarnRegistry extends NpmRegistry {
   constructor(cwd: string, registries: ConfigRegistries, requestManager: RequestManager, reporter: Reporter) {
     super(cwd, registries, requestManager, reporter);
 
-    this.homeConfigLoc = path.join(userHome, '.yarnrc');
+    this.homeConfigLoc = process.env.XDG_CONFIG_HOME || path.join(userHome, '.config', 'yarn', 'yarnrc');
```
bestander: Why process.env.XDG_CONFIG_HOME can't define userHome or config.userConfig

michael-zapata:
The change should be far less intrusive now. I'll look at how e2e tests are done when I have some time to dedicate to this.

bestander:
Yeah, that is way better, thanks @michael-zapata.
Waiting for tests.

bestander:
Thanks, @michael-zapata.
end_to_end_tests/data/run-yarn-test.sh is not the best place for very specific e2e tests, this script is hard to monitor and debug on CI and it is not really scalable.

Can you express the test similar to this https://github.com/yarnpkg/yarn/commit/d709c2fa2ca1d89e456fcbe9c0b3a4eedb3d5088?

michael-zapata:
I pushed to see if results were different in the CI, but that test is failing in local, as in both checks. I can get behind one of them failing, but the two of them seems rather strange.

I should dig further before doing tests as this is obviously not working. :)

BYK:
@michael-zapata the CI seems fine. Do you need to work on this still or can we merge?

@bestander any more concerns about this PR or do you think it is ready?

michael-zapata:
Hey, sorry for not getting back on that. The test that was added isn't running in the CI from what I can see, as it fails completely locally. I didn't have the time to take a look at why it's failing and that should be fixed before merging as this is not functional at the moment.

BYK:
@michael-zapata alright, how can I assist you with that? :)

michael-zapata:
I just need to investigate to understand why this isn't working, I am not really familiar with the codebase and I won't have the time to do that before this week-end. :/

Some pointers on where this might fizzle would be very much welcome though!

BYK:
>Some pointers on where this might fizzle would be very much welcome though!

I don't know that end-to-end script either so can't really help quickly :( That said once I have time to get back to this, I can help if you don't beat me to it :)

HaleTom:
Any chance of moving the `node_modules` directory to `$XDG_DATA_HOME`?

bestander:
Some time has passed, let's close the PR as there are too many conflicts.
If anyone wants to reopen a new one please go ahead.

PR #3675, opened 2017-06-19
Include top-level package.json in engines check

CrabDude:
**Summary**

Add support for enforcing engine versions (especially yarn) at the project level (as opposed to the [sub]dependencies).

This is especially important on large projects to prevent `yarn.lock` churn.

**Test plan**

`yarn test`. Manual verification by setting various versions in `"engines"`.

**TODO**

Needs language translation, and possibly a better log message.

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -412,6 +412,7 @@ export class Install {
       patterns: rawPatterns,
       ignorePatterns,
       workspaceLayout,
+      manifest,
     } = await this.fetchRequestFromCwd();
     let topLevelPatterns: Array<string> = [];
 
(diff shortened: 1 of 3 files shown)


src/cli/commands/install.js
```diff
@@ -422,6 +423,11 @@ export class Install {
     }
 
     steps.push(async (curr: number, total: number) => {
+      this.reporter.step(curr, total, this.reporter.lang('fetchingPackages'), emoji.get('mag'));
```
arcanis: You should replace `fetchingPackages` with `checkingManifest` here - however I'm a bit concerned it will clutter the display. Could you add this step only if there's an `engine` field?

CrabDude: ✅

bestander:
Thanks, @CrabDude

PR #3686, opened 2017-06-21
Add retry on 5xx errors. Fixes #2849

### LINKED ISSUE #2849: Ensure Yarn retries on 5xx errors
Both

Yarn sometimes shows 5xx errors when installing packages (eg. see #2769)

Yarn is supposed to be resilient to poor network conditions. It sounds like it might not be retrying correctly in some scenarios. We should verify that it does indeed retry on 5xx errors, by using a debugging proxy such as [Fiddler](http://www.telerik.com/fiddler) to intentionally inject 500 errors and see how Yarn reacts.

torifat:
This is not a very elegant solution but it will do the job for now.

During fetching tarball for any **5xx** errors, it will show `warning There appears to be trouble with our server. Retrying...` and try again after `3000ms`. If it fails again then it will show the error like before.

**Future Work**
Refactor `RequestManager` to add support for retry.

I spent some time trying to refactor `RequestManager`, but it will take some time and effort to finish it.  And, since `RequestManager` is a very important part of **Yarn**. So, it is important that it doesn't degrade the speed after refactor.

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -1,12 +1,13 @@
 /* @flow */
 
 import http from 'http';
-import {SecurityError, MessageError} from '../errors.js';
+import {SecurityError, MessageError, ResponseError} from '../errors.js';
 import type {FetchedOverride} from '../types.js';
 import * as constants from '../constants.js';
 import * as crypto from '../util/crypto.js';
 import BaseFetcher from './base-fetcher.js';
 import * as fsUtil from '../util/fs.js';
+import {sleep} from '../util/misc.js';
 
 const path = require('path');
 const tarFs = require('tar-fs');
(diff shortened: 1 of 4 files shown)


src/fetchers/tarball-fetcher.js
```diff
@@ -132,46 +132,64 @@ export default class TarballFetcher extends BaseFetcher {
...
+        if (retry) {
+          this.reporter.warn(this.reporter.lang('retryOnInternalServerError'));
+          setTimeout(() => {
+            makeRequest().then(resolve).catch(reject);
+          }, 3000);
+        } else {
+          reject(error);
+        }
+      });
+    });
```
Daniel15: Might be cleaner to implement this as a loop, something like:
```js
async function fetchStuff(): Promise<void> {
  let retriesRemaining = 3;
  do {
    try {
      await makeRequest();
      return;
    } catch (ex) {
      this.reporter.warn(this.reporter.lang('retryOnInternalServerError'));
      retriesRemaining--;
    }
  } while (retriesRemaining > 0);
  throw new Error('Ran out of retries!'); // or whatever
}
```

torifat: Looks much cleaner & reusable 😄

bestander:
Ping, what is next? :)

torifat:
@bestander I will update the PR with changes by tomorrow. A bit sick today.

torifat:
`RequestManager.execute timeout error with default maxRetryAttempts` is failing in CI but not locally 😕

Daniel15:
Looks like CI is passing now. This looks good to me! Thanks :D

Daniel15:
Is it possible to write a unit test for this? Mock the request such that it fails with a 500 error the first time, then succeeds the second time?

bestander:
@torifat, great job!
Any thoughts about a unit test?

torifat:
I'm stuck with the following problem.

I have mocked `request` module and sent a `500` `statusCode`.
```js
if (err instanceof ResponseError && err.responseCode >= 500 && retriesRemaining > 1) {
  retriesRemaining--;
  this.reporter.warn(this.reporter.lang('retryOnInternalServerError'));

  console.log(`Before Sleep: ${retriesRemaining}`);
  // ^------- I can see this output in console

  await sleep(3000);

  console.log('After Sleep');
  // ^------- I can't see this output in console :(

}
```

I tried `jest.runAllTimers();` after passing the fake response but it's not working. If I use `jest.useRealTimers()` then it works.

bestander:
@torifat, can you commit the test code, we'll have a look

torifat:
@balderdash pushed.

__tests__/fetchers.js
```diff
@@ -206,3 +206,34 @@ test('TarballFetcher.fetch properly stores tarball of scoped package in offline
...
   expect(exists).toBe(true);
 });
+
+test.only('TarballFetcher.fetch retry on 500', async () => {
+  // jest.useRealTimers();
+  const dir = await mkdir('tarball-fetcher');
+  jest.mock('request', factory => options => {
+    return {
+      on: (name, cb) => {
+        if (name === 'response') {
```
bestander: add a log here, this function is called before the request is called, at the moment when Yarn registers `on` callbacks.
See how mock is done in `__tests__/__mocks__/request.js`

torifat: hmm, thanks for pointing it out. `__tests__/__mocks__/request.js` doesn't have any support for testing stream. I'm trying to figure out a way.

bestander: Thanks, @torifat, if you are too busy right now it would be better to merge the fix and add the test to the backlog.

torifat: Yeah, I will be free next week. Let me remove the test commit.

bestander:
Thanks, @torifat!

src/fetchers/tarball-fetcher.js
```diff
@@ -130,48 +131,66 @@ export default class TarballFetcher extends BaseFetcher {
...
-          if (tarballCachePath) {
-            validateStream.pipe(fs.createWriteStream(tarballCachePath)).on('error', reject);
-          }
-
-          validateStream.pipe(extractorStream).on('error', reject);
-        },
-      },
-      this.packageName,
-    );
+    let retriesRemaining = 2;
```
BYK: @bestander shall we pull this out to a constant like `MAX_SERVER_ERROR_RETRIES` or something?

src/fetchers/tarball-fetcher.js
```diff
@@ -130,48 +131,66 @@ export default class TarballFetcher extends BaseFetcher {
...
+              validateStream.pipe(extractorStream).on('error', reject);
+            },
+          },
+          this.packageName,
+        );
+      } catch (err) {
+        if (err instanceof ResponseError && err.responseCode >= 500 && retriesRemaining > 1) {
+          retriesRemaining--;
+          this.reporter.warn(this.reporter.lang('retryOnInternalServerError'));
+          await sleep(3000);
```
BYK: Same as above, should we pull this out to `RETRY_DELAY`? We actually have this same delay in a few places now.

src/util/misc.js
```diff
@@ -78,3 +78,9 @@ export function compareSortedArrays<T>(array1: Array<T>, array2: Array<T>): bool
   }
   return true;
 }
+
+export function sleep(ms: number): Promise<void> {
```
BYK: <3 async/await

PR #3691, opened 2017-06-21
fixes hardlink collitions

### LINKED ISSUE #2734: --link-duplicates error

This is a bug in the pre-release `0.21.2` version with the `--link-duplicates` flag.

Installing a project with `@kadira/storybook` and `webpack@2` results in the following error:

> error An unexpected error occurred: "EEXIST: file already exists, link '/home/git/forks/yarn-link-bug/node_modules/@kadira/storybook/node_modules/enhanced-resolve/node_modules/memory-fs/.npmignore' -> '/home/git/forks/yarn-link-bug/node_modules/babel-loader/node_modules/enhanced-resolve/node_modules/memory-fs/.npmignore'".

Node 6.9.2, Yarn 0.21.2, OSX Sierra.

**Reproduction Repository**

A [reproduction is available](https://github.com/STRML/yarn-link-bug).

Ref: https://github.com/yarnpkg/yarn/pull/2620#issuecomment-281146567 and [this tweet](https://twitter.com/bestander_nz/status/833696608390836224). My preference would be for the feature to remain as it gives us significant (20%) size savings on our frontend project.

bestander:
**Summary**

Fixes #2734

The problem in hardlinking code trying to create a hardlink per package twice.
That might happen when the same there were a few packages hardlinked in a hierarchy.

**Test plan**

Added test from the issue.


src/util/fs.js
```diff
@@ -450,6 +450,8 @@ async function buildActionsForHardlink(
 
       // push all files to queue
       invariant(srcFiles, 'src files not initialised');
+      // hardlinking is per package, not going into sub packages
+      srcFiles = srcFiles.filter(f => f !== 'node_modules');
```
arcanis: Why is there paths from the node_modules here? Shouldn't they be handled outside of this function (not completely sure how works the linking pipeline yet)? + will it work with `bundledDependencies`?

src/fetchers/tarball-fetcher.js
```diff
@@ -175,13 +176,16 @@ export default class TarballFetcher extends BaseFetcher {
   }
 
   async _fetch(): Promise<FetchedOverride> {
+    const isFilePath = this.reference.startsWith('file:');
```
bestander: copied from https://github.com/yarnpkg/yarn/pull/3709 for bundledDependencies case

bestander:
@arcanis, you were right, this is trickier but I found a way around and added more tests

bestander:
Review please

src/util/fs.js
```diff
@@ -360,6 +360,11 @@ async function buildActionsForHardlink(
     const {src, dest} = data;
     const onFresh = data.onFresh || noop;
     const onDone = data.onDone || noop;
+    if (files.has(dest)) {
```
bestander: that is all the fix

arcanis: What happens if an hardlinked file needs to change? Will it be removed in an earlier pass? Unless it is, maybe we could try comparing the inodes as well

bestander: Oh it does compare inodes for sure, this check just prevents assessing the same path twice, a folder in particular.
We have `A -> B` structure that we want to hardlink to `A1 -> B1`, package-linker passes that modules A1 and B1 need to be hardlinked independently but part of hardlinking is that it is recursive, so files in B1 will be scheduled to be linked twice.

I could not find a more elegant way of doing this especially with bundledDependencies because they don't participate in hoisting.

bestander: I'll add this comment to the code for clarity

PR #3701, opened 2017-06-22
Improve handling of packages without manifest

sheerun:
This improves over #3624 by:

- Better guessing package names (added a test)
- Adding Support for hosted git, like github or gitlab (including shorthand notation)
- Adding Support for .tar.gz archives (both local and remote)


sheerun:
Failed build on appveyor seems unrelated

Issue #3693, opened 2017-06-22
Fix .includes issue once and for all


bestander:
**Do you want to request a *feature* or report a *bug*?**
Node 4 compat issue

**What is the current behavior?**

1. Jest tests on Node 4 don't catch compatibility issues like having `.includes` in `-legacy.js` builds.
2. We need to include proper polyfil for Node 4

Volune:
The simple way would be to add a dependency to [`babel-polyfill`](https://yarnpkg.com/en/package/babel-polyfill) and require it at the beginning.

Node 4 already supports several ES6 features ([compatibility table](https://kangax.github.io/compat-table/es6/)). I think it would be better to add a dependency to [`core-js`](https://yarnpkg.com/en/package/core-js), and only require the necessary polyfills.

`core-js` is internally used by `babel-polyfill` and `babel-runtime`. `babel-runtime` is already used in Yarn, but does not handle instance functions such as `Array.prototype.includes`.

bestander:
@Volune, would you send a PR for that please?

kaylieEB:
I also think that `babel-polyfill` is quite heavy, if you want to fix just .includes issue, then we should use airbnb's .includes polyfill https://www.npmjs.com/package/array-includes.

bestander:
Makes sense.
On top of that we need to make jest run without polyfills, these issues
should be caught by CI

kaylieEB:
hm, looking at [babelrc](https://github.com/yarnpkg/yarn/blob/c844df7236ec122ca6b803b9f2943352c9b09a7e/.babelrc)
it looks like jest already runs without the polyfill? I actually think we just need to add a [plugin](https://www.npmjs.com/package/babel-plugin-array-includes)
 to the pre-node setting. Happy to send a PR.

bestander:
Go ahead, thanks

kaylieEB:
Thanks! https://github.com/yarnpkg/yarn/pull/3705

Volune:
I think we should have a look at [babel-preset-env](http://babeljs.io/docs/plugins/preset-env/#optionsuse-built-ins) to fix all similar potential issues at once.

voxsim:
@bestander here, https://github.com/yarnpkg/yarn/pull/3652, I upgraded babel and used babel-preset-env.. If you want I can move some commits and arrange a pull request for the next week. This is propedeutic for the new release of Jest too.
[...]

PR #3705, opened 2017-06-23
Add new plugin to .babelrc to transpile Array.prototype.includes

kaylieEB:
**Summary**

Transpiles `Array.prototype.includes` correctly per Issue https://github.com/yarnpkg/yarn/issues/3693

**Test plan**
Ran `yarn run build` and manually checked the lib and lib-legacy files to check that only the latter had transpiled version.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,6 +4,7 @@
     "pre-node5": {
       "presets": ["es2015-node4"],
       "plugins": [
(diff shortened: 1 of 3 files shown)


hzoo:
I think y'all know this already but that plugin only works for `[].includes` not `var.includes` if var is an array

PR #3709, opened 2017-06-23
fixes transtive file: dependencies

bestander:
**Summary**

Transitive file: dependencies were losing relative path indication and you could not install `root -> a-> file:./b`

The problem was that between tarball-resolver and tarball-fetcher we were loosing information about file: protocol that clearly identifies that the tarball is local.

**Test plan**

Added test.


__tests__/fixtures/install/install-prod-deduped-direct-dev-dep/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "dependencies": {
-    "a": "file:a"
+    "a": "file:./a"
   },
   "devDependencies": {
-    "b": "file:b"
+    "b": "file:./b"
```
bestander: Without this fix the test was failing.
That is strange and needs a bit more investigation later on

src/resolvers/exotics/file-resolver.js
```diff
@@ -107,7 +107,7 @@ export default class FileResolver extends ExoticResolver {
         if (temp === section) {
           temp = Object.assign({}, section);
         }
-        temp[k] = `file:${path.relative(this.config.cwd, path.join(loc, util.removePrefix(v, 'file:')))}`;
+        temp[k] = `file:./${path.relative(this.config.cwd, path.join(loc, util.removePrefix(v, 'file:')))}`;
```
bestander: An alternative would be to make the fix at fetcher level that tries to fetch non relative paths from registry.

src/fetchers/tarball-fetcher.js
```diff
@@ -174,13 +175,16 @@ export default class TarballFetcher extends BaseFetcher {
   }
 
   async _fetch(): Promise<FetchedOverride> {
+    const isFilePath = this.reference.startsWith('file:');
```
bestander: copied from https://github.com/yarnpkg/yarn/pull/3709 to test bundledDependencies case

PR #3712, opened 2017-06-24
add a check that same file is not copied twice

bestander:
**Summary**

If the same file is copied twice during a bulk copy then we have a problem in out recursive logic.
Added an invariant to make sure we don't make extra IO operation.

**Test plan**

all tests pass

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -163,6 +163,7 @@ async function buildActionsForCopy(
     const {src, dest, type} = data;
     const onFresh = data.onFresh || noop;
     const onDone = data.onDone || noop;
+    invariant(!files.has(dest), `The same file ${dest} can't be copied twice in one bulk copy`);
     files.add(dest);
 
     if (type === 'symlink') {


src/util/fs.js
```diff
@@ -163,6 +163,7 @@ async function buildActionsForCopy(
     const {src, dest, type} = data;
     const onFresh = data.onFresh || noop;
     const onDone = data.onDone || noop;
+    invariant(!files.has(dest), `The same file ${dest} can't be copied twice in one bulk copy`);
```
arcanis: Is it efficient? It `O(N²)`, right?

bestander: Why would it be?
Set should be implemented by a HashSet and its has operation should be sublinear, O(1) in an ideal situation.
https://jsperf.com/array-indexof-vs-set-has

arcanis: Yeah I'm wrong, I thought it was an array (in which case it would be iterating over the array for each element to copy, hence square complexity), but with a set it's all fine 👍

PR #3715, opened 2017-06-24
lockfile optimized

bestander:
just updated lockfile that selfupdates after install

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2726,19 +2726,7 @@ istanbul-lib-hook@^1.0.7:
   dependencies:
     append-transform "^0.4.0"
 
-istanbul-lib-instrument@^1.1.1, istanbul-lib-instrument@^1.7.2:
-  version "1.7.2"
-  resolved "https://registry.yarnpkg.com/istanbul-lib-instrument/-/istanbul-lib-instrument-1.7.2.tgz#6014b03d3470fb77638d5802508c255c06312e56"
-  dependencies:
-    babel-generator "^6.18.0"
-    babel-template "^6.16.0"
-    babel-traverse "^6.18.0"
-    babel-types "^6.18.0"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -2726,28 +2726,28 @@ istanbul-lib-hook@^1.0.7:
...
     babylon "^6.13.0"
-    istanbul-lib-coverage "^1.1.1"
+    istanbul-lib-coverage "^1.1.0"
     semver "^5.3.0"
 
-istanbul-lib-instrument@^1.7.1:
-  version "1.7.1"
-  resolved "https://registry.yarnpkg.com/istanbul-lib-instrument/-/istanbul-lib-instrument-1.7.1.tgz#169e31bc62c778851a99439dd99c3cc12184d360"
+istanbul-lib-instrument@^1.7.2:
+  version "1.7.2"
```
SimenB: Why isn't this deduped away?

bestander: Currently Yarn does deduping on the go "for free", this PR should address it https://github.com/yarnpkg/yarn/pull/3729.

I think deduping should be either during upgrade operations or when deduping was requested with a command.
Not sure if --force should dedupe

SimenB: Will `yarn dedupe` become a thing? Right now I'm stuck manually deduping the lockfile, which really sucks.

>error The dedupe command isn't necessary. `yarn install` will already dedupe.

This is a lie

bestander: Sorry, @SimenB, we just had to mitigate the issue of Yarn being too eager to dedupe the packages during any installation which resulted in unexpdected behavior.

Please chime in into this issue https://github.com/yarnpkg/yarn/issues/3778 and help out to resolve it

SimenB: Ah, didn't know about that issue, thanks!

Issue #3617, opened 2017-06-09
Error message could be better


sarupbanskota:
**Do you want to request a *feature* or report a *bug*?**

bug / minor enhancement

**What is the current behavior?**

```
--- gig/babel ‹enforce_plugin_names* ?› » yarn --global add babel-register                       1 ↵
yarn install v0.24.5
error `install` has been replaced with `add` to add new dependencies. 
Run "yarn global add add babel-register" instead.
```

_Notice the double add_

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

would be nicer to have had something like:

```
Error: "--global" isn't supported. Run "yarn global add babel-register"
```

**Please mention your node.js, yarn and operating system version.**
node 8, yarn v0.24.5

aij:
There are similarly confusing error messages when other flags are specified before the command as well. Rather than recognizing the command that is after the flags, yarn will infer the command as `install`, and then complain about the `install` being used inappropriately. This is particularly surprising on GNU/Linux where the position of flags tends to be quite flexible. (Some of) the BSDs tend to be more strict.

For example, `yarn test --pure-lockfile --frozen-lockfile` will run tests, whereas changing the argument order results in this misleading error:

```
$ yarn --pure-lockfile --frozen-lockfile test
yarn install v0.24.6
info No lockfile found.
error `install` has been replaced with `add` to add new dependencies. Run "yarn add test" instead.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Would it be reasonable to allow options (or at least global options) to appear before the command? (I'm not familiar enough with older version of yarn to know whether that might produce surprising new interpretations of previously supported invocations or not.)

Issue #3620, opened 2017-06-10
yarn global list isn't supported


sarupbanskota:
**Do you want to request a *feature* or report a *bug*?**
bug / enhancement

**What is the current behavior?**
```
$ yarn global ls                             1 ↵
yarn global v0.24.5
warning No license field
✨  Done in 0.06s.
```

```
$ yarn global list                           1 ↵
yarn global v0.24.5
error Invalid subcommand. Try "add, bin, ls, remove, upgrade, upgrade-interactive"
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

**What is the expected behavior?**

the same as `yarn global ls`

also based on what the website says, it sounds like `list` should be supported as a yarn subcommand.
https://yarnpkg.com/lang/en/docs/cli/global/

> Read more about the commands that can be used together with yarn global:
>
> yarn add: add a package to use in your current package.
yarn bin: displays the location of the yarn bin folder.
yarn list: list installed packages.
yarn remove: remove a package that will no longer be used in your current package.
yarn upgrade: upgrades packages to their latest version based on the specified range.

**Please mention your node.js, yarn and operating system version.**
yarn version v0.24.5, node 8

rockchalkwushock:
@sarupbanskota 

How do you have your PATH setup? This will matter with where `yarn global ls` is looking at. You can refer to #630. I also made note of what this says in this [gist](https://gist.github.com/rockchalkwushock/a1b58cc38b0a59d11284a4715393d1fd#2-cli-installs). You'll have to scroll down to the bit about `yarn`. Hope this is what the issue is an you can get it working 👍

robsonrosa:
@rockchalkwushock I'm not sure if its related with current PATH setup. The documentation itself is ambiguos. At the first paragraph we have:
> yarn global is a prefix used for a number of commands like `add`, `bin`, `ls` and `remove`.

And at the "read more" section, the command list changes to:

> Read more about the commands that can be used together with yarn global:
> `yarn add`: add a package to use in your current package.
> `yarn bin`: displays the location of the yarn bin folder.
> `yarn list`: list installed packages.
> `yarn remove`: remove a package that will no longer be used in your current package.
> `yarn upgrade`: upgrades packages to their latest version based on the specified range.

So, locally we should use `yarn list` and globally `yarn global ls`. And it's pretty confusing to me.

sarupbanskota:
yeah, what @robsonrosa said, basically :-)

[...]

PR #3719, opened 2017-06-25
✎ Remove duplicate 'add' when using global flag

yeskunall:
**Summary**

Just a small typo fix as pointed out by @sarupbanskota as seen when warned to use `add` instead of `install` (using the `global` flag)

Cheers! ❤️

**Test plan**

N/A

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -857,7 +857,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
     let command = 'add';
     if (flags.global) {
-      command = 'global add';
+      command = 'global';
     }
     throw new MessageError(reporter.lang('installCommandRenamed', `yarn ${command} ${exampleArgs.join(' ')}`));
   }


bestander:
Thanks, @yeskunall.
Could you share a bit more context on the fix?

Maybe a screenshot of the output before and after?

sarupbanskota:
@bestander this has to do with https://github.com/yarnpkg/yarn/issues/3620

yeskunall:
Actually @sarupbanskota this simply fixes #3617, and nothing more. I haven't looked at what #3620 is, but this is merely a grammatical fix, nothing more.

I'm currently at work, but @bestander you can expect a screenshot today itself. Sorry for the delay.

yeskunall:
So following from the issue reported in #3617
```
yarn global add add babel-register
```
`yarn` _did_ do the right thing, and the double add only arises due to the fact that it recognises `add` as a package instead of a command.

BYK:
@bestander ping. (feel free to hand this over to me)

yeskunall:
@BYK so my original changes should probably be discarded, but can you maybe shed some light on what @bestander meant in his review?

BYK:
@yeskunall I don't like talking on behalf of people so that's why I pinged him directly. He's in a different time zone so I'm hoping he'd respond when he wakes up :) If not I'll dig in.

bestander:
I think we have an understanding here.

The current change can't be accepted because it breaks error message for `yarn install --global left-pad`, for example because it will say `Run" yarn global left-pad" instead`.

I don't have an opinion how this should be fixed, feel free to use your best judgement and cover different usages of error message.

yeskunall:
I'll look into this! Thanks for the feedback, you guys! ❤️

BYK:
@yeskunall do you need any help from us? I'd love to get this landed before Yarn 1.0! :)

bestander:
Thanks @yeskunall, I think a better solution came from #4040.
Don't feel bad about it, I hope to see more PRs from you here.

PR #3721, opened 2017-06-26
Fix yarn global prefix directory

KishanBagaria:
Previous PR: #3458

Fixes: #2064

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -14,7 +14,10 @@ import {run as runRemove} from './remove.js';
 import {run as runUpgrade} from './upgrade.js';
 import {run as runUpgradeInteractive} from './upgrade-interactive.js';
 import {linkBin} from '../../package-linker.js';
+import {POSIX_GLOBAL_PREFIX, FALLBACK_GLOBAL_PREFIX} from '../../constants.js';
 import * as fs from '../../util/fs.js';
+
+const nativeFs = require('fs');
 
 class GlobalAdd extends Add {
   maybeOutputSaveTree(): Promise<void> {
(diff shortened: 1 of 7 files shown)


src/cli/commands/global.js
```diff
@@ -69,44 +72,43 @@ async function getBins(config: Config): Promise<Set<string>> {
...
-    if (process.env.DESTDIR) {
-      prefix = path.join(process.env.DESTDIR, prefix);
+      return path.join(process.env.LOCALAPPDATA, 'Yarn');
     }
+    return FALLBACK_GLOBAL_PREFIX;
+  }
 
-    return prefix;
+  try {
+    await fs.access(path.join(POSIX_GLOBAL_PREFIX, 'bin'), nativeFs.constants.W_OK);
```
KishanBagaria: This `fs.access` call can be memoized by adding this line after it:

    await config.registries.yarn.saveHomeConfig({prefix: POSIX_GLOBAL_PREFIX});

KishanBagaria: If you want to ensure if we have permission to write to `%LOCALAPPDATA%\Yarn`, that can be done. It's likely redundant though, because it's almost guaranteed to be writable, at least as writable as `~/.yarn` (`FALLBACK_GLOBAL_PREFIX`).

I guess it won't hurt to do that.

src/cli/commands/global.js
```diff
@@ -69,44 +72,44 @@ async function getBins(config: Config): Promise<Set<string>> {
...
     return process.env.PREFIX;
-  } else if (process.platform === 'win32') {
+  }
+
+  let prefix = FALLBACK_GLOBAL_PREFIX;
+  if (process.platform === 'win32') {
+    // %LOCALAPPDATA%\Yarn --> C:\Users\Alice\AppData\Local\Yarn
     if (process.env.LOCALAPPDATA) {
-      return path.join(process.env.LOCALAPPDATA, 'Yarn', 'bin');
+      prefix = path.join(process.env.LOCALAPPDATA, 'Yarn');
```
BYK: Let's make this `yarn` as we are at it, instead of the upper-cased `Yarn`?

@Daniel15 @bestander any reason we were using `Yarn`?

src/constants.js
```diff
@@ -59,6 +59,9 @@ export const CONFIG_DIRECTORY = getDirectory('config');
 export const LINK_REGISTRY_DIRECTORY = path.join(CONFIG_DIRECTORY, 'link');
 export const GLOBAL_MODULE_DIRECTORY = path.join(CONFIG_DIRECTORY, 'global');
 
+export const POSIX_GLOBAL_PREFIX = '/usr/local';
```
BYK: May be we should do `path.dirname('path/to/node')` as suggested in https://docs.npmjs.com/files/folders#prefix-configuration ?

BYK:
Also this patch needs a rebase to get the tests passing on CI again.

src/cli/commands/global.js
Daniel15: @BYK The standard on Windows is to use uppercase. If you take a look in `%LocalAppData%` you'll see that most folders are uppercase. Some badly-behaved apps use lowercase, but it's not the norm.

BYK: Alright, NVM my comment then. Yarn should be a nice kitty.

src/constants.js
Daniel15: Perhaps we should try both. `/usr/local` is good if it's writable by the user, otherwise the directory that Node.js is installed from (I think there's some property on `process` that has the full path of the Node.js process?)

BYK: From what I understand from NPM docs, we should to this in the opposite order: try `node`s folder first. That said not sure how well it would play with homebrew or my local folder hacks (I alias `node` to `~/Desktop/node8` etc. for quick testing purposes and I heard this is not that uncommon)

KishanBagaria:
@BYK Rebased and tests pass.

src/constants.js
KishanBagaria: @BYK Implementing NPM's behavior of using parent directories of the node executable will come with bugs, I've described this in #3458 and https://github.com/yarnpkg/yarn/issues/2064#issuecomment-303823304.

[Daniel](https://github.com/yarnpkg/yarn/issues/2064#issuecomment-303592899) also said something similar:

> The issue is that Yarn symlinks the executables into the wrong directory (the Node.js directory rather than somewhere reasonable). It doesn't really make sense to use the Node.js directory for these scripts, and on many systems that directory is not even writable by regular users.

BYK: @KishanBagaria - oh okay. Sorry I got lost in those long conversations. Thanks a lot for clarifying!

Shall we put this into the code as a comment?

KishanBagaria: No worries, the whole global-prefix has been a quite long conversation indeed.

Maybe the website docs can be updated with how we get the prefix so that regular users find the info easily, and why we don't mimic NPM.

BYK: >Maybe the website docs can be updated with how we get the prefix so that regular users can find the info easily, and why we don't mimic NPM.

I think this is a great idea! @Daniel15 any suggestions? I can file an issue for this so we don't forget.

Issue #3677, opened 2017-06-20
Allow `git+file://` dependencies


Volune:
Yarn 0.24.6 and master branch

`yarn add git+file:///path/to/git/repo#branch` is not supported, it'll try to find the package in npm registry.
It does work in NPM (5.0.3).

Expected behavior: `file:///path/to/git/repo#branch` should be treated as a git remote repository, the same way as if it was accessed by SSH or HTTPS.

I traced the problem to [git-resolver.js:16](https://github.com/yarnpkg/yarn/blob/eded72393fd83717b80c5541199c7849acc66e6d/src/resolvers/exotics/git-resolver.js#L16). I think we should allow any `git+***://` protocols, and let git handle the rest. This would also allow `git+ftp://` for example.

PR #3735, opened 2017-06-27
Allow `git+***://` dependencies

Volune:
**Summary**

Fixing #3677

> `yarn add git+file:///path/to/git/repo#branch` is not supported, it'll try to find the package in npm registry.
> It does work in NPM (5.0.3).
>
> Expected behavior: file:///path/to/git/repo#branch should be treated as a git remote repository, the same way as if it was accessed by SSH or HTTPS.

**Test plan**

Adding some tests on `getExoticResolver` to ensure the Git resolver is used for `git+file:///` dependencies.

**Implementation notes**

I moved `getExoticResolver` to `src/resolvers/index.js`, I think it's more suitable than `src/package-request.js`, and more convenient to write the tests. I also refactored at little to make the returned type more specific.

I made the Git resolver handle any protocol prefixed with `git+`, not only `git+file`. That way, yarn will handle as many protocols as git itself does.

### DIFF
--- a/src/resolvers/exotics/git-resolver.js
+++ b/src/resolvers/exotics/git-resolver.js
@@ -12,8 +12,10 @@ import Git from '../../util/git.js';
 
 const urlParse = require('url').parse;
 
+const GIT_PROTOCOL_PATTERN = /git\+.+:/;
+
 // we purposefully omit https and http as those are only valid if they end in the .git extension
-const GIT_PROTOCOLS = ['git:', 'git+ssh:', 'git+https:', 'ssh:'];
+const GIT_PROTOCOLS = ['git:', 'ssh:'];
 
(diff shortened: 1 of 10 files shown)


bestander:
nice work, @Volune!

src/resolvers/exotics/git-resolver.js
```diff
@@ -12,8 +12,10 @@ import Git from '../../util/git.js';
 
 const urlParse = require('url').parse;
 
+const GIT_PROTOCOL_PATTERN = /git\+.+:/;
```
BYK: Since we're using patterns now, do you think we can remove `GIT_PROTOCOLS` and special casing for `https://` + ends with `.git` custom rules and bake them into the pattern in the future?

mbahiense:
Great!! o/

PR #3739, opened 2017-06-28
Yarn workspaces: running commands from subfolders that don't match glob

bestander:
This fixes an issue that if you run a Yarn command from any folder that is a subfolder of a workspace it will automatically change CWD to the root.

Before changing CWD the code now checks that workspaces glob matches the original CWD.

**Test plan**

Added test


arcanis:
So it will become impossible to run a command inside another workspace context if you're not located yourself in a workspace, is that correct?

The previous behavior was intentional, but I'm ok with changing this.

bestander:
No, you can run 'yarn install' in a sub folder of a workspace root but if
it is not a workspace itself.

Example - e2e tests inside Jest.

All those tests have their package.json files but we don't necessarily want
to add all of them to the root jest workspace.
At the same time we may need to run Yarn commands on the tests without
being moved to the workspace root.

PR #3743, opened 2017-06-28
Fix: inherit env variables from process when calling Git

### LINKED ISSUE #3742: Git does not have access to regular environment variables when cloning
Bug

Git fails to connect to repos behind a proxy since certain environment variables are not passed to it when making the call. Internal reference: t19705486.

Block git access without a proxy and then try to use the proxy. Normal `git xxx` commands from the shell passes but commands issued by `yarn` fails.

`yarn` commands to behave like the ones executed from the shell.

Node v6.10.3
yarn 0.27.1

BYK:
**Summary**

Fixes #3742.

With #3633 we started overriding the default `env` map that was passed
to `git` calls. It defaulted to `process.env` and since the option is
a whitelist, now it is only the ones we wanted to add, instead of
`process.env` + the additions. This patch fixes the problem by spreading
`process.env` before the new values.

Relevant part from Node docs:

https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options
>Use env to specify environment variables that will be visible to the new process, the default is process.env.

**Test plan**

Manual verification for t19705486. We should have automated tests for this
so I'll start working on them. Just want the fix our first.

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -32,7 +32,12 @@ const supportsArchiveCache: {[key: string]: boolean} = map({
 });
 
 // Suppress any password prompts since we run these in the background
-const env = {GIT_ASKPASS: '', GIT_TERMINAL_PROMPT: 0, GIT_SSH_COMMAND: 'ssh -oBatchMode=yes'};
+const env = {
+  ...process.env,
+  GIT_ASKPASS: '',
+  GIT_TERMINAL_PROMPT: 0,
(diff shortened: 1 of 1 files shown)


cpojer:
Do we want to add a regression test so this doesn't happen again?

BYK:
Working on that 😉

bestander:
Thanks, @BYK, I'll cherry-pick it into 0.27 now

Issue #2930, opened 2017-03-16
support for ~ (home dir) in cafile (and other directives accepting a file path)


kevinv11n:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
~ is not properly evaluated when used in the `cafile` directive in .npmrc

**If the current behavior is a bug, please provide the steps to reproduce.**

Consider the scenario of a private npm registry that uses a self-signed SSL certificate. A user's ~/.npmrc would look like this:
```
registry=https://internal-registry/...
cafile=~/custom-certs.pem
```

Running `yarn install` results in this error message: `error Could not open cafile: ENOENT: no such file or directory, open '~/custom-certs.pem'`

If you change the cafile directive to `cafile=/users/kevin/custom-certs.pem` then `yarn install` works fine.

**What is the expected behavior?**
Use of ~ in cafile should expand to the user's home directory. It should do the same expansion in other directives that accept file paths.

**Please mention your node.js, yarn and operating system version.**
node: v6.2.0
yarn: v0.21.3

PR #3756, opened 2017-06-29
Support for ~ (home dir) in path configurations

vbfox:
**Summary**

This fixes the handling of path that are present in npm configuration (#2930). What's now handled :

* Paths starting with `~/` (Also `~\` on windows) get expanded
* Path that aren't rooted are considered relative to the current directory
* All path are fully resolved

All of this is from npm ( https://github.com/npm/npm/blob/latest/lib/config/core.js#L394 ).
I'm not sure that I like the fact that relative paths are relative to cwd instead of the configuration file where they are found but it's what's npm is doing 😉 

**Motivation**

My original motivation was to stop having errors when i'm installing `node-sass` on node 8. As versions of `node-sass` that aren't the latest don't include pre-built binaries for node 8 the installer try to use `node-gyp` to build a fresh copy.

This build would work (I have the necessary dev tools) except that `node-gyp` since recently read the `cafile` property and open it. As the cwd for install script is their package, it fail and break most yarn commands.

**Test plan**

I added some unit tests tests.

Also verified on the output of an npm module dumping it's environment in it's install script (during `yarn install --verbose`) :

2017-06-29 16_05_09-vs 2015

With `.npmrc`:

```
cafile=cafile
prefix=~/.npm-global
```

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -1,0 +1,41 @@
+/* @flow */
+
+import NpmRegistry from '../../src/registries/npm-registry.js';
+import {resolve} from 'path';
+import homeDir from '../../src/util/user-home-dir.js';
+
+describe('normalizeConfig', () => {
+  beforeAll(() => {
+    process.env.REPLACE = 'REPLACED';
+  });
+
+  afterAll(() => {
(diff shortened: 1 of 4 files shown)


arcanis:
Thanks!

PR #3759, opened 2017-06-29
Update: add regression tests for Git.spawn env issues

BYK:
**Summary**

Regression tests for #3742 and follow up to #3743.

**Test plan**

Tests shall pass.

### DIFF
--- a/src/util/__mocks__/child.js
+++ b/src/util/__mocks__/child.js
@@ -2,8 +2,6 @@
 
 const realChild = (require: any).requireActual('./child.js');
 
-realChild.spawn = jest.fn(() => {
-  return Promise.resolve('');
-});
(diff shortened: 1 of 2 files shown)


src/util/__mocks__/child.js
```diff
@@ -0,0 +1,9 @@
+/* @flow */
+
+const realChild = (require: any).requireActual('./child.js');
+
+realChild.spawn = jest.fn(() => {
```
cpojer: Why do you like braces and return statements so much? :P

bestander:
src/utils/__mocks__ gets compiled into lib on build, I wonder if it affects Yarn in any way

bestander:
I'll revert this one for today

PR #3760, opened 2017-06-29
Update: Upgrade Jest to version 20

### LINKED ISSUE #3757: Upgrade to Jest 20

BYK:
**Summary**

Fixes #3757. Jest 20 is the latest version with many improvements,
most importantly for us, not automatically polyfilling `Array.includes`.

**Test plan**

CI should pass without issues.

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -741,13 +741,13 @@ test.concurrent("doesn't warn when peer dependency is met during add", (): Promi
 
       expect(
         warnings.some(warning => {
-          return warning.data.toString().toLowerCase().includes('unmet peer');
+          return warning.data.toString().toLowerCase().indexOf('unmet peer') > -1;
         }),
       ).toEqual(false);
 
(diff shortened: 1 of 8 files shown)


cpojer:
I'm not gonna nitpick too much, because I just merged it, but I would assume that `!==` is faster than `>` in the general case. It may not matter much for most callsites, but then there might be one or two where it actually matters.

BYK:
I love nitpicky, especially around these things so I created a jsperf test: https://jsperf.com/indexof-1-eq-vs-gt

For my Firefox 55 and Chrome 59, it says there no meaningful differences. Mystery solved!

voxsim:
@BYK I think we should upgrade https://github.com/yarnpkg/yarn/blob/master/flow-typed/npm/jest_v19.x.x_with_custom_matchers.js as well to use the new features of jest ;)

BYK:
@voxsim should we create a new issue for this with more details?

PR #3768, opened 2017-06-30
Revert "Update: add regression tests for Git.spawn env issues"

bestander:
Reverts yarnpkg/yarn#3759, see comments for reasons

- mock leaks into source code when built
- jest complains about seeing the mock twice


bestander:
quick landing, cc @BYK

PR #3770, opened 2017-06-30
fixes #3767 - silencing Yarn when running in jest

### LINKED ISSUE #3767: Running jest via `yarn run` breaks jest
After updating to 0.27.3 Travis started failing tests https://travis-ci.org/yarnpkg/yarn/jobs/248542037.

This can be reproduced locally:

- install yarn@0.27.3
- yarn run jest

However running jest directly works

- ./node_modules/.bin/jest

bestander:
**Summary**
fixes #3767 which was caused by #3536

**to cherry-pick into 0.27-stable**

The reason why it fails - when you run `yarn run test-ci` current Yarn binary will set env var  `YARN_SILENT=1` and then invoke jest passing process.env.
Later jest would run some tests in Yarn repo that check for output of Yarn commands but they are silenced and fail.

Apparently it affects only tests that are invoked via `yarn run` command and that test yarn's output.
I suppose that only Yarn repository tests for Yarn output and is invoked via Yarn run command so I just override `YARN_SILENT` in the run commands that invoke jest.

Otherwise we might revert automatic silencing #3536 and educate users to silence manually.

The logic is this now

`yarn run` => sets YARN_SILENT = 1 => `finds command in package.json ` => overrides YARN_SILENT = 0 => e2e tests run not silenced.

**Test plan**

- yarn build
- `./bin/yarn run jest`

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,10 +95,11 @@
     "lint": "yarn run lint-prettier && eslint . && flow check",
     "lint-prettier": "node scripts/prettier.js lint",
     "prettier": "node scripts/prettier.js write",
+    "jest": "YARN_SILENT=0 node ./node_modules/.bin/jest",
     "release-branch": "./scripts/release-branch.sh",
     "test": "yarn lint && yarn test-only",
     "test-ci": "yarn build && yarn test-only",
-    "test-only": "node --max_old_space_size=4096 ./node_modules/.bin/jest --coverage --verbose",
+    "test-only": "node --max_old_space_size=4096 YARN_SILENT=0 ./node_modules/.bin/jest --coverage --verbose",
     "watch": "gulp watch"
   },
   "jest": {


package.json
```diff
@@ -95,10 +95,11 @@
     "lint": "yarn run lint-prettier && eslint . && flow check",
     "lint-prettier": "node scripts/prettier.js lint",
     "prettier": "node scripts/prettier.js write",
+    "jest": "YARN_SILENT=0 node ./node_modules/.bin/jest",
     "release-branch": "./scripts/release-branch.sh",
     "test": "yarn lint && yarn test-only",
     "test-ci": "yarn build && yarn test-only",
-    "test-only": "node --max_old_space_size=4096 ./node_modules/.bin/jest --coverage --verbose",
+    "test-only": "node --max_old_space_size=4096 YARN_SILENT=0 ./node_modules/.bin/jest --coverage --verbose",
```
BYK: I don't think this is the correct way to set it here. See Travis logs.

package.json
```diff
@@ -95,10 +95,11 @@
     "lint": "yarn run lint-prettier && eslint . && flow check",
     "lint-prettier": "node scripts/prettier.js lint",
     "prettier": "node scripts/prettier.js write",
+    "jest": "YARN_SILENT=0 node ./node_modules/.bin/jest",
```
BYK: This wouldn't work on Windows :(

bestander:
Ok, deferring to you and Mael :)

package.json
arcanis: The fix is pretty minimal, it would only require to use [cross-event](https://yarnpkg.com/en/packages?q=cross-env&p=1)

bestander:
deprecated

PR #3771, opened 2017-06-30
Fix: fix e2e yarn output tests after #3536

BYK:
**Summary**

Fixes #3536. Explicitly and programmatically passes `YARN_SILENT=0`
to test runs that checks yarn's output.

**Test plan**

All CI tests and local test runs with yarn 0.27.3 should pass.

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -30,19 +30,29 @@ async function execCommand(
   }
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


arcanis:
Oh, an alternate PR is #3770

arcanis:
Hm, this PR will not fix the tests that run from within the jest process (only those that are spawned later). I don't think that's a problem right now, but maybe that's something to keep in mind, since it could cause unexpected issues later on.

bestander:
ok, this is better than #3770

Issue #3769, opened 2017-06-30
Follow up on release 0.27.3


bestander:
- [x] #3759 was reverted because of jest complains, need to reland it with the fixes
- [x] We don't have tests that cover how current Yarn will run Yarn's jest tests https://github.com/yarnpkg/yarn/issues/3767. Circle CI could compile Yarn and use it to drive `test-ci` command
- [ ] We don't cover private packages installation e2e https://github.com/yarnpkg/yarn/issues/3765

Daniel15:
> We don't have tests that cover how current Yarn will run Yarn's jest tests [#3767](https://github.com/yarnpkg/yarn/issues/3767). Circle CI could compile Yarn and use it to drive `test-ci` command

We could make CircleCI use the latest nightly build of Yarn, rather than the current release build. This would help us more easily catch issues with Yarn itself. That's a one-line change in `scripts/bootstrap-env-ubuntu.sh`:
```
-echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+echo "deb http://nightly.yarnpkg.com/debian/ nightly main" | sudo tee /etc/apt/sources.list.d/yarn.list
```

Or the RC instead.

Having said that, that also means that a broken version of Yarn would also break the build, and the build won't be fixed until the fix is committed 😕

bestander:
I think nightly on Circle and stable on Travis could be a good mix.

Daniel15:
Given CircleCI is the backbone of our release infra, it might be safer to use nightly on Travis 😛

bestander:
True!

bestander:
As a matter of fact Circle is using yarn v0.21.3 and is green.
We might configure one matrix rule on Travis to run on nightly (or compiled current)

arcanis:
Re: private packages; I've setup a private package at @arcanis/private-pkg for debugging and fixing the issue, but it would be better to have something on the @yarnpkg organization.

BYK:
@bestander I did the second item on Travis CI instead since it was easier: https://github.com/yarnpkg/yarn/pull/3805

BYK:
I've filed #4186 for the last checkbox and am going to close this ticket in favor of the more granular issues.

PR #3774, opened 2017-06-30
Fixes npm auth

### LINKED ISSUE #3765: yarn@0.27.3 consistently 404s on private packages

bug

yarn@0.27.3 fails to install private packages on every run of `yarn` across several different projects/lockfiles, with a 404.

eg
screen shot 2017-06-29 at 3 55 41 pm

presumably, run `yarn` with `yarn@0.27.3` installed against any `yarn.lock` that includes private packages.

`yarn@0.27.3` installs private packages.

node: 6.9.1
yarn: 0.27.3
OS: macOS@10.12.4, ubuntu 14.04.5 LTS

arcanis:
This commit should fix #3765. I'm not completely satisfied with hardcoding our registry, maybe that could extended at a later point with an option to "inherit" the auth token from npm. Not sure it would be useful, tho.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -5,6 +5,7 @@ import type RequestManager from '../util/request-manager.js';
 import type {RegistryRequestOptions, CheckOutdatedReturn} from './base-registry.js';
 import type Config from '../config.js';
 import type {ConfigRegistries} from './index.js';
+import {YARN_REGISTRY} from '../constants.js';
 import * as fs from '../util/fs.js';
 import NpmResolver from '../resolvers/registries/npm-resolver.js';
 import envReplace from '../util/env-replace.js';
@@ -215,8 +216,9 @@ export default class NpmRegistry extends Registry {
     const baseRegistry = this.getRegistry(packageName);
     const registries = [baseRegistry];
 
-    if (baseRegistry === `https://registry.yarnpkg.com/`) {
-      registries.push(`https://registry.npmjs.org/`);
+    // If sending a request to the Yarn registry, we must also send it the auth token for the npm registry
+    if (baseRegistry === YARN_REGISTRY) {
+      registries.push(DEFAULT_REGISTRY);
     }
 
     for (const registry of registries) {
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -87,7 +87,8 @@ export default class NpmRegistry extends Registry {
     const packageIdent = packageName || pathname;
     const isScoppedPackage = packageIdent.match(/^@|\/@/);
 
-    if (alwaysAuth || isScoppedPackage) {
+    // this.token must be checked to account for publish requests on non-scopped packages
+    if (this.token || alwaysAuth || isScoppedPackage) {
       const authorization = this.getAuth(packageName || pathname);
       if (authorization) {
         headers.authorization = authorization;
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -76,7 +76,6 @@ export default class NpmRegistry extends Registry {
     const registry = this.getRegistry(packageName || pathname);
     const requestUrl = url.resolve(registry, pathname);
     const alwaysAuth = this.getRegistryOrGlobalOption(registry, 'always-auth');
-    const customHostSuffix = this.getRegistryOrGlobalOption(registry, 'custom-host-suffix');
 
     const headers = Object.assign(
       {
@@ -84,7 +83,8 @@ export default class NpmRegistry extends Registry {
       },
       opts.headers,
     );
-    if (this.token || (alwaysAuth && isRequestToRegistry(requestUrl, registry, customHostSuffix))) {
+
+    if (alwaysAuth || (packageName || pathname)[0] === `@`) {
       const authorization = this.getAuth(packageName || pathname);
       if (authorization) {
         headers.authorization = authorization;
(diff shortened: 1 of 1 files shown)


bestander:
We definitely should extract the npmrc/yarnrc config parsing from npm-registry and yarn-registry.
There is a lot of confusion that those classes though in inheritance relationship read the config files exclusively.

bestander:
@arcanis, how did you test the PR?
I suppose e2e test will be coming within the follow-up task https://github.com/yarnpkg/yarn/issues/3769, right?

src/registries/npm-registry.js
```diff
@@ -208,26 +208,33 @@ export default class NpmRegistry extends Registry {
...
 
-    const registry = this.getRegistry(packageName);
+    const baseRegistry = this.getRegistry(packageName);
+    const registries = [baseRegistry];
 
-    // Check for bearer token.
-    const authToken = this.getRegistryOrGlobalOption(registry, '_authToken');
-    if (authToken) {
-      return `Bearer ${String(authToken)}`;
+    if (baseRegistry === `https://registry.yarnpkg.com/`) {
```
bestander: we have constants for these values

src/registries/npm-registry.js
```diff
@@ -76,15 +76,15 @@ export default class NpmRegistry extends Registry {
...
 
     const headers = Object.assign(
       {
         Accept: 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
       },
       opts.headers,
     );
-    if (this.token || (alwaysAuth && isRequestToRegistry(requestUrl, registry, customHostSuffix))) {
+
+    if (alwaysAuth || (packageName || pathname)[0] === `@`) {
```
bestander: I don't understand the reason for this change

src/registries/npm-registry.js
bestander: So if package comes from `registry.yarnpkg.com` then you want to add `registry.npmjs.org` to the list of registries in the attempt of getting username and password.

That looks like a lazy patch, it would be better to set up `getRegistryOrGlobalOption` to return same data for yarnpkg and npmjs rather than patch every callsite

src/registries/npm-registry.js
bestander: From @arcanis:

this.token is undefined
So the codepath is not executed if alwaysAuth is not set
The condition I put tries to set an auth token if alwaysAuth is enabled, or if the package is scoped

BYK: This should go into the code as a comment. Also a helper function named `isScoped` or `needsAuth` (or both?) would make the code more readable.

BYK: Also I think the `this.token` check here was to memoize the token. We may wanna keep that behavior.

src/registries/npm-registry.js
```diff
@@ -208,26 +208,33 @@ export default class NpmRegistry extends Registry {
...
+    if (baseRegistry === `https://registry.yarnpkg.com/`) {
+      registries.push(`https://registry.npmjs.org/`);
     }
 
-    // Check for basic auth token.
-    const auth = this.getRegistryOrGlobalOption(registry, '_auth');
-    if (auth) {
-      return `Basic ${String(auth)}`;
-    }
+    for (const registry of registries) {
```
BYK: Is this trying to get an auth token for the first matching registry? If so, I find that a bit dangerous. Should we keep a mapping of `registry: token` pairs and use the appropriate one when communicating. This would also leak tokens to other registries which may be an important security threat.

arcanis: Hence the check to make sure that the npm fallback is only added when the registry is the Yarn registry. The token won't be sent for any other hostname (wich is another issue).

That being said, the whole "multi-registry" logic is flawed, since we only support a single registry implementation (and adding more of them wouldn't make much sense, since it would complexify the codebase for little gain). I'd like to rework it so that we only support a single registry implementation, the npm one, and then make possible to configure what needs to be generic (mostly the hostname). But that's a second step.

BYK: I'm +1 to this until we do the refactor. The code can use a short comment explaining this safe-guard.

BYK:
Sup @arcanis? :)

src/registries/npm-registry.js
```diff
@@ -76,15 +77,19 @@ export default class NpmRegistry extends Registry {
...
     const headers = Object.assign(
       {
         Accept: 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
       },
       opts.headers,
     );
-    if (this.token || (alwaysAuth && isRequestToRegistry(requestUrl, registry, customHostSuffix))) {
+
+    const packageIdent = packageName || pathname;
+    const isScoppedPackage = packageIdent.match(/^@|\/@/);
```
BYK: What does the second part cover? (sorry, not very familiar with scoped packages except for the `@user/package` pattern.

src/registries/npm-registry.js
```diff
@@ -208,26 +213,34 @@ export default class NpmRegistry extends Registry {
...
+      if (auth) {
+        return `Basic ${String(auth)}`;
+      }
+
+      // Check for basic username/password auth.
+      const username = this.getRegistryOrGlobalOption(registry, 'username');
+      const password = this.getRegistryOrGlobalOption(registry, '_password');
+      if (username && password) {
+        const pw = new Buffer(String(password), 'base64').toString();
+        return 'Basic ' + new Buffer(String(username) + ':' + pw).toString('base64');
```
BYK: Wow, no idea we did double base64 for the password part.

src/registries/npm-registry.js
arcanis: In some case `packageIdent` will be `@user/package`, and in other cases it will be `https://example.org/@user/package.tgz`

arcanis:
The main issue with e2e test is that it would require a private package somewhere. I have my own test package on my account, but I obviously cannot use it in the Yarn testsuite 😄 

Ideally, we should create a new private package on the `yarnpkg` organization (we might need to upgrade the plan, cc @bestander, do you know who holds the keys of this account?), create a new `yarnpkg-test` user, and give to this user the read-only rights to access the test package ([npm access](https://docs.npmjs.com/cli/access)). From there we should be able to safely put the auth token in our tests (unfortunately it will still allow malicious users to publish under this user namespace, but that shouldn't make much difference since they won't have push access to any existing project).

BYK:
>That, or we build a stub registry server.

This. Or even better: we just do integration tests via mocks.

bestander:
I think we should have an end to end installation without stubs at least for one package.
It should take a few minutes to register @yarnpkg test org and create a package there.
The login won't be secure but what are the risks, malicious parties will be able to install that package?

BYK:
@bestander I'll merge this then since we have filed #3842

daniel-nagy:
This has broken installing private modules from our company's Bitbucket server.

```JSON
"private-module": "git+ssh://git@git:7999/project/private-module.git"
```

I rolled back to version `0.27.4` and it correctly installs the package.

bestander:
@daniel-nagy please raise a new issue.
Can you provide an example URL publicly for us to test?

BYK:
@daniel-nagy @bestander #3907

PR #3782, opened 2017-06-30
Chore: test yarn with yarn on CircleCI

BYK:
**Summary**

Refs #3769. Tries to use the built yarn to test yarn itself.

**Test plan**

Tests pass on CircleCI.

catception


BYK:
Forget it, CircleCI is not built for this.

Issue #3437, opened 2017-05-17
yarn issues warnings when cordova plugins are defined in `package.json`


fiznool:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

The latest version of the cordova CLI (7.0.0) adds cordova plugins to `package.json` when running `cordova plugin add <plugin-name>`. The spec for Cordova plugins [states that you can manage cordova version dependencies using the `engines.cordovaDependencies` property](https://cordova.apache.org/docs/en/latest/guide/hybrid/plugins/#specifying-cordova-dependencies). Unfortunately, yarn does not like this non-standard definition property and issues a warning when running `yarn` with the freshly included cordova plugins in `package.json`.

The warning resembles:
```
warning cordova-plugin-camera@2.4.0: The engine "cordovaDependencies" appears to be invalid.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

You don't need to use the Cordova CLI to replicate the bug. The following is enough:

``` bash
mkdir yarn-cordova && cd $_
yarn init -y
yarn add cordova-plugin-camera
```

**What is the expected behavior?**

A warning not to be issued.

**Please mention your node.js, yarn and operating system version.**

- node v6.10.3
- yarn v0.24.4
- macOS Sierra 10.12.4

bestander:
@fiznool I suppose you can send a PR to add cordova to exclusions https://github.com/yarnpkg/yarn/blob/master/src/package-compatibility.js#L48.
A slightly better approach could be to make this field configurable.

edwardpayton:
+1

PR #3786, opened 2017-07-01
Adds compatibility for cordovaDependencies engine

bytenik:
See https://cordova.apache.org/docs/en/latest/guide/hybrid/plugins/#specifying-cordova-dependencies for a description of when the `cordovaDependencies` engine is used.

Without this change, freshly installed Cordova plugins generate errors on yarn add. This was mentioned in #3437 a while back but no one ever made a pull request to resolve it.


BYK:
@bestander @arcanis may be we should extend this `ignore` list via an environment variable so we don't have to keep this manual list in the code?

arcanis:
I'm fine with the manual list in this particular case (I fear allowing to configure this field would trigger non-obvious issues where a package would work on some computers but not others because an `ignore` key hasn't been set), but if we are to make this configurable, it should probably be through the use of the yarnrc rather than an environment variable.

BYK:
>but if we are to make this configurable, it should probably be through the use of the yarnrc rather than an environment variable.

Great idea, yeah.

bytenik:
Prettier doesn't seem to do anything.

```C:\Users\bytenik\AppData\Local\Temp\yarn>yarn prettier
yarn prettier v0.27.5
$ node scripts/prettier.js write
Done in 7.03s.

C:\Users\bytenik\AppData\Local\Temp\yarn>git status
On branch patch-1
Your branch is up-to-date with 'origin/patch-1'.
nothing to commit, working tree clean

C:\Users\bytenik\AppData\Local\Temp\yarn>
```

BYK:
@bytenik oh interesting. Then would you mind rebasing onto master or merging from it to see if it would clear the CI failures?

anpage:
The linter is failing because the line is too long. The comment needs to be shortened, which could be why `yarn prettier` couldn't automatically fix it.

BYK:
@bytenik you may also us a link shortener like bit.ly

bytenik:
I took the bit.ly approach.

PR #3789, opened 2017-07-01
forward SIGTERM to spawned child processes

### LINKED ISSUE #3424: 'yarn run' does not forward SIGTERM to process
When running an application via 'yarn run' and a SIGTERM is signaled on it, yarn does not forward the SIGTERM to its child process(es). Hence an application's graceful shutdown code cannot run.

GAumala:
**Summary**

If yarn receives a SIGTERM signal, it forwards it to all the child processes that it has spawned. This fixes #3424.

module `src/util/child.js` now keeps a reference to every spawned child process until the child exits. it also exports a new function `forwardSignalToSpawnedProcesses()` which sends a signal to all currently held child processes.

new module `src/util/signal-handler.js` exports a single function `handleSignals()` which attaches a SIGTERM listener to the global `process` object that calls `forwardSignalToSpawnedProcesses()`

Currently the `handleSignals()` function is called in `src/cli/index.js`, please let me know if there is a better place for this.

**Test Plan**

Write a script with a long running operation and a SIGTERM handler:

```javascript
//longOperation.js
process.on('SIGTERM', () => {
  console.log('SIGTERM');
  process.exit(1)
});

setInterval(() => {
  console.log('tick')
}, 1000)
```
Add it to a package json as an executable script
```json
+ "longOperation": "node longOperation.js"
```
Run it wth yarn
```
yarn run longOperation
```
In a new terminal run `ps -axu | grep yarn` to find the yarn prcess id. You'll get something like this:

```
gabriel  11880  2.6  0.4 1107924 53612 pts/6   Sl+  05:32   0:00 node bin/yarn.js run longOperation
gabriel  11897  0.0  0.0  13020  2312 pts/4    S+   05:32   0:00 grep yarn
```
Run kill with the yarn process id
```
kill 11880
```

Yarn should be terminated and the `longOperation.js` script should be able to print `SIGTERM` to console before it exits.

### DIFF
--- a/__tests__/util/child.js
+++ b/__tests__/util/child.js
@@ -1,0 +1,74 @@
+/* @flow */
+import {spawn, forwardSignalToSpawnedProcesses} from '../../src/util/child.js';
+
+let mockSpawnedChildren = [];
+
+jest.mock('child_process', () => {
+  class MockedChildProcess {
+    stdout: Object;
+    stderr: Object;
+    receivedSignal: string;
+    exitWithCode: number => void;
+
+    constructor() {
+      this.stdout = {
+        on: () => {},
+      };
+      this.stderr = {
+        on: () => {},
+      };
+    }
+
+    kill(signal) {
+      this.receivedSignal = signal;
+    }
+
+    on(event, handler) {
+      if (event === 'close') {
+        this.exitWithCode = handler;
+      }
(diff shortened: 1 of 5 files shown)


bestander:
Thanks, @GAumala, nice work!

Issue #3696, opened 2017-06-22
Add a `yarn browse pkg` command


Haroenv:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

To go to a package detail page, you’d have to manually enter `open https://yarnpkg.com/en/package/whatever` instead of being able to do `yarn browse whatever`

**What is the expected behavior?**

`yarn browse whatever` opens <https://yarnpkg.com/en/package/whatever>  in your default browser.

bestander:
This feature can be done by a standalone tool.
I think Yarn should focus on package management.

Haroenv:
fish: 

```fish
function yarn-browse
  open https://yarnpkg.com/en/package/$argv[1]
end
```

bash: 

```bash
function yarn-browse { 
  open https://yarnpkg.com/en/package/$1
}
```

You can add these to you respective .profile

(for future reference)

Daniel15:
For what it's worth, I think this would be a useful feature.

I registered a short domain for these - https://yarn.fyi/ will redirect (eg. https://yarn.fyi/react)

neoziro:
This package does the job: https://github.com/sindresorhus/npm-home

Daniel15:
@neoziro  - That will open the npm detail page, not the Yarn one.

neoziro:
https://github.com/sindresorhus/npm-home/pull/3

PR #3790, opened 2017-07-01
Use babel-polyfill and babel-preset-env

voxsim:
**Summary**
In this pull request:
- I reviewed all the presets, plugin, etc.. related to babel
- Upgraded babel-jest
- Use babel-polyfill to fix https://github.com/yarnpkg/yarn/issues/3693 once for all (I hope 😱 )
- Use babel-preset-env instead of babel-preset-node5 and babel-preset-es2015-node4 to be futureproof
- Add some plugins that babel-preset-env doesn't have by default
- Rename babel env 'pre-node5' in 'node4'

**Test plan**
Tests in place

I am a babel newbie, if anyone found something erronous please let me know :D it took me many days to understand each part and how it works ;)


voxsim:
For some reasons, in node4 with babel-polyfill the errors behave differently and put some additional slashes:

```
    Expected string:
      "Command failed: node \"/home/travis/build/yarnpkg/yarn/bin/yarn.js\" unknown 
    error An unexpected error occurred: \"Command \\\"unknown\\\" not found.\".
    "
    To contain value:
      "Command \"unknown\" not found"
```

Volune:
That looks great. 👍 

A few questions:
- Can we move `import 'babel-polyfill';` into a separate file, to ensure it's imported only once (as recommended in the doc)?
- Can we put the `node6` babel configuration as default configuration?
- Can you also revert #3592 and #3654, to ensure `.includes` will work as expected?
- I like to see you replacing `stage-0` preset with specific plugins, but what was the politic about `stage-0` from Yarn developers?

.babelrc
```diff
@@ -1,24 +1,47 @@
...
+        ["transform-inline-imports-commonjs"]
+      ],
+      "presets": [["env", {
+        "targets": {
+          "node": 4
+        },
+        "useBuiltIns": true
+      }]]
+    },
+    "node6": {
```
kaylieEB: to what @volune said, maybe this can be "development", which babel uses as default.

scripts/build-webpack.js
```diff
@@ -58,7 +59,7 @@ const compilerLegacy = webpack({
         test: /\.js$/,
         exclude: /node_modules/,
         loader: 'babel-loader',
-        query: babelRc.env['pre-node5'],
+        query: babelRc.env.node4,
```
voxsim: Is it ok if we open another issue/pull request on that? This pr is becoming too big imho.

voxsim:
@Volune some answers:

*   Can we move `import 'babel-polyfill';` into a separate file, to ensure it's imported only once (as recommended in the doc)?

The file src/cli/index.js is always required one time, I included in that file because otherwise babel-preset-env does not do its magic (I mean replacing the require('babel-polyfill') with all require that really needs in base on the platform).

*   Can we put the `node6` babel configuration as default configuration?

Yes, we can, I will do asap

*   Can you also revert [#3592](https://github.com/yarnpkg/yarn/pull/3592) and [#3654](https://github.com/yarnpkg/yarn/pull/3654), to ensure `.includes` will work as expected?

I did it

*   I like to see you replacing `stage-0` preset with specific plugins, but what was the politic about `stage-0` from Yarn developers?

I think we should ask this question to @bestander or maybe @kittens, I can always revert and use the stage-0 preset, It's not a problem at all :D

scripts/build-webpack.js
kaylieEB: yup, thanks!

voxsim: @kaylieEB can you open the issue? Because I don't fully understand your comment, I am afraid to write something wrong

kaylieEB: np :) https://github.com/yarnpkg/yarn/issues/3809

voxsim: thanks :D

bestander:
Looping in @BYK.

bestander:
One thing to test - what is the impact on the bundle size?

voxsim:
That's a good answer, I will check it O.o @bestander do you think we can integrate https://github.com/siddharthkp/bundlesize in yarn?

bestander:
> do you think we can integrate https://github.com/siddharthkp/bundlesize in yarn?

That looks cool and useful.
My first impression - "YES!!!" but then I start thinking about the support burden and this thread comes to mind https://github.com/yarnpkg/yarn/issues/3696.
We can add so many things to CI and make it solve a wide range of issues but the wider it gets the harder it will be for people to discover the commands that solve really complex problems.


I would separate Yarn as a project into two areas

1. a package manager CLI and everything about it - keeping yarn.lock consistent, dealing with version conflicts and node_modules size and support for different runtime environments

2. a community project that makes package management easier - yarnpkg.com website, its blog, yarn.fyi website (yarn.fyi/react), yarn create command, github issues and RFCs

The latter is non homogenous and split across domains and tools.
Maybe we need to have a buddy tool that comes with Yarn and addresses the problem in a consistent way?

Daniel15:
@voxsim - The effect on the bundle size looks way too big, unfortunately. Comparing the yarn-legacy JS file, it's 4.23 MB for this build (https://4316-49970642-gh.circle-artifacts.com/0/home/ubuntu/yarn/artifacts/yarn-legacy-0.27.0.js) vs 3.58 MB for the latest nightly build (https://nightly.yarnpkg.com/yarn-legacy-0.27.0-20170703.1704.js).

src/util/execute-lifecycle-script.js
```diff
@@ -145,7 +145,7 @@ export async function executeLifecycleScript(
   // Add global bin folder if it is not present already, as some packages depend
   // on a globally-installed version of node-gyp.
   const globalBin = getGlobalBinFolder(config, {});
-  if (pathParts.indexOf(globalBin) === -1) {
+  if (!pathParts.includes(globalBin)) {
```
Daniel15: What's the advantage of this change?

Daniel15:
I've tried using `babel-preset-env` in the past, and was too scared to commit it since the generated JavaScript differed. For example, it started converting arrow functions to regular functions, due to the fact that Node.js 4.x has some very minor bugs around precedence (http://node.green/#ES2015-functions-arrow-functions-correct-precedence) that don't really affect us.

Please perform a build with the old Babel config, and then another build with the new Babel config, and diff the files to see what the differences are.

Daniel15:
We can probably just polyfill `includes` with something like:
```js
if (!String.prototype.includes) {
  String.prototype.includes = function(str, start) {
    return this.indexOf(str, start) > -1;
  }
}
```

voxsim:
@bestander I totally agree with you, btw I was talking on integrate bundle size in our travis/circleci build to check on every pr the bundlesize of yarn node6/node4 build (i don't like call it normal or legacy becauae with node8 it does not make sense). I can open a pr but I don't have the permissions to do anything on Travis or CircleCI.

@Daniel15 thanks for your feedbacks and I agree with you about the polyfill, I think you should comment on https://github.com/yarnpkg/yarn/issues/3693 and decide with us which are our next steps :)

@BYK thanks for your review, I agree with you that this PR is too big. Do you prefer one pr for each commit or can i open a pull request with:
- remove babylon
-upgrade babel-core
- remove babel-plugin-symbol
- update babel-jest
After this pr we will discuss our next steps  ;)

bestander:
> I totally agree with you, btw I was talking on integrate bundle size in our travis/circleci build to check on every pr the bundlesize of yarn node6/node4 build (i don't like call it normal or legacy because with node8 it does not make sense). I can open a pr but I don't have the permissions to do anything on Travis or CircleCI.

We can sort out the permission issue.
Thanks for looking into this.

Daniel15:
> Do you prefer one pr for each commit or can i open a pull request with:

One PR per commit please 😃

PR #3794, opened 2017-07-02
Upgrade eslint to 4.1.1

voxsim:
**Summary**
Upgrade eslint to 4.1.1 and other related plugins.

Some info:
- I do not understand why we need 'eslint-plugin-react' in eslint configuration, can someone explain this to me?
- I removed eslint-plugin-babel and babel-eslint that they were not used
- I turned off space-before-function-paren for https://github.com/prettier/prettier/issues/1139

**Test plan**
Tests in place.


voxsim:
For now we can't upgrade eslint for the dependencies, I will close it :P

PR #3796, opened 2017-07-02
Use eslint-plugin-prettier instead of custom scripts for prettier

voxsim:
**Summary**
Use eslint-plugin-prettier instead of custom scripts for prettier. I would thank @RomanovRoman for let me know. Maybe @cpojer you can do the same in jest ;)

**Test plan**
Tests in place


.eslintrc
```diff
@@ -16,6 +17,13 @@
...
+    "sort-keys": "off",
+    "prettier/prettier": ["error", {
+      "singleQuote": true,
+      "trailingComma": "all",
+      "bracketSpacing": false,
+      "printWidth": 120,
+      "parser": "flow"
+    }]
   }
 }
```
kaylieEB: What do you think about adding https://github.com/prettier/eslint-config-prettier to avoid future conflicts in eslint rules?

package.json
```diff
@@ -92,9 +93,7 @@
     "build-dist": "bash ./scripts/build-dist.sh",
     "build-win-installer": "scripts\\build-windows-installer.bat",
     "check-lockfile": "./scripts/check-lockfile.sh",
-    "lint": "yarn run lint-prettier && eslint . && flow check",
-    "lint-prettier": "node scripts/prettier.js lint",
-    "prettier": "node scripts/prettier.js write",
```
kaylieEB: So has the script been deleted? I don't see it in the changes. And if you're removing the write script, then prettier will no longer auto-magically format the code and only warn. Is that the intention? If not, I'd add a script like `lint-fix` and pass in the `eslint --fix` command.

voxsim: You are right I forgot to delete the script, i will do asap. Yes, it is that the intetion, so everything will be like before I added prettiee :)

.eslintrc
voxsim: Yeah I think it is better :)

voxsim: @kaylieEB I tried the cli tool https://github.com/prettier/eslint-config-prettier#cli-helper-tool but there are some problems: every yarn command put `yarn run v<version>` on top of the result of each command.. when I try to pipe with `eslint-config-prettier-check` it does not work.
Is it ok for you if we open an issue and we work on that in another pull request?

kaylieEB: sounds good to me!

package.json
kaylieEB: Thanks! So it wouldn't be exactly the same since we used to make people run `yarn run prettier` in the script to auto-fix the issues, but I'm assuming someone will run eslint --fix once in a while then.

voxsim:
if @kaylieEB agreed, I think @bestander or @arcanis can review and merge it in case ;)

Btw I opened this issue https://github.com/yarnpkg/yarn/issues/3808

bestander:
LGTM, please rebase

voxsim:
@bestander done!

bestander:
Nice work!

BYK:
What's the "right" way to run `prettier` in write-mode after this patch? I was relying on `yarn prettier` to format things for me automatically and since we don't use `eslint` in `--fix` mode either I'm not sure what should I do.

arcanis:
`yarn run eslint --fix`

arcanis:
I opened a PR to add the script back, it might make it easier for external contributors to step into the project :)

Issue #3710, opened 2017-06-23
Peer dependency resolved from sibling, not root


alunny:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

As far as I can tell, yarn satisfies peer dependencies by finding the closest match anywhere in the tree. This is unexpected when the top-level project depends on module A at version x, a dependency B depends on module A at version y, and another dependency C peer depends on module A for some version that y satisfies that x does not.

I would expect dependency C to use A@x (what's specified in the top-level package.json file), instead of copying A@y from dependency B.

**If the current behavior is a bug, please provide the steps to reproduce.**

repro repo is here: https://github.com/alunny/yarn-peerdeps-example

to repro from scratch:
``` shell
$ yarn init
$ yarn add -D webpack@3.0.0
$ yarn add -D parallel-webpack@1.6.1 # has dependency "webpack": "^1.12.9"
$ yarn add -D karma-webpack@2.0.3 # has peer dependency "webpack": "^1.1.0 || ^2 || ^2.1.0-beta.0 || ^2.2.0-rc.0"
$ grep version node_modules/karma-webpack/node_modules/webpack/package.json
  "version": "1.15.0",
$ yarn why node_modules/karma-webpack/node_modules/webpack/
yarn why v0.23.3
[1/4] 🤔  Why do we have the module "node_modules/karma-webpack/node_modules/webpack/"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
✨  Done in 0.37s.
```

**What is the expected behavior?**

Expected karma-webpack to not have its own version of webpack, but use the top-level copy (3.0.0), perhaps with a warning that the peer dependency is not satisfied.

Failing that, perhaps allow `yarn install --flat` for specific packages (webpack in this case) rather than needing to resolve every duplicate module in the project.

**Please mention your node.js, yarn and operating system version.**

node v6.10.2
yarn 0.23.3
macOS Sierra 10.12.5

bestander:
Thanks for reporting.

This looks crazy, node_modules/karma-webpack/node_modules/webpack/ should not be installed at all and Yarn should display a warning that webpack peer dependency is not satisfied.

PR #3803, opened 2017-07-03
Fix:  Peer dependencies should only be looked up from top level dependencies

kaylieEB:
**Summary**
This is a bug fix for https://github.com/yarnpkg/yarn/issues/3710

The previous implementation of satisfying peer dependencies of an added dependency looked through all the existing patterns of a dependency (including nested dependencies of other packages) and returned the first match, which ended up copying the nested modules of the other dependency and swallowing warnings about unmet/incorrect peer dependencies. This PR checks for whether the package is at root level or not and otherwise throws a warning if it doesn't meet the semver qualifications.

**Test plan**
Added tests, and tested on CLI that the peer dependencies don't get installed.

**BEFORE**
screen shot 2017-07-02 at 5 38 28 pm

**AFTER**
screen shot 2017-07-02 at 5 39 36 pm


yarn.lock
```diff
@@ -1597,10 +1597,6 @@ eslint-config-fbjs@^2.0.0-alpha.1:
   version "2.0.0-alpha.1"
   resolved "https://registry.yarnpkg.com/eslint-config-fbjs/-/eslint-config-fbjs-2.0.0-alpha.1.tgz#addda12e1e68cec9271436df01ddcfe67bd1e88d"
 
-eslint-plugin-babel@^3.3.0:
```
voxsim: I think you should revert this upgrade infact `yarn check-lockfile` fails.

voxsim:
I deleted my comment because I was wrong ;)

src/package-linker.js
```diff
@@ -368,16 +368,20 @@ export default class PackageLinker {
 
     for (const name in peerDeps) {
       const range = peerDeps[name];
-      const patterns = this.resolver.patternsByPackage[name] || [];
-      const foundPattern = patterns.find(pattern => {
-        const resolvedPattern = this.resolver.getResolvedPattern(pattern);
-        return resolvedPattern ? this._satisfiesPeerDependency(range, resolvedPattern.version) : false;
+      const pkgs = this.resolver.getAllInfoForPackageName(name);
+      let foundPattern;
+      pkgs.find(pkg => {
```
BYK: I think what you intend to do is this:
```
const found = pkgs.find(pkg => {
    const {root, version, patterns} = pkg._reference || {};
    return root && this._satisfiesPeerDependency(range, version) && patterns;
});

if (found) {
    ref.addDependencies(found._reference.patterns);
}
```

Modifying the outside scope in iteration callbacks are a bit dangerous and quite confusing to the reader :(

BYK: You can probably use `invariant()` if nothing helps with that warning. That said, with the way I suggested, it shouldn't be a problem.

src/package-linker.js
```diff
@@ -409,16 +409,17 @@ export default class PackageLinker {
...
-      const patterns = this.resolver.patternsByPackage[name] || [];
-      const foundPattern = patterns.find(pattern => {
-        const resolvedPattern = this.resolver.getResolvedPattern(pattern);
-        return resolvedPattern ? this._satisfiesPeerDependency(range, resolvedPattern.version) : false;
+      const pkgs = this.resolver.getAllInfoForPackageName(name);
+      const found = pkgs.find(pkg => {
+        const {root, version} = pkg._reference || {};
+        return root && this._satisfiesPeerDependency(range, version);
       });
+      const foundPattern = found && found._reference && found._reference.patterns;
```
BYK: Not sure if this is you or flow yelling at you but you don't need to check `found._reference` since it is required for `.find()` to return something for `found`. Actually you can even move `patterns` up there and just check `found` as I suggested earlier. I'm guessing you did it this way because of `flow` complaining tho :D

kaylieEB: yup exactly, made flow unhappy. thanks!

Issue #3306, opened 2017-05-02
Removing a git package from package.json and running `yarn` does not update yarn.lock


jbinto:
**Do you want to request a *feature* or report a *bug*?**
Bug report.

**What is the current behavior?**

For a regular package, things work fine:

```
yarn add express
# yarn.lock is edited to add express.

# Now edit package.json and remove express.
vi package.json

# Run `yarn` again.
yarn

# yarn.lock has been changed to remove express, as expected
```

But for git packages, this doesn't work:

```
yarn add git+https://github.com/jbinto/yup
# yarn.lock is edited to add the yup git repo.

# Now edit package.json and remove yup.
vi package.json

# Run `yarn` again.
yarn

# yarn.lock has NOT been changed to remove yup (not expected)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

See above. I also have a small repro repo, just read through the commits, it's fairly apparent what's happening:

https://github.com/jbinto/yarn-repro/commits/master

**What is the expected behavior?**

Removing a `git` package from `package.json` and then running `yarn` should update the `yarn.lock`.

To work around this I'd have to either:
a) update `yarn.lock` by hand (which is not advised),
b) delete `yarn.lock` and regenerate, but then lose track of which dependencies I'm currently locked at

**Please mention your node.js, yarn and operating system version.**

yarn 0.23.4, node v7.9.0, OS X 10.12.4

bestander:
Thanks for repro steps, I confirm it happens

aracarie:
I'm actually having this issue for regular packages too - Yarn 0.24.6 / Node.js v6.10.3.

I can work on this if you'd like?

bestander:
Of course, @aracarie, please do

PR #3811, opened 2017-07-03
Bugfix 3306: Check top level patterns during integrity check

aracarie:
**Summary**
Fixes [issue 3306](https://github.com/yarnpkg/yarn/issues/3306). Previously changes to the package.json dependencies would not throw an error during integrity checks, unless there were zero dependencies, in which case an empty yarn.lock would be written.

A condition was added to the integrity checker for top level patterns. If the top level patterns are not the same, the install task will continue running with the force flag set, as would be done with a reinstall after a package was removed.

`yarn check --integrity` is also affected by this, and will return an error if the top level patterns don't match those specified in the package.json.

**Test plan**
Before:
```
yarn add foo left-pad
# package.json is edited to remove left-pad
yarn
# left-pad is still in yarn.lock
# package.json is edited to remove foo
yarn
# yarn.lock is empty
```

After:
```
yarn add foo left-pad
# package.json is edited to remove left-pad
yarn
# yarn.lock only contains foo
# package.json is edited to remove foo
yarn
# yarn.lock is empty
```


src/cli/commands/install.js
```diff
@@ -359,6 +359,13 @@ export class Install {
       return true;
     }
 
+    if (!match.integrityMatches && match.integrityError == 'PATTERNS_DONT_MATCH') {
+      this.reporter.warn(this.reporter.lang('integrityPatternsDontMatch'));
+      this.reporter.info(this.reporter.lang('uninstallRegenerate'));
+      this.flags.force = true;
```
bestander: Not sure we should force here

bestander:
Thanks, @aracarie.
Would you add a test please?
Here is an example https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L445

src/cli/commands/install.js
aracarie: Okay, should the install command not regenerate the lockfile then, and instead just display a warning in this case?

bestander: I think it should cleanup lockfile, yes.
Just force may mean more things like forcing rebuilding binary dependencies.

bestander: Probably the logic for missing dependencies should be done not in bailout function but where the lockfile is saved

aracarie: Got it. I'll try that out instead.

aracarie:
:+1: done

Issue #3751, opened 2017-06-29
bundledDependencies are not handled properly causing issues and warnings


bestander:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
```json
{
  "dependencies": {
    "npm": "^3.10.5"
  }
}
```
```bash
$ yarn
yarn install v0.27.0
warning package.json: No license field
info No lockfile found.
warning No license field
[1/4] Resolving packages...
warning npm > request > node-uuid@1.4.8: Use uuid module instead
[2/4] Fetching packages...
[3/4] Linking dependencies...
error An unexpected error occurred: "The same file /Users/bestander/work/temp/3202/node_modules/npm/node_modules can't be copied twice in one bulk copy".
info If you think this is a bug, please open a bug report with the information provided in "/Users/bestander/work/temp/3202/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Check introduced in #3712

Another example
```
yarn add nyc@11.0.3
```

bestander:
Scheduled to be released in 0.28 - either revert the check or find out why a folder is copied twice

cibernox:
FWIW, I've also experienced this using `--nightly` with the install script

bestander:
Thanks, I'll investigate + convert the check into a warning

cibernox:
Pinning to 0.27.3 works for me.

bestander:
Yeah, it is only in master right now

bestander:
The problem is in bundledDependencies.
Sending a PR to make the issues non breaking installation

bestander:
The bug is mitigated, the PRI can be lowered

BYK:
>The bug is mitigated, the PRI can be lowered

I think we should keep this high-prio and fix the underlying issue in 1.0

silverwind:
Still seeing these warnings on `yarn install npm` using v0.28.4.

clintmod:
@byk why no longer high priority? and why no longer in backlog?

artlogic:
Still happening in v1.0.1.

BYK:
@artlogic yup, aware. Will look into this.

Howard-HS:
not sure it it's a proper solution, remove yarn.lock then re-run yarn worked for me :)

bestander:
@sudo-suhas, that is a separate bug.
Afaik `yarn global list` actually runs install command :(

[...]

PR #3814, opened 2017-07-04
tone down error for double copies #3751

bestander:
Turns out https://github.com/yarnpkg/yarn/issues/3751 revealed a bug in fs.copy.
Instead of breaking install it now displays a warning.
The underlying issue needs to be fixed anyway though


BYK:
Do we need this merged urgently? If so I'd rather fix the underlying issue.

BYK:
On second look, this makes sense this we already have a set. Gonna merge and investigate later.

PR #3820, opened 2017-07-04
Don't prompt for version on CI

SimenB:
**Summary**

Our CI (travis enterprise) was stuck because of a change in package.json that was untested with an install.

image

**Test plan**
I couldn't find a fitting place to add a test. Ideas?

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -335,7 +335,8 @@ export default class Config {
 
     this.disablePrepublish = !!opts.disablePrepublish;
 
-    this.nonInteractive = opts.nonInteractive != null ? !!opts.nonInteractive : isCi || !process.stdout.isTTY;
+    // $FlowFixMe$
+    this.nonInteractive = !!opts.nonInteractive || isCi || !process.stdout.isTTY;
 
     this.requestManager.setOptions({
       offline: !!opts.offline && !opts.preferOffline,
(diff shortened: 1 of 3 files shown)


BYK:
Hey, thanks for the PR!

I think `nonInteractive` should be `true` when `isCI` is true. What do you think about doing that instead?

Crafting a test case for this might seem tricky but I think we can create this scenario and then pass `CI: 1` as an environment variable and make sure the test simply passes. Because it would time out otherwise. Right?

SimenB:
@BYK pushed the requested change.

Where would the test be added?

SimenB:
Not sure why I get a flow failure from moving some code... Halp?

src/config.js
```diff
@@ -334,7 +335,7 @@ export default class Config {
 
     this.disablePrepublish = !!opts.disablePrepublish;
 
-    this.nonInteractive = !!opts.nonInteractive;
+    this.nonInteractive = opts.nonInteractive != null ? !!opts.nonInteractive : isCi || !process.stdout.isTTY;
```
BYK: Wow, this is some hard-to-follow logic! How about this:
```
this.nonInteractive = !!opts.nonInteractive || isCi || !process.stdout.isTTY;
```

I know understand what you are trying to do (honor when `non-interative` is forced to `false`) but it is a case that cannot happen currently. What do you think?

SimenB: The though was "if `opts.nonInteractive` is set to anything, respect that. Otherwise check if we are on CI or stdout is non-tty".
Can simplify it a bit?

Issue #2484, opened 2017-01-18
Yarn cannot handle "//" package


nkovacs:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

According to Isaac Schlueter, this is a valid way to add comments to your package.json: https://groups.google.com/forum/#!msg/nodejs/NmL7jdeuw0M/yTqI05DRQrIJ

Yarn 0.19.1 does not support this. You can add a package that has a package.json like this, but if you try to run yarn install after that, it will fail.

yarn-error.log:

```
Yarn version: 
  0.19.1

Node version: 
  6.9.4

Platform: 
  linux x64

npm manifest: 
  {
    "name": "yarntest",
    "version": "1.0.0",
    "main": "index.js",
    "repository": {},
    "license": "MIT",
    "dependencies": {
      "keyczarjs": "^0.1.0"
    }
  }

yarn manifest: 
  No manifest

Lockfile: 
  # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
  # yarn lockfile v1
  
  
  keyczarjs@^0.1.0:
    version "0.1.0"
    resolved "https://registry.yarnpkg.com/keyczarjs/-/keyczarjs-0.1.0.tgz#c2f57f256a13211e967b842e034d201b4d45e837"
    dependencies:
      "//":
        "0" "Pinned version to avoid breaking us when something changes"
        "1" "we have twice had something break, so this way we control it."
        "2" "To test with the latest development version, use this:"
        "3" "git://github.com/digitalbazaar/forge.git"
      node-forge "0.3.0"
  
  node-forge@0.3.0:
    version "0.3.0"
    resolved "https://registry.yarnpkg.com/node-forge/-/node-forge-0.3.0.tgz#c428ed8d858c549ea146eac8123703c636b21965"

Trace: 
  TypeError: Cannot convert object to primitive value
      at /usr/share/yarn/lib/package-request.js:322:42
      at next (native)
      at step (/usr/share/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at /usr/share/yarn/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13

```

I also tested with a non-array comment package in my package.json, in that case the error was: "Received malformed response from registry for undefined. The registry may be down."

**If the current behavior is a bug, please provide the steps to reproduce.**

1. yarn init
2. yarn add keyczarjs
[...]

PR #3829, opened 2017-07-05
Add support for "//" comments in package.json dependencies

jseminck:
**Summary**
New feature:  Adds support for the "//" package in package.json, by deleting it before we resolve the dependencies. This issue was reported: https://github.com/yarnpkg/yarn/issues/2484

**Test plan**
I added an integration test

My main question is that there are plenty of other places where the package.json file is being read - do we also have to delete the "//" in all those places? Or is this single place sufficient?

More-over, are there any other keys other than "//" that we need to deal with?


jseminck:
Thanks for the comments. Will look further into this.

I added a new dependency because it was suggested in the original issue: https://github.com/yarnpkg/yarn/issues/2484#issuecomment-309756803

> Yarn should remove "comment" items from the loaded package.json. There are several existing libraries that can be used to do this.

But if we just want to stick to `dependencies`, `devDepencies` and `peerDependencies`, then a simple `delete` would do as well.

BYK:
>I added a new dependency because it was suggested in the original issue: #2484 (comment) - It also made it easier because it goes through the whole JSON object and deletes the key from everywhere.

Yeah, you're right. Sorry for overlooking that part. I think we're good now though, thanks a lot! If people complain about more comments in `package.json` we can probably add that dependency.

PR #3832, opened 2017-07-05
Fixes link: protocol linking

### LINKED ISSUE #3810: Link: protocol dependencies are not removed when the dependency changes
bug

The symlink is not removed even once the dependency has been removed from the package.json

```bash
set -e

rm -rf node_modules yarn.lock package.json

echo '{"dependencies":{"foo":"link:./hello-world"}}' > package.json
yarn

if [[ ! -L ./node_modules/foo ]]; then
    echo 'Failing (1)'
    exit 1
fi

echo '{}' > package.json
yarn

if [[ -L ./node_modules/foo ]]; then
    echo 'Failing (2)'
    exit 1
fi
```

The symlink should be removed

yarn 0.27.5

arcanis:
**Summary**

Fixes #3810 

**Test plan**

Adds a new test

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -152,6 +152,31 @@ test.concurrent(
   },
 );
 
+test.concurrent('replace the symlink when it changes, when using the link: protocol', async () => {
+  await runInstall({}, 'install-link', async (config, reporter): Promise<void> => {
+    const lockfile = await Lockfile.fromDirectory(config.cwd);
+
+    const pkgJson = await fs.readJson(`${config.cwd}/package.json`);
+    pkgJson.dependencies['test-missing'] = 'link:barbaz';
+    await fs.writeFile(`${config.cwd}/package.json`, JSON.stringify(pkgJson));
+
+    const reInstall = new Install({}, config, reporter, lockfile);
+    await reInstall.init();
+
+    const expectPath = path.join(config.cwd, 'node_modules', 'test-missing');
+
+    const stat = await fs.lstat(expectPath);
+    expect(stat.isSymbolicLink()).toEqual(true);
+
+    const target = await fs.readlink(expectPath);
+    if (process.platform !== 'win32') {
+      expect(target).toEqual('../barbaz');
+    } else {
+      expect(target).toMatch(/[\\\/]barbaz[\\\/]$/);
+    }
+  });
+});
+
 test('changes the cache path when bumping the cache version', async () => {
   await runInstall({}, 'install-github', async (config): Promise<void> => {
     const inOut = new stream.PassThrough();
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -251,10 +251,51 @@ export default class PackageLinker {
       }
     }
 
-    // linked modules
-    for (const loc of possibleExtraneous) {
-      const stat = await fs.lstat(loc);
+    // If an Extraneous is an entry created via "yarn link", we prevent it from being overwritten.
+    // Unfortunately, the only way we can know if they have been created this way is to check if they
+    // are symlinks - problem is that it then conflicts with the newly introduced "link:" protocol,
+    // which also creates symlinks :( a somewhat weak fix is to check if the symlink target is registered
+    // inside the linkFolder, in which case we assume it has been created via "yarn link". Otherwise, we
+    // assume it's a link:-managed dependency, and overwrite it as usual.
+    const linkTargets = new Map();
+
+    for (const entry of await fs.readdir(this.config.linkFolder)) {
+      const entryPath = path.join(this.config.linkFolder, entry);
+      const stat = await fs.lstat(entryPath);
+
       if (stat.isSymbolicLink()) {
+        const packageName = entry;
+        linkTargets.set(packageName, await fs.readlink(entryPath));
+      } else if (stat.isDirectory() && entry[0] === '@') {
+        // if the entry is directory beginning with '@', then we're dealing with a package scope, which
+        // means we must iterate inside to retrieve the package names it contains
+        const scopeName = entry;
+
+        for (const entry2 of await fs.readdir(entryPath)) {
+          const entryPath2 = path.join(entryPath, entry2);
+          const stat2 = await fs.lstat(entryPath2);
+
+          if (stat2.isSymbolicLink()) {
+            const packageName = `${scopeName}/${entry2}`;
+            linkTargets.set(packageName, await fs.readlink(entryPath2));
+          }
+        }
+      }
+    }
+
+    for (const loc of possibleExtraneous) {
+      let packageName = path.basename(loc);
+      let scopeName = path.basename(path.dirname(loc));
+
+      if (scopeName[0] === `@`) {
+        packageName = `${scopeName}/${packageName}`;
+      }
+
+      if (
+        (await fs.lstat(loc)).isSymbolicLink() &&
+        linkTargets.has(packageName) &&
+        linkTargets.get(packageName) === (await fs.readlink(loc))
+      ) {
         possibleExtraneous.delete(loc);
         copyQueue.delete(loc);
       }
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -285,7 +285,7 @@ export default class PackageLinker {
 
     for (const loc of possibleExtraneous) {
       let packageName = path.basename(loc);
-      let scopeName = path.basename(path.dirname(loc));
+      const scopeName = path.basename(path.dirname(loc));
 
       if (scopeName[0] === `@`) {
         packageName = `${scopeName}/${packageName}`;
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -253,6 +253,11 @@ async function buildActionsForCopy(
       }
     }
 
+    if (destStat && destStat.isSymbolicLink()) {
+      await unlink(dest);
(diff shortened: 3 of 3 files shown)


src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
 
-    // linked modules
-    for (const loc of possibleExtraneous) {
-      const stat = await fs.lstat(loc);
+    // if an extraneous is an entry created via "yarn link", we prevent it from being overwritten
+    // unfortunately, the only mean we have to know if they have been created this way is to check if
+    // they are a symlink - problem is that it then conflicts with the newly introduced "link:" protocol,
+    // which also creates symlinks :( a somewhat weak fix is to check if the symlink target is registered
+    // inside the linkFolder, in which case we assume it has been created via "yarn link". Otherwise, we
+    // remove it as usual.
```
arcanis: Cf this comment for the rational behind this change

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+    const linkTargets = new Map();
+
+    for (const entry of await fs.readdir(this.config.linkFolder)) {
+      const entryPath = path.join(this.config.linkFolder, entry);
+      const stat = await fs.lstat(entryPath);
       if (stat.isSymbolicLink()) {
+        linkTargets.set(entry, await fs.readlink(entryPath));
+      } else if (stat.isDirectory()) {
+        for (const entry2 of await fs.readdir(entryPath)) {
```
BYK: This logic needs to be explained. Why do we not recurse but just check one-level deep?

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+          const entryPath2 = path.join(entryPath, entry2);
+          const stat2 = await fs.lstat(entryPath2);
```
BYK: May be use better variable names like `statsInner` or something?

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+          if (stat2.isSymbolicLink()) {
+            linkTargets.set(`${entry}/${entry2}`, await fs.readlink(entryPath2));
+          }
+        }
+      }
+    }
+
+    for (const loc of possibleExtraneous) {
```
BYK: `possiblyExtraneous` or `maybeExtraneous`?

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+      const basename = path.basename(loc);
+      const parentname = path.basename(path.dirname(loc));
```
BYK: May be use `path.parse()` instead? (https://nodejs.org/api/path.html#path_path_parse_path)

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+
+      const checkname = parentname[0] === `@` ? `${parentname}/${basename}` : basename;
```
BYK: No idea what are we doing or why we are we doing it here. Are we checking for scoped packages? If so can we use more descriptive variable names like `isScopedPackage` or similar?

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+
+      if (
+        (await fs.lstat(loc)).isSymbolicLink() &&
+        linkTargets.has(checkname) &&
```
BYK: No need for the `.has()` check since `.get()` returns `undefined` if the key doesn't exist.

src/util/fs.js
```diff
@@ -253,6 +253,11 @@ async function buildActionsForCopy(
       }
     }
 
+    if (destStat && destStat.isSymbolicLink()) {
+      await unlink(dest);
+      destStat = null;
```
BYK: A comment about why we set this to `null` would be nice.

__tests__/commands/install/integration.js
```diff
@@ -152,6 +152,31 @@ test.concurrent(
...
 
+test.concurrent('replace the symlink when it changes, when using the link: protocol', async () => {
+  await runInstall({}, 'install-link', async (config, reporter): Promise<void> => {
+    const lockfile = await Lockfile.fromDirectory(config.cwd);
+
+    const pkgJson = await fs.readJson(`${config.cwd}/package.json`);
+    pkgJson.dependencies['test-missing'] = 'link:barbaz';
+    await fs.writeFile(`${config.cwd}/package.json`, JSON.stringify(pkgJson));
+
+    const install2 = new Install({}, config, reporter, lockfile);
```
arcanis: runInstall does the first one, and this is the second one

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
       }
     }
 
-    // linked modules
-    for (const loc of possibleExtraneous) {
-      const stat = await fs.lstat(loc);
+    // if an extraneous is an entry created via "yarn link", we prevent it from being overwritten
```
arcanis: Nope, in this context `extraneous` is used as a name (like in `possibleExtraneous`)

__tests__/commands/install/integration.js
```diff
@@ -152,6 +152,31 @@ test.concurrent(
...
+    const expectPath = path.join(config.cwd, 'node_modules', 'test-missing');
+
+    const stat = await fs.lstat(expectPath);
+    expect(stat.isSymbolicLink()).toEqual(true);
+
+    const target = await fs.readlink(expectPath);
+    if (process.platform !== 'win32') {
+      expect(target).toEqual('../barbaz');
+    } else {
+      expect(target).toMatch(/[\\\/]barbaz[\\\/]$/);
```
arcanis: This test is took from the one above. Testing a single one might work, but I don't think we have any guarantee regarding which one is used.

src/package-linker.js
```diff
@@ -251,10 +251,41 @@ export default class PackageLinker {
...
+    // unfortunately, the only mean we have to know if they have been created this way is to check if
+    // they are a symlink - problem is that it then conflicts with the newly introduced "link:" protocol,
+    // which also creates symlinks :( a somewhat weak fix is to check if the symlink target is registered
+    // inside the linkFolder, in which case we assume it has been created via "yarn link". Otherwise, we
+    // remove it as usual.
+    const linkTargets = new Map();
```
arcanis: Because a `Map` is semantically better, and is already used at various places in the code

src/package-linker.js
arcanis: scopped packages, ie. `@fakescope/fake-package`

src/util/fs.js
arcanis: if we don't, the `destStat` (ie. the stat object that contains the infos regarding `dest`) will be wrong, since we've just deleted it.

src/package-linker.js
arcanis: This is an already existing variable.

src/package-linker.js
arcanis: Yeah but we don't want to execute the `readlink` if the key doesn't exists.

src/package-linker.js
arcanis: We need to get the key inside the `linkTargets` map - ie the package name.

src/package-linker.js
arcanis: I'm not sure it would be much clearer -

```js
const parsedPath = path.parse(loc);

const basename = parsedPath.base;
const parentname = path.basename(parsedPath.dir);
```

__tests__/commands/install/integration.js
BYK: Okay, how about `secondInstall` or `secondRound` for a more descriptive name then?

__tests__/commands/install/integration.js
BYK: Alright, thanks for the clarification :)

src/package-linker.js
BYK: Oh okay :(

src/util/fs.js
BYK: Gotcha. Didn't realize this was used further on.

src/package-linker.js
BYK: (y)

src/package-linker.js
BYK: Okay great. Can we put this in the code as a comment? Or may be have a function named `checkScopedPackage` or something to express the intent?

src/package-linker.js
BYK: Oh, right.

We can still have more descriptive variable names like `packageName` and `scopeName` instead of `basename` and `parentname` to make the intent clearer, right?

src/package-linker.js
BYK: Oh, now I understand. How about `fullPackageName` (assuming we also rename `basename` and `parentname` above to my suggestions: `packageName` and `scopeName`.

src/package-linker.js
BYK: Actually, if that's the intention I think there may be a more efficient way since `readLink` already does an `lstat` call internally. We can possibly do something like this:
```
const linkTarget = linkTargets.get(checkname);
let canDelete;
try {
    canDelete = linkTarget && (await fs.readLink(loc)) === linkTarget);
} catch (err) {
    canDelete = false;
}

BYK:
We can merge once the CI builds pass.

PR #3836, opened 2017-07-06
Refactor branch/tag resolution for git dependencies, fixes #3720

### LINKED ISSUE #3720: Branch/tag resolution for git dependencies

a *bug*

We have a repository where we have a branch for the minor version (eg `1.1`) and then tag releases for patch versions in this branch (`v1.1.0`, `v1.1.1`, ...)

I want to point a project the branch so I can have changes that are not released yet, if I put:
```json
{
  "mno-enterprise-angular": "git+https://git@github.com/<org>/<project>.git#1.1"
}
```
It actually resolves to the last patch release (eg: `v1.1.1`) instead of the tip of the branch.

See above, an example is:
```json
{
  "package": "git+https://git@github.com/maestrano/mno-enterprise-angular.git#1.1"
}
```
The resulting `yarn.lock` is:
```
"mno-enterprise-angular@git+https://git@github.com/maestrano/mno-enterprise-angular.git#1.1":
  version "1.1.0"
  resolved "git+https://git@github.com/maestrano/mno-enterprise-angular.git#4fd4c179f5ca315bcf32e25c2cdbfe85102a0ae1"
```
`4fd4c17` is the `v1.1.0` tag

It is expected to resolve to the tip of the branch.
At the moment, I specify the commit in the `package.json` as a workaround but I can't use `yarn upgrade`.

* yarn v0.24.6
* node v6.11.0
* ubuntu 14.04

Volune:
**Summary**

For git dependencies, the branch/tag resolution does not always work as expected.
Refactor the code to make the resolution algorithm more explicit and follow discussions in #3720

**Test plan**

New test in `__tests__/util/git-ref-resolver.js`.

Also benefit from the fact that travis-ci ran the test with an older Git version. I don't know if we have a minimum git version requirement, it seems we currently don't test against any specific version.

**Implementation details**

I refactored `parseRefs` to return the full ref name, so we can differentiate branches and tags.

I use `git ls-remote --symref` to get the default branch name. This works only in recent versions of git (thanks travis-ci for reporting the issue), so I fallback to another algorithm if `--symref` is unavailable.

I tried not to use the variable name `hash`, which is confusing between the git-url hash and the commit hash. Any suggestion on variable names is welcomed.

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -2,24 +2,20 @@
 
 import type Config from '../config.js';
 import type {Reporter} from '../reporters/index.js';
+import type {ResolvedSha, GitRefResolvingInterface, GitRefs} from './git/git-ref-resolver.js';
 import {MessageError, SecurityError} from '../errors.js';
-import {removeSuffix} from './misc.js';
+import {spawn as spawnGit} from './git/git-spawn.js';
+import {resolveVersion, isCommitSha, parseRefs} from './git/git-ref-resolver.js';
 import * as crypto from './crypto.js';
-import * as child from './child.js';
 import * as fs from './fs.js';
 import map from './map.js';
 
 const invariant = require('invariant');
-const semver = require('semver');
 const StringDecoder = require('string_decoder').StringDecoder;
 const tarFs = require('tar-fs');
 const tarStream = require('tar-stream');
 const url = require('url');
 import {createWriteStream} from 'fs';
-
-type GitRefs = {
-  [name: string]: string,
-};
 
 type GitUrl = {
   protocol: string, // parsed from URL
@@ -31,20 +27,7 @@ const supportsArchiveCache: {[key: string]: boolean} = map({
   'github.com': false, // not support, doubt they will ever support it
 });
 
-// Suppress any password prompts since we run these in the background
-const env = {
-  GIT_ASKPASS: '',
-  GIT_TERMINAL_PROMPT: 0,
-  GIT_SSH_COMMAND: 'ssh -oBatchMode=yes',
-  ...process.env,
-};
-
-// This regex is designed to match output from git of the style:
-//   ebeb6eafceb61dd08441ffe086c77eb472842494  refs/tags/v0.21.0
-// and extract the hash and tag name as capture groups
-const gitRefLineRegex = /^([a-fA-F0-9]+)\s+(?:[^/]+\/){2}(.*)$/;
-
-export default class Git {
+export default class Git implements GitRefResolvingInterface {
   constructor(config: Config, gitUrl: GitUrl, hash: string) {
     this.supportsArchive = false;
     this.fetched = false;
@@ -95,10 +78,6 @@ export default class Git {
     };
   }
 
-  static spawn(args: Array<string>, opts?: child_process$spawnOpts = {}): Promise<string> {
-    return child.spawn('git', args, {...opts, env});
-  }
-
   /**
    * Check if the host specified in the input `gitUrl` has archive capability.
    */
@@ -114,7 +93,7 @@ export default class Git {
     }
 
     try {
-      await Git.spawn(['archive', `--remote=${ref.repository}`, 'HEAD', Date.now() + '']);
+      await spawnGit(['archive', `--remote=${ref.repository}`, 'HEAD', Date.now() + '']);
       throw new Error();
     } catch (err) {
       const supports = err.message.indexOf('did not match any files') >= 0;
@@ -126,13 +105,9 @@ export default class Git {
    * Check if the input `target` is a 5-40 character hex commit hash.
    */
 
-  static isCommitHash(target: string): boolean {
-    return !!target && /^[a-f0-9]{5,40}$/.test(target);
-  }
-
   static async repoExists(ref: GitUrl): Promise<boolean> {
     try {
-      await Git.spawn(['ls-remote', '-t', ref.repository]);
+      await spawnGit(['ls-remote', '-t', ref.repository]);
       return true;
     } catch (err) {
       return false;
@@ -151,7 +126,7 @@ export default class Git {
    * Attempt to upgrade insecure protocols to secure protocol
    */
   static async secureGitUrl(ref: GitUrl, hash: string, reporter: Reporter): Promise<GitUrl> {
-    if (Git.isCommitHash(hash)) {
+    if (isCommitSha(hash)) {
       // this is cryptographically secure
       return ref;
     }
@@ -203,7 +178,7 @@ export default class Git {
 
   async _archiveViaRemoteArchive(dest: string): Promise<string> {
     const hashStream = new crypto.HashStream();
-    await Git.spawn(['archive', `--remote=${this.gitUrl.repository}`, this.ref], {
+    await spawnGit(['archive', `--remote=${this.gitUrl.repository}`, this.ref], {
       process(proc, resolve, reject, done) {
         const writeStream = createWriteStream(dest);
         proc.on('error', reject);
@@ -220,7 +195,7 @@ export default class Git {
 
   async _archiveViaLocalFetched(dest: string): Promise<string> {
     const hashStream = new crypto.HashStream();
-    await Git.spawn(['archive', this.hash], {
+    await spawnGit(['archive', this.hash], {
       cwd: this.cwd,
       process(proc, resolve, reject, done) {
         const writeStream = createWriteStream(dest);
@@ -249,7 +224,7 @@ export default class Git {
   }
 
   async _cloneViaRemoteArchive(dest: string): Promise<void> {
-    await Git.spawn(['archive', `--remote=${this.gitUrl.repository}`, this.ref], {
+    await spawnGit(['archive', `--remote=${this.gitUrl.repository}`, this.ref], {
       process(proc, update, reject, done) {
         const extractor = tarFs.extract(dest, {
           dmode: 0o555, // all dirs should be readable
@@ -265,7 +240,7 @@ export default class Git {
   }
 
   async _cloneViaLocalFetched(dest: string): Promise<void> {
-    await Git.spawn(['archive', this.hash], {
+    await spawnGit(['archive', this.hash], {
       cwd: this.cwd,
       process(proc, resolve, reject, done) {
         const extractor = tarFs.extract(dest, {
@@ -290,32 +265,13 @@ export default class Git {
 
     return fs.lockQueue.push(gitUrl.repository, async () => {
       if (await fs.exists(cwd)) {
-        await Git.spawn(['pull'], {cwd});
+        await spawnGit(['pull'], {cwd});
       } else {
-        await Git.spawn(['clone', gitUrl.repository, cwd]);
+        await spawnGit(['clone', gitUrl.repository, cwd]);
       }
 
       this.fetched = true;
     });
-  }
-
-  /**
-   * Given a list of tags/branches from git, check if they match an input range.
-   */
-
-  async findResolution(range: ?string, tags: Array<string>): Promise<string> {
-    // If there are no tags and target is *, fallback to the latest commit on master
-    // or if we have no target.
-    if (!range || (!tags.length && range === '*')) {
-      return 'master';
-    }
-
-    return (
-      (await this.config.resolveConstraints(
-        tags.filter((tag): boolean => !!semver.valid(tag, this.config.looseSemver)),
-        range,
-      )) || range
-    );
   }
 
   /**
@@ -332,7 +288,7 @@ export default class Git {
 
   async _getFileFromArchive(filename: string): Promise<string | false> {
     try {
-      return await Git.spawn(['archive', `--remote=${this.gitUrl.repository}`, this.ref, filename], {
+      return await spawnGit(['archive', `--remote=${this.gitUrl.repository}`, this.ref, filename], {
         process(proc, update, reject, done) {
           const parser = tarStream.extract();
 
@@ -370,7 +326,7 @@ export default class Git {
     invariant(this.fetched, 'Repo not fetched');
 
     try {
-      return await Git.spawn(['show', `${this.hash}:${filename}`], {
+      return await spawnGit(['show', `${this.hash}:${filename}`], {
         cwd: this.cwd,
       });
     } catch (err) {
@@ -385,89 +341,90 @@ export default class Git {
    */
   async init(): Promise<string> {
     this.gitUrl = await Git.secureGitUrl(this.gitUrl, this.hash, this.reporter);
+
+    await this.setRefRemote();
+
     // check capabilities
-    if (await Git.hasArchiveCapability(this.gitUrl)) {
+    if (this.ref !== '' && (await Git.hasArchiveCapability(this.gitUrl))) {
       this.supportsArchive = true;
     } else {
       await this.fetch();
     }
 
-    return this.setRefRemote();
+    return this.hash;
   }
 
   async setRefRemote(): Promise<string> {
-    const stdout = await Git.spawn(['ls-remote', '--tags', '--heads', this.gitUrl.repository]);
-    const refs = Git.parseRefs(stdout);
+    const stdout = await spawnGit(['ls-remote', '--tags', '--heads', this.gitUrl.repository]);
+    const refs = parseRefs(stdout);
+    return this.setRef(refs);
+  }
+
+  setRefHosted(hostedRefsList: string): Promise<string> {
+    const refs = parseRefs(hostedRefsList);
     return this.setRef(refs);
   }
 
   /**
-   * TODO description
+   * Resolves the default branch of a remote repository (not always "master")
    */
 
-  async setRef(refs: GitRefs): Promise<string> {
-    // get commit ref
-    const {hash} = this;
-
-    const names = Object.keys(refs);
-
-    if (Git.isCommitHash(hash)) {
-      for (const name in refs) {
-        if (refs[name] === hash) {
-          this.ref = name;
-          return hash;
-        }
-      }
-
-      // `git archive` only accepts a treeish and we have no ref to this commit
-      this.supportsArchive = false;
-
-      if (!this.fetched) {
-        // in fact, `git archive` can't be used, and we haven't fetched the project yet. Do it now.
-        await this.fetch();
-      }
-      return (this.ref = this.hash = hash);
-    }
-
-    const ref = await this.findResolution(hash, names);
-    const commit = refs[ref];
-    if (commit) {
-      this.ref = ref;
-      return (this.hash = commit);
-    } else {
-      throw new MessageError(this.reporter.lang('couldntFindMatch', ref, names.join(','), this.gitUrl.repository));
+  async resolveDefaultBranch(): Promise<ResolvedSha> {
+    try {
+      const stdout = await spawnGit(['ls-remote', '--symref', this.gitUrl.repository, 'HEAD']);
+      const lines = stdout.split('\n');
+      const [, ref] = lines[0].split(/\s+/);
+      const [sha] = lines[1].split(/\s+/);
+      return {sha, ref};
+    } catch (err) {
+      // older versions of git don't support "--symref"
+      const stdout = await spawnGit(['ls-remote', this.gitUrl.repository, 'HEAD']);
+      const [sha] = stdout.split(/\s+/);
+      return {sha, ref: undefined};
     }
   }
 
   /**
-   * Parse Git ref lines into hash of tag names to SHA hashes
+   * Resolve a git commit to it's 40-chars format and ensure it exists in the repository
+   * We need to use the 40-chars format to avoid multiple folders in the cache
    */
 
-  static parseRefs(stdout: string): GitRefs {
-    // store references
-    const refs = {};
+  async resolveCommit(shaToResolve: string): Promise<?ResolvedSha> {
+    try {
+      await this.fetch();
+      const revListArgs = ['rev-list', '-n', '1', '--no-abbrev-commit', '--format=oneline', shaToResolve];
+      const stdout = await spawnGit(revListArgs, {cwd: this.cwd});
+      const [sha] = stdout.split(/\s+/);
+      return {sha, ref: undefined};
+    } catch (err) {
+      // assuming commit not found, let's try something else
+      return null;
+    }
+  }
 
-    // line delimited
-    const refLines = stdout.split('\n');
+  /**
+   * Resolves the input hash / ref / semver range to a valid commit sha
+   * If possible also resolves the sha to a valid ref in order to use "git archive"
+   */
 
-    for (const line of refLines) {
-      const match = gitRefLineRegex.exec(line);
+  async setRef(refs: GitRefs): Promise<string> {
+    // get commit ref
+    const {hash: version} = this;
 
-      if (match) {
-        const [, sha, tagName] = match;
-
-        // As documented in gitrevisions:
-        //   https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html#_specifying_revisions
-        // "A suffix ^ followed by an empty brace pair means the object could be a tag,
-        //   and dereference the tag recursively until a non-tag object is found."
-        // In other words, the hash without ^{} is the hash of the tag,
-        //   and the hash with ^{} is the hash of the commit at which the tag was made.
-        const name = removeSuffix(tagName, '^{}');
-
-        refs[name] = sha;
-      }
+    const resolvedResult = await resolveVersion({
+      config: this.config,
+      git: this,
+      version,
+      refs,
+    });
+    if (!resolvedResult) {
+      throw new MessageError(
+        this.reporter.lang('couldntFindMatch', version, Object.keys(refs).join(','), this.gitUrl.repository),
+      );
     }
 
-    return refs;
+    this.hash = resolvedResult.sha;
+    this.ref = resolvedResult.ref || '';
+    return this.hash;
   }
 }
--- a/src/util/git/git-ref-resolver.js
+++ b/src/util/git/git-ref-resolver.js
@@ -1,0 +1,165 @@
+/* @flow */
+
+import type Config from '../../config.js';
+import {removeSuffix} from '../misc.js';
+
+const semver = require('semver');
+
+export type ResolvedSha = {sha: string, ref: ?string};
+export interface GitRefResolvingInterface {
+  resolveDefaultBranch(): Promise<ResolvedSha>,
+  resolveCommit(sha: string): Promise<?ResolvedSha>,
+}
+export type GitRefs = Map<string, string>;
+export type ResolveVersionOptions = {
+  version: string,
+  config: Config,
+  git: GitRefResolvingInterface,
+  refs: GitRefs,
+};
+type Names = {tags: Array<string>, heads: Array<string>};
+
+const REF_TAG_PREFIX = 'refs/tags/';
+const REF_BRANCH_PREFIX = 'refs/heads/';
+
+// This regex is designed to match output from git of the style:
+//   ebeb6eafceb61dd08441ffe086c77eb472842494  refs/tags/v0.21.0
+// and extract the hash and ref name as capture groups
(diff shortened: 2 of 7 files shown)


src/util/git.js
```diff
@@ -385,89 +367,82 @@ export default class Git {
...
+  async setRef(refs: GitRefs): Promise<string> {
+    // get commit ref
+    const {hash: version} = this;
 
-        refs[name] = sha;
-      }
+    const resolvedResult = await resolveVersion(this.config, version, refs);
+    if (resolvedResult === false) {
+      throw new MessageError(
+        this.reporter.lang('couldntFindMatch', version, Object.keys(refs).join(','), this.gitUrl.repository),
```
sheerun: This can result in some unreadable stuff like Couldn't find match for ">=1.5,<1.6" in "refs/heads/design_experimental,refs/heads/experimental_modal_scroll,refs/heads/master,refs/merge-requests/1/head,refs/merge-requests/10/head,refs/merge-requests/100/head,refs/merge-requests/101/head,refs/merge-requests/102/head,refs/merge-requests/103/head,refs/merge-requests/104/head,refs/merge-requests/105/head,refs/merge-requests/106/head,refs/merge-requests/107/head,refs/merge-requests/108/head,refs/merge-requests/11/head,refs/merge-requests/12/head,refs/merge-requests/13/head,refs/merge-requests/14/head,refs/merge-requests/15/head,refs/merge-requests/16/head,refs/merge-requests/17/head,refs/merge-requests/18/head,refs/merge-requests/19/head,refs/merge-requests/2/head,refs/merge-requests/20/head,refs/merge-requests/21/head,refs/merge-requests/22/head,refs/merge-requests/23/head,refs/merge-requests/24/head,refs/merge-requests/25/head,refs/merge-requests/26/head,refs/merge-requests/27/head,refs/merge-requests/28/head,refs/merge-requests/29/head,refs/merge-requests/3/head,refs/merge-requests/30/head,refs/merge-requests/31/head,refs/merge-requests/32/head,refs/merge-requests/33/head,refs/merge-requests/34/head,refs/merge-requests/35/head,refs/merge-requests/36/head,refs/merge-requests/37/head,refs/merge-requests/38/head,refs/merge-requests/39/head,refs/merge-requests/4/head,refs/merge-requests/40/head,refs/merge-requests/41/head,refs/merge-requests/42/head,refs/merge-requests/43/head,refs/merge-requests/44/head,refs/merge-requests/45/head,refs/merge-requests/46/head,refs/merge-requests/47/head,refs/merge-requests/48/head,refs/merge-requests/49/head,refs/merge-requests/5/head,refs/merge-requests/50/head,refs/merge-requests/51/head,refs/merge-requests/52/head,refs/merge-requests/53/head,refs/merge-requests/54/head,refs/merge-requests/55/head,refs/merge-requests/56/head,refs/merge-requests/57/head,refs/merge-requests/58/head,refs/merge-requests/59/head,refs/merge-requests/6/head,refs/merge-requests/60/head,refs/merge-requests/61/head,refs/merge-requests/62/head,refs/merge-requests/63/head,refs/merge-requests/64/head,refs/merge-requests/65/head,refs/merge-requests/66/head,refs/merge-requests/67/head,refs/merge-requests/68/head,refs/merge-requests/69/head,refs/merge-requests/7/head,refs/merge-requests/70/head,refs/merge-requests/71/head,refs/merge-requests/72/head,refs/merge-requests/73/head,refs/merge-requests/74/head,refs/merge-requests/75/head,refs/merge-requests/76/head,refs/merge-requests/77/head,refs/merge-requests/78/head,refs/merge-requests/79/head,refs/merge-requests/8/head,refs/merge-requests/80/head,refs/merge-requests/81/head,refs/merge-requests/82/head,refs/merge-requests/83/head,refs/merge-requests/84/head,refs/merge-requests/85/head,refs/merge-requests/86/head,refs/merge-requests/87/head,refs/merge-requests/88/head,refs/merge-requests/89/head,refs/merge-requests/9/head,refs/merge-requests/90/head,refs/merge-requests/91/head,refs/merge-requests/92/head,refs/merge-requests/93/head,refs/merge-requests/94/head,refs/merge-requests/95/head,refs/merge-requests/96/head,refs/merge-requests/97/head,refs/merge-requests/98/head,refs/merge-requests/99/head,refs/tags/1.0.0,refs/tags/1.0.1,refs/tags/1.0.10,refs/tags/1.0.11,refs/tags/1.0.3,refs/tags/1.0.4,refs/tags/1.0.5,refs/tags/1.0.6,refs/tags/1.0.7,refs/tags/1.0.8,refs/tags/1.0.9,refs/tags/1.1.0,refs/tags/1.1.1,refs/tags/1.1.2,refs/tags/1.1.3,refs/tags/1.1.4,refs/tags/1.2.0,refs/tags/1.2.1,refs/tags/1.2.2,refs/tags/1.3.0,refs/tags/1.4.0,refs/tags/1.4.1,refs/tags/1.4.2,refs/tags/1.4.3,refs/tags/1.4.4,refs/tags/1.4.5,refs/tags/1.4.6,refs/tags/1.4.7,refs/tags/1.5.0,refs/tags/1.5.1,refs/tags/1.6.0,refs/tags/1.6.1,refs/tags/1.6.2,refs/tags/1.7.0,refs/tags/1.7.1" for "https://gitlab.com/leanlabsio/kanban.git"

sheerun:
It seems to be working even if merged with #3855 One thing to notice is that yarn.lock will have "version" different than resolved tag, i.e. "version" field from package.json on resolved commit, or "0.0.0" if package has no package.json

```
yarn add IndexOverflow/orange-notification#2.0.x
```
```
orange-notification@IndexOverflow/orange-notification#2.0.x:
  version "0.0.0"
  resolved "https://codeload.github.com/IndexOverflow/orange-notification/tar.gz/0c39d4f2aab707bfe5cab21633010e436ca49a8f"
```

Another case with package.json but incorrect version:

```
yarn add "gitlab:leanlabsio/kanban#>=1.5 <1.6"
```
```
"gitlab-kanban-client@gitlab:leanlabsio/kanban#>=1.5 <1.6":
  version "0.0.1"
  resolved "https://gitlab.com/leanlabsio/kanban/repository/archive.tar.gz?ref=4c311001a77674ec79788d12c302afba32907b8e"

```

It's probably not an issue but makes one wonder why yarn.lock even has "version" field locked. It seems "resolved" is more than enough.

src/util/git.js
sheerun: Maybe also replace each "refs/tags/" with space in this message

Volune: The `refs/merge-requests/*` had nothing to do here, it should be solved with latest commit.

sheerun:
Please kindly prioritize this PR, it's the last thing missing in Yarn to allow migration of Bower projects :)

BYK:
@sheerun sorry for missing. Gonna look at it tomorrow!

Volune:
Regarding the version in the lockfile, I think it would be interesting to put the resolved commit sha as the value of the version, and improve the `outdated` command to show if the dependency will resolve to a new commit. But that's out of the scope of this PR.

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  config: Config;
+  version: string;
+  refs: GitRefs;
+  names: ?Names;
+
+  async resolve(): Promise<DefaultBranch | ResolvedSha | false> {
+    const testFunctions = [
+      () => this.tryVersionAsGitCommit(),
```
BYK: You can do
```js
GitRefResolver.testFunctions = [
    GitRefResolver.tryVersionAsFullRef,
    GitRefResolver.tryVersionAsTagName,
];
//...
async resolve(): Promise {
    for (const testFunction of testFunctions) {
        const result = await testFunction.call(this);
        if (result) {
            return result;
        }
    }
    return false;
}
```

Would save you from creating bound functions and an array each time `resolve` is called which could result in significant performance and memory savings.

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+      () => (this.version === '' ? this.useDefaultBranch() : false),
```
BYK: `this.version === '' && this.useDefaultBranch()` would have the same effect.

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+      () => this.tryVersionAsFullRef(),
+      () => this.tryVersionAsTagName(),
+      () => this.tryVersionAsBranchName(),
+      () => this.tryVersionAsTagSemver(),
+      () => this.tryVersionAsBranchSemver(),
+      () => (this.version === '*' ? this.useDefaultBranch() : false),
```
BYK: Same as above: `this.version === '*' && this.useDefaultBranch()`

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+      const result = await testFunction();
+      if (result !== false) {
+        return result;
+      }
+    }
+    return false;
+  }
+
+  useDefaultBranch(): DefaultBranch {
+    return {defaultBranch: true};
```
BYK: I suggest doing the following instead:
```
const DEFAULT_BRANCH = Object.freeze({defaultBranch: true});
//...
useDefaultBranch(): DefaultBranch {
    return DEFAULT_BRANCH;
}
```

Actually, after this change, I don't see a reason for using a method at all. Why not use the constant directly?

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  tryVersionAsGitCommit(): ResolvedSha | false {
+    if (isCommitSha(this.version)) {
```
BYK: Simpler:
```
if (!isCommitSha(this.version)) {
    return false;
}
// rest of the code, without the extra indents :)
```

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+      for (const ref in this.refs) {
```
BYK:
```
const result = this.refs.find(ref => this.refs[ref].startsWith(this.version));
return result || {sha: this.version, ref: null};
```

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+        if (sha.startsWith(this.version)) {
+          return {sha, ref};
+        }
+      }
+      return {sha: this.version, ref: undefined};
+    }
+    return false;
+  }
+
+  tryRef(ref: string): ResolvedSha | false {
```
BYK:
```
return this.refs[ref] && {sha: this.refs[ref], ref};
```

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+
+  tryVersionAsGitCommit(): ResolvedSha | false {
+    if (isCommitSha(this.version)) {
+      for (const ref in this.refs) {
+        const sha = this.refs[ref];
+        if (sha.startsWith(this.version)) {
+          return {sha, ref};
+        }
+      }
+      return {sha: this.version, ref: undefined};
```
BYK: Danger: if I pass something that looks like a commit SHA like `fb00cc` which actually is a branch or a tag, this would resolve it as a version. Not sure if this is actually harmful or not.

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+    return this.tryRef(ref);
+  }
+
+  tryVersionAsBranchName(): ResolvedSha | false {
+    const ref = `${REF_BRANCH_PREFIX}${this.version}`;
+    return this.tryRef(ref);
+  }
+
+  getSemverNames(): Names {
+    if (!this.names) {
```
BYK: Same trick as above, return early:
```
if (this.names) {
    return this.names;
}
//actual function body
```

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+      const names = (this.names = {
+        tags: [],
+        branches: [],
+      });
+      for (const ref in this.refs) {
+        const match = refNameRegexp.exec(ref);
+        if (match) {
```
BYK:
```
const names = (this.names = {
    tags: [],
    branches: [],
});
const actions = {
    tags: tag => names.tags.push(tag),
    heads: head => names.branches.push(head),
};

this.refs.forEach(ref => {
    if (!match || semver.valid(match[2], this.config.looseSemver)) {
        continue;
    }
    const [, type, name] = match;
    actions[type](name);
});
```

src/util/git-ref-resolver.js
Volune: I need to put `return this.refs[ref] ? {sha: this.refs[ref], ref} : false` to match the method signature, is it worth it?

src/util/git-ref-resolver.js
Volune: I prefer keeping a single `return` for lazy-initializing getters. I'll do the change to be consistent with the other one

src/util/git-ref-resolver.js
Volune: I don't think I can use that, `this.refs` is an object. Maybe I can improve the code by changing `this.refs` to an instance of `Map`?

src/util/git-ref-resolver.js
```diff
@@ -0,0 +1,192 @@
...
+const semver = require('semver');
+
+export type GitRefs = {
+  [name: string]: string,
+};
+export type DefaultBranch = {defaultBranch: true};
+export type ResolvedSha = {sha: string, ref: ?string};
+type Names = {tags: Array<string>, branches: Array<string>};
+
+export const isCommitSha = (target: string): boolean => Boolean(target) && /^[a-f0-9]{5,40}$/.test(target);
```
Volune: I'll keep it case-sensitive a it was previously, but handle uppercase arguments in `tryVersionAsGitCommit`

Volune:
Regarding git sha and tag/branch:
A branch/tag name may indeed be interpreted as a sha. The current PR allows the `repo#refs/tags/1234567` notation to workaround this issue.

I didn't put verification to ensure the sha exists in the repository, because it seemed to me that the verification was _too costly_ for this step. If there are no tags or branches matching the commit (and I assume it will be most of the time), it would require fetching the remote repository to test that the sha is valid.

But after a few more tests, I realize that the current behavior may produce multiple cache folders for the same content, for example:
- `npm-test-js-git-repo-0.0.0-90f974f`
- `npm-test-js-git-repo-0.0.0-90f974f6`
- `npm-test-js-git-repo-0.0.0-90f974f63`
- `npm-test-js-git-repo-0.0.0-90f974f63a1ab7f9f9030808896ff39b4d597591`

So maybe I should make the verification after all. Any opinion?

src/util/git-ref-resolver.js
BYK: The other regex to match SHA's is case-insensitive so I think it was not intentional to make this case sensitive in the old version.

src/util/git-ref-resolver.js
BYK: Or you can do `return Boolean(this.refs[ref]) && {sha: this.refs[ref], ref}`.

Also, as mentioned earlier, I don't fully agree with this signature. I'd honestly prefer `null` or `undefined` instead of returning `false` which would allow you to use my first suggestion and switch the return type to `?ResolveSha`.

src/util/git-ref-resolver.js
BYK: Of, sorry. `Map` doesn't support `.find` or `.forEach` either but you can try `Object.keys(this.refs).find(` instead.

src/util/git-ref-resolver.js
BYK: >I prefer keeping a single return for lazy-initializing getters. I'll do the change to be consistent with the other one

That preference leads to an unnecessary indentation, makes the intent less clear and makes the code harder to follow. I don't think having fewer return statements are objectively better. Would you mind explaining your thought process behind that preference so I can understand where you are coming from (instead of trying to impose my preference over yours which is not fun :))

BYK:
>Regarding git sha and tag/branch:
>A branch/tag name may indeed be interpreted as a sha. The current PR allows the repo#refs/tags/1234567 notation to workaround this issue.

I think we should put this both in docs and in the code as a comment somewhere. Thanks for explaining :)

>I didn't put verification to ensure the sha exists in the repository, because it seemed to me that the verification was too costly for this step. If there are no tags or branches matching the commit (and I assume it will be most of the time), it would require fetching the remote repository to test that the sha is valid.

Not necessarily. Can't we use this endpoint for a quick validation: https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference (agreed that it still means a network request) - This link works for me for arbitrarily shortened SHA-1 refs: https://api.github.com/repos/yarnpkg/yarn/commits/8816df3c51a4

>But after a few more tests, I realize that the current behavior may produce multiple cache folders for the same content
>
>So maybe I should make the verification after all. Any opinion?

That sounds bad. I'd try to normalize all the references, probably to SHAs since that seems to be the common denominator. What do you think?

src/util/git-ref-resolver.js
```diff
@@ -38,39 +39,36 @@ class GitRefResolver {
 
   async resolve(): Promise<DefaultBranch | ResolvedSha | false> {
     const testFunctions = [
```
BYK: Still defining a new array for each invocation and the array is actually the same for every invocation.

BYK: Btw. I'm not very convinced that we need a class for this. Just having these functions in the module may be a more efficient approach. This is up to you though, just sharing my observation.

Volune: My first motivation for the class was the multiple parameters (`version`, `config`, `refs`, ... possibly growing). I can refactor and pass all of these as an object argument, that should indeed be more efficient.

src/util/git-ref-resolver.js
Volune: I think it was intentional because there is a test for that specific case. And after thinking about it, I think it may help keeping the sha consistently lowercase over all the application (except user input)

src/util/git-ref-resolver.js
Volune: That's a preference only for getters, most of the time I also prefer putting an if/return at the beginning and avoid indentation.
In the case of the getters:
- A single `return` statement
- Closer to the implementation of a simple getter
  ```
  getValue() {
    // if you add/remove lazy-init later, just change this line
    return this.value;
  }
  ```
- Ensure that the "value" is saved, to avoid errors like:
  ```
  getValue() {
    if (this.value) {
      return this.value;
    }
    const value = computeValue();
    // ...
    return value;
  }
  ```

Volune:
> Not necessarily. Can't we use this endpoint for a quick validation: 

That's true for Github. That would require some research for Bitbucket/Gitlab and will not work for private repositories.
There is a lot work to do about optimisations for Github/Bitbucket/... (also see my comments on #3923), but I want to have the feature working for all repositories for the moment, so I only consider the general case performance.

> That sounds bad. I'd try to normalize all the references, probably to SHAs since that seems to be the common denominator. What do you think?

The tags/branches and semver ranges already resolve to a unique full-length sha, so no problem for them.
The short sha are (yarn current behavior and current PR implementation) assumed to be valid, and used that way. If we verify them, we can at the same time resolve them to the full-length sha, and use the full-length sha in the end.

So I'll have a look to make this verification, doing the git fetch (Github optimisations can come later), and cleaning the code according to other comments.

Volune:
I've changed the PR code a lot to fix most of what we talked about. Let me know if you see other problems.

After this PR, I'll work on the github/bitbucket optimizations.

src/util/git/git-ref-resolver.js
```diff
@@ -0,0 +1,195 @@
...
+
+export const isCommitSha = (target: string): boolean => Boolean(target) && /^[a-f0-9]{5,40}$/.test(target);
+
+const REF_TAG_PREFIX = 'refs/tags/';
+const REF_BRANCH_PREFIX = 'refs/heads/';
+
+// This regex is designed to match output from git of the style:
+//   ebeb6eafceb61dd08441ffe086c77eb472842494  refs/tags/v0.21.0
+// and extract the hash and ref name as capture groups
+const gitRefLineRegex = /^([a-fA-F0-9]+)\s+(refs\/(?:tags|heads)\/.*)$/;
```
BYK: Follow the `const` naming conventions above: `GIT_REF_LINE_REGEX`

src/util/git/git-ref-resolver.js
```diff
@@ -0,0 +1,195 @@
...
+
+const refNameRegexp = /^refs\/(tags|heads)\/(.+)$/;
```
BYK: Naming conventions as mentioned above: `GIT_REF_NAME_REGEX`.

src/util/git/git-ref-resolver.js
```diff
@@ -0,0 +1,195 @@
...
+
+const tryVersionAsGitCommit = ({version, refs, git}: ResolveVersionOptions): Promise<?ResolvedSha> => {
+  const lowercaseVersion = version.toLowerCase();
+  if (!isCommitSha(lowercaseVersion)) {
+    return Promise.resolve(null);
+  }
+  for (const ref in refs) {
```
BYK: Clearer:
```js
const existingResolution = refs.find(ref => refs[ref].startsWith(lowercaseVersion));

return existingResolution ? Promise.resolve({ sha: refs[existingResolution], ref: existingResolution }) : git.resolveCommit(lowercaseVersion);
```

arcanis:
@Volune I understand the issue itself, but could you explain what is actually causing it in the code before your patch? It seems to me it simply a matter of sorting branches before tags, so I'm not sure about the complexity of this PR, but I might be missing something :)

src/util/git/git-ref-resolver.js
```diff
@@ -0,0 +1,195 @@
...
+  const names = {
+    tags: [],
+    branches: [],
+  };
+  for (const ref in refs) {
+    const match = refNameRegexp.exec(ref);
+    if (match) {
+      const [, type, name] = match;
+      if (semver.valid(name, config.looseSemver)) {
+        switch (type) {
```
Volune: I missed your comment 😕

src/util/git/git-ref-resolver.js
```diff
@@ -0,0 +1,195 @@
...
+
+export const resolveVersion = async (options: ResolveVersionOptions): Promise<?ResolvedSha> => {
+  const {version, git} = options;
+  if (version.trim() === '') {
+    return git.resolveDefaultBranch();
+  }
+
+  for (const testFunction of VERSION_RESOLUTION_STEPS) {
+    const result = await testFunction(options);
+    if (result !== null) {
```
Volune: We want to try the next function if this one returns `null`

src/util/git/git-spawn.js
```diff
@@ -0,0 +1,15 @@
...
+
+// Suppress any password prompts since we run these in the background
+const env = {
+  GIT_ASKPASS: '',
+  GIT_TERMINAL_PROMPT: 0,
+  GIT_SSH_COMMAND: 'ssh -oBatchMode=yes',
+  ...process.env,
+};
+
+export const spawn = (args: Array<string>, opts?: child_process$spawnOpts = {}): Promise<string> => {
```
Volune: Made mocking easier in tests

BYK: Alright. Is it too hard to mock this without putting it into a separate module? If so I'm okay keeping this, it just looked a bit off to me, that's all.

src/util/git/git-ref-resolver.js
BYK: Ah, right.

src/util/git/git-ref-resolver.js
BYK: No worries :)

Volune:
@arcanis this small piece of feature (resolving a usable commit sha and git ref) had more issues than just the tag/branch order. I've listed them in the latest commit message, and some new code comments.
Also I refactored to make the algorithm more explicit in the code (that's the point of the `VERSION_RESOLUTION_STEPS` array), and more tests.

@BYK I changed GitRefs to an ES6 Map, I hope it makes the code cleaner.

src/util/git/git-spawn.js
voxsim: I agree with @BYK I think it will be better if we move everything in the git module. I think you can check http://facebook.github.io/jest/docs/mock-function-api.html#content for some examples (or [here](https://github.com/facebook/jest/issues/936) if you have some doubts, they discuss a lot on this topic).
Btw great work! :D

src/util/git.js
```diff
@@ -385,89 +341,90 @@ export default class Git {
...
-      this.ref = ref;
-      return (this.hash = commit);
-    } else {
-      throw new MessageError(this.reporter.lang('couldntFindMatch', ref, names.join(','), this.gitUrl.repository));
+  async resolveDefaultBranch(): Promise<ResolvedSha> {
+    try {
+      const stdout = await spawnGit(['ls-remote', '--symref', this.gitUrl.repository, 'HEAD']);
+      const lines = stdout.split('\n');
+      const [, ref] = lines[0].split(/\s+/);
+      const [sha] = lines[1].split(/\s+/);
```
BYK: I'd rather use a stronger regexp like `/^\s*([0-9a-f]{5,40})\s+/` and just use `match` on it. Same goes for the `ref` if possible.

src/util/git.js
```diff
@@ -385,89 +341,90 @@ export default class Git {
...
-    const ref = await this.findResolution(hash, names);
-    const commit = refs[ref];
-    if (commit) {
-      this.ref = ref;
-      return (this.hash = commit);
-    } else {
-      throw new MessageError(this.reporter.lang('couldntFindMatch', ref, names.join(','), this.gitUrl.repository));
+  async resolveDefaultBranch(): Promise<ResolvedSha> {
+    try {
+      const stdout = await spawnGit(['ls-remote', '--symref', this.gitUrl.repository, 'HEAD']);
```
BYK: May be [`git symbolic-ref`](https://git-scm.com/docs/git-symbolic-ref) would be more precise?

src/util/git.js
```diff
@@ -385,89 +341,90 @@ export default class Git {
...
+      const lines = stdout.split('\n');
+      const [, ref] = lines[0].split(/\s+/);
+      const [sha] = lines[1].split(/\s+/);
+      return {sha, ref};
+    } catch (err) {
+      // older versions of git don't support "--symref"
+      const stdout = await spawnGit(['ls-remote', this.gitUrl.repository, 'HEAD']);
```
BYK: I think we should print a warning here about this being slower and less precise, urging the user to upgrade their Git version, what do you think?

src/util/git.js
```diff
@@ -385,89 +341,90 @@ export default class Git {
...
-      const match = gitRefLineRegex.exec(line);
+  async resolveCommit(shaToResolve: string): Promise<?ResolvedSha> {
+    try {
+      await this.fetch();
+      const revListArgs = ['rev-list', '-n', '1', '--no-abbrev-commit', '--format=oneline', shaToResolve];
+      const stdout = await spawnGit(revListArgs, {cwd: this.cwd});
+      const [sha] = stdout.split(/\s+/);
+      return {sha, ref: undefined};
+    } catch (err) {
+      // assuming commit not found, let's try something else
```
BYK: May be add some verbose logging here?

src/util/git.js
Volune: I think it doesn't work on remote repositories

src/util/git/git-spawn.js
Volune: That mean I would mock the spawn function and test the spawn function in the same test file, I think I will end up testing the mock and not the actual code.
I will delay this question after another PR that I will prepare about git hosted resolvers.

Volune:
> @Volune - I'll wait for your response before merging but I'm okay merging this PR as it is and then keep working on certain improvements.
@BYK  Okay for me. As I said before, I also want to work on other issues about the Git Hosted resolvers, and various optimisations in these resolvers that I'd like to share with the Git Fetcher. So it will be an opportunity to improve this code too.

BYK:
@Volune merged! Thanks a lot for keeping up with all the change requests and making submitting the PR in the first place!

Looking forward to more contributions from you! 🎉

PR #3840, opened 2017-07-06
Fix: honor selected version when saving deps

### LINKED ISSUE #3817: Save selected version from version prompt to package.json
Bug

Yarn saves the entered version number to package.json, even if it did not exist and another version was selected from a list. This will make yarn prompt the user with the version list each time `yarn` is ran.

Example:
I run `yarn add typescript@99`. Version 99 does not exist, so Yarn will prompt me with a list of versions to choose from.
When I select 2.4.1 and hit enter, version 2.4.1 will be installed.
Yarn then outputs this message:
```
└─ typescript@2.4.1
```
But yarn actually saved `"typescript": "99"` to package.json.

Yarn saves the selected version from the version prompt list to package.json.

Node 6.11.0
Yarn 0.27.5
Kubuntu 17.04

BYK:
**Summary**

Fixes #3817. Fixes how the version to be saved being determined
when adding packages with different patterns.

**Test plan**

Existing tests.

### DIFF
--- a/src/cli/commands/add.js
+++ b/src/cli/commands/add.js
@@ -12,7 +12,8 @@ import {buildTree} from './list.js';
 import {wrapLifecycle, Install} from './install.js';
 import {MessageError} from '../../errors.js';
 
-const invariant = require('invariant');
+import invariant from 'invariant';
+import semver from 'semver';
 
 export class Add extends Install {
   constructor(args: Array<string>, flags: Object, config: Config, reporter: Reporter, lockfile: Lockfile) {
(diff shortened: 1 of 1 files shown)


src/cli/commands/add.js
```diff
@@ -55,23 +56,22 @@ export class Add extends Install {
...
+      version = range;
     } else {
-      // default to save prefix
-      version = `${String(this.config.getOption('save-prefix') || '')}${pkg.version}`;
+      const prefix = tilde
+        ? '~' // --save-tilde
+        : exact
+          ? '' // --save-exact
+          : String(this.config.getOption('save-prefix')) || '^';
+      version = `${prefix}${pkg.version}`;
```
arcanis: I feel like the previous style was clearer.

arcanis:
Looks good, feel free to decide whether you want to keep the style change or revert it, then merge it.

arcanis:
Note that there is a failing test, tho, and it doesn't look like a flaky test!

src/cli/commands/add.js
```diff
@@ -55,23 +56,22 @@ export class Add extends Install {
...
     const {exact, tilde} = this.flags;
-    const parts = PackageRequest.normalizePattern(pattern);
+    const {hasVersion, range} = PackageRequest.normalizePattern(pattern);
     let version;
+
     if (getExoticResolver(pattern)) {
       // wasn't a name/range tuple so this is just a raw exotic pattern
       version = pattern;
-    } else if (parts.hasVersion && parts.range) {
+    } else if (hasVersion && range && (semver.satisfies(pkg.version, range) || getExoticResolver(range))) {
```
arcanis: Is the `|| getExoticResolver(range)` required? It should be catched by the `getExoticResolver(pattern)`, right (in the sense that if a pattern isn't exotic, the range probably shouldn't)?

src/cli/commands/add.js
```diff
@@ -55,23 +56,22 @@ export class Add extends Install {
...
       // if the user specified a range then use it verbatim
-      version = parts.range === 'latest' ? `^${pkg.version}` : parts.range;
-    } else if (tilde) {
-      // --save-tilde
-      version = `~${pkg.version}`;
-    } else if (exact) {
-      // --save-exact
-      version = pkg.version;
+      version = range;
```
arcanis: You removed the `latest` special-casing, which might be why the test is failing

src/cli/commands/add.js
BYK: The previous version was incorrect since it was checking the prefix only if the `hasRange` condition was not met. That said I can convert this into if statements instead of ternaries. I also find nested ternaries harder to read.

src/cli/commands/add.js
BYK: Yeah, unfortunately `getExoticResolver` actually only uses `Resolver.isVersion` so if you pass `package@githubUser/packageRepo#master` for instance, it doesn't match. This was only visible after I made the fix.

src/cli/commands/add.js
BYK: It really shouldn't since `semver.satisfies` should return false when `range` is `"latest"`. I verified this locally and tests pass locally. I'll see what's wrong there and fix it.

PR #3843, opened 2017-07-06
Fixes package linking when a filename casing changes

arcanis:
**Summary**

Fixes an issue where files were being removed when their name casing changed (`State.js` => `state.js`). Only occured on case-insensitive filesystems (OSX, Windows). 

**Test plan**

Tests should pass, + a new test to make sure it doesn't happen again


arcanis:
Damn, the tests work fine on OSX, but apparently not on Linux. I'll need to check this at home.

src/package-linker.js
```diff
@@ -296,7 +296,17 @@ export default class PackageLinker {
     // remove all extraneous files that weren't in the tree
     for (const loc of possibleExtraneous) {
       this.reporter.verbose(this.reporter.lang('verboseFileRemoveExtraneous', loc));
-      await fs.unlink(loc);
+
```
bestander: Can we just ignore case in fs.copyBulk instead?

fs.js: 153
```
  for (const loc of possibleExtraneous) {
    if (files.has(loc)) {
      possibleExtraneous.delete(loc);
    }
  }
```

Just call `toLocaleLowerCase()` when dealing with files Set.

Same thing should be applied to buildActionsForHardlink function

arcanis: Hmm maybe ... do you think the following would work?

```js
files.add(loc);
```

to


```js
files.add(loc);
files.add(loc.toLowerCase());
```

arcanis: This way, the `.has()` checks wouldn't have to change. As far as I see, the `files` set doesn't look used apart from those checks, so maybe it's enough?

bestander: `files` set is used just in 4 places.
So yeah, I think doing `files.add(loc.toLowerCase());` and `file.has(loc.toLowerCase())` should work

bestander: I don't think you need to add loc twice, that would cause confusion

src/package-linker.js
```diff
@@ -296,7 +296,17 @@ export default class PackageLinker {
...
+      const base = path.basename(loc);
+      const dir = path.dirname(loc);
+
+      // we iterate once again on every directory of our extraneous files to find out if they really exist
+      // we need to do this because of case-insensitive file systems, where deleting a.js will delete A.js
+      for (const file of await fs.readdir(dir)) {
+        if (base === file) {
```
BYK: Wouldn't this be a single file? If so why can't we use `fs.exists` directly?

src/util/fs.js
```diff
@@ -532,7 +532,11 @@ export async function copyBulk(
       const cleanup = () => delete currentlyWriting[data.dest];
       reporter.verbose(reporter.lang('verboseFileCopy', data.src, data.dest));
       return (currentlyWriting[data.dest] = readFileBuffer(data.src)
-        .then(d => {
+        .then(async (d) => {
```
arcanis: Yep, we don't seem to do anything concurrent here, so it could be refactored to use `await`

src/util/fs.js
```diff
@@ -541,7 +543,11 @@ export async function copyBulk(
       const cleanup = () => delete currentlyWriting[data.dest];
       reporter.verbose(reporter.lang('verboseFileCopy', data.src, data.dest));
       return (currentlyWriting[data.dest] = readFileBuffer(data.src)
-        .then(d => {
+        .then(async d => {
+          // we need to do this because of case-insensitive filesystems, which wouldn't properly
+          // change the file name in case of a file being renamed
+          await unlink(data.dest);
```
bestander: Great work, @arcanis.
I am a bit worried about the extra IO operation here.
What impact would it cause for normal installation?

PR #3847, opened 2017-07-07
Upgrade version rework 3603

rally25rs:
** Related PR **

Documentation changes: https://github.com/yarnpkg/website/pull/578

**Summary**

This is a PR for work done so far for RFC: https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-upgrade-command-consistency.md

`upgrade` command now works more like `upgrade-interactive`.

`upgrade-interactive` is now actually an interactive version of `upgrade`.

See RFC for details.


src/types.js
```diff
@@ -153,6 +153,8 @@ export type Dependency = {
   latest: string,
   url: string,
   hint: ?string,
+  range: string,
+  latestPattern: string,
```
arcanis: (+ in the rest of the code)

I feel like the name isn't very clear - the difference between `latest` and `latestPattern` should be explained somewhere

arcanis:
Don't forget that automated tests (at least on `upgrade`) will be required to merge it, tho :)

arcanis:
@rally25rs Let me know when you want another review! Would love to see this merged 👍

src/cli/commands/upgrade-interactive.js
```diff
@@ -4,47 +4,52 @@ import type {Dependency} from '../../types.js';
...
+import {Add} from './add.js';
+import {getOutdated, setFlags as setUpgradeFlags} from './upgrade.js';
 
 export const requireLockfile = true;
 
 export function setFlags(commander: Object) {
-  commander.usage('upgrade-interactive');
-  commander.option('-E, --exact', 'upgrade to most recent release with exact version');
-  commander.option('-T, --tilde', 'upgrade to most recent release with patch version');
+  setUpgradeFlags(commander);
```
rally25rs: Yeah I was just trying to avoid the duplication. I can put the flags in both commands though.

src/cli/commands/upgrade.js
```diff
@@ -21,12 +34,95 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
...
+        name: normalized.name,
+        wanted: '',
+        latest: '',
+        url: '',
+        hint: '',
+        range: '',
+        current: '',
+        latestPattern: newPattern,
+      });
+    }
```
rally25rs: This was the case where a user specifies an exact version for something that isn't outdated. For example if you have foo@2.0.0 and that happens to be the latest, it won't be returned by getOutdatedPackages, but the user can still do `yarn upgrade foo@1.2.4` and we would change to that requested version.
This is preserving functionality that the previous `upgrade` had (because it would just forward the package with version through to `add`)

src/cli/commands/upgrade.js
```diff
@@ -21,12 +34,95 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
...
-  const lockfile = useLockfile ? await Lockfile.fromDirectory(config.lockfileFolder, reporter) : new Lockfile();
-  const {dependencies, devDependencies, optionalDependencies, peerDependencies} =
-    (await config.readRootManifest()) || {};
-  const allDependencies = Object.assign({}, peerDependencies, optionalDependencies, devDependencies, dependencies);
-  let addArgs = [];
+  const lockfile = await Lockfile.fromDirectory(config.lockfileFolder);
+  const deps = await getOutdated(config, reporter, flags, lockfile, args);
+
+  // if specific versions were requested for packages, override what getOutdated reported as the latest to install
+  args.forEach(requestedPattern => {
```
rally25rs: Is there a reason, or just personal preference? I always visually mistake `for ... of ...` as the old `for ... in ...` which you seldom actually want. The subtle difference always messes me up.

src/cli/commands/upgrade.js
```diff
@@ -21,12 +34,95 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
...
+  if (!deps.length) {
+    reporter.success(reporter.lang('allDependenciesUpToDate'));
+    return;
+  }
+
+  // remove deps being upgraded from the lockfile, or else Add will use the already-installed version
+  // instead of the latest for the range.
+  deps.forEach(dep => lockfile.removePattern(dep.latestPattern));
+
+  const addFlags = Object.assign({}, flags, {force: true});
```
rally25rs: Maybe. Probably. This is what the old/existing code did so I preserved it without much thought. Does the `commander` library that handles the command line options have anything to enforce only the allowed flags?

src/types.js
rally25rs: yeah, I agree... I had a hard time coming up with a meaningful name. Any suggestions? It's basically the `name@version` that we will upgrade to, which gets passed as input to the `add` command. Maybe something like `upgradeToPattern` or just `upgradeTo`?

rally25rs:
@arcanis I did some refactoring and linked a PR for the docs.
I also merged `master` as of a couple days ago and resolved some conflicts.
Hopefully this is good to go now...

src/cli/commands/upgrade.js
```diff
@@ -21,53 +67,110 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
...
+    flags.caret = false;
   }
 
-  return dependency;
+  normalizeScope();
+
+  const deps = (await PackageRequest.getOutdatedPackages(lockfile, install, config, reporter, patterns))
+    .filter(versionFilter)
+    .filter(scopeFilter);
+  deps.forEach(dep => (dep.upgradeTo = buildPatternToUpgradeTo(dep, flags)));
```
arcanis: It won't block the merge but I think we should try using `for ... of` when possible, instead of `forEach`

arcanis:
Thanks a lot, @rally25rs ! I'm really happy to see this merged !

cpojer:
Great work @rally25rs, thank you so much for your contributions to Yarn :)

danez:
This has changed what `yarn upgrade` is doing. Before this change it was updating all dependencies in the tree, by respecting package.json ranges (by simply ignoring the lockfile and creating a new one) now it only updates dependencies I have in my `package.json`.

The only way to achieve the old behaviour is to remove `yarn.lock` and then do `yarn install`. Is this expected? It was really one of the nice features in comparison to npm to upgrade everything without having to manually touch the filesystem.

bdwain:
@danez we're discussing that in #4476

Issue #3530, opened 2017-05-30
Yarn create fails on Windows with "spawn UNKNOWN"


guigrpa:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Running `yarn create` with `create-react-app` or with the tool I'm building (`create-monorepo`) fails, throwing a `spawn UNKNOWN` error (more details below).

**If the current behavior is a bug, please provide the steps to reproduce.**

```
$ yarn create react-app example-app
yarn create v0.24.6
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Installed "create-react-app@1.3.1" with binaries:
      - create-react-app
warning No license field
error An unexpected error occurred: "spawn UNKNOWN".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\ggrau\\AppData\\Local\\Yarn\\config\\global\\yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/create for documentation about this command.
```

Global installation of `create-react-app` actually succeeds, since running `create-react-app example-app` actually succeeds, but not with `yarn create`.

The same error occurs with `yarn create monorepo`.

**What is the expected behavior?**

Error should not be thrown, and the `create-react-app` script should be launched.

**Please mention your node.js, yarn and operating system version.**

Node v7.10.0
yarn 0.24.6
Windows 7 SP1

The problem also occurs on the same PC running Node v7.4.0, and on *another PC* running Node v7.9.0. It **does not occur on macOS Sierra (10.12), Node 7.8.0**.

guigrpa:
Actually, I'm a bit confused. Yarn's latest stable release seems to be 0.24.6 (the one I have installed via .msi installer), the [online docs](https://yarnpkg.com/en/docs/cli/create) include `yarn create`, there's even [a blog post](https://yarnpkg.com/blog/2017/05/12/introducing-yarn/) covering this feature, and `yarn create` seems to *do something*, but the [release notes for **0.25.0**](https://github.com/yarnpkg/yarn/releases/tag/v0.25.0) are the ones including `yarn create` as a new feature...

vinodronold:
I am also getting this error:
````
C:\EX>yarn create react-app FFREACTIPY
yarn create v0.24.6
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[...]

PR #3848, opened 2017-07-07
Fix `yarn create` command on Windows

Volune:
**Summary**

Yarn create fails on Windows, see #3530 
The issue is caused by Yarn trying to run a javascript file as an executable (the !# doesn't work in Windows command)

This PR uses the existing `linkBin` function to create any appropriate shim depending on the system.

**Test plan**

No existing test. Manually tested on Windows 10 and one linux. More tests by other people would be appreciated.
Also any idea to automate a test here is welcomed.

### DIFF
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -1,14 +1,12 @@
 // @flow
 
-import {resolve} from 'path';
-
 import type Config from '../../config.js';
 import {MessageError} from '../../errors.js';
-import {registryNames} from '../../registries/index.js';
 import type {Reporter} from '../../reporters/index.js';
 import * as child from '../../util/child.js';
-import * as fs from '../../util/fs.js';
-import {run as runGlobal} from './global.js';
+import {run as runGlobal, getBinFolder} from './global.js';
+
+const path = require('path');
 
 export function setFlags(commander: Object) {}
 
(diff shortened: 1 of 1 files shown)


src/cli/commands/create.js
```diff
@@ -51,7 +52,10 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       throw new MessageError(reporter.lang('createInvalidBin', packageName));
     }
 
-    await child.spawn(binPath, rest, {stdio: `inherit`});
+    const linkedBinPath = config.getTemp(`create-bin-${packageName}`);
+    await linkBin(binPath, linkedBinPath);
```
arcanis: I'm not sure calling `linkBin` is required, is it? It should have been done when adding the package (l30)

Volune: I assumed we were previously not using the already linked bin for a reason. If there are no reasons, I can indeed change to use the existing one in global.

arcanis: Nope, no particular reason - I just forgot about this behavior when I implemented the feature 😞

Volune: @arcanis Please review the new implementation. Tested on Windows.
Also I wonder, shouldn't we pass the flags to `runGlobal` (and now also to `getBinFolder`)?

Issue #3853, opened 2017-07-07
To avoid breaking the internet, use strong dependency versioning


jdbevan:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When installing yarn via npm (which I know is not the official way but is still in significant usage) the install can break if a dependency releases a patch or minor change. See https://github.com/yarnpkg/yarn/issues/3850, https://github.com/yarnpkg/yarn/issues/3851 and https://github.com/omnidan/node-emoji/pull/45

Given that part of the point of using yarn is to avoid this kind of problem, it would be nice if yarn did actually avoid this kind of problem :)

**If the current behavior is a bug, please provide the steps to reproduce.**
Release a broken patch or minor release of a dependency, install yarn via npm

**What is the expected behavior?**
Yarn installs successfully

**Please mention your node.js, yarn and operating system version.**
OS agnostic
Node agnostic
Yarn >= 0.24.0

MikeN123:
Yarn 0.21.3 broke as well, with a different error.

hbeachey:
Any version (we had all the way back to 0.16.1) which depend on the updating package breaks. Not sure previous versions really need the fix, but it's something to mention.

bestander:
As described here https://github.com/yarnpkg/yarn/pull/3854#issuecomment-313746862.

We need to publish only the single-file-build yarn.js to npm.
Looking forward for a PR

Daniel15:
> We need to publish only the single-file-build yarn.js to npm.

Publishing the `dist` directory after running `scripts/build-dist.sh` should be sufficient, we'd just need to have a modified `package.json` that removes all dependencies (since we'd no longer need any of the dependencies).

Daniel15:
https://github.com/yarnpkg/yarn/pull/3866

Issue #3850, opened 2017-07-07
npm install -g yarn -> Cannot set property 'getDocsInfo' of undefined


raphael22:
Last installation though npm lead to this error after cli usage.

```
npm install -g yarn

yarn

/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/lib/cli/commands/index.js:308
commands[key].getDocsInfo = getDocsInfo(key);

TypeError: Cannot set property 'getDocsInfo' of undefined
    at Object.<anonymous> (/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/lib/cli/commands/index.js:308:29)
    at Module._compile (/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at _load_index3 (/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/lib/cli/index.js:18:43)
    at Object.<anonymous> (/home/nikopol/.nvm/versions/node/v6.9.4/lib/node_modules/yarn/lib/cli/index.js:135:54)

```

**Please mention your node.js, yarn and operating system version.**
node : v6.9.4 / v6.11
yarn : tried v0.24 & v0.27.5
os : Lubuntu x64

emckay:
I'm also having this issue on v0.24.5 and v0.27.5 on Ubuntu 16.04 with node 6.10.3

eluleci:
Me too while creating Docker image from node:6.9.5-alpine .

talldan:
We've also just started experiencing this out of the blue:

Node v6.10.3
Yarn: 0.27.5
Ubuntu (not sure on the version, it's a CI server)

ot-juan-davila:
same issue 
Node v6.9.5
Yarn: 0.27.5
windows 10

jdbevan:
I got the same on:
Node v6.11.0
Yarn 0.27.5 / 0.27.3
macOS Sierra 10.12.5

[...]

PR #3854, opened 2017-07-07
Add npm shrinkwrap file.

danthewolfe:
Since most people and many CIs do the initial install of yarn by
running `npm install yarn`, please include the npm shrinkwrap file.

**Summary**
Adds the file 'npm-shrinkwrap.json'

This will prevent issues such as #3850 when installing yarn via npm, as suggested in #3853

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,6484 @@
+{
+  "name": "yarn",
+  "version": "0.27.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@gulp-sourcemaps/identity-map": {
+      "version": "1.0.1",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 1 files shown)


jiripospisil:
Be aware of this however https://github.com/npm/npm/issues/17185.

arcanis:
Hey, thanks for this PR! However, I'm not sure this is the right way to solve this issue. We should ideally publish the single-file releases, which wouldn't be affected by any dependency issue.

bestander:
I agree with @arcanis.
Yarn is not ready to maintain the shrinkwrap file in its repo because all developers are using Yarn for development.
As @arcanis described to avoid subdependencies breaking Yarn it is better to introduce a prepublish script that would bundle yarn.js and publish only this file.

I'll close this PR but looking forward for a PR with the prepublish script 🥇

PR #3855, opened 2017-07-07
Fix allowing to install packages without manifest

sheerun:
Fixes my previous PR: #3624

It also adds integration tests to ensure we cover this feature.


bestander:
Great job, @sheerun.
Thanks a lot.

PR #3857, opened 2017-07-07
Fix: exit with non-zero only when wanted pkgs are really outdated

lirantal:
**Summary**

Today, yarn `outdated` shows a yellow and red formatted list of outdated packages but regardless of each type of outdated packages it exists with an error.

It makes sense to return a non-zero exit code only in cases where the _current_ package installed is not aligned with the _wanted_ package. If they are on the same version and a more up to date version exists, it doesn't necessarily convey a problem.

The strongest motivation behind the error code is that users can run `yarn outdated` on their CI builds and if a version is out of date than what they expect then it will fail the build. This Pull Request will amend the outdated command for this expected behavior.

**Test plan**

Demonstrating the fix:

image

### DIFF
--- a/src/cli/commands/outdated.js
+++ b/src/cli/commands/outdated.js
@@ -28,11 +28,15 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   }
 
   const getNameFromHint = hint => (hint ? `${hint}Dependencies` : 'dependencies');
+  const isOutdatedPackage = ({current, wanted, name}) =>
+    current === wanted ? 0 : 1;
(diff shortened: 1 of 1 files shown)


arcanis:
> It makes sense to return a non-zero exit code only in cases where the current package installed is not aligned with the wanted package.

Aren't you looking for `yarn check`?

lirantal:
Looks exactly what I needed and the `--integrity` is a plus. Thanks @arcanis!
We can close.

arcanis:
👍

PR #3862, opened 2017-07-08
Fix various commands asking for a password

johnf:
**Summary**

Commands like ```publish``` were asking for a password instead of using the token in ```.npmrc```.

The core code path that was causing the issue is https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L218-L221

``` javascript
// If sending a request to the Yarn registry, we must also send it the auth token for the npm registry
if (baseRegistry === YARN_REGISTRY) {
  registries.push(DEFAULT_REGISTRY);
}
```

The problem was ```baseRegistry = "https://registry.yarnpkg.com"``` while ```YARN_REGISTRY = "https://registry.yarnpkg.com/"```. Notice the trailing slash.

I've reverted the code added in https://github.com/yarnpkg/yarn/commit/33820719e2c049ff93e86d29251b96d496f2b6e8#diff-b053bee294c216269844e5874039b6caR172
which solves the problem. Unclear to me why it was added though.

**Test plan**

yarn publish no longer asks for a password.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -192,7 +192,7 @@ export default class NpmRegistry extends Registry {
       const availableRegistries = this.getAvailableRegistries();
       const registry = availableRegistries.find(registry => packageName.startsWith(registry));
       if (registry) {
-        return addSuffix(registry, '/');
+        return String(registry);
       }
     }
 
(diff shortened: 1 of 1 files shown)


arcanis:
Ping @lumaxis? Do you remind the rational behind this change?

johnf:
@lumaxis @arcanis ping

Should I rewrite this to just make this bit of code above work instead of the current change?
This would remove any possible side effects.

lumaxis:
@arcanis @johnf Heya 👋 Sorry for the slow response on this, was out sick last week and am only slowly catching up on email.
I didn't add this specific change myself as I was also partly building on top of the work of someone else. 😕

arcanis:
No problem, thanks! Let's merge this then, all tests are green so we can be reasonably confident this won't break anything :)

Issue #3806, opened 2017-07-03
yarn upgrade --latest error


IgorAufricht:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
When using `yarn upgrade --latest`, in some cases there's an error (even though `yarn` and `yarn upgrade` work fine).

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Create an empty directory with the following `package.json` (this is the minimal example that shows the error):
```
{
  "name": "test",
  "dependencies": {
    "karma-coverage": "^1.1.1",
    "karma-coverage-istanbul-reporter": "^0.2.0",
    "karma-jasmine": "~1.1.0",
    "karma-junit-reporter": "^1.2.0",
    "karma-phantomjs-launcher": "^1.0.2",
    "karma-sourcemap-loader": "~0.3.7",
    "karma-webpack": "^1.8.1",
    "phantomjs-prebuilt": "^2.1.14"
  },
  "devDependencies": {
    "grunt": "~1.0.1"
  }
}
```
2. Run `yarn`. (This works correctly.)
3. Run `yarn upgrade --latest`.
4. I get an error like this (clearing yarn cache doesn't help):
```
$ yarn upgrade --latest
yarn upgrade v0.27.5
warning package.json: No license field
warning package.json: No license field
warning test: No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
warning Pattern ["phantomjs-prebuilt@^2.1.14"] is trying to unpack in the same destination "C:\\Users\\U001912\\AppData\\Local\\Yarn\\cache\\v1\\npm-phantomjs-prebuilt-2.1.14-d53d311fcfb7d1d08ddb24014558f1188c516da0" as pattern ["phantomjs-prebuilt@^2.1.7"]. This could result in a non deterministic behavior, skipping.
[3/4] Linking dependencies...
warning "webpack-dev-middleware@1.11.0" has unmet peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
warning "karma-jasmine@1.1.0" has unmet peer dependency "jasmine-core@*".
warning "karma-jasmine@1.1.0" has unmet peer dependency "karma@*".
warning "karma-junit-reporter@1.2.0" has unmet peer dependency "karma@>=0.9".
warning "karma-phantomjs-launcher@1.0.4" has unmet peer dependency "karma@>=0.9".
[...]

PR #3865, opened 2017-07-09
Bugfix 3806: Adding a condition for install lifecycle scripts

aracarie:
**Summary**
Install lifecycle scripts will no longer be run for hoisted
dependencies. Previously these were run against the cache, causing
errors. See issue #3806.


arcanis:
@aracarie Thanks! Would you mind adding a test to show the fix and make sure it doesn't happen again?

aracarie:
Yep, just added.

Issue #3851, opened 2017-07-07
TypeError: Cannot set property 'getDocsInfo' of undefined


yoyo837:
**Do you want to request a *feature* or report a *bug*?**

bug
**What is the current behavior?**
when i upgrade yarn：npm i yarn@0.27.5 -g and run yarn first，it reply :TypeError: Cannot set property 'getDocsInfo' of undefined。
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Please mention your node.js, yarn and operating system version.**
win 10 15063 x64，node 6.10.3 x64, npm 3.10.10

yoyo837:
Every time reply this error, actually.

ralt:
Yeah, same here. All our builds are failing because of this error. 3 other people working on different projects are also getting this.

igeligel:
Can report this too.

yoyo837:
I can't work, because yarn can't execute run dev, it's frustrating.

jdbevan:
This is a duplicate of https://github.com/yarnpkg/yarn/issues/3850

paulmillr:
Drop yarn folks, no reason to use it with npm v5.

yakke:
just updated the ubuntu package and having same issue as reported above. Also our builds using the npm's version are failing

hbeachey:
It's because the package `node-emoji` updated with a new dependency of `lodash.toArray`

AChusuei:
```
/Users/AlanChusuei/.meteor/packages/meteor-tool/.1.4.3_1.d4y6i2++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/yarn/lib-legacy/cli/commands/index.js:322
  commands[key].getDocsInfo = getDocsInfo(key);
                            ^

TypeError: Cannot set property 'getDocsInfo' of undefined
    at Object.<anonymous> (/Users/AlanChusuei/.meteor/packages/meteor-tool/.1.4.3_1.d4y6i2++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/yarn/lib-legacy/cli/commands/index.js:322:29)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
[...]

PR #3866, opened 2017-07-09
Publish tarball with single JS file build to npm

### LINKED ISSUE #3853: To avoid breaking the internet, use strong dependency versioning

Daniel15:
**Summary**
Updates our build to publish the regular tarball to npm, rather than the raw working directory. This will publish the single JS file build to npm

Advantages:
 - Much faster to install (no need to install any dependent packages)
 - Brings Yarn npm package in line with all other installation methods
 - Locks all dependency versions, preventing issues like #3850 and #3851 from happening
 - We can remove our npm access token from CircleCI

Disadvantages:
 - Code using "Yarn-as-an-API" (for example, `require('yarn/lib/lockfile/parse.js')`) will no longer work. We don't officially support this today, and I didn't see much [code on GitHub that does this](https://github.com/search?utf8=%E2%9C%93&q=%22require%28%27yarn%2Flib%22&type=Code)

Because of the API change, we should not do this until the next RC (ie. **do not** cherry-pick this diff into the stable branches). Any existing packages that depend on Yarn should continue to work, as they'll continue to use the old version of Yarn (assuming they don't have a bad version range like `"yarn": "*"` in their `package.json`).

The proper fix to the API thing would be to officially publish packages for the reusable parts (for example, publish the lockfile parser as `@yarnpkg/lockfile` or something like that).

**Test plan**

Tested by modifying `name` in `package.json` to be `@daniel15/yarn-test`, and then using `scripts/update-npm.sh` to publish that package: https://www.npmjs.com/package/@daniel15/yarn-test

I haven't actually tested the Jenkins job, however it's identical to the Homebrew one except it just executes a different script.

Closes #3853


circle.yml
```diff
@@ -41,17 +41,6 @@ test:
...
-deployment:
-  release:
-    tag: /v[0-9]+(\.[0-9]+)*/
-    owner: yarnpkg
-    commands:
-      # Only NPM is handled here - All other release files are handled in a webhook.
-      - echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
-      - ./scripts/set-installation-method.js "`pwd`/package.json" npm
-      # https://github.com/yarnpkg/yarn/issues/610, need to make yarn publish noninteractive
-      - npm publish --tag rc
```
Daniel15: This will now be ran by the Jenkins job. To clarify the overall flow, it's:

1. Someone tags the new release
2. CircleCI and AppVeyor build the release and then call their webhooks
3. Once both the CircleCI and AppVeyor builds are both completed, the webhook code kicks off the `yarn-version` Jenkins build job
4. `yarn-version` updates the version number on the Yarn site, then kicks off `yarn-chocolatey`, `yarn-homebrew` and `yarn-npm` build jobs
5. `yarn-npm` pushes to npm

(I'll update the docs at some point)

BYK: I think capturing the "why" of this process is also quite important. For instance, I currently don't understand why we need to do it this way instead of just doing everything from CircleCI on a tag push :)

BYK:
>Code using "Yarn-as-an-API" (for example, require('yarn/lib/lockfile/parse.js')) will no longer work. We don't officially support this today, and I didn't see much code on GitHub that does this

I'm very concerned about this aspect. May be we should also split yarn into `yarn` and `yarn-cli` where `yarn` is the actual project and `yarn-cli` is what we publish right now?

circle.yml
bestander: I agree with @BYK.
The advantage of Circle CI builds is that they are simple, easy to debug and maintain.
Moving everything to custom jenkins bot may be more cohesive and safe but the support becomes depending 99% on you, @Daniel15

Daniel15: @bestander - Sure. This one is easy to move back into CircleCI if we want to do that. It's pretty straightforward.

@BYK The main reason was to make it consistent with all the other builds. We could continue publishing the npm build here, but the `.tar.gz`, `.msi` and `.js` builds will always be published on infra we run though, as they're all signed (Authenticode for .msi, GPG for everything else). There's no way I'd be comfortable with having a signing key sitting on CircleCI's infra 😃

Daniel15:
> May be we should also split yarn into `yarn` and `yarn-cli` where `yarn` is the actual project and `yarn-cli` is what we publish right now?

People are almost always going to be interested in the CLI, and `yarn-cli` seems unnecessarily verbose. We could have `yarn` and `yarn-lib`, maybe?

arcanis:
Yup, using `yarn` for the cli means less friction for users that want to install yarn, which will be the most common occurence. Once we'll have settled on an API, third party developers will just have to use another name (it should even be possible to warn them of the mistake at install time if they depend on the `yarn` package).

Daniel15:
After spending a while thinking about this and chatting to @bestander, I think I'm going to submit a pull request to move this particular step back to publishing directly from CircleCI. The reasons that the other releases are pushed from infra under our own control (eg. signing them with GPG/Authenticode) don't really apply to the npm package as it just doesn't support this functionality. We'll have to revisit it if npm add extra security in the future, but it's fine for now.

PR #3872, opened 2017-07-09
Remove babel-plugin-transform-es2015-typeof-symbol

voxsim:
**Summary**
Remove `babel-plugin-transform-es2015-typeof-symbol` from package.json because it is not used.

**Test plan**
Everything is working like before

cc @Daniel15 @BYK


kaylieEB:
looks good to me! 👍

PR #3873, opened 2017-07-09
Remove babel-jest from package.json

voxsim:
**Summary**
This is an experiment: it seems that if we have jest, we don't need to declare babel-jest as our devDependency.

**Test plan**
The same tests.


kaylieEB:
thanks--it seems to be the case according to the docs! 🚢

PR #3874, opened 2017-07-09
Remove babel-eslint

voxsim:
**Summary**
Remove `babel-eslint` because we don't use it.

**Test plan**
The same tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
   },
   "devDependencies": {
     "babel-core": "^6.24.1",
-    "babel-eslint": "^7.2.3",
     "babel-loader": "^6.2.5",
     "babel-plugin-array-includes": "^2.0.3",
     "babel-plugin-transform-inline-imports-commonjs": "^1.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -267,7 +267,7 @@ babel-core@^6.0.0, babel-core@^6.0.2, babel-core@^6.24.1:
     slash "^1.0.0"
     source-map "^0.5.0"
 
-babel-eslint@^7.1.1, babel-eslint@^7.2.3:
+babel-eslint@^7.1.1:
   version "7.2.3"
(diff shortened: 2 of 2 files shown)


kaylieEB:
We need `babel-eslint` as a parser for eslint plugins like flowtype and it's a peer dependency of other packages we use.

voxsim:
Mmm I check https://github.com/gajus/eslint-plugin-flowtype/blob/master/package.json ant It seems it has only `eslint` as a peer dependency.
I will try to see if i make something wrong for flow, if everything work as expected

voxsim:
@kaylieEB @BYK I checked with deleting a https://github.com/gajus/eslint-plugin-flowtype#require-return-type and everything works like before. Now we are using `babel-eslint@^7.1.1` that it is a dependency of `eslint-config-fb-strict@^20.0.1` do we don't need to declare in our package.json IMHO.

kaylieEB:
So I had an issue with `babel-eslint` due to it being a peerDep of [eslint-config-fbjs](https://github.com/facebook/fbjs/blob/master/packages/eslint-config-fbjs/package.json#L23), ([CI error](https://circleci.com/gh/yarnpkg/yarn/4307?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)) but if `eslint-config-fb-strict`, has a strict dependency on that config package well as all of its correct peers including babel-eslint, we should be good. It actually makes more sense to remove it. Thanks for looking into this. I'll merge once CI passes.

voxsim:
Mmh the problems is that one test fail 😱  I checked on master and it fails on that branch too

kaylieEB:
hmm :( I don't have the permissions to merge without the CI passing. Maybe ask @BYK or try restarting the build with a new PR?

voxsim:
@kaylieEB I can merge it in case, but I prefer a double check from @BYK before ;)

BYK:
Hey there and sorry folks for the CI disruption. I'll see if I can get master green first so we can confidently merge this PR and others :)

BYK:
Hey there, can you rebase?

voxsim:
@BYK done ;)

Issue #3801, opened 2017-07-02
Yarn installs old versions in .bin


domoritz:
Reproduce with

```
git clone git@github.com:vega/vega-lite.git
cd vega-lite
rm yarn.lock
yarn
node_modules/.bin/uglifyjs --version
```

Prints `uglify-js 2.8.29` but should print version 3.

bestander:
Can repro this with latest Yarn.
Worked in 0.24

bestander:
```
uglifyjs -> ../handlebars/node_modules/uglify-js/bin/uglifyjs
```

bestander:
package-linker.js `determineTopLevelBinLinks` has incorrect logic inside.
It needs to go through top level dependencies first and then through the flat tree.

uhoh-itsmaciek:
As another data point, 0.24.6 works correctly for me, and 0.25.4 breaks (it means our yarn script uses the wrong eslint version and breaks our linting).

pselden:
Also, if there is one without a bin and one with, it won't even put anything in .bin.

Example:

```json
{
  "name": "test-yarn",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "sane": "^2.0.0",
    "watch": "^1.0.2"
  }
}
```

sane depends on watch 0.10.0, which has no .bin, but watch 1.0 does. No watch command is added to node_modules/.bin

timfish:
Our CI builds started failing only on linux because of this as our provider updated yarn to v0.27.5.
This repo reproduces the issue: https://github.com/timfish/yarn-test

Running `yarn build:html` used to work and now it fails as the version of `html-minifier` in `.bin` is an older version which is a dependency of `aurelia-bunder`.

kaylieEB:
I have a working PR to fix this issue here https://github.com/yarnpkg/yarn/pull/3877

PR #3310, opened 2017-05-03
link bins of transitive deps to top level

rally25rs:
**Summary**

Add transitive dependency bin symlinks to top-level `node_modules/.bin` to be more consistent with NPM behavior.

Fixes #2874 
Replaces #3272 

`package-linker` now iterates over the flat/hoisted tree and finds the newest version of each package. Each of those packages gets its bins linked at the top-level.

After that, during `PackageLinker.save()` each direct dependency has its bins linked (this was existing behavior, so unchanged) which will overwrite any existing bin links from transitive deps. This ensures that direct dependencies always take priority over the transitive ones.

**Test plan**

Stopped skipping a previously removed test, and added an additional test in `__tests__/commands/install/integration.js`

PR #3877, opened 2017-07-10
Prioritize bin links at the root level

kaylieEB:
**Summary**
This is a fix for https://github.com/yarnpkg/yarn/issues/3801
It seems like there's some history behind this issue, namely

**https://github.com/yarnpkg/yarn/pull/3310**, which 1) looked through all transient dependencies, 2) linked them to node_modules/.bin, then 3) overrode them with any top level dependencies

**https://github.com/yarnpkg/yarn/pull/3545**, removed the last part from above, which I assume was due to it incorrectly detecting what is actually at top level.

So now, the flat tree from `PackageHoister` gives us a list of all the bin paths that are hoisted to top level, but will include same names if their version ranges don't match, since hoisted has a different meaning from being added as an actual top level dependency. The latest implementation before this PR relied on the whatever order it came from and satisfied the bin link with the first one it found. 
So solution was to still allow for the override to happen when the package has `isDirectRequired` equal to true.

@bestander @rally25rs feel free to jump in with any additional knowledge :)

**Test plan**
Added tests to actually read versions of the bin links. It passes all the previous test cases with eslint, and passes the new test with uglifyjs bin that fails on master. (Got lucky that we already had uglifyjs tarball in the test fixtures!)

**BEFORE**
screen shot 2017-07-09 at 5 16 40 pm

**AFTER**
screen shot 2017-07-09 at 5 17 43 pm

Eslint behavior is the same as before

screen shot 2017-07-09 at 5 26 30 pm

### DIFF
--- a/__tests__/commands/install/bin-links.js
+++ b/__tests__/commands/install/bin-links.js
@@ -6,6 +6,7 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
 
 const request = require('request');
 const path = require('path');
+const exec = require('child_process').exec;
 import {runInstall} from '../_helpers.js';
 
 async function linkAt(config, ...relativePath): Promise<string> {
@@ -18,6 +19,36 @@ async function linkAt(config, ...relativePath): Promise<string> {
     const contents = await fs.readFile(joinedPath);
     return /node" +"\$basedir\/([^"]*\.js)"/.exec(contents)[1];
   }
+}
+
+function execCommand(cwd: string, binPath: Array<string>, args: Array<string>): Promise<Array<?string>> {
+  const cmd = path.join(...binPath);
+
+  return new Promise((resolve, reject) => {
+    exec(
+      `${cmd} ${args.join(' ')}`,
+      {
+        cwd,
+        env: {
+          ...process.env,
+          YARN_SILENT: 0,
+        },
+      },
+      (error, stdout) => {
+        if (error) {
+          reject({error, stdout});
+        } else {
+          const stdoutLines = stdout
+            .toString()
+            .split('\n')
+            .map((line: ?string) => line && line.trim())
+            .filter((line: ?string) => line);
+
+          resolve(stdoutLines);
+        }
+      },
+    );
+  });
 }
 
 beforeEach(request.__resetAuthedRequests);
@@ -49,6 +80,8 @@ test('direct dependency bin takes priority over transitive bin', (): Promise<voi
     expect(await linkAt(config, 'node_modules', 'standard', 'node_modules', '.bin', 'eslint')).toEqual(
       '../eslint/bin/eslint.js',
     );
+    const stdout = await execCommand(config.cwd, ['node_modules', '.bin', 'eslint'], ['--version']);
+    expect(stdout[0]).toEqual('v3.7.0');
   });
 });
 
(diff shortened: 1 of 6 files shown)


src/package-linker.js
```diff
@@ -380,15 +379,15 @@ export default class PackageLinker {
     }
   }
 
-  determineTopLevelBinLinks(flatTree: HoistManifestTuples): HoistManifestTuples {
+  determineTopLevelBinLinks(flatTree: HoistManifestTuples): Array<[string, Manifest]> {
```
arcanis: Does using `HoistManifestTuples` break something?

arcanis:
I've merged this PR since it would be nice to ship it along with the 0.28.1 :)

__tests__/commands/install/bin-links.js
```diff
@@ -20,6 +21,36 @@ async function linkAt(config, ...relativePath): Promise<string> {
   }
 }
 
+function execCommand(cwd: string, binPath: Array<string>, args: Array<string>): Promise<Array<?string>> {
```
BYK: I remember seeing a function like this in somewhere else. May be we should start using something like https://yarnpkg.com/en/package/jsinspect to help catch these automatically? (also it would help me be sure instead of saying "I suspect" :D)

__tests__/package-hoister.js
```diff
@@ -106,7 +106,7 @@ test('Produces valid destination paths for scoped modules', () => {
     _reference: (({}: any): PackageReference),
   }: any): Manifest);
 
-  const info = new HoistManifest(key, parts, pkg, '', true, false);
+  const info = new HoistManifest(key, parts, pkg, '', true, true, false);
```
BYK: We really should convert this method to accept and options object instead of boolean arguments. No idea what's happening here :(

rally25rs:
Oh man, this .bin link thing has been such a pain... So the bin linking behavior in <=0.24 was to only link the top-level bins.

In 3535 (which went out as part of v0.25) I started linking the transient deps, but didn't want to change too much code and dig myself a hole I couldn't get out of, so based it on the reference counts that were being used to determine which packages to hoist; Each dep version with the highest reference count would be hoisted the furthest up the tree, so would get a .bin link at top level.

However I also left the old/existing logic in place that had already been linking just the direct dependencies, so that no matter what dep had the highest ref count, which could be a transient dep, it would get overridden by the direct dep, just like <=v0.24 had linked.

3332 removed that last part, so would certainly result in transient deps being .bin linked instead of direct-deps. I think that might be a naming issue on my part that caused some confusion. The function `determineTopLevelBinLinks()` is used to "determine which transient deps should be linked at the top level (/node_modules/.bin)", not to determine the top-level modules.

I hadn't seen 3332 when it got merged, but it probably indicates that <=v0.24 was linking wrong bins too, since the code that got deleted there was the original (before I messed with it) code that did the bin links for the direct dependencies.

I hope you're attempt to fix these bin links has better luck than mine...

__tests__/commands/install/bin-links.js
kaylieEB: Yeah I copied it over from the test file that was testing `run` command and tweaked it for my purpose. I think it was out of scope for this diff but the right thing to do is probably have a place to store and dedupe all of these similar functions (you're right--I've definitely seen it being reused in several other test files!) inside some shared test utils folder.

On a separate note, that looks like a cool tool! I think I've seen similar functions under src/ which is probably more important to refactor :)

src/package-hoister.js
```diff
@@ -107,19 +117,12 @@ export default class PackageHoister {
...
         if (info) {
-          infos.push(info);
+          this.hoist(info);
         }
       }
-
-      //
-      for (const info of infos) {
-        this.hoist(info);
-      }
```
kaylieEB: fyi, this was unrelated, (so is the for loop below) but the second iteration really didn't seem necessary. lmk if they're need for some reason.

src/package-linker.js
kaylieEB: No but this function was only used once below, where it just pulled out the Manifest (`pkg`), and I thought it'd be better to reduce the size of object being passed from the HoistManifest. If you think we need to keep the previous signature, happy to send a refactor PR.

BYK:
>sorry to get to the comments after PR closed!

No worries at all :)

PR #3896, opened 2017-07-10
Link binaries defined in workspace packages

### LINKED ISSUE #3846: Workspace bin files aren't linked or installed in root

When running `yarn install` in a package with workspaces, if any of the workspace packages declare a `bin` in their package.json it's ignored when hoisting dependencies 

example repo here: https://github.com/jquense/workspaces-bin-repo

As a little aside if I used a published package name for my test workspaces the local folder would get replaced with the package in the npm registery. Not related to this bug, but maybe a bug anyway.

That `my-test-foo` package's bin will be accessible from packages that depend on it.

node 8.1
yarn 27.5
OSX Sierra

jgoz:
**Summary**

Bin files defined in workspace packages were not being linked into dependent packages because the mock package in `package-fetcher.js` [did not include `"bin"`](https://github.com/yarnpkg/yarn/blob/a3ce7c702f644efde783beb8e0b99dc08100f0df/src/package-fetcher.js#L29-L32) for `workspace` or `link` remote types.

This adds a `WorkspaceFetcher` implementation that reads the `package.json` file in the workspace, overriding the `version` and `_uid` fields similar to the mock package. But all the other fields are preserved, including `"bin"`, which allows the linker to do its thing.

Fixes #3846.

**Test plan**

Added a test case in `commands/install/workspaces-install.js` and verified the correct behaviour using the test repo linked in #3846 (https://github.com/jquense/workspaces-bin-repo).

Note that the test failures in this branch existed even before I made any changes. AFAIK, no existing tests fail as a result of this change.


src/fetchers/workspace-fetcher.js
```diff
@@ -0,0 +1,49 @@
...
+  dest: string;
+  registry: RegistryNames;
+  workspaceDir: string;
+
+  setupMirrorFromCache(): Promise<?string> {
+    return Promise.resolve();
+  }
+
+  async fetch(defaultManifest: ?Object): Promise<FetchedMetadata> {
+    const pkg = await (async () => {
```
arcanis: Why an IIFE, since `fetch` is already synchronous? :o

jgoz: Whoops, `fetch` should not be `async`.

arcanis: Why that, since it returns a promise?

jgoz: My mistake, was looking at the BaseFetcher implementation. This can be `async` and I removed the IIFE.

jgoz:
@arcanis Addressed your feedback.

arcanis:
Waiting for the tests to end, then I'll merge it. Good job!

Issue #3727, opened 2017-06-26
Consistency issue: why diff command -> ls in global and list in normal mode


sakivks:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Please mention your node.js, yarn and operating system version.**

bestander:
We are going away from shorthand commands.
Should be list everywhere

sakivks:
That was the issue 
shorthand(ls) works but list doesn't work in global
![screen shot 2017-06-27 at 12 49 40 am](https://user-images.githubusercontent.com/9475630/27556336-cac9d368-5ad2-11e7-84e9-c7f10c617bd0.png)

bestander:
This should be fixed in 0.26 https://github.com/yarnpkg/yarn/pull/3308

PR #3898, opened 2017-07-11
Breaking: Deprecate yarn <command> ls in favor of yarn <command> list

voxsim:
**Summary**
This pull request deprecates every `yarn <command> ls` in favor of `yarn <command> list`

**Test plan**
I check every command and every test that took these commands

If you like it, do you think we should do the same for `yarn <command> rm`? And delete everything deprecated when we roll out the version 1.0?

### DIFF
--- a/__tests__/commands/cache.js
+++ b/__tests__/commands/cache.js
@@ -20,11 +20,11 @@ const runCache = buildRun.bind(
   },
 );
 
-test('ls', async (): Promise<void> => {
+test('list', async (): Promise<void> => {
   await runInstall({}, 'artifacts-finds-and-saves', async (config): Promise<void> => {
     const out = new stream.PassThrough();
     const reporter = new reporters.JSONReporter({stdout: out});
-    await run(config, reporter, {}, ['ls']);
+    await run(config, reporter, {}, ['list']);
     const stdout = String(out.read());
     expect(stdout).toContain('dummy');
     expect(stdout).toContain('0.0.0');
@@ -35,7 +35,7 @@ test('ls with scoped package', async (): Promise<void> => {
   await runInstall({}, 'install-from-authed-private-registry', async (config): Promise<void> => {
     const out = new stream.PassThrough();
     const reporter = new reporters.JSONReporter({stdout: out});
-    await run(config, reporter, {}, ['ls']);
+    await run(config, reporter, {}, ['list']);
     const stdout = String(out.read());
     expect(stdout).toContain('@types/lodash');
     expect(stdout).toContain('4.14.37');
--- a/__tests__/commands/licenses.js
+++ b/__tests__/commands/licenses.js
@@ -20,7 +20,7 @@ const runLicenses = buildRun.bind(
 );
 
 test('lists all licenses of the dependencies with the --json argument', async (): Promise<void> => {
-  await runLicenses(['ls'], {json: true}, '', (config, reporter, stdout) => {
+  await runLicenses(['list'], {json: true}, '', (config, reporter, stdout) => {
     expect(stdout).toContain(JSON.stringify(expectedTable));
   });
 });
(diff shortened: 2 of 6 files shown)


arcanis:
Hm, `ls` is one of the rare alias (with `rm`) that I think we really should support. To some people (I'm one of them :D) they are *the* right way to list and remove things when working with the command line.

voxsim:
@arcanis I am a unix fan and I totally agree, but yesterday i saw that we deprecate `yarn global ls` in favor of `yarn global list`, so this pull request is for consistency between different commands

voxsim:
We can always choose to support both :)

arcanis:
I think we should, wdyt @BYK?

voxsim:
@arcanis @BYK this is the original issue https://github.com/yarnpkg/yarn/issues/3727 (see the comment of @bestander :p)

BYK:
I'd rather support `list` only for simplicity. LGTM but please make sure you tag the commit with `Breaking: ` when merging. We should probably be vocal about this breaking change with the next release (if that's not 1.0)

voxsim:
Yep ok :)

bestander:
Yeah, let's make it all clear and unambiguous.
People can have their own short commands as described here https://yarnpkg.com/blog/2017/06/19/adding-command-line-aliases-for-yarn/

bestander:
Thanks, @voxsim

Daniel15:
Shouldn't we also rename "rm" to "remove" then?

bestander:
We should

BYK:
Filed #3952.

tpina:
Hi @BYK, @bestander  and @voxsim 

`ls` command has not been deprecated for `yarn team`, should I raise an issue? or should I just go ahead a create the PR?

Can I also, as part of the same PR, do a small fix of the command template description for: `yarn owner` and `yarn tag`?

BYK:
@tpina I love PRs a lot more than issues. And if that template description fix is relevant for this fix, go ahead and bundle it. Otherwise, it should be in its own PR.

tpina:
@BYK PR will follow soon then. Thank you.

mikemaccana:
This is inconsistent with every other Unix program. Please just make an alias so `ls` works again.

bestander:
Yarn is not a unix program, if you want to have your favorite shortcut https://yarnpkg.com/blog/2017/06/19/adding-command-line-aliases-for-yarn/

mikemaccana:
@bestander I'm aware of that - I run yarn on Windows, which also implements `ls` in it's default shell and spells it consistently with all other programs too.  

I'm also aware of how to make an alias. Here's mine: 

	# yarn broke 'ls'
	# Scope private do we don't call yarn recursively!
	function Private:yarn() {
		$modifiedArgs = @()
		foreach ( $arg in $args ) {
			if ( $arg -cmatch '^ls$' ) {
				$arg = 'list'
			}
			$modifiedArgs += $arg
		}
		& yarn $modifiedArgs
	}
	echo 'Mike profile loaded.'

Which do you think is better @bestander?:

 - Every user with experience in a command line OS learns a new way of spelling 'list' and/or makes an alias
 - yarn implements an alias, because it cares about its users?

bestander:
@mikemaccana, thanks that is a very nice alias, please add a section with this example to https://yarnpkg.com/blog/2017/06/19/adding-command-line-aliases-for-yarn/.

Yarn cares about the user experience very much, this is why it was created.
But we have made a decision to avoid CLI shortcuts early on and still stand by it.
Otherwise we were spending too much time debating what -v should stand for: `version` or `verbose`.

mikemaccana:
@bestander Sure, done. 

Back on topic, I'm not sure avoiding shortcuts helps the experience. If you say 'list' to a Unix or Powershell user they generally spell it 'ls'. But my point is already made and I'll keep quiet for now.

mikemaccana:
I noticed `yarn init` still exists. Should this be deleted and replaced with `yarn initialize`?

bestander:
Good point, you got me :)

PR #3899, opened 2017-07-11
Fix: fix failing tests on master after release changes

BYK:
**Summary**

Fixes the failing builds all over the place.

**Test plan**

Builds shall pass.


__tests__/commands/info.js
```diff
@@ -42,16 +42,18 @@ const expectedKeys = [
   'bugs',
   'license',
   'scripts',
-  'dependencies',
-  'devDependencies',
   'dist',
   'directories',
 ];
 
+// yarn now ships as built, single JS files so it has no dependencies
```
Daniel15: I'm still confused as to why this wasn't broken in my original PR? The builds on #3866 are clean :/

BYK: It only started to fail after you've published that version since these tests were doing an `info` call on `yarn` which was published. Not really your fault :)

We should probably do those tests offline.

Daniel15: Ohhhhh... Wow, I didn't realise it was checking live data.

BYK: Yeah. I should go over most of the tests including git ones and make everything offline if possible leaving very few actual e2e tests.

PR #3902, opened 2017-07-11
Lower platform incompatibility warning to info for optional dependencies

### LINKED ISSUE #3738: Don't warn about incompatible optional dependencies
Currently Yarn behavior matches npm in that it warns about optional dependencies that are incompatible (and thus skipped). `fsevents` on Windows is a good example of this.

Since the warning is almost never actionable, and looks a bit frightening to beginners, I propose that we downgrade its log level. I'm not sure if Yarn even has a concept of log levels, but I wouldn't want to see this warning unless I'm specifically debugging Yarn itself.

Corresponding npm issue has wide support (although it got autoclosed): https://github.com/npm/npm/issues/11632

onurtemizkan:
**Summary**
Resolves #3738

Screenshots:
then : screen shot 2017-07-11 at 20 40 03
now : screen shot 2017-07-11 at 20 36 24

**Test plan**
This part doesn't seem to have tests yet so I didn't make any updates about tests. I'd be happy to add tests for `checkOne` and the other untested functions though.

### DIFF
--- a/src/package-compatibility.js
+++ b/src/package-compatibility.js
@@ -110,7 +110,7 @@ export function checkOne(info: Manifest, config: Config, ignoreEngines: boolean)
       ref.ignore = true;
       ref.incompatible = true;
 
-      reporter.warn(`${human}: ${msg}`);
+      reporter.info(`${human}: ${msg}`);
       if (!didIgnore) {
         reporter.info(reporter.lang('optionalCompatibilityExcluded', human));
         didIgnore = true;


voxsim:
@onurtemizkan can you share a screenshot of before/after the change? This would help us to understand what exactly change.

onurtemizkan:
@voxsim, I added the screenshots to the first post

voxsim:
Thanks @onurtemizkan! i will merge it!

jhabdas:
Please consider requiring unit tests for all work.  One should not justify not providing tests simply because they do not currently exist.

voxsim:
@jhabdas we usually require unit tests, but imho this thing is not useful to test..  Everytime we have a test that covers something, this should add value, otherwise it is a waste of effort

onurtemizkan:
@jhabdas I changed a little part of a function that is untested, I mentioned that I'm volunteering to test all functions there in my first post, but also think that this is out of the scope of this small update and belongs to another PR.

javatheist:
@voxsim as I don't know who to label, I just picked your name from this list of contributors.

First of all, thank you (pars pro toto) for quickfixing this. As a matter of fact, it's not done yet. A fresh run of installation works well as depicted above, but if you need to re-run the installation, it show up as an error again due to the fact that it is listed in the resolutions section of the package.json and thus the process breaks. A solution to succeed is to remove the resolution entry each time, but as you can image, this is unsatisfactory. Just wanted you and the community to know that.

Regards

PR #3904, opened 2017-07-11
Breaking: Remove unsupported aliases

### LINKED ISSUE #3892: Remove every unsupported-alias
We should remove every unsupported-alias (https://github.com/yarnpkg/yarn/blob/master/src/cli/unsupported-aliases.js) from the codebase.

onurtemizkan:
**Summary**
Resolves #3892

**Test plan**
I removed `unsupported-aliases` tests too


BYK:
According to CircleCI you missed some tests :)

You can run tests locally by running `yarn build; yarn test`

BYK:
@voxsim I labeled this as "breaking". Do you agree?

voxsim:
@BYK yes, in some sense is a breaking change ;)

voxsim:
Thanks @onurtemizkan !! I will merge it :)

Issue #3821, opened 2017-07-04
Yarn global not detecting license in package.json


OmgImAlexis:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Installing global packages doesn't always detect license field.

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn global add xo
```

```
➜  agenda git:(feature/update_dev_deps) ✗ yarn global add xo          
yarn global v0.27.5
warning package.json: No license field
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Installed "xo@0.18.2" with binaries:
      - xo
warning No license field
Done in 47.10s.
➜  agenda git:(feature/update_dev_deps) ✗ which xo
/usr/local/bin/xo
➜  agenda git:(feature/update_dev_deps) ✗ realpath /usr/local/bin/xo
/Users/xo/.config/yarn/global/node_modules/xo/cli.js
➜  agenda git:(feature/update_dev_deps) ✗ cat /Users/xo/.config/yarn/global/node_modules/xo/package.json | grep license
  "license": "MIT",
```

**What is the expected behavior?**
For it to install without a warning about missing license.

**Please mention your node.js, yarn and operating system version.**
Node: v8.1.3
Yarn: 0.27.5
OS: macOS Sierra 10.12.3 (16D32)

GAumala:
I tried to debug this, and it seems that what yarn is complaining about is not the license for package `xo`, but the license for the user's `package.json` file. When `yarn global` is used, yarn keeps a `package.json` with all the packages that the user has installed globally listed as dependencies.  Right now, mine looks something like this:

```json
{
  "dependencies": {
    "create-react-app": "^1.3.1",
    "typescript": "^2.3.2",
    "xo": "^0.18.2"
  }
}
```
Those 3 are the only packages that I have installed globally in my system. As you can see, this file does not have a license, not even a name, so that's why a warning is thrown. Maybe the `validate` function in `normalize-manifest` could receive an extra parameter to skip this check when running `yarn global`.

BYK:
>Maybe the validate function in normalize-manifest could receive an extra parameter to skip this check when running yarn global.

@bestander @arcanis I think this is the way to go. What do you think?

olingern:
@BYK  Looks like this can be closed a la #3921

kaylieEB:
[...]

PR #3921, opened 2017-07-13
#3821 Remove manifest check for global commands

kaylieEB:
**Summary**
From #3821
When `global` command is used, skip the validation check for manifest, such as missing license or name field listed here https://github.com/yarnpkg/yarn/blob/master/src/util/normalize-manifest/validate.js#L44-L98

**Test plan**
Added a test in `/__tests__/normalize-manifest.js`, refactored out the test function used for non-global commands

*BEFORE*
screen shot 2017-07-12 at 8 29 20 pm

*AFTER*

screen shot 2017-07-12 at 8 27 58 pm

### DIFF
--- a/__tests__/normalize-manifest.js
+++ b/__tests__/normalize-manifest.js
@@ -15,6 +15,51 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
 
 const fixturesLoc = path.join(__dirname, 'fixtures', 'normalize-manifest');
 
+async function _compareManifests(loc, options = {}): Promise<void> {
+  const actualWarnings = [];
+  const expectedWarnings = options.expectedWarnings || (await fs.readJson(path.join(loc, 'warnings.json')));
+  const reporter = new NoopReporter();
+
+  // $FlowFixMe: Investigate
+  reporter.warn = function(msg) {
+    actualWarnings.push(msg);
+  };
+
+  const config = await Config.create({cwd: loc, globalFolder: options.globalFolder}, reporter);
+
+  let actual = await fs.readJson(path.join(loc, 'actual.json'));
+  const expected = await fs.readJson(path.join(loc, 'expected.json'));
+
+  let isRoot = actual._root;
+  if (isRoot == null) {
+    isRoot = true;
+  } else {
+    delete actual._root;
+  }
+
+  const error = expected._error;
+  if (error) {
+    delete expected._error;
+  }
+
+  try {
+    actual = await normalizeManifest(actual, loc, config, isRoot);
+  } catch (err) {
+    if (error && err.message === error) {
+      return;
+    } else {
+      throw err;
+    }
+  }
+
+  if (error) {
+    throw new Error(`Expected to throw error: ${error}`);
+  }
+
+  expect(map(actual)).toEqual(expand(expected));
+  expect(actualWarnings).toEqual(expectedWarnings);
+}
+
 for (const name of nativeFs.readdirSync(fixturesLoc)) {
   if (name[0] === '.') {
     continue;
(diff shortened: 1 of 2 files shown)


arcanis:
Hmm. Technically, a package with no license is by default privative, and probably shouldn't be used by anyone except its author. I can see the point for adding a warning about it - if we don't, who's gonna notice it? :|

kaylieEB:
hey @arcanis, I thought this was an ask because global packages aren't meant to be used with package.json https://github.com/npm/npm/issues/2949#issuecomment-11408461. I got the warning above despite not even having a package.json in the location where my global bins are installed, and even when I created one with a valid license, the warning didn't disappear. Also, I'm unsure what would be the reasoning for publishing your own global dependencies since they're generally meant for running binaries like grunt, eslint, etc. It seems like it's a wrong use case to support whereas the warning probably happens too frequently for everyday use, and yarn doesn't create/add to package.json from `yarn global add`. What do you think?

kaylieEB:
Actually, rereading your comment--I think the confusion maybe that the current behavior validates the package.json in cwd itself and not for the manifests of its dependencies, but maybe that's a bug itself?

arcanis:
> I got the warning above despite not even having a package.json in the location where my global bins are installed, and even when I created one with a valid license, the warning didn't disappear.

It's because the license field should be inside the package.json of the global dependency you've just added, not inside a package.json where your local bins are installed.

> Also, I'm unsure what would be the reasoning for publishing your own global dependencies since they're generally meant for running binaries like grunt, eslint, etc.

Well, someone could just publish their own tools because they find it easier to install them later, but that's not the real issue. Let's say I create a `super-cli` package, everybody loves it. That's great! Except that when I published it, I forgot to put a license field. Now, if I'm a not-very-great-person, I can probably fill claims against companies and startups who happen to use my project, because I never explicitly allowed them to do so.

arcanis:
> I think the confusion maybe that the current behavior validates the package.json in cwd itself and not for the manifests of its dependencies, but maybe that's a bug itself?

Oh! I see what you mean :o Yeah, that might be a bug in itself, you're right. We shouldn't validate the cwd package.json when running a `yarn global` command.

__tests__/normalize-manifest.js
```diff
@@ -15,60 +15,71 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
...
+const globalfixturesLoc = path.join(__dirname, 'fixtures', 'normalize-manifest', 'name');
+
+test(
+  'license warnings should not be thrown with global commands',
+  _compareManifests.bind(null, globalfixturesLoc, {
+    expectedWarnings: [],
+    globalFolder: globalfixturesLoc,
+  }),
+);
+
```
arcanis: Did the old code break something?

kaylieEB: hmm? not sure if I understand..I moved the predicate passed inside the for loop in line 25 to its own function so I can reuse it to test the one for global.

kaylieEB:
Shouldn't we validate on normal `add` as well then?  I can also see value in validating the current package.json when you're running `yarn init` (it seems the file was originally intended for this use case, [detecting typos](https://github.com/yarnpkg/yarn/blob/master/src/util/normalize-manifest/validate.js#L48), [replacing fields](https://github.com/yarnpkg/yarn/blob/master/src/util/normalize-manifest/validate.js#L80), and such)

__tests__/normalize-manifest.js
arcanis: Ok! :) I was trying to understand what changed in this file - I can see now that you've factored the test so that we can call it multiple times with different parameters, perfect 👍

arcanis:
I think that would make sense to add this behavior to `yarn add` as well, yep. Not sure about `init`, tho - I feel like it is more often used to bootstrap a new `package.json` rather than update or validate an existing one.

BYK:
@arcanis @kaylieEB any reason to not mergE?

arcanis:
Nope

PR #3924, opened 2017-07-13
Do not use /usr/local/share as home under fakeroot

### LINKED ISSUE #2536: Yarn fakeroot problems

Bug

I have a Debian package that invokes Yarn during the install step.

Under Yarn 0.18.1 and previous versions, this worked fine. Starting with Yarn 0.19.1, it fails:

```
dh binary
   dh_testroot
   dh_prep
   dh_auto_install
make[2]: Entering directory `<http://jenkins:8080/job/Nightly/ws/build-area/package-2.0.2'>
make install
yarn  --pure-lockfile
/usr/share/yarn/node_modules/mkdirp/index.js:90
                    throw err0;
                    ^

Error: EACCES: permission denied, mkdir '/usr/local/share/.cache'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:922:18)
    at sync (/usr/share/yarn/node_modules/mkdirp/index.js:71:13)
    at Function.sync (/usr/share/yarn/node_modules/mkdirp/index.js:77:24)
    at Object.<anonymous> (/usr/share/yarn/bin/yarn.js:30:8)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

`fakeroot yarn add lodash`

`fakeroot yarn` should work as well as running `yarn` by itself does.

I believe that this issue was introduced by #2150. Since fakeroot is [commonly used](http://manpages.ubuntu.com/manpages/zesty/man1/debuild.1.html#contenttoc5) in building Debian packages, 

fakeroot sets the `FAKEROOTKEY` environment variable (see [here](http://manpages.ubuntu.com/manpages/precise/man1/fakeroot-tcp.1.html)), so one possible fix is to only switch to `/usr/local/share` if `isRootUser(getUid()) && process.env.FAKEROOTKEY == null`.

Node 6.9.4
Yarn 0.19.1
Ubuntu 14.04

Tebro:
Fixes #2536

**Summary**

When packaging software that uses yarn for dependency management, some distributions use fakeroot as a part of their build system (my case, Solus OS). 

What happens is that the uid is set to 0, and because of that yarn tried to use /usr/local/share as the home directory. When yarn tried to create this directory it fails, because the user is not really root and does not have the permissions. 

This added check makes sure that we use the right home directory when running under fakeroot  

**Test plan**

This is what was executed (in the yarn codebase) to prove the difference:

With current yarn:

```
 ~/tmp/yarn $ fakeroot yarn
/usr/lib64/node_modules/yarn/node_modules/mkdirp/index.js:90
                    throw err0;
                    ^

Error: EACCES: permission denied, mkdir '/usr/local'
    at Object.fs.mkdirSync (fs.js:895:18)
    at sync (/usr/lib64/node_modules/yarn/node_modules/mkdirp/index.js:71:13)
    at sync (/usr/lib64/node_modules/yarn/node_modules/mkdirp/index.js:77:24)
    at sync (/usr/lib64/node_modules/yarn/node_modules/mkdirp/index.js:77:24)
    at Function.sync (/usr/lib64/node_modules/yarn/node_modules/mkdirp/index.js:77:24)
    at Object.<anonymous> (/usr/lib64/node_modules/yarn/bin/yarn.js:31:8)
    at Object.Module._extensions..js (module.js:580:10)
```

and with the modified yarn:

```
 ~/tmp/yarn $ fakeroot bin/yarn 
yarn install v0.28.0
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.1.1: The platform "linux" is incompatible with this module.
info "fsevents@1.1.1" is an optional dependency and failed compatibility check. Excluding it from installation.
[4/5] Linking dependencies...
warning "ajv-keywords@1.5.1" has incorrect peer dependency "ajv@>=4.10.0".
warning "eslint-config-fbjs@2.0.0-alpha.1" has incorrect peer dependency "babel-eslint@^7.1.1".
[5/5] Building fresh packages...
Done in 2.48s.
```

### DIFF
--- a/src/util/user-home-dir.js
+++ b/src/util/user-home-dir.js
@@ -4,7 +4,7 @@ import ROOT_USER from './root-user.js';
 
 const path = require('path');
 
-const isNotFakeRoot = typeof process.env.FAKEROOTKEY === 'undefined';
+const isNotFakeRoot = Boolean(process.env.FAKEROOTKEY);
 
 const userHomeDir =
   process.platform === 'linux' && ROOT_USER && isNotFakeRoot
--- a/src/util/user-home-dir.js
+++ b/src/util/user-home-dir.js
@@ -4,11 +4,7 @@ import ROOT_USER from './root-user.js';
 
 const path = require('path');
 
-const isNotFakeRoot = Boolean(process.env.FAKEROOTKEY);
-
 const userHomeDir =
-  process.platform === 'linux' && ROOT_USER && isNotFakeRoot
-    ? path.resolve('/usr/local/share')
-    : require('os').homedir();
+  process.platform === 'linux' && ROOT_USER ? path.resolve('/usr/local/share') : require('os').homedir();
 
 export default userHomeDir;
--- a/src/util/user-home-dir.js
+++ b/src/util/user-home-dir.js
@@ -4,7 +4,11 @@ import ROOT_USER from './root-user.js';
 
 const path = require('path');
 
+const isNotFakeRoot = typeof process.env.FAKEROOTKEY === 'undefined';
+
 const userHomeDir =
-  process.platform === 'linux' && ROOT_USER ? path.resolve('/usr/local/share') : require('os').homedir();
+  process.platform === 'linux' && ROOT_USER && isNotFakeRoot
+    ? path.resolve('/usr/local/share')
+    : require('os').homedir();
 
 export default userHomeDir;
--- a/src/util/root-user.js
+++ b/src/util/root-user.js
@@ -7,7 +7,11 @@ function getUid(): ?number {
   return null;
 }
 
-export default isRootUser(getUid());
+export default isRootUser(getUid()) && isNotFakeRoot();
+
+export function isNotFakeRoot(): boolean {
+  return Boolean(process.env.FAKEROOTKEY);
+}
 
 export function isRootUser(uid: ?number): boolean {
   return uid === 0;


arcanis:
I'm not familiar with `fakeroot` but I don't understand something. Shouldn't creating the share folder work? According to the documentation it should:

```
       Here is an example session with fakeroot.  Notice that inside the  fake
       root  environment  file manipulation that requires root privileges suc-
       ceeds, but is not really happening.
```

Tebro:
I am not familiar with it either, it is used in Solus build system, and this combination made yarn commands fail. This was encountered in the build system, but the commands listed in the Test plan above were run "normally".

src/util/user-home-dir.js
```diff
@@ -4,7 +4,11 @@ import ROOT_USER from './root-user.js';
 
 const path = require('path');
 
+const isNotFakeRoot = typeof process.env.FAKEROOTKEY === 'undefined';
```
BYK: Why not just `Boolean(process.env.FAKEROOTKEY)`?

BYK: Also, I'd argue that this check should go into `./root-user.js`

Tebro: I can understand the recommendation for `Boolean(...)` but as to moving the check into `root-user.js` I am not sure if I agree. It would make the imported `ROOT_USER` contain the combination of two checks, which at least from the current naming is not directly clear what exactly it is checking.

BYK: Agree with the concern but looks like `ROOT_USER` is only used here and with the assumption of root permissions so probably it's a good idea to put it there. @bestander @arcanis thoguhts?

Tebro: No I see, and agree with your point. Will add that change as well.

arcanis:
Are the packages generated by this fakeroot build supposed to be installed on a per-user basis, or on the whole system? (in the second case, we *do* want to use `/usr/local`)

BYK:
@arcanis sorry I missed your message. Please block the PR if you have concerns.

Tebro:
This should affect Debian packaging as well as Solus. We build the packages in a container, and the files that end up in the actual package are placed in a specific directory, $installdir. 

So let's say we have an application like the Atom editor. It would be built from source in the container, and the resulting build artifacts are placed in $installdir. For example the main binary would be placed in $installdir/usr/bin/atom. 

When the package is installed on a Solus installation the package manager would then unpack the files that were place in $installdir into their correct locations. (/usr/bin/atom for the example above)

Tebro:
@arcanis @BYK do you have further concerns about this?

BYK:
I don't. Was waiting for @arcanis to jump in if he had any more concerns but I'm assuming he's okay for now. Merging.

PR #3929, opened 2017-07-13
chore(lint): Integrate `js-inspect` into lint stage

### LINKED ISSUE #3903: Chore: add code duplication detection tool to lint step
Forking from https://github.com/yarnpkg/yarn/pull/3877#discussion_r126381876, I think it would be nice to add a code duplication detector to the lint stage (or as a separate stage) to avoid copy/pasting of commonly used functions.

A good project is https://yarnpkg.com/en/package/jsinspect but we may prefer https://yarnpkg.com/en/package/jscpd too.

onurtemizkan:
**Summary**

Fixes #3903.

1. `dupe-check` step is added to `lint` stage, but it only warns about possible duplicates, doesn't fail CI.
2. The threshold to detect duplication is set to 30, which is the default value. Currently, it reports 9 possible duplicates for yarn codebase. This would be set according to the accuracy of detections.
3. The reporter shows all the code that are possibly duplicates of each other as _diffs_. Thus, the reports can be very verbose. There is an option for suppressing diffs longer than `n` lines in reports. That can be considered.

**Test plan**

Run `yarn lint` and observe the duplicated code output.


package.json
```diff
@@ -93,7 +94,7 @@
     "build-dist": "bash ./scripts/build-dist.sh",
     "build-win-installer": "scripts\\build-windows-installer.bat",
     "check-lockfile": "./scripts/check-lockfile.sh",
-    "lint": "eslint . && flow check",
+    "lint": "eslint . && flow check && gulp dupe-check",
```
BYK: Why not just use the CLI interface directly instead of going through `gulp`?

onurtemizkan:
@BYK I updated the PR, but fetch errors failed CI.

BYK:
@onurtemizkan can we try merging from master, resolving the conflicts and try this again?

BYK:
@onurtemizkan ping?

onurtemizkan:
@BYK, sorry for the late response. I've just updated the branch.

package.json
```diff
@@ -105,7 +106,9 @@
     "build-deb": "./scripts/build-deb.sh",
     "build-dist": "bash ./scripts/build-dist.sh",
     "build-win-installer": "scripts\\build-windows-installer.bat",
-    "lint": "eslint . && flow check",
+    "check-lockfile": "./scripts/check-lockfile.sh",
```
BYK: This is a merge error. This line is removed on latest master :)

package.json
```diff
@@ -28,6 +28,7 @@
     "is-builtin-module": "^1.0.0",
     "is-ci": "^1.0.10",
     "is-webpack-bundle": "^1.0.0",
+    "jsinspect": "^0.12.6",
```
BYK: Ah, one more. This should be in `devDependencies`.

PR #3938, opened 2017-07-15
Publish npm release of Yarn directly from CircleCI

Daniel15:
**Summary**
This is a followup to #3866. I originally moved the Yarn npm publish to our external build infra. After thinking about this for a while, I'm going to move this particular step back to publishing directly from CircleCI. The reasons that the other releases are pushed from infra under our own control (eg. signing them with GPG/Authenticode) don't really apply to the npm package as it just doesn't support this functionality. 

We'll have to revisit it if npm add extra security in the future, but it's fine for now.

**Test plan**

This is difficult to test without publishing a new Yarn version. I tested the script locally by commenting out the check for if the specific version already exists on npm, and it worked up until the point of publishing (which obviously threw an error "You cannot publish over the previously published version 0.28.1.")

### DIFF
--- a/scripts/update-npm.sh
+++ b/scripts/update-npm.sh
@@ -1,16 +1,16 @@
 #!/bin/bash
-# Pushes the latest Yarn version to npm
+# Pushes the locally built Yarn version to npm
 
 set -ex
 
-# See if YARN_VERSION was passed in the environment, otherwise get version
-# number from Yarn site
-if [ -z "$YARN_VERSION" ]; then
-  echo 'Getting Yarn version from https://yarnpkg.com/latest-version'
-  version=`curl --fail https://yarnpkg.com/latest-version`
-else
-  version="$YARN_VERSION"
-fi
+version=`./dist/bin/yarn --version`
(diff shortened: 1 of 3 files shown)


scripts/update-npm.sh
```diff
@@ -20,14 +20,11 @@ if [ "`npm view yarn@$version name`" = 'yarn' ]; then
   exit 0
 fi;
 
-# Grab the tarball so we can publish it as-is
-url=https://yarnpkg.com/downloads/$version/yarn-v$version.tar.gz
-tempfile=`mktemp -t 'yarn-release-XXXXXXXX.tar.gz'`
-curl --fail -L -o $tempfile $url
-
+# Determine if this is an RC or a stable release
+release_type=`curl --fail https://release.yarnpkg.com/release_type/$version`
```
Daniel15: This endpoint simply returns "stable" or "rc" based off the version number passed in. In theory we could do the version comparison directly in this script, but it would still need to hit an external URL (`https://yarnpkg.com/latest-version`) to get the latest version and determine if this release is RC or stable, so I'd rather keep this simply and reuse the existing logic.

Issue #3787, opened 2017-07-01
Prefix added to version of packages in package.json


rockchalkwushock:
```plaintext
yarn@0.27.5
node@8.1.3
npm @5.0.3
macOS@10.12.5

yarn bin: /Users/codybrunner/node_modules/.bin
yarn global bin: /Users/codybrunner/.node_modules_global/bin
```

I did not see a similar issue opened and this could just be based on my own local settings with `yarn`. Seems after upgrading to `@0.27.5` whenever I install a package I get the following:

```js
{
  dependencies: {
    react: "/Users/codybrunner15.16.1" // expect react: "~15.16.1"
  }
}
```

My guess is that `yarn` is now using the whole prefix of where to find the cached modules on my machine. Not very well versed in my understanding of how some of that works 😕 . Figured I'd report this...if it is my own settings causing this feel free to close.

rockchalkwushock:
I just noticed that support was added in `@0.27.0` for `~` to be seen as `$HOME`. I'm now wondering if the fact I have `save-prefix: "~"` could be resulting in a translation from `~15.16.1` to `/Users/codybrunner15.16.1`. I'm going to check this out and get back with an answer soon 😉

rockchalkwushock:
Yep absolutely the problem!

After running:

```sh
yarn config set save-prefix "^"
mkdir test
cd test
yarn init
yarn add react
```

## Result

```js
{
  "dependencies": {
    "react": "^15.16.1" // expected result!
  }
}
```

Unfortunately I have not the slightest idea of how to resolve something like this in the underlying source code :confused:; but at least I now know that the change made in `@0.270` to support `~` being seen as `$HOME` is the underlying issue. I will leave this to someone on the `yarn` team or someone that knows more about how and where to fix this in the source code...if it's something that should be fixed.

arcanis:
I guess the easy fix would be to check for `~/` instead of `~`. Thanks for investigating!

rockchalkwushock:
@arcanis 

I figured it would be something like this. I've never done much with CLI related libraries so I was hesitant to make any changes. I'll open a PR if I can find where in the source code to make the change and submit when the tests are passing locally.

rockchalkwushock:
DISCLAIMER: I am not strong in my regex-fu:

```js
path = path.replace(/^\s*~(?=$|\/|\\)/, userHome);
```

[...]

PR #3941, opened 2017-07-16
Fix: Don't assume all string configs are paths

BYK:
**Summary**

Refs #3787. When reading from the config, it was assumed that all string values are possible paths unless explicitly stated causing the `~` semver modifier to be expanded to user's home path if it was set as `save-prefix`. This patch makes `expand` default to false.

**Test plan**

Added a new test case for this specific scenario.


arcanis:
Ooooh I see - the regex doesn't take into account the case where the tilde is the only character in the string. I feel like this should be fixed into the regex rather than the code, wdyt? Only a matter of replacing the lookahead by a capture group.

BYK:
@arcanis  - There's no regex, this is done here: `https://github.com/yarnpkg/yarn/pull/3941/files#diff-9d9a6cd82f41984872a66a3ab0d440c4R197`.

If you're referring to the regex in `expandPath`, I disagree. `~` should be expanded. The problem is on the config reading part which assumes any string value is a path and I've fixed exactly that :)

arcanis:
> If you're referring to the regex in expandPath, I disagree. ~ should be expanded.

Why that? If someone needs to expand it, they can use `~/` to avoid the ambiguity. I'm just a bit wary of adding a per-option behavior when we could have a consistent behavior everywhere.

BYK:
>Why that? If someone needs to expand it, they can use "~/" instead to avoid the ambiguity.

This puts the onus on the users instead of us, the tool providers. Since it would create a surprising behavior, I don't think that's a good idea.

>I'm just a bit wary of adding a per-option behavior when we could have a consistent behavior everywhere - especially if we have to take care of passing the right arguments to getOption everytime.

We already had this and I made the lightest change I could do to not make things worse. That said if you want, you can remove the boolean argument and add the `expandPath` calls where they need to be used in a follow-up. Alternatively, we can add "option types" and separate "path" types from "string" types but that sounds a bit heavy-handed to me for now.

Issue #3897, opened 2017-07-11
Local packages with relative paths not correctly hoisted


aracarie:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Local packages with relative paths may not be hoisted correctly.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
mkdir a b
echo "{\"version\":\"1.0.0\",\"dependencies\":{\"b\":\"file:../b\"}}" > a/package.json
echo "{\"version\":\"1.0.0\"}" > b/package.json
echo "{\"dependencies\":{\"a\":\"file:a\",\"b\":\"file:./b\"}}" > package.json
yarn
```

Examine `yarn.lock` and execute `ls node_modules/a/node_modules` to see duplicated dependencies.

**What is the expected behavior?**
Common dependencies hoisted correctly.

I'll be able to provide a PR for this sometime this week.

PR #3945, opened 2017-07-18
Bugfix #3897: Fixing local package dependency resolving

aracarie:
**Summary**
File protocol dependencies will now be de-duped correctly. E.g. './a', 'a', and '/absolute/path/to/a' will be treated equivalently.

**Test plan**
Test added to cover the above cases.


src/lockfile/wrapper.js
```diff
@@ -184,6 +186,8 @@ export default class Lockfile {
 
       if (remote.resolved) {
         seen.set(remote.resolved, obj);
+      } else if (remote.type == 'copy' && remote.reference && remote.hash) {
```
BYK: May be you need a helper:
```js
const keyForRemote = remote => remote.reference && remote.hash ? `${remote.reference}#${remote.hash}` : null;
```

src/package-resolver.js
```diff
@@ -388,14 +388,18 @@ export default class PackageResolver {
       }
     }
 
+    if (getExoticResolver(version) && manifest) {
```
BYK: May be check `manifest` first to avoid the function call when it is missing?

src/package-resolver.js
```diff
@@ -388,14 +388,18 @@ export default class PackageResolver {
...
+      this.exoticRangeMatch(patterns.map(p => this.getStrictResolvedPattern(p)), manifest);
```
BYK: You can do `patterns.map(this.getStrictResolvedPattern.bind(this)), manifest);` instead.

src/package-resolver.js
```diff
@@ -410,7 +414,10 @@ export default class PackageResolver {
     });
 
     const maxValidRange = semver.maxSatisfying(versionNumbers, range);
-    if (!maxValidRange) {
+
+    if (!maxValidRange && getExoticResolver(range) && manifest) {
```
BYK: Same as above,  check `manifest` first?

src/package-resolver.js
```diff
@@ -410,7 +414,10 @@ export default class PackageResolver {
...
+      return this.exoticRangeMatch(resolvedPatterns, manifest);
+    } else if (!maxValidRange) {
```
BYK: Looks like this part needs some refactoring:
```js
if (!maxValidRange) {
    return manifest && getExoticResolver(range) ? this.exoticRangeMatch(resolvedPatterns, manifest) : null;
}
```

src/package-resolver.js
```diff
@@ -420,6 +427,26 @@ export default class PackageResolver {
     return maxValidRangeManifest;
   }
 
+  exoticRangeMatch(resolvedPkgs: Iterable<Manifest>, manifest: Manifest): ?Manifest {
+    if (!manifest._remote) {
+      return null;
+    } else if (manifest._remote.type != 'copy' || !manifest._remote.reference) {
```
BYK: You return `null` in both cases so why not:
```js
if (!(manifest._remote && manifest._remote.reference && manifest._remote.type === 'copy')) {
    return null;
}
```

This also makes your intent clearer.

src/package-resolver.js
```diff
@@ -420,6 +427,26 @@ export default class PackageResolver {
...
+      return null;
+    }
+
+    for (const pkg of resolvedPkgs) {
```
BYK:
```js
const matchedPkg = resolvedPkgs.find(pkg => {
    const remote = pkg._remote;
    return remote && remote.reference.manifest._remote.reference && remote.type === 'copy');
});

if (matchedPkg) { 
    manifest._remote = match._remote;
}

return matchedPkg;
```

__tests__/commands/install/integration.js
```diff
@@ -437,6 +437,24 @@ test.concurrent('install with file: protocol as default', (): Promise<void> => {
   });
 });
 
+test.concurrent('install file: dedupe dependencies', async (): Promise<void> => {
+  await runInstall({}, 'install-file-dedupe-dependencies', async (config, reporter) => {
+    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'a', 'node_modules'))).toEqual(false);
```
BYK: Add a comment above this stating what you are testing or be more explicit with your test, please. I'm guessing this is to make sure `b` is not added as a sub-dependency of `a`, right?

__tests__/commands/install/integration.js
```diff
@@ -437,6 +437,24 @@ test.concurrent('install with file: protocol as default', (): Promise<void> => {
...
+
+    // Add b as a dependency of c, as an absolute path
+    const pkgJsonContents = await fs.readFile(path.join(config.cwd, 'c/package.json'));
+    const pkgJson = JSON.parse(pkgJsonContents);
+    pkgJson.dependencies = {b: `file:${path.join(config.cwd, 'b')}`};
+    await fs.writeFile(path.join(config.cwd, 'c/package.json'), JSON.stringify(pkgJson));
```
BYK: Why are we doing this? Feels like this needs to be a separate test case or handled via a `yarn add b@file:...` command instead of directly manipulating the `package.json` file in the fixture.

__tests__/commands/install/integration.js
```diff
@@ -437,6 +437,24 @@ test.concurrent('install with file: protocol as default', (): Promise<void> => {
...
+
+    const lockfile = await Lockfile.fromDirectory(config.cwd);
+    const reInstall = new Install({}, config, reporter, lockfile);
+    await reInstall.init();
+
+    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'c', 'node_modules'))).toEqual(false);
```
BYK: Again, what are we testing? I'm guessing you are testing now `c` also doesn't have any sub-dependencies _even if_ we added `b` to it. I'd suggest creating a separate test case for this whole scenario. It would make your test cases more understandable and atomic so if something goes wrong in the future, we won't look at this test and try to decipher what exactly went wrong :)

aracarie:
Thanks for the review. I'll work on those changes.

aracarie:
Code refactored and test split into two separate cases with better comments.

src/lockfile/wrapper.js
```diff
@@ -156,7 +160,8 @@ export default class Lockfile {
       invariant(ref, 'Package is missing a reference');
       invariant(remote, 'Package is missing a remote');
 
-      const seenPattern = remote.resolved && seen.get(remote.resolved);
+      const altRemoteKey = keyForRemote(remote);
```
BYK: I think we should merge all this functionality into `keyForRemote` and just use it everywhere without the `remote.resolved` checks. What do you think?

src/package-resolver.js
```diff
@@ -421,6 +426,27 @@ export default class PackageResolver {
...
+   */
+
+  exoticRangeMatch(resolvedPkgs: Array<Manifest>, manifest: Manifest): ?Manifest {
+    const remote = manifest._remote;
+    if (!(remote && remote.reference && remote.type === 'copy')) {
+      return null;
+    }
+
+    const matchedPkg = resolvedPkgs.find(({_remote: pkgRemote}) => {
+      return pkgRemote && pkgRemote.reference === remote.reference && pkgRemote.type === 'copy';
```
BYK: No need for `return`:

```
resolvedPkgs.find(({_remote: pkgRemote}) => pkgRemote && pkgRemote.reference === remote.reference && pkgRemote.type === 'copy');
```

__tests__/commands/install/integration.js
```diff
@@ -437,6 +438,27 @@ test.concurrent('install with file: protocol as default', (): Promise<void> => {
...
+    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'a', 'node_modules'))).toEqual(false);
+  });
+});
+
+// When local packages are installed, dependencies with relative and absolute paths should be
+// deduped e.g. 'file:b' and 'file:/absolute/path/to/b'
+test.concurrent('install file: dedupe dependencies 2', (): Promise<void> => {
+  return runInstall({}, 'install-file-dedupe-dependencies-2', async (config, reporter) => {
+    // Add b as a dependency, using an absolute path
+    await add(config, reporter, {}, [`b@file:${path.join(config.cwd, 'b')}`]);
```
BYK: `fs.realpath` might be better than using `path.join` with `config.cwd` since it is guaranteed to be absolute where as `config.cwd` can still be a relative path. Unlikely, but just pointing it out.

src/lockfile/wrapper.js
aracarie: Yes this is better. It looks like there may be [one case](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L193-L197) where a remote may have resolved but a falsy hash, so I added a check for `remote.resolved` in `keyForRemote`.

src/package-resolver.js
aracarie: This one exceeded the max line length lint rule, so I left it as is for now.

__tests__/commands/install/integration.js
aracarie: Ended up using `path.resolve` for this since it allows multiple paths to be passed as arguments. Should give the same functionality as `fs.realpath`.

src/package-resolver.js
BYK: Prettier should fix it automatically. Or you can do it manually:
```
resolvedPkgs.find(
  ({_remote: pkgRemote}) =>
    pkgRemote && pkgRemote.reference === remote.reference && pkgRemote.type === 'copy'
);
```

__tests__/commands/install/integration.js
BYK: Nice!

src/package-resolver.js
```diff
@@ -388,14 +388,18 @@ export default class PackageResolver {
       }
     }
 
+    if (manifest && getExoticResolver(version)) {
+      this.exoticRangeMatch(patterns.map(this.getStrictResolvedPattern.bind(this)), manifest);
```
BYK: Did you mean to return this? Otherwise I'm a bit confused since we return `null` right after this.

aracarie: Yep, good catch. Updated.

BYK: Well, this means the tests don't cover this line :( we should add a test case in a follow-up.

BYK:
LGTM, will merge once Travis passes.

src/package-resolver.js
aracarie: Yes, I'll look into why the added tests don't seem to be covering this.

Issue #3276, opened 2017-04-28
Error: ENOENT: no such file or directory, symlink '' -> ...


slester:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**
When running `yarn`, I receive the following error: 
```Trace:
  Error: ENOENT: no such file or directory, symlink '' -> '/var/www/x/node_modules/browserify/test/symlink_dedupe/one/dir'
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Run `yarn` with browserify in the `package.json` (perhaps others)

**What is the expected behavior?**
For it not to try to create empty symlinks.

**Please mention your node.js, yarn and operating system version.**
node v7.9.0, yarn v0.23.2, `Linux x 4.10.10-1-ARCH #1 SMP PREEMPT Wed Apr 12 18:50:28 CEST 2017 x86_64 GNU/Linux`

bestander:
I'll need a package.json to verify if the fix works but I've put up a PR with a possible fix.

slester:
This issue still occurs with 0.24.5; node v7.10.0, same OS.

`Trace:
  Error: ENOENT: no such file or directory, symlink '' -> '/var/www/ba/node_modules/browserify/test/symlink_dedupe/one/dir'`

bestander:
Thanks, @slester.
Could you add steps to reproduce the issue please

slester:
I am running `yarn` on a `package.json` with a GitHub `username/repo` flag instead of a version number in `devDependencies`. What sort of other information would be helpful besides what I posted in the original issue?

bestander:
Unless there is an easy way to reproduce the issue it is unlikely that anyone will prioritize on fixing it over the other ones that have easy to repeat scenario.

The `ENOENT: no such file or directory, symlink` error looks very specific to the dependencies you have.
Thanks for reporting it, @slester, but your help to isolate the issue would be needed.
Ideally a shell script that would bootstrap a project with the error.

slester:
@bestander Here's a small script that will reproduce the issue. It seems to show up only if you're installing something from GitHub:

```mkdir yarn-ex && cd yarn-ex
echo '{"name":"yarn-ex", "dependencies": {"browserify": "substack/node-browserify"}, "version": "1.0.0"}' > package.json
yarn```

bestander:
Thanks, @slester, I can repro this.

PR #3946, opened 2017-07-18
Fix: Ensure valid target passed to fs.symlink (refs #3276)

aracarie:
**Summary**
[`path.relative`](https://nodejs.org/api/path.html#path_path_relative_from_to) returns an empty string for the current directory. This was being passed as the target to `fs.symlink`. Changed it so '.' is passed instead in these cases. Fixes issue #3276.

**Test plan**
The fix is very small and straightforward so haven't added a test. Can do if needed.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -725,7 +725,9 @@ export async function symlink(src: string, dest: string): Promise<void> {
       } else {
         relative = path.relative(path.dirname(dest), src);
       }
-      await fsSymlink(relative, dest);
+      // When path.relative returns an empty string for the current directory, we should instead use
+      // '.', which is a valid fs.symlink target.
+      await fsSymlink(relative || '.', dest);
     }
   } catch (err) {
     if (err.code === 'EEXIST') {


src/util/fs.js
```diff
@@ -721,9 +721,9 @@ export async function symlink(src: string, dest: string): Promise<void> {
       // use relative paths otherwise which will be retained if the directory is moved
       let relative;
       if (await exists(src)) {
-        relative = path.relative(fs.realpathSync(path.dirname(dest)), fs.realpathSync(src));
+        relative = path.relative(fs.realpathSync(path.dirname(dest)), fs.realpathSync(src)) || '.';
       } else {
-        relative = path.relative(path.dirname(dest), src);
+        relative = path.relative(path.dirname(dest), src) || '.';
       }
       await fsSymlink(relative, dest);
```
BYK: Do `await fsSymlink(relative || '.', dest);` instead?

Also please add a comment above why we default to `'.'` (basically, the summary of this PR as a comment :))

aracarie:
Changes made :+1:

Issue #3524, opened 2017-05-29
Child process is exited prematurely when running a info command


ncuillery:
When running the info command in the child process, the child process exits right after the reception of the first chunk of data:

```javascript
const exec = require('child_process').exec;

const child = exec('yarn info react-native@latest');

child.stdout.on('data', data => {
  console.log('DATA', data.length);
});

child.stderr.on('data', data => {
  console.log('ERROR', data);
});

child.on('exit', code => {
  console.log('EXIT', code);
});
```

Output: 
```
$ node debug-exec.js 
DATA 18
DATA 8192
EXIT 0
```

The first chunk of data contains `yarn info v0.24.5` (first line of the output). Then, the second chunk contains the first 8192b of the info command output. Then the process is closed properly (error code 0).

This bug has been discovered by using NPM 5.0.0. I investigated and I reproduced the same bug with yarn, so I report the issue here too.

- Original issue on React Native: https://github.com/facebook/react-native/issues/14239
- NPM issue I created after my investigation: https://github.com/npm/npm/issues/16780

bestander:
Thanks for reporting.
According to npm fix process.stdout.write need to have a callback, should Yarn do the same?
Help welcome

bestander:
I'll raise the priority to get keep it on radar.

BYK:
I'll look into console reporter to see how hard this is to fix. Aiming for the 1.0 release for now.

PR #3955, opened 2017-07-18
Fix: avoid using `process.exit()` when possible

BYK:
**Summary**

Refs #3524. We are using `process.exit()` pretty liberally although it is [strongly advised not to use it](https://nodejs.org/api/process.html#process_process_exit_code) since it may cause the `stdout` to get corrupted/terminated before a full flush. This patch changes all possible `process.exit(code)` calls with `process.exitCode = code` statements.

We'd also ideally enable [no-process-exit](http://eslint.org/docs/rules/no-process-exit) rule in ESLint but it requires an upgrade to ESLint v4 which should be handled separately.

**Test plan**

Expect tests to pass and actually finish (not run indefinitely due to `yarn` not exiting at all). Also, the script referenced in #3524 to output something like the following:
```
DATA 18
DATA 8192
DATA 8192
DATA 5392
DATA 15
EXIT 0
```

### DIFF
--- a/scripts/build-webpack.js
+++ b/scripts/build-webpack.js
@@ -35,6 +35,8 @@ const compiler = webpack({
   output: {
     filename: `yarn-${version}.js`,
     path: path.join(basedir, 'artifacts'),
+    library: 'yarn-cli',
+    libraryTarget: 'commonjs2',
   },
   target: 'node',
 });
@@ -71,6 +73,8 @@ const compilerLegacy = webpack({
   output: {
     filename: `yarn-legacy-${version}.js`,
     path: path.join(basedir, 'artifacts'),
+    library: 'yarn-cli',
+    libraryTarget: 'commonjs2',
   },
   target: 'node',
 });
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -20,347 +20,353 @@ const net = require('net');
 const onDeath = require('death');
 const path = require('path');
 
-loudRejection();
-handleSignals();
+export default function main({
+  startArgs,
+  args,
+  endArgs,
+}: {
+  startArgs: Array<string>,
+  args: Array<string>,
+  endArgs: Array<string>,
+}) {
+  loudRejection();
+  handleSignals();
 
-const startArgs = process.argv.slice(0, 2);
+  // set global options
+  commander.version(version, '--version');
+  commander.usage('[command] [flags]');
+  commander.option('-v, --verbose', 'output verbose messages on internal operations');
+  commander.option('--offline', 'trigger an error if any required dependencies are not available in local cache');
+  commander.option('--prefer-offline', 'use network only if dependencies are not available in local cache');
+  commander.option('--strict-semver');
+  commander.option('--json', '');
+  commander.option('--ignore-scripts', "don't run lifecycle scripts");
+  commander.option('--har', 'save HAR output of network traffic');
+  commander.option('--ignore-platform', 'ignore platform checks');
+  commander.option('--ignore-engines', 'ignore engines check');
+  commander.option('--ignore-optional', 'ignore optional dependencies');
+  commander.option('--force', 'install and build packages even if they were built before, overwrite lockfile');
+  commander.option('--skip-integrity-check', 'run install without checking if node_modules is installed');
+  commander.option('--check-files', 'install will verify file tree of packages for consistency');
+  commander.option('--no-bin-links', "don't generate bin links when setting up packages");
+  commander.option('--flat', 'only allow one version of a package');
+  commander.option('--prod, --production [prod]', '');
+  commander.option('--no-lockfile', "don't read or generate a lockfile");
+  commander.option('--pure-lockfile', "don't generate a lockfile");
+  commander.option('--frozen-lockfile', "don't generate a lockfile and fail if an update is needed");
+  commander.option('--link-duplicates', 'create hardlinks to the repeated modules in node_modules');
+  commander.option('--global-folder <path>', 'specify a custom folder to store global packages');
+  commander.option(
+    '--modules-folder <path>',
+    'rather than installing modules into the node_modules folder relative to the cwd, output them here',
+  );
+  commander.option('--cache-folder <path>', 'specify a custom folder to store the yarn cache');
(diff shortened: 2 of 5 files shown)


src/cli/index.js
```diff
@@ -142,7 +142,7 @@ if (outputWrapper) {
 if (command.noArguments && commander.args.length) {
   reporter.error(reporter.lang('noArguments'));
   reporter.info(command.getDocsInfo);
-  process.exit(1);
+  process.exitCode = 1;
```
arcanis: You need to shortcut the rest of the code to have the same behavior - `process.exit` doesn't return, but setting `process.exitCode` will continue executing the rest of the code.

src/util/signal-handler.js
```diff
@@ -3,7 +3,7 @@ import {forwardSignalToSpawnedProcesses} from './child.js';
 
 function forwardSignalAndExit(signal: string) {
   forwardSignalToSpawnedProcesses(signal);
-  process.exit(1);
+  throw new Error('Received SIGTERM, terminating.');
```
arcanis: Will that print a stacktrace on the console? That won't be helpful for the user.

BYK: I thought it would be. Why do you think it wouldn't?

src/cli/index.js
BYK: Not shortcutting the whole point with this change. We should not need the shortcut behavior and turns out we really don't since all tests are passing. It was only necessary at one point so I left it that way.

src/util/signal-handler.js
arcanis: `SIGTERM` is the default signal sent by `kill`. It doesn't make much sense to print a stacktrace (which heavily imply that something went wrong in the program implementation) if the termination came from the outside, don't you think?

BYK: Alright, makes sense. I'll find an alternative way for this one.

src/cli/index.js
arcanis: I don't think that's a good idea.

- it kinda violates the principle of the least surprise, since we would no longer have the guarantee that any part of this file would only be executed if no other command matched. It's only a matter of time before we make some mistake that will execute two commands in some edge case.

- Case in point, your current code has a bug that will make `yarn -v` print the version number before running `yarn install -v` (and `yarn -v -h` will print the version number then call `yarn help -v`).

- Even without that, not shortcutting also means that we will be running useless tasks (such as instanciating a reporter etc) even when not needed. That's an issue because those commands must be fast (some tools such as Nuclide might need to call them often, and if their results are late it will slow down the all process).

I'm not against using `exitCode`, you have a point that shutting down the program before the I/O has finished is bad, but I strongly think we then need to refactor the CLI code to make sure we never have to check that our code won't get called twice - might only be a matter of adding `else` clauses 😃

src/cli/index.js
```diff
@@ -20,347 +20,352 @@ const net = require('net');
...
-// if -v is the first command, then always exit after returning the version
-if (commandName === '-v') {
-  console.log(version.trim());
-  process.exitCode = 0;
-}
+export default function main() {
+  loudRejection();
+  handleSignals();
+
+  const startArgs = process.argv.slice(0, 2);
```
arcanis: What do you think about modularizing and putting this in the `main` parameters? I can see this function as an interesting entry point to run commands in our testcases - we currently need to instanciate each class and pass them the right properties, which might be tedious and error-prone, so it could be nice to just be able to use `run([ 'install', '--check-files' ])` instead and use the actual CLI pipeline.

arcanis: Actually my comment was more about [Line 31](https://github.com/yarnpkg/yarn/pull/3955/files#diff-867becf4a9c2c6c6d4e7c1278750724eR31), but it's also applicable here.

src/reporters/console/console-reporter.js
```diff
@@ -196,7 +196,7 @@ export default class ConsoleReporter extends BaseReporter {
         (err, answer) => {
           if (err) {
             if (err.message === 'canceled') {
-              process.exit(1);
```
arcanis: Shouldn't we exit immediately here? Or reject?

scripts/build-webpack.js
```diff
@@ -35,6 +35,8 @@ const compiler = webpack({
   output: {
     filename: `yarn-${version}.js`,
     path: path.join(basedir, 'artifacts'),
+    library: 'yarn-cli',
+    libraryTarget: 'commonjs2',
```
arcanis: Is this change required? What breaks without it?

BYK: Yes, this is the solution that fixes the single-file bundles. Without this, Webpack exports an empty object, causing `yarn-bundled-entry-point.js` to fail since `main` function can't be imported.

src/reporters/console/console-reporter.js
BYK: Yeah I think you're right. I thought about the same thing and don't know why I didn't merge the `reject` part. Update coming.

skevy:
I noticed when pulling down master and running `build-dist` that the webpack builds were no longer working.

I traced it back to this PR. You can see in CircleCI the output of the last working build here:

https://circleci.com/gh/yarnpkg/yarn/4533

Notice how `$ ./artifacts/yarn-`./dist/bin/yarn --version`.js --version` outputs a version number.

When this commit runs in CI, however (`https://circleci.com/gh/yarnpkg/yarn/4536`), there is no version number output, and I confirm that the webpack build does not work at all.

Daniel15:
Hmm, interesting, I wonder why the nightly end-to-end tests didn't fail. They use the nightly Debian packages, which contain the Webpack build inside them. https://build.dan.cx/job/yarn-e2e/label=docker,os=ubuntu-16.04/269/console

skevy:
I'm not sure why those don't fail (without looking into it more) -- but the reason CircleCI doesn't fail is because `yarn --version` still exists with code 0, even though it doesn't do anything.

Daniel15:
Right. Maybe the build should assert that it actually returns the correct version number.

skevy:
👍

BYK:
I'll work on this soon.

BYK:
Filed #4057, actively working on it.

PR #3958, opened 2017-07-19
Global Config at ${PREFIX}/etc/{nmp,yarn}rc

### LINKED ISSUE #1931: Yarn not reading /etc/npmrc while npm does
I'm using yarn 0.17.3, this AFAIK is a bug.

In CI we have a file /etc/npmrc to point to the proxy we use for internet access:

```
ci$ cat /etc/npmrc
proxy = http://npmcache:3128
https-proxy = http://npmcache:3128
```

npm is picking up this configuration:

```
ci$ npm config list
; cli configs
user-agent = "npm/3.10.3 node/v6.7.0 linux x64"

; globalconfig /usr/etc/npmrc
https-proxy = "http://npmcache:3128/"
proxy = "http://npmcache:3128/"

; node bin location = /usr/bin/nodejs
; cwd = /home/ci
; HOME = /home/ci
; "npm config ls -l" to show all defaults.
```

But yarn does not:

```
ci$ yarn config list
yarn config v0.17.3
info yarn config
{ 'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'strict-ssl': true,
  'user-agent': 'yarn/0.17.3 npm/? node/v6.7.0 linux x64' }
info npm config
{}
Done in 0.07s.
```

Parakleta:
This searches for the global RC files at `${PREFIX}/etc/{npm,yarn}rc`, whereas previously it looked in `${PREFIX}/.{npm,yarn}rc`.  By code inspection it looks like the NPM code does the same thing on both windows and posix, but the documentation for windows suggests it might not use the `/etc` path component.  I haven't verified behaviour on windows.

Resolves #1931.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -127,15 +127,18 @@ export default class NpmRegistry extends Registry {
   }
 
   async getPossibleConfigLocations(filename: string, reporter: Reporter): Promise<Array<[boolean, string, string]>> {
+    // npmrc --> ./.npmrc, ~/.npmrc, ${prefix}/etc/npmrc
+    const localfile = '.' + filename;
     const possibles = [
-      [false, path.join(this.cwd, filename)],
-      [true, this.config.userconfig || path.join(userHome, filename)],
-      [false, path.join(getGlobalPrefix(), 'etc', filename.slice(1))],
+      [false, path.join(this.cwd, localfile)],
+      [true, this.config.userconfig || path.join(userHome, localfile)],
+      [false, path.join(getGlobalPrefix(), 'etc', filename)],
     ];
 
+    // npmrc --> ../.npmrc, ../../.npmrc, etc.
     const foldersFromRootToCwd = getPosixPath(this.cwd).split('/');
     while (foldersFromRootToCwd.length > 1) {
-      possibles.push([false, path.join(foldersFromRootToCwd.join(path.sep), filename)]);
+      possibles.push([false, path.join(foldersFromRootToCwd.join(path.sep), localfile)]);
       foldersFromRootToCwd.pop();
     }
 
@@ -167,7 +170,7 @@ export default class NpmRegistry extends Registry {
     // docs: https://docs.npmjs.com/misc/config
     this.mergeEnv('npm_config_');
 
-    for (const [, loc, file] of await this.getPossibleConfigLocations('.npmrc', this.reporter)) {
+    for (const [, loc, file] of await this.getPossibleConfigLocations('npmrc', this.reporter)) {
       const config = NpmRegistry.normalizeConfig(ini.parse(file));
 
       // normalize offline mirror path relative to the current npmrc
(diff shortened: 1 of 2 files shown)


src/registries/npm-registry.js
```diff
@@ -130,7 +130,7 @@ export default class NpmRegistry extends Registry {
     const possibles = [
       [false, path.join(this.cwd, filename)],
       [true, this.config.userconfig || path.join(userHome, filename)],
-      [false, path.join(getGlobalPrefix(), filename)],
+      [false, path.join(getGlobalPrefix(), 'etc', filename.slice(1))],
```
BYK: Please add a comment about the `.slice(1)` part above this line. Or even better, add example paths as comments:
```
./.npmrc
~/.npmrc
$PREFIX/etc/npmrc
```

Parakleta: Actually to be properly correct it should probably take the RC name without the leading `.` and prefix it, rather than slicing it off.  If we're going to do this properly I can try and take a look tomorrow at filling it out a bit.

I don't know anything about the Windows install of this or NPM though, so someone else will need to check that aspect of it.

BYK: That sound great :) Your change seems okay on Windows too since it seems in-line with what NPM documents and it doesn't break AppVeyor builds.

Parakleta:
I think this structure is cleaner, and I added some comments.

Parakleta:
I just noticed similar path searching behaviour in https://github.com/yarnpkg/yarn/blob/master/src/util/rc.js but I'm not sure where this code fits in.  It does miss the `${prefix}/etc` version and looks in `/etc` instead which is probably not correct.  Not sure why there are multiple RC search algorithms.

BYK:
>I just noticed similar path searching behaviour in https://github.com/yarnpkg/yarn/blob/master/src/util/rc.js but I'm not sure where this code fits in. It does miss the ${prefix}/etc version and looks in /etc instead which is probably not correct. Not sure why there are multiple RC search algorithms.

Oh, that's quite concerning actually. Let me check that.

BYK:
@Parakleta I'll go ahead and merge your fix since it fixes an immediate problem. That said [`findRc` method is only used in one place](https://github.com/yarnpkg/yarn/search?utf8=%E2%9C%93&q=findRc&type=) and its logic should be fixed too. I'd argue we should use it here too if possible.

Would you be interested in taking a stab at it as a follow-up?

PS: 
```js
const home = isWin ? process.env.USERPROFILE : process.env.HOME;
```

line should actually be
```js
const home = process.env.HOME || process.env.USERPROFILE;
```

Sicne you can have a `HOME` environment variable on Windows too :)

Parakleta:
I think consolidating `findRc` is going to require refactoring and that's a bit much for me, I'll leave that for someone who understands the structure and direction of the project.

BYK:
@Parakleta It's actually only used in two places so it shouldn't be too complicated. If you don't have time, that's okay but if you don't wanna try just because you'd get drowned we're here to help ;)

PR #3960, opened 2017-07-19
Fix: pruneOfflineMirror handle undefined resolved

tmchng:
**Summary**

Bug: pruneOfflineMirror does not guard against `lockfile[dependency].resolved == undefined` (eg. the dependency is a local file)
This results in the following error when running `yarn install`:
`TypeError: Cannot read property 'split' of undefined`

To reproduce:

- set a `yarn-offline-mirror` path and set `yarn-offline-mirror-pruning` to `true`
- have a `file:` local dependency in `package.json`
- run `yarn install`

Simple fix - set basename only if resolved is valid.

**Test plan**

The fix is obvious - call `resolved.split()` only after making sure `resolved` is not undefined. Nothing changes at the functional level.

Please let me know if I should still create a test case for this scenario.

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -634,8 +634,8 @@ export class Install {
     const requiredTarballs = new Set();
     for (const dependency in lockfile) {
       const resolved = lockfile[dependency].resolved;
-      const basename = path.basename(resolved.split('#')[0]);
       if (resolved) {
+        const basename = path.basename(resolved.split('#')[0]);
         if (dependency[0] === '@' && basename[0] !== '@') {
           requiredTarballs.add(`${dependency.split('/')[0]}-${basename}`);
         }
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -6,6 +6,7 @@ import type {ReporterSelectOption} from '../../reporters/types.js';
 import type {Manifest, DependencyRequestPatterns} from '../../types.js';
 import type Config from '../../config.js';
 import type {RegistryNames} from '../../registries/index.js';
+import type {LockfileObject} from '../../lockfile/wrapper.js';
 import normalizeManifest from '../../util/normalize-manifest/index.js';
 import {MessageError} from '../../errors.js';
 import InstallationIntegrityChecker from '../../integrity-checker.js';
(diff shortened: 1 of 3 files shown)


arcanis:
Hey @tmchng, thanks for your PR! A testcase would indeed be great 😃 

To write one, you can add a new folder in the [fixture](https://github.com/yarnpkg/yarn/tree/master/__tests__/fixtures/install) directory, then add an entry in the [integration.js file](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L51-L59). I think you can even omit the `runInstall` callback, since in this case you only want to check that the install succeeds :)

BYK:
I think we should change `lockfile: Object` definition in the function to something that resembles the actual lockfile structure and mark `resolved` as `resolved: ?string`. This way `flow` would take care of the type safety and we won't need an extra test.

arcanis:
Yup, that would be a good alternative

tmchng:
@BYK Making sure that `resolved` is `string` type makes sense. What do you mean by "changing `lockfile: Object` definition to something that resembles the actual lock file structure?" I'll have to look more into how flow works.

The test case I thought might be missing is making sure that "yarn install succeeds when there's a `file:` dependency", and in our case with `yarn-offline-mirror-pruning` turned on. [This test case](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L394) is similar to what I have in mind, so we just need one with the offline mirror + pruning set.

arcanis:
@tmchng We use Flow to have static types in our code. It usually means that we can't have the kind of error you've reported because Flow will just abort the compilation (so we don't need tests - the Flow check *is* the test). Unfortunately, the type of the `lockfile` object is defined as `Object`, which basically disable any check on any variable extracted from it. A better type would be something along the line of `{ [string]: { resolved: ?string } }` (plus any other properties), which would allow Flow to know that `resolved` might be falsy, and would prevent us from using it without guard.

tmchng:
@arcanis Added Flow check on the lockfile object, and it caught the bug I fixed as expected. I'm not going to add a integration test then since it seems like you and @BYK don't think it's necessary.

PR #3976, opened 2017-07-20
Prevent optional subdependencies from being installed with --ignore-optional

### LINKED ISSUE #2666: ignore-optional flag works not like expected
Bug / feature ? 

`yarn install --ignore-optional` tries to install also optional dependencies of sub modules. The while install process fails if one of the optional module installation fails.

To skip the optional dependencies of sub modules or to continue on errors.

Node 7.5.0, Yarn v0.19.1, Ubuntu 14.04

arcanis:
Fix #2666

@blexrob Would you mind reviewing this PR? You worked a bit on the hoister in #3465, so you might have some additional insight :)


src/cli/commands/install.js
```diff
@@ -277,12 +277,7 @@ export class Install {
 
       pushDeps('dependencies', projectManifestJson, {hint: null, optional: false}, true);
       pushDeps('devDependencies', projectManifestJson, {hint: 'dev', optional: false}, !this.config.production);
-      pushDeps(
-        'optionalDependencies',
-        projectManifestJson,
-        {hint: 'optional', optional: true},
-        !this.flags.ignoreOptional,
-      );
+      pushDeps('optionalDependencies', projectManifestJson, {hint: 'optional', optional: true}, true);
```
BYK: Why always set `isUsed` to true?

src/package-hoister.js
```diff
@@ -142,15 +146,15 @@ export default class PackageHoister {
     //
     let parentParts: Parts = [];
     const isIncompatible = ref.incompatible;
-    let isRequired = isDirectRequire && !ref.ignore && !isIncompatible;
+    let isRequired = isDirectRequire && !ref.ignore && !isIncompatible && (!ref.optional || !this.ignoreOptional);
```
BYK: Better to write as `!(ref.optional && this.ignoreOptional)` since makes the intent clearer. Also possibly fewer instructions :P

src/package-hoister.js
```diff
@@ -142,15 +146,15 @@ export default class PackageHoister {
...
 
     if (parent) {
       if (!this.tree.get(parent.key)) {
         return null;
       }
       // non ignored dependencies inherit parent's ignored status
       // parent may transition from ignored to non ignored when hoisted if it is used in another non ignored branch
-      if (!isDirectRequire && !isIncompatible && parent.isRequired) {
+      if (!isDirectRequire && !isIncompatible && parent.isRequired && (!ref.optional || !this.ignoreOptional)) {
```
BYK: Same as above. Also, may convert this to a function instead of repeating the same logic in multiple places?

src/package-hoister.js
```diff
@@ -196,7 +200,12 @@ export default class PackageHoister {
 
       for (const depPattern of ref.dependencies) {
         const depinfo = this._lookupDependency(info, depPattern);
-        if (depinfo && !depinfo.isRequired && !depinfo.isIncompatible) {
+        if (
+          depinfo &&
+          !depinfo.isRequired &&
+          !depinfo.isIncompatible &&
+          (!depinfo.pkg._reference || !depinfo.pkg._reference.optional || !this.ignoreOptional)
```
BYK: Yeah, we definitely need to extract this logic out.

src/package-hoister.js
arcanis: That's purely stylistic ;)

src/package-hoister.js
arcanis: Yeah I agree, but there was some subtle changes in the conditions so I aimed to make the minimal change required and revisit if needed. I'll give it a second look.

src/cli/commands/install.js
arcanis: Because the ignore flag isn't managed anymore by this portion of the code - we do it directly in the hoister instead, so that we can operate recursively.

BYK: Gotcha. May be we should remove that flag in a follow-up since it is only used by `devDependencies` and is confusing a bit.

src/package-hoister.js
BYK: Not really stylistic. One reads "if `ref` is not optional or we should not ignore optionals" and the other one reads "if NOT (ref is optional and we should ignore optionals)". Human brains are bad at double negations ;)

src/package-hoister.js
BYK: We can defer this to a follow-up since it's not the main goal of the PR anyways.

src/cli/commands/install.js
blexrob: isUsed is also used to remove the package from consideration in flat mode, so I wouldn't remove it (if another package requires an incompatible version, this would create a conflict).

PR #3985, opened 2017-07-21
Fix: handle potential race conditions when using cache

BYK:
**Summary**

There's a delay between when we make sure a file in cache exists and when we actually start reading it. Our code wasn't doing any error handling when trying to open the file so if another process removes the file we assume existing we were throwing an unexpected `ENOENT ... open` error.

This patch removes all delayed existence checks and catches the error directly on the `ReadStream` instance directly. Moreover, it automatically falls back for local caches: direct, cache, offline mirror.

This revealed a bug in tarball path resolution which I fixed.

**Test plan**

Existing tests.

### DIFF
--- a/src/fetchers/git-fetcher.js
+++ b/src/fetchers/git-fetcher.js
@@ -37,29 +37,6 @@ export default class GitFetcher extends BaseFetcher {
     }
   }
 
-  async getLocalAvailabilityStatus(): Promise<boolean> {
-    // Some mirrors might still have files named "./reponame" instead of "./reponame-commit"
-    const tarballLegacyMirrorPath = this.getTarballMirrorPath({
-      withCommit: false,
-    });
-    const tarballModernMirrorPath = this.getTarballMirrorPath();
-    const tarballCachePath = this.getTarballCachePath();
-
-    if (tarballLegacyMirrorPath != null && (await fsUtil.exists(tarballLegacyMirrorPath))) {
-      return true;
-    }
-
-    if (tarballModernMirrorPath != null && (await fsUtil.exists(tarballModernMirrorPath))) {
-      return true;
-    }
-
-    if (await fsUtil.exists(tarballCachePath)) {
-      return true;
-    }
-
-    return false;
-  }
-
   getTarballMirrorPath({withCommit = true}: {withCommit: boolean} = {}): ?string {
     const {pathname} = url.parse(this.reference);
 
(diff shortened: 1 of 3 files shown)


BYK:
@cpojer may be I can simulate the race condition in tests with mocking. What do you think?

cpojer:
Possibly, if you mock the fs module in a test.

BYK:
Actually, hold it: I'm gonna try rewriting this part to have proper fallbacks instead of just crashing and burning in flames.

BYK:
@arcanis @cpojer this turned out to be a very interesting patch and larger than I expected. That said it revealed a bug which I fixed. Swift review appreciated! :)

BYK:
//cc @bestander

src/fetchers/git-fetcher.js
```diff
@@ -78,30 +55,47 @@ export default class GitFetcher extends BaseFetcher {
...
 
-    if (!tarballPath || !await fsUtil.exists(tarballPath)) {
-      throw new MessageError(this.reporter.lang('tarballNotInNetworkOrCache', this.reference, tarballPath));
+  async fetchFromLocal(override: ?string): Promise<FetchedOverride> {
+    let cachedStream;
+    const triedPaths = [];
+    for (const tarballPath of this.getLocalPaths(override)) {
+      if (tarballPath) {
+        try {
+          cachedStream = await new Promise((resolve, reject) => {
```
BYK: We may wanna add a comment here explaining the weird `await Promise` thing: `createReadStream` is synchronous but it immediately triggers an `error` event if the file is not found. If nothing handles this event, it crashes node. The problem is we want to try an array of files in order so we need this pattern.

arcanis: Looks reasonnable :+1:

src/fetchers/tarball-fetcher.js
```diff
@@ -208,11 +216,7 @@ export default class TarballFetcher extends BaseFetcher {
       return this.fetchFromLocal(this.reference);
     }
 
-    if (await this.getLocalAvailabilityStatus()) {
-      return this.fetchFromLocal();
-    } else {
-      return this.fetchFromExternal();
-    }
+    return this.fetchFromLocal().catch(err => this.fetchFromExternal());
```
arcanis: This will apply the fallback for any error - we probably only want to do this for a small subset of errors (basically ENOENT)

arcanis:
I don't see any problem in this PR, however I'm a bit concerned it will only hide the issue instead of fixing it - unless I'm missed something it doesn't make running two Yarn instances at the same time any safer, since they will still end up with a potentially corrupted cache that might randomly break. It will only prevent some early breaks, which would arguably be better than a corrupted installation 😕

arcanis:
What do you think about adding something like this?

```js
let lock;

try {
    lock = await fs.open(`${cachePath}/.yarn-lock`, `wx`)
} catch (error) {
    if (error.code === `EEXIST`) {
        throw new Error(`Two instances of Yarn seem to be running concurrently - if you're sure this is not the case, use "yarn unlock" to remove the locks`);
    } else {
        throw error;
    }
}

try {
    // ...
} finally {
    await fs.unlink();
}
```

The main problem is that `fetchFromLocal` exits before we've ended reading the cache, so we would need to have some way to persist the lock until it gets useless.

BYK:
>I don't see any problem in this PR, however I'm a bit concerned it will only hide the issue instead of fixing it - unless I'm missed something it doesn't make running two Yarn instances at the same time any safer, since they will still end up with a potentially corrupted cache that might randomly break. It will only prevent some early breaks, which would arguably be better than a corrupted installation 😕

The aim for this PR was not to solve the concurrent Yarn issues. It was to make Yarn more resilient to them and provide better error messages instead of cryptic error messages without a stacktrace like `ENOENT: cannot find file xxxx` which is what happens if you don't handle the error from `createReadStream`. Along the way it also removes many unnecessary file system I/O that checks for file/directory existence so I'm hoping it will have a positive impact on the perf.

I'll follow up with patches to address the concurrent Yarn instances problem without using a lock.

>What do you think about adding something like this (warning: potato code)?

As mentioned above, I'll try to avoid using locks for now and this patch is not about solving the concurrency problem completely.

src/fetchers/tarball-fetcher.js
BYK: Don't agree with this part is what the old code did mostly and `fetchFromLocal` doesn't actually throw `ENOENT` errors. It throws handmade "can't find file in cache" errors. That said we can check specifically for `ENOENT` errors inside `fetchFromLocal` trials but I'm not really sure if it matters since if it is something else, we still ended up not being able to read the file and we want to try the next file name.

BYK: My only concern around this now is with offline mode, I'll double check that this change doesn't affect that.

src/fetchers/git-fetcher.js
BYK: Will update the patch with a comment then :)

PR #3987, opened 2017-07-21
Fix: Only add auth token when registry url matches pathname url (fixes #3907)

### LINKED ISSUE #3907: Can't install from GitHub when always-auth is set to true

Bug report.

```bash
$ yarn add --dev npm-publish-git-tag
yarn add v0.28.0-20170710.0901
[2/5] Resolving packages...
[3/5] Fetching packages...
error An unexpected error occurred: "https://codeload.github.com/bahmutov/local-or-home-npmrc/tar.gz/7b9c4e239457b0810a98b84707730618e75c9ee2: Request failed \"404 Not Found\"".
```

Simply run `yarn add --dev npm-publish-git-tag` within any directory.

You may also run `yarn add --dev npm-utils`, as `npm-utils` is [the package that has the GitHub reference](https://github.com/bahmutov/npm-utils/blob/master/package.json#L40).

For `yarn` to successfully download, and install, the tarball provided by the URL above.

The following simple request works:

```bash
curl https://codeload.github.com/bahmutov/local-or-home-npmrc/tar.gz/7b9c4e239457b0810a98b84707730618e75c9ee2
```

`npm install --save-dev npm-utils` also works as desired.

```bash
$ yarn --version
0.28.0-20170710.0901

$ node --version
v6.11.0

$ uname -a
MINGW64_NT-10.0 HBETTSPC2 2.7.0(0.307/5/3) 2017-02-17 14:20 x86_64 Msys
```

Windows 10

soda0289:
**Summary**
Fixes issue: https://github.com/yarnpkg/yarn/issues/3907

In yarn version 0.27.4 a patch (https://github.com/yarnpkg/yarn/commit/5ff69228f6cc9fa27bbaeb67412f7f66f218b9da) introduced a change that caused an authorization header to be added to requests that were not being sent to the registry.

This commit brings back some older logic to ensure we only add the authorization header when we are sending requests to the registry.

**Test plan**

To test this change you need a private repository using a repository server such, as Verdaccio configured, to only allow authenticated calls to be made for both download and publish commands. This configuration requires that yarn has the `always-auth` option enabled.

If we install additional packages from github and save them to the `yarn.lock` file it should not send the authorization header for these requests.

I started to add some unit tests. This is my first time writing mocks with Jest and any suggestions are welcome. I can add more tests for additional cases as well.


__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+      mockReporter,
+    };
+  }
+
+  test('should call requestManager.request with pathname url', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const pathname = 'http://github.com/yarnpkg/yarn.tgz';
```
BYK: This is a URL not a path name. Variable names are important ;)

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+
+    npmRegistry.request(pathname);
+
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
+
+    expect(requestParams.url).toBe(pathname);
```
BYK: You should probably use [toHaveBeenLastCalledWith](https://facebook.github.io/jest/docs/expect.html#tohavebeenlastcalledwitharg1-arg2-) instead.

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const pathname = 'http://github.com/yarnpkg/yarn.tgz';
+
+    npmRegistry.request(pathname);
+
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
+
+    expect(requestParams.headers.authorization).toBe(undefined);
```
BYK: Same as above: use `toHaveBeenCalledWith` or `toHaveBeenLastCalledWith`

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+    const mockReporter = jest.fn();
+
+    return {
+      mockRequestManager,
+      mockRegistries,
+      mockReporter,
+    };
+  }
+
+  test('should call requestManager.request with pathname url', () => {
```
BYK: Just `url` nor `pathname url`

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+
+    expect(requestParams.url).toBe(pathname);
+  });
+
+  test('should not add authorization header if pathname not to registry', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const pathname = 'http://github.com/yarnpkg/yarn.tgz';
```
BYK: Use `HTTPS` URLs, please. Even for tests ;)

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+
+    npmRegistry.config = {
+      'always-auth': false,
+      _authToken: 'testAuthToken',
+    };
+    npmRegistry.request(pathname);
+
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
+
+    expect(requestParams.headers.authorization).toBe(undefined);
```
BYK: Use [toBeUndefined](https://facebook.github.io/jest/docs/expect.html#tobeundefined) instead.

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,122 @@ describe('normalizeConfig', () => {
...
+
+  test('should not add authorization header if pathname is to registry and always-auth is false', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const pathname = 'https://registry.npmjs.org/yarnpkg/yarn.tgz';
+
+    npmRegistry.config = {
+      'always-auth': false,
```
BYK: I think this is the default value anyways, right?

src/registries/npm-registry.js
```diff
@@ -72,10 +73,22 @@ export default class NpmRegistry extends Registry {
     return name.replace('/', '%2f');
   }
 
+  getRequestUrl(registry: string, pathname: string): string {
+    const isPathnameUrl = pathname.match(/^https?:/);
```
soda0289: The `.match` string method is used throughtout this class and I'm guessing those should be repalced as well. I can fix this in another PR.

I will rename the variable to `isUrl`. I think `pathname` can also be a path, `/foo/bar/`, as well as a URL.

src/registries/npm-registry.js
```diff
@@ -72,10 +73,22 @@ export default class NpmRegistry extends Registry {
...
+    }
+  }
+
   request(pathname: string, opts?: RegistryRequestOptions = {}, packageName: ?string): Promise<*> {
     const registry = this.getRegistry(packageName || pathname);
-    const requestUrl = url.resolve(registry, pathname);
+    const requestUrl = this.getRequestUrl(registry, pathname);
+
     const alwaysAuth = this.getRegistryOrGlobalOption(registry, 'always-auth');
+    const customHostSuffix = this.getRegistryOrGlobalOption(registry, 'custom-host-suffix');
```
soda0289: I will have to move the `isRequestToRegistry` inside the `NPMRegistry` class since it will needs access to the config to read the 'custom-host-suffix`.

src/registries/npm-registry.js
soda0289: `.test` is not part of the string prototype only the Regex prototype.

__tests__/registries/npm-registry.js
soda0289: Is there a way to use `.toHaveBeenLastCalledWith` and only match one property of a JSON object? Sinon has matchers that can be used but I don't see anything similar in the documentation.

__tests__/registries/npm-registry.js
soda0289: I think the default might be undefined do you think I should have a test for that as well.

soda0289:
@BYK Thanks for the review!

I have made the following changes:
     - Add test for non-registry and always-auth true
     - Rename variables
     - Use Regex.text instead of String.match
     - Move isRequesteRegistry into NPMRegistry class
     - Move customHostSuffix code inside isRequestToRegistry
     - Use .toBeUndefined

I did not use `.toHaveBeenLastCalledWith()` since I was unsure how to only check one JSON property in the argument. I made all of the other changes.

src/registries/npm-registry.js
```diff
@@ -74,21 +76,50 @@ export default class NpmRegistry extends Registry {
...
 
-    if (isPathnameUrl) {
+    if (isUrl) {
       return pathname;
     } else {
       return url.resolve(registry, pathname);
     }
   }
 
+  getPortOrDefaultPort(port: ?string, protocol: ?string): ?string {
```
BYK: Just say `getPort`? Also you can simplify your function as follows:
```
const DEFAULT_PORTS = Object.freeze({
    'http:': '80',
    'https:': '443',
});
/**
  * Remove the port if it is the default for the protocol, fallback to emptry string if no port provided.
*/
static getPort(port: ?string, protocol: ?string): string {
    return !port || port === DEFAULT_PORTS[protocol] ? '' : port;
}
```

src/registries/npm-registry.js
```diff
@@ -74,21 +76,50 @@ export default class NpmRegistry extends Registry {
...
+    const requestPath = requestParsed.path || '';
+    const registryPath = registryParsed.path || '';
+    const customHostSuffix = this.getRegistryOrGlobalOption(registry, 'custom-host-suffix');
+
+    return (
+      requestHost === registryHost &&
+      requestPort === registryPort &&
+      (requestPath.startsWith(registryPath) ||
+        // For some registries, the package path does not prefix with the registry path
+        (typeof customHostSuffix === 'string' && customHostSuffix.length > 0 && requestHost.endsWith(customHostSuffix)))
```
BYK: Why can't we just do `(customHostSuffix && requestHost.endsWith(customHostSuffix)`?

__tests__/registries/npm-registry.js
```diff
@@ -146,15 +164,68 @@ describe('request', () => {
...
+    expect(npmRegistry.isRequestToRegistry('http://foo.bar/foo/bar/baz', 'http://foo.bar/foo/')).toBe(true);
+
+    expect(
+      npmRegistry.isRequestToRegistry(
+        'http://foo.bar/foo/00000000-1111-4444-8888-000000000000/baz',
+        'http://foo.bar/foo/',
+      ),
+    ).toBe(true);
+
+    expect(npmRegistry.isRequestToRegistry('https://foo.bar:443/foo/bar/baz', 'https://foo.bar/foo/')).toBe(true);
```
BYK: How about this:
```
const validRegistryUrls = Object.feeze({
    'https://foo.bar:443/foo/bar/baz': 'https://foo.bar/foo/',
    'https://foo.bar/foo/bar/baz': 'https://foo.bar:443/foo/',
   ....
});

const invalidRegistryUrls = Object.freeze({
    'https://wrong.thing/foo/bar/baz': 'https://foo.bar/foo/',
    ...
});

Object.keys(validRegistryUrls).forEach(url => expect(npmRegistry.isRequestToRegistry(url, validRegistryUrls[url]).toBeTruthy());
Object.keys(invalidRegistryUrls).forEach(url => expect(npmRegistry.isRequestToRegistry(url, validRegistryUrls[url]).toBeFalsy());
```

BYK: You may also wanna look at this library which does all this logic for you: https://yarnpkg.com/en/package/normalize-url

src/registries/npm-registry.js
soda0289: I originally did have it like that but `flow check` was failing. I think the return value of  `getRegistryOrGlobalOption()` is mixed and therefore would need to assert it is a string before performing string operations.

I can remove the `.length` check since that should not be necessary.

__tests__/registries/npm-registry.js
soda0289: I used the normalize-url library cleans up the getPort() logic

BYK: >I used the normalize-url library cleans up the getPort() logic

Did you push these changes? I don't see them yet.

src/registries/npm-registry.js
BYK: Oh, makes sense. Well, may be you can do
```
const customHostSuffix = String(this.getRegistryOrGlobalOption(registry, 'custom-host-suffix') || '');
```

So it is always a string? The only downside is if someone someone sets this to `null` or something, it will be converted into the string `'null'`. Not too concerned about that since AFAIK `typeof` checks are expensive.

__tests__/registries/npm-registry.js
soda0289: I had to use arrays instead of Objects since some tests had duplicate keys.

soda0289:
@BYK 

I was able to remove the getPort() function and use the library normalize-url to handle that logic. Also normalizes unicode urls which might not have been checked before. I cleaned up the tests by creating an array of the tests cases. Let me know if anything else can be improved.

src/registries/npm-registry.js
```diff
@@ -72,9 +73,39 @@ export default class NpmRegistry extends Registry {
     return name.replace('/', '%2f');
   }
 
+  getRequestUrl(registry: string, pathname: string): string {
+    const isUrl = /^https?:/.test(pathname);
```
BYK: Nitpick: technically `git://xxx.com/yyy` is also a URL so you'd want to name this as `isHTTPUrl` or something.

src/registries/npm-registry.js
```diff
@@ -72,9 +73,39 @@ export default class NpmRegistry extends Registry {
...
+      return pathname;
+    } else {
+      return url.resolve(registry, pathname);
+    }
+  }
+
+  isRequestToRegistry(requestUrl: string, registryUrl: string): boolean {
+    const normalizedRequestUrl = normalizeUrl(requestUrl);
+    const normalizedRegistryUrl = normalizeUrl(registryUrl);
+    const requestParsed = url.parse(normalizedRequestUrl);
```
BYK: Simplification:
```
const {requestHost, requestPath} = url.parse(normalizeUrl(requestUrl));
const {registryHost, registryPath} = url.parse(normalizeUrl(registryUrl));
```

Also, `registryUrl` should not change frequently so we may consider memoizing parsing that in a future patch.

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,193 @@ describe('normalizeConfig', () => {
...
+  test('should call requestManager.request with url', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const url = 'https://github.com/yarnpkg/yarn.tgz';
+
+    npmRegistry.request(url);
+
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
```
BYK: [toHaveBeenCalledWith()](https://facebook.github.io/jest/docs/expect.html#tohavebeencalledwitharg1-arg2-)

__tests__/registries/npm-registry.js
```diff
@@ -39,3 +39,193 @@ describe('normalizeConfig', () => {
...
+
+    expect(requestParams.url).toBe(url);
+  });
+
+  test('should not add authorization header if pathname not to registry', () => {
```
BYK: s/pathname/request

src/registries/npm-registry.js
plmrry: This will break protocol relative URLs, which are generated by default in `.npmrc`

BYK: Oh. Issue filed somewhere?

BYK: https://github.com/yarnpkg/yarn/pull/4347Oh. Issue filed somewhere?

Issue #3954, opened 2017-07-18
Scripts that cause a segmentation fault exit with status 0


rubennorte:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

Yarn returns an exit status 0 when a script inside the package.json "scripts" causes a segmentation fault.

This only happens when there is a single command in the script (e.g.: `"test": "./segfault"`). When there are several commands in the script (e.g.: `"test-alt": "true && ./segfault"`) it returns an exit status 1.

**If the current behavior is a bug, please provide the steps to reproduce.**

See this [gist](https://gist.github.com/rubennorte/1ac1ed80ef3ef77d2aecc2335c1f59db)

**What is the expected behavior?**

When a script in the package.json exits due to a segmentation fault yarn exits with an error exit code.

Yarn v0.27.5, Node.js 6.9.1, Mac OS X & Linux

PR #3995, opened 2017-07-23
Handle scripts failing with exit signals (fixes: #3954)

onurtemizkan:
**Summary**
Fixes: #3954, Adds error handling for scripts that are failing with a signal such as `SIGBUS`, `SIGSEGV`, etc.

In current behavior, yarn checks the exit code of a script to show errors and fail the process. But certain errors make node child-processes to exit with a signal instead. (e.g. the case in #3954)

This fix adds support for handling exit signals from child processes and shows more informative error messages.

**Note:**  This implementation handles exit signals priorly, assuming exit signals are pointing to more critical low-level issues.

**Test plan**

Added 3 test cases (one for _zero_ exit code, one for _non-zero_ exit code, one for _exit signal_)

### DIFF
--- a/__tests__/fixtures/lifecycle-scripts/script-segfault/package.json
+++ b/__tests__/fixtures/lifecycle-scripts/script-segfault/package.json
@@ -1,9 +1,5 @@
 {
   "scripts": {
-    "compile": "gcc segfault.c -g -o segfault",
-    "pretest": "yarn run compile",
-    "test": "./segfault",
-    "pretest-alt": "yarn run compile",
-    "test-alt": "./segfault && true"
+    "test": "node app.js"
   }
 }
--- a/__tests__/fixtures/lifecycle-scripts/script-segfault/package.json
+++ b/__tests__/fixtures/lifecycle-scripts/script-segfault/package.json
@@ -1,0 +1,9 @@
+{
+  "scripts": {
+    "compile": "gcc segfault.c -g -o segfault",
+    "pretest": "yarn run compile",
+    "test": "./segfault",
+    "pretest-alt": "yarn run compile",
+    "test-alt": "./segfault && true"
+  }
+}
--- a/__tests__/lifecycle-scripts.js
+++ b/__tests__/lifecycle-scripts.js
@@ -142,3 +142,23 @@ test('should inherit existing environment variables when setting via yarnrc', as
   expect(stdout).toMatch(/^RAB$/m);
   expect(stdout).toMatch(/^FOO$/m);
 });
+
+test('should not show any error messages when script ends successfully', async () => {
+  const stdout = await execCommand('test', 'script-success');
+
+  expect(stdout).toMatch(/Done in /);
+});
+
+test('should show correct error message when the script ends with an exit code', async () => {
+  await execCommand('test', 'script-fail').catch(error => {
+    expect(error.message).toMatch(/Command failed with exit code /);
+    expect(error.code).not.toBe(0);
+  });
+});
+
+test('should show correct error message when the script ends an exit signal', async () => {
+  await execCommand('test', 'script-segfault').catch(error => {
+    expect(error.message).toMatch(/Command failed with signal /);
+    expect(error.code).not.toBe(0);
+  });
+});
--- a/__tests__/lifecycle-scripts.js
+++ b/__tests__/lifecycle-scripts.js
@@ -158,7 +158,7 @@ test('should show correct error message when the script ends with an exit code',
 
 test('should show correct error message when the script ends an exit signal', async () => {
   await execCommand('test', 'script-segfault').catch(error => {
-    expect(error.message).toMatch(/Command failed with signal /);
+    expect(error.message).toMatch(/Command failed with signal "SIGSEGV"/);
     expect(error.code).not.toBe(0);
   });
 });
--- a/__tests__/lifecycle-scripts.js
+++ b/__tests__/lifecycle-scripts.js
@@ -144,21 +144,15 @@ test('should inherit existing environment variables when setting via yarnrc', as
 });
 
 test('should not show any error messages when script ends successfully', async () => {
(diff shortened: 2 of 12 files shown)


__tests__/fixtures/lifecycle-scripts/script-segfault/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "scripts": {
+    "compile": "gcc segfault.c -g -o segfault",
```
BYK: 😮  Will this work on AppVeyor (Windows)

I'm not sure if Windows has the concept of "exiting with a signal" so it may be better to disable this test on Windows.

src/util/child.js
```diff
@@ -94,8 +94,20 @@ export function spawn(
...
+              [
+                'Command failed.',
+                `Exit signal: ${signal}`,
+                `Command: ${program}`,
+                `Arguments: ${args.join(' ')}`,
+                `Directory: ${opts.cwd || process.cwd()}`,
+                `Output:\n${stdout.trim()}`,
+              ].join('\n'),
+            );
+            err.EXIT_SIGNAL = signal;
```
BYK: I'd still add `EXIT_CODE` property.

src/util/child.js
```diff
@@ -94,8 +94,20 @@ export function spawn(
           processingDone = true;
         }
 
-        proc.on('close', (code: number) => {
-          if (code >= 1) {
+        proc.on('close', (code: number, signal: string) => {
+          if (signal) {
+            err = new SpawnError(
+              [
+                'Command failed.',
```
BYK: Looks like this message generation part is repeating itself quite a bit. May be you can DRY this up a bit?

__tests__/lifecycle-scripts.js
```diff
@@ -142,3 +142,23 @@ test('should inherit existing environment variables when setting via yarnrc', as
   expect(stdout).toMatch(/^RAB$/m);
   expect(stdout).toMatch(/^FOO$/m);
 });
+
+test('should not show any error messages when script ends successfully', async () => {
+  const stdout = await execCommand('test', 'script-success');
+
+  expect(stdout).toMatch(/Done in /);
```
BYK: This text may disappear in the future. May be just do `expect().not.toThrow`?

__tests__/lifecycle-scripts.js
```diff
@@ -142,3 +142,23 @@ test('should inherit existing environment variables when setting via yarnrc', as
...
+});
+
+test('should show correct error message when the script ends with an exit code', async () => {
+  await execCommand('test', 'script-fail').catch(error => {
```
BYK: This test would pass if no error is thrown (if the script doesn't fail). I'd use `expect().toThrow()` or even better, [toThrowErrorMatchingSnapshot()](https://facebook.github.io/jest/docs/expect.html#tothrowerrormatchingsnapshot)

__tests__/lifecycle-scripts.js
```diff
@@ -142,3 +142,23 @@ test('should inherit existing environment variables when setting via yarnrc', as
...
+    expect(error.message).toMatch(/Command failed with exit code /);
+    expect(error.code).not.toBe(0);
+  });
+});
+
+test('should show correct error message when the script ends an exit signal', async () => {
+  await execCommand('test', 'script-segfault').catch(error => {
```
BYK: Same as above, use `toThrow*()`

src/util/execute-lifecycle-script.js
```diff
@@ -268,7 +268,11 @@ export async function execCommand(stage: string, config: Config, cmd: string, cw
     return Promise.resolve();
   } catch (err) {
     if (err instanceof SpawnError) {
-      throw new MessageError(reporter.lang('commandFailed', err.EXIT_CODE));
+      if (err.EXIT_SIGNAL) {
```
BYK: DRY this up.

onurtemizkan:
@BYK while making updates on tests, I used [`resolves` and `rejects`](https://facebook.github.io/jest/docs/expect.html#resolves) helpers, which are added at Jest v20.0.0. We are currently using Jest 20.0.4 but our flow-type definitions for Jest do not include those two properties. So the flow checks will fail on CI. Should I raise another issue for updating the flow types or is it enough to add those two type defs with this PR?

__tests__/lifecycle-scripts.js
```diff
@@ -144,21 +144,15 @@ test('should inherit existing environment variables when setting via yarnrc', as
...
+  await expect(execCommand('test', 'script-success')).resolves.toBeDefined();
 });
 
-test('should show correct error message when the script ends with an exit code', async () => {
-  await execCommand('test', 'script-fail').catch(error => {
-    expect(error.message).toMatch(/Command failed with exit code /);
-    expect(error.code).not.toBe(0);
-  });
+test('should throw error when the script ends with an exit code', async () => {
+  await expect(execCommand('test', 'script-fail')).rejects.toBeDefined();
```
onurtemizkan: @BYK this test (also the test below) should fail when the script unexpectedly succeeds.

__tests__/lifecycle-scripts.js
```diff
@@ -142,3 +142,17 @@ test('should inherit existing environment variables when setting via yarnrc', as
   expect(stdout).toMatch(/^RAB$/m);
   expect(stdout).toMatch(/^FOO$/m);
 });
+
+test('should not show any error messages when script ends successfully', async () => {
+  await expect(execCommand('test', 'script-success')).resolves.toBeDefined();
```
BYK: You can simply return these, making it a one-liner:
```
test('should not show any error messages when script ends successfully', async () =>
  expect(execCommand('test', 'script-success')).resolves.toBeDefined()
);
```

onurtemizkan: 👍

PR #3996, opened 2017-07-23
added ignore patterns when resolving workspaces

### LINKED ISSUE #3986: Yarn more than one workspace
Bug

Doing some yarn command a second time using workspaces feature may fail with `There are more than one workspace with name "XXX"` error if a package is conflicting and not hoisted at top-level and when the `workspaces` option is a glob englobing pattern.

See https://github.com/maoueh/yarn-workspace-already-exist-bug that has all the instructions, copied here for easiest repro:

 1. Clone repository

    ```
    $ git clone https://github.com/maoueh/yarn-workspace-already-exist-bug.git /tmp/yarn-workspace-already-exist-bug
    ```

 1. Yarn install once

    ```
    $ cd /tmp/yarn-workspace-already-exist-bug
    $ yarn install
    ```

 1. Yarn install a second time to see the error:

    ```
    $ yarn install
    yarn install v0.28.1
    error There are more than one workspace with name "ramda"
    ```

There should not be a problem doing a second `yarn install`.

**Extra Info**

The problem I think is simply that Yarn is looking in `node_modules` folders to determined workspaces. For example, with the pattern `packages/**/*`, and with the tree:

```
├── node_modules
├── package.json
├── packages
│   ├── a
│   │   └── package.json
│   ├── b
│   │   ├── node_modules
│            └── ramda
│   │   └── package.json
│   └── c
│       ├── node_modules
│            └── ramda
│       └── package.json
└── yarn.lock
```

Yarn is finding `packages/b/nodes_modules/ramda` and `packages/c/nodes_modules/ramda` as being a valid workspace while it's a simple dependency.

Yarn 0.28.1
Node 7.9
Mac OS X

arslanarshad31:
**Summary**
Bug Fix - Resolves #3986 

Added an ignore pattern for getting `package.json` and `yarn.json` files when resolving workspaces.
Ignore pattern looks like:  `packages/**/*/node_modules/*/+(package.json|yarn.json)`
This should prevent any problems when running `yarn install` second time as the node_modules will be ignored.


src/config.js
```diff
@@ -604,10 +604,14 @@ export default class Config {
 
     const registryFilenames = registryNames.map(registryName => this.registries[registryName].constructor.filename);
     const trailingPattern = `/+(${registryFilenames.join(`|`)})`;
+    const ignorePatterns = this.registryFolders.map(folder => `/${folder}/*/+(${registryFilenames.join(`|`)})`);
```
BYK: You should "regexp-escape" `registryFilenames` before using them in a regex.

BYK:
Also, if we can have tests for this, that'd make the PR even better! :)

src/config.js
BYK: Turns out we don't need any escaping here :)

Issue #4006, opened 2017-07-24
yarn init doesn't understand repo shortcuts 


mohsen1:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**
when providing org/repo to yarn init repo section it remains as is. Example: 
https://github.com/mohsen1/universal-async-component/blob/master/package.json#L6
**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn init
// accept accept...
repo: org/repo
```

**What is the expected behavior?**

it should resolve the shortcut to proper GitHub repo URL

**Please mention your node.js, yarn and operating system version.**
yarn v0.23.3 node 8

FrederickGeek8:
The commit referenced in this Github issue should provide the functionality you are asking for. I'm going to see if I can't setup some tests and then I'll submit a PR.

olingern:
@BYK Looks like this can be closed?

BYK:
@olingern, yup. Thanks for the heads up! :)

PR #4012, opened 2017-07-24
Add support for Github URL shorthand to yarn init

FrederickGeek8:
**Summary**
This pull request fulfills #4006 – an issue that requests a feature be added to `yarn init` so that Github shorthand i.e. a value of `yarnpkg/yarn` can be used in the repository field.

**Test plan**
I have added a test to the test suite. Outside of that, this is the difference in usage:

**Before:** Running `yarn init` with default options other than the repository field which is set to `yarnpkg/yarn` would yield a package.json that has the repository URL set to `yarnpkg/yarn` 

**After:** Running `yarn init` with default options other than the repository field which is set to `yarnpkg/yarn` would yield a package.json that has the repository URL set to `https://github.com/yarnpkg/yarn`


src/cli/commands/init.js
```diff
@@ -77,6 +77,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       key: 'repository',
       question: 'repository url',
       default: extractRepositoryUrl(repository),
+      isShorthand: /^[a-zA-Z0-9-_]*\/[a-zA-Z0-9-_]+$/,
```
BYK: Let's reuse the existing matcher from https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/github-resolver.js#L17

src/cli/commands/init.js
```diff
@@ -136,6 +137,9 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
         }
       } else {
         answer = (await reporter.question(question)) || def;
+        if (entry.isShorthand && entry.isShorthand.exec(answer)) {
```
BYK: Use `entry.isShorthand.test(answer)` instead.

Also not sure why you need to test `entry.isShorthand` since you define it above?

FrederickGeek8: It was to discern the repository case similar to the `If (entry.validation)` a couple lines up. I'll make the changes you requested layer today.

FrederickGeek8:
I've made the requested changes and am now attempting to figure out the test system. Is there any documentation you could point me to, or any tips for working with the `init` command? I'm struggling to see how to pass parameters to it.

src/cli/commands/init.js
```diff
@@ -136,6 +138,9 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
         }
       } else {
         answer = (await reporter.question(question)) || def;
+        if (entry.isShorthand && entry.isShorthand(answer)) {
+          answer = 'https://github.com/' + answer;
+        }
```
arcanis: Please use backticks when concatenating strings :)

arcanis: Also, this is something that should be handled outside of the regular field resolution process imo, since it's only meant to be used on the github repository and is not generic at all. Something like this after all the questions have been asked:

```js
if (pkg.repository && GitHubResolver.isVersion(pkg.repository)) {
    pkg.repository = `http://github.com/${answer}`;
}
```

arcanis:
As for the `init` tests, the flags are passed as first parameter when instanciating the `Init` class. For instance, the [following code](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/init.js#L24) instanciates `yarn init -y`.

FrederickGeek8:
@arcanis I am aware, however what I was wondering was if there is any standard way to override `ConsoleReporter.question` and provide test inputs instead so that a fitting shorthand repository URL can be tested.

src/cli/commands/init.js
BYK: >Please use backticks when concatenating strings :)

We may wanna enable the following ESLint rules then:

- http://eslint.org/docs/rules/no-useless-concat
- http://eslint.org/docs/rules/prefer-template

BYK:
>@arcanis I am aware, however, what I was wondering was if there is any established way to override ConsoleReporter.question and provide test inputs instead so that a fitting shorthand repository URL can be tested

There's this test from https://github.com/yarnpkg/yarn/blob/master/__tests__/reporters/console-reporter.js#L134-L155
```
test('ConsoleReporter.select', async () => {
  expect(
    await getConsoleBuff(async function(r, streams): Promise<void> {
      streams.stdin.on('resume', function() {
        streams.stdin.send('1\n', 'ascii');
        streams.stdin.end();
      });

      const res = await r.select('Ayo?', 'Select one', [
        {
          name: 'foo',
          value: 'foo',
        },
        {
          name: 'bar',
          value: 'bar',
        },
      ]);
      expect(res).toBe('foo');
    }),
  ).toMatchSnapshot();
});
```

The core idea is using `stdin` to send the data. This kind of tests may be a bit more fragile tho, keep that in mind.

BYK:
Btw, LGTM except for the tests.

FrederickGeek8:
I've been mulling over relevant code and I'm starting to think that creating tests for this at this moment is a bit above my current understanding of the repository. I can maybe add tests later, but it will take a while and a bit more studying.

FrederickGeek8:
I created a new class called "TestReporter" which for now just allows you to override the question inputs for the BaseReporter class. Using this I've rolled out tests for the changes described in this PR, as well as tests (or rather test) for TestReport itself.

BYK:
@FrederickGeek8 no worries. I can try helping with tests. Which part confuses you the most?

Also you can hop over to our Discord channels and ping me directly there if you prefer real-time help.

FrederickGeek8:
@BYK see my previous comment. I've completed writing tests using a new Reporter I wrote.

src/reporters/test-reporter.js
```diff
@@ -0,0 +1,23 @@
+/* @flow */
+import type {QuestionOptions} from './types.js';
+import BaseReporter from './base-reporter.js';
+
+export default class TestReporter extends BaseReporter {
```
BYK: Nice! I'd base this on `BufferReporter` which really helps with testing. I'll do this and merge the patch afterward.

__tests__/commands/init.js
```diff
@@ -15,6 +15,9 @@ const execInit = buildRun.bind(
   ConsoleReporter,
   fixturesLoc,
   (args, flags, config, reporter, lockfile): Promise<void> => {
+    if (flags.hasOwnProperty('questionMap')) {
```
BYK: Not a fan of this change but I'm okay with it until I can find a better solution.

__tests__/commands/init.js
```diff
@@ -33,6 +36,29 @@ test.concurrent('init --yes should create package.json with defaults', (): Promi
...
+    'repository url': 'user/repo',
+    author: '',
+    license: '',
+  };
+  return execInit({questionMap: inputMap}, 'init-github', async (config): Promise<void> => {
+    const {cwd} = config;
+    const manifestFile = await fs.readFile(path.join(cwd, 'package.json'));
+    const manifest = JSON.parse(manifestFile);
+
+    expect(manifest.name).toEqual(path.basename(cwd));
```
BYK: I'd use the awesome `toMatchSnapshot` feature here.

PR #4026, opened 2017-07-26
Chore: enable no-process-exit lint rule

BYK:
**Summary**

A follow up to #3955 to disable further usage of `process.exit()`.

**Test plan**

`yarn lint` should not produce any errors.


cpojer:
Do you think it makes sense to add code comments on why the remaining `process.exit` calls are necessary?

BYK:
>Do you think it makes sense to add code comments on why the remaining process.exit calls are necessary?

Why, I do. Will update.

BYK:
@cpojer ready!

Issue #4016, opened 2017-07-25
Scoped packages 403 on yarn install with empty cache


lukeggchapman:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Having set up a private npm repo by running:
`npm login --registry=https://npm.< private host >.com --scope=@scoped`

Then after pulling down a repo with an existing `yarn.lock` file that contains a `@scoped/package` and running `yarn` you get a 403 on that scoped package even though you are logged in to that registry and the token exists in the ~/.npmrc file.

**If the current behavior is a bug, please provide the steps to reproduce.**
The problem seems to be warming up the cache because this issue is fixed for any package moving forward by running `yarn add @scoped/package` once and reintroduced after cleaning the cache.

Having a `.yarnrc` in the repo with `"@scoped:registry" "https://npm.< private host >.com/"` does not fix the issue.

**What is the expected behavior?**
Scoped packages that point to a seperate npm registry should work when running a fresh install (`yarn`) and should not require the packages to be individually added to the yarn cache.

**Please mention your node.js, yarn and operating system version.**
OSX 10.12.5
node.js 8.2.1
yarn 0.27.5

lukeggchapman:
@arcanis mentioned it here:
https://github.com/yarnpkg/yarn/issues/3765#issuecomment-312386851

Yarn 0.25.4 also has this issue.

BYK:
I'm leaning to closing this as a duplicate of #3765. What do you think? This should work on master if that's the case and fail otherwise.

lukeggchapman:
I believe #3765 was released as part of 0.27.5, which is where I was seeing this problem.
I've also just tested the latest master and it's still an issue.

BYK:
@lukeggchapman we have merged #3987 quite recently and it should address the issue AFAIK. Are you sure your credentials are up to date since NPM did some changes recently and some people reported their auth token not working.

If still broken, can you provide us/me a simple repro case so we can test and fix it?

lukeggchapman:
@BYK I pulled master down at the time I commented.
The credentials are against a privately hosted registry, and works fine with `yarn add @scoped/pkg`

I've tracked down the bug and should have a PR for it shortly.

lukeggchapman:
This piece of code only works for packageName (`yarn add`), not pathName `yarn install`.
https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L221-L223

[...]

PR #4027, opened 2017-07-26
Scoped packages 403 on yarn install with empty cache

lukeggchapman:
**Summary**

Fix scoped packages not adding correct auth headers (#4016)

**Test Plan**

Having set up config for a scoped registry with an auth token. It should be possible to run `yarn cache clean` followed by `yarn install` without having the scoped packages 403

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -175,6 +175,24 @@ describe('request', () => {
 
     expect(requestParams.headers.authorization).toBe('Bearer testAuthToken');
   });
+
+  test('should add authorization header with correct token for scoped package', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    const url = 'https://some.other.registry/@testScope/yarn.tgz';
+
+    npmRegistry.config = {
+      '//some.other.registry/:_authToken': 'testScopedAuthToken',
+      '@testScope:registry': '//some.other.registry/',
+    };
+    npmRegistry.request(url);
+
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
+
+    expect(requestParams.headers.authorization).toBe('Bearer testScopedAuthToken');
+  });
 });
 
 describe('isRequestToRegistry functional test', () => {
@@ -227,5 +245,40 @@ describe('isRequestToRegistry functional test', () => {
     expect(npmRegistry.isRequestToRegistry('http://pkgs.host.com:80/foo/bar/baz', 'http://pkgs.host.com/bar/baz')).toBe(
       true,
     );
+  });
+});
+
+describe('getScope functional test', () => {
+  describe('matches scope correctly', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+
+    test('in package names', () => {
+      const packageNames = [
+        ['normal', ''],
+        ['normal-package', ''],
+        ['@scopedNoPkg', ''],
+        ['@scoped/pkg', '@scoped'],
+        ['invalid@scope/pkg', ''],
+      ];
+
+      packageNames.forEach(([packageName, scope]) => {
+        expect(npmRegistry.getScope(packageName)).toEqual(scope);
+      });
+    });
+
+    test('in pathname', () => {
+      const pathnames = [
+        ['http://foo.bar:80/foo/bar/baz', ''],
+        ['http://foo.bar:80/@scope/bar/baz', '@scope'],
+        ['http://foo.bar:80/@scope%2fbar/baz', '@scope'],
+        ['http://foo.bar:80/invalid@scope%2fbar/baz', ''],
+      ];
+
+      pathnames.forEach(([pathname, scope]) => {
+        expect(npmRegistry.getScope(pathname)).toEqual(scope);
+      });
+    });
   });
 });
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -176,7 +176,7 @@ describe('request', () => {
     expect(requestParams.headers.authorization).toBe('Bearer testAuthToken');
   });
 
-  test('should add authorization header with token for scope if pathname is to registry and is scoped package', () => {
+  test('should add authorization header with token for custom registries with a scoped package', () => {
     const testCwd = '.';
     const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
     const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
@@ -258,7 +258,7 @@ describe('getScope functional test', () => {
       const packageNames = [
         ['normal', ''],
         ['normal-package', ''],
-        ['@scopedNoPkg', ''],
+        ['@scopedNoPkg', '@scopedNoPkg'],
         ['@scoped/pkg', '@scoped'],
         ['invalid@scope/pkg', ''],
       ];
@@ -271,7 +271,7 @@ describe('getScope functional test', () => {
     test('in pathname', () => {
       const pathnames = [
         ['http://foo.bar:80/foo/bar/baz', ''],
-        ['http://foo.bar:80/@scopedNoPkg', ''],
+        ['http://foo.bar:80/@scopedNoPkg', '@scopedNoPkg'],
         ['http://foo.bar:80/@scope/bar/baz', '@scope'],
         ['http://foo.bar:80/@scope%2fbar/baz', '@scope'],
         ['http://foo.bar:80/invalid@scope%2fbar/baz', ''],
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -164,7 +164,7 @@ describe('request', () => {
     const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
     const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
 
-    const url = 'https://registry.npmjs.org/@testScope/yarn.tgz';
+    const url = 'https://registry.npmjs.org/@testScope%2fyarn.tgz';
 
     npmRegistry.config = {
       _authToken: 'testAuthToken',
@@ -181,7 +181,7 @@ describe('request', () => {
     const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
     const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
 
-    const url = 'https://some.other.registry/@testScope/yarn.tgz';
+    const url = 'https://some.other.registry/@testScope%2fyarn.tgz';
 
     npmRegistry.config = {
       '//some.other.registry/:_authToken': 'testScopedAuthToken',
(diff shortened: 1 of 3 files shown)


src/registries/npm-registry.js
```diff
@@ -48,6 +48,11 @@ function isPathConfigOption(key: string): boolean {
   return PATH_CONFIG_OPTIONS.indexOf(key) >= 0;
 }
 
+function isScopedPackage(packageIdent: string): boolean {
+  // scoped package names will begin with '@', scoped path names will contain '/@'
+  return !!packageIdent.match(/^@|\/@/);
```
BYK: Use `/^@|\/@/.test(packageIdent)` instead please.

src/registries/npm-registry.js
```diff
@@ -218,21 +221,27 @@ export default class NpmRegistry extends Registry {
     }
   }
 
-  getScope(packageName: string): string {
-    return !packageName || packageName[0] !== '@' ? '' : packageName.split(/\/|%2f/)[0];
+  getScope(packageIdent: string): string {
+    // Matches '@scope' in a packageName or pathname, otherwise returns ''
+    if (packageIdent && isScopedPackage(packageIdent)) {
```
BYK: No need for the `isScopedPackage` check since the below matcher will fail anyways.

src/registries/npm-registry.js
```diff
@@ -218,21 +221,27 @@ export default class NpmRegistry extends Registry {
     }
   }
 
-  getScope(packageName: string): string {
-    return !packageName || packageName[0] !== '@' ? '' : packageName.split(/\/|%2f/)[0];
+  getScope(packageIdent: string): string {
```
BYK: This function better written as:
```
getScope(packageIdent: string): string {
    const match = packageIdent.match(/(^@|(?!\/)@).+?((?=%2f)|(?=\/))/);
    return match && match[1] || '';
}

File: src/registries/npm-registry.js [hunk 4]
```diff
@@ -218,21 +221,27 @@ export default class NpmRegistry extends Registry {
...
+    // Matches '@scope' in a packageName or pathname, otherwise returns ''
+    if (packageIdent && isScopedPackage(packageIdent)) {
+      const matched = packageIdent.match(/(^@|(?!\/)@).+?((?=%2f)|(?=\/))/);
+      return (matched && matched[0]) || '';
```
BYK: You want `matched[1]` if you want the first group, not `matched[0]`

src/registries/npm-registry.js
```diff
@@ -218,21 +221,27 @@ export default class NpmRegistry extends Registry {
     }
   }
 
-  getScope(packageName: string): string {
-    return !packageName || packageName[0] !== '@' ? '' : packageName.split(/\/|%2f/)[0];
+  getScope(packageIdent: string): string {
+    // Matches '@scope' in a packageName or pathname, otherwise returns ''
+    if (packageIdent && isScopedPackage(packageIdent)) {
+      const matched = packageIdent.match(/(^@|(?!\/)@).+?((?=%2f)|(?=\/))/);
```
BYK: This regular expression is very hard to follow, can use some comments. Also what's with the `%2f` part? We should unescape the URL and run the matcher on it instead of doing this.

__tests__/registries/npm-registry.js
```diff
@@ -175,6 +175,24 @@ describe('request', () => {
 
     expect(requestParams.headers.authorization).toBe('Bearer testAuthToken');
   });
+
+  test('should add authorization header with correct token for scoped package', () => {
```
BYK: This test case is very similar to the previous one so I'd either make them a bit more differentiated or fix the previous one.

src/registries/npm-registry.js
lukeggchapman: You want the complete match (matched[0]), not any of the groups. The first group is '^@' or '/@'

src/registries/npm-registry.js
lukeggchapman: I matched on `%2f` because I didn't have the domain knowledge to know for what reason it was escaped in the pathname. If it's not to differentiate them from other forward slashes I could add an extra `.replace('%2f', '/')` call in this line to make the regex more readable.

src/registries/npm-registry.js
lukeggchapman: I've reworked the regex so this is possible.

__tests__/registries/npm-registry.js
lukeggchapman: I don't believe it is.
This tests that a scope configured with a non-default registry is receiving the token specified for that scope.
The previous test is testing the use of the default _authToken when fetching from a scope that has no configured registry.
I'm happy to remove the previous test if you believe it does not add value.

lukeggchapman:
Thanks for the review @BYK! I've cleaned up some things and It's ready for you to review again.

soda0289:
@lukeggchapman Thanks for fixing this and adding some more tests!

 I have a question about how we determine if a package is scoped and what that scope is.
Using the regex: `/(^@|(?!\/)@).+?((?=%2f)|(?=\/))/`
Won't this mark the following url as having a scope:
`http://my-registry.com/@registry/foo/bar.tgz`
 where registry could be: `http://my-registry.com/@registry/foo`

OR this url having an incorrect scope:
`http://my-registry.com/test@test/@foo/bar.tgz` 
where registry could be `http://my-registry.com/test@test/` and the actual scope `@foo`

Not sure there can be a perfect solution or if it can be improved. Maybe for URL we only selected the last `/` as scope and package name separator but that might not work if package name contains a `/`, which I'm not sure is allowed

lukeggchapman:
@soda0289 You're welcome, happy to contribute.

I've recently refactored the regex to be `/(^@|\/@)(.+?)(?=\/)/` which will resolve the problem with `http://my-registry.com/test@test/@foo/bar.tgz`.

With the first issue I've made the assumption that scoped packages will only ever have one `@/` based off an [existing assumption.](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L119)

It would be possible to rewrite the regex to grab only the last `@scope/pkg.tgz`, but with registries that have an `@` already always being marked as scoped then more work would need to be done to support them properly.

src/registries/npm-registry.js
BYK: Oh, let's make those groups non-capturing then? (haven't seen the latest code yet)

src/registries/npm-registry.js
BYK: You need `decodeURIComponent(str)` instead.

BYK:
>It would be possible to rewrite the regex to grab only the last @scope/pkg.tgz, but with registries that have an /@ already always being marked as scoped then more work would need to be done to support them properly

This is a bit concerning. Should we file an issue for this so we don't forget? :(

src/registries/npm-registry.js
```diff
@@ -222,13 +222,9 @@ export default class NpmRegistry extends Registry {
...
   getScope(packageIdent: string): string {
-    // Matches '@scope' in a packageName or pathname, otherwise returns ''
-    if (packageIdent && isScopedPackage(packageIdent)) {
-      const matched = packageIdent.match(/(^@|(?!\/)@).+?((?=%2f)|(?=\/))/);
-      return (matched && matched[0]) || '';
-    }
-
-    return '';
+    // Matches strings beginning with @ or contain /@ up until the next forward slash
+    const match = packageIdent.replace('%2f', '/').match(/(^@|\/@)(.+?)(?=\/)/);
```
BYK: Looking at this closely, I really can't understand a few things:

 - Why do we even have URL-encoded parts in the ident? That doesn't look right so it should be fixed where the value originally comes from.
 - What does the `?` at the end of `(.+?)` part stand for? Is it for the query string delimiter? If so, why do we need to include it in the match?

I feel like the following version is clearer and more accurate:
```
// Matches the first path segment that starts with an `@`. The RegEx is constructed as follows:
// `(?:^|\/)` Match either the start of the string or a `/` but don't capture
// `(@[^?\/]+)` Match a string starting with an `@` and not having any `/` or `?` in it and capture
// `(?:[?\/]|$)/)` Match a path delimiter, a query string delimiter or the end of the string but don't capture
const match = packageIdent.match(/(?:^|\/)(@[^?\/]+)(?:[?\/]|$)/);
return match && match[1] || '';
```

__tests__/registries/npm-registry.js
```diff
@@ -175,6 +175,24 @@ describe('request', () => {
 
     expect(requestParams.headers.authorization).toBe('Bearer testAuthToken');
   });
+
+  test('should add authorization header with token for scope if pathname is to registry and is scoped package', () => {
```
BYK: Let's explicitly say "custom registry" in the test name:
`should add authorization header with token for custom registries with a scoped package`

src/registries/npm-registry.js
lukeggchapman: * It's only the forward slashes in the package name (eg (`https://private.npm.repo.com/@foo%2fbar`) and It would be good to find where the pathname is getting stored/retrieved from to see if it is necessary to keep it in that format. I assume it is related to [this](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L72).
* I'm using the `?` after a quantifier for [non-greedy matching](https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions#special-questionmark), so it matches only up to the next forward slash that is in a lookahead so it doesn't get captured.
* I haven't seen a query string in a pathname yet, so it is not something I am catering for. Is that possible? Do you have some examples I could add to the test cases?

Your regex passes the test cases so I'll update it with that, with the added benefit of you having considered the query strings. :+1:

lukeggchapman:
I'll create an issue for registries that contain `/@` always being marked as issues tomorrow.
Also I can look more into where the escaped forward slashes in package names within paths are coming from.

__tests__/registries/npm-registry.js
```diff
@@ -271,7 +271,7 @@ describe('getScope functional test', () => {
     test('in pathname', () => {
       const pathnames = [
         ['http://foo.bar:80/foo/bar/baz', ''],
-        ['http://foo.bar:80/@scopedNoPkg', ''],
+        ['http://foo.bar:80/@scopedNoPkg', '@scopedNoPkg'],
```
BYK: I'm guessing this was intentional?

src/registries/npm-registry.js
```diff
@@ -218,21 +221,29 @@ export default class NpmRegistry extends Registry {
     }
   }
 
-  getScope(packageName: string): string {
-    return !packageName || packageName[0] !== '@' ? '' : packageName.split(/\/|%2f/)[0];
+  getScope(packageIdent: string): string {
+    // removing escaped / from package names in full paths
+    packageIdent = packageIdent.replace('%2f', '/');
```
BYK: I'm trying to say this normalization should not be done here, and actually should not be done at all. `http://foo.bar:80/@scope%2fbar/baz` is not equivalent to `http://foo.bar:80/@scope/bar/baz`. See https://stackoverflow.com/a/14658209/90297

__tests__/registries/npm-registry.js
lukeggchapman: Yeah I was on the fence about this one anyway. Semantically getScope is correctly getting the scope, but there is no package and will likely error when trying to fetch the package from the endpoint.

src/registries/npm-registry.js
lukeggchapman: Agreed, it is not ideal and smells. Hopefully I'll have some time tomorrow (I'm in Australia) to follow the pathname around and see why it has it in the first place. So far this is the only file I've looked at in the yarn codebase.

lukeggchapman:
So `2%f` was originally introduced when [support for scoped packages was added](https://github.com/lukeggchapman/yarn/commit/1acee606cf21ae725f7aa1ee8802774539f22ec2).

In my case It is being added to the package name when [initially added with `yarn add`](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/registries/npm-resolver.js#L79) and then stored in the path when stored and fetched from the lock file.

Doing a search on the repo for `NpmRegistry.escapeName` will see that it is common practice to replace the forward slash in our package names as this is how the npm registry expects our scoped package names to be formatted.

lukeggchapman:
Knowing that forward slashes in scoped package names are always escaped then I can rewrite both the isScopedPackage check and the getScope method. I'll update the PR shortly.

lukeggchapman:
@BYK Let me know what you think of this latest solution, it's a different approach to the one I was taking before.
@soda0289 latest fix should cover your concerns for registries with `/@`

BYK:
Oh wow, great detective work @lukeggchapman. I'll be honest, I'm not a fan of this latest solution but seems like the "real" solution is not very relevant to this patch and we have passing tests. So I'll merge this and then follow up with a PR in the hopes of fixing this madness.

NPM itself [says the following](https://docs.npmjs.com/misc/scope):
>A scope follows the usual rules for package names (URL-safe characters, no leading dots or underscores). 

So we should already unescape package names when working on them?

src/registries/npm-registry.js
```diff
@@ -23,6 +23,12 @@ const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 const REGEX_REGISTRY_PREFIX = /^https?:/;
 const REGEX_REGISTRY_SUFFIX = /registry\/?$/;
 
+// All scoped package names are of the format `@scope%2fpkg` from the use of NpmRegistry.escapeName
+// `(?:^|\/)` Match either the start of the string or a `/` but don't capture
+// `[^\/?]*?` Match any character that is not '/' or '?' and capture, up until the first occurance of:
+// `%2f` Match '%2f' the escaped forward slash and don't capture
+const SCOPED_PKG_REGEXP = /(?:^|\/)(@[^\/?]*?)(?=%2f)/;
```
BYK: This matches an empty scope like `@%2f`. Also, matches `@scope%2f`without the package name. Gonna merge anyways since I don't think this is a problem but I'm quite confused about the `*?` part in `(@[^\/?]*?)`. Why not just `(@[^\/?]+)` instead. What does even `*?` mean since both of them are quantifiers and you're not supposed to use two quantifiers back to back.

_Goes away, reads about `*?` making `*` not greedy._

I'm now convinced that we need `+` instead of `*?` since our aim is to match a scope and the full name of the scope. So gonna replace that and then merge it. LMK if you disagree.

lukeggchapman:
@BYK I agree that the solution is not ideal as it would be much cleaner if the escape name was done only once in one place. If you search for `request(NpmRegistry.escapeName(` it gets used quite frequently.

This sums up my knowledge on the presence of the %2f in npm package names:
https://github.com/npm/npm/issues/11738#issuecomment-190930512
The extensive discussion was a little too extensive for me today 😆

src/registries/npm-registry.js
lukeggchapman: I wasn't sure if a eg `@scope%2ffoo%2fbar` should have scope `@scope` or `@scope%2ffoo` so your knowledge is appreciated there.

BYK:
>@BYK I agree that the solution is not ideal as it would be much cleaner if the escape name was done only once in one place. If you search for request(NpmRegistry.escapeName( it gets used quite frequently.

Yup, gonna fix that later on.

>This sums up my knowledge on the presence of the %2f in npm package names:
npm/npm#11738 (comment)
>The extensive discussion was a little too extensive for me today 😆

Yeah, I've also traced back to the weird behavior of NPM registry. I think it is madness but it is the world we live in :D Sorry for all the trouble. Also fixing your PR soon. I messed it up :D

src/registries/npm-registry.js
BYK: Turns out we need `+?` so it stops matching at the first sight of `%2f`. TIL

lukeggchapman:
Hey @BYK, sorry to pester you but are you able to give me an indication of when this will be released?
With npm 5 also having [similar issues](https://github.com/npm/npm/issues/17918#issuecomment-318140270) It's a blocker for using private modules in our builds, I'll need to use my fork until it's released.

PR #4030, opened 2017-07-27
Forces workspaces to always be installed from their root

### LINKED ISSUE #4028: Running yarn add in a workspace break the lockfile
```bash
rm -rf foo

(
    mkdir -p foo/a && cd foo/a
    echo '{"name":"a", "version":"1.0.0"}' > package.json
)

(
    mkdir -p foo/b && cd foo/b
    echo '{"name":"b", "version":"1.0.0"}' > package.json
)

(
    mkdir -p foo/root && cd foo/root
    echo '{"private": true, "workspaces": ["workspace"], "dependencies": {"a":"file:../a"}}' > package.json
    echo 'workspaces-experimental true' > .yarnrc
)

(
    mkdir -p foo/root/workspace && cd foo/root/workspace
    echo '{"name":"workspace", "version":"1.0.0"}' > package.json
)

(
    cd foo/root
    yarn
    cp yarn.lock good-yarn.lock
)

(
    cd foo/root/workspace
    yarn add b@file:../../b
)

(
    cd foo/root
    diff good-yarn.lock yarn.lock
)
```

arcanis:
This PR fixes #4028.

- When installing, we will now always start installing from the root of the workspace, or the cwd if there's none (which is what defines `lockfileFolder`).

- The relative paths are now resolved when reading the json files rather than when resolving them. It allows us to factor the logic, and I also think it also solves a pair of issues where relative subdependencies were not correctly resolved when installed from non-`link:` and non-`file:` dependencies (this is only my belief based on my lecture of tbe source code, I haven't yet took the time to validate this hypothesis).

I still have to write integration tests, but the testcase linked with #4028 is fixed, and the regular testsuite is still green.


bestander:
New tests? :)

arcanis:
>  I still have to write integration tests, but the testcase linked with #4028 is fixed, and the regular testsuite is still green.

They'll come in time, before merge 😛

arcanis:
Added a test

__tests__/commands/install/integration.js
```diff
@@ -61,6 +61,25 @@ test.concurrent('installing with --ignore-optional should not install optional s
...
+      pkgJson.dependencies['b'] = 'file:../b';
+      await fs.writeFile(`${config.cwd}/workspace/package.json`, JSON.stringify(pkgJson));
+
+      const workspaceConfig = await Config.create({cwd: `${config.cwd}/workspace`}, reporter);
+
+      const reInstall = new Install({}, workspaceConfig, reporter, await Lockfile.fromDirectory(config.cwd));
+      await reInstall.init();
+
+      expect((await fs.readFile(`${config.cwd}/yarn.lock`)).indexOf(`"b@file:b"`)).not.toEqual(-1);
+      expect((await fs.readFile(`${config.cwd}/yarn.lock`)).indexOf(`"a@file:./a"`)).not.toEqual(-1);
```
BYK: Awoid reading the same file twice and store the file contents in a variable, please.

src/util/git.js
```diff
@@ -267,7 +267,7 @@ export default class Git implements GitRefResolvingInterface {
       if (await fs.exists(cwd)) {
         await spawnGit(['pull'], {cwd});
       } else {
-        await spawnGit(['clone', gitUrl.repository, cwd]);
+        await spawnGit(['clone', gitUrl.repository, cwd], {cwd: this.config.cwd});
```
BYK: Why do we need this change?

src/util/normalize-manifest/fix.js
```diff
@@ -306,4 +306,10 @@ export default (async function(
       info.license = inferredLicense;
     }
   }
+
+  for (const hint of ['dependencies', 'devDependencies', 'optionalDependencies', 'peerDependencies']) {
```
BYK: Didn't we have this in somewhere else? Somewhere more central?

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
+import {FILE_PROTOCOL_PREFIX} from '../../resolvers/exotics/file-resolver.js';
+import {LINK_PROTOCOL_PREFIX} from '../../resolvers/exotics/link-resolver.js';
+
+const path = require('path');
+
+export default function(info: Object, moduleLoc: string, lockfileFolder: string): void {
+  // It won't work if we don't yet know what's the folder we'll use as root. It's not a
+  // big deal tho, because it only happens when trying to figure out the root, and we
+  // don't need to know the dependencies / devDependencies at this time.
+  if (!lockfileFolder) {
```
BYK: `lockfileFolder` can only be an empty string since you define its type as `string`. Is this check really necessary?

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
...
+    return;
+  }
+
+  for (const hint of ['dependencies', 'devDependencies', 'optionalDependencies', 'peerDependencies']) {
```
BYK: Can you move the `['dependencies', 'devDependencies', 'optionalDependencies', 'peerDependencies']` part into a constant so we don't have to create this array ever again in anywhere in the app?

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
...
+    if (!info[hint]) {
```
BYK: `hint` is not a great name. How about `dependencyType`?

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
...
+
+      let prefix;
+
+      if (value.indexOf(FILE_PROTOCOL_PREFIX) === 0) {
+        prefix = FILE_PROTOCOL_PREFIX;
+      } else if (value.indexOf(LINK_PROTOCOL_PREFIX) === 0) {
+        prefix = LINK_PROTOCOL_PREFIX;
+      }
+
+      if (prefix) {
```
BYK: Early exit:
```
if (!prefix) {
  return;
}
```

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
...
+        const unprefixed = value.substr(prefix.length);
+        const hasPrefix = /^\.(\/|$)/.test(unprefixed);
+
+        const absoluteTarget = path.resolve(lockfileFolder, moduleLoc, unprefixed);
+        let relativeTarget = path.relative(lockfileFolder, absoluteTarget) || '.';
+
+        if (hasPrefix) {
+          // If the original value was using the "./" prefix, then we output a similar path.
+          // We need to do this because otherwise it would cause problems with already existing
```
BYK: Why not normalize everything in the lockfile so we don't need this? Also I have concerns around the regular expression above and Windows. May be we should use something else?

src/util/normalize-manifest/resolve-relative.js
```diff
@@ -0,0 +1,52 @@
...
+          // lockfile, which would see some of their entries being unrecognized.
+          relativeTarget = relativeTarget.replace(/^(?!\.{0,2}\/)/, `./`);
```
BYK: No idea what this is doing but I'm guessing it can be replaced with `path.normalize()` or something similar?

src/util/normalize-manifest/resolve-relative.js
arcanis: No, this has to be executed for each dependency. `continue` would be the right keyword, but it would behave no differently than it currently does.

src/util/normalize-manifest/resolve-relative.js
arcanis: If the dependency is `link:./foo`, we need it to be `link:./foo` after this code. If it is `link:foo`, we need it to be `link:foo`. If we don't, we change the output of the lockfile, and it might cause lockfiles to change. Normalize doesn't do the job, because we actually need to do the opposite of normalizing :)

src/util/normalize-manifest/resolve-relative.js
arcanis: This commit is to be shipped in patch version, so I'd prefer not to change the lockfile output.

src/util/normalize-manifest/resolve-relative.js
BYK: Sorry, I was going for `continue`. It makes the code easier to follow and takes a level of indent.

src/util/normalize-manifest/resolve-relative.js
BYK: Ah! Well, may be add a `todo` comment to follow up with this then?

src/util/normalize-manifest/resolve-relative.js
BYK: Ping :)

__tests__/commands/install/integration.js
BYK: Ping

src/util/normalize-manifest/resolve-relative.js
arcanis: > It makes the code easier to follow and takes a level of indent.

That's kinda subjective :)

BYK: Taking a level of indent, making the code closer to the left are both facts. I'd argue text closer to left side is easier to read text indented otherwise we'd have large margins on the left side for all text. That said I'm gonna stop here :D

Issue #3773, opened 2017-06-30
Yarn run no longer works on windows with executables not mentioned in package.json/scripts


betalb:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When running executable, that is not mentioned in scripts section of package.json using `yarn run`, yarn fails

```
yarn run v0.27.3
$ "C:\work\repos\repo\node_modules\.bin\grunt" "--version"
'C:\work\repos\repo\node_modules\.bin\grunt" "--version' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
* yarn init
* yarn add grunt
* yarn run grunt -- --version

Adding grunt to scripts resolves an issue
```
"scripts": {
    "grunt": "grunt"
}
```

**What is the expected behavior?**
Worked fine in v0.24.6
```
yarn run v0.24.6
$ "C:\work\repos\repo\node_modules\.bin\grunt" --version
grunt-cli v1.2.0
grunt v1.0.1
Done in 0.51s.
```

**Please mention your node.js, yarn and operating system version.**
Yarn version: 0.27.3
Node version: 7.10.0
Platform: win32 x64 (Windows 10)

arcanis:
Hm, we do have a [test](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/run.js#L69-L76) to check this, and it seems to be passing, even on Windows :/ Does anybody reproduce this? I can't make this happen on OSX.

bdukes:
This appears to only be an issue when there are additional arguments passed.  `yarn run gulp` works for me, but `yarn run gulp -- --verbose` or `yarn run gulp -- watch` do not

jods4:
Same here: `yarn run webpack -- --config webpack.config.vendor.js` used to work. 
Not anymore.

craigjennings11:
Ran into this when I upgraded from 0.26.1 to 0.27.0

![image](https://user-images.githubusercontent.com/1683368/27884184-f8b1f0ea-6198-11e7-8bd6-f113211e296b.png)

My hunch is that this release [comment](https://github.com/yarnpkg/yarn/releases/tag/v0.27.0) has something to do with it
> Fix arguments passed to scripts so that they are correctly escaped

Volune:
[...]

PR #4031, opened 2017-07-27
Fix `yarn run` on window

Volune:
**Summary**

Fix for #3773
> Yarn run no longer works on windows with executables not mentioned in package.json/scripts

**Test plan**

I think we don't have automated tests for this case, it's OS specific and the fix relies on the `shell` option of `child_process.spawn`, so we cannot use mock to reliably test this issue.
(Any idea welcomed)

I ran manual tests on Windows 10 and a linux VM.
We should get more manual tests from the community before merging.

### DIFF
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -166,23 +166,9 @@ export async function executeLifecycleScript(
   env[constants.ENV_PATH_KEY] = pathParts.join(path.delimiter);
 
   // get shell
-  const conf = {windowsVerbatimArguments: false};
-  let sh = 'sh';
-  let shFlag = '-c';
   if (process.platform === 'win32') {
-    // cmd or command.com
-    sh = process.env.comspec || 'cmd';
-
-    // d - Ignore registry AutoRun commands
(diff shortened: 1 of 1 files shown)


BYK:
`shell` option is added in Node 5 and we need to support Node 4 for some more. For this reason, we can't go with the current solution. That said the code you removed was trying to mimic the `shell: true` behavior as documented in Node docs.

I think the main issue is with getting a full path or something on Windows so we can look there.

Volune:
`shell` is in [Node 4 documentation](https://nodejs.org/docs/latest-v4.x/api/child_process.html#child_process_child_process_spawn_command_args_options)

I just tested with node 4.8.4 to confirm, it worked.

BYK:
Oh, wow. Apologies. I looked at the latest docs and it says "v5.7.0 | The shell option is supported now." in the history part. I was also curious why our Node 4 builds didn't fail :D

This means we may use this in one more place.

Vanuan:
Looks like it was introduced in 4.8, since it fails in 4.7
Couldn't find any information about node requirement. Looks like it's the time to upgrade to 6.x

PR #4036, opened 2017-07-27
Chore: clear no-console warnings from lint

BYK:
**Summary**

After upgrading ESlint, we started seeing a bunch of `no-console`
lint warnings. This patch fixes these or supresses them if necessary.

**Test plan**

Lint should pass without any errors or warnings. All existing tests
should pass.


src/cli/commands/licenses.js
```diff
@@ -101,65 +99,4 @@ export const {run, setFlags, examples} = buildSubCommands('licenses', {
   async list(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
     await list(config, reporter, flags, args);
   },
-
-  async generateDisclaimer(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
```
BYK: To my very surprise, this entire block of code was unused.

arcanis: It is - it's an entry inside the object that is passed in parameter to `buildSubCommands`. It is run when you call `yarn licenses generate-disclaimer`.

src/cli/commands/help.js
```diff
@@ -14,45 +14,50 @@ export function hasWrapper(flags: Object, args: Array<string>): boolean {
 
 export function setFlags(commander: Object) {}
 
+function rawText(str: string): {inspect(): string} {
+  return {
+    inspect(): string {
+      return str;
+    },
+  };
+}
```
arcanis: Isn't that a bit hacky? :<

BYK: It is but the alternative is to fix the `JSON.stringify` transformation that the reporter uses on strings, which requires a `typeof x === 'string'` check and then fixing all translatable strings to add the now-missing quotes which would invalidate all translations. If you think it's still worth it, I'm okay doing that in a follow-up since this stringification issue popped up in a few other PRs recently.

src/cli/commands/licenses.js
BYK: Well, since no tests cover this, I think I'll add a test case for this since it was very confusing to me not finding a single call to this function.

src/cli/commands/help.js
arcanis: Ok - what do you think about putting this in the BaseReporter class, so that it can become:

```js
reporter.log(reporter.lang('helpExamples', reporter.rawText(examples.join('\n'))));
```

It would become a tool whose implementation would be entirely contained into the reporter. Whether it's hacky or not would be something the commands would have to care or not.

BYK: Great idea, update coming.

src/cli/commands/licenses.js
```diff
@@ -124,15 +125,14 @@ export const {run, setFlags, examples} = buildSubCommands('licenses', {
...
-        `IN PORTIONS OF THE ${String(manifest.name).toUpperCase().replace(/-/g, ' ')} PRODUCT.`,
+        `IN PORTIONS OF THE ${String(manifest.name).toUpperCase().replace(/-/g, ' ')} PRODUCT.` +
+        '\n',
     );
-    console.log();
 
     for (const [licenseText, manifests] of manifestsByLicense) {
-      console.log('-----');
-      console.log();
+      reporter.log('-----\n');
```
bestander: I am not 100% sure but afaik @kittens had a few console.log instead of reporters for a reason.
I don't remember which.

kittens: `reporter.log` dumps a bunch of ansi escapes to clear the current line and is for abstracting the console output so it can be consumed by other tools (json output being the primary one). This command is only for text consumption and you should just be dumping it to a txt file. Using a reporter here has the potential to mess up the output since it might print ansi escapes.

Issue #3610, opened 2017-06-08
Yarn Version returns fatal


khaosdoctor:
**What is the current behavior?**
Running `yarn version` works ok until it executes `git add <thefullpathtomyrepo>`, then I get a "Fatal: <thefullpathtomyrepo>/package.json is outside repository"

Line: https://github.com/yarnpkg/yarn/blob/c1fe8a16c6c66d4c0fe7a0955b8432aa01a01587/src/cli/commands/version.js#L150

**If the current behavior is a bug, please provide the steps to reproduce.**
Checkout a repo, go into it, run `yarn version`

**What is the expected behavior?**

- Update version
- Create commit
- Commit
- Tag

**Please mention your node.js, yarn and operating system version.**
Node 8
NPM 5
Yarn 0.24
MacOSX Sierra

BYK:
I'm assuming your repo doesn't have a `package.json` file in it?

khaosdoctor:
No, it has. When I run the install command it works just fine

rally25rs:
If you manually run `git add <thefullpathtomyrepo>` then I would expect Git to output the same error message. What is `<thefullpathtomyrepo>`? Is your Git repo set up in some non-standard way?

khaosdoctor:
Let me give a more customized example.

I usually set all my repos on `~/Documents/Repositories`, so I created a new folder called `personal` inside it and then cloned one of my repos called `knoblr`, so the full path is `~/Documents/Repositories/Personal/knoblr`.

When I run `yarn version` (or `npm version`) it gives me this error. If I run `git add /Users/myuser/Documents/Repositories/Personal/knoblr/package.json` it outputs me the same error. I have to manually do `git add package.json` and things flow smoothly.

I think this is because git treats things relatively, so I think it would be better to run just `git add package.json` (or some relative path to it).

BYK:
I couldn't find any git documentation around absolute paths and `add` but I think your assessment is correct.

So the solution then would be normalizing the path according to the git root and seems like we can get that easily by running `git rev-parse --show-toplevel`.

Sounds good? //cc @arcanis @bestander

BYK:
@khaosdoctor can you let us know if you can run `git rev-parse --show-toplevel` and if #4038 fixes the problem for you"?

khaosdoctor:
I tested it here, ran `git add $(git rev-parse --show-toplevel)/package.json` and got the same error. Tried again with Yarn 0.27 and same error too.

[...]

PR #4038, opened 2017-07-27
Fix: use `git add` with relative path to git root in version command

BYK:
**Summary**

Refs #3610.

Some versions of `git` doesn't support passing absolute paths to
`git add` causing it to fail. This patch runs
`git rev-parse --show-toplevel` to git root and uses the relative
path of `package.json` to that directory when calling `git add`.

**Test plan**

Manual tests for now.

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -5,7 +5,7 @@ import type Config from '../../config.js';
 import {registryNames} from '../../registries/index.js';
 import {execCommand} from '../../util/execute-lifecycle-script.js';
 import {MessageError} from '../../errors.js';
-import {spawn} from '../../util/child.js';
+import {spawn as spawnGit} from '../../util/git/git-spawn.js';
 import * as fs from '../../util/fs.js';
 import map from '../../util/map.js';
 
@@ -118,9 +118,7 @@ export async function setVersion(
   // check if committing the new version to git is overriden
   if (!flags.gitTagVersion || !config.getOption('version-git-tag')) {
     // Don't tag the version in Git
-    return function(): Promise<void> {
-      return Promise.resolve();
-    };
+    return Promise.resolve.bind(Promise);
   }
 
   return async function(): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/cli/commands/version.js
```diff
@@ -118,9 +118,7 @@ export async function setVersion(
   // check if committing the new version to git is overriden
   if (!flags.gitTagVersion || !config.getOption('version-git-tag')) {
     // Don't tag the version in Git
-    return function(): Promise<void> {
-      return Promise.resolve();
-    };
+    return Promise.resolve.bind(Promise);
```
arcanis: If you want to use a single line statement, I'd prefer we use `() => Promise.resolve()` instead of `bind`, I find it more easily parsable by my brain and it has the added benefit that it can have a lower runtime overhead.

src/cli/commands/version.js
```diff
@@ -146,14 +144,16 @@ export async function setVersion(
       const flag = sign ? '-sm' : '-am';
       const prefix: string = String(config.getOption('version-tag-prefix'));
 
+      const gitRoot = (await spawnGit(['rev-parse', '--show-toplevel'])).trim();
+
       // add manifest
-      await spawn('git', ['add', pkgLoc]);
+      await spawnGit(['add', path.relative(gitRoot, pkgLoc)]);
```
arcanis: I think you need to use the `{cwd: gitRoot}` option, or it won't work if you're in a deep directory.

src/cli/commands/version.js
```diff
@@ -146,14 +144,16 @@ export async function setVersion(
       const flag = sign ? '-sm' : '-am';
       const prefix: string = String(config.getOption('version-tag-prefix'));
 
+      const gitRoot = (await spawnGit(['rev-parse', '--show-toplevel'])).trim();
```
arcanis: You probably want to use `{cwd: config.cwd}` so that if we add tests, it will still work.

src/cli/commands/version.js
```diff
@@ -146,14 +144,16 @@ export async function setVersion(
...
-      await spawn('git', ['add', pkgLoc]);
+      await spawnGit(['add', path.relative(gitRoot, pkgLoc)]);
 
       // create git commit
-      await spawn('git', ['commit', '-m', message]);
+      await spawnGit(['commit', '-m', message]);
 
       // create git tag
-      await spawn('git', ['tag', `${prefix}${newVersion}`, flag, message]);
+      await spawnGit(['tag', `${prefix}${newVersion}`, flag, message]);
```
arcanis: Those two commands should also have the `{cwd: gitRoot}` option.

src/cli/commands/version.js
BYK: Sure. That said using an arrow function for this turned out to be slightly slower on V8: https://jsperf.com/resolved-promise

src/cli/commands/version.js
BYK: Good catch!

PR #4045, opened 2017-07-28
`yarn <command> ls` command cleanup

tpina:
**Summary**

From https://github.com/yarnpkg/yarn/pull/3898#issuecomment-318609494

This PR is to clean up the remaining undeprecated `yarn <command> ls` command and update the usage template.

**Test plan**

Some rounds of functional manual testing to assert that the correct messages are displayed and previous commands work as per expected. Touched files do not have any associated tests.


src/cli/commands/team.js
```diff
@@ -38,14 +43,19 @@ function explodeScopeTeam(arg: string, requireTeam: boolean, reporter: Reporter)
   };
 }
 
-function warnDeprecation(reporter: Reporter) {
-  reporter.warn(`\`yarn team rm\` is deprecated. Please use \`yarn team remove\`.`);
+function warnDeprecation(reporter: Reporter, deprecationWarning: DeprecationWarning) {
+  reporter.warn(`\`yarn team ${deprecationWarning.deprecatedCommand}\` is deprecated.
```
BYK: You should be using `reporter.lang` and add a language key for these so they can be translated.

src/cli/commands/team.js
```diff
@@ -38,14 +43,19 @@ function explodeScopeTeam(arg: string, requireTeam: boolean, reporter: Reporter)
...
+  Please use \`yarn team ${deprecationWarning.currentComand}\`.`);
 }
 
-function wrapRequired(callback: CLIFunctionWithParts, requireTeam: boolean, isDeprecated?: boolean): CLIFunction {
+function wrapRequired(
+  callback: CLIFunctionWithParts,
+  requireTeam: boolean,
+  subCommandDeprecated?: DeprecationWarning,
```
BYK: I'd rename this to `deprecationMessage` or `deprecationInfo`

tpina:
@BYK  and @voxsim 

Done. Awaiting for feedback.

src/cli/commands/team.js
```diff
@@ -71,23 +81,31 @@ function wrapRequired(callback: CLIFunctionWithParts, requireTeam: boolean, isDe
...
+): CLIFunction {
+  return wrapRequired(
+    function(
+      parts: TeamParts,
+      config: Config,
+      reporter: Reporter,
+      flags: Object,
+      args: Array<string>,
+    ): CLIFunctionReturn {
+      if (args.length === 1) {
```
BYK:
```
return args.length === 1 && callback(...);
```

Also would be nice to explain why we do this only when argument count is exactly 1.

BYK:
@tpina thanks a lot! This looks good to me. Would you like to address my changes or just merge it as it is?

tpina:
@BYK  I'll address your changes/comments. Let me take another look at it.

src/cli/commands/team.js
tpina: `yarn team list` only takes one arg which is either: `<scope>` or `<scope:team>` more than one arg and the command should fail.

src/cli/commands/team.js
tpina: Will do

src/cli/commands/team.js
tpina: Ok not a problem!

tpina:
@BYK  a really annoying thing with the `reporter.lang` - it keeps putting quote on the params so the final string reads: `warning yarn "team" "ls" is deprecated. Please use `yarn "team" "list"`.

Is there any way around this?

tpina:
meh... workaround added. See if you agree once checks have completed.

src/reporters/lang/en.js
```diff
@@ -324,6 +324,8 @@ const messages = {
     'Installing Yarn via Yarn will result in you having two separate versions of Yarn installed at the same time, which is not recommended. To update Yarn please follow https://yarnpkg.com/en/docs/install .',
 
   scopeNotValid: 'The specified scope is not valid.',
+
+  yarnDeprecatedCommand: '`yarn $0 $1` is deprecated.\n Please use `yarn $0 $2`.',
```
BYK: This would repeat the `yarn` part in the message :(

I'll fix it and then merge.

tpina:
@BYK you're right. Apologies for that.

Lucas-C:
I think the current doc does not reflect the `ls` -> `list` change:
https://yarnpkg.com/en/docs/cli/owner

BYK:
@Lucas-C thanks for the heads up! Filed https://github.com/yarnpkg/website/issues/660

Issue #886, opened 2016-10-12
fails if bundledDependency is missing in .tgz file


raytiley:
Running yarn on `0.15.1` on OS X and having issues installing an ember-cli project.

![ui -zsh 97x25 2016-10-12 10-06-36](https://cloud.githubusercontent.com/assets/41680/19313205/abfac170-9063-11e6-8788-157b8961d58d.png)

package.json

```
{
  "name": "carousel-ui",
  "version": "0.0.1",
  "description": "User interface for Carousel API",
  "private": true,
  "directories": {
    "doc": "doc",
    "test": "tests"
  },
  "scripts": {
    "start": "ember server",
    "build": "ember build",
    "test": "ember test"
  },
  "repository": "",
  "engines": {
    "node": ">= 0.10.0"
  },
  "author": "",
  "license": "MIT",
  "devDependencies": {
    "broccoli-asset-rev": "^2.4.2",
    "ember-ajax": "^2.0.1",
    "ember-cli": "2.7.0",
    "ember-cli-app-version": "^1.0.0",
    "ember-cli-autoprefixer": "0.6.0",
    "ember-cli-babel": "^5.1.6",
    "ember-cli-bootstrap-sassy": "0.0.18",
    "ember-cli-dependency-checker": "^1.3.0",
    "ember-cli-deploy": "0.6.4",
    "ember-cli-deploy-build": "0.1.1",
    "ember-cli-deploy-gzip": "0.2.3",
    "ember-cli-deploy-s3": "0.3.1",
    "ember-cli-deploy-slack": "0.1.0",
    "ember-cli-deploy-webhooks": "0.3.0",
    "ember-cli-flash": "1.3.16",
    "ember-cli-htmlbars": "^1.0.3",
    "ember-cli-htmlbars-inline-precompile": "^0.3.5",
    "ember-cli-ic-ajax": "0.2.5",
    "ember-cli-inject-live-reload": "^1.4.0",
    "ember-cli-jshint": "^1.0.0",
    "ember-cli-mirage": "0.2.1",
    "ember-cli-moment-shim": "2.0.0",
    "ember-cli-nouislider": "0.9.1",
    "ember-cli-pagination": "2.2.2",
    "ember-cli-qunit": "^3.0.0",
    "ember-cli-release": "^0.2.9",
    "ember-cli-rollbar": "0.3.0",
    "ember-cli-sass": "5.5.1",
    "ember-cli-selectize": "0.5.5",
[...]

PR #4046, opened 2017-07-28
Warn for missing bundledDependencies

rally25rs:
**Summary**

For #886 

* Display a WARNING during install if a dependency is missing a `bundledDependency` (this indicates a problem with that published package)
* No longer error when attempting to create .bin links for `bundledDependency`s that are missing.

**Test plan**

Added a test in `__tests__/commands/install/integration` that installs `tap@0.3.1` which is a broken package according to the original issue #886


src/package-linker.js
```diff
@@ -99,11 +99,17 @@ export default class PackageLinker {
...
 
-        const dep = await this.config.readManifest(loc, remote.registry);
-
-        if (dep.bin && Object.keys(dep.bin).length) {
-          deps.push({dep, loc});
+            if (dep.bin && Object.keys(dep.bin).length) {
+              deps.push({dep, loc});
+            }
+          } catch (ex) {
+            // intentionally ignoring error.
```
arcanis: We should probably check that the exception is of a specific type (ENOENT). We still want to throw if we can't read it for any other reason, instead of silencing the error (since it won't get displayed by the missing bundled dependency check either).

arcanis: Also, I'm not sure you need to run `await fs.exists` here, since you will get the same error if `loc` doesn't exists.

rally25rs:
@arcanis Pushed requested changes. Not sure why CI tests are failing. Seems unrelated to my changes. All tests pass locally.

rally25rs:
Oh thanks, I was about to resolve the conflict myself...

arcanis:
Thanks 👍

bestander:
Awesome!

src/package-linker.js
```diff
@@ -380,6 +387,10 @@ export default class PackageLinker {
         linkBinConcurrency,
       );
     }
+
+    for (const [, {pkg}] of flatTree) {
+      await this._warnForMissingBundledDependencies(pkg);
```
BYK: Why do we report these later on, instead of as we encounter above around line 112?

BYK: I also don't like the fact that we need to check for file/directory existence later on when we already know they are missing above.

rally25rs: I made a separate `_warnForMissingBundledDependencies` function primarily as a Single-Responsibility Principle / Separation of concerns nod. `linkSelfDependencies` does not sound like a function that would check bundled dependencies.

If I thought it would make a big performance difference I probably would have combined them, but I can't imagine that many packages actually have bundled deps, so I'm figuring the vast majority of installs won't see any noticeable performance difference.

I'm open to refactoring it though, if you want.

BYK: >linkBinDependencies does not sound like a function that would check bundled dependencies.

Not sure if I agree since it already checks them via that `try`-`catch` block. I don't see the value of indirection here.

>If I thought it would make a big performance difference I probably would have combined them, but I can't imagine that many packages actually have bundled deps, so I'm figuring the vast majority of installs won't see any noticeable performance difference.

I'm a bit concerned about this but I think your reasoning is sound. Still, if there was an "obvious", faster way, I'd prefer that. As my comment above, if these were completely separate things, I'd be okay but looks like we just defer the work we could have done earlier without a very clear separation and paying the cost of separation. Or I'm missing something here.

Issue #3920, opened 2017-07-12
yarn doesn't use .npmrc from $HOME when run as root


also:
**Do you want to request a *feature* or report a *bug*?**

Bug; incompatibility with npm.

**What is the current behavior?**

When run as  the `root` user on Linux, yarn does not use `$HOME/.npmrc` but uses `/usr/local/share/.npmrc` instead.

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/yarnpkg/yarn/blob/v0.28.2/src/util/user-home-dir.js#L8
https://github.com/yarnpkg/yarn/blob/v0.28.2/src/registries/npm-registry.js#L132

Run `yarn config list --verbose` as root on Linux (while not in any subdirectory of`$HOME`) and observe that yarn looks for `.npmrc` first in the current directory, then in `/usr/local/share/`, but never in `$HOME` (`/root`).

**What is the expected behavior?**

Yarn defaults to `$HOME/.npmrc` as [documented by npm](https://docs.npmjs.com/files/npmrc#per-user-config-file).

A module named `user-home-dir` doesn't export a string which isn't the user's home directory.

**Please mention your node.js, yarn and operating system version.**

yarn v0.27.5
node v8.1.3

dylanfoster:
It looks it doesn't use `./.npmrc` either. I had to downgrade to get it to work

vinicius73:
Same problem here

![image](https://user-images.githubusercontent.com/1561347/28141208-f9015fec-6731-11e7-8e62-90d13d275143.png)

BYK:
Thanks for reporting. Anyone willing to submit a PR for this?

BYK:
@kaylieEB still working on this?

kaylieEB:
hey @BYK sorry for the delay, here it is :) https://github.com/yarnpkg/yarn/pull/4047

BYK:
@kaylieEB no worries, just wanted to make sure to avoid duplicate work :)

Jessevanbekkum:
We also had this problem, and we were able to solve it by placing the `.npmrc` in the directory next to the git repository (one directory up from the source dir). Then yarn was able to pick up the `.npmrc`. Not a pretty solution, but it worked

BYK:
Oh, looks like we forgot to close this. This is fixed on master.

PR #4047, opened 2017-07-29
Check for npmrc in home directory for linux root users

kaylieEB:
**Summary**
Fix for https://github.com/yarnpkg/yarn/issues/3920.

In this [commit](https://github.com/wtgtybhertgeghgtwtg/yarn/commit/f4fe7431ed992862a1c1d2357e6521c268bad7a7), `userHome` for linux users running as root was changed to `/usr/local/share`, mainly to allow for other users to run globally added bins. However this introduced a bug where npmrc wasn't being looked in the root directory after checking `/user/local/share`. So this change pushes another location to check in case `userHome` is different from native home directory returned by `os.homedir()`

**Test plan**
I've added a test to generally test the `getPossibleConfigLocations` method, but I haven't been able to properly mock out running it as a root user on linux. Suggestions welcome!

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -50,12 +50,17 @@ function createMocks(): Object {
       getScopedOption: jest.fn(),
     },
   };
-  const mockReporter = jest.fn();
+  const mockLog = [];
+  const mockReporter = {
+    lang: (...args) => mockLog.push(...args),
+    verbose: jest.fn(),
+  };
 
   return {
     mockRequestManager,
     mockRegistries,
     mockReporter,
+    mockLog,
   };
 }
 
@@ -290,5 +295,18 @@ describe('getScope functional test', () => {
     packageIdents.forEach(([pathname, scope]) => {
       expect(npmRegistry.getScope(pathname)).toEqual(scope);
     });
+  });
+});
+
+describe('getPossibleConfigLocations', () => {
+  describe('searches recursively to home directory', async () => {
+    const testCwd = './project/subdirectory';
+    const {mockRequestManager, mockRegistries, mockReporter, mockLog} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+    await npmRegistry.getPossibleConfigLocations('npmrc', mockReporter);
+
+    expect(mockLog.indexOf('project/subdirectory/.npmrc')).toBeGreaterThan(-1);
+    expect(mockLog.indexOf('project/.npmrc')).toBeGreaterThan(-1);
+    expect(mockLog.indexOf(`${homeDir}/.npmrc`)).toBeGreaterThan(-1);
   });
 });
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -304,9 +304,9 @@ describe('getPossibleConfigLocations', () => {
     const logs = reporter.getBuffer().map(logItem => logItem.data);
     expect(logs).toEqual(
       expect.arrayContaining([
-        expect.stringContaining(pathJoin('project', 'subdirectory', '.npmrc')),
-        expect.stringContaining(pathJoin('project', '.npmrc')),
-        expect.stringContaining(pathJoin(homeDir, '.npmrc')),
+        expect.stringContaining(JSON.stringify(pathJoin('project', 'subdirectory', '.npmrc'))),
+        expect.stringContaining(JSON.stringify(pathJoin('project', '.npmrc'))),
+        expect.stringContaining(JSON.stringify(pathJoin(homeDir, '.npmrc'))),
       ]),
     );
   });
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -299,7 +299,7 @@ describe('getScope functional test', () => {
 });
 
 describe('getPossibleConfigLocations', () => {
-  describe('searches recursively to home directory', async () => {
+  test('searches recursively to home directory', async () => {
     const testCwd = './project/subdirectory';
     const {mockRequestManager, mockRegistries, mockReporter, mockLog} = createMocks();
     const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import {resolve} from 'path';
+import {resolve, join as pathJoin} from 'path';
 
 import NpmRegistry from '../../src/registries/npm-registry.js';
 import {BufferReporter} from '../../src/reporters/index.js';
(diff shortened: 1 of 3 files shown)


src/util/user-home-dir.js
```diff
@@ -4,7 +4,8 @@ import ROOT_USER from './root-user.js';
 
 const path = require('path');
 
-const userHomeDir =
-  process.platform === 'linux' && ROOT_USER ? path.resolve('/usr/local/share') : require('os').homedir();
+export const home = require('os').homedir();
+
+const userHomeDir = process.platform === 'linux' && ROOT_USER ? path.resolve('/usr/local/share') : home;
```
BYK: Why limit to `linux` instead of blacklisting Windows only?

kaylieEB: Hmm, I'm not sure  I fully understand the question. Would you mind explaining a bit? Windows is already filtered out in the ROOT_USER  here https://github.com/yarnpkg/yarn/blob/7a053e2ca07d19b2e2eebeeb0c27edaacfd67904/src/util/root-user.js#L4

Daniel15: What about other operating systems like BSD? I wonder if we should use this root logic for those, too.

src/registries/npm-registry.js
```diff
@@ -14,7 +14,7 @@ import {addSuffix} from '../util/misc';
 import {getPosixPath, resolveWithHome} from '../util/path';
 
 const normalizeUrl = require('normalize-url');
-const userHome = require('../util/user-home-dir').default;
+const {default: userHome, home} = require('../util/user-home-dir');
```
Daniel15: While you're here, could you please change this to an `import` statement? I've been meaning to revisit one of my old pull requests where I updated all `require` statements to use `import`, but just haven't had time.

src/util/user-home-dir.js
kaylieEB: It seems safe to add `'freebsd'` to the list, are there any other platforms we should add from here? https://nodejs.org/api/process.html#process_process_platform

src/registries/npm-registry.js
kaylieEB: sure thing!

src/util/user-home-dir.js
kaylieEB: Ah actually @BYK  I see what you're saying. Yeah I don't think the platform check is even necessary given that line of code in root-user.js

src/registries/npm-registry.js
```diff
@@ -12,12 +12,11 @@ import envReplace from '../util/env-replace.js';
...
 import {getPosixPath, resolveWithHome} from '../util/path';
-
-const normalizeUrl = require('normalize-url');
-const {default: userHome, home} = require('../util/user-home-dir');
-const path = require('path');
-const url = require('url');
-const ini = require('ini');
+import normalizeUrl from 'normalize-url';
+import {default as userHome, home} from '../util/user-home-dir';
+import path from 'path';
```
BYK: Nice! I think the convention is to list stdlib imports at the top, right?

__tests__/registries/npm-registry.js
```diff
@@ -50,12 +50,17 @@ function createMocks(): Object {
       getScopedOption: jest.fn(),
     },
   };
-  const mockReporter = jest.fn();
+  const mockLog = [];
+  const mockReporter = {
+    lang: (...args) => mockLog.push(...args),
```
BYK: I think we can also use the `BufferReporter` for tests?

__tests__/registries/npm-registry.js
```diff
@@ -292,3 +297,16 @@ describe('getScope functional test', () => {
     });
   });
 });
+
+describe('getPossibleConfigLocations', () => {
+  describe('searches recursively to home directory', async () => {
```
BYK: This should be a `test` or `it` call not `describe`.

src/registries/npm-registry.js
```diff
@@ -178,6 +177,12 @@ export default class NpmRegistry extends Registry {
       [false, path.join(getGlobalPrefix(), 'etc', filename)],
     ];
 
+    // When home directory for global install is different from where $HOME/npmrc is stored,
+    // E.g. /usr/local/share vs /root on linux machines, check the additional location
+    if (home !== userHome) {
+      possibles.push([true, path.join(home, localfile)]);
```
BYK: I think we should change this data structure to be a normal object instead of an array sometime in the future: `{isHomeDir: true, path: path.join(home, localfile)}`

kaylieEB:
@BYK thanks fro being awesome and making those changes!

felixfbecker:
Any ETA for a release containing this fix?

Daniel15:
@felixfbecker   You can use a nightly build if you like. https://nightly.yarnpkg.com/

felixfbecker:
Tried to, but unfortunately I need to add it to an Alpine Docker image and running the install script just results in 
```
Step 8/16 : RUN curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --nightly
 ---> Running in dc6404f17920
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6746    0  6746    0     0  33679      0 --:--:-- --:--:-- --:--:-- 54845
Installing Yarn!
> Downloading tarball...

[1/2]: https://nightly.yarnpkg.com/latest.tar.gz --> /tmp/yarn.tar.gz.XXXXLPmEmH
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  821k  100  821k    0     0  1764k      0 --:--:-- --:--:-- --:--:-- 1764k

[2/2]: https://nightly.yarnpkg.com/latest.tar.gz.asc --> /tmp/yarn.tar.gz.XXXXLPmEmH.asc
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   900  100   900    0     0   7945      0 --:--:-- --:--:-- --:--:--  7945
> WARNING: GPG is not installed, integrity can not be verified!
> Extracting to ~/.yarn...
tar: unrecognized option: strip
BusyBox v1.24.2 (2016-03-29 19:24:15 GMT) multi-call binary.

Usage: tar -[cxtZzJjahmvO] [-X FILE] [-T FILE] [-f TARFILE] [-C DIR] [FILE]...

Create, extract, or list files from a tar file

Operation:
        c       Create
        x       Extract
        t       List
        f       Name of TARFILE ('-' for stdin/out)
        C       Change to DIR before operation
        v       Verbose
        Z       (De)compress using compress
        z       (De)compress using gzip
        J       (De)compress using xz
        j       (De)compress using bzip2
        a       (De)compress using lzma
        O       Extract to stdout
        h       Follow symlinks
        m       Don't restore mtime
        exclude File to exclude
        X       File with names to exclude
        T       File with names to include
The command '/bin/sh -c curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --nightly' returned a non-zero code: 1
```
Really don't want to change our whole image just for a hotfix. I guess the alternative is switching to npm 5

BYK:
@felixfbecker that's unfortunate :( I'll file an issue to see why the script is not working on Alpine. Guessing some mismatch between GNU `tar` and others.

In the meantime, you can use the single-file bundle from https://nightly.yarnpkg.com/yarn-1.0.0-20170825.2124.js (just `curl` this wherever and use it directly run it.

Daniel15:
> In the meantime, you can use the single-file bundle from nightly.yarnpkg.com/yarn-1.0.0-20170825.2124.js

https://nightly.yarnpkg.com/latest.js will also give you the latest version.

You could also manually grab the tarball from https://nightly.yarnpkg.com/latest.tar.gz, check the GPG signature, and extract it. That's what the install script does :)

PR #4063, opened 2017-08-01
Fix: Fix single-file builds and simplify bundles

### LINKED ISSUE #4057: Single bundle builds are broken
Currently, our single-file, standalone builds are broken. See https://github.com/yarnpkg/yarn/pull/3955#issuecomment-318846631

BYK:
**Summary**

Fixes #4057.

**Test plan**

CI should pass, especially with `build-dist` job.

Manual:

 - Run `yarn build-dist`
 - Run `node artifacts/yarn-0.28.0.js --version`
 - Run `./dist/bin/yarn --version`

Make sure all above commands work without errors.


bin/yarn.js
```diff
@@ -4,41 +4,24 @@
...
-ver = ver.split('-')[0]; // explode and truncate tag from version #511
+var majorVer = parseInt(ver.split('.')[0], 10);
 
 var dirPath = null;
 
-if (semver.satisfies(ver, '>=5.0.0')) {
-  dirPath = '../lib';
-} else if (semver.satisfies(ver, '>=4.0.0')) {
-  dirPath = '../lib-legacy';
+if (majorVer >= 5) {
```
BYK: We should not have any dependencies in this file anymore, hence all the changes. Also these new ones are simpler :)

scripts/build-dist.sh
```diff
@@ -16,25 +17,33 @@ case "$(uname -s)" in
     ;;
 esac
 
+version=`node -e "console.log(JSON.parse(fs.readFileSync('package.json')).version)"`
```
BYK: We could use `jq` instead of this but we are guaranteed to have `node` so why not take advantage of that.

bin/yarn.js
```diff
@@ -4,41 +4,31 @@
...
 }
 
-// load v8-compile-cache
-if (semver.satisfies(ver, '>=5.7.0')) {
-  require('v8-compile-cache');
+if (dirPath) {
+  var v8CompileCachePath = dirPath + 'v8-compile-cache';
+  var fs = require('fs');
+  // We don't have/need this on legacy builds and dev builds
+  fs.existsSync(v8CompileCachePath) && require(dirPath + 'v8-compile-cache');
```
bestander: It won't crash on Node 4?

BYK: Nope: 
>Requiring v8-compile-cache in Node <5.7.0 is a noop – but you need at least Node 4.0.0 to support the ES2015 syntax used by v8-compile-cache.

bin/yarn.js
```diff
@@ -4,41 +4,31 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
 'use strict';
 
-// validate that used node version is supported
-var semver = require('semver');
 var ver = process.versions.node;
-ver = ver.split('-')[0]; // explode and truncate tag from version #511
+var majorVer = parseInt(ver.split('.')[0], 10);
```
Daniel15: :+1:

bin/yarn.js
```diff
@@ -4,41 +4,31 @@
...
+  dirPath = '../lib/';
+} else if (majorVer >= 4) {
+  dirPath = '../lib-legacy/';
 } else {
-  console.log(require('chalk').red('Node version ' + ver + ' is not supported, please use Node.js 4.0 or higher.'));
-  // We need to skip the rest and terminate here since the Node version is not supported.
-  // `console.log` calls are sync so we don't have a risk of exiting early before stdout flushes
-  process.exit(1); // eslint-disable-line no-process-exit
+  console.error('Node version ' + ver + ' is not supported, please use Node.js 4.0 or higher.');
+  process.exitCode = 1;
```
Daniel15: This seems like a legit place to call `process.exit`

Don't you need a `return` here? Otherwise it'll keep running even after hitting an incompatible Node.js version.

bin/yarn.js
Daniel15: Would prefer an `if` statement rather than `&&`, for clarity. Also I assume you meant to reuse the `v8CompileCachePath` variable 😛 
```
if (fs.existsSync(v8CompileCachePath)) {
  require(v8CompileCachePath)
}
```

bin/yarn.js
```diff
@@ -4,41 +4,31 @@
...
+  var v8CompileCachePath = dirPath + 'v8-compile-cache';
+  var fs = require('fs');
+  // We don't have/need this on legacy builds and dev builds
+  fs.existsSync(v8CompileCachePath) && require(dirPath + 'v8-compile-cache');
+
+  // Just requiring this package will trigger a yarn run since the
+  // `require.main === module` check inside `cli/index.js` will always
+  // be truthy when built with webpack :(
+  var cli = require(dirPath + 'cli');
+  if (!cli.webpackPolyfill) {
```
Daniel15: Is this relying on internal implementation details of Webpack? Would it break if we upgrade Webpack to a newer version?

Maybe we could export something more obvious like `IS_BUNDLE` from the Webpack build.

circle.yml
```diff
@@ -29,13 +29,11 @@ test:
         ./scripts/set-dev-version.js
       fi;
     - yarn lint
+    # Limit maxWorkers to 3 to avoid OOM on CircleCI
     - yarn test-ci -- -- --maxWorkers 3
     - yarn check-lockfile
     - yarn run build-dist
     - yarn run build-deb
-    # Test that the standalone .js build works as expected
-    - ./artifacts/yarn-`./dist/bin/yarn --version`.js --version
```
Daniel15: Shouldn't we still include this? We should verify that the output is what we expect, though.

scripts/build-dist.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
-set -ex
+set -Exu
```
Daniel15: What does capital E do? I don't see it on http://tldp.org/LDP/abs/html/options.html nor https://ss64.com/bash/set.html

scripts/build-dist.sh
```diff
@@ -16,25 +17,33 @@ case "$(uname -s)" in
     ;;
 esac
 
+version=`node -e "var fs=require('fs');console.log(JSON.parse(fs.readFileSync('package.json')).version)"`
```
Daniel15: This can be simplified:
```
version=`node -p -e "require('./package.json').version"`
```

This is what we do in `release-branch.sh` for example: https://github.com/yarnpkg/yarn/blob/9347540b09cc71d71f8651b2423ff8bbe3eeaf50/scripts/release-branch.sh#L6-L6

scripts/build-dist.sh
```diff
@@ -16,25 +17,33 @@ case "$(uname -s)" in
...
 mkdir artifacts
-mkdir dist{,/bin,/lib}
+mkdir dist{,/bin,/lib,/lib-legacy}
 
 # Workaround for https://github.com/yarnpkg/yarn/issues/2591
 eval $system_yarn run build
 eval $system_yarn run build-bundle
 chmod +x artifacts/*.js
+# Verify that it works as expected
+[[ "$version" == "$(node artifacts/yarn-$version.js --version)" ]] || exit 1
```
Daniel15: Oh, okay, doing it here is a good idea.

scripts/build-dist.sh
```diff
@@ -16,25 +17,33 @@ case "$(uname -s)" in
...
+[[ "$version" == "$(node artifacts/yarn-legacy-$version.js --version)" ]] || exit 1
 
 cp package.json dist/
 cp README.md dist/
 cp LICENSE dist/
-cp artifacts/yarn-legacy-*.js dist/lib/yarn-cli.js
-cp bin/yarn-bundle-entry.js dist/bin/yarn.js
-cp bin/{yarn,yarnpkg,*.cmd} dist/bin/
+cp artifacts/yarn-legacy-$version.js dist/lib-legacy/cli.js
+cp artifacts/yarn-$version.js dist/lib/cli.js
```
Daniel15: Hmm, including both the legacy **and** the non-legacy version is going to double the on-disk footprint for Yarn. The Node Docker people aren't going to be happy about that as it'll add an extra 3 MB to their image, which is significant given the Alpine image is only ~20 MB or so 😃

Can we **just** bundle the legacy one, like before? I don't really see any major advantages to building and bundling both.

src/cli/index.js
```diff
@@ -371,3 +371,17 @@ export default function main({
...
+  // ignore all arguments after a --
+  const doubleDashIndex = process.argv.findIndex(element => element === '--');
+  const startArgs = process.argv.slice(0, 2);
+  const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
+  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
+
+  main({startArgs, args, endArgs});
+}
+
+if (require.main === module) {
```
Daniel15: What is this doing?

bin/yarn.js
BYK: `process.exit()` is deprecated for all uses. We can use a `return` here but it is essentially covered by the `if` statement below.

bin/yarn.js
BYK: I think it can be. I'm not super happy with the approach but I didn't want to add more code/variables etc. either. We can improve this in a follow-up.

scripts/build-dist.sh
BYK: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#The-Set-Builtin
>If set, any trap on ERR is inherited by shell functions, command substitutions, and commands executed in a subshell environment. The ERR trap is normally not inherited in such cases.

I'm not sure if it makes a huge difference but looked safer to me since it covers more areas.

scripts/build-dist.sh
BYK: Oh, I didn't realize we used only one. What you say makes sense, I'll update this part.

src/cli/index.js
BYK: This runs yarn when you do `node lib/cli`. This part of the code is essentially what makes the single-file builds directly runnable since this file is the entry point.

PR #4068, opened 2017-08-01
Fix: Don't run pre/post hooks if they are not scripts

### LINKED ISSUE #3994: Yarn build script ignored whenever prebuild is installed
bug / request clarification

When `prebuild` is installed, `yarn build` always runs prebuild regardless of whether there's a `build` script already defined.

```
yarn init
yarn add prebuild
```

then add the following block to `package.json`:

```
  "scripts": {
    "build": "echo foo"
  }
```

then run `yarn run build`.

I expect to see `foo`. Instead `prebuild` is run.

Would like clarification here, as [Yarn documentation](https://yarnpkg.com/lang/en/docs/cli/run/) doesn't mention that npm packages can override scripts, nor that there are default behaviors. [NPM docs](https://docs.npmjs.com/misc/scripts) have more extensive discussion, but again do not suggest this behavior.

yarn 0.27.5, Mac OS Sierra 10.12.5, node v8.1.3

This bug is not present on yarn 0.24.6. I have not tested any other versions.

BYK:
**Summary**

Fixes #3994. We should not run installed binaries if
they correspond to `pre/post` script names. Similarly,
`pre/post` scripts should not be run if a binary is run.

**Test plan**

Added 2 new tests.


BYK:
>Seems ok, CI is red

Looks like I hit a nerve there 😛

Will fix the CI.

PR #4074, opened 2017-08-02
chore: enable flow languge integration in VScode

BurtHarris:
**Summary**

These two settings are needed to use the flow language add-in for Visual Studio Code IDE.   Without this, VScode displays flow features in JavaScript code with red squiggles.  Configuring this in the project directory, rather than a per-user setting, allows the right version of flow to be selected on a per-package.json basis.

https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode has add-in info, including docs on settings.

**Test plan**

This should only impact visual studio code users.   Test can be accomplished by opening a few flow-enabled source files and verifying no red squiggles appear.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+// These settings enable flow language feature integration in the vscode IDE, using the version of flow
+// loaded based on package.json
+{
+  "flow.useNPMPackagedFlow": true,
+  "javascript.validate.enable": false
+}


BYK:
Thanks for submitting this!

I'm usually against checking in editor-specific files into the codebase. That said I'd like the opinion of @yarnpkg/core before accepting or rejecting.

bestander:
Yeah, it would open the door to all editor configs and will increase maintenance burden.
I would suggest finding an alternative

BurtHarris:
I understand your reluctance.   But note that there are no "personal" preferences in there.     

Unfortunately because enabling flow checking requires disabling the built-in java checking, I'm very reluctant to do this in my per-user configuration.      

I guess I might make the change in my own repository's master, but try not to include it in any PR's I send your way, but I'm not sure how that sort of thing works with __git__.   

(I spent the last 17 years at Microsoft, we used a version of Perforce, where a per-user change like that would have worked out OK.)    Any advice from git wizards?

bestander:
Just add .vscode folder to your global .gitginore https://help.github.com/articles/ignoring-files/

Daniel15:
I don't think this is too bad, given that it's project configuration rather than per-user configuration. IntelliJ also suggest committing the `.idea` directory to source control: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

bestander:
And we'll end up having a folder for every possible IDE.

bestander:
The problem is - we allow one, we'll have to manage discussions "why don't you add another"

BurtHarris:
I'm fine with a .gitignore approach, thanks.

BurtHarris:
@BYK thanks for making me feel welcome.

PR #4075, opened 2017-08-02
Convert gulpfile 'var's to 'const's

onurtemizkan:
Converted two constants to `const` inside `gulpfile.js`


Daniel15:
Nice profile picture, @onurtemizkan! 😛

PR #4078, opened 2017-08-02
Adding additional test for pull request #3945

aracarie:
**Summary**
Adding an additional test for an uncovered method from pull request #3945. The previous tests did not cover this, since the method, `getExactVersionMatch`, is only called if a package's `fresh` prop is false or the frozen flag is set. The fixtures directory for this test contains a lockfile to meet this condition.

If the `manifest && getExoticResolver(version)` condition does not evaluate properly the `multiplePackagesCantUnpackInSameDestination` warning is logged. This test checks for that.


__tests__/commands/install/integration.js
```diff
@@ -491,6 +491,19 @@ test.concurrent('install file: dedupe dependencies 2', (): Promise<void> => {
   });
 });
 
+// When local packages are installed from a repo with a lockfile, the multiple packages
+// unpacking in the same location warning should not occur
+test.concurrent('install file: dedupe dependencies 3', (): Promise<void> => {
+  return runInstall({}, 'install-file-dedupe-dependencies-3', (config, reporter, install, getStdout) => {
+    const stdout = getStdout();
+    // Need to check if message is logged, but don't need to check for any specific parameters
+    // so splitting on undefined and testing if all message parts are in stdout
```
BYK: I feel like we can simply use `expect(getStdout()).toMatchSnapshot()` here with a comment similar to what you already put there. Thoughts?

BYK:
Waiting for you to reply my comment before merging, in case you wanna change this to a snapshot.

__tests__/commands/install/integration.js
aracarie: So the only downside I could see for that, is the test would then expand to cover all the warning, info, and success messages logged when an install is done. In the situation when one of those changes, it might be confusing for a test for file protocol installs to fail? What do you think?

BYK: Fair :)

PR #4080, opened 2017-08-03
Fix yarn workspace command and add test

### LINKED ISSUE #4021: Yarn workspace commands running in wrong location
Bug

Running `yarn workspace my-package upgrade classnames` changes the `package.json` at the top-level instead of the one at `./packages/my-package/package.json`

Yarn v0.27.5
Node v7.1.0

thejameskyle:
Resolves #4021

This fixes the workspace command so that it runs in the correct directory and adds a test to validate it.

Previously the command was also relying on `process.argv` to get the executable path for `node` and `yarn`. However, that ends up being unreliable if yarn was run as part of another process (aka `jest`).

`process.execPath` will give you the correct path to `node` and I'm assuming `require.resolve('../bin/yarn.js')` will work for getting the correct path to the `yarn` executable (from the `constants.js` file). I have a feeling it might not work perfectly though... @Daniel15 might know?

cc @arcanis


Daniel15:
I haven't looked at this PR in detail yet, but bin/yarn.js won't work when using the single-JS-file build (which I think is what we use internally at Facebook for example). The single file build only has the yarn-vX.X.X.js file. 

_Sent from my Samsung SAMSUNG-SM-G925A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

thejameskyle:
@Daniel15 looked into it, turns out webpack avoids that issue by replacing `require.resolve` with the module id:

```js
YARN_BIN_PATH = /*require.resolve*/(341);
```

thejameskyle:
Okay I am now handling the case where we are inside of a bundle separately so that it doesn't look for the `../bin/yarn.js` file.

src/cli/commands/workspace.js
```diff
@@ -39,8 +40,13 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     throw new MessageError(reporter.lang('workspaceUnknownWorkspace', workspaceName));
   }
 
+  const workspace = workspaces[workspaceName];
+
   try {
-    await child.spawn(process.argv[0], [process.argv[1], ...rest], {stdio: 'inherit'});
+    await child.spawn(NODE_BIN_PATH, [YARN_BIN_PATH, ...rest], {
```
BYK: May be we should use [child_process.fork()](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options) for this instead?

scripts/build-webpack.js
```diff
@@ -38,6 +38,10 @@ const compiler = webpack({
     libraryTarget: 'commonjs2',
   },
   target: 'node',
+  node: {
```
BYK: I think this needs a comment explaining why.

src/constants.js
```diff
@@ -59,6 +60,17 @@ export const CONFIG_DIRECTORY = getDirectory('config');
...
 export const GLOBAL_MODULE_DIRECTORY = path.join(CONFIG_DIRECTORY, 'global');
 
+export const NODE_BIN_PATH = process.execPath;
+export const YARN_BIN_PATH = getYarnBinPath();
+
+function getYarnBinPath(): string {
+  if (isWebpackBundle) {
+    return __filename;
+  } else {
+    return path.join(__dirname, '..', 'bin', 'yarn.js');
```
BYK: We shouldn't rely on this path since it can change drastically as you already mentioned in your comments.

arcanis: Hm, if it changes, it will get caught by the tests, and we will simply have to update it.

src/constants.js
```diff
@@ -2,6 +2,7 @@
 
 const path = require('path');
 const userHome = require('./util/user-home-dir').default;
+const isWebpackBundle = require('is-webpack-bundle');
```
arcanis: I'm not sure we should use an external module for this - inlining it would allow webpack to remove the unneeded code at compile time.

arcanis:
@BYK Of note, that's a different issue than #4030 - the problem here is that `yarn workspace <workspace>
 <command>` will run in the current folder instead of the workspace folder. The important bit is the added `cwd` option on the `spawn` call.

src/constants.js
Daniel15: >  inlining it would allow webpack to remove the unneeded code at compile time.

That's a good point. Is Webpack smart enough to eliminate the unneeded branch at compile time?

BYK:
@thejameskyle  - I'd really love to move forward with this. Please let us know what you think about our comments.

src/constants.js
arcanis: I think so, but I'm not completely sure now that you mention it 🤔

thejameskyle: Is it that important to remove one line of code?

src/constants.js
thejameskyle: There's not a lot of option otherwise, there's no reliable way of automatically getting the entry point to yarn. You have to manually handle each case

src/cli/commands/workspace.js
thejameskyle: I think in the future it's probably worthwhile not to use a child process at all, would save a whole bunch of overhead

thejameskyle:
Updated

PR #4086, opened 2017-08-03
Prevents hoisting through peer dependencies

### LINKED ISSUE #3933: Improper hoisting of peerDependencies: webpack, ajv, ajv-keywords
bug

1. (Top-level) `foo` lists `dependencies`:
    - `bar@5`
    - `baz`
1. `baz` lists `peerDependencies` `bar@5`.
1. (Top-level) `qux` & `quux` list `dependencies` `bar@4`
    - Causes `bar@4` to hoist, invalidating `baz`'s peerDependency of `bar@5`)

The result is the following tree:

```
├─foo
│   └─bar@5
├─baz // (Why is this hoisted? Causes invalid bar peerDependency)
├─bar@4 // (due to qux & quux)
├─qux
└─quux
```

```
// In empty directory
$ yarn add  babel-cli@6.24.1 stylelint@7.9.0 webpack@3.1
$ yarn check
yarn check v0.28.1
[...]
error "webpack#ajv-keywords#ajv@>=5.0.0" doesn't satisfy found match of "ajv@4.11.8"
[...]
error Found 1 errors.
```

Don't hoist dependencies (i.e., `baz` & `ajv-keywords`) w/o hoisting peerDependencies  (i.e., `bar` & `ajv`).

In the example above, `ajv-keywords` should not be hoisted, and left as a sub-dependency of `webpack`.

yarn@0.28.1
node@4.8.2

**Related:**

https://github.com/yarnpkg/yarn/pull/3465
https://github.com/yarnpkg/yarn/issues/2688

**Dupes:**

https://github.com/yarnpkg/yarn/issues/3916
https://github.com/webpack/webpack/issues/5262

arcanis:
Should fix #3933 - not yet ready, the simple reproduction script pass, but not the real-world case.


arcanis:
K, I think I got it. Seems like the issue is the order the dependencies are processed. We currently sort them by name, so in some case we might do the check "does the peer dependency exists?" before the peer dependency has been traversed. Only issue with this theory is that `ajv` should be traversed before `ajv-keyword`, but apparently not.

arcanis:
Aaaah no, the compared patterns are `ajv-keywords@^1.0.0` and `ajv@^4.7.0`, not the package names, so `ajv-keywords` comes first.

src/package-hoister.js
```diff
@@ -121,7 +121,42 @@ export default class PackageHoister {
         return sortAlpha(aPattern, bPattern);
       });
 
-      for (const [pattern, parent] of queue) {
+      // sort the queue again to hoist packages without peer dependencies first
+      let sortedQueue = [];
+      const availableSet = new Set();
+
+      for (let hasChanged = true; queue.length > 0 && hasChanged;) {
```
BYK: This really begs for a `while` loop

src/package-hoister.js
```diff
@@ -121,7 +121,42 @@ export default class PackageHoister {
...
+        hasChanged = false;
+
+        for (let t = 0; t < queue.length; ++t) {
+          const pattern = queue[t][0];
+          const pkg = this.resolver.getStrictResolvedPattern(pattern);
+
+          const peerDependencies = Object.keys(pkg.peerDependencies || {});
+          const areDependenciesFulfilled = peerDependencies.every(peerDependency => availableSet.has(peerDependency));
```
BYK: This code inside the 2nd-level for loop looks a bit expensive to me. Do you think we can find more efficient ways to do this? Open for discussion.

src/package-hoister.js
```diff
@@ -291,8 +326,24 @@ export default class PackageHoister {
       }
     }
 
+    const peerDependencies = Object.keys(info.pkg.peerDependencies || {});
```
BYK: May be we need a helper function, `getPeerDependencies` since I see this `Object.keys(pkg.peerDependencies || {})` in multiple places now.

src/package-hoister.js
```diff
@@ -291,8 +326,24 @@ export default class PackageHoister {
...
+      for (const peerDependency of peerDependencies) {
+        const checkParts = parts.concat(peerDependency);
+        const checkKey = this.implodeKey(checkParts);
+        info.addHistory(`Looked at ${checkKey} for a peer dependency match`);
+
+        const existing = this.tree.get(checkKey);
+
+        if (existing) {
+          info.addHistory(`Found a peer dependency requirement at ${checkKey}`);
+          break hoistLoop;
```
BYK: It almost looks like this part should become a separate method to avoid this label usage.

src/package-hoister.js
arcanis: In theory it could be possible to avoid checking every fulfilled dependency every time, since we can just skip the ones we've already validated. In practice, the gain would be inconsequential and would depend on the engine optimizations more than anything else. Additionally, peer dependencies are rare enough for this code not to have a major (or even minor) impact on the perf. It's unlikely more than a single pass will be ever required.

That being said, there's maybe one optimization that can be useful, which is to just skip this whole codepath if there's no peer dependency (avoids us to do what is essentially a manual slice). Will do this.

src/package-hoister.js
arcanis: I feel like the indirection would make the code less clear. Why make a function call to an helper when you can just call the function this helper calls? It only matters if we plan to change the `info.pkg` underlying type and suddenly it wasn't a mapping of the package.json anymore, but I don't see what we would have to gain in doing that.

src/package-hoister.js
arcanis: I don't know - despite being usually avoided, the label has a great use case here, and I feel like the hoister is complex enough to warrant some of its logic to be kept inline, so that we don't have to jump everywhere in the file to understand its process.

src/package-hoister.js
arcanis: Nevermind, one needs to iterate on the dependencies to know if they have peer dependencies, so it can't really be optimized either. Eh. 😞

arcanis:
The next to last appveyor build failed because of a Github 500, and the last one failed because of a segfault. Third time's the charm ...

src/package-hoister.js
BYK: >Additionally, peer dependencies are rare enough for this code not to have a major (or even minor) impact on the perf. It's unlikely more than a single pass will be ever required.

I think this is a dangerous assumption in the long run since people are using `peerDependencies` more and more with many libraries as they are the "right way" for things like plugins etc.

src/package-hoister.js
BYK: >I feel like the indirection would make the code less clear.

It would actually make the code clearer since it gives a name to what you are doing and it is shorter: less cognitive overload. Also makes that part of the code separately testable and less prone to errors and easier to refactor since you hoisted up repeated code.

src/package-hoister.js
BYK: >I don't know - despite being usually avoided, the label has a great use case here, and I feel like the hoister is complex enough to warrant some of its logic to be kept inline

With this line of thinking, we can increase complexity infinitely instead of trying to find simpler ways of doing things.

>There's already too many functions imo :p

Too many functions >> too much code in a single function

mikegreiling:
Will this end up in an upcoming release anytime soon?  I see this was part of the `Yarn 1.0` card.  Will there be any releases between now and v1.0?

PR #4088, opened 2017-08-03
Fix: Fix numeric versions resolving to local files

### LINKED ISSUE #4013: install fails if directory has empty file with name "1"
bug

Weird install failure

package.json
```
{
  "name": "test",
  "dependencies": {
    "glob": "^7.1.1"
  }
}
```

```bash
touch 1
```

```bash
yarn install
yarn install v0.28.1
warning test: No license field
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "ENOTDIR: not a directory, scandir '/Users/bestander/Library/Caches/Yarn/v1/npm-wrappy-0.0.0-a3898aef-4262-4dd5-afa7-204e8e8cb0b2-1500935336291'".
info If you think this is a bug, please open a bug report with the information provided in "/Users/bestander/work/temp/weird/yarn-error.log".
```

Yarn 0.28.1, MacOSX

BYK:
**Summary**

Fixes #4013. We were blindly priotizing local files (even if they
are not directories) if there's anything on the file system matching
a pattern for a package. This patch ensures that we only do this if
the pattern is not a valid semver range and the matched local entity is a directory.
**Test plan**

Added a new unit test and modified an incorrect old one. Also manually
verified that the issue described in #4013 does not happen anymore.


bestander:
Afaik "foo": "bar" is legit in npm, and this looks like a breaking Chang anyway

arcanis:
If we want a perfect compatibility we also need to support the following:

```js
{
    "dependencies": {
        "a": "a/b/c"
    }
}
```

Starting from the third level of depth, NPM tries to install it from `./a/b/c`. Note that two levels of depth shouldn't match this codepath, because `a/b` is a github url (`github.com:a/b`).

arcanis:
@bestander `"foo": "bar"` is a valid selector, but it's not resolved on the filesystem. I think it's resolved as a tag of the package.

BYK:
>Afaik "foo": "bar" is legit in npm, and this looks like a breaking Chang anyway

Damn, you're right. Even if it is not documented, npm accepts this. The problem is, it installs the package as `bar` not `foo` which is not what we do AFAIK.

BYK:
Also, npm doesn't priotize folders over remote versions. I just tried it with having a folder named `1` and `npm i foo@1` and it installed the remote version.

BYK:
Looks like the rule is "try file if it is not a valid semver range". Updating accordingly.

__tests__/commands/install/integration.js
```diff
@@ -459,17 +459,30 @@ test.concurrent('install file: link file dependencies', async (): Promise<void>
...
 test.concurrent('install with file: protocol as default', (): Promise<void> => {
-  return runInstall({noLockfile: true}, 'install-file-as-default', async config => {
+  return runInstall({}, 'install-file-as-default', async config => {
     expect(await fs.readFile(path.join(config.cwd, 'node_modules', 'foo', 'index.js'))).toEqual('foobar;\n');
   });
 });
 
+test.concurrent("don't install with file: protocol as default if target is a file", (): Promise<void> => {
+  // $FlowFixMe
+  return expect(runInstall({lockfile: false}, 'install-file-as-default-no-file')).rejects.toBeDefined();
```
voxsim: I think the problem here is `rejects`, Do we use this somewhere else?

src/package-request.js
```diff
@@ -132,21 +132,27 @@ export default class PackageRequest {
     }
   }
 
-  async normalizeRange(pattern: string): Promise<string> {
+  async normalizeRange(name: string, pattern: string): Promise<string> {
```
voxsim: If I remember correctly we can delete every  `Promise.resolve` because `async/await` world we should not use them (i.e. It automatically change the result in Promise.resolve).

Do you think we can start to add some unit tests for package-request functions? 

I am afraid that every function in this file is not try to adress one responsability and we couple the code in function just for code duplication. One another proof of this is that we have many similar names. If we can't name something, maybe there is a reason.

voxsim:
@BYK I added some comments, If you think that they are out of scope, ignore them :) and thanks for change the flag 'noLockfile' in 'lockfile' :D

__tests__/commands/install/integration.js
BYK: It is because of that. This was introduced by Jest 20 and @onurtemizkan upgraded flow type definitions and used it in one of his patches.

I think Flow gets confused by how `runInstall` is defined by `bind` and cannot determine the output type of `expect` here.

src/package-request.js
BYK: >If I remember correctly we can delete every Promise.resolve because async/await world we should not use them (i.e. It automatically change the result in Promise.resolve).

Oh, right. This now makes it a double promise. Great catch!

>Do you think we can start to add some unit tests for package-request functions?

I'd love to but I'd also love to tie them to a "business" purpose. Right now most of these seem to be covered via integration tests. That said I want to do a pass on tests to make them more approachable so may be then?

>I am afraid that every function in this file is not try to adress one responsability and we couple the code in function just for code duplication. One another proof of this is that we have many similar names. If we can't name something, maybe there is a reason.

That's a good observation. I think we can give this file a closer look. That said I'd rather not try that in this patch since this PR resolves an important issue and refactoring those may create some back and forths.

BYK:
>@BYK I added some comments, If you think that they are out of scope, ignore them :) and thanks for change the flag 'noLockfile' in 'lockfile' :D

Hey, thanks for the comments! I'll go ahead and fix the `Promise.resolve()` parts. For more, I'll defer them to follow up to get this important fix in as soon as possible.

src/package-request.js
```diff
@@ -132,21 +132,27 @@ export default class PackageRequest {
...
-      return Promise.resolve(pattern);
+      return pattern;
     }
 
-    if (await fs.exists(path.join(this.config.cwd, pattern))) {
-      return Promise.resolve(`file:${pattern}`);
+    if (!semver.validRange(pattern)) {
+      try {
+        if ((await fs.stat(path.join(this.config.cwd, pattern))).isDirectory()) {
+          return `file:${pattern}`;
```
cpojer: Should we warn if this happens? I kinda feel like this file pattern isn't great.

cpojer: And I'm saying, we should support it, but hint at the user that it's silly.

__tests__/commands/install/integration.js
cpojer: I recommend turning it into an async function and awaiting the expect call. Looks a bit cleaner.

src/package-request.js
BYK: Can you be more specific? Do you think we should warn people when they omit the `file:` prefix? I think that's reasonable but would it be a warning or info? And what if "they just prefer it that way?". May be this should be controlled this via a config flag: `warn-unsafe-file false`?

__tests__/commands/install/integration.js
BYK: @cpojer you don't like returning promises from tests?

BYK:
>I'm approving to unblock but I think @arcanis came back to live and can glance over it before merging.

(y)

__tests__/commands/install/integration.js
cpojer: Not explicitly, no. Turning the function into an async function will implicitly return a promise, which Jest then uses to make the test wait for the promise.

BYK: Alright. I'll defer this to a later diff tho since it would require a large number of changes in this file (to be consistent) which are not very related to this particular problem. I know I did the `lockfile: false` changes but it was required to make sure tests were running correctly.

cpojer: Fine with me, file it under the post-post-1.0 tag ;P

src/package-request.js
```diff
@@ -134,14 +134,20 @@ export default class PackageRequest {
...
     if (pattern.indexOf(':') > -1 || pattern.indexOf('@') > -1 || getExoticResolver(pattern)) {
-      return Promise.resolve(pattern);
+      return pattern;
     }
 
-    if (await fs.exists(path.join(this.config.cwd, pattern))) {
-      return Promise.resolve(`file:${pattern}`);
+    if (!semver.validRange(pattern)) {
+      try {
+        if ((await fs.stat(path.join(this.config.cwd, pattern))).isDirectory()) {
```
arcanis: I'm not sure this check is the right one:

- It seems that NPM reports "Could not install from "a/b/c/d" as it does not contain a package.json file" even with an unexisting path
- It should work not only if the path is a directory, but also if it is a tarball

WDYT?

arcanis: Is there something preventing us from just fallbacking to the filesystem if the `pattern` doesn't match `validRange`?

BYK: Don't know if NPM supports tarballs. I'm fine with adding a bit more checks. My initial approach was a lot stricter actually: read the package.json file and make sure the name matches the wanted package name, hence that extra `name` argument in the previous version.

That said when we try to read a package manifest, we automatically fallback to the directory name as package name. This, combined with NPM being quite liberal about paths, lead to this final approach which just checks if it is a directory and it is not a valid semver range.

Awaiting further opinions :)

BYK: >Is there something preventing us from just fallbacking to the filesystem if the pattern doesn't match validRange?

Except for the fact that I find it very dangerous (even this current version), I don't think so.

BYK: Well, here's one: if there's a typo, it would make the error message quite cryptic.

arcanis: > make sure the name matches the wanted package name, hence that extra name argument in the previous version.

Yeah we shouldn't check that, especially since we have a feature to rename packages, so we might end up installing packages under a different name than their original one.

> Except for the fact that I find it very dangerous (even this current version), I don't think so.

How so? It seems less dangerous to me, since it means that whatever the filesystem is, Yarn would behave the same way on two different systems. Otherwise, there's a risk it will fallback to fetching the package from different ways if the file doesn't exists, instead of reporting about the missing file.

arcanis: That being said, I checked on NPM and it *seems* that they just silently ignore `"foo": "foo.tgz"` ... so I guess this PR should be fine :)

Issue #3788, opened 2017-07-01
`yarn list` shows devDependencies on a production install. 


pixelpax:
I boot a production version of my container with a Dockerfile that includes commands

**Do you want to request a *feature* or report a *bug*?**
I believe this is a bug as I receive misleading information. 

**What is the current behavior?**
I have a container built from an image with the following in its Dockerfile
```
ENV NODE_ENV=production
COPY package.json .
RUN yarn install --quiet
```

I am expecting it to be building in production and ignoring devDependencies. It does so, and I can confirm this with `ls node_modules` however, when I do `yarn ls` I see devDependencies listed alongside. 

**If the current behavior is a bug, please provide the steps to reproduce.**
In your package.json: 
```
  "dependencies": {
    "express": "^4.15.3"
  },
  "devDependencies": {
    "nodemon": "^1.11.0"
  }

```
then do
`NODE_ENV=production; yarn; yarn list | grep nodemon;`
and note that nodemon is present in list, but not present in `node_modules`
**What is the expected behavior?**
In this way, yarn is showing me a list of packages which I don't actually have stored on my machine. Perhaps this is intentional because I also see a list of packages _before_ executing `yarn` to install my `package.json`. I would expect to look in my package.json if I wanted to see what packages _should_ be installed and to use `yarn list` to see what packages are _actually_ installed. This is also what I'm used to with `npm`.

**Please mention your node.js, yarn and operating system version.**
Using yarn v0.24.6 within a docker container (from node:6.11 official)

gowthamgts:
@pixelpax Thanks for reporting this, I'm able to reproduce this in my local environment. Working on this.

bestander:
@gowthamgts, that is great.
Go ahead, the list command source is here https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/list.js

reznord:
Is this still open for a commit? If so, I'd like to take this and push a PR for this issue.

gowthamgts:
Yes @reznord, I'm caught up with work, so would be great if you take this up.

olingern:
Jumped in on this one, hope that's okay @reznord

reznord:
It's fine :)

PR #4092, opened 2017-08-04
Fix: Remove devDependencies from list output when environment is production

olingern:
**Summary**

Fixes: #3788.

This PR loads `devDependencies` from the manifest, re-builds the patterns package structure, and filters packages if the `NODE_ENV` is production.

**Test plan**

Added new tests.


BYK:
Related #4095.

src/constants.js
```diff
@@ -74,6 +74,15 @@ export const SINGLE_INSTANCE_PORT = 31997;
...
 
 export const ENV_PATH_KEY = getPathKey(process.platform, process.env);
+export const NODE_ENV = getEnv(process.env);
+
+export function getEnv(env: Env): string {
+  return env && env.NODE_ENV ? env.NODE_ENV : 'development';
+}
+
+export function isProduction(env: Env): boolean {
+  return getEnv(env) === 'production';
```
BYK: You should make `'development` and `'production` strings constants too ;)

src/cli/commands/list.js
```diff
@@ -180,21 +181,33 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
...
 }
 
+const getDevDeps = (manifest): Array<string> => Object.keys(manifest.devDependencies).map(key => `${key}@${manifest.devDependencies[key]}`);
+
 export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const lockfile = await Lockfile.fromDirectory(config.lockfileFolder, reporter);
   const install = new Install(flags, config, reporter, lockfile);
-  const {requests: depRequests, patterns, workspaceLayout} = await install.fetchRequestFromCwd();
+
+
```
BYK: These extra blank lines would probably fail lint.

src/cli/commands/list.js
```diff
@@ -180,21 +181,33 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
...
+  const {requests: depRequests, patterns, manifest, workspaceLayout} = await install.fetchRequestFromCwd();
   await install.resolver.init(depRequests, {
     isFlat: install.flags.flat,
     isFrozen: install.flags.frozenLockfile,
     workspaceLayout,
   });
 
+  let pkgPatterns = [];
+  if (isProduction(process.env)) {
```
BYK: I would just use `process.env` at all times instead of passing this. Also this should probably be a method on `config` since we also support passing `--prod` flag. Not sure that flag applies to `list` but we can make it so?

src/cli/commands/list.js
```diff
@@ -180,21 +181,33 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
...
+    const devDeps = getDevDeps(manifest);
+    pkgPatterns = patterns.filter(pattern => !devDeps.includes(pattern));
```
BYK: I would convert `devDeps` to a `Set` for faster look up. Technically, both can be sets since this is a set subtraction but the `Set` class doesn't have that method unless you use a third-party library.

olingern: Good point. This is probably O(n)^n since `filter` loops over the entire collection and then `includes` checks the entire collection for existence there, too. 

`devDeps` being a set could get the filtering down to constant access to see if a package is a dependency, but I don't see how converting `patterns` to a set would benefit this particular piece of code, since its creation would be linear as well. I could be blindly missing something, though :)

src/cli/commands/list.js
olingern: I thought about that, but it seemed that it would decrease the testability of `isProduction`

src/constants.js
olingern: Will do!

src/cli/commands/list.js
BYK: You should be able to mock `process.env` during tests ;)

src/cli/commands/list.js
BYK: Where `n` is the length of `devDeps` and `m` being the length of patterns it should go as follows:

 - O(n) for creating the set
 - O(1) for each look up
 - O(m) for creating the other set
 - O(m) for walking over the set

So I think you're right. Just converting one to a set for faster look up should be the best :)

BYK: Oh wait, there's one important nuance tho:

If `m` > `n` which usually is the case, then converting `m` to a set and then going over `devDeps` (O(n)) and removing items from the set should be faster.

olingern: Hmm ... 

So, let's say `a` is our dependencies and `b` is our dev dependencies.

For production, we're looking for the subset, `c`, which is found by `a` - `b`

In both cases where `a` > `b` and `b` > `a` you would have to create a set of the largest ( or smallest depending on where you want to take the linear hit ) and then iterate over the other collection to find the subset of `a` - `b`.

I think converting `devDeps` to a hash will get us to O(n) + m, which is pretty _okay_ :)

src/cli/commands/list.js
olingern: Added tests for these. Feels a little weird to store and restore the `NODE_ENV` instead of injecting, but if there's no reason for other portions of the code to care about `process.env`, then this makes sense :)

__tests__/constants.js
```diff
@@ -10,3 +10,45 @@ test('getPathKey', () => {
   expect(getPathKey('linux', {})).toBe('PATH');
   expect(getPathKey('darwin', {})).toBe('PATH');
 });
+
+test('NODE_ENV_DEV', () => expect(NODE_ENV_DEV).toBe('development'));
```
BYK: Don't think these are useful :)

BYK:
Well, you broke the build :D

olingern:
@BYK Build fixed. Was going to test stdout. Let me know what you think:

1.) Install packages from a fixture that has a devDependency
2.) Run in `development`. Make sure devDependency is in the output
3.) Run in `production`. Make sure devDependency is not in the output

BYK:
@olingern that plans sounds good to me. Keep in mind that you can create your own small fixtures using `file:` protocol so you don't need "real" packages for anything.

__tests__/constants.js
```diff
@@ -10,3 +10,22 @@ test('getPathKey', () => {
   expect(getPathKey('linux', {})).toBe('PATH');
   expect(getPathKey('darwin', {})).toBe('PATH');
 });
+
+test('isProduction', () => {
```
BYK: Protip: you can have `describe('isProduction')` at the top that wraps all this and then put each separate scenario in their own `test()` calls.

Btw. changing globals like `process.env.NODE_ENV` in tests is a bit dangerous since we run tests in parallel. It may yield to other problems. A no-to-great way is to define `isProduction` as follows that lets you do dependency injection:
```
function isProduction(env = process.env) {
    return env.NODE_ENV === 'production';
}
```

__tests__/fixtures/list/dev-deps-prod/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "left-pad": "^1.1.3"
```
BYK: Instead of relying on "real" dependencies, you can just construct local packages which may simplify and speed up things.

src/cli/commands/list.js
```diff
@@ -180,21 +181,36 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
   return notDim && (found || hasChildren);
 }
 
+export function getDevDeps(manifest: Object): Set<string> {
+  const devDepSet = new Set();
+  Object.keys(manifest.devDependencies).forEach(key => devDepSet.add(`${key}@${manifest.devDependencies[key]}`));
```
BYK: Even better:
```
return new Set(Object.keys(manifest.devDependencies).map(key => `${key}@${manifest.devDependencies[key]}`));
```

src/cli/commands/list.js
```diff
@@ -180,21 +181,36 @@ export function filterTree(tree: Tree, filters: Array<string>): boolean {
...
-  const {requests: depRequests, patterns, workspaceLayout} = await install.fetchRequestFromCwd();
+
+  const {requests: depRequests, patterns, manifest, workspaceLayout} = await install.fetchRequestFromCwd();
   await install.resolver.init(depRequests, {
     isFlat: install.flags.flat,
     isFrozen: install.flags.frozenLockfile,
     workspaceLayout,
   });
 
+  let pkgPatterns = [];
```
BYK: May be a more explanatory name than `pkgPatterns`? May be `usedPatterns` or `activePatterns`?

olingern: Good point. `activePatterns` has a nice "ring" to it.

src/cli/commands/list.js
olingern: Nice, will amend

__tests__/fixtures/list/dev-deps-prod/package.json
olingern: Hmmm ... could you point me to a test that follows that pattern? I was basing most of my structure off of the `no-args` tests / fixtures

BYK:
@olingern ping me when you push the final update then!

__tests__/fixtures/list/dev-deps-prod/package.json
BYK: Try this: https://github.com/yarnpkg/yarn/tree/master/__tests__/fixtures/install/install-file-relative ;)

__tests__/constants.js
olingern: Yah, wasn't sure of the overall test structure / style. It seemed that mosts tests are at parent level to each file, so I didn't know if nesting tests inside of a describe was good or bad style here.

> Btw. changing globals like process.env.NODE_ENV in tests is a bit dangerous since we run tests in parallel
Agreed. Wasn't sure of how to successfully mock this, but your advice above sorts that out :)

BYK: >It seemed that mosts tests are at parent level to each file, so I didn't know if nesting tests inside of a describe was good or bad style here.

Yeah, that's something I want to address soon. Apologies for the confusion :(

__tests__/commands/list.js
```diff
@@ -25,115 +25,153 @@ const runList = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, c
...
+      const children = [{name: 'is-plain-obj@^1.0.0', color: 'dim', shadow: true}];
+      const trees = [
+        makeTree('left-pad@1.1.3', {color: 'bold'}),
+        makeTree('sort-keys@1.1.2', {children, color: 'bold'}),
+        makeTree('is-plain-obj@1.1.0'),
+      ];
+
+      rprtr.tree('list', trees);
+
+      expect(tree).toEqual(rprtr.getBuffer());
```
olingern: I'm new-er to Jest, and just watched @cpojer's talk on Jest's new features. As I understand it, snapshots don't ensure validity, but keep a historical record (snapshot) of the current state. I could see snapshots being useful for quick smoke tests, but I'm curious to how you would use them here.

__tests__/commands/list.js
```diff
@@ -25,115 +25,153 @@ const runList = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, c
...
-});
 
-test.concurrent('expands a glob', (): Promise<void> => {
-  return runList(['gulp*'], {}, 'glob-arg', (config, reporter): ?Promise<void> => {
-    const rprtr = new BufferReporter({});
-    const tree = reporter.getBuffer().slice(-1);
-    const trees = [makeTree('gulp@3.9.1', {color: 'bold'}), makeTree('gulp-babel@6.1.2', {color: 'bold'})];
+  test('lists all dependencies when not production', (): Promise<void> => {
+    const nodeEnv = process.env.NODE_ENV;
+    process.env.NODE_ENV = 'development';
```
olingern: Gotcha

olingern:
@BYK I have couple questions / comments, but will get around to making all the fixes later today. 🎉

__tests__/commands/list.js
BYK: Here's my reasoning:

If you manually tested this and know this output is correct, we can safely assume that the output should not change unless the behavior changes. Thus, we can store this sample output as a reference and in this output we should/should not have devDependencies based on the prod state. Makes sense?

olingern: If, for some reason, the functionality of how `yarn list` changed -- `.toMatchSnapshot()` would catch the error, but it wouldn't give a diff on why / how to fix it (or maybe it would, I'm not privy to snapshot's expected / result output).

BYK: > but it wouldn't give a diff on why / how to fix it (or maybe it would, I'm not privy to snapshot's expected / result output).

It would :) It'd say "I expected this, and I got this other thing instead. Here's the diff."

olingern:
@BYK Made most of the updates, but couldn't find a way to mock `isProduction` via requiring the constants file.

Seems like there may be an answer [here](https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L23), but I'll have to take a look at it tomorrow.

BYK:
>Seems like there may be an answer here, but I'll have to take a look at it tomorrow.

That seems like the right place for me :) Seems like you have a few lint issues (`yarn lint` to see them locally). Fix them and we can merge the patch!

Thanks a lot for all the follow-ups, very much appreciated!

olingern:
@arcanis I can make sure linting / tests pass today

arcanis:
That would be great 😃 I tried running prettier, but unfortunately some Flow errors remain

olingern:
@BYK @arcanis linting fixed, tests updated

PR #4094, opened 2017-08-04
Update: use a named version folder in NPM tarballs

### LINKED ISSUE #3758: Cleanup release artifacts from having dist folder

BYK:
**Summary**

Fixes #3758. Makes the top-level folder in the tar archives have a name like `yarn-vX.Y.Z` instead of `dist` using the `--transform` and `-s` options in `tar` (they are different in GNU and BSD `tar`).

**Test plan**

Run `yarn build-dist` and then `tar -ztvf artifacts/yarn-v1.0.0.tar.gz`. Make sure the output lists all the files under `yarn-v1.0.0` directory.


BYK:
@Daniel15  - I'm now worried that this may break `deb` packages so any guidance around that would be appreciated.

Daniel15:
Looks good to me! I was going to say that we need to update the installation script, but we were clever and added `--strip 1` to the `tar` extract command so it's just stripping off the first piece of the path and thus doesn't care what it is.

`deb` looks fine, you can check the "artifacts" tab of CircleCI: https://circleci.com/gh/yarnpkg/yarn/4701#artifacts/containers/0. The file name has the correct version, and checking the package with `dpkg-deb` (ie. `curl https://4701-49970642-gh.circle-artifacts.com/0/home/ubuntu/yarn/artifacts/yarn_1.0.0_all.deb | dpkg-deb -I -`) shows the right version number on the package itself.

However, the Windows build is broken, as it's looking for a `dist` directory: https://github.com/yarnpkg/yarn/blob/cbcdede995ecbd94cb8ee037a51c55597721d4e4/resources/winsetup/YarnSetup.wixproj#L44-L48

`build-npm.sh` (which pushes Yarn to npm) also looks for a `dist` directory: https://github.com/yarnpkg/yarn/blob/77328c398e70eadc2780c8c58510a2c575953801/scripts/update-npm.sh#L6

Maybe the easiest solution here would be to rename the directory when building the tarball, then rename it back? Then all the scripts in Yarn itself could keep on using `dist`.

Daniel15:
Oh, even better @BYK: GNU tar has a `--transform` option to transform the file names when archiving: https://stackoverflow.com/a/9729944/210370. We could do `--transform s/^dist/yarn-vx.x.x/` when archiving to change the directory name when creating the archive, without actually physically renaming it.

BYK:
>Note that Mac OS comes with BSD tar rather than GNU tar but you can use Homebrew to get the GNU version.

Ah, this is very unfortunate. I want this code to work across platforms without issues and looks like GNU `tar` also has an `-s` option with a completely different meaning. I don't want to branch code by trying to guess which `tar` version people has so I think I'll go ahead and fix other places where `dist` is used. It would have been great of we could simply use this option tho :(

An alternative would be to just have a symlink named `dist` to the new, named folder but I guess that'd break on Windows then.

BYK:
I stand corrected: we need that `dist` folder so I'm gonna bite the bullet and detect the type of tar and tailor the command to it. Also leave the `dist` folder intact for the Windows installer.

BYK:
@Daniel15 ready

scripts/build-deb.sh
```diff
@@ -13,7 +13,7 @@ ensureAvailable lintian
 ensureAvailable rpmbuild
 
 PACKAGE_TMPDIR=tmp/debian_pkg
-VERSION=`dist/bin/yarn --version`
+VERSION=`./artifacts/yarn-legacy-* --version`
```
Daniel15: We can probably revert this back, right?

scripts/update-npm.sh
```diff
@@ -3,7 +3,7 @@
 
 set -ex
 
-version=`./dist/bin/yarn --version`
+version=`./artifacts/yarn-legacy-* --version`
```
Daniel15: This too?

scripts/build-deb.sh
BYK: Yeah we can. I'd leave this one though since this script doesn't rely on `dist` being there, but the other one does so I'll revert that one before landing.

PR #4105, opened 2017-08-06
Selective version resolution feature

kaylieEB:
**Summary**

An implementation of https://github.com/yarnpkg/rfcs/pull/68

1. Based on a glob pattern specified in the `resolutions` field of package.json, yarn will override specific versions of packages. This now includes ranges such as async@>0.9 <2.0 and exotic versions such as "async@https://github.com/caolan/async.git
2.  Warnings are generated when
- The glob pattern doesn't end with a valid package name (currently it only checks if it's empty after the last slash, since it's difficult to know what sub-dependencies exist at the time of install)
- When the given resolution version is incompatible with the version specified in the request package.json (see screenshot)
- When the given resolution version is an invalid semver (currently doesn't support exotic versions)
3. Edge cases that won't be supported yet

- Yarn's resolution method `getResolvedPattern` has no idea about the parent requests, so this means if one nested dependency _happens_ to have the same semver (say, package a has `left-pad~1.0.0` and package b also has `left-pad~1.0.0`), then they can end up being resolved to the same version defined in resolutions. However if they are different by any one character, such as semver range or version number, then they can individually be resolved. Imho I don't think it's too negative given that under normal circumstances, yarn will resolve both requests to the same package anyways.
- I have not yet implemented '**' case where all nested dependencies would resolve to particular version. I can't find any good use case for this. 

**Test plan**
I've modified / added scenarios to @victornoel's existing tests, and it didn't break existing functionalities. More tests coming soon!
//TODO: Add more test cases for warnings and ranges / exotic versions

**(Directly nested dependencies pattern)**
screen shot 2017-08-06 at 10 17 20 pm

**(All nested dependencies pattern)**
screen shot 2017-08-06 at 10 19 16 pm


src/cli/index.js
```diff
@@ -170,6 +170,7 @@ export function main({
   const run = (): Promise<void> => {
     invariant(command, 'missing command');
     return command.run(config, reporter, commander, commander.args).then(exitCode => {
+      reporter.close();
```
arcanis: Do you really need to call `close` on the reporter? It will break the `footer` call below

src/package-resolver.js
```diff
@@ -23,11 +24,12 @@ export type ResolverOptions = {|
 |};
 
 export default class PackageResolver {
-  constructor(config: Config, lockfile: Lockfile) {
+  constructor(config: Config, lockfile: Lockfile, resolutions: ?Resolutions) {
     this.patternsByPackage = map();
     this.fetchingPatterns = map();
     this.fetchingQueue = new BlockingQueue('resolver fetching');
     this.patterns = map();
+    this.resolutions = resolutions || new Resolutions(config);
```
arcanis: You can use a default parameter instead of `|| new Resolutions(config)`

src/resolutions.js
```diff
@@ -0,0 +1,101 @@
...
+  find(reqPattern: string, parentNames: Array<string>): ?string {
+    const {name, range} = PackageRequest.normalizePattern(reqPattern);
+    const resolutions = this.resolutionsByPackage[name];
+    let resolvedPattern = null;
+
+    if (!resolutions) {
+      return resolvedPattern;
+    }
+
+    const {version} = resolutions.find(({pattern}) => minimatch(path.join(...parentNames, name), pattern)) || {};
```
arcanis: We should probably use an explicit `.join('/')`, since `path.join` is system-specific.

src/resolutions.js
```diff
@@ -0,0 +1,101 @@
...
+
+    if (version) {
+      resolvedPattern = `${name}@${version}`;
+
+      if (!semver.valid(version)) {
+        this.reporter.warn(this.reporter.lang('invalidResolutionVersion', resolvedPattern));
+      }
```
arcanis: This warning will cause issues when people will use the exotic resolvers such as `file:` or git, which I assume will account for a fair percent of the feature usage (will allow people to fix a bug in a fork).

src/resolutions.js
```diff
@@ -0,0 +1,101 @@
...
+
+      if (!semver.satisfies(version, range)) {
+        this.reporter.warn(this.reporter.lang('incompatibleResolutionVersion', resolvedPattern, reqPattern));
+      }
```
arcanis: The resolution value is a range, not a pinned reference. We should delay printing a warning until the range has been resolved (we can do the check directly in the resolvers).

arcanis:
Nice PR! Left a few notes here and there, let me know what you think :)

src/cli/commands/install.js
```diff
@@ -165,13 +166,13 @@ export class Install {
...
     this.rootPatternsToOrigin = map();
-    this.resolutions = map();
     this.lockfile = lockfile;
     this.reporter = reporter;
     this.config = config;
     this.flags = normalizeFlags(config, flags);
-
-    this.resolver = new PackageResolver(config, lockfile);
+    this.resolutions = map();
+    this._resolutions = new Resolutions(config);
```
kaylieEB: hm so `this.resolutions` was what was being previously used for flat mode - I didn't necessarily want to mess with it but the rfc said the plan was to eventually have flat mode replaced by resolutions so I kind of left both as is. How would you rename it?

src/package-resolver.js
kaylieEB: you're right! thanks :)

src/resolutions.js
kaylieEB: Ok, my thought was that most people will put in versions and not exotic resolvers. But I can look into supporting the latter for v1 if you think it'll be most of the use case.

src/cli/index.js
kaylieEB: the reporter stuff is from the revert of the commit 280b6eb59ac6a8aaafdde1d86b48372ac76149f4 :(  it was necessary to have the feature working after rebasing, but let me re-revert it once a fix lands.

src/resolutions.js
kaylieEB: gotcha

src/package-resolver.js
cpojer: Note that they have different semantics. The default parameter will only work if the argument is omitted. If the argument is falsy, it will be null in the end:

```javascript
const fn = (arg = 42) => arg;
fn(null); // null

const fn = arg => arg || 42;
fn(null); // 42
```

It depends on what you are looking for.

src/cli/commands/install.js
arcanis: `ResolutionMap` / `ResolutionRegistry`?

src/resolutions.js
arcanis: I think it would be best if we could support it for 1.0, yep. But we can merge this PR before that I you prefer, and iterate on it.

kaylieEB:
I've resolved the TODO item above and been working on supporting exotic ranges -- will try to post the revision and merge this weekend :)

arcanis:
I've ran the tests locally and they failed to pass, I'll take some time to investigate a bit and see if my environment is responsible 👍

bestander:
Best feature ever!!!

cpojer:
Party time! @kaylieEB this work was impressive and we are all extremely grateful that you did all this work :)

victornoel:
Thanks a lot for taking care of that :)

EmielM:
Sorry for hijacking the topic, but was it ever a consideration to also allow filtering (nested) dependencies with this, eg by allowing "none", "skip" or "filter" in the resolution map? Seems like simple way to implement #4611.

PR #4107, opened 2017-08-07
Fix: deadlock when using network mutex and concurrent runs

BYK:
**Summary**

The current network mutex code relies on the old `process.exit()` to
terminate all existing connections. Since that has been removed now,
when a "master" `yarn` instance gets another one connected to it and
waiting for it to end, it stays in a zombie state due to the TCP
server not terminating with clients on it. This patch makes it shut
the server down properly when everything is done, fixing the deadlock.

**Test plan**

Adds a new integration test which hangs and times out without the fix.
Also can be tested manually, running two `yarn` instances side by side
with the `--mutex=network` option, concurrently.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -236,13 +236,13 @@ export function main({
       });
 
       const onServerEnd = async () => {
-        clients.forEach(client => {
+        for (const client of clients) {
           try {
             client.destroy();
           } catch (err) {
             // pass
           }
-        });
+        }
 
         await server.close();
         server.unref();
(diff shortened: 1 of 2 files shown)


src/cli/index.js
```diff
@@ -222,29 +223,43 @@ export function main({
...
+              resolve(runEventuallyWithNetwork(mutexPort));
             });
           });
       });
 
-      const onServerEnd = (): Promise<void> => {
-        server.close();
-        return Promise.resolve();
+      const onServerEnd = async () => {
+        clients.forEach(client => {
```
arcanis: `for ... of` would have been preferable

PR #4111, opened 2017-08-07
Fix: fix unknown single-instance type error

BYK:
**Summary**

When `:` is not used in the `mutex` argument to provide a specifier,
`mutexType` and `mutexSpecifier` are parsed incorrectly. This patch
fixes that.

**Test plan**

Run `yarn --mutex=network` before the patch, observe the error. Run
the same command after the patch and see it pass.


BYK:
>How come the code was working before?

It was using `.split(':')` followd by a `.shift()` and `.join(':')` before which was not prone to this mistake. Not to say there were no tests for the mutex behavior :D

Issue #3879, opened 2017-07-10
Publish lockfile parser as separate npm package


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Feature 

**What is the current behavior?**
Yarn lockfile parser is not available as a separate package

**What is the expected behavior?**
Yarn lockfile parser should be published (eg. as `@yarnpkg/lockfile` or something similar) so that people can use it in their own apps. Also see #3397

BYK:
@Daniel15 @arcanis I think we should include this in 1.0 plans if possible since it would be even harder to undo #3397 after 1.0 (breaking change)

Daniel15:
Would workspaces help with this? I'm not very familiar with how workspaces work, but I understand that they're similar to Lerna in that they help manage when you have multiple npm packages in the same repo?

Otherwise, we could just have a separate deployment process where we manually publish the lockfile parser to npm if it changes. It could be versioned separately as it should be fairly standalone and separate to Yarn itself, right?

arcanis:
Not sure about adding this to the 1.0 goal - we really need to make it right and I think that would require extra time. A few notes:

- I don't think internal APIs should count as breaking changes. We very explicitly don't support plugins atm, and until then making guarantees about our code and file hierarchy seems perilous. Doing it too early would lock us without any possibility to fix things.

- That being said, I can see plugins being the target goal for a 2.0 release. 1.0 is about stability, 2.0 could be about modularity.

- The parser should definitely be extracted at some point. It has nothing specific to Yarn (except maybe the newly merged "conflict merge" feature), and makes it harder for people to write tools outside of Yarn (*including us*).

All in all, I think we should extract the parser to its own package post-1.0, pre-2.0.

arcanis:
@Daniel15 Yup, workspaces would be the way to go for this. Yarn itself is a good candidate for showcasing the workspace feature 😃

cdaringe:
i want to vocalize my support of this.  dependency reports are so important!  if we don't expose the parser soon, can we at least expose a command in the short term to do a JSON dump?  that should be cheap and non invasive.

`yarn lockfile --json` or `yarn dependencyReport`, etc?

ggascoigne:
I'd like to add a "me too" to this issue as well, we also have a dependency reporting problem that would be simplified by access to the parser.

wmhilton:
Should I open a pull request to work on this feature?

wmhilton:
[...]

PR #4114, opened 2017-08-08
Begin separating lockfile into its own npm package.

wmhilton:
**Summary**
Publish lockfile parser as separate npm package as per #3879

**Test plan**

- [x] ~~Eliminate `../` requires~~ Since we're just bundling `yarn-lockfile` separately, not separating the codebase, these kinds of `../` dependencies aren't a show-stopper.
  - [x] `parse.js`
  - [x] `stringify.js`
  - [x] `wrapper.js`

NOTE: I put the static function `PackageRequest.normalizePattern()` into its own file because importing `PackageRequest` blew up the Webpack build for `yarn-lockfile` from 200KB to 2MB.

- [x] Change code that import `parse.js` or `stringify.js` directly to import the exported versions from `wrapper.js`
- [x] Rename `wrapper.js` to `index.js` and change imports from `../lockfile/wrapper.js` to `../lockfile`
- [x] Add a step to build `yarn-lockfile` to the Webpack config
- [x] ~~Change imports from `../lockfile` to `yarn-lockfile` and add `yarn-lockfile` as a dependency of `yarn`~~ we don't have to make `yarn-lockfile` a dependency, we can make it a by-product
- [x] ~~Automate publishing the `yarn-lockfile` package~~ Future work
- [x] Live happily ever after


wmhilton:
FYI I am being a little over-zealous with inlining imports. (e.g. I copied and pasted `sortAlpha` into three files) I'm just trying to get a feel for how easy or difficult it is to lift the parser/stringifier out of the rest of the code.

src/util/normalize-pattern.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
+/**
+ * Explode and normalize a pattern into it's name and range.
```
Daniel15: Small thing: `it's` should be `its` (no apostrophe)

scripts/build-webpack.js
```diff
@@ -33,7 +36,7 @@ const compiler = webpack({
     }),
   ],
   output: {
-    filename: `yarn-${version}.js`,
+    filename: `[name]-${version}.js`,
```
Daniel15: Hmmmm. I wonder if we should version the lockfile parser separately. We could call the current one `1.0`, and bump it to `2.0` if there's any major changes. What do you think?

BYK:
Oh also, let's fix CI first, forgot to mention that :)

cpojer:
@BYK I think this is fine as a first try, we can do the more ambitious separation later and more incrementally.

cpojer:
Also, @wmhilton thank you so much for contributing to Yarn!

arcanis:
Yep, I'm fine with this approach. Workspaces will eventually be integrated, but we'll probably have some questions to answer first, and this PR is a good way to see what needs we'll have.

Daniel15:
A simple-ish approach I can think of to publish this separately (without having to workspace-ify everything) is: 
1. Create `packages/lockfile/` folder
2. Create `packages/lockfile/package.json` with a simple package.json for the lockfile parser, with its own version number. I don't think it'd need any dependencies if we use the Webpack build
3. Make Webpack build it into `packages/lockfile/index.js`

We should then be able to run `npm publish` in `packages/lockfile` whenever we want to publish it. We could just publish it manually given how infrequently the lockfile parser changes.

src/util/normalize-pattern.js
wmhilton: 😁 I didn't write that. https://github.com/yarnpkg/yarn/pull/4114/files#diff-846935dd1790b71f1d658d684a6c7305L160 We can still fix the grammar though (in both places).

Actually I'm not entirely sure what this function does... I just concluded after inspecting it that it didn't depend on the rest of `src/package-request.js`, so I lifted it out to save ~2MB of weight in the bundled result.

scripts/build-webpack.js
wmhilton: Personally, I'd use the same version number everywhere (whether changes were made or not) because I think that is the clearest and easiest system to understand. I think Babel does this, and Lerna calls this the "synchronized" versioning scheme and is used by default. With the synchronized scheme it would be simple to publish a new lockfile package every time you publish a new yarn package and they could be part of the same build and release script.

Versioning the packages independently has advantages too... but those advantages are usually for the consumer (less frequent updates) and creates more work for the developers. Deciding on the version number of the lockfile package and whether or not to publish a new version becomes one more manual task that could be forgotten instead of part of the regular release process.

wmhilton: Mmm. But (as per your comment in the main thread) it may be simpler to get something out the door that's published manually, in which case independent versioning might be easier.

Daniel15: > because I think that is the clearest and easiest system to understand.

I do the same thing for one of my C# projects - Every package uses the same version number, even if it doesn't change as part of a release. Makes it easier to reason about the version number.

> Deciding on the version number of the lockfile package and whether or not to publish a new version becomes one more manual task that could be forgotten instead of part of the regular release process.

On the other hand, we might want to have a proper schema for the lockfile, and bump its major version separately for any major lockfile changes. Hmmmmmm

I'll leave this up to the other Yarn team members; I'm happy either way 😃

wmhilton:
> Oh, very interesting approach. So instead of releasing this as a separate package first, you make it a separate bundle so it can be distributed.

Yup! There's two categories of separation of concerns:
- developer-oriented separation of concerns, which would be modularizing the input (refactoring the source code)
- user-oriented separation of concerns, which is modularizing the output (the bundles / packages)

They are both valid, but I'm concentrating on the second category as it has the most bang for the buck.

wmhilton:
@Daniel15 I like that idea. Hold my avocado...

wmhilton:
@BYK Getting the CI tests passing again is my top priority

wmhilton:
> If we publish this to npm as a separate package, perhaps we should call it @yarnpkg/lockfile to better signify that it's "official".

I think that's a good idea, although AFAIK only Angular 2 has really taken advantage of scoped packages. React and Vue and Babel all still use prefixes... I'm not sure if it's a style thing or for practical reasons?

FWIW, I'd also be very happy/excited to hand over the [yarn-lockfile](https://npm.im/yarn-lockfile) package name!

wmhilton:
Looks like the tests are going to pass. Time for a squash and rebase!

wmhilton:
@Daniel15 I went ahead and granted you publishing control over [yarn-lockfile](https://npm.im/yarn-lockfile) so you have that as an option. The new "official" version should be 100% backwards compatible with my old module, and would fix https://github.com/wmhilton/yarn-lockfile/issues/1. Unfortunately I already released a 1.0.0 version, so this would have to be a 1.0.1 if you wanted to claim the name and take it over.

And if not and you want to publish it under `@yarnpkg/lockfile` or something, I will update my module to add a deprecation message pointing users to the official version and I will un-maintain mine.

cdaringe:
in the interim, i've updated https://www.npmjs.com/package/parse-yarn-lock. it now delivers a webpacked and es5 friendly build of _just_ the parser (+ it's deps) for each prod version of yarn.

that is the `parse-yarn-lock@0.28.4` tarball has a single .js file that exports the parser for that version.

wmhilton:
@cdaringe Lol. That is the wonderful thing about the npm registry - lots of redundancy! 👍

Daniel15:
> I think that's a good idea, although AFAIK only Angular 2 has really taken advantage of scoped packages. React and Vue and Babel all still use prefixes... I'm not sure if it's a style thing or for practical reasons?

I think the main reason is that they used to only be available to paying npm customers. Now you can make a free org so that's not an issue for open source any more. I think Babel has expressed interest in moving to scoped packages.

rarkins:
I haven't tried it but the current docs imply it's possible to publish scoped packages as a regular user too: https://docs.npmjs.com/misc/scope

> Each npm user/organization has their own scope, and only you can add packages in your scope. This means you don't have to worry about someone taking your package name ahead of you. Thus it is also a good way to signal official packages for organizations.

wmhilton:
> the current docs imply it's possible to publish scoped packages as a regular user too

Absolutely, I've published (public) scoped packages before. You just name the package `@username/foo`. It's free.

FWIW I noticed there is already a [@yarnpkg org](https://www.npmjs.com/org/yarnpkg). Hopefully you guys control that account? It hasn't published any packages.

Daniel15:
> I haven't tried it but the current docs imply it's possible to publish scoped packages as a regular user too: docs.npmjs.com/misc/scope

Oh yeah, true that! I think that was added recently too.

>FWIW I noticed there is already a @yarnpkg org. Hopefully you guys control that account? It hasn't published any packages.

Yep, we're in control of that one 😃

BYK:
>Is this ready to merge? Do I need to sign a CLA? Anything else I can do or are y'all happy to take it from here?

Hey, sorry for not getting back to this earlier. I'll give another pass of review. You don't need to sign a CLA AFAIK.

BYK:
In the meantime, looks like your branch can benefit from a merge from master or rebasing?

wmhilton:
@BYK here, have a rebase!

src/lockfile/index.js
```diff
@@ -102,14 +102,14 @@ export default class Lockfile {
...
-    const lockfileLoc = path.join(dir, constants.LOCKFILE_FILENAME);
+    const lockfileLoc = path.join(dir, LOCKFILE_FILENAME);
 
     let lockfile;
     let rawLockfile = '';
     let parseResult;
 
-    if (await fs.exists(lockfileLoc)) {
-      rawLockfile = await fs.readFile(lockfileLoc);
+    if (await exists(lockfileLoc)) {
```
BYK: I'd rather keep this the old way to find all `fs.exists` calls easily in the future but I'm guessing this is to reduce output size of the bundle by limiting the imports?

src/package-request.js
```diff
@@ -154,36 +155,7 @@ export default class PackageRequest {
...
-      }
-    }
-
-    // add back @ scope suffix
-    if (isScoped) {
-      name = `@${name}`;
-    }
-
-    return {name, range, hasVersion};
+    return normalizePattern(pattern);
```
BYK: Why do we need to wrap this external function instead of just requiring it from where it is already being used?

wmhilton: It originally lived inside package-request, but bundling package-request blew up the bundle by 10x, so I moved it outside of package-request, but didn't want to upset the code that expected it to live on the class static method.

That was before I boldly found & replaced all the `wrapper.js`s in all the files. I could probably do the same here with `normalizePattern`

src/lockfile/index.js
wmhilton: Yes, I think there was a transitive dependency that blew up if you included all of fs.js. I'll do a quick with-and-without to compare.

wmhilton: Using

`import {exists, readFile} from './util/fs.js'`, `exist()` and `readFile`

instead of

`import fs from './util/fs.js'`, `fs.exist()` and `fs.readFile()`

appears to have no effect on the bundle output size. Huh. I was probably just being overly careful. I can return to the olden way.

wmhilton:
I went ahead and ripped `PackageRequest.normalizePattern` out totally and changed all the places that referenced it to import `normalizePattern` directly. Interestingly, I'd say about half the files changed were including PackageRequest for that single static function. Also, I reverted the `import {exists, readFile} from './fs.js'` back to `import fs from './fs.js'` after testing showed it didn't make a difference on bundle size.

wmhilton:
✨ 🎺 ✨ 🎉

Haroenv:
👌

cpojer:
@wmhilton great work, thank you so much for your contribution and splitting the lockfile parser off of Yarn :)

wmhilton:
I look forward to seeing what cool community inventions are invented as a result. In particular I think there is opportunity for plugins to bundlers like webpack or dynamic module loaders like SystemJS to use lock files for module resolution.

_Sent from my Motorola XT1635-02 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

modosc:
dumb question, but how do i install this? `yarn add yarn-lockfile` pulls in what i think is the old version which then pulls in `yarn *` as a dependency. or is this not published yet?

Daniel15:
@modosc  - It's not published yet. It'll likely be published as `@yarnpkg/lockfile`.

modosc:
@Daniel15 thanks - is there an issue i can watch to track the status?

BYK:
@modosc now there is: #4221

wmhilton:
@Daniel15 Quick question: going forward the `@yarnpkg/lockfile` module's exports will be a public-facing API. Should `implodeEntry` and `explodeEntry` be part of that? A quick grep shows `implodeEntry` is only used once, in `cli/commands/generate-lock-entry.js` and `explodeEntry` is never used outside of `lockfile/index.js`.

```
> node                                       
> yarn = require('@yarnpkg/lockfile')        
{ parse: [Getter],                           
  stringify: [Getter],                       
  implodeEntry: [Function: implodeEntry],    
  explodeEntry: [Function: explodeEntry],    
  default: [Function: Lockfile] }            
>                                            
```

Issue #3863, opened 2017-07-08
workspaces - not linking .bin and not installing unhosited package if top level node_modules was cached


Bnaya:
**Do you want to request a *feature* or report a *bug*?**
bug

**Prologue**:
CI services, (In my case codeship) are caching the top level `node_modules` dir on the project
To speedup future npm/yarn installs

**What is the current behavior?**
 - On the first run on the CI, when's there's no node_modules dir, yarn will go and link the needed bins to the sub packages, install unhoisted deps, etc
 - On a repetitive run of the CI, `yarn install` will look at the cached `node_modules` and will think he is all good and won't perform the bin linking and unhoisted install  

**What is the expected behavior?**
yarn should link and install unhoisted deps for the workspaces

**Further thinking**
This issue also happens not only when the node_modules was cached, but if for some reason i've delete packages dir and checkedin it again from git and similar scenarios 

**Immediate workaround**
Delete the top level `node_module` dir/disable CI `node_module` cache. but not yarn cache!
For my project `yarn install` is very quick so i can live with that for now

**Please mention your node.js, yarn and operating system version.**
yarn 0.27.5
node 8.1.2

bestander:
Thanks for reporting.
Yarn has a bug in integrity check for workspaces.
A workaround for you now is `rm node_modules/.yarn-integrity` before running `yarn install`.

arcanis:
After some investigations, a few details:

  - The behavior as described in the OP (ie running Yarn with no flags) is not a bug. When yarn starts, it tries to find a yarn-integrity file. If it exists, and if the content of the file match what we should have should we generate a new yarn-integrity file, then we bail out and print "Already up-to-date". The default behavior only include the lockfile entries inside the integrity file, so even if something changes inside the node_modules filesystem, we won't pick it up **by default**. This is not related to the workspaces, the same thing can be observed with regular installs:

    ```
    echo '{"dependencies":{"babel-cli":"*"}}' > package.json
    yarn
    rm -rf node_modules/*
    yarn # Already up-to-date
    ```

  - If you need to bypass this behavior on a one-time basis (for example when running your CI tests), you can disable this logic by using the `--skip-integrity-check` or `--force` flags, which will both force Yarn to run the regular install whatever the integrity file says.

[...]

Issue #4109, opened 2017-08-07
Remove the registry codepaths


arcanis:
**What is the current behavior?**

Yarn currently tries to support multiple registries (think npm vs bower). However, the code that does it is bogus at best, isn't implemented in a lot of places, isn't tested, isn't really enabled, and frankly isn't useful at the moment. I think we should remove everything related to it and only support npm modules, it would make the code much easier to read. Once we want to actually support a new registry type we will be able to design it properly.

Note that this doesn't affect private registries or things like this - I'm only talking here about registry types. We will still support changing the origin of the registry queries.

PR #4122, opened 2017-08-08
Fix: Reimplements part of the integrity check

arcanis:
**Summary**

The integrity file doesn't currently support workspaces, which leads
to #3863 (removing a workspace node_modules folder and running yarn
doesn't do anything, because Yarn incorrectly assumes that everything
is already done).

This PR rewrites a good chunk of the integrity logic to support and
fix this behavior.

**Test plan**

Tests are passing, I still have to write an automated test before
merging this PR.

### DIFF
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -193,6 +193,7 @@ export default class InstallationIntegrityChecker {
 
     if (workspaceLayout) {
       result.topLevelPatterns = result.topLevelPatterns.filter(p => {
+        // $FlowFixMe
         return !workspaceLayout.getManifestByPattern(p);
       });
 
@@ -203,13 +204,17 @@ export default class InstallationIntegrityChecker {
 
         const manifest = workspaceLayout.workspaces[name].manifest;
 
-        for (const dependencyType of constants.DEPENDENCY_TYPES) {
-          if (!manifest[dependencyType]) {
-            continue;
-          }
+        if (manifest) {
+          for (const dependencyType of constants.DEPENDENCY_TYPES) {
+            const dependencies = manifest[dependencyType];
 
-          for (const dep of Object.keys(manifest[dependencyType])) {
-            result.topLevelPatterns.push(`${dep}@${manifest[dependencyType][dep]}`);
+            if (!dependencies) {
+              continue;
+            }
+
+            for (const dep of Object.keys(dependencies)) {
+              result.topLevelPatterns.push(`${dep}@${dependencies[dep]}`);
+            }
           }
         }
       }
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -185,7 +185,21 @@ export default class InstallationIntegrityChecker {
       artifacts,
     };
 
-    result.topLevelPatterns = patterns.sort(sortAlpha);
+    result.topLevelPatterns = patterns.sort(sortAlpha).filter(p => {
+      return !workspaceLayout || !workspaceLayout.getManifestByPattern(p);
+    });
+
+    // If using workspaces, we also need to add the workspaces patterns to the top-level, so that we'll know if a
+    // dependency is added or removed into one of them. We must take care not to read the aggregator.
+    if (workspaceLayout) {
+      for (const name of Object.keys(workspaceLayout.workspaces)) {
+        if (workspaceLayout.workspaces[name].loc) {
+          result.topLevelPatterns = result.topLevelPatterns.concat(
+            workspaceLayout.workspaces[name].manifest._reference.patterns,
+          );
+        }
+      }
+    }
 
     if (flags.checkFiles) {
       result.flags.push('checkFiles');
@@ -222,9 +236,7 @@ export default class InstallationIntegrityChecker {
       const modulesRoot = this._getModulesRootFolder();
 
       result.files = (await this._getIntegrityListing({workspaceLayout}))
-        .map(entry => {
-          return path.relative(modulesRoot, entry);
-        })
+        .map(entry => path.relative(modulesRoot, entry))
         .sort(sortAlpha);
     }
 
@@ -266,11 +278,7 @@ export default class InstallationIntegrityChecker {
       return 'FLAGS_DONT_MATCH';
     }
 
-    const actualPatterns = actual.topLevelPatterns.filter(p => {
-      return !workspaceLayout || !workspaceLayout.getManifestByPattern(p);
-    });
-
-    if (!compareSortedArrays(actualPatterns, expected.topLevelPatterns || [])) {
+    if (!compareSortedArrays(actual.topLevelPatterns, expected.topLevelPatterns || [])) {
       return 'PATTERNS_DONT_MATCH';
     }
 
@@ -292,16 +300,19 @@ export default class InstallationIntegrityChecker {
         return 'FILES_MISSING';
       }
 
-      // Since we know the "files" entry is sorted alphabetically, we can optimize the thing
+      // Since we know the "files" entry is sorted (alphabetically), we can optimize the thing
       // Instead of storing the files in a Set, we can just iterate both arrays at once. O(n)!
       for (let u = 0, v = 0; u < expected.files.length; ++u) {
-        // Skip over files that have been added
-        while (v < actual.files.length && actual.files[v] !== expected.files[u]) {
+        // Number of iterations after which there won't be enough entries remaining for the arrays to match
+        const max = actual.files.length - expected.files.length;
+
+        // Skip over files that have been added (not present in 'expected')
+        while (v < max && actual.files[v] !== expected.files[u]) {
           v += 1;
         }
 
         // If we've reached the end of the actual array, the file is missing
-        if (v === actual.files.length) {
+        if (v === max) {
           return 'FILES_MISSING';
         }
       }
@@ -332,16 +343,10 @@ export default class InstallationIntegrityChecker {
 
     const expected = await this._getIntegrityFile(loc.locationPath);
     let integrityMatches = this._compareIntegrityFiles(actual, expected, flags.checkFiles, workspaceLayout);
-    console.log(actual, expected, integrityMatches);
 
     if (integrityMatches === 'OK') {
       invariant(expected, "The integrity shouldn't pass without integrity file");
       for (const modulesFolder of expected.modulesFolders) {
-        console.log(
-          modulesFolder,
-          this.config.lockfileFolder,
-          await fs.exists(path.join(this.config.lockfileFolder, modulesFolder)),
-        );
         if (!await fs.exists(path.join(this.config.lockfileFolder, modulesFolder))) {
           integrityMatches = 'MODULES_FOLDERS_MISSING';
         }
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -300,18 +300,18 @@ export default class InstallationIntegrityChecker {
         return 'FILES_MISSING';
       }
 
-      // Since we know the "files" entry is sorted (alphabetically), we can optimize the thing
+      // Since we know the "files" array is sorted (alphabetically), we can optimize the thing
       // Instead of storing the files in a Set, we can just iterate both arrays at once. O(n)!
       for (let u = 0, v = 0; u < expected.files.length; ++u) {
-        // Number of iterations after which there won't be enough entries remaining for the arrays to match
-        const max = actual.files.length - expected.files.length;
+        // Index that, if reached, means that we won't have enough food to match the remaining expected entries anyway
+        const max = v + (actual.files.length - v) - (expected.files.length - u) + 1;
 
-        // Skip over files that have been added (not present in 'expected')
+        // Skip over files that have been added (ie not present in 'expected')
         while (v < max && actual.files[v] !== expected.files[u]) {
           v += 1;
         }
 
-        // If we've reached the end of the actual array, the file is missing
+        // If we've reached the index defined above, the file is either missing or we can early exit
         if (v === max) {
           return 'FILES_MISSING';
         }
--- a/src/integrity-checker.js
+++ b/src/integrity-checker.js
@@ -64,7 +64,7 @@ export default class InstallationIntegrityChecker {
   config: Config;
 
   /**
-   * Get the common ancestor of every node_modules - it might be a node_modules directory itself, but isn't required to.
+   * Get the common ancestor of every node_modules - it may be a node_modules directory itself, but isn't required to.
    */
 
   _getModulesRootFolder(): string {
@@ -131,7 +131,7 @@ export default class InstallationIntegrityChecker {
       }
     }
 
-    return locations;
+    return locations.sort(sortAlpha);
   }
 
   /**
@@ -141,10 +141,6 @@ export default class InstallationIntegrityChecker {
     const files = [];
 
     const recurse = async dir => {
-      if (!await fs.exists(dir)) {
-        return;
-      }
-
       for (const file of await fs.readdir(dir)) {
         const entry = path.join(dir, file);
         const stat = await fs.lstat(entry);
@@ -158,7 +154,9 @@ export default class InstallationIntegrityChecker {
     };
 
     for (const modulesFolder of this._getModulesFolders({workspaceLayout})) {
-      await recurse(modulesFolder);
+      if (await fs.exists(modulesFolder)) {
+        await recurse(modulesFolder);
+      }
     }
 
     return files;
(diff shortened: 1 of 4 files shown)


arcanis:
The tests are failing consistently on CI, but not in local oõ

src/integrity-checker.js
```diff
@@ -118,23 +109,59 @@ export default class InstallationIntegrityChecker {
   }
 
   /**
-   * returns a list of files recursively in a directory sorted
+   * Get the list of the directories that contain our modules (there might be multiple such folders b/c of workspaces).
```
BYK: s/might/may

src/integrity-checker.js
```diff
@@ -118,23 +109,59 @@ export default class InstallationIntegrityChecker {
...
-      for (const file of await fs.readdir(currentDir)) {
-        const entry = path.join(currentDir, file);
+
+  _getModulesFolders({workspaceLayout}: {workspaceLayout: ?WorkspaceLayout} = {}): Array<string> {
+    const locations = [];
+
+    if (this.config.modulesFolder) {
+      locations.push(this.config.modulesFolder);
+    } else {
+      locations.push(path.join(this.config.lockfileFolder, 'node_modules'));
```
BYK: No `'node_modules'` please.

src/integrity-checker.js
```diff
@@ -118,23 +109,59 @@ export default class InstallationIntegrityChecker {
...
+  }
+
+  /**
+   * Get a list of the files that are located inside our module folders.
+   */
+  async _getIntegrityListing({workspaceLayout}: {workspaceLayout: ?WorkspaceLayout} = {}): Promise<Array<string>> {
+    const files = [];
+
+    const recurse = async dir => {
+      if (!await fs.exists(dir)) {
```
BYK: Why do we need this check?

src/integrity-checker.js
```diff
@@ -118,23 +109,59 @@ export default class InstallationIntegrityChecker {
...
+    return locations;
+  }
+
+  /**
+   * Get a list of the files that are located inside our module folders.
+   */
+  async _getIntegrityListing({workspaceLayout}: {workspaceLayout: ?WorkspaceLayout} = {}): Promise<Array<string>> {
+    const files = [];
+
+    const recurse = async dir => {
```
BYK: I'd use a more descriptive name. Something like `getDeepFileList` or `getFullDirectoryTree`

src/integrity-checker.js
```diff
@@ -118,23 +109,59 @@ export default class InstallationIntegrityChecker {
...
+          await recurse(entry);
         } else {
-          files.push(path.relative(rootDir, entry));
+          files.push(entry);
         }
       }
+    };
+
+    for (const modulesFolder of this._getModulesFolders({workspaceLayout})) {
+      await recurse(modulesFolder);
```
BYK: No need to serialize each of these. Why not use `Promise.all` here:
```
const allFolders = [];
for (const modulesFolder of this._getModulesFolders({workspaceLayout})) {
  allFolders.push(recurse(modulesFolder));
});
await Promise.all(allFolders);
```

Alternatively:
```
await Promise.all(this._getModulesFolders({workspaceLayout}).map(folder => recuse(folder)));
```

Also, instead of changing `files` implicitly, it may be better to return it from `recurse` and then `.concat()` at the end. May be slower and churn memory tho.

src/integrity-checker.js
```diff
@@ -171,16 +203,29 @@ export default class InstallationIntegrityChecker {
...
+        result.modulesFolders.push(path.relative(this.config.lockfileFolder, modulesFolder));
+      }
+    }
 
     if (flags.checkFiles) {
-      result.files = await this._getFilesDeep(modulesFolder);
+      const modulesRoot = this._getModulesRootFolder();
+
+      result.files = (await this._getIntegrityListing({workspaceLayout}))
+        .map(entry => {
```
BYK: `.map(entry => path.relative(modulesRoot, entry))`

src/integrity-checker.js
```diff
@@ -196,52 +241,68 @@ export default class InstallationIntegrityChecker {
...
     if (!compareSortedArrays(actual.linkedModules, expected.linkedModules)) {
       return 'LINKED_MODULES_DONT_MATCH';
     }
-    if (!compareSortedArrays(actual.flags, expected.flags)) {
+
+    let relevantExpectedFlags = expected.flags.slice();
+
+    // If we run "yarn" after "yarn --check-files", we shouldn't fail the less strict validation
+    if (actual.flags.indexOf('checkFiles') === -1) {
+      relevantExpectedFlags = relevantExpectedFlags.filter(flag => flag !== 'checkFiles');
```
BYK: Feels like `'checkFiles'` and possible other flags should be in constants.

src/integrity-checker.js
```diff
@@ -196,52 +241,68 @@ export default class InstallationIntegrityChecker {
...
     }
+
     if (checkFiles) {
-      if (expected.files.length === 0) {
-        // edge case handling - --check-fies is passed but .yarn-integrity does not contain any files
-        // check and fail if there are file in node_modules after all.
-        const actualFiles = await this._getFilesDeep(locationFolder);
-        if (actualFiles.length > 0) {
-          return 'FILES_MISSING';
+      // Early bailout if we expect more files than what we have
```
BYK: Smart

src/integrity-checker.js
```diff
@@ -196,52 +241,68 @@ export default class InstallationIntegrityChecker {
...
         }
-      } else {
-        // TODO we may want to optimise this check by checking only for package.json files on very large trees
-        for (const file of expected.files) {
-          if (!await fs.exists(path.join(locationFolder, file))) {
-            return 'FILES_MISSING';
-          }
+
+        // If we've reached the end of the actual array, the file is missing
+        if (v === actual.files.length) {
```
BYK: Another optimization: whenever `actual.files.length - v` < `expected.files.length - u` we're done.

src/integrity-checker.js
```diff
@@ -259,28 +320,33 @@ export default class InstallationIntegrityChecker {
...
     const expected = await this._getIntegrityFile(loc.locationPath);
-    const integrityMatches = await this._compareIntegrityFiles(
-      actual,
-      expected,
-      flags.checkFiles,
-      loc.locationFolder,
-      workspaceLayout,
-    );
+    let integrityMatches = this._compareIntegrityFiles(actual, expected, flags.checkFiles, workspaceLayout);
+    console.log(actual, expected, integrityMatches);
```
BYK: Don't forget to remove this guy :)

src/integrity-checker.js
```diff
@@ -259,28 +320,33 @@ export default class InstallationIntegrityChecker {
...
+
+    if (integrityMatches === 'OK') {
+      invariant(expected, "The integrity shouldn't pass without integrity file");
+      for (const modulesFolder of expected.modulesFolders) {
+        console.log(
+          modulesFolder,
+          this.config.lockfileFolder,
+          await fs.exists(path.join(this.config.lockfileFolder, modulesFolder)),
+        );
+        if (!await fs.exists(path.join(this.config.lockfileFolder, modulesFolder))) {
```
BYK: This needs comment since I don't follow the logic. This should have been caught by the checks above already, right?

src/integrity-checker.js
```diff
@@ -63,51 +63,42 @@ export default class InstallationIntegrityChecker {
...
-    if (this.config.enableMetaFolder) {
-      locationFolder = path.join(this.config.lockfileFolder, constants.META_FOLDER);
+  _getIntegrityFileFolder(): string {
+    if (this.config.modulesFolder) {
+      return this.config.modulesFolder;
+    } else if (this.config.enableMetaFolder) {
+      return path.join(this.config.lockfileFolder, constants.META_FOLDER);
     } else {
-      locationFolder = await this._getModuleLocation(usedRegistries);
+      return path.join(this.config.lockfileFolder, 'node_modules');
```
arcanis: No, workspaceRootFolder should only be used for resolving relative paths. In workspaces it will be the package root folder, and we don't want to store the `.yarn-integrity` file there.

src/integrity-checker.js
arcanis: Parallel execution would require sorting the results afterward, so it should make little difference in the end.

BYK: I think it still would since sorting is CPU intensive and it is quite cheaper than IO.

src/integrity-checker.js
arcanis: Yeah I'm debugging why it breaks

src/integrity-checker.js
arcanis: Nope

  - the `files` entry contains the list of files, but is only there if we use the `--check-files` option. Otherwise it does no check at all.

  - the `modulesFolders` entry is a static list of directory that isn't baked by the filesystem. They are not the actual physical folders, they are the theoretical folders. So when we compare them in `_compareIntegrityFiles`, we don't actually check they exist. That's done here.

BYK: >We still want to check if at the very least the directories exist.

See, we need to add a comment about this :D

>the modulesFolders entry is a static list of directory that isn't baked by the filesystem. They are not the actual physical folders, they are the theoretical folders. So when we compare them in _compareIntegrityFiles, we don't actually check they exist. That's done here.

Wow. Again, needs comment so any future person would know this when they look at the code.

src/integrity-checker.js
```diff
@@ -185,7 +185,21 @@ export default class InstallationIntegrityChecker {
       artifacts,
     };
 
-    result.topLevelPatterns = patterns.sort(sortAlpha);
+    result.topLevelPatterns = patterns.sort(sortAlpha).filter(p => {
```
BYK: I'd do `filter` first and `sort` second to reduce the runtime of sort.

src/integrity-checker.js
```diff
@@ -185,7 +185,21 @@ export default class InstallationIntegrityChecker {
...
+      return !workspaceLayout || !workspaceLayout.getManifestByPattern(p);
```
BYK: Since `workspaceLayout` won't ever change during the lifetime of this function, you can do the following and speed up the cases where workspaces are not used:
```
const topLevelPatterns = workspaceLayout ? patterns.filter(p => !workspaceLayout.getManifestByPattern(p)) : patterns;
result.topLevelPatterns = topLevelPatterns.sort(sortAlpha);
```

src/integrity-checker.js
```diff
@@ -185,7 +185,21 @@ export default class InstallationIntegrityChecker {
...
+    });
+
+    // If using workspaces, we also need to add the workspaces patterns to the top-level, so that we'll know if a
+    // dependency is added or removed into one of them. We must take care not to read the aggregator.
+    if (workspaceLayout) {
+      for (const name of Object.keys(workspaceLayout.workspaces)) {
+        if (workspaceLayout.workspaces[name].loc) {
+          result.topLevelPatterns = result.topLevelPatterns.concat(
```
BYK: This breaks sorting, right? Is that intentional?

src/integrity-checker.js
```diff
@@ -292,16 +300,19 @@ export default class InstallationIntegrityChecker {
...
       }
 
-      // Since we know the "files" entry is sorted alphabetically, we can optimize the thing
+      // Since we know the "files" entry is sorted (alphabetically), we can optimize the thing
       // Instead of storing the files in a Set, we can just iterate both arrays at once. O(n)!
       for (let u = 0, v = 0; u < expected.files.length; ++u) {
-        // Skip over files that have been added
-        while (v < actual.files.length && actual.files[v] !== expected.files[u]) {
+        // Number of iterations after which there won't be enough entries remaining for the arrays to match
+        const max = actual.files.length - expected.files.length;
```
BYK: Move this out of the loop? :)

src/integrity-checker.js
```diff
@@ -292,16 +300,19 @@ export default class InstallationIntegrityChecker {
...
+
+        // Skip over files that have been added (not present in 'expected')
+        while (v < max && actual.files[v] !== expected.files[u]) {
           v += 1;
         }
 
         // If we've reached the end of the actual array, the file is missing
-        if (v === actual.files.length) {
+        if (v === max) {
```
BYK: Don't think this logic is correct. `max` is just a "small" difference where `v` is the absolute index which has to gi past `max`. If you kept a separate `advance` variable and calculated `v` as `u + advance` and did `if (advance === max)` then the logic would be correct.

src/integrity-checker.js
arcanis: It's still sorted, just not sorted alphabetically. As long as it has a deterministic order, it's should be fine.

src/integrity-checker.js
arcanis: I did this originally but I find it more readable for this variable and its comment to be close to the place we're using it. Since this is something that could/should be optimized by a decent compiler I think it's fine.

BYK: Well, you can move it to the initialization part of the `for` loop if you wanna keep it cozy? :D

src/integrity-checker.js
BYK: `Object.keys(workspaceLayout.workspaces)` doesn't have a deterministic order AFAIK so you may wanna sort the keys first. Also we should add a comment about this into the code.

src/integrity-checker.js
```diff
@@ -196,52 +253,67 @@ export default class InstallationIntegrityChecker {
...
-        if (actualFiles.length > 0) {
-          return 'FILES_MISSING';
+      // Early bailout if we expect more files than what we have
+      if (expected.files.length > actual.files.length) {
+        return 'FILES_MISSING';
+      }
+
+      // Since we know the "files" array is sorted (alphabetically), we can optimize the thing
+      // Instead of storing the files in a Set, we can just iterate both arrays at once. O(n)!
+      for (let u = 0, v = 0; u < expected.files.length; ++u) {
```
BYK: I honestly think the following is easier to follow:
```
const maxDiff = actual.files.length - expected.files.length;
if (maxDiff < 0) {
    return 'FILES_MISSING';
}
const advance = 0;
for (const u = 0; u < expected.files.length; ++u) {
  while (advance < maxDiff && actual.files[u + advance] !== expected.files[u]) {
    advance += 1;
  }
  if (advance === maxDiff) { 
    return 'FILES_MISSING';
  }
}
```

src/integrity-checker.js
arcanis: Yeah I guess. It's actually a bit more complex: the sorting order is de facto implementation-defined, so it will stay the same for any given engine, which is all we need. I thought about sorting the array but it wouldn't change anything ... except be a bit safer in some weird edge cases.

arcanis: Still you're right, I'll add it, just in case. Better be safe :)

src/integrity-checker.js
arcanis: Hm, I think your code won't early exit when the check cannot succeed anymore. It will always iterate on all of `actual`.

src/integrity-checker.js
arcanis: The `dir` might not exists at the root. I'll move it inside the l160 loop, since it only matter there.

src/integrity-checker.js
arcanis: Do you have numbers on perf benefits when parallelizing I/O? Intuitively I would think that in the end, the disk must execute all the operations sequentially, but I've never really looked.

src/integrity-checker.js
BYK: I don't think assuming engine internals will stay the same across runs is safe enough, yeah. Thanks! :)

src/integrity-checker.js
```diff
@@ -63,51 +63,42 @@ export default class InstallationIntegrityChecker {
...
-        loc = possibleLoc;
-        break;
-      }
+  _getModulesRootFolder(): string {
+    if (this.config.modulesFolder) {
+      return this.config.modulesFolder;
+    } else if (this.config.workspaceRootFolder) {
+      return this.config.workspaceRootFolder;
+    } else {
+      return path.join(this.config.lockfileFolder, 'node_modules');
```
arcanis: I'd like to avoid using `this.config.registry`, since its semantic isn't clear (cf #4109). However I can put `node_modules` in the `constants.js` file, if you prefer.

src/integrity-checker.js
BYK: >It will always iterate on all of actual (minus advance).

Not really. It keeps incrementing `advance` as long as there are non-matching so if the non-matching files are at the start of the array, as soon as we exhaust them (`maxDiff` iterations) we stop.

src/integrity-checker.js
BYK: It's not necessarily parallelizing. With this way, you tell the OS which files/dirs you want without an explicit order so it can optimize for look up times based on the actual byte-layout on the disk. Also it should be possible to parallelize I/O with the new SSDs and some not-so-new capabilities of the SATA NCQ.

src/integrity-checker.js
BYK: Until we actually act on #4109 we should keep things the same way so the clean up itself is easier too.

arcanis:
@BYK Code is now using a constant. Good to merge? :)

src/integrity-checker.js
```diff
@@ -118,23 +109,57 @@ export default class InstallationIntegrityChecker {
...
+      locations.push(this.config.modulesFolder);
+    } else {
+      locations.push(path.join(this.config.lockfileFolder, constants.NODE_MODULES_FOLDER));
+    }
+
+    if (workspaceLayout) {
+      for (const workspaceName of Object.keys(workspaceLayout.workspaces)) {
+        const loc = workspaceLayout.workspaces[workspaceName].loc;
+
+        if (loc) {
```
BYK: I think after my fixes, we don't need the emptiness check here anymore.

Referring to these changes:
 - https://github.com/yarnpkg/yarn/pull/4129/files#diff-ffb8cf10e4b2d16cc7df603574698af6R287
 - https://github.com/yarnpkg/yarn/pull/4131/files#diff-1ff3de58ebf278d554d14bcbe160d981R12

src/integrity-checker.js
```diff
@@ -171,16 +238,27 @@ export default class InstallationIntegrityChecker {
...
     }
 
-    Object.keys(lockfile).forEach(key => {
+    for (const key of Object.keys(lockfile)) {
       result.lockfileEntries[key] = lockfile[key].resolved || '';
-    });
+    }
+
+    for (const modulesFolder of this._getModulesFolders({workspaceLayout})) {
+      if (await fs.exists(modulesFolder)) {
```
BYK: I'm still not entirely getting how we may have missing folders in workspaces. Can you provide more details about what this is protecting us from?

Bnaya:
Thanks @arcanis @BYK :D

src/integrity-checker.js
arcanis: The `_getModulesFolders` function only returns the theoretical folders, it does not check that those folders actually exist on the disk. So for example, if we have a workspace named `foobar`, the function will return `foobar/node_modules` without checking that this folder actually exists.

PR #4124, opened 2017-08-08
Upgrade flow to ^0.52.0

wbinnssmith:
Summary: This upgrades flow, both in the flowconfig and in the flow-bin dependency to ^0.52.0

Test plan: `yarn lint`

*Note*: Should we reconsider using carets for flow versions? While flow is still < 1.0 and shouldn't introduce breaking changes in patch versions, @nmote has told me not to rely on this in the past.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -16,4 +16,4 @@ suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
 unsafe.enable_getters_and_setters=true
 
 [version]
-^0.49.1
+^0.52.0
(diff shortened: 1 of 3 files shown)


BYK:
>Note: Should we reconsider using carets for flow versions? While flow is still < 1.0 and shouldn't introduce breaking changes in patch versions, @nmote has told me not to rely on this in the past.

Good thinking. I think we're good for now but if it ever breaks, I'll go to "safer" ranges :)

PR #4131, opened 2017-08-09
Fix: invalid versions for workspace package

BYK:
**Summary**

When resolving workspace packages, we were always using the version `0.0.0` instead of the version defined in that package. This patch fixes it to use the correct version available from the package.json file inside.

**Test plan**

Existing tests.

### DIFF
--- a/src/fetchers/workspace-fetcher.js
+++ b/src/fetchers/workspace-fetcher.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import type {PackageRemote, FetchedMetadata, Manifest} from '../types.js';
+import type {PackageRemote, FetchedMetadata} from '../types.js';
 import type Config from '../config.js';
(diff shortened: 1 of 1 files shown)


src/fetchers/workspace-fetcher.js
```diff
@@ -39,8 +39,7 @@ export default class WorkspaceFetcher {
       dest: this.dest,
       package: {
         ...pkg,
-        _uid: '',
-        version: '0.0.0',
+        _uid: (pkg && pkg.version) || '',
```
arcanis: `pkg.version` should be enough - workspaces without versions are invalid and [ignored](https://github.com/yarnpkg/yarn/blob/master/src/config.js#L632-L635)

BYK: I needed to do another change to be able to remove the checks. Please review again.

PR #4132, opened 2017-08-09
Fix: prevent hoister from adding an ancestor when there is none

BYK:
**Summary**

Just fixes a small bug in the hoisting code where it pushed `undefined` to the ancestor list when traversing the-level dependencies.

**Test plan**

N/A

### DIFF
--- a/src/package-hoister.js
+++ b/src/package-hoister.js
@@ -411,14 +411,13 @@ export default class PackageHoister {
    */
 
   hoist(info: HoistManifest) {
-    const {key, parts: rawParts} = info;
+    const {key: oldKey, parts: rawParts} = info;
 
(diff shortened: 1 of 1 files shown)


src/package-hoister.js
```diff
@@ -428,7 +427,7 @@ export default class PackageHoister {
     // update to the new key
     if (oldKey === newKey) {
       info.addHistory("Didn't hoist - conflicts above");
-      this.setKey(info, oldKey, parts);
+      this.setKey(info, oldKey, rawParts);
```
arcanis: Is it intended? This changes the behavior

BYK: It is and it corrects the behavior actually since the intention was to re add the removed one to cause no changes if there are conflicts. Right?

PR #4135, opened 2017-08-10
fix(run): improve escaping for script arguments

rhendric:
**Summary**

Extra command-line arguments to scripts were not being escaped correctly. This patch adds robust shell quoting logic for both Windows and Linux/macOS.

**Test plan**

On *nix, create a `package.json` containing `"scripts":{"echo":"echo"}`. Run `yarn run -s echo -- '$X \"blah\"'`. Expect to observe ` \blah\` prior to this patch, and `$X \"blah\"` after it.

Testing on Windows should be similar, but may require fancier escaping to get the arguments into yarn in the first place. (I don't have access to a Windows box to verify the exact procedure to follow, sorry—but I did confirm that my automated tests succeed in AppVeyor.)

### DIFF
--- a/__tests__/commands/run.js
+++ b/__tests__/commands/run.js
@@ -64,7 +64,7 @@ test('properly handles extra arguments and pre/post scripts', (): Promise<void> 
     const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
     const poststart = ['poststart', config, pkg.scripts.poststart, config.cwd];
     const prestart = ['prestart', config, pkg.scripts.prestart, config.cwd];
-    const start = ['start', config, pkg.scripts.start + ' "--hello"', config.cwd];
+    const start = ['start', config, pkg.scripts.start + ' --hello', config.cwd];
 
     expect(execCommand.mock.calls[0]).toEqual(prestart);
     expect(execCommand.mock.calls[1]).toEqual(start);
@@ -75,7 +75,7 @@ test('properly handles extra arguments and pre/post scripts', (): Promise<void> 
 test('properly handle bin scripts', (): Promise<void> => {
   return runRun(['cat-names'], {}, 'bin', config => {
     const script = path.join(config.cwd, 'node_modules', '.bin', 'cat-names');
-    const args = ['cat-names', config, `"${script}"`, config.cwd];
+    const args = ['cat-names', config, script, config.cwd];
 
     expect(execCommand).toBeCalledWith(...args);
   });
@@ -114,19 +114,21 @@ test('properly handle env command', (): Promise<void> => {
   });
 });
 
-test('retains string delimiters if args have spaces', (): Promise<void> => {
+test('adds string delimiters if args have spaces', (): Promise<void> => {
   return runRun(['cat-names', '--filter', 'cat names'], {}, 'bin', config => {
     const script = path.join(config.cwd, 'node_modules', '.bin', 'cat-names');
-    const args = ['cat-names', config, `"${script}" "--filter" "cat names"`, config.cwd];
+    const q = process.platform === 'win32' ? '"' : "'";
+    const args = ['cat-names', config, `${script} --filter ${q}cat names${q}`, config.cwd];
 
     expect(execCommand).toBeCalledWith(...args);
   });
 });
 
-test('retains quotes if args have spaces and quotes', (): Promise<void> => {
+test('adds quotes if args have spaces and quotes', (): Promise<void> => {
   return runRun(['cat-names', '--filter', '"cat names"'], {}, 'bin', config => {
     const script = path.join(config.cwd, 'node_modules', '.bin', 'cat-names');
-    const args = ['cat-names', config, `"${script}" "--filter" "\\"cat names\\""`, config.cwd];
+    const quotedCatNames = process.platform === 'win32' ? '^"\\^"cat^ names\\^"^"' : `'"cat names"'`;
+    const args = ['cat-names', config, `${script} --filter ${quotedCatNames}`, config.cwd];
 
     expect(execCommand).toBeCalledWith(...args);
   });
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -10,11 +10,7 @@ import map from '../../util/map.js';
 
 const leven = require('leven');
 const path = require('path');
-
-// Copied from https://github.com/npm/npm/blob/63f153c743f9354376bfb9dad42bd028a320fd1f/lib/run-script.js#L175
-function joinArgs(args: Array<string>): string {
-  return args.reduce((joinedArgs, arg) => joinedArgs + ' "' + arg.replace(/"/g, '\\"') + '"', '');
-}
+const {quoteForShell, sh, unquoted} = require('puka');
 
 export function setFlags(commander: Object) {}
 
@@ -35,7 +31,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       if (await fs.exists(binFolder)) {
         for (const name of await fs.readdir(binFolder)) {
           binCommands.push(name);
-          scripts[name] = `"${path.join(binFolder, name)}"`;
+          scripts[name] = quoteForShell(path.join(binFolder, name));
         }
       }
       visitedBinFolders.add(binFolder);
@@ -82,8 +78,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       process.env.YARN_SILENT = '1';
       for (const [stage, cmd] of cmds) {
         // only tack on trailing arguments for default script, ignore for pre and post - #1595
-        const defaultScriptCmd = cmd + joinArgs(args);
-        const cmdWithArgs = stage === action ? defaultScriptCmd : cmd;
+        const cmdWithArgs = stage === action ? sh`${unquoted(cmd)} ${args}` : cmd;
         await execCommand(stage, config, cmdWithArgs, config.cwd);
       }
     } else if (action === 'env') {
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -2,9 +2,9 @@
 /* eslint max-len: 0 */
 
 import execa from 'execa';
+import {sh} from 'puka';
 import makeTemp from './_temp.js';
 import * as fs from '../src/util/fs.js';
-import * as misc from '../src/util/misc.js';
 import * as constants from '../src/constants.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000;
@@ -76,7 +76,7 @@ async function runYarn(args: Array<string> = [], options: Object = {}): Promise<
     options['extendEnv'] = false;
   }
   options['env']['FORCE_COLOR'] = 0;
-  const {stdout, stderr} = await execa(path.resolve(__dirname, '../bin/yarn'), args, options);
+  const {stdout, stderr} = await execa.shell(sh`${path.resolve(__dirname, '../bin/yarn')} ${args}`, options);
 
   return [stdout, stderr];
 }
@@ -211,9 +211,8 @@ test('yarnrc binary path (executable)', async () => {
   expect(stdoutOutput.toString().trim()).toEqual('override called');
 });
 
-// Windows could run these tests, but we currently suffer from an escaping issue that breaks them (#4135)
-if (process.platform !== 'win32') {
-  test('yarn run <script> --opt', async () => {
+for (const withDoubleDash of [false, true]) {
+  test(`yarn run <script> ${withDoubleDash ? '-- ' : ''}--opt`, async () => {
     const cwd = await makeTemp();
 
     await fs.writeFile(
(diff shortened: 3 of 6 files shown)


src/cli/commands/run.js
```diff
@@ -11,9 +11,22 @@ import map from '../../util/map.js';
 const leven = require('leven');
 const path = require('path');
 
-// Copied from https://github.com/npm/npm/blob/63f153c743f9354376bfb9dad42bd028a320fd1f/lib/run-script.js#L175
+function quoteForCmd(arg: string): string {
+  // See the below blog post for an explanation of what's going on here:
+  // eslint-disable-next-line max-len
+  // https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/
```
BYK: Hah! I'm glad there is another one who read the same article 🗡 

I think you should release this as a library so `yarn` and others can use it from there because I've yet to see a proper Windows-compatible shell escaping library on NPM.

src/cli/commands/run.js
```diff
@@ -11,9 +11,22 @@ import map from '../../util/map.js';
...
+  const caretEscape = /"/.test(arg);
+  arg = `"${arg.replace(/\\+(?=$|")/g, '$&$&').replace(/"/g, '\\"')}"`;
+  if (caretEscape) {
+    arg = arg.replace(/[()%!^"<>&|]/g, '^$&');
+  }
+  return arg;
+}
+
+const quoteForShell = process.platform === 'win32' ? quoteForCmd : arg => `'${arg.replace(/'/g, "'\\''")}'`;
```
BYK: May be we should have the following in `constants.js`:
```
export const IS_WINDOWS = process.platform === 'win32';
```

What do you think?

src/cli/commands/run.js
```diff
@@ -11,9 +11,22 @@ import map from '../../util/map.js';
...
+
 function joinArgs(args: Array<string>): string {
-  return args.reduce((joinedArgs, arg) => joinedArgs + ' "' + arg.replace(/"/g, '\\"') + '"', '');
+  return args.length ? ' ' + args.map(quoteForShell).join(' ') : '';
```
BYK: I don't think you need the `args.length ?` check. `[].map(whatever).join(' ')` still returns `''`.

__tests__/fixtures/run/escape-args/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "scripts": {
+    "write-args": "node write-args.js"
```
BYK: Can do `node -p -e "process.argv[2], process.argv.slice(3).join(' ')"` instead and then read the output from stdout. I'd prefer not to write to file system if/when possible.

__tests__/commands/run.js
```diff
@@ -13,6 +13,7 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 const execCommand: $FlowFixMe = require('../../src/util/execute-lifecycle-script').execCommand;
 
 const path = require('path');
+const q = process.platform === 'win32' ? '"' : "'";
```
BYK: Please rename this to `shellQuotes` or something. `q` is a very ambiguous variable name.

__tests__/commands/run-integration.js
```diff
@@ -0,0 +1,57 @@
...
+        cases.push(a + a + b + b);
+        cases.push(a + b + a + b);
+        cases.push(a + b + b + a);
+        cases.push(a + a + a + b);
+        cases.push(a + a + b + a);
+        cases.push(a + b + a + a);
+      }
+    }
+  }
+  cases.sort(() => Math.random() - 0.5);
```
BYK: This is a very bad way to randomize an array. Also, why do we need randomization anyways?

__tests__/commands/run-integration.js
```diff
@@ -0,0 +1,57 @@
...
+import {run} from '../../src/cli/commands/run.js';
+
+const path = require('path');
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'run');
+const runRun = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, config, reporter): Promise<void> => {
+  return run(config, reporter, flags, args);
+});
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
```
BYK: Why?

__tests__/commands/run-integration.js
```diff
@@ -0,0 +1,57 @@
...
+        cases.push(a + a + a + b);
+        cases.push(a + a + b + a);
+        cases.push(a + b + a + a);
+      }
+    }
+  }
+  cases.sort(() => Math.random() - 0.5);
+  const tempDir = await makeTemp();
+  const prefix = path.resolve(__dirname, '../..');
+  const origPrefix = process.env.PREFIX;
```
BYK: Don't modify `process.env` directly, please. There should be a way to pass this to `runRun`.

src/cli/commands/run.js
BYK: Looks like there's a bit more to `bash` escaping too: https://github.com/xxorax/node-shell-escape/blob/master/shell-escape.js

src/cli/commands/run.js
rhendric: I completely agree; I started writing that library this morning. I was a little shy about writing it first and then proposing that yarn take a dependency on a 0.0.1 library that nobody's used yet.

src/cli/commands/run.js
rhendric: Sounds reasonable to me. There are a few instances of that check throughout the code; I looked to make sure there wasn't some other pattern already in place that I should follow.

If I understand correctly, I think those extra replaces in node-shell-escape are just cosmetic; they remove some redundant quotes but don't affect the final interpreted string. I'll probably add something like that though if I do go and release this as a library.

src/cli/commands/run.js
rhendric: Yes, but the entire function would return the `' '` that gets prepended to the joined string.

__tests__/fixtures/run/escape-args/package.json
rhendric: Good suggestion. I'll give it a try.

__tests__/commands/run.js
rhendric: I'm with you in general, but in this case, this guy gets interpolated several times into several strings below, and having it be a short name really helps readability in my opinion. Compare:
```js
const args = ['cat-names', config, `${q}${script}${q} ${q}--filter${q} ${q}cat names${q}`, config.cwd];

const args = [
  'cat-names',
  config,
  `${shellQuote}${script}${shellQuote} ${shellQuote}--filter${shellQuote} ${shellQuote}cat names${shellQuote}`,
  config.cwd,
];
```

Easy enough to change if you still disagree though.

__tests__/commands/run-integration.js
rhendric: Ugh. Something is happening here that I don't understand.

Originally I wrote this test to invoke the script serially, and that was consistently timing out on AppVeyor, even with a much higher `DEFAULT_TIMEOUT_INTERVAL` (2.5 minutes!), and even with only three iterations of the script invocation. Once I parallelized, the tests stopped timing out and in fact finished much faster than timeout/degree of parallelism, so there was probably something I was doing wrong with the serial implementation (I was `await`ing on all the right things, I thought!), but I was nervous that there was a chance that Windows just wanted to take its time spawning child processes on that build machine sometimes. So I took this down a little but it is still higher than the current test times seem to justify. What's your general sense here; should I try to keep the value conservatively high to reduce the chance of random build failure, or lower to try to crack down on build time?

__tests__/commands/run-integration.js
rhendric: It's bad if true randomness is required, but it's good enough for a quick scramble (would you prefer I implement/`require` a Knuth shuffle just for this test?). I'm randomizing because I'm not actually testing every combination; I didn't want to take too long on this test but I also wanted to get a decent sampling of the space. Basically it's a poor man's property-based test. If you think it's a good idea, I'd be happy to pull in jsverify or jssmartcheck and write a real property-based test. Alternatively, if this is a library, the heavy property checking can go there and yarn can test a much smaller sampling of cases.

__tests__/commands/run-integration.js
rhendric: Can you give me more of a hint? I looked through the code for such a way and couldn't find a straightforward one.

src/cli/commands/run.js
BYK: Well then, add lots of tests and have a high coverage and mark it as 1.0 ;)

src/cli/commands/run.js
BYK: >Sounds reasonable to me. There are a few instances of that check throughout the code; I looked to make sure there wasn't some other pattern already in place that I should follow.

On second thought, I think that change deserves its own diff so I'll work on it. For now, your diff is fine. :)

>If I understand correctly, I think those extra replaces in node-shell-escape are just cosmetic; they remove some redundant quotes but don't affect the final interpreted string. I'll probably add something like that though if I do go and release this as a library.

Fine by me if you say they are safe. It's frightening for me to realize that I know more about Windows escaping than Bash escaping at this point :D

src/cli/commands/run.js
BYK: Oh, sorry I missed that part! That said, then the name of this function should be `stringifyArgs` or `getArgsString` to convey the meaning better.

Optionally, you can pass in the first part and name the function `serializeShellCommand` and do something like:
```
function serializeShellCommand(cmd: string, args: Array<string>): string {
    return [cmd].concat(args.map(quoteForShell)).join(' ');
}
```

That would take care all of it. Anyways, I'm just rambling. Just changing the name of the function would be enough for me but I'm still a bit concerned that people may miss that extra space at the beginning or don't understand the reason at first glance.

__tests__/commands/run.js
BYK: Well, let's agree to add a comment above it stating this then and call it a day? :)

__tests__/commands/run-integration.js
BYK: Well, now I see. Actually, I think you can benefit from using `test.concurrent` and splitting your tests into individual test cases, generated programmatically, wrap them in a `describe` block so you can use `beforeEach` hooks and then `jest` should take care of it. If even after that and not writing to the file system (remember, we agreed to try using stdout instead above?) I'm okay leaving the timeout higher.

__tests__/commands/run-integration.js
BYK: Any non-determinism in tests would create doubt when running things on the CI: did it fail for reals or was that a random blip? Did I catch a real edge case thanks to randomness, or the CI had a bad network connection? So let's have this test all possible combinations instead of random sampling for each test run or agree to ahve the same random sample for each run.

__tests__/commands/run-integration.js
BYK: Search for `getGlobalPrefix` in the code. Spoiler: you can use `flags` or `config` to override the prefix.

__tests__/fixtures/run/escape-args/package.json
BYK: Turns out it should just be `node -p "..."` without the `-e` part.

BYK:
Hey, @rhendric! Long time no see. Any updates?

rhendric:
Hi! Sorry for the delay; I still expect to finish the swing on this but I think it'll be another week or so before I have the time.

BYK:
No worries, just wanted to check if you're still working on it :)

rhendric:
Hi, I'm back on the case! Quick question: I see that there is now [a comment in \_\_tests\_\_/integration.js](/yarnpkg/yarn/blob/75bc450/__tests__/integration.js#L214) linking back to this PR, claiming that this is blocking those tests from being enabled on Windows, but I don't see the connection—those tests look like they don't use any characters that are likely to cause problems. Looks like it was added in #4152. @arcanis, or anyone, what's the story with that? Should I try to remove that gating `if` as part of this PR?

rhendric:
I figured out what the deal with those tests is. It's a combination of the below:
* the tests use native `echo`,
* `echo` on Windows does not strip quotation marks from its arguments (in Windows, unlike in Unix-based OSes, the choice of how to handle command line argument quotation is left to the executable being invoked, not to the invoking shell—strictly speaking, all commands in Windows receive a single command line instead of an array of command line arguments), and
* Yarn puts quotes around all of its arguments whether they need them or not.

I see three ways to get these tests passing, the first of which I'm going to go with unless I hear objections:
1. Make Yarn not put quotes around arguments unless necessary (my library handles this, and I've put it through quite a series of torture tests). This has the advantage of being less likely to surprise the next Yarn user who tries to use `echo` in a script. This is also perhaps the intent indicated by leaving these tests disabled until my work here is done, because either of the below things could have been done earlier.
1. Replace `echo` in these tests with something else: `node -p`, perhaps, or add a dependency on [`echo-cli`](https://www.npmjs.com/package/echo-cli).
1. Just make the tests expect to see `'"--opt"'` on Windows instead of `'--opt'`.

rhendric:
@BYK, re-review please?

BYK:
@rhendric thank you so much for this!

PR #4143, opened 2017-08-10
Fallbacks to /tmp if the preferred cache folder isn't writable

### LINKED ISSUE #4141: Fix: yarn should check if the cache folder is usable before committing to it

Bug

Yarn directly assumes it can use the cache folder and goes ahead to create it. The thing is, sometimes even if it can create the folder, it may not be able to write into it or read from it. Even more, the creation is also not guaranteed and there are no safe guards for this. Thus, if for some reason this fails, yarn falls flat on its face.

Try running `yarn --cahe-folder=/root/.yarn-cache` as a normal user for instance.

Yarn should detect the issue right from the start and send out a clear message. Even better: print a warning about the cache folder not being usable and falling back to the system-provided tmp folder so things work properly.

Latest master.

arcanis:
**Summary**

Fixes #4141. If `$(yarn cache dir)` isn't writable, Yarn will crash and burn with EACCESS / EPERM. This patch adds a fallback mechanism so that we try different folders until we find one that fits. As a first step, I've added `/tmp/.yarn-cache`.

**Test plan**

Tested with

```
chmod 000 /Users/mael/Library/Caches/Yarn && ~/yarn/bin/yarn
```

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -70,3 +70,20 @@ test('--mutex network', async () => {
     execa(command, ['add', 'foo'].concat(args), options),
   ]);
 });
+
+test('cache folder fallback', async () => {
+  const cwd = await makeTemp();
+  const cacheFolder = path.join(cwd, '.cache');
+
+  await fs.mkdirp(cacheFolder);
+  await fs.chmod(cacheFolder, 0o000);
+
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const args = ['--preferred-cache-folder', cacheFolder];
+
+  const options = {cwd};
+
+  await Promise.all([
+    execa(command, ['cache', 'dir'].concat(args), options),
+  ]);
+});
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -77,38 +77,30 @@ test('cache folder fallback', async () => {
   const cwd = await makeTemp();
   const cacheFolder = path.join(cwd, '.cache');
 
-  await fs.mkdirp(cacheFolder);
-
   const command = path.resolve(__dirname, '../bin/yarn');
   const args = ['--preferred-cache-folder', cacheFolder];
 
   const options = {cwd};
 
-  {
+  function runCacheDir(): Promise<Array<Buffer>> {
     const {stderr, stdout} = execa(command, ['cache', 'dir'].concat(args), options);
 
     const stdoutPromise = misc.consumeStream(stdout);
     const stderrPromise = misc.consumeStream(stderr);
 
-    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
-
-    expect(stdoutOutput.toString().trim()).toEqual(path.join(cacheFolder, `v${constants.CACHE_VERSION}`));
-    expect(stderrOutput.toString()).not.toMatch(/Skipping preferred cache folder/);
+    return Promise.all([stdoutPromise, stderrPromise]);
   }
 
-  await fs.chmod(cacheFolder, 0o000);
+  const [stdoutOutput, stderrOutput] = await runCacheDir();
 
-  {
-    const {stderr, stdout} = execa(command, ['cache', 'dir'].concat(args), options);
+  expect(stdoutOutput.toString().trim()).toEqual(path.join(cacheFolder, `v${constants.CACHE_VERSION}`));
+  expect(stderrOutput.toString()).not.toMatch(/Skipping preferred cache folder/);
 
-    const stdoutPromise = misc.consumeStream(stdout);
-    const stderrPromise = misc.consumeStream(stderr);
+  await fs.unlink(cacheFolder);
+  await fs.writeFile(cacheFolder, `not a directory`);
 
-    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+  const [stdoutOutput2, stderrOutput2] = await runCacheDir();
 
-    expect(stdoutOutput.toString().trim()).toEqual(
-      path.join(constants.PREFERRED_MODULE_CACHE_DIRECTORIES[0], `v${constants.CACHE_VERSION}`),
-    );
-    expect(stderrOutput.toString()).toMatch(/Skipping preferred cache folder/);
-  }
+  expect(stdoutOutput2.toString().trim()).toEqual(path.join(constants.PREFERRED_MODULE_CACHE_DIRECTORIES[0], `v${constants.CACHE_VERSION}`));
+  expect(stderrOutput2.toString()).toMatch(/Skipping preferred cache folder/);
 });
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -83,7 +83,5 @@ test('cache folder fallback', async () => {
 
   const options = {cwd};
 
-  await Promise.all([
-    execa(command, ['cache', 'dir'].concat(args), options),
-  ]);
+  await Promise.all([execa(command, ['cache', 'dir'].concat(args), options)]);
 });
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -101,6 +101,8 @@ test('cache folder fallback', async () => {
 
   const [stdoutOutput2, stderrOutput2] = await runCacheDir();
 
-  expect(stdoutOutput2.toString().trim()).toEqual(path.join(constants.PREFERRED_MODULE_CACHE_DIRECTORIES[0], `v${constants.CACHE_VERSION}`));
+  expect(stdoutOutput2.toString().trim()).toEqual(
+    path.join(constants.PREFERRED_MODULE_CACHE_DIRECTORIES[0], `v${constants.CACHE_VERSION}`),
+  );
   expect(stderrOutput2.toString()).toMatch(/Skipping preferred cache folder/);
 });
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -4,6 +4,8 @@
 import execa from 'execa';
 import makeTemp from './_temp.js';
 import * as fs from '../src/util/fs.js';
+import * as misc from '../src/util/misc.js';
+import * as constants from '../src/constants.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 
@@ -76,12 +78,37 @@ test('cache folder fallback', async () => {
   const cacheFolder = path.join(cwd, '.cache');
 
   await fs.mkdirp(cacheFolder);
-  await fs.chmod(cacheFolder, 0o000);
 
   const command = path.resolve(__dirname, '../bin/yarn');
   const args = ['--preferred-cache-folder', cacheFolder];
 
   const options = {cwd};
 
-  await Promise.all([execa(command, ['cache', 'dir'].concat(args), options)]);
+  {
+    const {stderr, stdout} = execa(command, ['cache', 'dir'].concat(args), options);
+
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+
+    expect(stdoutOutput.toString().trim()).toEqual(path.join(cacheFolder, `v${constants.CACHE_VERSION}`));
+    expect(stderrOutput.toString()).not.toMatch(/Skipping preferred cache folder/);
+  }
+
+  await fs.chmod(cacheFolder, 0o000);
+
+  {
+    const {stderr, stdout} = execa(command, ['cache', 'dir'].concat(args), options);
+
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+
+    expect(stdoutOutput.toString().trim()).toEqual(
+      path.join(constants.PREFERRED_MODULE_CACHE_DIRECTORIES[0], `v${constants.CACHE_VERSION}`),
+    );
+    expect(stderrOutput.toString()).toMatch(/Skipping preferred cache folder/);
+  }
 });
--- a/src/cli/commands/cache.js
+++ b/src/cli/commands/cache.js
@@ -55,7 +55,7 @@ export const {run, setFlags, examples} = buildSubCommands('cache', {
   },
 
   dir(config: Config, reporter: Reporter) {
-    reporter.log(config.cacheFolder);
+    reporter.log(config.cacheFolder, {force: true});
   },
 
   async clean(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -61,7 +61,7 @@ export function main({
     'rather than installing modules into the node_modules folder relative to the cwd, output them here',
   );
   commander.option('--preferred-cache-folder <path>', 'specify a custom folder to store the yarn cache if possible');
-  commander.option('--cache-folder <path>', 'specify a custom folder to store the yarn cache');
+  commander.option('--cache-folder <path>', 'specify a custom folder that must be used to store the yarn cache');
   commander.option('--mutex <type>[:specifier]', 'use a mutex to ensure only one yarn instance is executing');
   commander.option('--emoji [bool]', 'enable emoji in output', process.platform === 'darwin');
   commander.option('-s, --silent', 'skip Yarn console logs, other types of logs (script output) will be printed');
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -60,6 +60,7 @@ export function main({
     '--modules-folder <path>',
     'rather than installing modules into the node_modules folder relative to the cwd, output them here',
   );
+  commander.option('--preferred-cache-folder <path>', 'specify a custom folder to store the yarn cache if possible');
   commander.option('--cache-folder <path>', 'specify a custom folder to store the yarn cache');
   commander.option('--mutex <type>[:specifier]', 'use a mutex to ensure only one yarn instance is executing');
   commander.option('--emoji [bool]', 'enable emoji in output', process.platform === 'darwin');
@@ -324,6 +325,7 @@ export function main({
       binLinks: commander.binLinks,
       modulesFolder: commander.modulesFolder,
       globalFolder: commander.globalFolder,
+      preferredCacheFolder: commander.preferredCacheFolder,
       cacheFolder: commander.cacheFolder,
       preferOffline: commander.preferOffline,
       captureHar: commander.har,
--- a/src/config.js
+++ b/src/config.js
@@ -288,7 +288,7 @@ export default class Config {
       }
 
       for (let t = 0; t < preferredCacheFolders.length && !cacheRootFolder; ++t) {
-        const tentativeCacheFolder = preferredCacheFolders[t];
+        const tentativeCacheFolder = String(preferredCacheFolders[t]);
 
         try {
           await fs.mkdirp(tentativeCacheFolder);
--- a/src/config.js
+++ b/src/config.js
@@ -291,7 +291,6 @@ export default class Config {
         const tentativeCacheFolder = String(preferredCacheFolders[t]);
 
         try {
-
           await fs.mkdirp(tentativeCacheFolder);
 
           const testFile = path.join(tentativeCacheFolder, 'testfile');
@@ -302,7 +301,6 @@ export default class Config {
           await fs.unlink(testFile);
 
           cacheRootFolder = tentativeCacheFolder;
-
         } catch (error) {
           this.reporter.warn(this.reporter.lang('cacheFolderSkipped', tentativeCacheFolder));
         }
--- a/src/config.js
+++ b/src/config.js
@@ -279,15 +279,29 @@ export default class Config {
 
     let cacheRootFolder = opts.cacheFolder || this.getOption('cache-folder', true);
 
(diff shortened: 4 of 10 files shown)


src/config.js
```diff
@@ -276,9 +276,27 @@ export default class Config {
...
-    );
+
+    let cacheRootFolder = opts.cacheFolder || this.getOption('cache-folder', true);
+
+    for (let t = 0; t < constants.PREFERRED_MODULE_CACHE_DIRECTORIES.length && !cacheRootFolder; ++t) {
+      const tentativeCacheFolder = constants.PREFERRED_MODULE_CACHE_DIRECTORIES[t];
+
+      try {
+        await fs.mkdirp(tentativeCacheFolder);
+        await fs.writeFile(path.join(tentativeCacheFolder, constants.WTEST_FILENAME), `testing write access`);
```
BYK: I think we should use [fs.access](https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback) with both read and write checks.

src/config.js
```diff
@@ -276,9 +276,27 @@ export default class Config {
...
-      opts.cacheFolder || this.getOption('cache-folder', true) || constants.MODULE_CACHE_DIRECTORY,
-    );
+
+    let cacheRootFolder = opts.cacheFolder || this.getOption('cache-folder', true);
+
+    for (let t = 0; t < constants.PREFERRED_MODULE_CACHE_DIRECTORIES.length && !cacheRootFolder; ++t) {
+      const tentativeCacheFolder = constants.PREFERRED_MODULE_CACHE_DIRECTORIES[t];
+
+      try {
+        await fs.mkdirp(tentativeCacheFolder);
```
BYK: I'd separate this part out and try cleaning up the cache directory if we created it but it is not usable. Kind of an edge case tho, so not feeling too strongly about this.

src/config.js
```diff
@@ -276,9 +276,27 @@ export default class Config {
...
+        cacheRootFolder = tentativeCacheFolder;
+      } catch (error) {
+        this.reporter.warn(this.reporter.lang('cacheFolderSkipped', tentativeCacheFolder));
+      }
+    }
+
+    if (!cacheRootFolder) {
+      throw new MessageError(this.reporter.lang('cacheFolderMissing'));
+    } else {
+      this._cacheRootFolder = String(cacheRootFolder);
```
BYK: Why do you need the `String()` casting?

src/constants.js
```diff
@@ -46,15 +46,21 @@ function getDirectory(category: string): string {
   return path.join(userHome, `.${category}`, 'yarn');
 }
 
-function getCacheDirectory(): string {
+function getPreferredCacheDirectories(): Array<string> {
+  const preferredCacheDirectories = [getDirectory('cache')];
+
+  if (process.platform !== 'win32') {
+    preferredCacheDirectories.unshift('/tmp/.yarn-cache');
```
BYK: Nope. Use [os.tmpdir()](https://nodejs.org/api/os.html#os_os_tmpdir) please. Note that this is also platform independent.

src/constants.js
```diff
@@ -46,15 +46,21 @@ function getDirectory(category: string): string {
...
+  }
+
   if (process.platform === 'darwin') {
-    return path.join(userHome, 'Library', 'Caches', 'Yarn');
+    preferredCacheDirectories.unshift(path.join(userHome, 'Library', 'Caches', 'Yarn'));
```
BYK: This kind of changes our previous logic so not sure about that. We should directly fallback to `tmp` when our first choice fails. With this one, it will try this on OS X and then try the Linux version second.

src/config.js
```diff
@@ -276,9 +276,27 @@ export default class Config {
       networkConcurrency: this.networkConcurrency,
       networkTimeout: this.networkTimeout,
     });
-    this._cacheRootFolder = String(
-      opts.cacheFolder || this.getOption('cache-folder', true) || constants.MODULE_CACHE_DIRECTORY,
-    );
+
+    let cacheRootFolder = opts.cacheFolder || this.getOption('cache-folder', true);
```
BYK: This should be pushed to the top of the list we try and we should fall back to `tmp` at all times for resilliency. Right now this override seems to be ignored.

BYK:
We also need automated tests for this.

src/config.js
arcanis: I disagree - if the explicitely specified cache folder cannot be used, that's an hard error. This option contract forces us to use this folder. By (almost) silently using a different directory, we might end up with subtle bugs. For example, two Yarn process might run concurrently with different caches. In such a scenario, if for some reason their respective cache could not be reached, they would end up corrupting themselves.

src/config.js
arcanis: Was there before. Probably because of Flow, I'd guess :)

src/config.js
arcanis: Thinking of it, I think this would still have issues (since the cache folders have to be assumed empty, we would have to make the check on the directory itself, which might yield different results than what we would have add on a file). However, I'll also add a read on the file to account for the read permissions. Does that sounds ok?

arcanis: After discussing this with @BYK we came to the conclusion that `--cache-folder` would keep its current semantic (and would crash if the directory is not writable) since it would break the commonly assumed contract, potentially exposing the application to vulnerabilities (the `/tmp` folder could be world-readable whereas the `--cache-folder` wouldn't). However, we'll also add a `--preferred-cache-folder` option with less restrictions. Will even make it easier to test the feature :)

BYK: >Thinking of it, I think this would still have issues (since the cache folders have to be assumed empty, we would have to make the check on the directory itself, which might yield different results than what we would have add on a file). However, I'll also add a read on the file to account for the read permissions. Does that sounds ok?

I don't see how this can be the case? If you can create a file in the directory it should be equivalent to `fs.access` with write permissions. Same goes for read. We can make the case for space exhaustion in the cache directory (if it is on a different drive etc.) but this doesn't protect us against that anyways.

src/config.js
BYK: May be we can remove it now?

src/config.js
BYK: For the readers of this thread, if there are any: we have talked about getting a unique folder once we fallback to `tmp` like by using a timestamp or the PID of the process, or both, but then decided to do this: https://github.com/yarnpkg/yarn/pull/4143#discussion_r132509455

We should still probably try to ensure a unique folder in `tmp` when we fallback for safety.

src/config.js
```diff
@@ -276,9 +276,27 @@ export default class Config {
...
+
+    for (let t = 0; t < constants.PREFERRED_MODULE_CACHE_DIRECTORIES.length && !cacheRootFolder; ++t) {
```
BYK: `&& !cacheRootFolder` part totally gets lost in the code. Please make this an explicit `if` statement for clarity.

Also half-joking, half-serious: why are we using `t` as the counter and not `i`? 🗡

src/config.js
arcanis: > If you can create a file in the directory it should be equivalent to fs.access with write permissions.

Yes but the file we've created might not be accessible. Umask, etc.

src/config.js
arcanis: Good question, but I don't really know 😆 I've been using it for single-level loop since I started programming (and `u`/`v`/`w` or `x`/`y`/`z` when multiple loops are combined).

Could be amusing to run a search on the github repos to find what variable name is preferred by developpers :)

arcanis: Looking on stack exchange, it seems that literally noone else is using `t` as a variable name ... I wonder how I came to that. Meh ¯\(ツ)/¯

src/config.js
BYK: I really think we should just use `fs.access` here.

src/config.js
arcanis: Just tried, still fails.

src/config.js
```diff
@@ -276,9 +276,41 @@ export default class Config {
...
+        preferredCacheFolders = [preferredCacheFolder].concat(preferredCacheFolders);
+      }
+
+      for (let t = 0; t < preferredCacheFolders.length && !cacheRootFolder; ++t) {
+        const tentativeCacheFolder = preferredCacheFolders[t];
+
+        try {
+          await fs.mkdirp(tentativeCacheFolder);
+          // eslint-disable-next-line
+          await fs.access(tentativeCacheFolder, fs.constants.R_OK | fs.constants.W_OK | fs.constants.X_OK);
```
BYK: Why do we need to check for execution? (`fs.constants.X_OK`)

src/reporters/lang/en.js
```diff
@@ -174,6 +174,11 @@ const messages = {
   workspaceNameMandatory: 'Missing name in workspace at $0, ignoring.',
   workspaceNameDuplicate: 'There are more than one workspace with name $0',
 
+  cacheFolderSkipped: 'Skipping preferred cache folder $0 because it is not writable.',
+  cacheFolderMissing:
+    "Yarn hasn't been able to find a cache folder. Please use an explicit --cache-folder option to tell it what location to use, or make one of the preferred locations writable.",
```
BYK: >Yarn hasn't been able to find a cache folder.

find a cache folder it can use

>Please use an explicit --cache-folder option

use *the* explicit --cache-folder option

src/reporters/lang/en.js
```diff
@@ -174,6 +174,11 @@ const messages = {
...
+  cacheFolderSelected: 'Selected the next writable cache folder in the list, will be $0.',
```
BYK: I think it is better to make this a verbose message and say "Using cache folder: $0".

src/constants.js
```diff
@@ -46,15 +47,24 @@ function getDirectory(category: string): string {
...
   if (process.platform === 'darwin') {
-    return path.join(userHome, 'Library', 'Caches', 'Yarn');
+    preferredCacheDirectories.push(path.join(userHome, 'Library', 'Caches', 'Yarn'));
+  }
+
+  preferredCacheDirectories.push(getDirectory('cache'));
+
+  if (process.platform !== 'win32') {
+    preferredCacheDirectories.push(path.join(os.tmpdir(), '.yarn-cache'));
+    preferredCacheDirectories.push('/tmp/.yarn-cache');
```
BYK: You don't need this anymore.

src/constants.js
```diff
@@ -46,15 +47,24 @@ function getDirectory(category: string): string {
...
+
   if (process.platform === 'darwin') {
-    return path.join(userHome, 'Library', 'Caches', 'Yarn');
+    preferredCacheDirectories.push(path.join(userHome, 'Library', 'Caches', 'Yarn'));
+  }
+
+  preferredCacheDirectories.push(getDirectory('cache'));
+
+  if (process.platform !== 'win32') {
+    preferredCacheDirectories.push(path.join(os.tmpdir(), '.yarn-cache'));
```
BYK: Why omit Windows??

src/util/fs.js
```diff
@@ -13,6 +13,12 @@ const globModule = require('glob');
 const os = require('os');
 const path = require('path');
 
+export const constants = typeof fs.constants !== 'undefined' ? fs.constants : {
```
BYK: Comment about the reason for this switch please. (Node 4/Node 5+)

src/config.js
arcanis: You can't traverse folders if they are not executable.

BYK: TIL. Thanks!

src/cli/index.js
```diff
@@ -60,6 +60,7 @@ export function main({
     '--modules-folder <path>',
     'rather than installing modules into the node_modules folder relative to the cwd, output them here',
   );
+  commander.option('--preferred-cache-folder <path>', 'specify a custom folder to store the yarn cache if possible');
   commander.option('--cache-folder <path>', 'specify a custom folder to store the yarn cache');
```
Timo82: Hallo und weg

__tests__/integration.js
```diff
@@ -70,3 +72,43 @@ test('--mutex network', async () => {
...
+  const cacheFolder = path.join(cwd, '.cache');
+
+  await fs.mkdirp(cacheFolder);
+
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const args = ['--preferred-cache-folder', cacheFolder];
+
+  const options = {cwd};
+
+  {
```
BYK: I don't understand these custom blocks. Why do we need them and how do they work?

__tests__/integration.js
```diff
@@ -70,3 +72,43 @@ test('--mutex network', async () => {
...
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+
+    expect(stdoutOutput.toString().trim()).toEqual(path.join(cacheFolder, `v${constants.CACHE_VERSION}`));
+    expect(stderrOutput.toString()).not.toMatch(/Skipping preferred cache folder/);
+  }
+
+  await fs.chmod(cacheFolder, 0o000);
```
BYK: It looks like these should be two separate test cases:
```
describe('cache folder fallback', () => {
  test('use --preferred-cache-folder when available', async () => {
    ...
  });

  test('fallback to default cache when --preferred-cache-folder is not available', async () => {
    ...
  });
});

File: __tests__/integration.js [hunk 17]
```diff
@@ -70,3 +72,43 @@ test('--mutex network', async () => {
...
+
+  {
+    const {stderr, stdout} = execa(command, ['cache', 'dir'].concat(args), options);
+
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+
+    expect(stdoutOutput.toString().trim()).toEqual(
```
BYK: We can probably use `.toMatchSnapshot()` in both cases which is easier?

src/cli/commands/cache.js
```diff
@@ -55,7 +55,7 @@ export const {run, setFlags, examples} = buildSubCommands('cache', {
   },
 
   dir(config: Config, reporter: Reporter) {
-    reporter.log(config.cacheFolder);
+    reporter.log(config.cacheFolder, {force: true});
```
BYK: I am not happy with this solution since I think it needs a better overall solution as we have discussed. Let's have a `TODO` comment here saying "Don't use force option and fix this ASAP" or something along those lines so we don't forget?

src/cli/index.js
```diff
@@ -60,7 +60,8 @@ export function main({
     '--modules-folder <path>',
     'rather than installing modules into the node_modules folder relative to the cwd, output them here',
   );
-  commander.option('--cache-folder <path>', 'specify a custom folder to store the yarn cache');
+  commander.option('--preferred-cache-folder <path>', 'specify a custom folder to store the yarn cache if possible');
```
BYK: Better to say "prefer a custom cache folder when it is available". We can make this change in a follow-up.

src/cli/index.js
```diff
@@ -60,7 +60,8 @@ export function main({
...
+  commander.option('--cache-folder <path>', 'specify a custom folder that must be used to store the yarn cache');
```
BYK: A+ with the added emphasis with "must be". 👍

src/config.js
```diff
@@ -276,9 +276,41 @@ export default class Config {
...
+
+    if (!cacheRootFolder) {
+      let preferredCacheFolders = constants.PREFERRED_MODULE_CACHE_DIRECTORIES;
+      const preferredCacheFolder = opts.preferredCacheFolder || this.getOption('preferred-cache-folder', true);
+
+      if (preferredCacheFolder) {
+        preferredCacheFolders = [preferredCacheFolder].concat(preferredCacheFolders);
+      }
+
+      for (let t = 0; t < preferredCacheFolders.length && !cacheRootFolder; ++t) {
```
BYK: Let's just use `i` when it is available in loops? :D

src/config.js
```diff
@@ -276,9 +276,41 @@ export default class Config {
...
+        const tentativeCacheFolder = String(preferredCacheFolders[t]);
```
BYK: I'm guessing this is to silence Flow. Note to self: look into this since it may be pointing to a possibly `undefined` value which would break Yarn.

src/config.js
BYK: On second thought, we may wanna add a comment about this since I'm not sure if "You can't traverse folders if they are not executable." is common knowledge. What do you think?

src/config.js
```diff
@@ -276,9 +276,41 @@ export default class Config {
...
+          await fs.mkdirp(tentativeCacheFolder);
+          // eslint-disable-next-line
+          await fs.access(tentativeCacheFolder, fs.constants.R_OK | fs.constants.W_OK | fs.constants.X_OK);
+          cacheRootFolder = tentativeCacheFolder;
+        } catch (error) {
+          this.reporter.warn(this.reporter.lang('cacheFolderSkipped', tentativeCacheFolder));
+        }
+
+        if (cacheRootFolder && t > 0) {
+          this.reporter.warn(this.reporter.lang('cacheFolderSelected', cacheRootFolder));
```
BYK: Note to self: make this information an info or debug level message for all cases to signal which cache folder is being used for each run.

src/constants.js
```diff
@@ -46,15 +47,21 @@ function getDirectory(category: string): string {
...
+
   if (process.platform === 'darwin') {
-    return path.join(userHome, 'Library', 'Caches', 'Yarn');
+    preferredCacheDirectories.push(path.join(userHome, 'Library', 'Caches', 'Yarn'));
+  } else {
+    preferredCacheDirectories.push(getDirectory('cache'));
   }
 
-  return getDirectory('cache');
+  preferredCacheDirectories.push(path.join(os.tmpdir(), '.yarn-cache'));
```
BYK: We should probably append a timestamp or something to the `'.yarn-cache'` name to make it unique across instances and runs? What do you think? (can be a follow-up)

src/reporters/base-reporter.js
```diff
@@ -195,7 +195,7 @@ export default class BaseReporter {
   success(message: string) {}
 
   // a simple log message
-  log(message: string) {}
+  log(message: string, {force = false}: {force?: boolean} = {}) {}
```
BYK: Similar to above comment: let's add a TODO discouraging using this option.

src/util/fs.js
```diff
@@ -13,6 +13,15 @@ const globModule = require('glob');
 const os = require('os');
 const path = require('path');
 
+export const constants =
+  typeof fs.constants !== 'undefined'
```
BYK: Comment about Node 4 compatibility here would be great so in the future we don't get confused about this code.

__tests__/integration.js
arcanis: The `let` variables declared inside are lexically bound. It allows me to reuse the variable name and underlying that those two blocks are checking the same thing. It could as well be a function, but since there are only two invocations I figured it was simpler to just duplicate the part of the code.

__tests__/integration.js
arcanis: I'm not sure snapshots are needed here since the result shouldn't ever change. I prefer to test with the explicit value because otherwise, it wouldn't test this PR (for example, the test suite fails on Windows because for some reason the fallback is used when it shouldn't. I wouldn't have seen this problem with snapshots).

__tests__/integration.js
BYK: I'd prefer if you can make this a function since it would make the code easier to follow. I don't remember seen usage unnamed, custom blocks in a project I have worked on and it would really surprise me as it did right now. I understand that we can leverage the block-scoping with `let` and `const` but I think this makes the code less maintainable and harder to reason about.

shaunc:
Is this potentially a security hole (as /tmp is often writable by other users)?

BYK:
>Is this potentially a security hole (as /tmp is often writable by other users)?

If you specify a cache folder explicitly with `--cache-folder` Yarn fails if it cannot use it, precisely because of this concern. For the common case, I'm not so sure.

shaunc:
I would prefer that the default be secure and explicit config required for insecure usage, rather than the other way around. (E.g. have `--no-cache-folder` or some such to allow search... or maybe more sensible --cache-folder-fallback ...)

BYK:
>I would prefer that the default be secure and explicit config required for insecure usage, rather than the other way around. (E.g. have --no-cache-folder or some such to allow search... or maybe more sensible --cache-folder-fallback ...)

For `--no-cache`, follow https://github.com/yarnpkg/rfcs/pull/53 please. We intend to ship that now. For the fallback part, are you suggesting we have `--no-cache-fallback` by default and when people enable it explicitly, use the fallback mechanisms?

shaunc:
> you suggesting we have --no-cache-fallback by default and when people enable it explicitly, use the fallback mechanisms?

Yes ... or at the very least warn unless they have explicitly enabled. IMHO if someone doesn't have default cache location writable then 1) there is a reason for it or 2) there is no reason for it, but, either way, getting the user to think about it seems a good idea. Using /tmp without the user being conscious of it seems like a hack.

BYK:
>Yes ... or at the very least warn unless they have explicitly enabled. 

This is what happens right now.

>IMHO if someone doesn't have default cache location writable then 1) there is a reason for it or 2) there is no reason for it, but, either way, getting the user to think about it seems a good idea. Using /tmp without the user being conscious of it seems like a hack.

Again, a warning is printed and we are thinking about a `--strict` mode, where having warnings would cause yarn to quit early with an error code. I think that combination would give you what you expect, right?

shaunc:
Ah ... vg -- didn't realize there was an automatic warning. Yes -- "strict mode" sounds like a good idea; its documentation should point out that it has security benefits (as well as any other features -- I can imagine that over time various checks that would be a pain to configure separately could be included). Seems like it should be recommended for production builds.

Thanks!

PR #4152, opened 2017-08-11
Fix: Prevents interpreting run options as yarn options

arcanis:
**Summary**

Running the following command doesn't have the expected effect, because the `--no-coverage` option is interpreter by Yarn instead of being forwarded to the script. This patch fix this.

```
$> yarn test-only --no-coverage     # currently bad
$> yarn test-only -- --no-coverage  # currently good
```

This is a breaking change: it will change some edge-case behaviors.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -75,59 +75,57 @@ test('--mutex network', async () => {
 
 // Windows doesn't have the "echo" utility
 // Since this feature isn't platform-specific, running them on OSX and Linux only should be fine
-if (process.platform !== 'win32') {
-  test('yarn run <script> --opt', async () => {
-    const cwd = await makeTemp();
+test('yarn run <script> --opt', async () => {
+  const cwd = await makeTemp();
 
-    await fs.writeFile(
-      path.join(cwd, 'package.json'),
-      JSON.stringify({
-        scripts: {echo: 'echo'},
-      }),
-    );
+  await fs.writeFile(
+    path.join(cwd, 'package.json'),
+    JSON.stringify({
+      scripts: {echo: `echo`},
+    }),
+  );
 
-    const command = path.resolve(__dirname, '../bin/yarn');
-    const options = {cwd, env: {YARN_SILENT: 1}};
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const options = {cwd, env: {YARN_SILENT: 1}};
 
-    const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--opt'], options);
+  const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--opt'], options);
 
-    const stdoutPromise = misc.consumeStream(stdout);
-    const stderrPromise = misc.consumeStream(stderr);
+  const stdoutPromise = misc.consumeStream(stdout);
+  const stderrPromise = misc.consumeStream(stderr);
 
-    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+  const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
 
-    expect(stdoutOutput.toString().trim()).toEqual('--opt');
-    expect(stderrOutput.toString()).not.toMatch(
-      /From Yarn 1\.0 onwards, scripts don't require "--" for options to be forwarded/,
-    );
-  });
+  expect(stdoutOutput.toString().trim()).toEqual('--opt');
+  expect(stderrOutput.toString()).not.toMatch(
+    /From Yarn 1\.0 onwards, scripts don't require "--" for options to be forwarded/,
+  );
+});
 
-  test('yarn run <script> -- --opt', async () => {
-    const cwd = await makeTemp();
+test('yarn run <script> -- --opt', async () => {
+  const cwd = await makeTemp();
 
-    await fs.writeFile(
-      path.join(cwd, 'package.json'),
-      JSON.stringify({
-        scripts: {echo: 'echo'},
-      }),
-    );
+  await fs.writeFile(
+    path.join(cwd, 'package.json'),
+    JSON.stringify({
+      scripts: {echo: `echo`},
+    }),
+  );
 
-    const command = path.resolve(__dirname, '../bin/yarn');
-    const options = {cwd, env: {YARN_SILENT: 1}};
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const options = {cwd, env: {YARN_SILENT: 1}};
 
-    const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--', '--opt'], options);
+  const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--', '--opt'], options);
 
-    const stdoutPromise = misc.consumeStream(stdout);
-    const stderrPromise = misc.consumeStream(stderr);
+  const stdoutPromise = misc.consumeStream(stdout);
+  const stderrPromise = misc.consumeStream(stderr);
 
-    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
+  const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
 
-    expect(stdoutOutput.toString().trim()).toEqual('--opt');
-    expect(stderrOutput.toString()).toMatch(
-      /From Yarn 1\.0 onwards, scripts don't require "--" for options to be forwarded/,
-    );
-  });
-}
+  expect(stdoutOutput.toString().trim()).toEqual('--opt');
+  expect(stderrOutput.toString()).toMatch(
+    /From Yarn 1\.0 onwards, scripts don't require "--" for options to be forwarded/,
+  );
+});
 
 test('cache folder fallback', async () => {
   const cwd = await makeTemp();
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -73,59 +73,60 @@ test('--mutex network', async () => {
   ]);
 });
 
-// Windows doesn't have the "echo" utility
-// Since this feature isn't platform-specific, running them on OSX and Linux only should be fine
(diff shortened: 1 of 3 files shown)


BYK:
Also, tests or didn't happen.

arcanis:
> Do you know if we have an open or closed issue around this?

No, this is an issue reported by the Oculus people, and something I personally struggle with on a daily basis. :p

> The problem with this approach is it makes yarn take no arguments at all such as --prefix or --modules-folder.

Not really an issue since the only potential option concerned is `--prefix`, which is not implemented and has its own set of issues + the solution is obvious: any option before the script name would have to be yarn options. We currently don't need this behavior, tho.

kaylieEB:
Although it's convenient to not have to use the `--`, this would change the behavior from `npm` and how we write scripts inside `package.json`, so I'd prefer it if we could keep them the same.

arcanis:
Hm I don't know, this behavior really is a pain, especially since it is completely silent, which means that unless people actually look at the command executed, they won't see that the options have been ignored. It's potentially dangerous (just imagine something like `yarn rm / --dry-run` being executed).

A less invasive possibility would be to only add this behavior when running `yarn <pkg>`. Running `yarn run <pkg>` would still require the `--`. I don't really like this because it causes an unnecessary split between the two commands behaviors, but maybe it's still better than the alternative.

cpojer:
I think I'm on board with this change, but agree that we need tests for it. We had an internal discussion about this at FB previously and would like to make sure other people are involved in a vote for this feature. I seem to remember that there were some issues with this approach that made this less workable, but I do think that `yarn --options run --foo` makes a lot of sense to overall and will definitely be a better UX for beginners. Can we make this work at all in a backwards compatible way at all? 

@kaylieEB I think diverging here is ok, given that in scripts you'd write "yarn run …" instead of "npm run" or a generic command, like just "run", so it's clear that you are using either tool.

arcanis:
I don't think we can - the only way I could see this working would be to add some kind of "if the command line doesn't begins with --, add one" (so that we wouldn't break people already using --), but that seems extremely confusing and unexpected.

BYK:
I think the point we ended up is pretty great so 👍 . Thanks everyone, especially @arcanis. We need to be very loud and clear about this change though.

cpojer:
I think that's confusing and will break projects with people on different versions of Yarn. Can we add a deprecation warning to 1.0 if the command starts with `<bin> run --` and continue on with the old behavior? I understand it is confusing, but I think breaking this in the 1.0 release will cause more pain than it provides value.

kaylieEB:
> Can we add a deprecation warning to 1.0 if the command starts with <bin> run -- and continue on with the old behavior?

I think if this could be supported, it would be great since devs don't have to alter their current use habits, similar to how we support both `yarn [command]` and `yarn run [command]`. If not, definitely a big warning would be helpful.

arcanis:
@BYK @kaylieEB good for review

src/reporters/lang/en.js
```diff
@@ -188,6 +188,8 @@ const messages = {
 
   execMissingCommand: 'Missing command name.',
 
+  dashDashDeprecation:
+    "Using -- to pass arguments to your scripts isn't required anymore. Doing this may cause issues in future versions.",
```
cpojer: How about: "Yarn does not require \`--\` to pass arguments to scripts. In a future version of Yarn, passing \`--\` will fail with an error." ?

src/cli/index.js
```diff
@@ -403,7 +418,7 @@ export default function start() {
   const doubleDashIndex = process.argv.findIndex(element => element === '--');
   const startArgs = process.argv.slice(0, 2);
   const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
-  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
+  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex, process.argv.length);
```
kaylieEB: can you do `process.argv.slice(doubleDashIndex)` since it'll just default to length of the array?

src/reporters/lang/en.js
arcanis: It won't fail, it will just be forwarded to the binary as-is.

```
$> yarn test -- --no-coverage
```

will become

```
$> jest -- --no-coverage
```

instead of

```
$> jest --no-coverage
```

This is important because in some case you really want to prevent something from being interpreted as an option.

src/cli/index.js
```diff
@@ -110,6 +110,16 @@ export function main({
     command = commands.run;
   }
 
+  let warnAboutRunDashDash = false;
+  // we using "yarn <script> -abc" or "yarn run <script> -abc", we want -abc to be script options, not yarn options
+  if (command === commands.run) {
+    if (endArgs.length === 0) {
+      endArgs = ['--', ...args.splice(1, args.length)];
```
kaylieEB: same for here? `args.splice(1)`?

src/reporters/lang/en.js
cpojer: Right, in that case then, maybe this: "From Yarn 1.0, scripts do not require \`--\` to be passed. In a future version of Yarn, passing \`--\` will be forwarded to the script directly." ?

arcanis: Much better than the current text, thanks 👍

src/cli/index.js
arcanis: Sure thing 👍

__tests__/integration.js
```diff
@@ -73,6 +73,62 @@ test('--mutex network', async () => {
...
+
+    const command = path.resolve(__dirname, '../bin/yarn');
+    const options = {cwd, env: {YARN_SILENT: 1}};
+
+    const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--opt'], options);
+
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
```
BYK: Is there an advantage to this indirection instead of just doing
```
const stdoutOutput = await misc.consumeStream(stdout);
const stderrOutput = await misc.consumeStream(stderr);
```

__tests__/integration.js
```diff
@@ -73,6 +73,62 @@ test('--mutex network', async () => {
...
+
+    expect(stdoutOutput.toString().trim()).toEqual('--opt');
```
BYK: Are we testing the "$ <command>" echo string here? If so we may wanna add a comment since this may change.

__tests__/integration.js
```diff
@@ -73,6 +73,62 @@ test('--mutex network', async () => {
...
+
+    const command = path.resolve(__dirname, '../bin/yarn');
+    const options = {cwd, env: {YARN_SILENT: 1}};
+
+    const {stderr: stderr, stdout: stdout} = execa(command, ['run', 'echo', '--', '--opt'], options);
+
+    const stdoutPromise = misc.consumeStream(stdout);
+    const stderrPromise = misc.consumeStream(stderr);
+
+    const [stdoutOutput, stderrOutput] = await Promise.all([stdoutPromise, stderrPromise]);
```
BYK: Lots of repeated code here. You may clean it up with `beforeEach` and `afterEach` hooks with a `describe('run with arguments', ` block.

src/cli/index.js
```diff
@@ -110,6 +110,16 @@ export function main({
     command = commands.run;
   }
 
+  let warnAboutRunDashDash = false;
+  // we using "yarn <script> -abc" or "yarn run <script> -abc", we want -abc to be script options, not yarn options
```
BYK: we _are_ using

src/cli/index.js
```diff
@@ -171,6 +181,11 @@ export function main({
   //
   const run = (): Promise<void> => {
     invariant(command, 'missing command');
+
+    if (warnAboutRunDashDash) {
+      reporter.warn(reporter.lang('dashDashDeprecation'));
```
BYK: Arguably, this should be an `info` message, not a `warning` since with the `--strict` option it will break and I don't think should break just because you are still using the harmless `--` thing. What do you think?

__tests__/integration.js
arcanis: Problem with this writing is that it doesn't work, because the stdin and stdout streams are started and stopped as soon as the event loop gets a chance, ie. before the second line gets executed. Before of this, the second `consumeStream` call never receives the `end` event, and never returns. The synchronous version makes sure we attach both of our listeners before allowing the event loop to continue.

__tests__/integration.js
arcanis: No, because we run with `YARN_SILENT: 1`, so Yarn doesn't print the `yarn run` header.

__tests__/integration.js
BYK: Got it. I think this deserves a comment. Even better, find a way that doesn't require this workaround (not necessarily in this patch). Maybe use something else rather than the stream consumer? This may also help: https://nodejs.org/api/stream.html#stream_readable_resume

src/cli/index.js
arcanis: Makes sense, but the PR that will implement the `--strict` feature will need to revisit all the loglevels anyway.

arcanis:
@BYK Long term plan is to switch to a different command line parser more suited to our use cases. I have it on my backlog, but time has been a scarce resource lately :p

src/cli/index.js
arcanis: me hulk

src/cli/index.js
BYK: Eh, fair point. I'd still get certain things "right" from the start but up to you, not gonna block.

BYK:
>I have it on my backlog, but time has been a scarce resource lately :p

Shall we make it explicit by creating an issue?

__tests__/integration.js
BYK: Oh nice. Then what are we testing here? I'm confused :D

arcanis:
@BYK https://github.com/yarnpkg/yarn/issues/4225

Issue #4076, opened 2017-08-02
Yarn local mirror creates invalid file name when Git repository is at server root


itsananderson:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

My project's local `.yarnrc` is configured with this setting:

```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


yarn-offline-mirror "./virtual_env/yarn_packages"
```

In my `package.json` I have a reference to the following internal package:

```
"private_package": "git+ssh://git@git.corp.net:private_package.git",
```

Notice that Git repository is at the root of the server, rather than being a subfolder as with Github repositories: `git@github.com:itsananderson/some-repo.git`

Unfortunately, with the way Yarn parses the Git URL to generate the cache name, it generates this:

```
:private_repo.git-198dac0ae3f10765058a69d02c4f987527cd9b51
```
I believe that's due to this code: https://github.com/yarnpkg/yarn/blob/5e1eee74f18b7f229e8accb49e7ba9ac9c4655bf/src/fetchers/git-fetcher.js#L40

When a URL like `git+ssh://git@git.corp.net:private_package.git` is parsed, the `pathname` property is set to `:private_package.git`. One potential fix for this would be to simply trim the colon from the beginning.

**If the current behavior is a bug, please provide the steps to reproduce.**

I tried to create a GitHub demo of this issue, but since the issue only appears for repositories where the Git repo name is at the root of the server, the issue can't be reproduced with GitHub repositories.

Here are the steps I used to create a minimal repro by creating a local `--bare` Git clone and adding it as `git+ssh//$USER@localhost:node-extend.git`. I used `node-extend` because it's a simple module with no dependencies that can be installed from its Git repository.

```
# Create local bare git repo
git clone --bare git@github.com:justmoon/node-extend.git

# Clone local repo to make sure host signature is accepted etc.
git clone $USER@localhost:node-extend.git

# Create test project
mkdir git-cache-repro
cd git-cache-repro
yarn init -y
echo 'yarn-offline-mirror "./yarn-packages"' >> .yarnrc
yarn add git+ssh://$USER@localhost:node-extend.git
ls yarn-packages
# Note that the listed package is prefixed by a colon
```

Note that you may need to authorize your own SSH key to log into your machine. This should work for a typical setup:

```
cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
[...]

PR #4159, opened 2017-08-12
Fix local mirror invalid file name (#4076)

jespino:
There is a problem related to how url.parse and path.basename is used to build the repository name. So the simplest and easiest solution to this, is after the parsing remove ":" when is needed. Looks quick and dirty but understanding that the objective of the function is to transform reference to the packageFilename looks like the correct place to manage exceptional behavior.

### DIFF
--- a/__tests__/fetchers.js
+++ b/__tests__/fetchers.js
@@ -76,6 +76,25 @@ test('GitFetcher.fetch', async () => {
   expect(name).toBe('beeper');
 });
 
+test('GitFetcher.getTarballMirrorPath without slashes in the repo path', async () => {
+  const dir = await mkdir('git-fetcher');
+  const config = await Config.create();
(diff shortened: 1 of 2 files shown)


arcanis:
Thanks for your help! Can you give an example of something that demonstrates the issue, and write a test in the `__tests__` folder to make sure we don't unexpectedly have a regression later on?

jespino:
Hi @arcanis , the demonstration process is in the issue #4076 (you can follow the issue steps and you get a cache file that starts with ":" which is an invalid character in windows).

About the test, of course, I'll add a test for it.

jespino:
Added the new test, this check that the generated cache directory doesn't include the ":" in this especial case.

arcanis:
Thanks! 👍

PR #4165, opened 2017-08-14
Add private flag for yarn init command

### LINKED ISSUE #4113: yarn init --private
feature

Should we add the option to use `yarn init` to create a private project?
Like:
```
yarn init -p
yarn init --private
yarn init -yp
yarn init --yes --private
```
Every time I want to create a private project, I have to edit `package.json`. Very time consuming.
This feature also is not available in the `npm`.
One day I have to create a lot of private projects to test and it goes on and on

Create a `package.json` file with format:
```
{
  "name": "test",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "private": true
}
```

pierreneter:
**Summary**
Close #4113, this PR create the `private` flag for `yarn init` command. Can use both `yes` and `private` flags. Or can use `private` flag alone to set `private` is `true` when run command.

**Test plan**
Have added tests


src/cli/commands/init.js
```diff
@@ -89,12 +91,17 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       question: 'license',
       default: String(config.getOption('init-license')),
     },
+    {
+      key: 'private',
+      question: 'private',
+      default: '',
```
rally25rs: Could also add `validation` and `validationError` fields to ensure the user enters something that "yn" can convert to boolean (or only accept "true" and "false" and eliminate the dependency on "yn").

rally25rs: If the default was `default: defaultFlag ? 'true' : 'false'` then it would eliminate a conditional below

src/cli/commands/init.js
```diff
@@ -114,14 +121,18 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       def = val;
     }
 
+    if (manifestKey === 'private' && privateFlag) {
+      def = true;
+    }
+
     if (def) {
-      question += ` (${def})`;
+      question += ` (${def.toString()})`;
```
rally25rs: Why the additional `.toString()`?

src/cli/commands/init.js
```diff
@@ -114,14 +121,18 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       def = val;
     }
 
+    if (manifestKey === 'private' && privateFlag) {
```
rally25rs: This conditional could be eliminated if the default is set above in the array of questions

src/cli/commands/init.js
```diff
@@ -114,14 +121,18 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
     if (def) {
-      question += ` (${def})`;
+      question += ` (${def.toString()})`;
     }
 
     let answer;
     let validAnswer = false;
 
-    if (yes) {
+    if (yes || (privateFlag && manifestKey === 'private')) {
```
rally25rs: I think this additional `||` would also be eliminated if the `default` was set based ont he flag in the array of questions.

src/cli/commands/init.js
```diff
@@ -141,6 +152,9 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
 
     if (answer) {
+      if (manifestKey === 'private') {
```
rally25rs: It would be a better design to add something like `dataType: 'boolean'` to the schema of questions when they are defined, then here check that data type and use "yn" in the case of booleans only.
Or perhaps even `inputFormatter: yn` in the questions schema, then here: `answer = question.inputFormatter ? question.inputFormatter(answer) : answer`

src/cli/commands/init.js
rally25rs: also the default should probably be "false" anyway so that when the question is printed tot he user they can see that it is explicitly false by default. Empty string might make someone question if the default is actually going to end up true or false (in reality it depends how "yn" interprets empty string)

src/cli/commands/init.js
```diff
@@ -141,6 +152,9 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
+        answer = yn(answer);
```
BYK: We probably should make use of `yn` more since we added it?

src/cli/commands/init.js
pierreneter: I think `1` i shorter way user want to fill this key, so i use `yn`. And everything can be become a boolean when we use `yn`. So, if users want to change it, they can change it later by edit `package.json` file. Just `boolean` value is approved.

pierreneter: I keep `''` because we don't need add line `private: {boolean}` into `package.json` If users don't set value for private when they run: `yarn init`. And one thing, `true` and `false` are a `boolean` value, not a `string`, so I don't wrapper it in `''`.

src/cli/commands/init.js
pierreneter: because after i add this flag, `def` could be `string` or possibly `boolean`. I want to make sure it will be a `string` when added to `question`.

src/cli/commands/init.js
pierreneter: no need, just add `true` when user uses this flag.

src/cli/commands/init.js
rally25rs: But it's already inside a string interpolation, so would get converted to a string already.

BYK:
@rally25rs @pierreneter ping! Do you need anything from us to move this forward? I'd love to get this merged.

rally25rs:
@BYK I would go refactor out some of the conditionals myself, but I don't think there is a way for me to directly modify this PR?

src/cli/commands/init.js
rally25rs: This is a valid point. I was thinking yn might error if the user types in something like "i guess so" as the answer, but I suspect it will just try to convert it anyway. As long as yn doesn't throw an exception, then I would be OK with no validation.

pierreneter:
Hi,
@rally25rs you can create a PR to my fork, here: https://github.com/pierreneter-repositories/yarn/tree/init/private
I will merge it, no matter how.

src/cli/commands/init.js
pierreneter: Like I said below. No default value.

src/cli/commands/init.js
pierreneter: If removed it will generate 2 errors:
```
$ eslint . && flow check
Error: src/cli/commands/init.js:130
130:       question += ` (${def})`;
                            ^^^ boolean. This type cannot be coerced to
130:       question += ` (${def})`;
                       ^^^^^^^^^^^ string
```

pierreneter:
@rally25rs 
So, I move `yn` to array of question like you want. And consolidate the tests to make sure all I want.
You can do anything free to satisfy the above cases that I give: https://github.com/yarnpkg/yarn/pull/4165#issuecomment-322264770 or agree with what I do.
Please respond soon.
This RP has taken 10 days.
When I respond, no one responds. You can not ask others to do things and ignore them, not listen to them.

src/cli/commands/init.js
rally25rs: That's interesting. That is Flow complaining about the type. I didn't think it would complain inside an interpolated string. Oh well, in that case the `.toString()` is fine.

rally25rs:
The failing CircleCI tests seems unrelated to this PR, and passes on other integrations.

@BYK I can't merge this unless those pass. I also can't trigger a rebuild on CircleCI because I don't have permissions. Could you re-kick Circle? (or just merge this is github gives you that ability)

BYK:
>or just merge this is github gives you that ability

Never, unless I really have to :D

BYK:
>When I respond, no one responds. You can not ask others to do things and ignore them, not listen to them.

Hey @pierreneter, sorry for the communication issues. It is easy to miss notifications from GitHub as they tend to be a lot, for active projects like Yarn. Would you like to join our [Discord server](https://discord.gg/yarnpkg) so you can ping members directly in case they don't notice the updates? Also, apologies, since I try to go over all open PRs regularly to make sure they are not forgotten and I missed this one.

Thanks a lot for your hard work and sticking to the end with this one!

PR #4166, opened 2017-08-14
Enhancement: Prevents "yarn add" from being run from the workspace root directory

arcanis:
**Summary**

This PR prevents using `yarn add <pkg>` from within a workspace root folder, in order to prevent common mistakes where a package is added to the workspace root folder instead of being added to the right workspace (which, because of the node resolution algorithm, would be effectively invisible to the user).

If the user really wants to add a dependency to the workspace root folder, then we suggest them to use `yarn add <pkg> --dev` instead. This should be fine since the workspaces will probably only be used in development mode anyway. An alternative would be to introduce a new flag `--force` that would bypass this check, but I think the current heuristic is fine until we receive complaints.

**Test plan**

Added two tests.

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -42,6 +42,18 @@ const runAdd = buildRun.bind(
   },
 );
 
+test.concurrent('add without --dev should fail on the workspace root', async () => {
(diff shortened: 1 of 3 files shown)


arcanis:
Build failed because of an intermittent failure on the Yarn website. Retrying.

Issue #651, opened 2016-10-11
Bug with @types/package dependencies


lholznagel:
**Do you want to request a *feature* or report a *bug*?**
Possible bug found

**What is the current behavior?**
When the `package.json` contains dependencies like `@types/node` it fails on `yarn outdated` with the following error.

Error:
```
error https://registry.yarnpkg.com/@types/node: Not found
    at Request.params.callback [as _callback] (/usr/share/yarn/lib-legacy/util/request-manager.js:285:18)
    at Request.self.callback (/usr/share/yarn/node_modules/request/request.js:187:22)
    at emitTwo (events.js:87:13)
    at Request.emit (events.js:172:7)
    at Request.<anonymous> (/usr/share/yarn/node_modules/request/request.js:1048:10)
    at emitOne (events.js:77:13)
    at Request.emit (events.js:169:7)
    at IncomingMessage.<anonymous> (/usr/share/yarn/node_modules/request/request.js:969:12)
    at emitNone (events.js:72:20)
    at IncomingMessage.emit (events.js:166:7)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Install `@types/node` (`yarn add @types/node`)
2. Run `yarn outdated` or `yarn outdated @types/node`
3. The error should occur

**What is the expected behavior?**
It should check if `@types/node` is outdated and should not throw an error

**Please mention your node.js, yarn and operating system version.**
node.js: `6.7.0`
yarn: `v0.15.0`
os: `Ubuntu 16.04`

ssreekanth:
Duplicate of #620

lholznagel:
Oh sorry, didn´t find that while searching.

This then can be closed in favor of #620

PR #4174, opened 2017-08-15
Add cwd flag

### LINKED ISSUE #692: Analog to npm's --prefix command
In certain scenarios it is useful to be able to run npm from a seperate directory. With npm I can install node_modules from the root directory to the build directory. `npm --prefix ./build install`

Is there an analogous command for yarn?

kittens:
**Summary**

Fixes #692. This adds a `--cwd` flag to change the current working directory. This is so you can specify another working directory to perform an operation that's not the current one. But Sebastian, why not just cd into it and run Yarn there?

1. In Node the cwd is `realpath`d which means that if you cd into a symlinked directory then the cwd will be at the real location. I have a weird use-case where I have a symlinked folder in a workspace that is outside of the workspace root.
2. It makes Yarn wrapper scripts nicer since you don't need to mess with cding into a folder then cding back out. 

**Test plan**

Added tests.


arcanis:
But Sebastian, why not just cd into it and run Yarn there? 😆 

Related to #692 - I'm personally not convinced this feature should be in Yarn itself, but if people really want it I see no point preventing this from landing altogether :) First point is interesting tho. Do you have a detailed example?

arcanis:
I see. It's unfortunate Node is realpathing the cwd 😞 The only thing that makes me cautious with this patch is that unless I'm mistaken, the yarnrc options will not be read relative to the option `cwd`, which might be quite unexpected. @BYK wdyt?

kittens:
> unless I'm mistaken, the yarnrc options will not be read relative to the option cwd

[`options.cwd` is passed down to all `Registry` instances](https://github.com/yarnpkg/yarn/blob/fa0fb69a567e1c44320d6f2a5f2a8ee42bc226ea/src/config.js#L245). Everything in the Yarn codebase references `config.cwd` instead of `process.cwd()` (at least it should).

arcanis:
Not the command line options. They are extract in [cli/index.js](https://github.com/yarnpkg/yarn/blob/3bfa1e3d6303c58aed9660e1a2af3a6d6f030bce/src/cli/index.js#L108) because we need to have them before we run the command line argument parser, which occurs before we create the configuration. Adding a way to change the cwd would in turn have a weird behavior, where half the options (all command line options) will be read from a file (the original directory yarnrc) whilst the others will be read from another file (the redirected dir yarnrc).

kittens:
Updated to handle `--cwd` inside of `.yarnrc`. This makes the initial args `.yarnrc` search work if you pass a `--cwd` on the CLI and when it finds a `--cwd` flag in a `.yarnrc` file then it will reset and load `.yarnrc` relative to it. It detects `.yarnrc` recursion too and will bail out. I've added a bunch of tests that cover the new and previous behaviour of the code.

src/util/rc.js
```diff
@@ -55,13 +26,42 @@ export function findRc(name: string, parser: Function): Object {
     addConfigPath(home, `.${name}rc`);
   }
 
-  addRecursiveConfigPath(process.cwd(), `.${name}rc`);
+  // add .yarnrc locations relative to the cwd
+  while (true) {
+    configPaths.unshift(path.join(cwd, `.${name}rc`));
+
+    const upperCwd = path.dirname(cwd);
```
BYK: Shouldn't we stop at the project root?

src/rc.js
```diff
@@ -25,57 +23,76 @@ const buildRcConf = () =>
...
+// extract the value of a --cwd arg if present
+function extractCwdArg(args: Array<string>): ?string {
+  for (let i = args.length - 1; i >= 0; i--) {
+    const arg = args[i];
+    if (arg === '--cwd') {
+      return args[i + 1];
+    }
   }
+  return null;
+}
```
arcanis: Not completely sure about this one. Running `yarn run -- --cwd foo` will break, the `--cwd` will be interpreted when it really shouldn't. Additionally, making a pseudo-parsing here means that it might get out of sync if we change the cli parsing in some unexpected ways (for example #4152).

I feel like a better option might be to do a first pass of cli parsing using the yarnrc options available in the current directory branch, then extract the `--cwd` option and, if present, run a second pass using the new directory as base. It would always be synced, and wouldn't have much perf impact in most cases. I don't know of feasable this is with commander tho.

src/reporters/lang/en.js
```diff
@@ -116,6 +116,8 @@ const messages = {
   multiplePackagesCantUnpackInSameDestination:
     'Pattern $0 is trying to unpack in the same destination $1 as pattern $2. This could result in a non deterministic behavior, skipping.',
   incorrectLockfileEntry: 'Lockfile has incorrect entry for $0. Ignoring it.',
+  cwdInYarnrcWarning:
+    'There was a --cwd flag in a .yarnrc file relative to this cwd. .yarnrc files relative to the --cwd have been ignored.',
```
BYK: Better to say "relative to where you ran yarn" instead of "relative to this cwd". What do you think?

src/rc.js
```diff
@@ -5,17 +5,15 @@ import parse from './lockfile/parse.js';
...
-  rcUtil.findRc('yarn', (fileText, filePath) => {
+// given a cwd, load all .yarnrc files relative to it
+function getRcConfigForCwd(cwd: string): {[key: string]: string} {
+  return rcUtil.findRc('yarn', cwd, (fileText, filePath) => {
     const {object: values} = parse(fileText, 'yarnrc');
-    const keys = Object.keys(values);
 
-    for (const key of keys) {
+    // some keys reference directories so keep their relativity
+    for (const key in values) {
```
BYK: Aren't `for ... in` loops "considered harmful? :D

src/rc.js
```diff
@@ -25,57 +23,76 @@ const buildRcConf = () =>
...
-  return result;
-}
+  // concat wildcard arguments and arguments meant for this specific command
+  const newArgs = [].concat(argMap.get('*') || [], argMap.get(commandName) || []);
+
+  // check if the .yarnrc args specified a cwd
+  const newCwd = extractCwdArg(newArgs);
+  if (newCwd && newCwd !== origCwd) {
+    // ensure that we don't enter into a loop
+    if (previousCwds.includes(newCwd)) {
```
arcanis: We need to use `indexOf` for the Node 4 compatibility

src/util/rc.js
arcanis: Not really, because if we were to do that we wouldn't read the yarnrc files in $HOME, for example.

src/rc.js
kittens: Not really. Mutated prototypes are the only dangerous thing and they aren't really a practical concern.

src/rc.js
kittens: I thought that too but `includes` is already used [here](https://github.com/yarnpkg/yarn/blob/master/src/lockfile/parse.js#L374).

src/rc.js
kittens: It would require abstracting out a lot of stuff from `cli/index.js` which I'm not comfortable doing at the moment.

src/rc.js
arcanis: It's because `String#includes` is implemented in Node 4, but not `Array#includes`

src/rc.js
arcanis: Makes sense. Can you at least iterate from start to end and stop iterating if you encounter a `--` token?

src/rc.js
arcanis: I'd say the problem is more readability than anything. It's really unusal to see `for ... in` loops nowadays, and it's not obvious when reading the code whether it's intended or not. Flow helps a bit, but not much.

That being said, `for ... in` loops are used here and there in the source code so if you want to use it there I'm fine with it. If we eventually choose to stop using them it will be in another PR.

src/reporters/lang/en.js
arcanis: I don't see this message being referenced anywhere - @kittens is it still used or can we remove it?

arcanis:
I pushed a change for the two things I think should be updated. Only remains the reporter key that doesn't seem used, so if you confirm me it's safe to remove it I'll do it and merge this PR if the tests pass 😃

taion:
Would it make sense to make `-C` be an alias for `--cwd`? npm's `--prefix` is not a great name for the flag, but if nothing else is in the way, having `-C` for compatibility seems like a free win.

Also, the `--cwd` flag appears to be missing from the online documentation.

BYK:
@taion a quick search in the code for `-C` revealed that this is used for `--caret` in `upgrade` and `upgrade-interactive` commands.  Although I agree that it would be nice to have a shortcut, it would be a breaking change at this point which I'm not willing to make.

BYK:
>Also, the --cwd flag appears to be missing from the online documentation.

Thanks for the heads up! Filed #651. Feel free to submit something yourself if you're feeling like it :)

guilhermecvm:
Doesn't work with symlinks (trying to run from root package.json when using yarn workspaces)
```json
{
  "scripts": {
    "start": "yarn --cwd packages/dolphino react-scripts start"
  }
}
```
```
➜ dolphino (master) $ yarn start
yarn run v1.3.2
$ yarn --cwd packages/dolphino react-scripts start
$ packages/dolphino/node_modules/.bin/react-scripts start
/bin/sh: packages/dolphino/node_modules/.bin/react-scripts: No such file or directory
````

But file do exists and is symlinked
```
➜ dolphino (master) $ ll packages/dolphino/node_modules/.bin/react-scripts
lrwxr-xr-x  1 guilhermecvm  staff    59B 15 Nov 20:02 packages/dolphino/node_modules/.bin/react-scripts -> ../../../../node_modules/react-scripts/bin/react-scripts.js
```

BYK:
@guilhermecvm please file a new bug report so this can be tracked and addressed.

arcanis:
Or better yet, a pull request 👐

AlastairTaft:
I hit a similar issue where I've symlinked to another mono-repo to include all the packages there as dependencies but that tries to run babel as part of the install, unfortunately babel uses `process.cwd()` so it falls over. Can't work out if this will solve my use case.

PR #4183, opened 2017-08-16
Fix: Use semver tag to colorize package name for outdated/upgrade-interactive commands

### LINKED ISSUE #3815: upgrade-interactive row colours are incorrect
### Current behaviour:

screen shot 2017-07-04 at 5 07 42 pm

### Reproduction steps:

Have a repo with various out of date packages, and run `yarn upgrade-interactive` to see the incorrect colours applied to the rows.

### Expected behaviour:

- `@domain-group/eslint-config-domain`, `@domain-group/fe-brary` & `@domain-group/fe-build` should be neither red nor yellow - they have a major version bump with API incompatible changes.
- `babel-core` should be yellow as it is a minor update.

### Environment:

```
$ yarn --version
0.27.5
$ node --version
v7.10.0
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: OS X 10.10.5 (14F2411)
      Kernel Version: Darwin 14.5.0
```

CarlRosell:
**Summary**
This will fix #3815 by implementing a solution mentioned here https://github.com/yarnpkg/yarn/issues/3815#issuecomment-322496710.
As the `outdated`-command displayed pretty much the same data I changed the colors of that command as well.

New output: 
screen shot 2017-08-16 at 10 22 05
screen shot 2017-08-16 at 11 48 54

**Test plan**

I did not create any new tests as this is only a visual change and was not tested before. But if we decide to move the new "helper files" to a utils folder I can write a few tests to test the output.


src/cli/commands/_color-key-for-versions.js
```diff
@@ -0,0 +1,24 @@
+/* @flow */
+
+import type {FormatKeys} from '../../reporters/format.js';
+import semver from 'semver';
+
+const versionColorScheme = {
```
BYK: Nitpick: I'd use `VERSION_COLOR_SCHEME` and move this to `constants.js`

I'd also move this module under `utils` directory and remove the preceding `_` since this is a useful thing that is not necessarily a private module.

Finally, this may also use it's own Flow type for safety.

src/cli/commands/_color-key-for-versions.js
```diff
@@ -0,0 +1,24 @@
...
+  preminor: 'yellow',
+  patch: 'green',
+  prepatch: 'green',
+  prerelease: 'red',
+  unchanged: 'white',
+  unknown: 'red',
+};
+
+export default function(from: string, to: string): FormatKeys {
+  const [validFrom, validTo] = [semver.valid(from), semver.valid(to)];
```
BYK: Destructuring looks nice here but you now add the overhead of creating an array and destructuring it only for aesthetics. I'd just do two separate `const`s as follows:
```
const validFrom = semver.valid(from);
const validTo = semver.valid(to);
```

src/cli/commands/_color-key-for-versions.js
```diff
@@ -0,0 +1,24 @@
...
+  if (validFrom && validTo) {
```
BYK: The following may be easier to reason about:
```
let versionBump = 'unknown';
if (validFrom && validTo) {
    versionBump = semver.diff(validFrom, validTo) || 'unchanged';
}
return versionColorScheme[versionBump];
```

src/cli/commands/_colorize-diff.js
```diff
@@ -0,0 +1,14 @@
+/* @flow */
```
BYK: Again, move this under `utils` and drop the preceding `_` please :)

src/cli/commands/upgrade-interactive.js
```diff
@@ -57,26 +59,15 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   const headerPadding = (header, key) =>
     `${reporter.format.bold.underline(header)}${' '.repeat(maxLengthArr[key] - header.length)}`;
 
-  const colorizeName = ({current, wanted}) => (current === wanted ? reporter.format.yellow : reporter.format.red);
+  const colorizeName = (from, to) => reporter.format[colorKeyFromVersions(from, to)];
```
BYK: I think it'd be better to keep object destructuring here for the sake of named arguments. What do you think?

CarlRosell: The reason that I changed this not to use destructuring was to keep it pure as the second argument could be either `wanted` or `latest` if the `--latest` flag is set.

CarlRosell:
I got the same error locally but managed to solve it by rebasing from master. I think that e92ac93ab3eb0aa3af570bff73754d83e1fba287 fixed that issue. I have no idea what will happen if I just force push after rebasing to a PR 🙂. What should I do?

CarlRosell:
Ok. Tried to force push now 🤞

src/cli/commands/upgrade-interactive.js
BYK: Alright, that seems okay to me :)

Issue #3752, opened 2017-06-29
Yarn removes compiled libraries if "Linking dependencies" was interrupted by user


d4rky-pl:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
If I stop yarn during `Linking dependencies` (via ^C, for example because I made a mistake and wanted to stop and add `--dev`) then when installing any new package, compiled files in unrelated packages are removed and not recompiled properly.

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
mkdir test-bug
cd test-bug
yarn init
yarn add node-sass
ls node_modules/node-sass # the vendor directory is correctly there

yarn add react # press ^C (ctrl+C) during "Linking dependencies" phase
ls node_modules/node-sass # the vendor directory is still there

yarn add react
ls node_modules/node-sass # (!) the vendor directory is gone
```

**What is the expected behavior?**
Yarn should either not remove compiled binaries if "Linking dependencies" step has been interrupted or recompile them afterwards.

**Please mention your node.js, yarn and operating system version.**
`yarn`: 0.24.6
`node`: v8.1.2
`OS`: macOS 10.12.15

bestander:
Thanks, @d4rky-pl.
Yeah, yarn should be atomic and should indicate that install scripts started but have not finished.
The logic is in package-linker.js, help is welcome

jamsinclair:
Furthering d4rky-pl's reproduction steps: 

```shell
mkdir test-bug
cd test-bug
yarn init
yarn add node-sass
ls node_modules/node-sass/vendor # the vendor contents are correctly there

# Simulate the user/system exiting during the linking step
# The linking step removes the `.yarn-integrity` file
rm node_modules/.yarn-integrity

yarn
ls node_modules/node-sass/vendor # (!) the vendor contents are gone
```

It seems to be the deletion of the integrity file means the package-linker no longer has context of module artifacts the next time yarn is invoked.  

If the integrity file is missing on init of yarn, should it do a force install of all modules (like `yarn remove`)?

Happy to help, if someone can point me in the right direction 😸

bestander:
@jamsinclair I think you are right.

Some pointers:
https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L350 - this is where Yarn does the integrity check and can know if .yarn-integrity exists

https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L481 - this is where it runs scripts.

[...]

PR #4185, opened 2017-08-16
Bugfix 3752 - Artifacts disappearing when missing integrity file

jamsinclair:
**Summary**

I encountered personally similar problems as described in #3752 after interrupting a yarn install.

After some debugging I isolated this to being that the integrity file is removed during the `linking` step (with good reasons). 

However, if this file is never rewritten (e.g. user/system interrupted yarn) the next time yarn is invoked it no longer has context of package artifacts, as they are retrieved from the integrity file. 

I played around with some solutions and would like to put forward an idea of forcing package installations if the integrity file is missing.

**Test plan**

I've added two test cases for `add` and `install` scripts that replicate the reproduction steps listed in https://github.com/yarnpkg/yarn/issues/3752#issuecomment-322021848.

Without changes to the src code you can see these test cases fail (running Node v8.2.1):
screen shot 2017-08-17 at 00 09 06
screen shot 2017-08-17 at 00 11 24

With the src codes changes you can see on CI that all tests pass ✌️ 

**Concerns**

I'm not sure but, are there any scenarios where setting `force` in the `PackageInstallScripts` would cause the lockfile to be rewritten? Thinking of whether this change would affect flags like `--frozen-lockfile` 🤔


arcanis:
Great work @jamsinclair, thanks! Have you checked if it works fine with the `--ignore-scripts` flag?

src/package-install-scripts.js
```diff
@@ -34,6 +34,10 @@ export default class PackageInstallScripts {
   force: boolean;
   artifacts: InstallArtifacts;
 
+  setForce(force: boolean) {
```
BYK: Do we really need these setters and getters?

jamsinclair: Good question.

The force property was formerly only set on construction of the `PackageInstallScripts` class:
https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L178

When the install bailout method runs the `PackageInstallScripts` instance has already been created. Adding the setter, allowing overrides, seemed like a simpler solution rather than reinstantiating. Happy to hear another approach.

jamsinclair:
@arcanis, yup just tested changes now with `--ignore-scripts` 👍 

Looks right. The package install scripts do not seem to take place and reports a warning:
```
warning Ignored scripts due to flag.
```

src/package-install-scripts.js
arcanis: I think a setter is fine, it's how we already do it (`setArtifacts`), and I have the feeling it makes the code a bit safer (because JS doesn't (yet) have private variables, setters are one of the only ways to clearly state that a variable might be updated from the outside).

arcanis:
Thanks @jamsinclair! A simple solution to an important issue 👌

anhkind:
Will this fix be included in 0.27.x and 0.28.x versions _soon enough_?

arcanis:
It's already available through the [nightly builds](http://nightly.yarnpkg.com/). Otherwise, it will be in the next tagged release, which should be the 1.0.0 and should come in the next few weeks.

PR #4192, opened 2017-08-17
Fixes the info tests

### LINKED ISSUE #4191: master is randomly failing
All the failures are on CircleCI, and look like this:

```
Summary of all failing tests
 FAIL  __tests__/commands/info.js
  ● without arguments and in directory containing a valid package file

    expect(array).toContain(value)
    
    Expected array:
      ["name", "description", "dist-tags", "versions", "maintainers", "time", "repository", "readmeFilename", "homepage", "bugs", "license", "users", "version", "installationMethod", "preferGlobal", "engines", "bin", "dist", "directories"]
    To contain value:
      "scripts"
      
      at expectedKeys.forEach.key (__tests__/commands/info.js:55:52)
      at Array.forEach (native)
      at runInfo (__tests__/commands/info.js:55:18)
      at __tests__/commands/info.js:29:13
      at next (native)
```

Have we changed something related to this? @BYK @Daniel15

arcanis:
**Summary**

This PR fixes #4191. Now that we are using the single-file build with npm, the `scripts` field isn't included into the `package.json` file anymore. It's not clear why the problem only appeared now, but my guess is that this is related to the recent publication of the 0.28.4 release.

**Test plan**

Tests should now pass again.

### DIFF
--- a/__tests__/commands/info.js
+++ b/__tests__/commands/info.js
@@ -41,13 +41,12 @@ const expectedKeys = [
   'repository',
   'bugs',
   'license',
-  'scripts',
   'dist',
   'directories',
 ];
 
-// yarn now ships as built, single JS files so it has no dependencies
-const unexpectedKeys = ['dependencies', 'devDependencies'];
+// yarn now ships as built, single JS files so it has no dependencies and no scripts
+const unexpectedKeys = ['dependencies', 'devDependencies', 'scripts'];
 
 test.concurrent('without arguments and in directory containing a valid package file', (): Promise<void> => {
   return runInfo([], {}, 'local', (config, output): ?Promise<void> => {


arcanis:
Note that the problem occured because we're using unmocked requests for the info.js tests. I wouldn't say it's an issue tho, since we actually want to test what's on the npm registry. The problem is that we're checking it too late, only after it's already been deployed. Not sure how we could fix this tho.

https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/info.js#L10-L11

Daniel15:
I'm getting a sense of déjà vu here... Didn't we already fix this??

Daniel15:
@BYK wasn't https://github.com/yarnpkg/yarn/pull/3899 supposed to fix this?

BYK:
@Daniel15 it was. Apparently, something has changed in between or there was some very aggressive caching somewhere.

Issue #4177, opened 2017-08-15
yarn pack doesn’t run prepack


alfonsomunozpomer:
**Do you want to request a *feature* or report a *bug*?**
A bug.

**What is the current behavior?**
`yarn pack` creates a tarball without running `prepack`.

**If the current behavior is a bug, please provide the steps to reproduce.**
Creae a package with a `prepack` script. When you run `yarn pack` the script is not executed.

**What is the expected behavior?**
The script should be run before the tarball is created by `yarn pack`.

**Please mention your node.js, yarn and operating system version.**
Node v8.2.1
yarn v0.27.5
macOS Sierra 10.12.6

joaolucasl:
I'll take this one! PR coming soon :)

iwllyu:
if it helps my npm v3.10.10 also does not run `prepack` on `npm pack`

BYK:
>if it helps my npm v3.10.10 also does not run prepack on npm pack

~~May be this is intentional then?~~

Here it says this needs to be run: https://docs.npmjs.com/misc/scripts

BYK:
Thanks a lot for the swift action @joaolucasl!

kaylieEB:
Cloing this via https://github.com/yarnpkg/yarn/pull/4193

PR #4193, opened 2017-08-17
Added missing pre/post hooks for `pack` (#4177)

joaolucasl:
**Summary**
Yarn wouldn't run the `prepack` and `postpack` scripts, as it is expected when `pack` is ran (#4177).
This PR adds those missing calls and updates tests accordingly.

**Test plan**
Added tests.


BYK:
Thanks a lot!

deecewan:
Nice! Man, I should really get around to implementing my RFC relating to the hooks.

Issue #4115, opened 2017-08-08
Add the yarn re-route feature to .yarnc


cpojer:
The way we use Yarn at Facebook is by specifying a path in the `.yarnc` file that points to a single file bundle or entry point of Yarn. If the global version of Yarn is executed in a folder (or parent folder) with a `.yarnrc` with that setting, it should use that version of Yarn instead. This ensures that the same version of Yarn is used for the same project.

Beware of infinite re-directs.

arcanis:
So the plan is:

- first we try to execute the file as a Javascript file with `child_process.fork`
- then if it failed, we try to execute it via `child_process.spawn`

cpojer:
Instead of running it in a child process, is there any way we could require the file, if it's JS?

arcanis:
We could, but I'm not completely sure what would be the possible side effects, for example regarding `process.argv[1]`.

arcanis:
Implemented by #4204

PR #4204, opened 2017-08-18
Adds a yarn-path rc config parameter

arcanis:
**Summary**

This PR implements #4115 by adding a yarn-path configuration parameter (yarnrc only) that, if found, abort the current process and instanciate a new one on the given path, forwarding any command line argument.

**Implementation details**

- I `spawn` before `fork` because, otherwise, given a shell script or any non-javascript file, the call to `fork` would succeed (because the node binary exists), but the node process thus spawned would crash because of a syntax error. We wouldn't be able to detect if the exit reason is legit or not without inspecting stdout, which would be pretty bad.

- I avoid infinite loops by setting a variable in the environment that disable this behavior (so that the newly spawned process doesn't open a new process ad vitam aeternam). I think this should be fine for now, but I'm open to alternatives.

**Test plan**

Don't worry, I'm in the process of writing the tests :)


arcanis:
Had to switch the prettify parser to babylon, since [the flow parser is broken on `export default async function` structures](https://github.com/facebook/flow/issues/3264) 😞

arcanis:
Switching to babylon wasn't enough, since the Flow check is still failing. I changed the structure to export the symbol in two step, which avoids hitting this particular edge case.

.eslintrc
```diff
@@ -24,7 +24,7 @@
       "trailingComma": "all",
       "bracketSpacing": false,
       "printWidth": 120,
-      "parser": "flow"
+      "parser": "babylon"
```
BYK: Do we still need this change after your last change?

src/cli/index.js
```diff
@@ -398,14 +399,32 @@ export function main({
...
-  const doubleDashIndex = process.argv.findIndex(element => element === '--');
-  const startArgs = process.argv.slice(0, 2);
-  const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
-  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
+async function start(): Promise<void> {
+  const rc = getRcConfigForCwd(process.cwd());
+  const yarnPath = rc['yarn-path'];
 
-  main({startArgs, args, endArgs});
+  if (yarnPath && process.env.YARN_IGNORE_PATH !== '1') {
```
BYK: Maybe we can check if the current yarn path matching `yarnPath` too to get rid of the environment variable? I mean I'm fine with the env variable since it also allows us to bypass this behavior. Just asking what you think :)

src/util/child.js
```diff
@@ -15,6 +15,34 @@ let uid = 0;
...
+      reject(error);
+    });
+
+    proc.on('close', exitCode => {
+      resolve(exitCode);
+    });
+  });
+}
+
+export function spawnp(program: string, args: Array<string>, opts?: Object): Promise<number> {
```
BYK: We already have a promisified `spawn` below. Why do we need a new one?

src/cli/index.js
```diff
@@ -398,14 +399,32 @@ export function main({
...
+    const argv = process.argv.slice(2);
+    const opts = {stdio: 'inherit', env: Object.assign({}, process.env, {YARN_IGNORE_PATH: 1})};
+
+    try {
+      await spawnp(yarnPath, argv, opts);
```
BYK: I remember you mentioning the reason for trying `spawn` first. Can you put that as a comment here?

Also, I feel like we should try `require`ing the file (after checking if it ends with a `.js` extension or in a try-catch` block?

src/cli/index.js
```diff
@@ -398,14 +399,32 @@ export function main({
...
+    } catch (firstError) {
+      try {
+        await forkp(yarnPath, argv, opts);
```
BYK: Why discard this error?

src/cli/index.js
```diff
@@ -398,14 +399,32 @@ export function main({
...
+      } catch (error) {
+        throw firstError;
+      }
+    }
+  } else {
+    // ignore all arguments after a --
+    const doubleDashIndex = process.argv.findIndex(element => element === '--');
+    const startArgs = process.argv.slice(0, 2);
+    const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
+    const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
```
BYK: Is this on top of your `run` patch?

src/cli/index.js
```diff
@@ -398,14 +399,32 @@ export function main({
     });
 }
 
-export default function start() {
-  // ignore all arguments after a --
-  const doubleDashIndex = process.argv.findIndex(element => element === '--');
-  const startArgs = process.argv.slice(0, 2);
-  const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
-  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex + 1, process.argv.length);
+async function start(): Promise<void> {
```
BYK: Maybe we can just "fire and forget"? Or make this function sync but attach the `then` handler to set `process.exitCode`?

src/util/child.js
arcanis: The other one isn't the actual promisified `spawn` :/ it prints extra logs, and more importantly it buffers the output.

src/cli/index.js
arcanis: Because I rethrow the first error (not an executable) rather than the second one (not a javascript file).

src/cli/index.js
arcanis: > Also, I feel like we should try requireing the file (after checking if it ends with a .js extension or in a try-catch` block?

Why that? Javascript files are already supported by the `forkp` call below. Required modules are affected by side effects, so I'm not super confortable using it (let's say we use polyfills, if we use `require` then they will get attached twice to the same environment).

src/cli/index.js
arcanis: Nope, this is the same code as the one removed [here](https://github.com/yarnpkg/yarn/pull/4204/files/890bab1bd8e0767704bf29bb1b908d500369a10c#diff-867becf4a9c2c6c6d4e7c1278750724eL402), except that it now is in an `else` branch.

.eslintrc
arcanis: Nope, will revert it.

src/cli/index.js
arcanis: I think it's fine as is, but if you prefer it another way I can change it (note that the symbol is still exported [here](https://github.com/yarnpkg/yarn/pull/4204/files/890bab1bd8e0767704bf29bb1b908d500369a10c#diff-867becf4a9c2c6c6d4e7c1278750724eR438), it's just that the function declaration and symbol export have to be split for Flow to correctly parse it).

src/cli/index.js
arcanis: Thought about doing this, but I'm a bit wary about symlinks shenanigans, and path normalization :/ An environment variable felt safer and less error-prone. Plus as you mentioned, it allows disabling this behavior which might be useful in some particular cases.

src/cli/index.js
BYK: This was to reduce the startup time of Yarn. Forking processes are not cheap. They are even more expensive on Windows. I think it is fine as it is but if we receive performance complaints, we should probably think about using a `require`. Otherwise, this functionality can easily be achieved using a `bash` script more efficiently which reduces the value of the feature IMO.

src/cli/index.js
BYK: What if the second error was something else like a syntax error etc.?

src/cli/index.js
```diff
@@ -413,14 +414,32 @@ export function main({
...
 }
 
-export default function start() {
-  // ignore all arguments after a --
-  const doubleDashIndex = process.argv.findIndex(element => element === '--');
-  const startArgs = process.argv.slice(0, 2);
-  const args = process.argv.slice(2, doubleDashIndex === -1 ? process.argv.length : doubleDashIndex);
-  const endArgs = doubleDashIndex === -1 ? [] : process.argv.slice(doubleDashIndex);
+async function start(): Promise<void> {
+  const rc = getRcConfigForCwd(process.cwd());
```
BYK: I'm hoping this doesn't add too much of an overhead to the startup.

src/cli/index.js
arcanis: Then it will be rejected as not an executable. The alternative is to hide both errors and to throw something generic instead (like "Referenced file is not executable nor a valid javascript file").

PR #4218, opened 2017-08-21
Fix: make integrity-check resillient against old integirty files

BYK:
**Summary**

We've added new field to the integrity file, `modulesFolders`, which will
be missing from the old versions. Since the code assumes this field is
always an array, it causes failures with old integrity files. This patch
fixes the issue by always providing some defaults.

**Test Plan**

Added new unit test.


arcanis:
What would you think about wrapping the whole `_compareIntegrityFiles` function inside a `try/catch` block? This way we would actually catch any possible mistake.

BYK:
>What would you think about wrapping the whole _compareIntegrityFiles function inside a try/catch block instead? This way we would actually catch any possible mistake.

I think that's a safer option but I'm concerned about the performance penalty we pay for try-catch. It is pretty large, even on the most recent Node versions AFAIK.

BYK:
I'd actually prefer using a JSON-schema validator like ESLint does but I feel like that's a bit heavy-handed just for this file. If we use that for `package.json` validation too, that may be a nice addition.

arcanis:
I think that's a bit too heavy to be used here. We just need something really simple.

As for the `catch` perf time, this function is only called a single time, so it will probably be irrelevant anyway, since optimizations usually kick in after the function has been called a few times (otherwise V8 doesn't have enough stats to correctly optimize them).

BYK:
>Approving because those changes are sound, but I think try/catch would improve it. Your call :)

I think this is a good idea but it a bit more involved to simply bake into this patch. I need to add a new error type: `EXCEPTION_OCCURRED` and a message to go along with it. I'll merge this as it is and file a ticket to possibly wrap this in a try-catch block. I'm not sure if we should assume integrity is broken when an exception happens or simply stop. Stopping seems like the safer option for now.

Issue #4008, opened 2017-07-24
Because of package hoisting accounting for popular, linking phase takes over 15 minutes


kenotron:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Since #2676, the package hoister now accounts for the popularity of a package pattern to determine which version to hoist. This requires a complete traversal of the graph. Given our dependency graph, somehow this bit of code takes an unexpectedly long amount of time (>30 minutes).

I've verified that this occurs on a Mac or Windows box.

**If the current behavior is a bug, please provide the steps to reproduce.**
Given that this bug is caused by our internal packages, it is difficult to replicate in a bug. 

**What is the expected behavior?**
I worked around this problem with a hack to stop the prepass from going through more than one occurrence of a pattern temporarily and saw that time to install drop from >30 minutes down to 40 seconds. There should be an option in Yarn to allow for this popularity prepass to be turned off in a config.

**Please mention your node.js, yarn and operating system version.**
node.js 8.1.4
yarn 0.27.5
os Windows 10 / OS X 10.12

rally25rs:
Can you add your `package.json`?

kenotron:
Not sure at the moment what the root cause is: installation is slow even for 0.21.3 before that checkin went into place.

kenotron:
Okay confirmed that the 0.20.4 is fast. Somehow the release notes indicates that 0.21.x didn't have the hoisting fix, but it does.

kenotron:
I have a fix for the slow hoisting issue - please take a look at this #4222

kenotron:
Seeing as this is merged, I'm marking this closed. Thanks @arcanis!

PR #4222, opened 2017-08-21
Major optimization on hoisting

kenotron:
Keep track of side-effects of each visit of package, and replay side-effects rather than re-crawling through the subtree given the same exact pattern

**Summary**
This is a bug fix for #4008. Since Package Hoisting prepass was introduced, the algorithm crawls through the dependency tree regardless of whether the package has been seen before or not. This is okay for small repositories. To scale this to a large repository that has packages numbered in the 50+k, this prepass eats up CPU and hangs for upwards of 15-30minutes. 

The fix is to cache the side-effects of adding occurrences of the packages. This PR preserves the ability for the occurrence to be accounted for as well - it remembers what changed last time the same pattern was encountered and simply reruns the same side-effects (visits) as last time.

**Test plan**
I re-ran the unit tests available - the most important one being the integration-hoisting.js tests. These demonstrate that the hoisting functionality is preserved.

### DIFF
--- a/src/package-hoister.js
+++ b/src/package-hoister.js
@@ -487,13 +487,14 @@ export default class PackageHoister {
   prepass(patterns: Array<string>) {
     patterns = this.resolver.dedupePatterns(patterns).sort();
 
-    const visited: {
-      [pattern: string]: {
+    const visited: Map<
+      string,
+      Array<{
         pkg: Manifest,
         ancestry: Array<Manifest>,
         pattern: string,
-      }[],
-    } = {};
+      }>,
+    > = new Map();
 
     const occurences: {
       [packageName: string]: {
@@ -525,10 +526,12 @@ export default class PackageHoister {
         return;
       }
 
-      if (visited[pattern]) {
+      let visitedPattern = visited.get(pattern);
+
+      if (visitedPattern) {
         // if a package has been visited before, simply increment occurrences of packages
         // like last time this package was visited
-        visited[pattern].forEach(visitPkg => {
+        visitedPattern.forEach(visitPkg => {
           visitAdd(visitPkg.pkg, visitPkg.ancestry, visitPkg.pattern);
         });
 
@@ -540,8 +543,6 @@ export default class PackageHoister {
       const ref = pkg._reference;
       invariant(ref, 'expected reference');
 
-      visited[pattern] = visited[pattern] || [];
-
       visitAdd(pkg, ancestry, pattern);
 
       for (const depPattern of ref.dependencies) {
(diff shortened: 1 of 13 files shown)


src/package-hoister.js
```diff
@@ -496,30 +504,52 @@ export default class PackageHoister {
...
         // prevent recursive dependencies
-        return;
+        return pkg;
+      }
+
+      if (visited[pattern]) {
+        // if a package has been visited before, simply increment occurrences of packages
+        // like last time this package was visited
+        visited[pattern].forEach(visitPkg => {
+          visitAdd(pkg, ancestry, pattern);
```
arcanis: Shouldn't this line be?:

```
visitAdd(visitPkg.pkg, visitPkg.ancestry, visitPkg.pattern);
```

arcanis:
Thanks @kenotron! By any chance, would you have a simple example of the broken execution flow? For example, let's say we have the following tree:

```
/my-pkg
    /foo
        /baz
            /hello
            /world
    /bar
        /baz
            /hello
            /world
```

If I understand correctly, when traversing `/my-pkg/foo/baz` you'd like to store the result of the iteration in order to use it directly when iterating over the node in `/my-pkg/bar/baz`, is that correct? Would that work with nested dependencies like `foobar` in the following example? Since they are repeated patterns (`foobar` is used twice) that live in repeated patterns (both `hello` and `baz` are used twice).

```
/my-pkg
    /foo
        /baz
            /hello
                /foobar
            /world
    /bar
        /baz
            /hello
                /foobar
            /world
```

kenotron:
Hmm. The best way is through a test. Lemme try and put that example in there...

src/package-hoister.js
kenotron: Things got messed up as I was refactoring... Will fix this!

kenotron:
Updated based on review comments. Also added an integration test case suggested by @arcanis

src/package-hoister.js
```diff
@@ -487,6 +487,14 @@ export default class PackageHoister {
   prepass(patterns: Array<string>) {
     patterns = this.resolver.dedupePatterns(patterns).sort();
 
+    const visited: {
```
arcanis: Would you mind changing this to use a `Map` instead? They're available starting from Node 4, so it should be fine for the compatibility, and I expect it to improve the perfs an extra bit :)

kenotron: sure

kenotron:
The Travis CI job seems stuck... https://travis-ci.org/yarnpkg/yarn/jobs/267254272 not sure how to kick it again

arcanis:
Yeah, I think Travis OSX instances are overloaded :/ Maybe because it's midday in the US.

PR #4240, opened 2017-08-23
Fix #3697 Improve error handling when tarball fetcher receives bad hash

### LINKED ISSUE #3697: yarn does not verify tarballs put to the offline folder
bug

Yarn likely does not verify tarball written to the configured offline folder. On chocked (mobile) connection
happens that it writes even zero sized one. Then yarn fails with uninformative message of mysterious bad hash:
```
error Bad hash. Expected "8c2a5ef2472fd9ea742b04c77a75093ba2757c93" but got "da39a3ee5e6b4b0d3255bfef95601890afd80709" 
```
without even a failed package name.

*Note, that "__da39a .... 80709__" is a sha1 hash of the __zero__ size file*.

Full incident described at [create-react-native-app 272 issue](https://github.com/react-community/create-react-native-app/issues/272).  Also there is separate yarn [issue about uninformative message](https://github.com/yarnpkg/yarn/issues/3689).

Use [toxiproxy](https://github.com/Shopify/toxiproxy) to emulate cellular network conditions (bandwidth<1Mbps, latency>1000, jitter 800, with slicer and timeout toxics) set for registry then invoke yarn to install something with many dependencies (e.g. react-native) on a system with cleaned global, offline storage and cache. Or just use real poor cellular connection instead of toxiproxy.

 * Yarn should be able to deal with network errors and outages.
 * Yarn should write to offline cache and use only VERIFIED tarballs.
 * Error message should give full path to the offending tarball. 

node v6.11.0
yarn v0.24.6
Linux 4.4.70 #23 SMP Wed May 27 05:11:55 UTC 2017 (d45547b) x86_64 x86_64 x86_64 GNU/Linux

```
$ yarn config list|sed -r"s/\x1B\[([0-9]{1,3}((;[0-9]{1,3})*)?)?[m|K]//g"
yarn config v0.24.6
info yarn config
{ 'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': true,
  'version-git-message': 'v%s',
  'init-version': '0.1.0',
  'init-license': 'GPLv3',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'strict-ssl': true,
  'user-agent': 'yarn/0.24.6 npm/? node/v4.7.3 linux x64',
  'cache-folder': '[scrubbed]/.yarn_cache',
  lastUpdateCheck: 1498045409214,
  'yarn-offline-mirror': '[scrubbed]/.yarn_offline' }
info npm config
{}
```

kaylieEB:
**Summary**
Fix for https://github.com/yarnpkg/yarn/issues/3697

- Previously, the tarball fetcher always piped the content to offline cache path when the request succeeded (<=400), regardless of whether the received hash matched the requested hash or not. So I moved the write process to the `finish` block of the stream when it checks for the matching hash.
- Improved warning message to include package name and be more user friendly

**Test plan**
Added a test case in fetchers

*BEFORE*
screen shot 2017-08-23 at 8 50 34 am

*AFTER*

screen shot 2017-08-23 at 8 47 01 am
*note hashes are the same in the screenshots bc I made the code block fail intentionally

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -78,25 +78,26 @@ export default class TarballFetcher extends BaseFetcher {
         error.message = `${error.message}${tarballPath ? ` (${tarballPath})` : ''}`;
         reject(error);
       })
-      .on('finish', () => {
+      .on('finish', async () => {
         const expectHash = this.hash;
         const actualHash = validateStream.getHash();
+
         if (!expectHash || expectHash === actualHash) {
-          const tarballMirrorPath = this.getTarballMirrorPath();
-          const tarballCachePath = this.getTarballCachePath();
-
-          if (tarballMirrorPath) {
-            validateStream.pipe(fs.createWriteStream(tarballMirrorPath)).on('error', reject);
-          }
-
-          if (tarballCachePath) {
-            validateStream.pipe(fs.createWriteStream(tarballCachePath)).on('error', reject);
-          }
-
           resolve({
             hash: actualHash,
           });
         } else {
+          const tarballMirrorPath = this.getTarballMirrorPath();
+          const tarballCachePath = this.getTarballCachePath();
+
+          if (await fsUtil.exists(tarballMirrorPath)) {
+            fsUtil.unlink(tarballMirrorPath);
+          }
+
+          if (await fsUtil.exists(tarballCachePath)) {
+            fsUtil.unlink(tarballCachePath);
+          }
+
           reject(
             new SecurityError(
               this.config.reporter.lang(
@@ -158,6 +159,9 @@ export default class TarballFetcher extends BaseFetcher {
             process: (req, resolve, reject) => {
               // should we save this to the offline cache?
               const {reporter} = this.config;
+              const tarballMirrorPath = this.getTarballMirrorPath();
+              const tarballCachePath = this.getTarballCachePath();
+
               const {validateStream, extractorStream} = this.createExtractor(resolve, reject);
 
               req.on('response', res => {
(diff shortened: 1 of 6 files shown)


src/fetchers/tarball-fetcher.js
```diff
@@ -82,13 +82,30 @@ export default class TarballFetcher extends BaseFetcher {
...
+          const tarballMirrorPath = this.getTarballMirrorPath();
+          const tarballCachePath = this.getTarballCachePath();
+
+          if (tarballMirrorPath) {
+            validateStream.pipe(fs.createWriteStream(tarballMirrorPath)).on('error', reject);
+          }
+
+          if (tarballCachePath) {
+            validateStream.pipe(fs.createWriteStream(tarballCachePath)).on('error', reject);
+          }
```
arcanis: Won't that cause the tarballs to still be written on the disk even if the validation fails?

BYK: Agree with @arcanis. I think what we need is keeping this in the old place and just deleting the file on validation failure before we reject the promise. This would also prevent going over the file two times.

kaylieEB: yup initial approach was `only write when hash matches` but realized there are other reasons that would cause failure. updated!

BYK:
LGTM but las word is yours @arcanis

src/fetchers/tarball-fetcher.js
```diff
@@ -178,6 +185,17 @@ export default class TarballFetcher extends BaseFetcher {
           this.reporter.warn(this.reporter.lang('retryOnInternalServerError'));
           await sleep(3000);
         } else {
+          const tarballMirrorPath = this.getTarballMirrorPath();
+          const tarballCachePath = this.getTarballCachePath();
+
+          if (tarballMirrorPath && (await fsUtil.exists(tarballMirrorPath))) {
+            await fsUtil.unlink(tarballMirrorPath);
```
kaylieEB: @arcanis just curious, why does the unlinking have to be synchronous here?

arcanis: One benefit I see is that uncaught exceptions will be logged to the console, and will possibly abort the execution in future Node versions. Apart from that, it shouldn't really matter, but if you do it I suggest you also add a comment to explain it's made on purpose, since it can also be seen as a typo.

PR #4241, opened 2017-08-23
Fix: Ensure bin folder exists before trying to use it

### LINKED ISSUE #4164: Bug: `yarn create x` throws ENOENT on fresh Windows installs

Bug

On a fresh Windows installation, running `yarn create react-app hello` fails at the very beginnging with an `ENOENT` error for the global bin directory.

On a fresh user account which never had `yarn` or had a `yarn global` installation, run `yarn create react-app something` and see the ENOENT error for the global bin folder.

No errors. The directory should be automatically created.

Windows 10
Node 6
Yarn: latest nightly

BYK:
**Summary**

Fixes #4164. Ensures we have the bin folder before trying to link
binaries into it and throws a user-friendly error if it encounters
any permission issues.

**Test plan**
```
rm -rf $(yarn global bin);
yarn create react-app hello-world
```

Make sure it doesn't throw.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -202,12 +202,12 @@ test('yarn create', async () => {
   const command = path.resolve(__dirname, '../bin/yarn');
   const options = {cwd, env: {YARN_SILENT: 1}};
 
-  const {stderr: stderr, stdout: stdout} = execa(command, ['create', 'index', '.'], options);
+  const {stderr: stderr, stdout: stdout} = execa(command, ['create', 'html'], options);
 
   const stdoutPromise = misc.consumeStream(stdout);
   const stderrPromise = misc.consumeStream(stderr);
 
   const [stdoutOutput, _] = await Promise.all([stdoutPromise, stderrPromise]);
 
-  expect(stdoutOutput.toString()).toMatch(/\[created index\]/);
+  expect(stdoutOutput.toString()).toMatch(/<!doctype html>/);
 });
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -196,3 +196,18 @@ test('cache folder fallback', async () => {
   );
   expect(stderrOutput2.toString()).toMatch(/Skipping preferred cache folder/);
 });
+
+test('yarn create', async () => {
+  const cwd = await makeTemp();
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const options = {cwd, env: {YARN_SILENT: 1}};
+
+  const {stderr: stderr, stdout: stdout} = execa(command, ['create', 'index', '.'], options);
+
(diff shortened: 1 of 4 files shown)


BYK:
NOTE: Planning to add automated tests. Created the PR for CI and to get early feedback.

src/cli/commands/global.js
```diff
@@ -90,11 +90,15 @@ async function getGlobalPrefix(config: Config, flags: Object): Promise<string> {
...
+
+  const binFolder = path.join(prefix, 'bin');
   try {
-    await fs.access(path.join(prefix, 'bin'), (nativeFs.constants || nativeFs).W_OK);
+    await fs.access(binFolder, (nativeFs.constants || nativeFs).W_OK);
   } catch (err) {
     if (err.code === 'EACCES') {
       prefix = FALLBACK_GLOBAL_PREFIX;
+    } else if (err.code === 'ENOENT') {
+      await fs.mkdirp(binFolder);
```
kaylieEB: should you just return prefix here since you call `mkdirp` again in the function below?

BYK: Hmm, that actually makes sense. Let me try that.

__tests__/integration.js
```diff
@@ -196,3 +196,18 @@ test('cache folder fallback', async () => {
...
+  const cwd = await makeTemp();
+  const command = path.resolve(__dirname, '../bin/yarn');
+  const options = {cwd, env: {YARN_SILENT: 1}};
+
+  const {stderr: stderr, stdout: stdout} = execa(command, ['create', 'html'], options);
+
+  const stdoutPromise = misc.consumeStream(stdout);
+  const stderrPromise = misc.consumeStream(stderr);
+
+  const [stdoutOutput, _] = await Promise.all([stdoutPromise, stderrPromise]);
```
arcanis: If you rebase on master, you can replace lines 202-210 by `runYarn`

src/cli/commands/global.js
```diff
@@ -90,11 +90,15 @@ async function getGlobalPrefix(config: Config, flags: Object): Promise<string> {
   } else {
     prefix = POSIX_GLOBAL_PREFIX;
   }
+
+  const binFolder = path.join(prefix, 'bin');
   try {
-    await fs.access(path.join(prefix, 'bin'), (nativeFs.constants || nativeFs).W_OK);
+    await fs.access(binFolder, (nativeFs.constants || nativeFs).W_OK);
```
arcanis: You should be using the `constants` exposed by `util/fs`, which already do the fallback if required. Also, shouldn't we also test for `R_OK` and `X_OK`?

BYK: Good catch! This code was here before but I'll make the improvements.

PR #4243, opened 2017-08-24
Fix: Fix scoped packages' selective resolutions

### LINKED ISSUE #4242: Selective module resolution doesn't work with scoped packages
#4105 cc @kaylieEB

Bug

```json
"resolutions": {
  "@types/react": "15.0.22"
}
```

Result: 

```
Couldn't find any versions for "react" that matches "15.0.22"
```

The error is that yarn thinks that package `@types/react` is a path `@types`/`react`.

Yarn should correctly install the dep.

The source of bug is [this line](https://github.com/yarnpkg/yarn/blob/840acbf7b0e4c1cce13f8aed54db873e6a7f802e/src/resolution-map.js#L51). We cannot simply split by `/` here.

We either have to
1. Improve parsing rules to properly split scoped packages
2. Use different package separator
3. Use some kind of escaping.

yarn@1.0.0-20170823.1414
node@6.11.2
osx 10.12.6

s-panferov:
**Summary**

Fixes #4242 

1. [x] Write tests
2. [x] Write a fix


s-panferov:
@kaylieEB It seems that I made all obvious fixes (and it even works for several cases), but I still can't make it work for the specific commited test case. I feel that I have no enough knowledge of yarn's resolution algorithm to make it work. Could you please take a look?

src/util/parse-package-path.js
```diff
@@ -0,0 +1,26 @@
...
+
+const PACKAGE_SEP = '/';
+
+/**
+ * Parse input strings like `package-1/package-2` to an array of packages
+ */
+export default function parsePackagePath(input: string): string[] {
+  let prevScoped = false
+  let packages = []
+  input.split(PACKAGE_SEP).forEach(pkg => {
```
kaylieEB: can you use `for .. of` loop here instead?

src/util/parse-package-path.js
```diff
@@ -0,0 +1,26 @@
...
+    } else {
+      packages.push(pkg);
+    }
+
+    if (pkg.indexOf('@') !== -1) {
+      prevScoped = true;
+    }
+  });
+
+  return packages
```
arcanis: I think this function can just be replaced by:

```
input.match(/(@[^\/]+\/)?([^/]+)/g);
```

kaylieEB: +100! <3

s-panferov:
Thank you for the review. I'll make changes.

src/util/parse-package-path.js
s-panferov: Ok. I thought about it but came up with functional version. These parsing regexps rapidly becomes unreadable 😄.

s-panferov:
@kaylieEB I've made changes. You are right. When `@scoped/b` is not present at the top level — everything works as expected: package `@scoped/a/@scoped/b` correctly resolved to `2.0.0` according to resolution and hoisted to the top level. 

Second test case with `@scoped/b` present at the top level doesn't work.

s-panferov:
As a first thought: I think that the main problem is [this line](https://github.com/yarnpkg/yarn/blob/840acbf7b0e4c1cce13f8aed54db873e6a7f802e/src/package-resolver.js#L573). The code removes pattern for `@scoped/b@file:../scoped-b1` from the lockfile, but it should be there because it required for the top-level package. Am I right?

s-panferov:
Yes, I believe that the logic should be more sophisticated. We should check if the lockfile entry is used anywhere else before removing it. I'll try to investigate how to achieve this. I will appreciate any help because I'm unfamiliar with Yarn internals.

kaylieEB:
Why don't we try to merge this PR for resolving scoped packages per original issue with the first test case, then you can file a different issue / PR for the lockfile? WDYT?

s-panferov:
@kaylieEB sure! Let's re-review then. This one is rebased, tests pass. The failed test remains in the tree but skipped for now. Also, note that I chose not to use regexp in `parsePackagePath` function, because it get's tricky to handle another case from tests `invalidname/`.

src/resolution-map.js
```diff
@@ -48,7 +49,7 @@ export default class ResolutionMap {
   }
 
   parsePatternInfo(globPattern: string, range: string): ?Object {
-    const directories = globPattern.split(DIRECTORY_SEPARATOR);
+    const directories = parsePackagePath(globPattern);
     const name = directories.pop();
 
     if (!name) {
```
kaylieEB: I'd still prefer the regex version -- you can catch the name error by doing something like
```
if (globPattern.match(/(\/|\*+)$/)) {
  this.reporter.warn(this.reporter.lang('invalidResolutionName', globPattern));
}

const directories = parsePackagePath(globPattern); // input.match(/(@[^\/]+\/)?([^/]+)/g);
const name = directories.pop();
```

this way we can catch invalid names beyond just trailing slash, and in general, less line of code is easier to maintain :)

__tests__/commands/install/resolutions.js
```diff
@@ -51,3 +51,25 @@ test.concurrent('should warn when resolution entries are incorrrect or incompati
...
+  });
+});
+
+test.skip('install with resolutions should correctly install nested scoped packages', (): Promise<void> => {
+  return runInstall({}, {source: 'resolutions', cwd: 'scoped-nested'}, async config => {
+    expect(await getPackageVersion(config, '@scoped/a')).toEqual('1.0.0');
+    expect(await getPackageVersion(config, '@scoped/b')).toEqual('1.0.0');
+    expect(await getPackageVersion(config, '@scoped/a/@scoped/b')).toEqual('2.0.0');
+  });
+});
```
kaylieEB: nit, would you mind just deleting this test case and the related files then add them back later, in case you don't get to the issue? It also makes it easier to review in the pr.

s-panferov:
@kaylieEB @arcanis Do you have any other suggestions? If not, could you merge this, please?  We're struggling with dependency hell in TypeScript's `@type` packages and hope to see this feature work in nightly builds.

s-panferov:
Thanks!

PR #4246, opened 2017-08-24
Better active directory selection

### LINKED ISSUE #2709: yarn ignores existing project package.json
I've been using yarn happily for a while. For most things, it's pretty much a drop-in replacement for npm. However, it behaves very differently if you are working in folders other than the project root.  This is unintuitive, unlike npm in a bad way, and results in multiple _package.json_ files peppered throughout the source, all needing to be maintained separately.

For example, let's say my _package.json_ contains a "build" script. From a terminal, I run `yarn run build` or `npm run build` and both work as expected. Let's say I cd into directory"test/". `npm run build` still works but `yarn run build` complains that there is no _package.json_. 

I would expect that running `yarn run build` would descend the tree until it finds a _package.json_ file and then build the project as though in that directory (just like _npm_, _gulp_, and everything else). I suspect the cause is related to the fact that `yarn add` "conveniently"  creates a package.json file for you if there is no package.json in the current directory, saving you the npm init step. Again, I think yarn add should only do this if it does not find a _package.json_ when descending the tree.

This seems like an unintentional bug to me.

Using yarn version 0.20.1 and Windows 10.

arcanis:
**Summary**

Fix #2709 - Yarn will now try to move to the closest ancestor directory that contains a `package.json` file if it can. If none is to be found, the current directory will be used as usual. Note that the same behavior if applied to the `--cwd` option, so `--cwd foo/bar` will go to `foo` if `foo/package.json` exists but not `foo/bar/package.json`.

I thought about doing this in the same pass than the rc file selection, but it wouldn't change anything because we're not stat'ing the same files, so we could not factorize the checks.

**Test plan**

Added a test case, and the old one should still pass as expected.

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -244,7 +244,7 @@ if (process.platform !== 'win32') {
 
 test.concurrent('should run bin command', async () => {
   const stdout = await execCommand('bin', [], '', true);
-  expect(stdout[0].trim()).toMatch(/[\\\/]node_modules[\\\/]\.bin$/);
+  expect(stdout[0]).toMatch(/[\\\/]node_modules[\\\/]\.bin\n?$/);
   expect(stdout.length).toEqual(1);
 });
 
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -244,7 +244,7 @@ if (process.platform !== 'win32') {
 
 test.concurrent('should run bin command', async () => {
   const stdout = await execCommand('bin', [], '', true);
-  expect(stdout[0]).toEqual(path.join(fixturesLoc, 'node_modules', '.bin'));
+  expect(stdout[0].trim()).toMatch(/[\\\/]node_modules[\\\/]\.bin$/);
   expect(stdout.length).toEqual(1);
 });
 
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -243,7 +243,7 @@ if (process.platform !== 'win32') {
 }
 
 test.concurrent('should run bin command', async () => {
-  const stdout = await execCommand('bin', [], '', false);
+  const stdout = await execCommand('bin', [], '', true);
   expect(stdout[0]).toEqual(path.join(fixturesLoc, 'node_modules', '.bin'));
   expect(stdout.length).toEqual(1);
 });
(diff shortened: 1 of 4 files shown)


__tests__/index.js
```diff
@@ -243,7 +243,7 @@ if (process.platform !== 'win32') {
 }
 
 test.concurrent('should run bin command', async () => {
-  const stdout = await execCommand('bin', [], '', false);
+  const stdout = await execCommand('bin', [], '', true);
```
arcanis: If not run in a temp directory, the new behavior makes us go to the repository root. Other tests don't seem to break.

__tests__/integration.js
```diff
@@ -82,6 +82,22 @@ test('--mutex network', async () => {
   ]);
 });
 
+test('--cwd option', async () => {
+  const cwd = await makeTemp();
+  const cacheFolder = path.join(cwd, '.cache');
+
+  const subdir = path.join(cwd, 'a/b/c/d/e/f/g/h/i');
```
BYK: Since you're using `path.join` I'd recommend writing this as `path.join(cwd, 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i')` for cross-platform compatibility. You can even do `path.join(cwd, ...Array.from('abcdefghi'))` if you wanna look cool :P

src/cli/index.js
```diff
@@ -21,6 +21,22 @@ const net = require('net');
 const onDeath = require('death');
 const path = require('path');
 
+function findActiveDirectory(base): string {
```
BYK: I'd rename this to` findProjectRoot`

src/cli/index.js
```diff
@@ -21,6 +21,22 @@ const net = require('net');
...
+  let prev = null;
+  let dir = base;
+
+  do {
```
BYK: Can be simplified as:
```
while (prevDir !== currentDir && !fs.existsSync(path.join(currentDir, constants.NODE_PACKAGE_JSON))) {
    prevDir = currentDir;
    currentDir= path.dirname(currentDir);
}

return currentDir;
```

PR #4257, opened 2017-08-25
Fix: Deprecate implicit `file:` protocol and only allow package dirs

### LINKED ISSUE #4251: `yarn add jest@test` fails if `test` is a folder
Repro:

```
mkdir test
yarn add jest@test
# node_modules/jest is ../test
```

BYK:
**Summary**

Fixes #4251. Follow up to #4088. Instead of just checking whether the
target is a valid directory, we now check if it contains a
`package.json` file too. This is still different from `npm`'s behavior.
Apparently, `npm` fetches the package info upfront to favor dist-tags
over directories but this comes at a distinct performance penalty and
makes static, deterministic resolution impossible so we are now
deprecating the implicit `file:` protocol in patterns. After a certain
point, we'll remove this code and will require everyone to use `file:`
or at least one of the following path identifiers: `./`, `../`. `/`.

**Test plan**

Updated the existing test for warning check and added a new test for
invalid directories.

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -5,6 +5,12 @@ import type PackageResolver from './package-resolver.js';
 import type {Reporter} from './reporters/index.js';
 import type Config from './config.js';
 import type {Install} from './cli/commands/install';
+
+import path from 'path';
+
+import invariant from 'invariant';
+import semver from 'semver';
+
 import {cleanDependencies} from './util/normalize-manifest/validate.js';
 import Lockfile from './lockfile';
 import PackageReference from './package-reference.js';
@@ -16,10 +22,6 @@ import WorkspaceResolver from './resolvers/contextual/workspace-resolver.js';
 import {getExoticResolver} from './resolvers/index.js';
 import * as fs from './util/fs.js';
 import {normalizePattern} from './util/normalize-pattern.js';
-
-const path = require('path');
-const invariant = require('invariant');
-const semver = require('semver');
 
 type ResolverRegistryNames = $Keys<typeof registryResolvers>;
 
@@ -127,7 +129,8 @@ export default class PackageRequest {
 
     if (!semver.validRange(pattern)) {
       try {
-        if ((await fs.stat(path.join(this.config.cwd, pattern))).isDirectory()) {
+        if (await fs.exists(path.join(this.config.cwd, pattern, constants.NODE_PACKAGE_JSON))) {
+          this.reporter.warn(this.reporter.lang('implicitFileDeprecated', pattern));
           return `file:${pattern}`;
         }
       } catch (err) {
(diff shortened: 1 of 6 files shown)


src/package-request.js
```diff
@@ -5,6 +5,12 @@ import type PackageResolver from './package-resolver.js';
 import type {Reporter} from './reporters/index.js';
 import type Config from './config.js';
 import type {Install} from './cli/commands/install';
+
+import path from 'path';
+
+import invariant from 'invariant';
+import semver from 'semver';
```
arcanis: Be careful with this, I think some of our third-party modules (or their dependencies) are sometimes mocked in our tests, so using import will break them. Case in mind: `request`.

src/resolvers/exotics/file-resolver.js
```diff
@@ -23,6 +23,11 @@ export default class FileResolver extends ExoticResolver {
   loc: string;
 
   static protocol = 'file';
+  static prefixMatcher = /^.{,2}\//;
+
+  static isVersion(pattern: string): boolean {
+    return super.isVersion.call(this, pattern) || this.prefixMatcher.test(pattern);
```
arcanis: Maybe update the regex to also match `C:\...` ?

BYK: Hmm, good catch. I'll go with [path.isAbsolute](https://nodejs.org/api/path.html#path_path_isabsolute_path) which takes care of it for me.

src/resolvers/exotics/file-resolver.js
```diff
@@ -23,6 +24,11 @@ export default class FileResolver extends ExoticResolver {
   loc: string;
 
   static protocol = 'file';
+  static prefixMatcher = /^.{1,2}\//;
```
dylang: Oops, looks like the `.` isn't escaped, so it's matching _any_ two characters, including single-character scopes like `@m`.

I have a fix here: https://github.com/yarnpkg/yarn/pull/4338

sheerun:
Unfortunately this breaks possibility to install local packages without package.json #3855 :(

BYK:
@sheerun - yeah that was kind of intentional since this caused lots of headaches and I think this is safer. Is it too much trouble to require people to have a simple `package.json` for local packages?

sheerun:
For non-npm packages (like bower's or assets) there isn't package.json available. Also checking for package.json won't solve #4251 if `test` directory contains package.json (likely to happen).

The real fix would be to require local dependencies to have `../`, `./`, `/`, or `file:` prepended, just like you noticed. I'd agree that package.json requirement could be enforced when they aren't present.

BYK:
So, we are doing both right now if I'm not wrong. Are you suggesting making it an either / or check?

sheerun:
Yes, either correct path format, or package.json :) But rather deprecate incorrect path next major

BYK:
@sheerun you have angered the open source gods and now they want a PR. Are you up to the challenge?

PR #4264, opened 2017-08-25
Fix: Do not create globalFolder and linkedFolder unless necessary

### LINKED ISSUE #4229: Fix: Don't directly create the global folder if it's not gonna be used
Bug

Whenever you run `yarn`, it simply creates `this.globalFolder` even if it is not going to be used. This may cause issues on systems where the default global folder is not writeable but it's never gonna be used anyways (like CI machines).

Delete your `~/.config/yarn/global` folder and simply run `yarn install`. It will be created again.

`yarn` should not create this folder until it actually needs it.

Yarn 0.28.4

BYK:
**Summary**

Fixes #4229. Config used to ensure these folders existed at init.
This causes unnecessary and sometimes unexpected file system changes.
This patch defers creation of these folders unless necessary and adds
necessary protections for places where they used to be expected to exist.

**Test plan**

Existing unit tests.


src/package-linker.js
```diff
@@ -271,7 +271,18 @@ export default class PackageLinker {
...
+      linkedModules = await fs.readdir(this.config.linkFolder);
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        linkedModules = [];
+      } else {
+        throw err;
+      }
+    }
+
+    for (const entry of linkedModules) {
```
arcanis: This code looks redundant with the one in config.js - can't we use `this.linkedModules`?

BYK: Good catch! I think we can consolidate both code paths but I don't think it is as easy as it looks since there's some repeated logic with branching. I'll try to do it in a follow-up patch.

Issue #3824, opened 2017-07-05
yarn installs binaries for dependencies after a global add


fundon:
### Yarn

```
 λ ~ > yarn --version
0.27.5
```

### Install nodemon

nodemon dependence on touch@v1.0.0 and touch overrides the system touch command

```
 λ ~ > which nodemon
/Users/fundon/.nvm/versions/node/v8.1.3/bin/nodemon
 λ ~ > which touch
/Users/fundon/.nvm/versions/node/v8.1.3/bin/touch

# fish-shell
 λ ~ > yarn global bin
/Users/fundon/.nvm/versions/node/v8.1.3/bin
 λ ~ > echo $PATH
/Users/fundon/.nvm/versions/node/v8.1.3/bin /Users/fundon/.yarn/bin /Users/fundon/.config/yarn/global/node_modules/.bin /Users/fundon/.cargo/bin /Users/fundon/Dev/goroot/bin /Users/fundon/Library/Android/sdk/tools /Users/fundon/Library/Android/sdk/platform-tools /usr/local/bin /usr/bin /bin /usr/sbin /sbin
```

### Problem

See https://github.com/isaacs/node-touch/blob/v1.0.0/bin/touch.js#L30

```
 λ ~ > brew update
/usr/local/Homebrew/Library/Taps/michaeldfallen/homebrew-formula/.git/FETCH_HEAD Stats {
  dev: 16777222,
  mode: 33188,
  nlink: 1,
  uid: 501,
  gid: 80,
  rdev: 0,
  blksize: 4096,
  ino: 1099076,
  size: 104,
  blocks: 8,
  atimeMs: 1499220846000,
  mtimeMs: 1499220846000,
  ctimeMs: 1499220846000,
  birthtimeMs: 1487724022000,
  atime: 2017-07-05T02:14:06.000Z,
  mtime: 2017-07-05T02:14:06.000Z,
  ctime: 2017-07-05T02:14:06.000Z,
  birthtime: 2017-02-22T00:40:22.000Z }
/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/.git/FETCH_HEAD Stats {
  dev: 16777222,
  mode: 33188,
  nlink: 1,
```

BYK:
I'm not sure if `yarn` is at fault here. If there's a module named `rm` with a binary, named `rm` and if you install it globally, it would override the native `rm` command too.

This seems more like an issue with the package and global installation. What do you think @fundon?

[...]

Issue #2874, opened 2017-03-08
"bin" symlinks not created for sub-dependencies


rally25rs:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Yarn does not create `node_modules/.bin` symlinks for nested dependencies that have a [bin](https://docs.npmjs.com/files/package.json#bin) specified in their `package.json`

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Add [grunt-eslint](https://www.npmjs.com/package/grunt-eslint) to your package.json
2. `yarn install`

"grunt-eslint" depends on "eslint", which has this line in its package.json:

```
  "bin": {
    "eslint": "./bin/eslint.js"
  },
```

After install, no symlink is made in `node_modules/.bin`

**What is the expected behavior?**

If you `npm install` instead, the symlink is made.

```
$ ls -l node_modules/.bin/eslint

lrwxr-xr-x  1 user  computer\Domain Users  23 Mar  8 09:48 node_modules/.bin/eslint -> ../eslint/bin/eslint.js
```

**Please mention your node.js, yarn and operating system version.**

yarn v0.21.3
node v6.3
OSX

Daniel15:
If you want to use the `eslint` command, shouldn't your project have a direct dependency on `eslint`? Relying on transitive dependencies can be risky (for example, what if `grunt-eslint` were to remove its dependency on `eslint`?)

rally25rs:
I actually only noticed this bug because SublimeLinter-eslint plugin for SublimeText3 tries to use the local symlinked `node_modules/.bin/eslint`, so suddenly when I changed from `npm install` to `yarn install`, the linting in my editor stopped working.

-----

The real-life situation is a little more complex; I have something like 7 web UI projects are all basically clones of each other (or all based on a common framework of dependencies anyway).

All these projects share 1 "common" dependency module that sets them up (like a boilerplate template that has all the UI dependencies and grunt build stuff all set up already). So when I update a version of a library or a build step, I can change it in 1 code repo, not 7.

So all these projects just have 1 dependency that is to a private repo that I set up. That private repo has the 2-dozen actual dependencies, like eslint (this was originally done through peerDependencies before NPM stopped auto-installing those back around npm v3).

The grunt-eslint -> eslint relation just publicly demonstrates the same issue.

rally25rs:
Did a little more digging and I think this issue should have been fixed by: #1210 (Nested Executables Fix).

[...]

PR #4265, opened 2017-08-25
Disables linking transitive dependencies on global installs

arcanis:
**Summary**

The feature discussed in #2874 (linking all binaries to the top-level even if they aren't direct dependencies, in order to mimic npm's behavior) ended up breaking the global install (#3824 - if a package was globally installed and happened to depend on the `touch` package, the native `touch` binary would be changed for everyone).

In order to fix this, I made this behavior configurable, and disabled it in all global installs.

**Test plan**

Old tests should pass, plus I added a new test that makes sure that the `touch` binary isn't linked to the top-level when installing a package that depends on it (`fs-kit`).


src/package-linker.js
```diff
@@ -400,7 +406,7 @@ export default class PackageLinker {
     for (const [dest, {pkg, isDirectRequire}] of flatTree) {
       const {name} = pkg;
 
-      if (!linksToCreate.has(name) || isDirectRequire) {
+      if (isDirectRequire || (this.topLevelBinLinking && !linksToCreate.has(name))) {
```
BYK: We may wanna go one step back and prevent these from getting added to `linksToCreate` at all may be?

BYK: Oh, NVM. I'm just being stupid lol

src/cli/commands/global.js
```diff
@@ -20,6 +20,12 @@ import * as fs from '../../util/fs.js';
 const nativeFs = require('fs');
 
 class GlobalAdd extends Add {
+  constructor(...args) {
+    super(...args);
```
kaylieEB: I think this is not transpiling correctly in CI due to us using a deprecated [node 4 babel plugin](https://github.com/jbach/babel-preset-es2015-node4). I'll look into this. Otherwise we'd have to use named arguments or .apply

src/package-linker.js
```diff
@@ -48,17 +48,23 @@ export default class PackageLinker {
...
   reporter: Reporter;
   resolver: PackageResolver;
   config: Config;
+  topLevelBinLinking: boolean;
 
   setArtifacts(artifacts: InstallArtifacts) {
     this.artifacts = artifacts;
   }
 
+  setTopLevelBinLinking(topLevelBinLinking) {
```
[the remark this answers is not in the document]
BYK: May not be necessary since he added the type to the class.

PR #4268, opened 2017-08-25
Fix #4214: Align column headers for upgrade-interactive command

### LINKED ISSUE #4214: Column headers not aligned for upgrade-interactive command
When running the `upgrade-interactive`-command with the `--latest` flag, the "from", "to" and "url" column headers are off:
screen shot 2017-08-20 at 22 12 28

Someone already thought of this and made a special case when running with latest; the only problem is that the next lines will just override that:
https://github.com/yarnpkg/yarn/blob/2b09caff06151d6055705068c7edbc306fee9f68/src/cli/commands/upgrade-interactive.js#L45-L53

One solution could be to have an array of keys with dynamic length and only add the "range" key to that array if not latest. It's a bit more complex, but one win is that it will not have to loop over it for every dependency and only create the array of keys once.
```js
  const maxLengthArr = {
    name: 'name'.length,
    current: 'from'.length,
    range: 'latest'.length,
    [outdatedFieldName]: 'to'.length,
  };

  const keysWithDynamicLength = ['name', 'current', outdatedFieldName];

  if (!flags.latest) {
    maxLengthArr.range = 'range'.length;
    keysWithDynamicLength.push('range');
  }

  deps.forEach(dep =>
    keysWithDynamicLength.forEach(key => {
      maxLengthArr[key] = Math.max(maxLengthArr[key], dep[key].length);
    }),
  );
```

The more naive approach would be just to move the entire block last and override the range length. That would be more readable and easier to understand, but it would also do more work for each dependency (just in this case it might not be noticeable).
```js
 const maxLengthArr = {
    name: 'name'.length,
    current: 'from'.length,
    range: 'range'.length,
    [outdatedFieldName]: 'to'.length,
  };

  deps.forEach(dep =>
    ['name', 'current', 'range', outdatedFieldName].forEach(key => {
      maxLengthArr[key] = Math.max(maxLengthArr[key], dep[key].length);
    }),
  );

  if (flags.latest) {
    maxLengthArr.range = 'latest'.length;
  }
```

Both solutions would create the same output:
screen shot 2017-08-20 at 22 17 58

I can create a PR for any of the cases, I'm just not sure which you would prefer 🤷‍♂️

CarlRosell:
**Summary**
Fix for issue #4214 by using the first suggested solution from the issue itself. For more info see the issue 😃

Before: 
screen shot 2017-08-20 at 22 12 28
After:
screen shot 2017-08-20 at 22 17 58

### DIFF
--- a/src/cli/commands/upgrade-interactive.js
+++ b/src/cli/commands/upgrade-interactive.js
@@ -40,21 +40,19 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   const maxLengthArr = {
     name: 'name'.length,
     current: 'from'.length,
-    range: 'range'.length,
+    range: 'latest'.length,
     [outdatedFieldName]: 'to'.length,
   };
 
-  if (deps.length === 0) {
-    reporter.success(reporter.lang('allDependenciesUpToDate'));
-    return;
-  }
+  const keysWithDynamicLength = ['name', 'current', outdatedFieldName];
 
-  if (flags.latest) {
-    maxLengthArr.range = 'latest'.length;
+  if (!flags.latest) {
+    maxLengthArr.range = 'range'.length;
+    keysWithDynamicLength.push('range');
   }
 
   deps.forEach(dep =>
-    ['name', 'current', 'range', outdatedFieldName].forEach(key => {
+    keysWithDynamicLength.forEach(key => {
       maxLengthArr[key] = Math.max(maxLengthArr[key], dep[key].length);
     }),
   );
--- a/src/cli/commands/upgrade-interactive.js
+++ b/src/cli/commands/upgrade-interactive.js
@@ -37,6 +37,11 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
   const deps = await getOutdated(config, reporter, flags, lockfile, args);
 
+  if (deps.length === 0) {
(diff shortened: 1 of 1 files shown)


src/cli/commands/upgrade-interactive.js
```diff
@@ -40,21 +40,19 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
     current: 'from'.length,
-    range: 'range'.length,
+    range: 'latest'.length,
     [outdatedFieldName]: 'to'.length,
   };
 
-  if (deps.length === 0) {
-    reporter.success(reporter.lang('allDependenciesUpToDate'));
-    return;
-  }
```
kaylieEB: I don't think you meant to delete this :)

CarlRosell: Oh. Sorry for that, will fix in next commit.

CarlRosell: Also moved the check before the creation of `maxLengthArr` to return as quickly as possible

BYK:
LGTM. @kaylieEB approve and merge at will!

BYK:
@kaylieEB please edit the commit log when merging so it matches the PR log, instead of what GitHub automatically generates from each commit in the PR.

PR #2947, opened 2017-03-18
Update to CircleCI 2.0

charpeni:
Related to #2933.

**Summary**

Update CircleCI config to support 2.0.

I'm not sure about the webhooks part, it doesn't seem supported from `config.yml` in 2.0.

**Test Plan**

In my own CircleCI project

PR #4271, opened 2017-08-26
ci: Migrate to CircleCI 2.0 with Workflows

serima:
**Summary**

Migrate to CircleCI 2.0 using Workflows feature.
I have already checked #2947. I changed it to build by using workflows feature with reference to the config file. Thank them.

With this change, the total time for the build has reached 4 minutes 39 seconds. It is about 1/3 of the build time of the current master branch!
See: https://circleci.com/workflow-run/7a139f1f-f3b5-46e4-a93b-5ff914330652

There was no mention of webhooks in the document of circleci 2.0, but there was a question and answer in the forum.
The same specification also in 2.0. Just to be sure, I tried to build it as webhooks of gitter.im, but it did not matter.

https://discuss.circleci.com/t/how-do-i-configure-version-2-yml-to-have-webhooks-section/11746

**Test plan**

In my own CircleCI Project.

https://circleci.com/gh/serima/workflows/yarn/tree/feature%2Fcircleci-20-with-workflows


serima:
The reason I decided to work on this was because it seemed that there was no activity of @Daniel15 's branch `circleci-2` and this PR #2947 .
As @Daniel15 mentioned (https://github.com/yarnpkg/yarn/pull/2947#issuecomment-315271919) , we can reduce the build time using the workflows feature of CircleCI 2.0.

I think the deploy section works only when it is merged with `yarnpkg` 's master It is difficult to test, but what should I do? @BYK

BYK:
Hey @serima, thanks a lot for doing this! The speed gains look quote impressive and this will makes us go even faster. I'm out right now but I'll review your patch tomorrow.

If it needs to be pushed to a local branch for testing, I'll try that too.

Thank you again!

serima:
Thanks, @BYK! I am waiting for your feedback. Let me know if you have anything you want to ask.

BYK:
Looks pretty good to me. The only thing missing is testing multiple node versions. Ideally we'd run tests for Node 4, Node 6 and Node 8.

Is it hard to add them? Also would love to @Daniel15 to give this a shot.

BYK:
Pushed your branch and seems like the build is doing fine: https://circleci.com/gh/yarnpkg/yarn/tree/feature%2Fcircleci-20-with-workflows

Daniel15:
I'll take a look!

>  The only thing missing is testing multiple node versions. Ideally we'd run tests for Node 4, Node 6 and Node 8.

I wonder if we should try and keep the CircleCI build fast, and do all the slow stuff (like testing multiple Node versions) on Travis. In order to keep the testing of multiple Node.js versions fast, we'd need to build multiple Docker containers (one for each Node.js version). We can take the slow approach of installing it fresh on every build, but I'd rather keep the CircleCI build fast and high-signal, and leave the slow stuff in Travis.

circle.yml
```diff
@@ -1,57 +0,0 @@
...
-
-test:
-  override:
-    - node -v
-    - >
-      if [ "$CIRCLE_BRANCH" == 'master' ]; then
-        ./scripts/set-dev-version.js
-      fi;
-    - yarn lint
-    # Limit maxWorkers to 3 to avoid OOM on CircleCI
```
Daniel15: Keep this comment in the new version?

circle.yml
```diff
@@ -1,57 +0,0 @@
...
-    # Test that installing as root works and that it also works
-    # behind a user namespace which Circle CI tests are run under
-    - sudo env "PATH=$PATH" bin/yarn install --force
-
-deployment:
-  release:
-    tag: /v[0-9]+(\.[0-9]+)*/
-    owner: yarnpkg
-    commands:
-      # Only NPM is handled here - All other release files are handled in a webhook.
```
Daniel15: Keep this comment in the new version please

.circleci/config.yml
```diff
@@ -0,0 +1,99 @@
...
+          at: ~/project
+      - run:
+          name: Build distribution
+          command: |
+            node -v
+            if [ "$CIRCLE_BRANCH" == 'master' ]; then
+              ./scripts/set-dev-version.js
+            fi;
+            yarn build-dist
+            node ./scripts/build-webpack.js
```
Daniel15: There's no call to `build-webpack.js` in the old one... Why do we need it here? How was the old one doing it?

BYK:
>I wonder if we should try and keep the CircleCI build fast, and do all the slow stuff (like testing multiple Node versions) on Travis.

I don't think we need that since we need to wait for all CI builds to finish before merging anyway. I actually want to use one CI too if possible but we need Travis for macOS builds and AppVeyor for Windows.

>In order to keep the testing of multiple Node.js versions fast, we'd need to build multiple Docker containers (one for each Node.js version). 

I like that idea. Are there any downsides to this to hesitate?

Daniel15:
> I don't think we need that since we need to wait for all CI builds to finish before merging anyway.

CircleCI is what we do our releases from though, so slowing it down will slow down the release process a bit. Maybe that's not such a big deal.

Having said that... It can run these jobs in parallel, so maybe it won't actually slow down at all.

Let's merge this for now and add support for multiple Node.js versions in a separate PR 😃 

> I like that idea. Are there any downsides to this to hesitate?

We'd just need to maintain multiple Dockerfiles, but that's not such a big issue. We could create a dockerfiles directory, move the current one (https://github.com/yarnpkg/yarn/blob/master/Dockerfile.dev) into it, and create a few copies of it. You'd just need to change the base Docker image in the new Dockerfiles:
```
FROM node:7
```

Daniel15:
@BYK I can add you to the yarnpkg Docker org if you create an account on the Docker hub (https://hub.docker.com/), then you can publish Docker images for different node.js versions if you like.

serima:
> Hmm it looks like it calls the webhook multiple times (once per job, maybe?):



This is the behavior when added a webhooks to gitter.im in the webhook part on my branch for test.
Since four jobs are executed in one build, I think that webhook will be called each time one of them ends.

> I suspect they'll be fine as they should just ignore the builds with 0 artifacts

Yes, if so, I think that it will probably work without problems.

serima:
> Let's merge this for now and add support for multiple Node.js versions in a separate PR 😃

COOL. I also agree with this approach.

circle.yml
serima: ed6f133

circle.yml
serima: same as above

.circleci/config.yml
serima: Hmm, I will check. Actually we may not need to run this command here.

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Deploy
+          command: |
+            # Only NPM is handled here - All other release files are handled in a webhook.
+            if [ "${CIRCLE_PROJECT_USERNAME}" == "yarnpkg" ]; then
+              if git log -1 --pretty=%B | grep "^v[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
+                echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
+                ./scripts/set-installation-method.js "`pwd`/package.json" npm
```
serima: `./scripts/set-installation-method.js` is not found. 😨  will remove this line.

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+                npm publish --tag rc
```
serima: Do we have to run `./scripts/update-npm.sh` instead of this line?

serima:
We also need to do this:

> If you have implemented workflows on a branch in your GitHub repository, but the status check never completes, there may be status settings in GitHub that you need to reset.

https://circleci.com/docs/2.0/workflows-waiting-status/

.circleci/config.yml
serima: It is included in the processing of `yarn build-dist`

.circleci/config.yml
serima: We should use `./scripts/update-npm.sh` to controll rc/stable releases.

serima:
Hi @BYK @Daniel15, I tried the deployment process in my own repository, and found some problems. Then I did as follows (f00c2e9):

- Integrated the `deploy` section as `build_and_deploy`. This is because `dist` will be created only after build.
- Deleted redundant `node ./scripts/build-webpack.js`.
- Fixed it because the method of getting the release tag was wrong.

Please let me know if you need additional information to ship it.

serima:
By the way, the CI job failed for only NodeJS:4 of Travis CI. Could you try rebuilding?

serima:
Hey @BYK @Daniel15, Sorry to bother you again.

Fixed to run test and build parallelly without getting a deploy error.
By setting "persist_to_work" to the build section, we passed the dist directory as it is to the "deploy" image.
Compared with sequential execution, the total build time was shortened by about 2 minutes.

As usual, the build of TravisCI does not seem to be completed normally 😿  Please rebuild 🙏

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Deploy
+          command: |
+            # Only NPM is handled here - All other release files are handled in a webhook.
+            if [ "${CIRCLE_PROJECT_USERNAME}" == "yarnpkg" ]; then
+              if git describe --tags --abbrev=0 `git rev-parse HEAD` | grep "^v[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
```
Daniel15: Could this use `CIRCLE_TAG` instead? https://circleci.com/docs/2.0/env-vars/#build-details

Daniel15:
This looks pretty good to me. I wonder whether we should hold off merging this until Yarn 1.0 is out, to avoid any potential issues with the release process for 1.0. It looks safe to me, but just in case.

Thank you so much for working on it 😃

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Deploy
+          command: |
+            # Only NPM is handled here - All other release files are handled in a webhook.
+            if [ "${CIRCLE_PROJECT_USERNAME}" == "yarnpkg" ]; then
```
Daniel15: Can you confirm that this is **not** sent for pull requests, and only for builds in the repo itself?

.circleci/config.yml
serima: Oops, checking. Thanks.

.circleci/config.yml
serima: Does that mean that we do not want to publish when some feature branches are merged into master?
I want to know the release flow when publish to npm.

on master branch:

```
$ git tag - a v 0.27.0 (example)
$ git push origin master
```

and only waiting for CircleCI?

Daniel15: > I want to know the release flow when publish to npm.

Basically what you mentioned. `scripts/release-branch.sh` is what is used for new major or minor releases (eg. `0.29.0` or `1.0.0`)

We want to publish when a tag is created and the repo is the `yarnpkg/yarn` repo (that is, we need to be very careful to **not** publish a pull request from someone else's repo, for example if they tag it and then send a PR). This is likely fine with your changes though. Just trying to be cautious :)

serima: Thank you for more information.

> for example if they tag it and then send a PR

We can't include the tag we've created into the Pull Request. Therefore, there is no worry that external people tagging.
https://stackoverflow.com/questions/12278660/adding-tags-to-a-pull-request

.circleci/config.yml
serima: I do not know why, but `CIRCLE_TAG` is always empty.(like this https://discuss.circleci.com/t/build-on-tag/9864/43) Instead, I've just tried git tag execution but filtering did not work as we expected.

When I have more time I would like to settle down to study, but regarding this pull request, I want to proceed as it is.

serima:
Thanks, @Daniel15 .
> I wonder whether we should hold off merging this until Yarn 1.0 is out, to avoid any potential issues with the release process for 1.0.

Of course we will leave the timing of merging to you. I think that there is a need to pay attention to whether any problems will occur during the process of tagging and releasing.

.circleci/config.yml
tusbar: In CircleCI 2.0, this is done using the `filter` prop of the workflow jobs.

You could do something like this:

```yml
default_filters: &default_filters
  tags:
    only: '/v[0-9]+(\.[0-9]+)*/'


# …

workflows:
  version: 2
  install-test-build-and-deploy:
    jobs:
      - install:
          filters: *default_filters
      - test:
          filters: *default_filters
          requires:
            - install
      - build:
          filters: *default_filters
          requires:
            - install
      - deploy:
          filters:
            << *default_filters
            branches:
              ignore: '/.*/'
          requires:
            - test
            - build
```

Branches are opt-out and tags are opt-in. So this will run jobs for all tags matching the regexp, but only the `install`, `test` and `build` jobs will be ran for all branches.

serima: thanks! I will try it.

serima: @tusbar CircleCI tag filter is not recognized properly. I would be happy if you could tell me the specific steps of git tag release.

tusbar: @serima I’m using it successfully here: https://github.com/etalab/inspire/blob/master/.circleci/config.yml


- Here’s a successful tag workflow: https://circleci.com/workflow-run/f0b8eb73-e846-4727-b0ff-1ed724d1375f
- And a branch: https://circleci.com/workflow-run/f608ea8f-5c3e-4631-9b94-07d01bee2bf2

serima: @tusbar cool. thanks 🤓 I will try it again.

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+          paths:
+            - .cache/yarn
+      - persist_to_workspace:
+          root: ~/project
+          paths:
+            - yarn
+  test:
+    <<: *defaults
+    steps:
+      - checkout
```
tusbar: If you’re `persist_workspace`, you probably don’t have to `checkout` here, since you’re using `attach_workspace` below.

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+      - restore_cache:
+          key: yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install Dependencies
+          command: yarn install
+      - save_cache:
+          key: yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
tusbar: - The `{{ .Branch }}` specifier here seems useless as the checksum will always be more specific than the branch name.
- Also, it is always interesting to have fallbacks if the specific cache key is not found (e.g.: `yarn-`, `yarn-{{ .Branch }}-` and `yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}`)

serima: Fixed! 6ec592a

.circleci/config.yml
serima: I did not know that. thank you. it seems work well. 68d21e0

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+          name: Install Dependencies
+          command: yarn install
+      - save_cache:
+          key: v1-yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}
+          paths:
+            - .cache/yarn
+      - save_cache:
+          key: v1-yarn-{{ .Branch }}-
+          paths:
+            - .cache/yarn
```
tusbar: you don’t need to save multiple times!

Always save `v1-yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}` and in order the 3:
- `v1-yarn-{{ .Branch }}-{{ checksum "yarn.lock" }}`
- `v1-yarn-{{ .Branch }}-`
- `v1-yarn-`

The `restore_cache` command basically works as `<key>.*`, so when you load `v1-yarn-{{ .Branch }}-` it will restore the last cache that matches `v1-yarn-{{ .Branch }}-.*`

serima: Thank you for your precious advice 🙇 
Fixed. c5e4c5a

tusbar:
@serima, a word of advice, I see that you’re adding everything to the workspace. I’ve had performance issues with that in the past, I found that using the cache for node_modules was way faster.

In my `install` task I usually `rm -rf node_modules` before adding to workspace.

```yml
  install:
    <<: *default_container
    steps:
      - checkout

      - restore_cache:
          name: Restore yarn cache
          keys:
            - v1-app-yarn-{{ checksum "yarn.lock" }}
            - v1-app-yarn-

      - *restore_node_modules

      - run:
          name: Install dependencies
          command: yarn

      - save_cache:
          name: Save yarn cache
          key: v1-app-yarn-{{ checksum "yarn.lock" }}
          paths:
            - ~/.cache/yarn/

      - save_cache:
          name: Save node_modules cache
          key: v1-app-node-{{ checksum "yarn.lock" }}
          paths:
            - node_modules/

      - run:
          name: Remove node_modules to cleanup workspace
          command: rm -r node_modules/

      - persist_to_workspace:
          root: *workspace_root
          paths:
            - ./
```

Then in all of the other tasks, I restore the workspace, and the node_modules cache:

```yml
  lint:
    <<: *default_container
    steps:
      - *attach_workspace
      - *restore_node_modules

      - run:
          name: Lint JavaScipt
          command: yarn lint:scripts

      - run:
          name: Lint Stylesheets
          command: yarn lint:styles
```

with 

```yml
  attach_workspace: &attach_workspace
    attach_workspace:
      at: *workspace_root

  restore_node_modules: &restore_node_modules
    restore_cache:
      name: Restore node_modules cache
      keys:
        - v1-app-node-{{ checksum "yarn.lock" }}
        - v1-app-node-
```

.circleci/config.yml
```diff
@@ -0,0 +1,112 @@
...
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Tests
+          command: |
+            node -v
+            if [ "$CIRCLE_BRANCH" == 'master' ]; then
+              ./scripts/set-dev-version.js
+            fi;
+            yarn lint
```
tusbar: You’ll probably gain in build time if you split into multiple tasks :)

serima: You may be right. will try it.

serima: f807b7b

serima:
@tusbar Thank you for sharing knowledge. I tried it, but in this case it seems almost unchanged. The larger the size of node_modules, the more effective it may be.

* Before
  * https://circleci.com/workflow-run/fc8831ed-f123-45ba-9b4c-d88f49489a79
* After
  * https://circleci.com/workflow-run/5a9d6a07-8319-464d-a798-246294840214

serima:
I checked this pull request #4486 , it seems that we switched to the test on node 8.
If possible, I think that switching Docker image of `yarnpkg/dev` to node 8 is better before mergind this pull request.

serima:
Resolved conflicts 😸

serima:
Oops, a test failed after merged master.

```
Summary of all failing tests
 FAIL  __tests__/integration.js (55.486s)
  ● yarn add react-scripts@1.0.13

    expect(received).not.toMatch(expected)
    
    Expected value not to match:
      /^warning /gm
    Received:
      "warning You are using Node \"7.10.1\" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: \"^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0\""
      
      at __tests__/integration.js:82:1793
```

> If possible, I think that switching Docker image of yarnpkg/dev to node 8 is better before mergind this pull request.

It seems we have to do this.

serima:
@Daniel15 Could you update `yarnpkg/dev` image first?
I created the original image changed to `FROM node:8` and tried it.

https://hub.docker.com/r/serima/yarnpkg-node8/
https://circleci.com/gh/serima/yarn/260

and it passed.

Daniel15:
@serima I pushed an  update to the Docker image.

serima:
Thanks a lot! CI passed!

serima:
@tusbar Is there anything better to improve? If not, please approve 😄

.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
...
+      - v1-node-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - v1-node-{{ .Branch }}-
+      - v1-node-
+
+jobs:
+  install:
+    <<: *defaults
+    steps:
+      - checkout
+      - restore_cache:
```
tusbar: you can use `- *restore_node_modules` here :)

serima:
I am waiting patiently for merging 😟  Is there anything I can do for that?

tusbar:
@serima I just came up with this: https://github.com/etalab/geo.data.gouv.fr/pull/532 – this is actually cleaner than removing the node_modules manually, and actually runs slightly faster.

BYK:
@serim - oops, sorry for the wait. I didn't realize this was finished since CI status was red. I now see that it was a hiccup with Travis.

@Daniel15 - can you give this one last look and merge if you think it looks good?

arcanis:
@Daniel15 Seems like the update broke the releases :( Looking at the diff, the only thing that looks fishy is this:

https://github.com/yarnpkg/yarn/pull/4271/files#diff-1d37e48f9ceff6d8030570cd36286a61R137

Have you checked that CircleCI doesn't match "master" against this regex? Since it's technically a branch as well.

arcanis:
#4834 seems to have fixed the missing deploy step, however we do it 500 errors: cc @Daniel15 

https://circleci.com/gh/yarnpkg/yarn/5797

Daniel15:
@arcanis The 500 error looks like a bug in the archive script from when I refactored some stuff:

```
ErrorException: Undefined variable: build_num in /var/www/nightly.yarnpkg.com/nightly/public/archive_circleci.php:19
```

https://github.com/yarnpkg/release-infra/blob/master/nightly/public/archive_circleci.php#L19

Looks like it'll be thrown when the build status is 'invalid' (ie. a failed build) as I forgot to refactor one of the variables.

I filed an issue for it: https://github.com/yarnpkg/release-infra/issues/21

Daniel15:
1.3.2 worked: https://github.com/yarnpkg/yarn/releases/tag/v1.3.2

serima:
I am sorry for this 😢 Thanks all to fix the bug.

PR #4273, opened 2017-08-26
Replace deprecated `babel-preset-node4` and `babel-preset-node5` with preset-env

kaylieEB:
**Summary**
Current setup of presets are deprecated ([babel-preset-node5](https://github.com/leebenson/babel-preset-node5), [babel-preset-es2015-node4](https://github.com/jbach/babel-preset-es2015-node4)) and haven't been updated with the latest stages/plugins.

Particular issue was seen w/ using `constructor(...args) {super(...args)}`. 

> For example, Node 4 supports native classes but not spread. If super is used with a spread argument, then the transform-es2015-classes transform needs to be included, as it is not possible to transpile a spread with super otherwise.
[source](https://github.com/babel/babel-preset-env#include)

Last time we tried to use this, the size of artifacts was a concern, so I excluded plugins not being used and got the size to approximately be where they were before.

BEFORE
screen shot 2017-08-26 at 1 37 28 pm

AFTER
screen shot 2017-08-26 at 4 31 43 pm

**Test plan**
CI builds!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -27,13 +27,17 @@
           "targets": {
             "node": 4
           },
-          "loose": true
+          "loose": true,
+          "exclude": [
+            "transform-regenerator"
+          ]
         }],
         "flow",
         "stage-0"
       ],
       "plugins": [
-        ["transform-es2015-modules-commonjs"]
+        ["transform-inline-imports-commonjs"],
+        ["transform-runtime", { "polyfill": true, "regenerator": false }]
       ]
     }
   },
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,19 @@
   "only": "*.js",
   "env": {
     "pre-node5": {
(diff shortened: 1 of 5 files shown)


kaylieEB:
wip, need further investigation on how to make test pass on ci without relying on babel-polyfill

BYK:
Thanks a lot @kaylieEB! Anything you need from us?

kaylieEB:
Thanks @byk! I think I need to take the time to get docker working because it was building / passing fine in my local env but ci isn't too happy.

BYK:
@kaylieEB this may be an issue with older versions of Node (6 and 4) which we test against. I have Node 8 locally and your branch works fine for me.

kaylieEB:
moved to https://github.com/yarnpkg/yarn/pull/4374 (got tests working with node 6 builds 🎉 )

Issue #4276, opened 2017-08-27
Cannot remove dependency from workspace with yarn


yacinehmito:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Install a new dependency in a workspace, then try to uninstall it with `yarn remove`. You'll have the following output:

```
yarn remove v0.28.4
error No lockfile in this directory. Run `yarn install` to generate one.
```

It is however perfectly normal for a workspace to not have a lock file.

Removing from the root folder doesn't work either (which is expected).

**What is the expected behavior?**

There shouldn't be an error message. Like the normal case, yarn should:

- Remove the dependency from the workspace's `package.json`
- Link dependencies to trim the root folder's `yarn.lock`
- Delete the appropriate packages from the relevant `node_modules` folders

The RFC [briefly documents](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-workspaces-install-phase-1.md#running-yarn-install-in-workspaces-folders) the expected behaviour.

**Environment**

- yarn v0.28.4
- node v8.3.0
- Linux v4.9.44 (NixOS unstable distro)

arcanis:
Can you try the [nightly builds](http://nightly.yarnpkg.com/)?

yacinehmito:
Sure. I got the same error:

```
yarn remove v1.0.0-20170825.2124
error No lockfile in this directory. Run `yarn install` to generate one.
```

The slight difference is that this time the process doesn't exit, so the shell effectively hangs. This is however unrelated to this current issue.

yacinehmito:
I am currently writing a test and will submit a PR once I'm done.

yacinehmito:
Test submitted in #4277.

It doesn't fail, which means the problem is unrelated to the removing functionality itself. I don't know enough about the yarn codebase to keep investigating but my bet is that there's a check somewhere for the lock file that we need to modify. Instead of checking the current folder for a `yarn.lock` file, it should walk up the folder hierarchy.

kaylieEB:
FWIW I don't think it's specific to workspaces. The command seems to hang in a normal package without a lockfile.

arcanis:
[...]

PR #4277, opened 2017-08-27
Adding a test that checks remove in a workspace

yacinehmito:
I wanted to write a failing test for #4276. Turns out the test passes so the problem might not be related to the remove functionality itself. I'm still sharing the test though.

### DIFF
--- a/__tests__/commands/remove.js
+++ b/__tests__/commands/remove.js
@@ -100,6 +100,14 @@ test.concurrent('removes a single scoped package', (): Promise<void> => {
   });
 });
 
+test.concurrent("removes a workspace's dependency", (): Promise<void> => {
+  return runRemove(['left-pad'], {}, 'workspaces-remove-workspace-dependency/workspace-child', async (config): Promise<
+    void,
+  > => {
+    expect(JSON.parse(await fs.readFile(path.join(config.cwd, 'package.json'))).dependencies).toEqual({});
+  });
+});
+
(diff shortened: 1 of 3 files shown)


__tests__/commands/remove.js
```diff
@@ -100,6 +100,14 @@ test.concurrent('removes a single scoped package', (): Promise<void> => {
   });
 });
 
+test.concurrent("removes a workspace's dependency", (): Promise<void> => {
+  return runRemove(['left-pad'], {}, 'workspaces-remove-workspace-dependency/workspace-child', async (config): Promise<
```
arcanis: This test probably doesn't make what you want - the test framework will copy the folder referenced as third parameter, and by doing so will "forget" that "workspace-child" is a child of a "workspaces-remove-workspace-dependency".

yacinehmito: Understood. How would you go about writing such a test?

arcanis: Just use runInstall using the base fixture, then manually run remove (`new Remove(...)`) using the `{cwd: "..."}` option to execute it in the workspace subfolder.

BYK:
@yacinehmito  - are you interested in updating this or shall we close the PR?

yacinehmito:
Will do this weekend. If I don't before Monday feel free to close.

BYK:
@yacinehmito - Closing due to inactivity. LMK if you're still interested.

Issue #4263, opened 2017-08-25
`yarn upgrade-interactive` doesn't work in Jest


cpojer:
In the Jest repo, run `yarn upgrade-interactive`. It will not update any packages that are selected, but instead re-link (?) the workspaces.

rally25rs:
That's weird. 0.27.5 or 1.0.0? In the end it just feeds the package names to `add`

```
      const deps = answers.filter(isHint(hint)).map(getPattern);
      if (deps.length) {
        reporter.info(reporter.lang('updateInstalling', getNameFromHint(hint)));
        const add = new Add(deps, flags, config, reporter, lockfile);
        return add.init();
      }
```

which should be functionally equivalent to `yarn add whatever@^x.y.z something@^x.y.z morethings@^x.y.z`

so that would make it seem like `add` isn't updating the existing package.

cpojer:
I was on master as of yesterday.

kaylieEB:
Could verify this odd behavior on master. After choosing to upgrade typescript, it seems to be installing its own packages? Same result regardless of package that's chosen to be upgraded.
```
Choose which packages to update. typescript@2.5.1
info Installing "devDependencies"...
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "ajv-keywords@1.5.1" has incorrect peer dependency "ajv@>=4.10.0".
warning "karma-browserify@5.1.1" has unmet peer dependency "watchify@>=3 <4".
[4/4] 📃  Building fresh packages...
success Saved 34 new dependencies.
├─ babel-jest@20.0.3
├─ babel-plugin-jest-hoist@20.0.3
├─ babel-preset-jest@20.0.3
├─ eslint-config-fb-strict@20.0.3
├─ eslint-plugin-jest@20.0.3
├─ jest-changed-files@20.0.3
├─ jest-circus@20.0.3
├─ jest-cli@20.0.4
├─ jest-config@20.0.4
├─ jest-diff@20.0.3
├─ jest-docblock@20.0.3
├─ jest-editor-support@20.0.4
├─ jest-environment-jsdom@20.0.3
├─ jest-environment-node@20.0.3
├─ jest-get-type@20.0.3
├─ jest-haste-map@20.0.4
├─ jest-jasmine2@20.0.4
├─ jest-matcher-utils@20.0.3
├─ jest-matchers@20.0.3
├─ jest-message-util@20.0.3
[...]

PR #4278, opened 2017-08-27
Fix 4263: Make upgrade and upgrade-interactive compatible with workspaces

kaylieEB:
**Summary**
Fix for https://github.com/yarnpkg/yarn/issues/4263

When called at the project root level with workspaces, upgrade and upgrade-interactive should only change the root dev dependencies. Previously it was resolving / reinstalling workspace dependencies as well.  Also it didn't allow for existing dev dependency to update (see screenshot)

**Test plan**

Added a test in `upgrade` with workspaces

*BEFORE* (only chalk should be updated)

screen shot 2017-08-27 at 12 28 08 pm

*AFTER*

screen shot 2017-08-27 at 12 25 45 pm

*BEFORE* (typescript is a dev dependency)

screen shot 2017-08-27 at 12 25 23 pm

*AFTER*

screen shot 2017-08-27 at 12 25 32 pm


src/cli/commands/install.js
```diff
@@ -284,7 +284,7 @@ export class Install {
       pushDeps('devDependencies', projectManifestJson, {hint: 'dev', optional: false}, !this.config.production);
       pushDeps('optionalDependencies', projectManifestJson, {hint: 'optional', optional: true}, true);
 
-      if (this.config.workspaceRootFolder) {
+      if (this.config.workspaceRootFolder && !this.ignoreWorkspaces) {
```
arcanis: Hm, doesn't this break running `yarn install` at the root of the workspace? It won't install the workspaces dependencies anymore, and its own dependencies won't have the same layout than what they would have been if the hoister had been aware of the workspaces.

kaylieEB: `this.ignoreWorkspaces` is only set on `add` (defaults to false), so install should be ok. I'd also imagine this test (https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/workspaces-install.js#L71) would catch it?

arcanis: Oh I see! Ok, I missed that `ignoreWorkspaces` was undefined in `Install`. I think you should make it a flag defined in `Install` (rather than `Add`) and add a setter to toggle it (like the `topLevelBinLinkFlag` implemented [here](https://github.com/yarnpkg/yarn/pull/4265/files#diff-cebc41615330398e66d604a4a42bcd0e)).

kaylieEB: you got it!

src/cli/commands/upgrade-interactive.js
```diff
@@ -137,9 +137,11 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       flags.dev = hint === 'dev';
       flags.peer = hint === 'peer';
       flags.optional = hint === 'optional';
-
       const deps = answers.filter(isHint(hint)).map(getPattern);
       if (deps.length) {
+        for (const pattern of deps) {
+          lockfile.removePattern(pattern);
+        }
```
kaylieEB: upgrade-interactive seemed to have been previously broken, regardless of workspaces or not. When `--latest` flag wasn't being passed, it needed https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade.js#L84-L86, like upgrade, otherwise nothing would get re-installed.

src/cli/commands/add.js
```diff
@@ -32,6 +36,10 @@ export class Add extends Install {
     ]
       .filter(Boolean)
       .shift();
+
+    if (flags.existing) {
+      this.flagToOrigin = '';
+    }
```
arcanis: Why is this required? Isn't it already handled by [this line](https://github.com/kaylieEB/yarn/blob/93b2aaeb33aad23218177f3a1ed2bd9870e4d077/src/cli/commands/add.js#L196) (according to the comment above, `depType` is the dependency source when upgrading a package)? I think it also might break the warning [here](https://github.com/kaylieEB/yarn/blob/93b2aaeb33aad23218177f3a1ed2bd9870e4d077/src/cli/commands/add.js#L132-L134), since the string will now always be empty

src/cli/commands/add.js
```diff
@@ -120,7 +128,7 @@ export class Add extends Install {
    */
 
   async init(): Promise<Array<string>> {
-    if (this.config.workspaceRootFolder && this.config.cwd === this.config.workspaceRootFolder) {
+    if (this.ignoreWorkspaces) {
```
arcanis: Unless I'm mistaken, this condition cannot be replaced this way; it is meant to ensure that running `yarn add <something>` inside the workspace root will warn the user that they probably don't want to do this. With this change, no warning will be emitted in such a case because `ignoreWorkspaces` will be `false` in traditional `add` settings. I think a better condition is:

```js
this.config.workspaceRootFolder && this.config.cwd === this.config.workspaceRootFolder && !this.ignoreWorkspaces
```

src/cli/commands/add.js
kaylieEB: So the issue I was running into was that when I ran `yarn upgrade typescript` (a dev dependency) in the workspace root directory, it would incorrectly warn me that workspaces prefer dev dependencies. This is because `flagsToOrgin` defaults to being a hard dependency when flags aren't passed. I don't think yarn should ever throw that warning if the command is an upgrade and save it for when it's a brand new add (hence the check for `flags.existing` passed in the upgrade file below).

src/cli/commands/add.js
kaylieEB: That conditional wouldn't ever be true though according to this logic right? https://github.com/kaylieEB/yarn/blob/93b2aaeb33aad23218177f3a1ed2bd9870e4d077/src/cli/commands/add.js#L27-L29. I've also tested that it still warns when I try to add a non-dev dependency.

src/cli/commands/add.js
BYK: An inline comment about this would have been great.

cpsubrian:
How would one run `upgrade-interactive` inside a particular workspace? The lack of lockfile prevents it.

PR #4283, opened 2017-08-29
Fix: Always close the reporter before exiting

### LINKED ISSUE #4276: Cannot remove dependency from workspace with yarn

arcanis:
**Summary**

Fix #4276

In some particular cases, the reporters are not correctly closed before Yarn exits. It makes the process hang, because the reporters are expected to periodically check for the RAM usage. This patch does two things:

- First the setInterval loop itself is `unref`, so that it will never prevent the process from exiting.
- More importantly, the reporter is now correctly closed when exiting in the two cases I've noticed.

**Test plan**

I don't think tests are required for this fix, since the reproduction steps are fairly narrow (the bug only occurred when using Yarn in an invalid way) and won't allow us to catch any other bug than this one.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -176,7 +176,7 @@ export function main({
   if (command.noArguments && commander.args.length) {
     reporter.error(reporter.lang('noArguments'));
     reporter.info(command.getDocsInfo);
-    process.exitCode = 1;
+    exit(1);
     return;
   }
 
@@ -193,7 +193,7 @@ export function main({
   //
   if (command.requireLockfile && !fs.existsSync(path.join(config.cwd, constants.LOCKFILE_FILENAME))) {
     reporter.error(reporter.lang('noRequiredLockfile'));
-    process.exitCode = 1;
+    exit(1);
     return;
   }
 
@@ -411,10 +411,10 @@ export function main({
     return errorReportLoc;
   }
 
-  const exit = exitCode => {
+  function exit(exitCode) {
     process.exitCode = exitCode || 0;
-    return reporter.close();
-  };
+    reporter.close();
+  }
 
   const cwd = findProjectRoot(commander.cwd);
 
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -164,6 +164,11 @@ export function main({
     isSilent: process.env.YARN_SILENT === '1' || commander.silent,
   });
 
+  const exit = exitCode => {
+    process.exitCode = exitCode || 0;
+    reporter.close();
+  };
+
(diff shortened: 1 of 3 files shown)


arcanis:
Hm Flow seems to recognize `setInterval` as returning a number, which is true for browsers but not for Node. I'll add a `$FlowFixMe`.

Re: tests, if you think it's better I'll add one.

BYK:
If it is not too much trouble, a test would be great. If not, we'll... 🤷‍♂️

src/reporters/base-reporter.js
```diff
@@ -152,6 +152,8 @@ export default class BaseReporter {
     this.peakMemoryInterval = setInterval(() => {
       this.checkPeakMemory();
     }, 1000);
+    // $FlowFixMe: Node's setInterval returns a Timeout, not a Number
+    this.peakMemoryInterval.unref();
```
BYK: I'm 👍 with this change but curious why `clearInterval(this.peakMemoryInterval);` in `close()` below isn't enough.

src/cli/index.js
```diff
@@ -349,10 +349,10 @@ export function main({
     return errorReportLoc;
   }
 
-  const exit = exitCode => {
+  function exit(exitCode) {
```
BYK: Why convert to "normal" functions from arrow functions?

arcanis: Because otherwise `exit` is undefined at lines 178 & 195, since the assignment hasn't been executed yet.

BYK: Ah, let's move this above then? Or add a comment if moving up is not possible?

yacinehmito:
I don't think it fixes #4276, although this is a welcome PR.
The problem with the process hanging was just mentioned in passing. The real issue is that `yarn remove` in a workspace doesn't work.

PR #4286, opened 2017-08-30
Refactors the network mutex implementation

arcanis:
**Summary**

I'm currently debugging a concurrency issue that I noticed when running some internal tests, and the current mutex implementation makes it difficult to know exactly what's happening, because we only know that the process are waiting, with no indication as to *who* they're waiting.

This PR solves this issue by refactoring the mutex to now be an actual webserver (no extra dependency is required, I simply use the `http` native module). When a process fails to create a server, it sends an http request to the one already existing, and print the response (usually the active process `$PWD`) to its standard output as part of the warning message. Once this is done, it opens a new raw socket to the server, and patiently wait for it to be terminated (like in the current workflow).

screen shot 2017-08-30 at 3 29 35 pm

Note that pretty much anything can now be printed: I've chosen the `$CWD` since we're not supposed to run Yarn concurrently in the same directory, but the command line arguments and/or the PID might also be good candidates.

**Test plan**

This feature should be partially covered by the [already existing test](https://github.com/yarnpkg/yarn/blob/master/__tests__/integration.js#L70-L83), but I'm open to write new ones.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -58,11 +58,13 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
 
 const getRandomPort = () => Math.floor(Math.random() * PORT_RANGE) + MIN_PORT_NUM;
 
-function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
-  const {stderr, stdout} = execa(path.resolve(__dirname, '../bin/yarn'), args, options);
+async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
+  const process = execa(path.resolve(__dirname, '../bin/yarn'), args, options);
 
-  const stdoutPromise = misc.consumeStream(stdout);
-  const stderrPromise = misc.consumeStream(stderr);
+  const stdoutPromise = misc.consumeStream(process.stdout);
+  const stderrPromise = misc.consumeStream(process.stderr);
+
+  await process;
 
   return Promise.all([stdoutPromise, stderrPromise]);
 }
@@ -80,6 +82,29 @@ test('--mutex network', async () => {
     execa(command, ['add', 'left-pad'].concat(args), options),
     execa(command, ['add', 'foo'].concat(args), options),
   ]);
+});
+
+test('--mutex network 2', async () => {
+  const cwd = await makeTemp();
+  await fs.writeFile(path.join(cwd, '.yarnrc'), '--mutex network\n');
+
+  const promises = [];
+
+  for (let t = 0; t < 100; ++t) {
+    const subCwd = path.join(cwd, String(t));
+
+    await fs.mkdirp(subCwd);
+    await fs.writeFile(
+      path.join(subCwd, 'package.json'),
+      JSON.stringify({
+        scripts: {test: 'node -e "setTimeout(function(){}, process.argv[1])"'},
+      }),
+    );
+
+    promises.push(runYarn(['run', 'test', '100'], {cwd: subCwd}));
+  }
+
+  await Promise.all(promises);
 });
 
 test('--cwd option', async () => {
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -90,7 +90,7 @@ test('--mutex network 2', async () => {
 
   const promises = [];
 
-  for (let t = 0; t < 100; ++t) {
+  for (let t = 0; t < 40; ++t) {
     const subCwd = path.join(cwd, String(t));
 
     await fs.mkdirp(subCwd);
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -59,34 +59,16 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
 const getRandomPort = () => Math.floor(Math.random() * PORT_RANGE) + MIN_PORT_NUM;
 
 async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
-  const process = execa(path.resolve(__dirname, '../bin/yarn'), args, options);
+  const {stdout, stderr} = await execa(path.resolve(__dirname, '../bin/yarn'), args, options);
 
-  const stdoutPromise = misc.consumeStream(process.stdout);
-  const stderrPromise = misc.consumeStream(process.stderr);
-
-  await process;
-
-  return Promise.all([stdoutPromise, stderrPromise]);
+  return [stdout, stderr];
 }
 
 test('--mutex network', async () => {
   const cwd = await makeTemp();
-  const cacheFolder = path.join(cwd, '.cache');
 
-  const command = path.resolve(__dirname, '../bin/yarn');
-  const args = ['--cache-folder', cacheFolder, '--verbose', '--mutex', `network:${getRandomPort()}`];
-
-  const options = {cwd};
-
-  await Promise.all([
-    execa(command, ['add', 'left-pad'].concat(args), options),
-    execa(command, ['add', 'foo'].concat(args), options),
-  ]);
-});
-
-test('--mutex network 2', async () => {
-  const cwd = await makeTemp();
-  await fs.writeFile(path.join(cwd, '.yarnrc'), '--mutex network\n');
+  const port = 56000 + Math.floor(Math.random() * 100);
+  await fs.writeFile(path.join(cwd, '.yarnrc'), `--mutex "network:${port}"\n`);
 
   const promises = [];
 
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -307,6 +307,7 @@ export function main({
             console.error('Process stalled');
             if (process._getActiveHandles) {
               console.error('Active handles:');
+              // $FlowFixMe: getActiveHandles is undocumented, but it exists
               for (const handle of process._getActiveHandles()) {
                 console.error(`  - ${handle.constructor.name}`);
               }
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -17,6 +17,7 @@ const fs = require('fs');
 const invariant = require('invariant');
 const lockfile = require('proper-lockfile');
 const loudRejection = require('loud-rejection');
+const http = require('http');
 const net = require('net');
 const onDeath = require('death');
 const path = require('path');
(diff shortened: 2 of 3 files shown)


src/cli/index.js
```diff
@@ -234,8 +235,104 @@ export function main({
...
+              resolve(res);
+          }, err => {
+              killSockets();
+              reject(err);
+          });
+        });
+
+        function manager(request, response) {
+          response.writeHead(200);
+          response.write(config.cwd);
```
BYK: I'd add pid too when it's available.

src/cli/index.js
```diff
@@ -234,8 +235,104 @@ export function main({
...
+          response.on('error', () => {
+            tryServer();
+          });
+
+          response.on('data', buffer => {
+            buffers.push(buffer);
+          });
+
+          response.on('end', () => {
+            reporter.warn(reporter.lang('waitingNamedInstance', Buffer.concat(buffers).toString()));
```
BYK: I'd say this shouldn't be a warning.

src/cli/index.js
```diff
@@ -234,8 +235,104 @@ export function main({
...
+          }
+        }
+      }
+
+      function reportServerName() {
+        const request = http.get({ port: mutexPort }, response => {
+          const buffers = [];
+
+          response.on('error', () => {
+            tryServer();
```
BYK: Where is this function defined?

src/cli/index.js
arcanis: I agree, but it's a warning in the current code

src/cli/index.js
```diff
@@ -234,8 +235,104 @@ export function main({
...
+      const connectionOptions = {
+        port: +mutexPort || constants.SINGLE_INSTANCE_PORT,
+      };
+
+      function startServer() {
+        const clients = new Set();
+        const server = http.createServer(manager);
+
+        // The server must not prevent us from exiting
+        server.unref();
```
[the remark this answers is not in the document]
arcanis: It should, but it didn't work when I simply added it to the build

src/cli/index.js
arcanis: Typo, should be `startServer`

arcanis:
> I don't think this would work since http is designed to close the TCP connection as soon as possible.

It can be [disabled](https://github.com/yarnpkg/yarn/pull/4286/files#diff-867becf4a9c2c6c6d4e7c1278750724eR252), so that no socket will timeout by default.

> Why not use the same logic with plain TCP sockets?

My reasoning was that at least two sockets would be needed (one for the data, one for the keepalive), or a custom protocol would have to be implemented on top of TCP (to know when the server finished sending data).

In both cases I felt it was simpler and less error prone to just use http to communicate between socket and clients. Less logic in our codebase, more guarantees that we don't forget some edge case!

BYK:
I think that's fair. I just don't like the mutex code getting more involved but currently, we need this to debug the issues.

src/cli/index.js
BYK: May be we can use the forced log here too?

src/cli/index.js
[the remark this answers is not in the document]
BYK: Then may be the problem is not with the mutex? 🤔

arcanis:
I noticed there was no stress test on the mutex code. I added one (100 concurrent processes, it checks they all exit properly).

__tests__/integration.js
```diff
@@ -58,11 +58,13 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
...
-  const {stderr, stdout} = execa(path.resolve(__dirname, '../bin/yarn'), args, options);
+async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
+  const process = execa(path.resolve(__dirname, '../bin/yarn'), args, options);
 
-  const stdoutPromise = misc.consumeStream(stdout);
-  const stderrPromise = misc.consumeStream(stderr);
+  const stdoutPromise = misc.consumeStream(process.stdout);
+  const stderrPromise = misc.consumeStream(process.stderr);
+
+  await process;
```
BYK: This doesn't look right to me since it looks like `stdout` and `stderr` are only available after `await`ing `process`: https://github.com/sindresorhus/execa/blob/master/index.js#L259

I guess the following is what we need:
```
  const {stdout, stderr} = await execa(path.resolve(__dirname, '../bin/yarn'), args, options);
  
  return Promise.all([
    misc.consumeStream(stdout),
    misc.consumeStream(stderr),
  ]);
```

If I'm missing something, may be a comment here explaining the ordering would help?

__tests__/integration.js
```diff
@@ -82,6 +84,29 @@ test('--mutex network', async () => {
   ]);
 });
 
+test('--mutex network 2', async () => {
```
BYK: Let's name this `'stress test for "--mutex network"'`?

Alternatively, just use this one remove the old test if it is not enough/useful since this should cover that scenario already.

__tests__/integration.js
```diff
@@ -82,6 +84,29 @@ test('--mutex network', async () => {
...
+  const cwd = await makeTemp();
+  await fs.writeFile(path.join(cwd, '.yarnrc'), '--mutex network\n');
```
BYK: You should probably use a random port for each run to make sure we don't clash with any parallel tests.

__tests__/integration.js
```diff
@@ -82,6 +84,29 @@ test('--mutex network', async () => {
...
+
+    await fs.mkdirp(subCwd);
+    await fs.writeFile(
+      path.join(subCwd, 'package.json'),
+      JSON.stringify({
+        scripts: {test: 'node -e "setTimeout(function(){}, process.argv[1])"'},
+      }),
+    );
+
+    promises.push(runYarn(['run', 'test', '100'], {cwd: subCwd}));
```
BYK: This made me realize that we should probably limit `mutex` to places where we touch cache. It doesn't make much sense to block `yarn run` calls with a mutex, right? (future breaking TODO)

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
-                resolve(runEventuallyWithNetwork(mutexPort));
-              });
-            }
-          })
-          .on('error', () => {
-            // No server to listen to ? Let's retry to become the next server then.
-            process.nextTick(() => {
-              resolve(runEventuallyWithNetwork(mutexPort));
-            });
+        // No socket must timeout
```
BYK: `// We are going to hold all connections until this current process is finished so disable timeouts`

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
+          // This timeout must not prevent us from exiting
+          // $FlowFixMe: Node's setTimeout returns a Timeout, not a Number
+          timeout.unref();
         }
+      }
 
-        await server.close();
-        server.unref();
-      };
+      function reportServerName() {
```
BYK: May be rename this to `connectToExistingMutexServer`?

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
-            });
+        // No socket must timeout
+        server.timeout = 0;
+
+        // If we fail to setup the server, we ask the existing one for its name
+        server.on('error', () => {
+          reportServerName();
+        });
+
+        // If we succeed, keep track of all the connected sockets to delete them later
```
BYK: s/delete/close/

src/cli/index.js
BYK: What do you think about this one?

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
+              killSockets();
+              reject(err);
+            },
+          );
+        });
+
+        function manager(request, response) {
+          response.writeHead(200);
+          response.write(config.cwd);
+          response.end();
```
BYK: You can combine this with the line above: `response.end(config.cwd)` - https://nodejs.org/api/http.html#http_response_end_data_encoding_callback

That said, I'm a bit concerned about this part since `response.end` would close the connection. Why don't we just do `response.write()`, omit the `response.end()` call? That way we wouldn't need `reportServerName` and `waitForNetwork` together. We can merge their logic into one.

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
+            waitForTheNetwork();
+          });
+        });
 
-        socket.on('close', () => clients.delete(socket));
-      });
+        request.on('error', () => {
+          startServer();
+        });
+      }
```
BYK: Looks like you don't call `request.end()` in anywhere. According to Node docs, it should always be invoked:

https://nodejs.org/api/http.html#http_http_request_options_callback
>Note that in the example req.end() was called. With http.request() one must always call req.end() to signify the end of the request - even if there is no data being written to the request body.

src/cli/index.js
```diff
@@ -234,68 +235,129 @@ export function main({
...
+          timeout.unref();
         }
+      }
 
-        await server.close();
-        server.unref();
-      };
+      function reportServerName() {
+        const request = http.get(connectionOptions, response => {
+          const buffers = [];
```
BYK: I'd move this closer to where it is first used: right above `response.on('data'` to clarify the reason for it.

__tests__/integration.js
arcanis: Yep, I also witnessed this error first-hand last week, an internal build was failing exactly because of this. I solved it temporarily by restricting the option range to `install`, `add`, `remove`, and `upgrade`, but a more robust solution would to give commands the ability to export a variable when they need to be "mutex-aware", kinda like what we're doing with [`noArgument`](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/import.js#L33).

src/cli/index.js
arcanis: Doing this would require us to implement our own protocol on top of HTTP to detect that we received the full server message and that we can print it, plus 'hacking' the HTTP protocol to keep the request alive, which is not usual. As simple as it can be, I strongly prefer keeping the logic separated: one function to get information, one function to get a keepalive.

src/cli/index.js
arcanis: I use `http.get()`, not `http.request()`.

> The only difference between this method and `http.request()` is that it sets the method to `GET` and calls `req.end()` automatically.

src/cli/index.js
arcanis: With the split logic, I prefer the names to reflect what is the connection doing

src/cli/index.js
BYK: Ah, that's a very good reason: we don't have any means to detect the actual flushing. Looks like the ideal solution here would have been websockets, if they came built-in to node.js

Anyway, I think a short comment above the `response.end()` call saying something like the following would be great:```
// Call `.end()` to ensure the response is flushed.
```

That said I read the docs for `response.write()` and it reads like node.js always flushes each write: https://nodejs.org/api/http.html#http_response_write_chunk_encoding_callback

I'm not unhappy with the current implementation, I just find it a bit hairy and hacky since our second connection is a raw TCP connection and relies on the client technically DoS'ing the mutex by not responding with an ACK.

src/cli/index.js
BYK: Yeah, I haven't realized the split until a bit later. I still think this should be named `getServerName` or something to express what it is doing. You can argue that `reportServerName` also does that from the CLI viewpoint but I feel like the actual goal here is to "get" the name from the mutex to make sure it is valid, right?

PR #4129, opened 2017-08-09
Fix: invalid linking in workspaces for unused packages

BYK:
**Summary**

Fixes #4106. When a local package is listed under `workspaces` but that specific version is never used by anything, `yarn` still tries to install its bin links because it is listed as a dependency of the virtual workspace aggregator (but never get installed). This patch makes sure we have a location inside `node_modules` when trying to create bin links.

**Test plan**

Added a new test case which fails on master and passes with the fix.

Manual:
```
git clone git@github.com:babel/babel.git
yarn
```

Fails before the patch with latest nightly and passes after the patch.

Issue #4041, opened 2017-07-27
yarn why - couldn't find package workspace-aggregator-[UUID]


martin-css:
**Do you want to request a *feature* or report a *bug*?**
I think this is a bug - *only applies with the experimental yarn workspaces enabled*.

I was attempting to get the new experimental Yarn workspaces working in my mono repo and discovered `yarn why` isn't working. Issue  #3841 referred to `yarn list` which is working. I wonder if further updates are required for `why` (and possibly any other relevant commands).

**What is the current behavior?**

Running `yarn why` fails with the error below:
 
```
error Couldn't find package "workspace-aggregator-2e092624-1b04-4de1-9847-260953ff91b5" on the "npm" registry.
```
The v4  UUID is random and will change each time.

**If the current behavior is a bug, please provide the steps to reproduce.**

The shell steps below will create a contrived simple repo that reproduces the problem.

```
rm -rf repo
mkdir repo
cd repo
mkdir -p packages/p1
echo '{"name":"repo", "private":true, "workspaces":["packages/*"]}' >package.json
echo '{"name":"p1", "version": "0.0.1"}' >packages/p1/package.json
yarn install
yarn why p1
```

**What is the expected behavior?**

Yarn should show information on why package is installed (i.e. it is package present in the workspace). It should not fail with an error.

**Please mention your node.js, yarn and operating system version.**

Component | Version
--------------- | ---------
Node.JS | v8.2.1
Yarn | v0.28.4
OS | Docker container  `node:8.2.1`

cpojer:
I'm also getting this on Jest. What is this package? cc @bestander

arcanis:
It's a "fake" package that is created during the internal workspace resolution. It should never be publicly visible, but we need to add barriers in some places :)

BYK:
Closed by #4304.

levrik:
Just saw this when doing a `yarn install`:

![image](https://user-images.githubusercontent.com/9491603/34960027-6d309690-fa39-11e7-96e1-c05708ab0896.png)

sabrehagen:
And again on `v1.12.3`

![image](https://user-images.githubusercontent.com/1094600/50371275-a7bb7d00-060b-11e9-9e9d-a4a825490246.png)

PR #4288, opened 2017-08-30
Fix: Prevents linking binaries from workspaces

arcanis:
**Summary**

When running `yarn install` on some kind of workspaces (I still have to find out the exact criteria that trigger this bug), Yarn crashes with the following error message:

```
ENOENT: No such file or directory, lstat '/some/path/.../node_modules/workspace-aggregator-...'
```

This fix is an hotfix to prevent this error from happening. I still need to investigate to understand what exactly happens, and under what circumstances. This is probably related to #3896.

**Test plan**

Tests will be wrote. Do not merge.

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -134,7 +134,9 @@ export default class PackageLinker {
 
     // write the executables
     for (const {dep, loc} of deps) {
-      await this.linkSelfDependencies(dep, loc, dir);
+      if (dep._reference && dep._reference.location) {
+        await this.linkSelfDependencies(dep, loc, dir);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/package-linker.js
```diff
@@ -398,7 +398,7 @@ export default class PackageLinker {
       await promise.queue(
         topLevelDependencies,
         async ([dest, pkg]) => {
-          if (pkg.bin && Object.keys(pkg.bin).length) {
+          if (pkg._reference && pkg._reference.location && pkg.bin && Object.keys(pkg.bin).length) {
```
BYK: I did something very similar here: https://github.com/yarnpkg/yarn/pull/4129/files#diff-306a2caea89d886b11134fb770d1ebf6R92

May be we need a wider check for this?

arcanis:
Managed to reproduced the issue - it happens when there's both a `workspaces` and a `bin` field in the package.json.

BYK:
Ah, so we are trying to install the bins for the workspace itself, which should not be there at all?

arcanis:
Yeah, we setup the `bin` for the workspace aggregator. For some reason I can reproduce it manually on a small testcase, but it works fine when run from inside Jest :/

BYK:
So the root cause is likely to be the same for https://github.com/yarnpkg/yarn/issues/4041

Issue #992, opened 2016-10-13
Windows fails with EPERM: operation not permitted, all files are read-only


oscar-b:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
When performing any operation with yarn (add for instance) it errors with:

```
EPERM: operation not permitted, open 'C:\foo\node_modules\firebase\app-node.js'
at Error (native)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Use Windows :(
2. Use yarn
3. Failure

To fix the issue, right click the node_modules folder and un-check "Read-only". Wait. Re-run yarn command. Now it works, like magic! After yarn is done, all files are back to being read-only though. 

**What is the expected behavior?**
Shouldn't have to mark the files as readable.

**Please mention your node.js, yarn and operating system version.**
node v6.3.1
yarn v0.15.1
Windows 10 (anniversary edition)

Daniel15:
> To fix the issue, right click the node_modules folder and un-check "Read-only". 

Is Yarn marking `node_modules` as readonly, or is something else doing that? I can't repro on Windows with a brand new app (ie. `yarn init && yarn add react`)

oscar-b:
Hi @Daniel15 

Could you please try this in a clean directory?

```
yarn init
yarn add firebase
yarn add react
```

The last command gives me:

```
C:\Dev\yarntest>yarn add react
yarn add v0.15.1
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
error EPERM: operation not permitted, open 'C:\Dev\yarntest\node_modules\firebase\app-node.js'
    at Error (native)
info Visit http://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

I'm not sure if it's something the firebase packages does?

oscar-b:
For clarification; I tried the same steps but with react and eslint and didn't see the same error.

Daniel15:
Thanks for the information! I can replicate the issue on Windows  but not on Debian Linux.

FredyC:
I would say this is more of a _critical error_ as it prevents even basic use of Yarn in any project using `firebase` module.

Removing _read-only_ flag recursively works, but only temporarily. Even running just `yarn` will return those flags as they were most probably during `linking dependencies` phase. I assume that removing that flag yields different hash.

[...]

PR #4301, opened 2017-09-04
Fix: make sure all extracted tarballs are r/w enabled

BYK:
**Summary**

Fixes: #992. This issue was supposed to be fixed by #2826 but it
was not setting directory permissions wide enough. This patch uses
the `readable` and `writable` options provided by the `tar-fs`
package which essentially sets everything to `0o777`.

**Test plan**

N/A


BYK:
Relevant code: https://github.com/mafintosh/tar-fs/blob/master/index.js#L78-L85

arcanis:
Why not using the regular dir `0o755` / file `0o644`? Using `777` looks quite dangerous.

BYK:
>Why not using the regular dir 0o755 / file 0o644? Using 777 looks quite dangerous.

That's fair. Honestly, I just wanted to use the named options instead of Unix file permission codes, that's why. If you think this is unsafe, I can change it. Actually, I can imagine a scenario where a private package should not be readable by other users so I'll go ahead and use your suggestion instead.

PR #4304, opened 2017-09-04
Fixes yarn why when using workspaces

arcanis:
**Summary**

We were not passing the `workspaceLayout` to the resolver, which was breaking the resolution process for workspaces (the resolver was trying to resolve the workspaces on NPM). This PR fix this.

A followup will be to remove (or rather reformat) the aggregator from the display. Given the current time constraints I haven't took time to completely remove it:

screen shot 2017-09-04 at 6 15 05 pm

**Test plan**

Existing tests should pass + added a new one + manually tested `yarn why` inside a workspace

### DIFF
--- a/__tests__/commands/why.js
+++ b/__tests__/commands/why.js
@@ -59,6 +59,10 @@ test.concurrent('throws error with too many arguments', (): Promise<void> => {
   });
 });
 
+test.concurrent("doesn't throw when using it inside a workspace", (): Promise<void> => {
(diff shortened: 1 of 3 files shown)


__tests__/fixtures/why/workspace/.yarnrc
```diff
@@ -0,0 +1 @@
+workspaces-experimental true
```
BYK: Not necessary.

PR #4316, opened 2017-09-05
Chore: fix info tests

BYK:
**Summary**

Looks like bumping the version on npm invalidated a cache that
revealed the new API response from npm for yarn. Now the
`scripts` field always exists even if it is empty.

**Test plan**

CI should pass.

### DIFF
--- a/__tests__/commands/info.js
+++ b/__tests__/commands/info.js
@@ -43,10 +43,11 @@ const expectedKeys = [
   'license',
   'dist',
   'directories',
+  'scripts',
 ];
 
 // yarn now ships as built, single JS files so it has no dependencies and no scripts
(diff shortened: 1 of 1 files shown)


Daniel15:
Clearly I need to improve the BuildSize output to better handle when the file size hasn't changed, and also figure out what happened with the `.rpm` file here (seems like BuildSize thinks there's two files when it's really just one). Sorry for the noise, I might turn off commenting for now and only keep the status thing. 😛 

https://github.com/Daniel15/BuildSize/issues/8

Daniel15:
Okay I updated BuildSize to not post a comment for trivial changes in file size (less than 200 bytes) which should prevent noisy comments like this one. I also fixed the issue with the .rpm file appearing twice, which is also what caused the overall number to be incorrect here. Let me know if you see any weirdness with it.

PR #4318, opened 2017-09-05
Add inline comments to improve readability of PR #4278

kaylieEB:
**Summary**

Addresses some concerns around readability from PR https://github.com/yarnpkg/yarn/pull/4278
Also moved the flagsToOrigin override inside the conditional so that the check is only performed when running from workspace root.

**Test plan**


BYK:
I'll leave the final approval and merging to @arcanis since he was the one who reviewed the original PR.

PR #4325, opened 2017-09-06
Fix: More resillient global and cache folder determination

### LINKED ISSUE #4320: Failing build on Heroku with 1.0.0
Getting an error on Heroku with the build failing when trying to build the `"bcrypt": "^1.0.2",` package:

```
EROFS: read-only file system, access '/usr/local/bin'".
```

Reverting to yarn 0.27.5 fixes the issue.

### LINKED ISSUE #4323: Chore: consolidate writeable directory finding logic
Brought up at https://github.com/yarnpkg/yarn/issues/4320#issuecomment-327417225. We currently have two (or more?) different code paths for finding a writeable directory: one in cache, one in finding the global prefix.

Ideally, we'd have a single method that takes a list of directories and returns the first one that is writable.

BYK:
**Summary**

Fixes #4320 and fixes #4323. We were using `fs.access` when selecting
the global prefix folder automatically and expecting only an `EACCES` error.
This caused issues on Heroku where one of our first tries had threw an
`EROFS` error.

This patch also adds `$DESTIR` support back, which was removed in #3721,
probably inadvertently.

**Test plan**

Existing cache folder fallback tests should be enough for now. We should
move the core of those tests for the newly added `fs.getFirstWriteableFolder`
method.

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -103,7 +103,7 @@ async function getGlobalPrefix(config: Config, flags: Object): Promise<string> {
 
   if (!prefix) {
     throw new MessageError(
-      config.reporter.lang('noGlobalFolder', prefixFolderQueryResult.skipped.map(item => item.folder)),
+      config.reporter.lang('noGlobalFolder', prefixFolderQueryResult.skipped.map(item => item.folder).join(', ')),
     );
   }
 
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -102,9 +102,14 @@ async function getGlobalPrefix(config: Config, flags: Object, mode): Promise<str
   const prefix = prefixFolderQueryResult.folder && path.dirname(prefixFolderQueryResult.folder);
 
   if (!prefix) {
-    throw new MessageError(
-      config.reporter.lang('noGlobalFolder', prefixFolderQueryResult.skipped.map(item => item.folder).join(', ')),
+    config.reporter.warn(
+      config.reporter.lang(
+        'noGlobalFolder',
+        prefixFolderQueryResult.skipped.map(item => path.dirname(item.folder)).join(', '),
+      ),
     );
+
+    return FALLBACK_GLOBAL_PREFIX;
   }
 
   return prefix;
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -87,29 +87,26 @@ async function getGlobalPrefix(config: Config, flags: Object): Promise<string> {
     return process.env.PREFIX;
   }
 
-  let prefix = FALLBACK_GLOBAL_PREFIX;
+  const potentialPrefixFolders = [FALLBACK_GLOBAL_PREFIX];
   if (process.platform === 'win32') {
     // %LOCALAPPDATA%\Yarn --> C:\Users\Alice\AppData\Local\Yarn
     if (process.env.LOCALAPPDATA) {
-      prefix = path.join(process.env.LOCALAPPDATA, 'Yarn');
+      potentialPrefixFolders.unshift(path.join(process.env.LOCALAPPDATA, 'Yarn'));
     }
   } else {
-    prefix = POSIX_GLOBAL_PREFIX;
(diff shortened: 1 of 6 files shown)


src/constants.js
```diff
@@ -82,7 +82,7 @@ function getYarnBinPath(): string {
 export const NODE_MODULES_FOLDER = 'node_modules';
 export const NODE_PACKAGE_JSON = 'package.json';
 
-export const POSIX_GLOBAL_PREFIX = '/usr/local';
+export const POSIX_GLOBAL_PREFIX = `${process.env.DESTIR || ''}/usr/local`;
```
arcanis: typo

src/util/fs.js
```diff
@@ -880,3 +890,30 @@ export async function readFirstAvailableStream(
 
   return {stream, triedPaths};
 }
+
+export async function getFirstWriteableFolder(paths: Iterable<string>): Promise<FolderQueryResult> {
+  const result: FolderQueryResult = {
+    skipped: [],
+    folder: null,
+  };
```
arcanis: Why don't we just return a raw string? The list of skipped directories can be deduced relatively easily by the caller anyway.

src/config.js
```diff
@@ -290,29 +290,17 @@ export default class Config {
...
-          await fs.writeFile(testFile, 'content');
-          await fs.readFile(testFile);
-
-          cacheRootFolder = tentativeCacheFolder;
-        } catch (error) {
-          this.reporter.warn(this.reporter.lang('cacheFolderSkipped', tentativeCacheFolder));
-        }
+      const cacheFolderQuery = await fs.getFirstWriteableFolder(preferredCacheFolders);
+      for (const failedFolder of cacheFolderQuery.skipped) {
+        this.reporter.warn(this.reporter.lang('cacheFolderSkipped', failedFolder));
```
arcanis: `failedFolder` is not a string, so the display will probably be wrong. I think you meant `failedFolder.folder`? The variable name would probably be better as `skippedEntry` btw, so that it would be `skippedEntry.folder`

src/constants.js
arcanis: (`DESTIR` instead of `DESTDIR`)

BYK: darn

src/util/fs.js
BYK: Because I also attach the error, in case that matters. May be it is EACCESS, may be it is EROFS or may be it is something else?

Also deducing the list would require another traversal of the entire list until the selected folder is found which is a bit unnecesary.

src/config.js
BYK: Yup to both, good catch!

src/cli/commands/global.js
```diff
@@ -87,29 +87,31 @@ async function getGlobalPrefix(config: Config, flags: Object): Promise<string> {
...
-  } catch (err) {
-    if (err.code === 'EACCES') {
-      prefix = FALLBACK_GLOBAL_PREFIX;
-    } else if (err.code === 'ENOENT') {
-      // ignore - that just means we don't have the folder, yet
-    } else {
-      throw err;
-    }
+  const binFolders = potentialPrefixFolders.map(prefix => path.join(prefix, 'bin'));
+  const prefixFolderQueryResult = await fs.getFirstSuitableFolder(binFolders);
```
BYK: Here's the problem with this code:
We only need write permissions when doing `yarn global add` which is usually called with `sudo` or similar. That said if we change our permission checks depending on usage, then we may end up using two different global folders based on how `yarn` is called (with sudo and without) which would make running `yarn global add` useless.

I think we should just check `R_OK` and `X_OK` here but doing that caused some failing tests (which can be fixed, I'm sure). This would also be a breaking change. That said I'm almost certain that `POSIX_GLOBAL_PREFIX/bin` won't be writeable to yarn by default. What do you think?

thomaswrenn:
Can v1.0.1 be released/published to npm? 

We're having the heroku issue on our build systems.

BYK:
@thomaswrenn should be on npm right now?

PR #4341, opened 2017-09-07
Don't create bin links within link:// deps or from sub-packages of linked deps

skevy:
**Summary**

When installing a package with link:// deps, Yarn will link bin files from the parent package into the linked package and from any sub-packages of the linked package. This is undesirable behavior.

For link:// deps, Yarn doesn't do an install in the linked dep, so IMO it shouldn't link bin's either.

It's particularly weird when you have a lot of linked deps throughout your repo, interlinked with each other. You end up with .bin directories in your packages that have links to places all over your monorepo:

```
- package-foo/node_modules/.bin/
  - eslint ---> ../eslint/bin/eslint.js
  - mkdirp ---> ../../../package-bar/node_modules/mkdirp/bin/cmd.js

- package-bar/node_modules/.bin
  - babel-cli ---> ../../package-foo/node_modules/babel-cli/bin/babel.js
```

Since `link://` deps are treated like remotes, I think not modifying the linked modules node_modules dir is the correct behavior.

In addition, when I would run simultaneous `yarn install`'s in packages that were linked together with `link://`, I'd occasionally encounter ENOENT errors (during the link bin step) because it would be trying to link bin files in directories that didn't exist yet.

**Test plan**

Ran tests. Tested in a repo that uses lots of `link://` deps.


src/package-linker.js
```diff
@@ -468,6 +479,16 @@ export default class PackageLinker {
     return range === '*' || satisfiesWithPreleases(version, range, this.config.looseSemver);
   }
 
+  _isChildOfLinkedDep(linkedDepLocations: Array<string>, location: string): boolean {
+    let isWithinLink = false;
+    linkedDepLocations.forEach(loc => {
+      if (location.startsWith(loc)) {
```
skevy: Also, I will admit that this isn't my favorite way to solve this problem, but it is the least intrusive.

There's no real concept, that I could tell, of "give me the parent PackageReference that asked for this dependency". There is a `requests` array, but in the scenario where package-a (a non-linked dep) _and_ package-b (a link:// dep) requests `foo-bar`, it's not clear to me how, within the context of looping through the `flatTree` data structure, how we would tell which one of those "requests" applies.

Seeing as the flat tree structure provides us with an absolute path (that is not resolved into a symlink), this seemed like a reasonable (albeit slightly hacky) way to determine whether or not it was a child.

Curious to hear thoughts here or if I missed something in the underlying data structures.

skevy:
I've updated this PR, after realizing I did was checking the wrong thing.

I didn't want to check whether or not the package being linked was within a linked dep -- I wanted to check whether the destination that it was being linked to was within a linked dep.

skevy:
I made an amendment to this PR -- in addition to not linking deps elsewhere that are sub deps of a linked package, we _also_ don't want to link deps from the hoisted node_modules into the linked package. There should be no modification of `link:`'ed packages by yarn'ing in a dependent package.

src/package-linker.js
```diff
@@ -483,6 +500,16 @@ export default class PackageLinker {
     return range === '*' || satisfiesWithPreleases(version, range, this.config.looseSemver);
   }
 
+  _isChildOfLinkedDep(linkedDepLocations: Array<string>, location: string): boolean {
+    let isWithinLink = false;
+    linkedDepLocations.forEach(loc => {
+      if (path.join(location, 'node_modules').startsWith(loc)) {
```
BYK: We actually have a constant for this in constants.js: `NODE_MODULES_FOLDER` can you use that instead?

BYK: Instead of the "heavy" `path.join` can we not just check for `location.startsWith(loc)`?. Also I think using a short-circuited for-of loop would be better here.

src/package-linker.js
```diff
@@ -403,6 +413,13 @@ export default class PackageLinker {
         topLevelDependencies,
         async ([dest, pkg]) => {
           if (pkg._reference && pkg._reference.location && pkg.bin && Object.keys(pkg.bin).length) {
+            if (
+              this._isChildOfLinkedDep(linkTypeDepsLocations, pkg._reference.location) ||
+              this._isChildOfLinkedDep(linkTypeDepsLocations, dest)
```
BYK: Looks like we only use `_isChildOfLinkedDep` in a combination of `dest` and `pkg._reference.location` so may be we can pass it two values instead and be more efficient?

src/package-linker.js
```diff
@@ -179,6 +180,8 @@ export default class PackageLinker {
         // with a symlink source
         src = remote.reference;
         type = 'symlink';
+        // store the dest location for later usage
+        linkTypeDepsLocations.push(dest);
```
BYK: Maybe a `Set()` would be more efficient here?

BYK:
@skevy - Still interested in working on this?

skevy:
@BYK yah definitely. We're using a version of this patch internally on our fork. Just have been busy with some other things. I'll get back to upstreaming some of my changes soon...I also have fixes for file: deps and `yarn pack`.

arcanis:
Ping @skevy - I ran a rebase to resolve the conflict, but can you check @BYK comments before I can merge the PR?

BYK:
Closing this PR since I don't think the approach is good. We should not operate on the final destination of links but have higher-level checks to avoid putting stuff under `link://` dependencies.

That said, looking at this and #5112 I feel like we should deprecate `link://` in favor of workspaces. @rally25rs what do you think?

rally25rs:
IMO `link:` causes more problems than it fixes. The place I had hopes for using it was for local dev between packages that depend on each other, but I had to stop using it locally because if I had something like:

```
proj1
  node_modules

proj2
  node_modules
```

and added `link:../proj1` to proj2, it would mess up dependency resolution because node.js would find the packages in `proj1/node_modules` which may or may not be in sync with the versions specified in `proj2`'s `yarn.lock`, and also caused multiple versions of packages for web projects where they should have been flattened and hoisted to a common version.

I think `yarn link` or `file:` or `workspace` are all viable solutions instead of `link:`

PR #4347, opened 2017-09-08
Fix: support protocol-relative registry URLs

BYK:
**Summary**

Fixes matching protocol-relative registry URLs from config.
Reported here: https://github.com/yarnpkg/yarn/pull/3987#discussion_r137624001

**Test plan**

Added one new test.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -19,7 +19,7 @@ import url from 'url';
 import ini from 'ini';
 
 const DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
-const REGEX_REGISTRY_PREFIX = /^(https?:)?\/\//;
+const REGEX_REGISTRY_PREFIX = /^(https?:)?\/\//i;
 const REGEX_REGISTRY_SUFFIX = /registry\/?$/;
 
 export const SCOPE_SEPARATOR = '%2f';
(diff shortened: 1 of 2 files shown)


src/registries/npm-registry.js
```diff
@@ -314,8 +314,8 @@ export default class NpmRegistry extends Registry {
     // 3nd attempt, remove the 'registry/?' suffix of the registry URL
     return (
       this.getScopedOption(reg, option) ||
-      (reg.match(pre) && this.getRegistryOption(reg.replace(pre, ''), option)) ||
-      (reg.match(suf) && this.getRegistryOption(reg.replace(suf, ''), option))
+      (pre.test(reg) && !reg.startsWith('//') && this.getRegistryOption(reg.replace(pre, '//'), option)) ||
```
arcanis: Why this `reg.startsWith('//')` check? If it starts with `//` the subsequent replace will simply be a noop. This change makes `getRegistryOption` not being applied to protocol-relative url, is it intentional?

BYK: This was due to the regex now matching the whole `https://` part instead of just the protocol. We want the `//` kept intact but then we need the additional `.startsWith()` check to prevent this method from recursing infinitely on the same registry value.

That said I did this to share the same regex with others. I tried adding a comment but it took three lines so instead I created another regex 😆

PR #4351, opened 2017-09-08
Fix: remove the `accept` header from tar fetcher requests

### LINKED ISSUE #3833: `406 Not Acceptable` when npm succeeds
bug

`yarn add http://prerelease.componentone.com/wijmo5/npm-images/C1Wijmo-Enterprise-Eval-System-5.20172.328.tgz`
results in
```
yarn add v0.27.5
info No lockfile found.
[1/4] Resolving packages...
error An unexpected error occurred: "http://prerelease.componentone.com/wijmo5/npm-images/C1Wijmo-Enterprise-Eval-System-5.20172.328.tgz: Request failed \"406 Not Acceptable\"".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\User\\Source\\yarn-test\\yarn-error.log".
```

`yarn add http://prerelease.componentone.com/wijmo5/npm-images/C1Wijmo-Enterprise-Eval-System-5.20172.328.tgz`

install wijmo package like

`npm install http://prerelease.componentone.com/wijmo5/npm-images/C1Wijmo-Enterprise-Eval-System-5.20172.328.tgz`
```
npm notice created a lockfile as package-lock.json. You should commit this file.
+ wijmo@5.20172.328
added 1 package in 0.883s
```

Yarn version: 
  0.27.5

Node version: 
  8.1.3

Platform: 
  win32 x64

BYK:
**Summary**

Fixes #3833. Removes `accept` header from tar requests as @arcanis suggested since `application/octet` wasn't ensuring any valid tar file and `npm` client does not send this header anyway.

**Test plan**

Run `yarn add http://prerelease.componentone.com/wijmo5/npm-images/C1Wijmo-Enterprise-Eval-System-5.20172.328.tgz`

It fails without the patch, installs correctly with the patch.

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -142,7 +142,7 @@ export default class TarballFetcher extends BaseFetcher {
           {
             headers: {
               'Accept-Encoding': 'gzip',
-              Accept: 'application/octet-stream',
+              Accept: 'application/octet-stream, application/x-tar',
             },
             buffer: true,
             process: (req, resolve, reject) => {
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -142,7 +142,6 @@ export default class TarballFetcher extends BaseFetcher {
           {
             headers: {
               'Accept-Encoding': 'gzip',
-              Accept: 'application/octet-stream, application/x-tar',
             },
             buffer: true,
             process: (req, resolve, reject) => {


BYK:
We already do though. We just support more formats now.

arcanis:
Sure, but why not just remove the header? If we send it we're always gonna have potentially broken packages, especially since this is something that we have no control over with (the remote servers).

BYK:
Removing it would be a potentially breaking change since it changes a previously established behavior and we don't know why it was put there in the first place. I'm assuming it was for an edge case. Adding this extra type however, is just extending an already existing behavior, hence a safer change.

BYK:
Context for public: @arcanis and I had a chat and agreed that having the `Accept` header was technically the correct thing but if people have custom proxies or npm servers out there, relying on `npm` client's behavior of not sending an accept header at all, this may cause breakages like this so it is more user-friendly to remove the header completely which doesn't have much value to begin with.

Daniel15:
Can we merge this into the next 1.0.x? `application/octet-stream` seems wrong as any server that sends a proper content-type will fail the request.

BYK:
@Daniel15 that's the plan :)

PR #4367, opened 2017-09-08
Fix the scope regex, allow / in addition to %2f in the url

### LINKED ISSUE #4366: Can't install from private scoped registries
This is a regression in 1.0.0.

**Explanation**
This regex `const SCOPED_PKG_REGEXP = /(?:^|\/)(@[^\/?]+?)(?=%2f)/`:
https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L30

introduced in this commit:
https://github.com/yarnpkg/yarn/commit/cbb27f4fb0bb5860435ac0ae5f0dd3f0062825eb

or this PR:
https://github.com/yarnpkg/yarn/pull/4027

completely breaks support for private npm registries.

The `request` function is sometimes called with a package name (`@my/pkg`) and sometimes with a full URL (`https://myregistry.dev/@my/pkg/download/@my/pkg-1.0.0.tgz`).

The old regex `/^@|\/@/` matched such URLs, but the new one`/(?:^|\/)(@[^\/?]+?)(?=%2f)/` doesn't. This means that authorization token is not sent with any `tgz` requests.

cc due to involvement in the original PR @lukeggchapman @BYK

KidkArolis:
Fixes #4366

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -251,7 +251,7 @@ describe('isRequestToRegistry functional test', () => {
 const packageIdents = [
   ['normal', ''],
   ['@scopedNoPkg', ''],
-  ['@scoped/notescaped', ''],
+  ['@scoped/notescaped', '@scoped'],
   ['not@scope/pkg', ''],
   ['@scope?query=true', ''],
   ['@scope%2fpkg', '@scope'],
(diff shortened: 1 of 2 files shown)


KidkArolis:
Will add a test in a bit.

KidkArolis:
Now with tests

BYK:
See https://github.com/yarnpkg/yarn/issues/4366#issuecomment-328349888

I don't think this is the correct solution.

lukeggchapman:
Following on from https://github.com/yarnpkg/yarn/issues/4366#issuecomment-328462851

I think this might be the best solution.
The request method having to handle both:
`@scope%2fpkg` and `https://npm.registry.com/@scope/pkg/-/pkg-0.0.1.tgz`
makes it difficult, without handling of package names and path requests more separately.

I've [added some more tests and a fix for `getRegistry`](https://github.com/lukeggchapman/yarn/commit/5883016353568a889057972c23a8180979b1ec3c) so that when request is called with a path it uses the registry to get auth rather than the scope, it still relies on `isScopedPackage` (or always-auth) to flag it as needing auth.
@KidkArolis if you merge/rebase this to the latest master I can add them to your PR, depending on what @BYK wants to do.

lukeggchapman:
Hey @KidkArolis I improved some tests while investigating but couldn't PR against your fork.
Have a look and if you think they're worth while, either cherry-pick or I can add them.
https://github.com/yarnpkg/yarn/commit/7a01b96489f098b361efc458432bc6df753b29b2

KidkArolis:
@BYK this PR fixes the private registry issue. Is it possible to get this into the next patch release?

Re https://github.com/yarnpkg/yarn/compare/master...lukeggchapman:more-scope-tests, I'm happy to:

1. Cherry pick that into this PR, if we think more tests are better.
2. First get this PR merged, and then have a separate PR improving the tests.

Whatever gets this fixed sooner.. as it's causing a bit of disturbance, all our Codeship builds are failing since Codeship upgraded to the latest yarn..

lukeggchapman:
Agreed, lets get this in as soon as possible. I can create a seperate PR after.

BYK:
@KidkArolis it is my aim to get this into 1.0.2 and release that ASAP.

PR #4374, opened 2017-09-09
chore(build): Replace preset-node-5 and babel-preset-es2015-node4 w/ babel-preset-env

kaylieEB:
**Summary**
Replace [babel-preset-node 5](https://github.com/leebenson/babel-preset-node5) and [babel-preset-es2015-node4](https://github.com/jbach/babel-preset-es2015-node4) with babel-preset-env since the first two packages are deprecated and no longer maintained.

The downside is that `preset-env` is has the comprehensive list of plugins for each environment type, whereas previously packages only selected a few that we were using, so our build size increased. I'm curious to see why @yarnpkg/core 's thoughts are on this, and happy to close the PR if we'd rather stick with the old packages.

**Test plan**
Builds should build and tests should pass since there weren't any code changes 😃


Daniel15:
@kaylieEB Hmm, I wonder why this is going to increase the .js build size by ~100 KB? Any ideas?

kaylieEB:
@Daniel15 I'd assume so since js files still go through a [transpilation/build step via gulp](https://github.com/yarnpkg/yarn/blob/master/gulpfile.js#L36)?

BYK:
@kaylieEB still WIP? What do you need to get unblocked?

kaylieEB:
I guess node 6 part is actually good to go, depends if we want the update node 4 in one same pr. I'll get to it later this week, or otherwise feel free to take over the PR.

BYK:
>I guess node 6 part is actually good to go, depends if we want the update node 4 in one same pr. I'll get to it later this week, or otherwise feel free to take over the PR.

Isn't it better to upgrade them together at once so there are no discrepancies or code/dependency duplication? I think we can wait a bit more until you have time to get to this :)

BYK:
@kaylieEB ping! 😄

kaylieEB:
@BYK ready for review!

__tests__/index.js
```diff
@@ -100,7 +100,7 @@ function expectAnErrorMessage(command: Promise<Array<?string>>, expectedMessage:
     .then(function() {
       throw new Error('the command did not fail');
     })
-    .catch(error => expect(error.message).toContain(expectedMessage));
```
arcanis: What is this checking for?

arcanis:
Will close/reopen to trigger a new CircleCI test, it seems one of them bugged :(

PR #4377, opened 2017-09-09
feat(config): Add custom config for init private flag

pierreneter:
**Summary**

Here is a small custom; I add this because it relates to https://yarnpkg.com/en/docs/cli/init#toc-setting-defaults-for-yarn-init. I discovered this was necessary while writing introduces for this flag.

**Test plan**

New init and config tests.


src/cli/commands/init.js
```diff
@@ -31,6 +31,8 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     email: config.getOption('init-author-email'),
     url: config.getOption('init-author-url'),
   };
+  const _defaultPrivateFlag = String(config.getOption('init-private'));
+  const defaultPrivateFlag = _defaultPrivateFlag === 'undefined' ? '' : _defaultPrivateFlag;
```
BYK: Why not simply
```
const defaultPrivateFlag = config.getOption('init-private') || '';
```

This is actually safer since it allows the literal `'undefined'` value too. (not that I expect it to be used but it is "more accurate")

BYK:
I do not understand this change fully, can you explain your aim a bit more? Are you trying to make private-flag the default? Do we have to cast this to a string? Can we just do `config.getOption('private-flag') === true` instead?

src/cli/commands/init.js
pierreneter: Hi. I tried this way first. But this not pass the lint test. :(

pierreneter: You can see here: https://circleci.com/gh/yarnpkg/yarn/tree/pull%2F4377
Ok, I just thought of a shorter way. Let me try.

pierreneter:
@BYK about this, `config.getOption` return a string value. Wait a second. Let me push a commit.

pierreneter:
In this PR, I would like to add the ability for users to use `yarn config set init-private <value>`, same as another argument in `init`.

src/cli/commands/init.js
```diff
@@ -94,7 +94,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     {
       key: 'private',
       question: 'private',
-      default: '',
+      default: String(config.getOption('init-private') || ''),
```
pierreneter: This's a `string` value, and (L155)[https://github.com/yarnpkg/yarn/pull/4377/files#diff-527026e7decfe80ca4f832022aeb0d86R155] will pass `string` value of `private` flag to `boolean`

BYK:
Hey, sorry for the very late turnaround :( I tried adding `--init.private true` to my `.yarnrc` file and it seems to do the trick. Is that not enough?

pierreneter:
Hey, I tried your trick with my `.yarnrc` but it doesn't work. (`Windows`, `1.0.1`). Hm. If it works ok, so, it will be strange to what the user is used to https://yarnpkg.com/en/docs/cli/init#toc-setting-defaults-for-yarn-init

BYK:
It did work for me. What do you see when you run `yarn init` at the "private" stage?

pierreneter:
@BYK here is my screen and `.yarnrc`
yarn init

.yarnrc

BYK:
>@BYK here are my screen and .yarnrc

I've tested this locally and it works but apparently, the lockfile parser is allergic to Windows-style line ending so it may be that if you edited the file manually. Make sure you have Unix-style line endings and it should work.

BYK:
Added a PR for `\r\n` here: #4495

pierreneter:
But, what about enable setting default config like another init option :'( ?

src/cli/commands/init.js
BYK: Yarn automatically detects "magic" boolean strings `true` and `false` and converts the value into booleans. Try and see yourself ;)

BYK:
>But, what about enable setting default config like another init option :'( ?

Good point. Added a comment. LMK if that's not the case. Also shouldn't we have at least one test case for this?

BYK:
@pierreneter let's finish this with a test case! 😉

src/cli/commands/init.js
```diff
@@ -96,7 +96,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     {
       key: 'private',
       question: 'private',
-      default: '',
+      default: config.getOption('init-private') || '',
```
kaylieEB: any reason this shouldn't default to `false`?

BYK: No idea :D

pierreneter: Hm, because, default this's not set yet. It will not add to `package.json`. If set it's `false`, always it will be add to `package.json`: `private: false`.

PR #4381, opened 2017-09-10
Free tests from FORCE_COLOR environment variable dependency

hron:
FORCE_COLOR environment variable allows to force yarn to use
colors. This makes some of tests fail because of comparing
colorized output with plain strings.

For example, if you run a test in environment where
FORCE_COLOR is set you get this failure:

```
 FAIL  __tests__\index.js
  ● should add package
    expect(received).toEqual(expected)
    Expected value to equal:
      "success Saved lockfile."
    Received:
success Saved lockfile."
```

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -30,14 +30,15 @@ async function execCommand(
   }
 
   return new Promise((resolve, reject) => {
+    const cleanedEnv = {...process.env};
+    cleanedEnv['YARN_SILENT'] = 0;
+    delete cleanedEnv['FORCE_COLOR'];
+
     exec(
       `node "${yarnBin}" ${cmd} ${args.join(' ')}`,
       {
         cwd: workingDir,
-        env: {
-          ...process.env,
-          YARN_SILENT: 0,
-        },
+        env: cleanedEnv,
       },
       (error, stdout) => {
         if (error) {
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -59,6 +59,11 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
 const getRandomPort = () => Math.floor(Math.random() * PORT_RANGE) + MIN_PORT_NUM;
 
 async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
+  if (!options['env']) {
+    options['env'] = {...process.env};
+    options['extendEnv'] = false;
+  }
+  delete options['env']['FORCE_COLOR'];
   const {stdout, stderr} = await execa(path.resolve(__dirname, '../bin/yarn'), args, options);
 
   return [stdout, stderr];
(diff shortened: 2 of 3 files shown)


__tests__/index.js
```diff
@@ -30,14 +30,15 @@ async function execCommand(
   }
 
   return new Promise((resolve, reject) => {
+    const {...cleanedEnv} = {...process.env};
```
BYK: Isn't `const cleanedEnv = {...process.env};` enough?

__tests__/index.js
```diff
@@ -30,14 +30,15 @@ async function execCommand(
...
+    cleanedEnv['YARN_SILENT'] = 0;
+    delete cleanedEnv['FORCE_COLOR'];
```
BYK: Isn't it safe to force this value to be `0` as well?

__tests__/integration.js
```diff
@@ -59,6 +59,11 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
 const getRandomPort = () => Math.floor(Math.random() * PORT_RANGE) + MIN_PORT_NUM;
 
 async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
+  if (!options['env']) {
+    options['env'] = Object.assign({}, process.env);
```
BYK: Why `Object.assign()` here instead of the spread notation like other places?

__tests__/integration.js
```diff
@@ -59,6 +59,11 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
...
+    options['extendEnv'] = false;
```
BYK: What is this? I've searched the code and couldn't find any references to `extendEnv` :(

__tests__/index.js
hron: Unfortunately, setting FORCE_COLOR to 0 is not enough. All of the packages just checks if this variable is defined. Here is an example from `supports-color` library (which is a dependency of yarn):

```javascript
if ('FORCE_COLOR' in process.env) {
	return true;
}
```

__tests__/integration.js
hron: Agreed. This issue has been fixed.

__tests__/integration.js
hron: `extendEnv` is an option of `execa` function. Without passing this option set to false `execa` executes the following piece of code:

```javascript
if (opts && opts.env && opts.extendEnv !== false) {
	opts.env = Object.assign({}, process.env, opts.env);
}
```

This merges `process.env` and `opts.env` and FORCE_COLOR is set again.

__tests__/index.js
hron: You're right! Thank you very much for the hint.

__tests__/integration.js
```diff
@@ -59,6 +59,11 @@ const PORT_RANGE = MAX_PORT_NUM - MIN_PORT_NUM;
 const getRandomPort = () => Math.floor(Math.random() * PORT_RANGE) + MIN_PORT_NUM;
 
 async function runYarn(args: Array<string> = [], options: Object = {}): Promise<Array<Buffer>> {
+  if (!options['env']) {
+    options['env'] = {...process.env};
+    options['extendEnv'] = false;
```
BYK: Inline comment about this and `execa` would be great.

BYK:
Thanks a lot and sorry for the delay!

PR #4386, opened 2017-09-10
Chore: add instructions about using local builds

BYK:
**Summary**

I got questions about how to run the latest master locally and realized this is not document. This patch documents how to run local yarn builds for potential contributors. 

**Test plan**

Test the suggested change locally and verify it actually works.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,30 +10,36 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 1. Run the following commands:
 
 ```sh
-$ git clone YOUR_YARN_REPO_URL
-$ cd yarn
-$ yarn
-$ yarn run build
+git clone YOUR_YARN_REPO_URL
+cd yarn
+yarn
+yarn run build
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -10,30 +10,36 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
...
 
 ```sh
-$ yarn run watch
+yarn run watch
+```
+
+## Using the local builds
+
+```sh
+alias yarn="node /path/to/yarn/lib/cli/index.js"
```
Daniel15: This is an interesting approach... I usually just use the full path (eg. `c:\src\yarn\bin\yarn`)

I'd suggest using `bin/yarn` rather than `lib/cli/index.js`.

BYK:
>I wonder if we should move this document to the website... Isn't it strange having one piece of documentation in the repo, but everything else on the site?

Totally agree, I was actually surprised that this document was here. That said it is also a good practice to have some basic info inside th repo itself, right? May be we can cross-link them?

Issue #3347, opened 2017-05-08
On 'yarn install', yarn should report on bandwidth usage along with time taken


NiloCK:
**Do you want to request a *feature* or report a *bug*?**
Feature request. Not substantial (as far as I can guess).

**What is the current behavior?**
After running `yarn install`, yarn reports on the time taken by the installation - `Done in Xs`.

**What is the (desired) behavior?**
It should also report on the bandwidth usage.

I'm on a bandwidth budget, and I decided to adopt yarn after learning that it actually uses the local cache, without the user having to resort to odd hacks ([as in this npm issue](https://github.com/npm/npm/issues/2568)).

I suspect that there are others who switched for the same reason. This feedback is useful for us, and fits nicely with the current completion report.

bestander:
Send a PR, sounds cool

PR #4389, opened 2017-09-11
Log network usage on installs

olingern:
I took a stab at #3347, and thought it could be a useful feature for visibility in terms of what overhead does a package bring along.

**Summary**:
~~This initial implementation looks at the response header's `content-length` and aggregates them into the console reporter over the lifetime of the install.~~

This updated implementation hooks into the request manager's stream 'data' event and aggregates stream length over the lifetime of an add / install.

Looking around more closely, I found that `request-manager` is the funnel for all requests, so this is now updated to account for that.

Into the footer, it appends a new line and dynamically adjusts the size for easier human parsing :) 

screen shot 2017-09-10 at 11 25 42 pm

**Test Plan**:

- Verify tarball size against content length (currently a discrepancy)
- Verify output against multiple package installs
- Verify output when some packages are taken from cache and others resolved

_... ^ not an exhaustive list. I'm sure there are many different combinations of install sources I'll need to consider._

I believe these tarballs should be fixtures that I'm able to compare the logged network size to, i.e. local `left-pad` fixture should be same same size as resolved `left-pad` package of the same version. 

@BYK @bestander I'd love to get some feedback on the direction ( using the console reporter to maintain state ), the emoji ⬇️ , ~~and also some direction into the other places I should look into logging network usage ( like [here](https://github.com/yarnpkg/yarn/blob/b8f8b7c3b81544e616c97c9aefc2dd909f808a65/src/util/git.js#L234) )~~

Cheers

**EDIT**: This definitely breaks all a number of snapshots. Let me know if there's a good way to verify that this maintains stability ( only showing when a flag is present, like `showPeakMemory` functions in the reporter )


olingern:
A couple of updates after digging through a few things.

BYK:
@olingern thanks for the PR! We're a bit busy trying to release 1.0.2 soon. I'll get back to this PR once that's out :)

olingern:
ping @BYK I have some extra time over the next few days. Let me know if this is something you guys would be interested in merging :)

src/reporters/base-reporter.js
```diff
@@ -87,6 +88,10 @@ export default class BaseReporter {
   peakMemory: number;
   startTime: number;
 
+  logNetworkUsage(bytes: number) {
+    this.networkUsage += parseInt(bytes);
```
BYK: If it is already defined as `number`, you shouldn't need the extra `parseInt` call, right?

src/reporters/base-reporter.js
```diff
@@ -170,6 +175,27 @@ export default class BaseReporter {
     }
   }
 
+  getSizeUnit(iterations: number) {
```
BYK: Why not use something already available like https://yarnpkg.com/en/package/humanize-bytes instead of rolling your own?

src/reporters/base-reporter.js
```diff
@@ -170,6 +175,27 @@ export default class BaseReporter {
...
+    }
+  }
+  convertBytes(bytes: number, iterations: number): string {
+    return bytes < 100
+    ? `${bytes.toFixed(2)} ${this.getSizeUnit(iterations)}`
+    : this.convertBytes(bytes / 1000, ++iterations);
+  }
+
+  getTotalNetworkUsage(): number {
+    return this.networkUsage > 0 ? this.convertBytes(this.networkUsage, 0) : 0;
```
BYK: Do you really need the ternary? I think that module I referred about should be good enough :)

src/util/request-manager.js
```diff
@@ -424,18 +425,27 @@ export default class RequestManager {
...
   }
 
+  handleRequestData(data: object, queue: object) {
+
+    if (data && data.length > 0) {
+      this.reporter.logNetworkUsage(data.length);
+    }
+
+    if (queue) {
+      queue.stillActive.bind(queue);
```
BYK: What does this line do?

src/reporters/base-reporter.js
olingern: Oh, yes. Probably a good idea. Retrospect tells me I was trying to be too clever here.

olingern:
@BYK Exciting. I'll pick this back up this week, add tests / address comments in more detail. 

One question:
Should _always_ be shown on installs, or only with a flag? The former breaks many snapshots, the latter would not.

src/util/request-manager.js
olingern: From my memory, this checks to see if there are still requests resolving and keeps the spinners active in the console reporter.

[Current implementation](https://github.com/yarnpkg/yarn/blob/ae3cf9971d73350d6b52a0bfb3cc708e7a22d80b/src/util/request-manager.js#L441)

src/reporters/base-reporter.js
olingern: Maybe not :) I'll give [humanize-bytes](https://yarnpkg.com/en/package/humanize-bytes) a look and see what I can do with it

src/reporters/base-reporter.js
olingern: Was trying to be 'defensive' here, but maybe that's detrimental to performance. I'll go back and see if there are any cases where `parseInt` would be required.

olingern:
ping @BYK 
Going to go forward as this being the default behavior. Let me know if deserves a flag. Shaping this up over the holiday weekend. Excited to get it landed in the near future!

src/reporters/base-reporter.js
BYK: Just trust `flow` ;)

src/util/request-manager.js
BYK: It doesn't seem to be doing anything to me since `.bind()` simply returns a new function but doesn't call it. Since this is a "dangling" expression, I don't think the returned function is used anywhere and `.bind()` shouldn't have any side effects. Am I missing something here?

BYK:
@olingern I'd put this behind a flag or `yarnrc` option and default to off to keep things backward compatible. This may break people relying on the specific structure of the output of yarn. We can make this the default in a major version release if people demand it.

I'd be sure to promote this with a tweet after it lands and I think we should add some docs around it so people can find it. What do you think?

olingern:
> @olingern I'd put this behind a flag or yarnrc option and default to off to keep things backward compatible. This may break people relying on the specific structure of the output of yarn. We can make this the default in a major version release if people demand it.

This is what I thought, but wouldn't have objected to it being seen on every install :P

> I'd be sure to promote this with a tweet after it lands and I think we should add some docs around it so people can find it. What do you think?

For sure. I think making it optional for now, having good documentation around it, promoting it a la Twitter will all help gain traction.

If it's super popular / desired, perhaps introducing it as the default in a major release could work, since it's a potential breaking change.

src/util/request-manager.js
olingern: > It doesn't seem to be doing anything to me since .bind() simply returns a new function but doesn't call it. Since this is a "dangling" expression, I don't think the returned function is used anywhere and .bind() shouldn't have any side effects. Am I missing something here?

Nope, not missing anything. You're correct, it seems to be a 'dangling' expression, and wasn't doing anything. Thanks for catching that.

src/reporters/base-reporter.js
olingern: So, that lib you linked, humanize-bytes, only does "human -> bytes" and not the other way around, per their docs.

```javascript
var bytes = require('humanize-bytes');
bytes('1kb') // => 1024
bytes(1024) // => 1024
```

Poking around, not wanting to reinvent the wheel, I found [bytes](https://github.com/visionmedia/bytes.js), which seemed to have some ideal features that I was implementing, i.e. `decimalPlaces`

olingern:
@BYK Pushed the latest changes. Working on the flag / tests now

src/reporters/base-reporter.js
```diff
@@ -170,6 +176,14 @@ export default class BaseReporter {
     }
   }
 
+  convertBytes(bytes: number): string {
+    return humanizeBytes(bytes);
```
BYK: Is this indirection really needed?

src/reporters/console/console-reporter.js
```diff
@@ -157,6 +157,8 @@ export default class ConsoleReporter extends BaseReporter {
       const peakMemory = (this.peakMemory / 1024 / 1024).toFixed(2);
       msg += ` Peak memory usage ${peakMemory}MB.`;
     }
+
+    this.log(this._prependEmoji(`${this.getTotalNetworkUsage()} transferred`, '⬇️'));
```
BYK: Please move this message to `en.js` so it can be translated.

src/util/humanize-bytes.js
```diff
@@ -0,0 +1,5 @@
+import bytes from 'bytes';
+
+export default function humanizeBytes(bytesToConvert: number): string {
```
BYK: I'd put this into `src/util/misc.js`

src/util/request-manager.js
```diff
@@ -433,18 +434,22 @@ export default class RequestManager {
     const req = request(params);
     this.reporter.verbose(this.reporter.lang('verboseRequestStart', params.method, params.url));
 
-    req.on('error', onError);
-
     const queue = params.queue;
-    if (queue) {
```
BYK: Why remove this?

src/reporters/base-reporter.js
olingern: I think it's nice to be able to swap libs as needed. If any other parts of the codebase decide to use this utility, swapping the external libs would become easier this way. To your point, though, I'm not sure it will ever be used outside of this context.

olingern:
@BYK  A few changes for the flag pushed. Going to try to write some tests around flag output, and that should get things in a merge-able state.

BYK:
@olingern make sure to merge from master since you seem to have multiple merge conflicts right now.

src/reporters/base-reporter.js
BYK: I think swapping out libraries can still be done if needed by aliasing `humanizeBytes` to something else.

__tests__/reporters/base-reporter.js
```diff
@@ -125,3 +125,18 @@ test('BaseReporter.prompt', async () => {
   expect(error).not.toBeUndefined();
   reporter.close();
 });
+
+test('BaseReporter.getTotalNetworkUsage', async () => {
+  const reporter = new BaseReporter();
+
+  try {
+    const buffer1 = Buffer.alloc(10000);
```
BYK: Why do you need to allocate a buffer to report its size? :)

__tests__/util/humanize-bytes.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  test('appends the correct unit', () => {
+
+    const zero = 0;
+    const tenThousand = 10000;
+    const oneMillion = 1000000;
+    const oneHundredMillion = 100000000;
+    
+    const zeroHumanized = humanizeBytes(zero);
```
BYK: Don't think these intermediary variables are adding any value.

__tests__/util/humanize-bytes.js
```diff
@@ -0,0 +1,49 @@
...
+    const tenThousandHumanized = humanizeBytes(tenThousand);
+    const oneMillionHumanized = humanizeBytes(oneMillion);
+    const oneHundredMillionHumanized = humanizeBytes(oneHundredMillion);
+    
+
+    expect(getLastChars(zeroHumanized, 1)).toEqual('B');
```
BYK: Why not check the whole string to make sure the formatting and the numbers are also correct?

src/cli/index.js
```diff
@@ -243,7 +244,7 @@ export function main({
 
     return command.run(config, reporter, commander, commander.args).then(exitCode => {
       if (outputWrapper) {
-        reporter.footer(false);
+        reporter.footer(true, false);
```
BYK: Not sure what's happening here 😕

src/config.js
```diff
@@ -367,13 +369,17 @@ export default class Config {
     this.ignoreScripts = !!opts.ignoreScripts;
 
     this.disablePrepublish = !!opts.disablePrepublish;
+    this.showNetworkUsage = !!opts.showNetworkUsage;
 
     // $FlowFixMe$
     this.nonInteractive = !!opts.nonInteractive || isCi || !process.stdout.isTTY;
 
+    this.showNetworkUsage = !!opts.showNetworkUsage;
```
BYK: Why do we need this in two different places?

src/reporters/base-reporter.js
```diff
@@ -170,6 +172,14 @@ export default class BaseReporter {
     }
   }
 
+  logNetworkUsage(bytes: number) {
```
BYK: Why is this indirection needed if people can access `this.networkUsage` directly?

src/reporters/base-reporter.js
```diff
@@ -210,7 +220,7 @@ export default class BaseReporter {
   header(command: string, pkg: Package) {}
 
   // the screen shown at the very end of the CLI
-  footer(showPeakMemory: boolean) {}
+  footer(showPeakMemory: boolean, showNetworkUsage: boolean) {}
```
BYK: Oh, now I see what is happening above. Please avoid `boolean` arguments and convert this into an `options` object:
```
footer({showPeakMemory?: boolean, showNetworkUsage?: boolean} = {showPeakMemory: false, showNetworkUsage: false})
```

src/reporters/console/console-reporter.js
```diff
@@ -157,6 +157,11 @@ export default class ConsoleReporter extends BaseReporter {
       const peakMemory = (this.peakMemory / 1024 / 1024).toFixed(2);
       msg += ` Peak memory usage ${peakMemory}MB.`;
     }
+
+    if (showNetworkUsage) {
+      this.log(this._prependEmoji(`${this.getTotalNetworkUsage()} transferred`, '⬇️'));
```
BYK: Don't forget to move this to `en.js` and use `reporter.message` for translations.

src/util/humanize-bytes.js
BYK: Ping :)

src/util/request-manager.js
```diff
@@ -434,17 +443,23 @@ export default class RequestManager {
     this.reporter.verbose(this.reporter.lang('verboseRequestStart', params.method, params.url));
 
     req.on('error', onError);
-
-    const queue = params.queue;
-    if (queue) {
-      req.on('data', queue.stillActive.bind(queue));
+    
+    // only bind to request event emitter if --show-netowrk-usage is passed via cmd line
```
BYK: Nice. I think you still need to keep the old `queue.stillActive` logic.

olingern:
Will rebase and address all questions in the next day. Thanks for the review!

olingern:
@BYK  Addressed most of your comments, fixed linting, and rebased. Still have to go back and address the translations comment and the active queue comment.

BYK:
@olingern thanks a lot! Ping me when you think the PR is ready for another round.

arcanis:
Ping @olingern? 😃

olingern:
@arcanis 👋 thanks for the ping :)

I have an abundance of free time over the next couple of weeks, so I should be able to wrap this up.

Sounds like I need to:
1. Rebase master over this branch
2. Address @BYK's comment on `this.networkUsage`
3. Get the tests passing and ensure the feature doesn't affect anything when behind a flag
4. Ensure output when the flag tests matches what's expected, and create snapshot tests for that.

olingern:
I now have the oldest PR 😅 ... closing for now.

If people would still be interested in this, I'll look at re-implementing this off the latest

PR #4406, opened 2017-09-11
Bolds the titles in a list

orta:
**Summary**

Not substantial. It makes it easier to differentiate the items in a list vs the hint. Here is the before:

screen shot 2017-09-11 at 3 18 25 pm

and the after:

screen shot 2017-09-11 at 3 20 57 pm

**Test plan**

I ran the tests, and only ones that failed on master failed for me here 🥇

### DIFF
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -134,7 +134,7 @@ export default class ConsoleReporter extends BaseReporter {
 
     if (hints) {
       for (const item of items) {
-        this._log(`${' '.repeat(gutterWidth)}- ${item}`);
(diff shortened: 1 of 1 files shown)


BYK:
>I ran the tests, and only ones that failed on master failed for me here 🥇

Master should not have any failing tests :)

orta:
Aye, which I believe, figure it's just some weird edge-case  with my high sierra / node setup

BYK:
Thanks! Will merge once TravisCI passes. Looks like we have some flaky tests :)

BYK:
Will merge after 1.0.2 is released.

PR #4408, opened 2017-09-11
Fix: yarn bin should produce output even when silent

### LINKED ISSUE #4371: yarn bin returns blank when included in a package.json script

Report a bug.

Using `yarn bin` from within a script results in an empty value.

Create a script that utilizes `yarn bin`. Example:

```json
{
  "scripts": {
    "bug": "echo $(yarn bin)"
  }
}
```

Run `yarn bug` and see that no value is output. See the following output from my [test repo](https://github.com/s992/yarn-bug):

Yarn 0.27.5: 

```
~/code/yarn-bug
▶ yarn --version
0.27.5

~/code/yarn-bug
▶ yarn bug
yarn bug v0.27.5
$ echo $(yarn bin)
/Users/me/code/yarn-bug/node_modules/.bin
Done in 0.40s.
```

Yarn 1.0.0:

```
~/code/yarn-bug
▶ yarn --version
1.0.0

~/code/yarn-bug
▶ yarn bug
yarn bug v1.0.0
$ echo $(yarn bin)

✨  Done in 0.58s.
```

Using `yarn bin` from within a script results in the path to the Yarn `.bin` directory.

- Node version: 6.9.4 and 8.4.0
- Yarn version: 1.0.0
- OS version: MacOS 10.12.6

BYK:
**Summary**

Fixes #4371.

**Test plan**

Manual verification: `yarn --silent bin` or `YARN_SILENT=1 yarn bin` should produce output.

### DIFF
--- a/src/cli/commands/bin.js
+++ b/src/cli/commands/bin.js
@@ -14,6 +14,6 @@ export function setFlags(commander: Object) {}
 
 export function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const binFolder = path.join(config.cwd, config.registries[RegistryYarn.registry].folder, '.bin');
-  reporter.log(binFolder);
+  reporter.log(binFolder, {force: true});
   return Promise.resolve();
 }


src/cli/commands/bin.js
```diff
@@ -14,6 +14,6 @@ export function setFlags(commander: Object) {}
 
 export function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const binFolder = path.join(config.cwd, config.registries[RegistryYarn.registry].folder, '.bin');
-  reporter.log(binFolder);
+  reporter.log(binFolder, {force: true});
```
kaylieEB: I came across this [line](https://github.com/yarnpkg/yarn/blob/593e243364d28d4f106cd60b35a3d7de4ba37ae0/src/reporters/base-reporter.js#L200) but I can't think of a better solution atm :)

BYK: I asked @arcanis to add that line 😃

Once we have some more time, we need to think about logging levels and warnings again.

s992:
Thank you for the quick fix!

lukebatchelor:
We were affected by this as well, and I was wondering if anyone might be able to shed some light on how it happened.

We have a post install script which has something like

```sh
#!/usr/bin/env bash
set -e

BASEDIR=$(dirname $0)
BIN_PATH=$(yarn bin)
CHALK="$BIN_PATH/chalk"
```

As you found, in 1.0.1 `$(yarn bin)` returns an empty string, but interestingly, if the postinstall script is called by npm rather than yarn (`npm install` rather than `yarn install`) it actually still works!

My two questions are, what caused yarn to run silently when running in a subshell? But also, why would this be different when called when npm? Does npm spawn its tasks differently to yarn maybe?

In either case, thanks heaps for this fix!

arcanis:
All nested Yarn process are implicitely silenced (we set a `YARN_SILENT=1` in the environment when running the scripts).

PR #4411, opened 2017-09-12
Fix #4369: Include pull requests in git ref search

### LINKED ISSUE #4369: Yarn 1.X - Can't install Github Pull Requests anymore
Bug

Yarn 0.X allowed you to install PR commits by referencing the PR number.  I've seen a few ways to do this, but it's always worked for me as:
`yarn add angular-ui/ui-layout#228/head`

Using 1.X, you now get:
```
yarn add v1.0.1
info No lockfile found.
[1/4] Resolving packages...
error Couldn't find match for "228/head" in "_c" for "https://github.com/angular-ui/ui-layout.git".
```

`yarn init; yarn add angular-ui/ui-layout#228/head`

It installs the head commit of the Pull Request.

Node 8.4
Yarn 1.0.1
Linux Mint 18.3

kaylieEB:
**Summary**

Fixes https://github.com/yarnpkg/yarn/issues/4369, a regression from previous refactor.
Include `pull` requests in ref search and test for it later when it's called by github resolver.

*BEFORE*
before

*AFTER*
after

**Test plan**

Added a unit test in `git/git-ref-resolver.js`


arcanis:
FYI I don't think this was a regression but rather a non-feature - cf #3012 which was implementing it, but that we closed (rational at the time was that if people want to use PR branches as a package source, then they should use the source repository rather than the target repository).

PR #4416, opened 2017-09-12
Fix: actual and expected hashes are swapped

dotchev:
**Summary**

This PR fixes the error message for hash mismatches where the actual and expected hashes were swapped in the error message.

**Test plan**

Manual verification.


BYK:
@dotchev hey, thanks for the fix! Looks like you need to update the snapshots for tests regarding this. You can do it by running `yarn jest -u`.

Can you update the patch with those changes to tests pass?

PR #4419, opened 2017-09-12
Fix: always output reporter.inspect values and `config get` result

### LINKED ISSUE #3922: Yarn config is empty when run from within an npm script

Bug

The output of `yarn config` is empty if run from within a yarn script.

1. Create an empty project with a script in `package.json`: 
```
"scripts": {
    "debug": "yarn config list"
}
```
2. Run the script with `yarn run debug`.

Expect `yarn run debug` to produce the same output as `npm run debug` or `yarn config list`.

node.js v6.11.0
yarn v0.27.5
OS Ubuntu 16.04

Seems this was introduced sometime between v0.24.5 and v0.27.5. We were using this behavior as part of a docker build script that would internally grab an NPM authorization token (from `yarn config`) to provide it to the docker container being built. Would be nice to see the regression fixed, to bring this back into line with npm's behavior.

BYK:
**Summary**

Fixes #3922. Certain commands in yarn should always produce an
output, regardless of the silent status. This makes
`reporter.inspect` for console reporter to always produce an
output. It also forces the output of `yarn config get` to override
silent mode.

**Test plan**

Should add integration tests.

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -26,7 +26,7 @@ export const {run, setFlags, examples} = buildSubCommands('config', {
       return false;
     }
 
-    reporter.log(String(config.getOption(args[0])));
+    reporter.log(String(config.getOption(args[0])), {force: true});
     return true;
   },
 
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -126,7 +126,7 @@ export default class ConsoleReporter extends BaseReporter {
       });
     }
 
-    this.log('' + value);
+    this.log('' + value, {force: true});
   }
 
   list(key: string, items: Array<string>, hints?: Object) {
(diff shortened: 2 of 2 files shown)


BYK:
@kaylieEB I think you were referring to `this.log('' + value, {force: true});`. I changed it to use `String()` instead for explicit string casting.

kaylieEB:
Sorry I was referring to using template literals ``` `${value}` ```, but this already looks good to merge :).

BYK:
>Sorry I was referring to using template literals `${value}`, but this already looks good to merge :).

That is still implicit string conversion. I opted for the explicit `String()` one :) We should probably enable https://eslint.org/docs/rules/no-implicit-coercion in our code base.

kaylieEB:
interesting, i was suggesting it more for readability but being explicit makes better sense, thanks :)

PR #4431, opened 2017-09-13
Fix fakeroot detection

### LINKED ISSUE #4430: Global installs with root are installed in wrong directory
A bug

When installing something globally with the `root` user it gets installed into `/root/.config/yarn` instead of `/usr/local/share`.
The command I used was: `yarn global add selenium-standalone`. 
The installed binary gets linked to `/usr/local/bin/selenium-standalone`. It now only works for the `root` user, since no other user can access the install directory. 

Call `yarn global add selenium-standalone` as root. 

The package should be installed in `/usr/local/share`. 

node v8.5.0`
yarn 1.0.1
Ubuntu 16.04

martinstuecklschwaiger:
I think that fixes #4430

`fakeroot` detection was added in #3924 (7a053e2ca07d19b2e2eebeeb0c27edaacfd67904); it actually tests the exact opposite though. 

For testing this I added the line
```JavaScript
console.log("isFakeRoot", isFakeRoot());
```

I then tried the lines
```
./yarn global add
sudo ./yarn global add
fakeroot ./yarn global add
```

I also tried if `ROOT_USER` is now correct in [user-home-dir.js](https://github.com/yarnpkg/yarn/blob/0ef3bf1f03d128e4793eca3fe3b5f93542451f15/src/util/user-home-dir.js) for these 3 commands. 
```JavaScript
console.log("ROOT_USER", ROOT_USER);
```

### DIFF
--- a/src/util/root-user.js
+++ b/src/util/root-user.js
@@ -7,9 +7,9 @@ function getUid(): ?number {
   return null;
 }
 
-export default isRootUser(getUid()) && isNotFakeRoot();
+export default isRootUser(getUid()) && !isFakeRoot();
 
-export function isNotFakeRoot(): boolean {
+export function isFakeRoot(): boolean {
   return Boolean(process.env.FAKEROOTKEY);
 }
 


BYK:
Oh boy, thank you very much for catching this and for the quick fix! I can't believe how we missed the negation in that function in the original PR.

I feel like we should have cursory tests for this so it never happens again (and it wouldn't have happened at all if we wrote tests). Would you mind adding one or two tests for this root detection?

martinstuecklschwaiger:
Since these are all single line functions the only meaningful tests would be by actually executing this using `fakeroot`, or am I wrong here?

It's my first time digging into `yarn` code, would you mind helping me a bit with adding tests here? Or at least give me some pointers on where to begin?

BYK:
>Since these are all single line functions the only meaningful tests would be by actually executing this using fakeroot, or am I wrong here?

Not really :) You can always "fake" things.

>It's my first time digging into yarn code, would you mind helping me a bit with adding tests here? Or at least give me some pointers on where to begin?

Of course. Take a look at https://github.com/yarnpkg/yarn/blob/master/__tests__/util/git/git-spawn.js#L77-L90 for instance which modifies `process.env` before calling a function and then restores it.

What I'd do is to write at least two tests, one where I explicitly set `process.env.FAKEROOTKEY=''` to test "no fakeroot" case and the other where I explicitly set it to something truthy to test the positive case.

I'm going to merge this PR though since it solves an important bug and I want to get it into 1.0.2 so please send a new PR just for tests :)

PR #4434, opened 2017-09-13
Fix #4346 yarn upgrade interactive exit

### LINKED ISSUE #4346: yarn upgrade-interactive exits with error
Bug

Prior to v1, if you CTRL+C out of `yarn upgrade-interactive`, it would exit silently. Now, it will exit with this error:

```
Error: Promise rejected with value: undefined
    at /usr/local/Cellar/yarn/1.0.1/libexec/lib/cli.js:71593:12
    at Array.forEach (native)
    at EventEmitter.<anonymous> (/usr/local/Cellar/yarn/1.0.1/libexec/lib/cli.js:71589:24)
    at emitTwo (events.js:106:13)
    at EventEmitter.emit (events.js:191:7)
    at emit (/usr/local/Cellar/yarn/1.0.1/libexec/lib/cli.js:67713:11)
    at processEmit [as emit] (/usr/local/Cellar/yarn/1.0.1/libexec/lib/cli.js:67786:5)
```

1. run `yarn upgrade-interactive`
2. type `CTRL+C`

Exits without error

node.js v6.11.1
yarn v1.0.1
macOS 10.12.6

olingern:
**Summary**

A combination of changes have caused `yarn upgrade-interactive` to exit with a promise rejection. 

In short, I believe it has always been a problem, but #3995 exposed it to the prompt. 

The child rejection inside of [upgrade-interactive](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade-interactive.js#L152) is fine, as it was handled by the `Promise.race` condition; however, rejecting at the parent level inside of [console-reporter](https://github.com/yarnpkg/yarn/blob/master/src/reporters/console/console-reporter.js#L458) causes[ loud-rejection](https://github.com/sindresorhus/loud-rejection) to handle this.

I believe @arcanis 's PR #4283 is what would allow us not to hook into `SIGINT` inside of the console reporter and allow the reporter to cleanly close itself.

**Test Plan**

Will work on some scenarios! This PR needs some more verification on my end ... @BYK @torifat @arcanis please jump in and provide any feedback you think could be helpful! Opened early for visibility :)

### DIFF
--- a/src/reporters/console/console-reporter.js
+++ b/src/reporters/console/console-reporter.js
@@ -453,24 +453,9 @@ export default class ConsoleReporter extends BaseReporter {
       output: this.stdout,
     });
 
-    let rejectRef = () => {};
-    const killListener = () => {
-      rejectRef();
-    };
-
-    const handleKillFromInquirer = new Promise((resolve, reject) => {
-      rejectRef = reject;
-    });
-
-    rl.addListener('SIGINT', killListener);
-
     const {name = 'prompt', type = 'input', validate} = options;
-    const answers: InquirerResponses<string, T> = await Promise.race([
-      prompt([{name, type, message, choices, pageSize, validate}]),
-      handleKillFromInquirer,
-    ]);
+    const answers: InquirerResponses<string, T> = await prompt([{name, type, message, choices, pageSize, validate}]);
 
-    rl.removeListener('SIGINT', killListener);
     rl.close();
 
     return answers[name];


BYK:
I think @arcanis has the most context here so assigned this PR to him.

arcanis:
Hm I don't know much to this part of the code, but this seems sound. I'm not entirely sure why the SIGINT catch was required; according to my test inquirer doesn't return when it happens, so we don't need to catch it.

cc @torifat, do you remember why this code was needed?

olingern:
@arcanis Probably not ready to merge yet. I'll need to do some more testing, but, at the least, I believe `upgrade-inteactive` will need some small updates since there's nothing handling rejection [here](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade-interactive.js#L152) that `Promise.race` previously was.

I'll need to figure out:
1. When do rejections happen? i.e. failed upgrade
2. What is the right recourse? i.e. ask for a retry, etc.

BYK:
@olingern I was hoping to merge this and publish 1.0.2 with this included. If you want to hold it off though, I'll wait until you say we're good to merge.

torifat:
Sorry, for being late here. I don't have internet for about a week 😭 

The reason for using that `SIGINT` listener was to prevent yarn from quiting unexpectedly. Unless they have changed their code if you <kbd>Ctrl + C</kbd> then yarn will not quit properly and you will not get done with the timings.

But, like @olingern mentioned it is being handle properly after @arcanis's PR.

BYK:
>Sorry, for being late here. I don't have internet for about a week 😭

No worries and I'm glad you're back!

PR #4435, opened 2017-09-13
Add test for isFakeRoot

martinstuecklschwaiger:
This adds a test for `isFakeRoot()` as proposed by @BYK in #4431.

### DIFF
--- a/__tests__/util/root-user.js
+++ b/__tests__/util/root-user.js
@@ -1,9 +1,20 @@
 /* @flow */
 
-import {isRootUser} from '../../src/util/root-user.js';
+import {isRootUser, isFakeRoot} from '../../src/util/root-user.js';
 
 test('isRootUser', () => {
   expect(isRootUser(null)).toBe(false);
   expect(isRootUser(1001)).toBe(false);
   expect(isRootUser(0)).toBe(true);
+});
+
+test('isFakeRoot', () => {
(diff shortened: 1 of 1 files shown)


__tests__/util/root-user.js
```diff
@@ -1,9 +1,20 @@
...
+import {isRootUser, isFakeRoot} from '../../src/util/root-user.js';
 
 test('isRootUser', () => {
   expect(isRootUser(null)).toBe(false);
   expect(isRootUser(1001)).toBe(false);
   expect(isRootUser(0)).toBe(true);
 });
+
+test('isFakeRoot', () => {
+  delete process.env.FAKEROOTKEY;
```
BYK: You should explicitly set this to `''` or something since deleting would just restore it to the default value if the system running this. So if the tests were run with `fakreoot` then deleting that env variable would just reset it to the system value.

Makes sense?

martinstuecklschwaiger: I tried it out, and `delete process.env.FAKEROOTKEY` actually removes the key from the object. 
Running the tests with `fakeroot` actually works.

BYK: Alright. I've tested locally to see the same result. That said be advised that deleting it like this actually removes any existing value without any way for restoring it so your test is actually affectin the global state permanently.

Are you sure you want to do that?

PR #4444, opened 2017-09-13
Fix: always use origin/master instead of master for homebrew

### LINKED ISSUE #3415: update-homebrew.sh needs to run "git pull" before running " brew bump-formula-pr"
Otherwise the build fails (#3414) and it needs to be ran a second time to actually work.

BYK:
**Summary**

Fixes #3415. The old homebrew update script updated the repo
from remote but did not update the local `master` branch, thus
had the potential for conflicts and mismatches. This patch
forces it to use `origin/master` instead and uses the latest
up-to-date version from the source.

**Test plan**

N/A

### DIFF
--- a/scripts/update-homebrew.sh
+++ b/scripts/update-homebrew.sh
@@ -23,11 +23,10 @@ PATH=$PATH:$HOME/.linuxbrew/bin/
 pushd ~/.linuxbrew/Library/Taps/homebrew/homebrew-core
 #git remote set-url origin https://github.com/Daniel15/homebrew-core # for testing
 git remote set-url origin https://github.com/homebrew/homebrew-core
-git reset --hard HEAD
-git clean -fd
 git fetch --prune origin
+git reset --hard origin/master
+git clean -fd
 # Remove any existing branch (eg. if the previous attempt failed)
-git checkout master
 git branch -D yarn-$version || true
 popd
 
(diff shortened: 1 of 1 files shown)


scripts/update-homebrew.sh
```diff
@@ -23,11 +23,10 @@ PATH=$PATH:$HOME/.linuxbrew/bin/
 pushd ~/.linuxbrew/Library/Taps/homebrew/homebrew-core
 #git remote set-url origin https://github.com/Daniel15/homebrew-core # for testing
 git remote set-url origin https://github.com/homebrew/homebrew-core
-git reset --hard HEAD
-git clean -fd
 git fetch --prune origin
+git reset --hard origin/master
+git clean -fd
 # Remove any existing branch (eg. if the previous attempt failed)
-git checkout master
```
Daniel15: Will `git reset --hard origin/master` switch to the `master` branch? I guess that's why this is not needed?

Could you try switching to a different branch, then running this script, and ensuring it switches to master?

BYK: Ah, my bad. Didn't realize we had to be on master.

BYK:
@Daniel15 ping :)

Daniel15:
Thanks! Let's see if this makes the automated update work next time :)

PR #4447, opened 2017-09-13
Update: add DNS caching to request manager

### LINKED ISSUE #746: Doesn't work well on a flaky connection
Not sure.

I'm at an airport today, so I've been tethering on my phone and using in-flight WiFi.  Every time I try to run `yarn`, I eventually get a `socket hang up` error on an arbitrary package.  Presumably, it doesn't handle mobile connectivity very well.  `npm install` works in these same scenarios.

- Be able to install a particular module without worrying about the rest of the installed modules.
- Retry requests if one times out.

yarn v0.15.1
node v6.2.1
macOS v10.11.6

BYK:
**Summary**

Fixes #746. Unfortunately, neither Node, nor many systems come with
built-in DNS caching so the many parallel requests that Yarn makes
sometimes overwhelm the DNS servers, and most of the time, for the
very same domain(s). Even worse, we pay the DNS look up cost for
each request, which is quite sad at best. This patch introduces
the `dnscache` module which intercepts all DNS look ups and answers
them from an in-memory cache when possible. This applies to the
built-in `http` and `https` modules, used by `request`.

**Test plan**

Existing tests should pass, and hopefully be faster. Total number of
DNS look ups should decrease dramatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "death": "^1.0.0",
     "debug": "^2.2.0",
     "detect-indent": "^5.0.0",
+    "dnscache": "^1.0.1",
     "eslint-plugin-jest": "^20.0.3",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-relay": "0.0.8",
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -1,5 +1,11 @@
 /* @flow */
 
+import fs from 'fs';
+import url from 'url';
+import dnscache from 'dnscache';
+import invariant from 'invariant';
+import RequestCaptureHar from 'request-capture-har';
+
 import type {Reporter} from '../reporters/index.js';
 import {MessageError} from '../errors.js';
 import BlockingQueue from './blocking-queue.js';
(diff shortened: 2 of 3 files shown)


Daniel15:
70 KB increase seems like a lot, but I guess it's worth it 😛 

I'm surprised the operating system doesn't do DNS caching? Does this mean that every app needs to implement its own cache? 😢

bestander:
Good catch.
Is it JS only btw, does it bundle well?

BYK:
@bestander see https://github.com/yarnpkg/yarn/pull/4447#issuecomment-329346701

I think it is acceptable since it is a 2% increase.

>I'm surprised the operating system doesn't do DNS caching? Does this mean that every app needs to implement its own cache? 😢

Some do some doesn't and apparently even if they do, they are not great. I know Chrome comes with its own DNS cache and resolution system for instance. Windows is supposed have the cache but those Docker instances sure don't and I'm fairly certain that leaving the app for the exact same domain thousands of times (even if you have the OS cache) is not great.

staabm:
any numbers on which scale this changes improves performance?

BYK:
@staabm I didn't see any noticeable improvements in performance for my installs since almost all of them are from cache + lock file. I added this mostly for #746 where apparently Yarn was saturating DNS servers in those circumstances.

PR #4448, opened 2017-09-13
Fix: incorrect optional ignore in nested dependencies

### LINKED ISSUE #4445: error installing phantomjs-prebuilt, with 1.0.x, --ignore-optional, and 2 other deps
bug

Example error (it doesn't always fail on this module)

```
error /Users/billy/Dev/phantomjs-prebuilt/node_modules/phantomjs-prebuilt: Command failed.
Exit code: 1
Command: node install.js
Arguments:
Directory: /Users/billy/Dev/phantomjs-prebuilt/node_modules/phantomjs-prebuilt
Output:
module.js:491
    throw err;
    ^

Error: Cannot find module 'process-nextick-args'
    at Function.Module._resolveFilename (module.js:489:15)
    at Function.Module._load (module.js:439:25)
    at Module.require (module.js:517:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/billy/Dev/phantomjs-prebuilt/node_modules/readable-stream/lib/_stream_readable.js:26:23)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
```

https://github.com/billyvg/yarn-phantomjs-repro

Above repo will fail to install with
`yarn install --ignore-optional`

To install.  Note this works in `0.27.5_1`

node 8.4.0
yarn 1.0.2
OSX 10.12.6

BYK:
**Summary**

Fixes #4445. The bug was introduced in
https://github.com/yarnpkg/yarn/pull/3976/files#diff-7066979f95168f7bb59e4f9cfb3ba8fcR211
with an incomplete package optional check. This caused many transient
dependencies to be ignored at linking stage when `--ignore-optional`
flag is passed. This patch fixes the check to combine the flag with
the package's actual optional mark.

**Test plan**

Manually verified #4445 is resolved. Also added a unit test.


BYK:
>Also added a unit test.

This is not true yet, obviously. I will add a unit test soon and won't merge without tests for this.

src/package-hoister.js
```diff
@@ -242,7 +242,7 @@ export default class PackageHoister {
           continue;
         }
 
-        const isMarkedAsOptional = !depinfo.pkg._reference || this.ignoreOptional;
+        const isMarkedAsOptional = !depinfo.pkg._reference || (depinfo.pkg._reference.optional && this.ignoreOptional);
```
kaylieEB: i think it'd be more readable to do `depinfo.pkg._reference && depinfo.pkg._reference.optional && this.ignoreOptional`

src/package-request.js
```diff
@@ -193,6 +193,7 @@ export default class PackageRequest {
     invariant(ref, 'Resolved package info has no package reference');
     ref.addRequest(this);
     ref.addPattern(this.pattern, resolved);
+    ref.addOptional(this.optional);
```
BYK: The new test case revealed another bug here: if a package is listed as optional by one package and as a strict dependency by another, it should be non-optional at the end. However, due to incorrect request caching below and the lack of this line here, it's optional status was random :(

BYK:
>If tests are passing I am fine with it.
>I would suggest adding a separate test case for this specific issue instead of making existing one bigger

I was only able to find the additional issues when writing tests. Are you suggesting that I split this into two tests one with nested optional dependencies and keep the old one in addition to them?

src/package-request.js
kaylieEB: nice catch!

src/package-resolver.js
```diff
@@ -477,12 +475,12 @@ export default class PackageResolver {
...
 
-    const fetchKey = `${req.registry}:${req.pattern}`;
+    const fetchKey = `${req.registry}:${req.pattern}:${String(req.optional)}`;
 
-    if (this.fetchingPatterns[fetchKey]) {
+    if (this.fetchingPatterns.has(fetchKey)) {
       return;
     } else {
-      this.fetchingPatterns[fetchKey] = true;
+      this.fetchingPatterns.add(fetchKey);
```
kaylieEB: this else block isn't necessary since other wise it'll just return?

src/package-hoister.js
```diff
@@ -242,7 +242,7 @@ export default class PackageHoister {
           continue;
         }
 
-        const isMarkedAsOptional = !depinfo.pkg._reference || this.ignoreOptional;
+        const isMarkedAsOptional = depinfo.pkg._reference && depinfo.pkg._reference.optional && this.ignoreOptional;
```
kaylieEB: sorry I deleted the previous comment because I realized the logic isn't quite the same. The previous one is saying mark as optional if `depinfo.pkg._reference` doesn't exist (or if ignoreOptional flag). That seemed a little weird to be but if that's still the case, we should probably restore it to the way you had it.

src/package-resolver.js
BYK: That's right, I'll fix it.

BYK: Cost me about 3 hours to catch it :D

src/package-hoister.js
BYK: I know they are not equivalent but no tests failed and I feel like it makes more sense with the new version. What do you think?

kaylieEB: Works for me. LET'S LIVE ON THE EDGE. 🎉

acuper:
Should the night build contain the fix ?

BYK:
@acuper this change went live with the 1.3.2 release

acuper:
OK, I have 1.3.2, so for me --ignore-optional is still not working properly.
Will try to provide the steps how to reproduce but it might hard because this is happening in our internal projects where dependency hierarchy is huge.

Is there older version of yarn witch should has no the bug introduced inhttps://github.com/yarnpkg/yarn/pull/3976/files#diff-7066979f95168f7bb59e4f9cfb3ba8fcR211
?

acuper:
I think I found a way to reproduce it:

`
{
  "dependencies": {
    "internal-package": "^4.1.0",
    "yargs": "^9.0.1"
  }
}
`
now internal-package has this as optionalDependency:
`"yargs": "^7.0.2"`

the result of `yarn --ignore-optional` is missing package: is-fullwidth-code-point 
which dependency path is: yargs -> string-width -> is-fullwidth-code-point

BYK:
@acuper can you please create a new issue so we can track and debug this properly. I'd love to fix of this is broken.

PR #4461, opened 2017-09-14
Fix: Handle `null` versions in package.json

### LINKED ISSUE #4429: An unexpected error occurred: "Path must be a string. Received null".

bug

after updating yarn to latest version(1.0.1) getting error during `yarn install`
```
yarn install v1.0.1
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "Path must be a string. Received null".
```

here's the [yarn-error.log](https://gist.github.com/hbakhtiyor/4599c23300998e8cd33c0692bb810ab1#file-yarn-error-log) file

node.js: v8.5.0
yarn: 1.0.1
os: ubuntu 16.04 64bit

BYK:
**Summary**

Fixes #4429. Apparently there's an advice on the internet to use
`null` as the package version to install the latest and even worse,
some people are actually doing this. This patch makes sure the 
version is at least a string (albeit an empty one) when fixing the 
manifest.

**Test plan**

Added new integration test.

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -623,6 +623,12 @@ test.concurrent('install with comments in manifest', (): Promise<void> => {
   });
 });
 
+test.concurrent('install with null versions in manifest', (): Promise<void> => {
+  return runInstall({}, 'install-with-null-version', async config => {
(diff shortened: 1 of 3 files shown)


bestander:
typeof null is "object"!
I had to see this bug with my own eyes.

BYK:
@bestander or just use flow? 😃

PR #4463, opened 2017-09-15
Simplify import modules in tests

Connormiha:
**Summary**
There are many relative imports in tests files like `../../../src/module-folder/module-name`.
I think import without relative parent folder is more comfortable
```js
import {Install, run as install} from '../../../src/cli/commands/install.js';
// vs
import {Install, run as install} from 'cli/commands/install.js';
```
It's enough to rewrite default Jest's rule for resolve modules.
```json
"moduleDirectories": ["node_modules", "src"]
```

**Test plan**

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -1,15 +1,15 @@
 /* @flow */
 
-import Lockfile from '../../src/lockfile';
-import {ConsoleReporter} from '../../src/reporters/index.js';
-import {Reporter} from '../../src/reporters/index.js';
-import {parse} from '../../src/lockfile';
-import * as constants from '../../src/constants.js';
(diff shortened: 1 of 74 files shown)


BYK:
Looks like it is not working on CI for some reason.

Btw. I do not think we should be doing this. Custom module search paths remove the distinction between dependencies and local requires. What do you think? Also may be others think differently? //cc @yarnpkg/core

rally25rs:
IMO the relative paths are a) the way Node has worked by default since forever, and b) more intuitive for the majority of JS devs who may contribute to the project (when most devs see `import * from 'foo'` they assume `foo` is a dependency).

BYK:
@Connormiha thanks a lot for the PR but I'll go ahead and close this since this is not something we want to do.

PR #4464, opened 2017-09-15
fix(init): prevent init command from traversing up (#4441)

### LINKED ISSUE #4441: Bug: `yarn init` should not traverse up to find a package.json
This is with `yarn 1.0.2` and `node 8.4.0` installed via FreeBSD 11 packages.  In `~/tmp/test-yarn` I was stumped why neither `package.json` nor `yarn.lock` files were being created with `yarn init`.  Because `pacakage.json` and `yarn.lock` files existed in `~/`, it seems `yarn` walked up the directory tree to find the project root and run the command from there.  If this is intended behaviour, maybe it should be documented to prevent the confusion I had.

```
jrm@phe ~/tmp/test_yarn % yarn init --verbose
yarn init v1.0.2
verbose 0.69 Checking for configuration file "/usr/home/jrm/.npmrc".
verbose 0.69 Checking for configuration file "/home/jrm/.npmrc".
verbose 0.691 Checking for configuration file "/usr/local/etc/npmrc".
verbose 0.691 Checking for configuration file "/usr/home/jrm/.npmrc".
verbose 0.691 Checking for configuration file "/usr/home/.npmrc".
verbose 0.691 Checking for configuration file "/usr/.npmrc".
verbose 0.692 Checking for configuration file "/usr/home/jrm/.yarnrc".
verbose 0.692 Found configuration file "/usr/home/jrm/.yarnrc".
verbose 0.693 Checking for configuration file "/home/jrm/.yarnrc".
verbose 0.693 Found configuration file "/home/jrm/.yarnrc".
verbose 0.693 Checking for configuration file "/usr/local/etc/yarnrc".
verbose 0.693 Checking for configuration file "/usr/home/jrm/.yarnrc".
verbose 0.693 Found configuration file "/usr/home/jrm/.yarnrc".
verbose 0.693 Checking for configuration file "/usr/home/.yarnrc".
verbose 0.693 Checking for configuration file "/usr/.yarnrc".
verbose 0.696 current time: 2017-09-13T19:06:24.994Z
question name (test_yarn):
question version (1.0.0):
question description (Blah):
question entry point (index.js):
question repository url (none):
question author (blah):
question license (AGPL-3.0):
question private (true):
success Saved package.json
Done in 11.95s.
jrm@phe ~/tmp/test_yarn % ls -lah
total 33
drwxr-xr-x   2 jrm  jrm     2B Sep 13 16:05 .
drwxr-xr-x  20 jrm  jrm    54B Sep 13 13:44 ..
```

yerguden:
**Summary**

Fixes #4441.

**Test plan**

Added new test case.


src/cli/index.js
```diff
@@ -443,7 +443,8 @@ export function main({
     return errorReportLoc;
   }
 
-  const cwd = findProjectRoot(commander.cwd);
+  const shouldUseProjectRoot = commandName !== 'init';
```
BYK: This is a good start! That said I think at this point we already know the command itself and this information is better stored on the command class so here's what I'd do:

 - Put a `shouldRunInCurrentCwd` property on the `init` command here: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/init.js#L25 (you'd need to split the `export` and `function` definitions there to be able to do this.
 - on the line below do this instead:
   ```
    const cwd = command.shouldRunInCurrentCwd ? commander.cwd : findProjectRoot(commander.cwd);
    ```

You can add tests to https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/init.js and the fixtures for these commands are at https://github.com/yarnpkg/yarn/tree/master/__tests__/fixtures/init

I'd add a new fixture with two nested folders where the higher one has a package.json and then run the command in the nested one which doesn't have `package.json`.

BYK:
@yerguden need help writing tests?

yerguden:
Hi @BYK, I have updated my pull request however i noticed that my test doesn't cover changes in `src/cli/index.js`. as i wasn't able to find any tests that import `src/cli/index.js` i wasn't sure importing that file for test purposes was the right course of action.

BYK:
@yerguden take a look at https://github.com/yarnpkg/yarn/blob/master/__tests__/integration.js which does integration tests by calling yarn through the command line. `cli/index.js` is the entry point so you cannot easily test it. The best way is the integration test. I'd to the following:

- Create a `package.json` in the test during test run
- Create another directory in there without a `package.json` file.
- Invoke `yarn init -y` in the nested directory.
- Check the nested directory now has a `package.json` file and the parent `package.json` is untouched.

You can have a look at the test I added for inspiration here: https://github.com/yarnpkg/yarn/pull/4498/files#diff-971abd8c6c8db00bc212414ea19e1807

__tests__/commands/init.js
```diff
@@ -12,6 +12,25 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
...
+    fixturesLoc,
+    (args, flags, config, reporter, lockfile): Promise<void> => {
+      return runInit(config, reporter, flags, args);
+    },
+    [],
+    {yes: true},
+    {source: 'init-nested', cwd: 'inner-folder'},
+    async (config): Promise<void> => {
+      const {cwd} = config;
+      const fileExistsInCurrentCwd = await fs.exists(path.join(cwd, 'package.json'));
```
BYK: Don't need this variable, hust put this into `expect()` below directly.

__tests__/commands/init.js
```diff
@@ -12,6 +12,25 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
...
+
+      expect(fileExistsInCurrentCwd).toEqual(true);
```
BYK: You should also check the parent `package.json` is unchanged after the command.

yerguden:
@BYK thanks for the pointers :) i will be updating accordingly, i was wondering one thing since we are moving test to integration should i remove the unit test for this case as it seems to serve no purpose.

BYK:
> i was wondering one thing since we are moving test to integration should i remove the unit test for this case as it seems to serve no purpose.

I think having both is fine.

PR #4471, opened 2017-09-15
Support save-exact true

### LINKED ISSUE #4343: Yarn save-prefix option changed/ignored
Bug

I've just upgraded from v0.20.3 to v1.0.1. The behavior described by @meznaric in #3045 was what I was previously seeing.

When I run `yarn config set save-prefix false` (the previous recommendation, as noted in #3045), installed packages now have the string `false` placed in front of the version. For instance, when I run `yarn add --dev rollup-plugin-uglify`, this is what Yarn adds to my `package.json`.

```javascript
"rollup-plugin-uglify": "false2.0.1"
```

If I run `yarn config set save-prefix ''` to try and remove the prefix, the option is ignored. Below is the string Yarn adds to `package.json` when I run the same command as above with the new config.

```javascript
"rollup-plugin-uglify": "^2.0.1",
```

Note that Yarn lists my NPM config as having the same options (as shown by `yarn config list`).

Yarn config excerpt:
```javascript
'save-prefix': '',
```

NPM config excerpt:
```javascript
'save-exact': true,
'save-prefix': '',
```

I have no idea if I am the problem here, as I cannot find any documentation about this setting.

This is related to #1088 and #3320.

I should be able to configure Yarn such that the versions of packages being added are exact (unless there is a good reason I should not be doing this? - if so, a warning is in order).

Node v6.11.2
Yarn v1.0.1
macOS 10.11

ahmedelgabri:
**Summary**

Fixes #4343. Currently there is no way to remove the package prefix inside `.yarnrc` file, this PR add support for `save-exact` in `.yarnrc` as discussed in #4343. Full credit goes to @jambonrose

```
save-exact true
```

One small thing, should `yarn` be backwards compatible with the old behavior which is `save-prefix ''`? We can just add an extra check here for do this. What do you think @BYK?

```js
} else if (exact || Boolean(this.config.getOption('save-exact')) || Boolean(this.config.getOption('save-prefix'))) {

```

**Test plan**

New unit test.

### DIFF
--- a/__tests__/fixtures/add/install-strict-all/.yarnrc
+++ b/__tests__/fixtures/add/install-strict-all/.yarnrc
@@ -1,0 +1,2 @@
+yarn-offline-mirror "./mirror-for-offline"
+save-exact true
--- a/src/cli/commands/add.js
+++ b/src/cli/commands/add.js
@@ -69,7 +69,9 @@ export class Add extends Install {
    * returns version for a pattern based on Manifest
    */
   getPatternVersion(pattern: string, pkg: Manifest): string {
-    const {exact, tilde} = this.flags;
+    const tilde = this.flags.tilde;
+    const configPrefix = String(this.config.getOption('save-prefix'));
+    const exact = this.flags.exact || Boolean(this.config.getOption('save-exact')) || configPrefix === '';
     const {hasVersion, range} = normalizePattern(pattern);
     let version;
 
@@ -86,7 +88,7 @@ export class Add extends Install {
       } else if (exact) {
         prefix = '';
       } else {
-        prefix = String(this.config.getOption('save-prefix')) || '^';
+        prefix = configPrefix || '^';
       }
 
       version = `${prefix}${pkg.version}`;
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -223,6 +223,17 @@ test.concurrent('add save-prefix should not expand ~ to home dir', (): Promise<v
   });
 });
 
+test.concurrent('add save-exact should make all package.json strict', (): Promise<void> => {
+  return runAdd(['left-pad'], {}, 'install-strict-all', async config => {
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+
+    expect(lockfile[0]).toMatch(/^left-pad@\d+\.\d+\.\d+:$/);
+    expect(JSON.parse(await fs.readFile(path.join(config.cwd, 'package.json'))).dependencies['left-pad']).toMatch(
+      /^\d+\.\d+\.\d+$/,
+    );
+  });
+});
+
 test.concurrent('add with new dependency should be deterministic 3', (): Promise<void> => {
   return runAdd([], {}, 'install-should-cleanup-when-package-json-changed-3', async (config, reporter) => {
     // expecting yarn check after installation not to fail


StefanoMagrassi:
This should be so obvious: if I set `save-exact` to true the `prefix` must be automatically set to an empty string (otherwise there is some kind of inconsistency).

The yarn parser should also allow to set `save-prefix` key to an empty string. Now it throwns an error ("Invalid value type in .yarnrc").

BYK:
>This should be so obvious: if I set save-exact to true the prefix must be automatically set to an empty string (otherwise there is some kind of inconsistency).

Agreed.

>The yarn parser should also allow to set save-prefix key to an empty string. Now it throwns an error ("Invalid value type in .yarnrc").

This is not true. You can set it to an empty string.

BYK:
>One small thing, should yarn be backwards compatible with the old behavior which is save-prefix ''? We can just add an extra check here for do this. What do you think @BYK?

It should definitely be backwards compatible, otherwise this change would require a major version upgrade.

src/cli/commands/add.js
```diff
@@ -83,7 +83,7 @@ export class Add extends Install {
       let prefix;
       if (tilde) {
         prefix = '~';
-      } else if (exact) {
+      } else if (exact || Boolean(this.config.getOption('save-exact'))) {
```
BYK: I think it is better to do
```
const exact = this.flags.exact || Boolean(this.config.getOption('save-exact'));
const tilde = this.flags.tilde;
```

above. `save-exact` should take precedence over `save-prefix` option.

StefanoMagrassi:
> This is not true. You can set it to an empty string.

Really? My colleague got that error when tried to set `save-prefix` to an empty string (yarn version 1.0.2).

Maybe it was his fault or caused by some kind of other wrong configuration... In this case :+1:

ahmedelgabri:
> This is not enough :) You also need to add a new test(s) to test the new behavior.

True :) I might need some pointers on how to test the `.yarnrc`, a quick look at https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/add.js & I can't seem to find anything related to testing `.yarnrc`, I might be wrong.

ahmedelgabri:
I added a test & changed the logic to be backward compatible.

```js
const exact =
      this.flags.exact ||
      Boolean(this.config.getOption('save-exact')) ||
      String(this.config.getOption('save-prefix')) === ''; 
```

The `save-prefix` check can't be `Boolean(this.config.getOption('save-prefix'))` becasue it will always return true for non-empty strings.

__tests__/fixtures/add/install-strict-all/.npmrc
```diff
@@ -0,0 +1 @@
+yarn-offline-mirror=./mirror-for-offline
```
BYK: Why a separate file? Let's move this into `.yarnrc` below.

__tests__/fixtures/add/install-strict-all/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "mime-types": "^2.0.0"
```
BYK: I don't think you need this at all since you already do `yarn add left-pad` in the test

src/cli/commands/add.js
```diff
@@ -69,7 +69,11 @@ export class Add extends Install {
    * returns version for a pattern based on Manifest
    */
   getPatternVersion(pattern: string, pkg: Manifest): string {
-    const {exact, tilde} = this.flags;
+    const tilde = this.flags.tilde;
+    const exact =
+      this.flags.exact ||
+      Boolean(this.config.getOption('save-exact')) ||
+      String(this.config.getOption('save-prefix')) === '';
```
BYK: Assigning this to a variable like `configPrefix` might be good since it is now used here and also below.

__tests__/fixtures/add/install-strict-all/.npmrc
ahmedelgabri: Honestly, I just copied `install-no-strict-all` and just added `.yarnrc` :)

__tests__/fixtures/add/install-strict-all/package.json
ahmedelgabri: I think I can remove `package.json` & `mirror-for-offline` folder too then?

__tests__/fixtures/add/install-strict-all/.npmrc
ahmedelgabri: Maybe actually this is not needed at all?

ahmedelgabri:
@BYK can you have a look at travis, I think it's failing because of disk/memory issues or something.

all the errors are like this

```
Error: ENOSPC: no space left on device, mkdir <some path>
```

__tests__/fixtures/add/install-strict-all/.yarnrc
```diff
@@ -0,0 +1,2 @@
+yarn-offline-mirror "./mirror-for-offline"
```
BYK: Offline mirror is for not using the internet to install packages during test AFAIK.

BYK:
>@BYK can you have a look at travis, I think it's failing because of disk/memory issues or something.

This is probably due to the RAMDISK on OS X builds. If it fails again, I'll edit your patch and increase the size.

BYK:
Merged! Thanks a lot for helping us with this ❤️

ahmedelgabri:
Thanks @BYK 😊

PR #4479, opened 2017-09-15
Fix: Better npm_config_argv emulation

### LINKED ISSUE #2226: `npm_config_argv` is not set properly in yarn scripts

Not clear. To be compatible with npm I guess it's a bug.

Environment variables are not set correctly.

    // File: env-test.js
    console.log(process.env);

and

    // File: package.json
    {
      "scripts": {
        "env-test": "node ./test.js"
      }
    }

`npm_config_argv` should be set in `process.env`, as well as the`npm_config_demo_arg` in this case.

    $ npm run env-test --demo-arg="abc" | grep 'abc'`
    # output:
    # npm_config_argv: '{"remain":[],"cooked":["run","env","--demo-arg","abc"],"original":["run","env","--demo-arg=abc"]}',
    # npm_config_demo_arg: 'abc',

    $ npm run env-test --demo-arg="abc" | grep 'abc'`
    # output is empty

    $ node --version
    # output: v7.2.0
    $ yarn --version
    # output: 0.18.0
    $ uname -a
    # output: Darwin name.local 15.6.0 Darwin Kernel Version 15.6.0: Thu Sep  1 15:01:16 PDT 2016; root:xnu-3248.60.11~2/RELEASE_X86_64 x86_64
    # (it's Mac OS X El Capitan 10.11.6 (15G1108))

BYK:
**Summary**

Fixes #2226. Better emulates `npm_config_argv` by passing
`process.argv.slice(2)` as the `original` portion and both the
command name and the script name in `cooked` portion.

**Test case**

Added integration tests.


BYK:
>Added integration tests.

Will actually add these soon.

indiescripter:
Thanks @BYK, this helps #2226 a little as you commented over there.

However, if I understand correctly (I've only recently started using yarn),
`yarn <scriptname>` is shorthand for `yarn run <scriptname>` so passing just `process.argv.slice(2)` as per summary above is not quite sufficient as `npm_config_argv` should ideally expand into the `npm` canonical form including the word "run" prior to `<scriptname>` in the (probably) `cooked` array.

You could go one better than npm and make that `run-script` as `run` is shorthand for `run-script` in npm so `run-script` is arguably the correct canonical form.

BYK:
@indiescripter thanks for the feedback! So here's what happens:
 - `original` is always the untouched `process.argv.slice(2)`
 - `cooked` is the "interpreted" version, which would always be in the form `['run', 'script_name']` even when the command invoked is `yarn script_name`

The aim here is to make this variable more useful but not achieve 100% npm cimpatibility since that's not possible and overly limiting.

TL;DR
Is this version making things easier and "better" for you or is it as useless as the current version?

indiescripter:
Thank you so much @BYK for taking such an interest in my situation.

So long as one way or another we end up with at least

`['run', 'script_name']` in `cooked`

that is useful to/better for me.

This means my use case will work if the user types `yarn run <scriptname>`

It would be nice, though, as I suggested, if the user typed just the shorthand `yarn <scriptname>`
then `yarn run <scriptname>` or `yarn run-script <scriptname>` was automagically inferred but asking for that extra step might be stretching a friendship!

BYK:
>It would be nice, though, as I suggested, if the user typed just the shorthand yarn <scriptname> then yarn run <scriptname> or yarn run-script <scriptname> was automagically inferred but asking for that extra step might be stretching a friendship!

Oh, I already fixed some of this part :) When you run `yarn foobar` you still get `['run', 'foobar']` in `cooked`. That was the core of your issue, right?

indiescripter:
Fantastic, correct, thank you @BYK, that was the core issue!

I have a new project in development which I humbly feel quite excited about as it a somewhat novel approach to monorepo setups (i.e. multiple subpackages under a master, top-level, development package).

As you have taken a special interest in "Better npm_config_argv emulation", I would like to show my appreciation for your effort by giving you a heads-up well ahead of making any social media announcements about it.

The project, called *buildverse* is on npmjs.com at

https://www.npmjs.com/package/buildverse

The "novelty" of my approach is in the executable program `npm-apply`. This utility provides much flexibility in running package scripts in a [depth-first traversal](https://en.wikipedia.org/wiki/Tree_traversal) of a tree of subprojects. Judicial use of `npm-apply` allows the package scripts developer to implement preorder, inorder & postorder traversal patterns as described in the preceding Wikipedia link.

If you check out the GitHub repo for *buildverse* you will find that the `npm-apply` code makes use of 'npm_config_argv' to recursively run a `npm ...` command in subproject directories.

With your fix, I hope to be able to say "Works with yarn as well as npm" in the package README!

BYK:
@indiescripter that's great, congrats! And I'm happy to be able to help :)

Looks like you may also contribute to yarn in the future since you seem to be interested in these kind of things ;)

PR #4482, opened 2017-09-16
Upgrade Chalk to take advantage of the latest (colour) fixes

KayLeung:
The latest Chalk has fixes to Windows 10 / MacOS and more would improve the color detection in different env.

https://github.com/chalk/supports-color/blob/master/index.js#L48
https://github.com/chalk/supports-color/blob/master/index.js#L78


KayLeung:
Hmm... needs some fix on vars access.

I'm waiting to see if it's an expected change: https://github.com/chalk/chalk/issues/209 before another patch

BYK:
@KayLeung anything you need from us to get the builds pass?

__tests__/reporters/console-reporter.js
```diff
@@ -12,8 +12,8 @@ const stream = require('stream');
 // ensures consistency across environments
 require('chalk').enabled = true;
 require('chalk').supportsColor = true;
-require('chalk').styles.blue.open = '\u001b[34m';
-require('chalk').styles.bold.close = '\u001b[22m';
+require('chalk').blue._styles[0].open = '\u001b[34m';
```
BYK: It says the following on their 2.0 release notes:
>Removed chalk.styles. Use the ansi-styles package directly instead.

BYK:
>that's odd but if I read the source code correctly, it's the right way. https://github.com/chalk/chalk/blob/master/index.js#L61

Yeah, looks like it.

>BTW, should we really have to comment how to code?

This is not "now to code" though :) You are doing something not expected, which is reaching into `chalk` internals. That said after reading the already existing comments there, I think it is fine.

I'll merge as soon as the builds pass.

KayLeung:
Hmm. The test is passed on my local Windows 10.

Could it possibly affected by the loose assumption in https://github.com/chalk/supports-color/pull/61
We might try --color to force color on the test.

BYK:
@KayLeung possibly. That said we already try to do that:
https://github.com/KayLeung/yarn/blob/0750d6444096d322bc0e27099691da44f1527796/__tests__/reporters/console-reporter.js#L14

May be that code needs updating?

BYK:
Also same here, it passes on my local Windows.

__tests__/reporters/console-reporter.js
```diff
@@ -12,8 +12,10 @@ const stream = require('stream');
 // ensures consistency across environments
 require('chalk').enabled = true;
 require('chalk').supportsColor = true;
-require('chalk').styles.blue.open = '\u001b[34m';
-require('chalk').styles.bold.close = '\u001b[22m';
+// $FlowFixMe
```
BYK: What was the error it was giving?

KayLeung: property `level`. Property not found. @see https://circleci.com/gh/yarnpkg/yarn/5295?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Just seen Yarn bundled the typed file. Not sure what's the best way to update it.

KayLeung: Also,

Chalk v2 internal use `level` to determine the value:
https://github.com/chalk/chalk/blob/master/index.js#L76

and we cannot pass options into Class constructor since that return a new instance.

`supports-color` return an object and used in Chalk
https://github.com/chalk/supports-color/blob/master/index.js#L7

Both `supportsColor` & `level` is required in Chalk v2.

src/reporters/console/console-reporter.js
```diff
@@ -28,11 +28,6 @@ const tty = require('tty');
 type Row = Array<string>;
 type InquirerResponses<K, T> = {[key: K]: Array<T>};
 
-// fixes bold on windows
-if (process.platform === 'win32' && process.env.TERM && !/^xterm/i.test(process.env.TERM)) {
-  chalk.styles.bold.close += '\u001b[m';
-}
-
```
KayLeung: We have to revert this commit I thought.

BYK: Isn't it fixed on the new version?

KayLeung: Nope?

__tests__/reporters/console-reporter.js
BYK: This file is taken from https://github.com/flowtype/flow-typed/tree/master/definitions/npm

Bundling was most probably to reduce the compiled bundle size. Since this repo doesn't have the 2.x.x version yet, there's not much we can do. I'll add a note to the ignore comment about this.

Issue #4480, opened 2017-09-15
Regression: No breadcrumbs in package warnings


spencerwilson-optimizely:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
Package warnings are not listed with 'breadcrumbs' to show the ancestry of the warning. See the example:

**0.28.4**
```
~/code/client-rum
» rm -rf node_modules/ yarn.lock && node ~/Downloads/yarn-0.28.4.js
yarn install v0.28.4
info No lockfile found.
[1/4] 🔍  Resolving packages...
warning raven > node-uuid@1.4.8: Use uuid module instead
warning mocha > glob > minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning mocha > to-iso-string@0.0.2: to-iso-string has been deprecated, use @segment/to-iso-string instead.
warning mocha > jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 8.47s.
```

**1.0.0** (regression)
```
~/code/client-rum
» rm -rf node_modules/ yarn.lock && node ~/Downloads/yarn-1.0.0.js
yarn install v1.0.0
info No lockfile found.
[1/4] 🔍  Resolving packages...
warning node-uuid@1.4.8: Use uuid module instead
warning minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning to-iso-string@0.0.2: to-iso-string has been deprecated, use @segment/to-iso-string instead.
warning jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 8.38s.
```

**1.0.2**
```
~/code/client-rum
» rm -rf node_modules/ yarn.lock && yarn
yarn install v1.0.2
info No lockfile found.
[1/4] 🔍  Resolving packages...
warning node-uuid@1.4.8: Use uuid module instead
warning minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
[...]

PR #4484, opened 2017-09-16
fix(resolver): `parentNames` is missing in package warning breadcrumbs

motnik:
**Summary**

Bugfix for #4480. Change suggested by @BYK 

**Test plan**
Running yarn on local.

_**Before Changes**_
```bash
warning node-uuid@1.4.8: Use uuid module instead
```

_**After Changes**_
```bash
warning raven > node-uuid@1.4.8: Use uuid module instead
```

### DIFF
--- a/__tests__/package-request.js
+++ b/__tests__/package-request.js
@@ -6,16 +6,25 @@ import PackageResolver from '../src/package-resolver.js';
 import Lockfile from '../src/lockfile';
 import Config from '../src/config.js';
 
-async function prepareRequest(pattern, version, resolved): Object {
+async function prepareRequest(pattern: string, version: string, resolved: string, parentRequest?: Object): Object {
   const privateDepCache = {[pattern]: {version, resolved}};
-  const lockfile = new Lockfile({cache: privateDepCache});
   const reporter = new reporters.NoopReporter({});
   const depRequestPattern = {
     pattern,
     registry: 'npm',
     hint: null,
     optional: false,
+    parentNames: [],
+    parentRequest,
   };
+  if (parentRequest) {
+    depRequestPattern.parentRequest = parentRequest;
+    depRequestPattern.parentNames = parentRequest.parentNames.slice(0);
+    depRequestPattern.parentNames.push(parentRequest.pattern);
+    const lock = parentRequest.getLocked();
+    privateDepCache[parentRequest.pattern] = {version: lock.version, resolved: lock._remote.resolved};
+  }
+  const lockfile = new Lockfile({cache: privateDepCache});
   const config = await Config.create({}, reporter);
   const resolver = new PackageResolver(config, lockfile);
   const request = new PackageRequest(depRequestPattern, resolver);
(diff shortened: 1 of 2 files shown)


motnik:
**Running Results**

_**Before Change**_
```bash
LM-BNG-26500185:test nmotiani$ yarn install --force
yarn install v1.0.2
info No lockfile found.
warning demo@1.0.0: No license field
warning node-uuid@1.4.8: Use uuid module instead
[2/4] 🚚  Fetching packages...
info koa@2.3.0: The engine "node" is incompatible with this module. Expected version ">= 6.0.0".
[4/4] 📃  Rebuilding all packages...
success Saved lockfile.
```

_**After Change**_
```bash
LM-BNG-26500185:test nmotiani$ yarn install --force
yarn install v1.0.2
info No lockfile found.
warning demo@1.0.0: No license field
warning raven > node-uuid@1.4.8: Use uuid module instead
[2/4] 🚚  Fetching packages...
info koa@2.3.0: The engine "node" is incompatible with this module. Expected version ">= 6.0.0".
[4/4] 📃  Rebuilding all packages...
success Saved lockfile.
```

BYK:
Looks good, thanks! Now we need to add tests so nobody can ever break this again!

BYK:
@motnik do you need help writing some tests for this?

motnik:
@BYK I am trying to figure out the tests structure. Would really appreciate some help in this regard.

BYK:
@motnik sorry for the late response. So here are my suggestions:

 - Either go to https://github.com/yarnpkg/yarn/blob/master/__tests__/integration.js and create a simple integration test
 - Or, go here: https://github.com/yarnpkg/yarn/blob/master/__tests__/package-request.js and add a request with a parent and then make sure the `parentNames` array is filled up properly.
 - Option c) all of the above ;)

motnik:
@BYK but these `parentNames` only come with warnings, when the packages are getting deprecated. Any ideas on how to go about that? is there a way to get a mock response like the one we get?

BYK:
>@BYK but these parentNames only come with warnings, when the packages are getting deprecated. Any ideas on how to go about that? is there a way to get a mock response like the one we get?

If you go with the second option, you should be able to create the request yourself and only check if `parentNames` is populated correctly instead of the warnings.

If you wanna test the actual warnings, then you need the first path with integration tests and set up a situation where you would get warnings. I'd got with the second option since it is simpler and more robust.

BYK:
Ping!

BYK:
@motnik need assistance? :)

motnik:
@BYK apologies for the delay. Just sent the test cases. Can you have a look ?

__tests__/package-request.js
```diff
@@ -6,16 +6,25 @@ import PackageResolver from '../src/package-resolver.js';
...
     pattern,
     registry: 'npm',
     hint: null,
     optional: false,
+    parentNames: [],
+    parentRequest,
   };
+  if (parentRequest) {
+    depRequestPattern.parentRequest = parentRequest;
+    depRequestPattern.parentNames = parentRequest.parentNames.slice(0);
```
BYK: You can use `Array.from(parentRequest.parentNames)` instead ;)

motnik: True!!!  Good Catch!! I was working on JS libraries that needed to support IE8 😢 😭 , so picked up a few habits there... Also, i think `Array.prototype.slice` is faster than `Array.from` https://jsperf.com/arrayclonetest1/1

PR #4486, opened 2017-09-16
Update: use fs.copyFile when available

### LINKED ISSUE #4331: Use `fs.copyFile`
This should be much faster on systems like macos High Sierra. Let's feature detect it and support it: https://github.com/nodejs/node/pull/15034

BYK:
**Summary**

Fixes #4331. Supersedes #3290. Uses the newly added `fs.copyFile` 
on Node 8.5 hen available and falls back to the old buffer based 
method otherwise. This patch also refactors the file copy code a 
bit making it more efficient. Here are the durations on my computer 
with [this package.json](https://github.com/sciolist/fcopy/blob/c0ded8b82bbaa3f4cdb0a67e951eb3f8b0f97fb7/perf/project/alottafiles/package.json):

| master | w/o copyFile | w/ copyFile |
| - | - | - |
| ~23s | ~19s | ~14s |

This is with `yarn.lock` in place and w/o `node_modules`.

**Test plan**

CI should pass.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -40,7 +40,10 @@ export const chmod: (path: string, mode: number | string) => Promise<void> = pro
 export const link: (src: string, dst: string) => Promise<fs.Stats> = promisify(fs.link);
 export const glob: (path: string, options?: Object) => Promise<Array<string>> = promisify(globModule);
 
-const CONCURRENT_QUEUE_ITEMS = fs.copyFile ? 16 : 4;
+// fs.copyFile uses the native file copying instructions on the system, performing much better
+// than any JS-based solution and consumes fewer resources. Repeated testing to fine tune the
+// concurrency level revealed 128 as the sweet spot on a quad-core, 16 CPU Intel system with SSD.
+const CONCURRENT_QUEUE_ITEMS = fs.copyFile ? 128 : 4;
 
 const open: (path: string, flags: string | number, mode: number) => Promise<number> = promisify(fs.open);
 const close: (fd: number) => Promise<void> = promisify(fs.close);
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -566,7 +566,7 @@ export function copy(src: string, dest: string, reporter: Reporter): Promise<voi
  * `data` contains target file attributes like mode, atime and mtime. Built-in copyFile copies these
  * automatically but our polyfill needs the do this manually, thus needs the info.
  */
-const safeCopyFile = async function(data: CopyFileAction, cleanup: Function): Promise<void> {
+const safeCopyFile = async function(data: CopyFileAction, cleanup: () => void): Promise<void> {
   try {
     await unlink(data.dest);
     await copyFile(data.src, data.dest, 0, data);
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -40,8 +40,31 @@ export const chmod: (path: string, mode: number | string) => Promise<void> = pro
 export const link: (src: string, dst: string) => Promise<fs.Stats> = promisify(fs.link);
 export const glob: (path: string, options?: Object) => Promise<Array<string>> = promisify(globModule);
 
-const CONCURRENT_QUEUE_ITEMS = 4;
+const CONCURRENT_QUEUE_ITEMS = fs.copyFile ? 16 : 4;
 
+const open: (path: string, flags: string | number, mode: number) => Promise<number> = promisify(fs.open);
+const close: (fd: number) => Promise<void> = promisify(fs.close);
+const write: (
+  fd: number,
+  buffer: Buffer,
+  offset: ?number,
+  length: ?number,
+  position: ?number,
+) => Promise<void> = promisify(fs.write);
+const futimes: (fd: number, atime: number, mtime: number) => Promise<void> = promisify(fs.futimes);
+const copyFile: (src: string, dest: string, flags: number, data: CopyFileAction) => Promise<void> = fs.copyFile
+  ? // Don't use `promisify` to avoid passing  the last, argument `data`, to the native method
+    (src, dest, flags, data) =>
+      new Promise((resolve, reject) => fs.copyFile(src, dest, flags, err => (err ? reject(err) : resolve(err))))
+  : async (src, dest, flags, data) => {
+      const fd = await open(dest, 'w', data.mode);
+      try {
+        await write(fd, await readFileBuffer(src));
+        await futimes(fd, data.atime, data.mtime);
+      } finally {
+        await close(fd);
+      }
+    };
 const fsSymlink: (target: string, path: string, type?: 'dir' | 'file' | 'junction') => Promise<void> = promisify(
   fs.symlink,
 );
@@ -61,7 +84,6 @@ export type CopyQueueItem = {
 type CopyQueue = Array<CopyQueueItem>;
 
 type CopyFileAction = {
-  type: 'file',
   src: string,
   dest: string,
   atime: number,
@@ -70,19 +92,21 @@ type CopyFileAction = {
 };
 
 type LinkFileAction = {
-  type: 'link',
   src: string,
   dest: string,
   removeDest: boolean,
 };
 
 type CopySymlinkAction = {
-  type: 'symlink',
   dest: string,
   linkname: string,
 };
 
-type CopyActions = Array<CopyFileAction | CopySymlinkAction | LinkFileAction>;
+type CopyActions = {
+  file: Array<CopyFileAction>,
+  symlink: Array<CopySymlinkAction>,
+  link: Array<LinkFileAction>,
+};
 
 type CopyOptions = {
   onProgress: (dest: string) => void,
@@ -154,7 +178,11 @@ async function buildActionsForCopy(
   events.onStart(queue.length);
 
   // start building actions
-  const actions: CopyActions = [];
+  const actions: CopyActions = {
+    file: [],
+    symlink: [],
+    link: [],
+  };
 
   // custom concurrency logic as we're always executing stacks of CONCURRENT_QUEUE_ITEMS queue items
   // at a time due to the requirement to push items onto the queue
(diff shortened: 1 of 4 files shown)


src/util/fs.js
```diff
@@ -527,6 +552,15 @@ export function copy(src: string, dest: string, reporter: Reporter): Promise<voi
   return copyBulk([{src, dest}], reporter);
 }
 
+const copyFileWithAttributes = async function(data: CopyFileAction, cleanup: Function): Promise<void> {
```
BYK: This is a terrible name, I should probably change this to `safeCopyFile` with an explanation of `unlink` and `cleanup` parts. What do you think?

src/util/fs.js
```diff
@@ -40,8 +40,31 @@ export const chmod: (path: string, mode: number | string) => Promise<void> = pro
...
+) => Promise<void> = promisify(fs.write);
+const futimes: (fd: number, atime: number, mtime: number) => Promise<void> = promisify(fs.futimes);
+const copyFile: (src: string, dest: string, flags: number, data: CopyFileAction) => Promise<void> = fs.copyFile
+  ? // Don't use `promisify` to avoid passing  the last, argument `data`, to the native method
+    (src, dest, flags, data) =>
+      new Promise((resolve, reject) => fs.copyFile(src, dest, flags, err => (err ? reject(err) : resolve(err))))
+  : async (src, dest, flags, data) => {
+      const fd = await open(dest, 'w', data.mode);
+      try {
+        await write(fd, await readFileBuffer(src));
```
arcanis: Why not writeFile?

src/util/fs.js
arcanis: The name looks fine, but a comment to explain the difference with copyFile would be useful.

I also feel like the arguments are also kinda obfuscated because of the custom CopyFileAction type.

src/util/fs.js
```diff
@@ -547,57 +581,31 @@ export async function copyBulk(
...
+  const currentlyWriting: Map<string, Promise<void>> = new Map();
 
   await promise.queue(
     fileActions,
-    async (data): Promise<void> => {
-      let writePromise: Promise<void>;
-      while ((writePromise = currentlyWriting[data.dest])) {
+    async (data: CopyFileAction): Promise<void> => {
+      const writePromise = currentlyWriting.get(data.dest);
+      if (writePromise) {
```
arcanis: You can't replace the while by a if, otherwise multiple waiting instances will all get released all at once. The while ensures that a single one will ever be able to cross the condition.

src/util/fs.js
BYK: `writeFile` and `utimes` expect file paths and use `open()` and `close()` behind the scenes. Using `open`, `write` and `close` manually here along with `futimes` which expects a file descriptor, makes the operation more efficient by avoiding an extra file open/close.

I think this is why the patch is faster than master even when `fs.copyFile` is not used.

Should I add a comment about this in the code?

src/util/fs.js
BYK: I thought about this but here's my reasoning:
 - Event loop is single-threaded
 - If I do `await promise` the rest of the code is essentially gets into `promise.then()`
 - Since all code after this point is synchronous and since the event loop is single threaded, there can't be a new entry in the map between when the promise get's released and we add a new promise there.

Makes sense, or am I missing something?

arcanis: Even if the event loop is single threaded, because all iterations will wait on the same promise, all the promises will be released at once (but sequentially) because they won't re-check to make sure they are the first promise to wake up. 

That being said, I just saw that this call is inside a promise.queue call, so maybe this utility will already take care of this (but then why await at all?).

BYK: Ah, I see! Yes, you're right. For the promise queue part, I don't know why we simply don't return the existing promise. Doesn't make much sense to me since right now we are essentially writing the same file more than 1 time once the promsie is released. I'll change the code accordingly.

BYK:
Oh boy, that speed:

src/util/fs.js
```diff
@@ -40,8 +40,34 @@ export const chmod: (path: string, mode: number | string) => Promise<void> = pro
 export const link: (src: string, dst: string) => Promise<fs.Stats> = promisify(fs.link);
 export const glob: (path: string, options?: Object) => Promise<Array<string>> = promisify(globModule);
 
-const CONCURRENT_QUEUE_ITEMS = 4;
-
+const CONCURRENT_QUEUE_ITEMS = fs.copyFile ? 16 : 4;
```
cpojer: How did you pick these numbers?

src/util/fs.js
```diff
@@ -547,57 +592,31 @@ export async function copyBulk(
...
-          },
-          err => {
-            cleanup();
-            throw err;
-          },
-        ));
+      const copier = safeCopyFile(data, () => currentlyWriting.delete(data.dest));
+      currentlyWriting.set(data.dest, copier);
+      events.onProgress(data.dest);
+      return copier;
```
cpojer: I read: cpojer. lol

cpojer:
Wow, awesome speedup for Yarn!

src/util/fs.js
BYK: I borrowed them from #3290. I did some testing myself with various numbers going as high as 64 and as low as 1. `1` definitely hurts but anything more than 8 doesn't make a huge difference. I think coupling this with the CPU count would be a better idea. What do you think?

cpojer: How about an elaborate code comment on how you came up with this truth? Numbers tend to be arbitrary in code unless we give explicit meaning to them.

BYK: I'll experiment a bit more and then see if I can come up with a better logic. I think simply tying this tp process count should be fine and self-documenting. Otherwise, I'll add the comment as you suggested.

langpavel:
Does node support shallow copy for copy on write filesystems like btrfs? https://btrfs.wiki.kernel.org/index.php/Deduplication
It can not only speed up operation but also save disk space

BYK:
@langpavel I think the new `fs.copyFile` will be able to leverage that, yeah.

src/util/fs.js
BYK: Glad you brought this up @cpojer! I did some repeated testing for powers of 2 and found that 128 is consistently the best concurrency level. Updated the code and added a comment about this

BYK:
Another record after setting concurrency to 128 on Linux & Node 8: TravisCI test run time

vjeux:
I'm curious, for node < 8.5 could you shell out to "cp" and if you do, do you get the same wins?

BYK:
>I'm curious, for node < 8.5 could you shell out to "cp" and if you do, do you get the same wins?

Why you hate cross-platform? :( (Windows)

BYK:
@vjeux also "shelling out" would probably come with its own costs which may make this option a lot less desirable.

sciolist:
I did give shelling out to cp a look and you could get similar speed benefits, but you'd pretty pretty much need to prepare one massive bash (+ powershell for windows) script with all the copies you want to run and then execute that, to offset the shelling out cost.. especially on Windows where it's very, very slow. Terribly ugly.

Daniel15:
@langpavel

> Does node support shallow copy for copy on write filesystems like btrfs? 

Not yet, but coming soon! See https://github.com/libuv/libuv/pull/1491 and the discussion on https://github.com/libuv/libuv/pull/1465

src/util/fs.js
```diff
@@ -527,6 +558,23 @@ export function copy(src: string, dest: string, reporter: Reporter): Promise<voi
...
 
+/**
+ * Unlinks the destination to force a recreation. This is needed on case-insensitive file systems
+ * to force the correct naming when the filename has changed only in charater-casing. (Jest -> jest).
+ * It also calls a cleanup function once it is done.
+ *
+ * `data` contains target file attributes like mode, atime and mtime. Built-in copyFile copies these
+ * automatically but our polyfill needs the do this manually, thus needs the info.
+ */
+const safeCopyFile = async function(data: CopyFileAction, cleanup: Function): Promise<void> {
```
Daniel15: Use `() => void` instead of  `Function` for the type of `cleanup`.

src/util/fs.js
```diff
@@ -40,8 +40,37 @@ export const chmod: (path: string, mode: number | string) => Promise<void> = pro
 export const link: (src: string, dst: string) => Promise<fs.Stats> = promisify(fs.link);
 export const glob: (path: string, options?: Object) => Promise<Array<string>> = promisify(globModule);
 
-const CONCURRENT_QUEUE_ITEMS = 4;
-
+// fs.copyFile uses the native file copying instructions on the system, performing much better
+// than any JS-based solution and consumes fewer resources. Repeated testing to fine tune the
+// concurrency level revealed 128 as the sweet spot on a quad-core, 16 CPU Intel system with SSD.
+const CONCURRENT_QUEUE_ITEMS = fs.copyFile ? 128 : 4;
```
Daniel15: I wonder if 128 will essentially lock up the system on slower hard drives (particularly if the user is multitasking)

BYK: It shouldn't since this is still single threaded and we leave the scheduling job to the OS and Node as far as I understand. The reason for having a much lower limit for the fallback is mostly old versions of Node having buffering issues with `pipe` sometimes and possible memory pressure after #3539 since it now reads the whole file contents into memory.

BYK: If this becomes a problem we can always make this a configurable variable.

cpojer:
🎉

bestander:
john-wesley-shipp-the-flash

langpavel:
For CoW file systems support see #5456

droslean:
@BYK I found this PR while trying to understand the memory usage of yarn. Is there any way to disable this feature and not use any cache at all? Instead, I would assume yarn to download the packages directly to the node_modules directory instead of using the default cache folder and then copy-paste millions of files.

PR #3779, opened 2017-06-30
Add a failing test for version conflicts

magicmark:
**Summary**
This is a failing test to highlight the issue the issue here https://github.com/yarnpkg/yarn/issues/3780 https://github.com/mxmul/yarn_flat_regression

tl;dr:

```
a => b => left-pad@>=0.0.1
a => left-pad@1.0.0
```

yarn fails when doing `--flat` with the following:

```
info Unable to find a suitable version for "left-pad", please choose one by typing one of the numbers below:
  1) "left-pad@>=0.0.1" which resolved to "1.1.3"
  2) "left-pad@1.0.0" which resolved to "1.0.0"
```

(yarn version: 0.27.1)

(this seems to be a regression introduced in 0.27.0, this worked in 0.26.1)

**Test plan**
This is a test only change.

---

This test is currently failing because it's prompting for a resolution, with the following stacktrace
```
(node:23637) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 2)
 FAIL  __tests__/commands/install/integration.js (264.799s)
  ● unbound transitive dependencies should not conflict with top level dependency

    Error: Can't answer a question unless a user TTY
      
      at ConsoleReporter.select (src/reporters/console/console-reporter.js:322:29)
      at src/cli/commands/install.js:932:182
      at next (native)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:35:14
      at Promise.F (node_modules/core-js/library/modules/_export.js:35:28)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:14:12
      at Install.flatten (src/cli/commands/install.js:933:2215)
      at src/cli/commands/install.js:922:3570
      at next (native)
      Console output:
       warning package.json: No license field
      warning No license field
      [1/4] Resolving packages...
      
      at __tests__/commands/_helpers.js:144:13
      at throw (native)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:30:13
```

PR #4488, opened 2017-09-17
Fix: 'yarn --flat' unbound transitive deps shouldn't conflict with top level deps

### LINKED ISSUE #3780: yarn --flat sees a version conflict where there is none
bug

`yarn --flat` sees a version conflict where there is none (in >=0.27.0).

If I have package A that requires 'left-pad@>=1.0.1" and package B that requires both A and "left-pad@1.1.1", and install package B with `yarn --flat` there is a conflict.

https://github.com/mxmul/yarn_flat_regression

yarn would choose the one version that satisfies the version constraints (1.1.1)

`yarn@0.27.1` `node@5.11.0` `ubuntu@10.04`

mxmul:
**Summary**

Fixes #3780, and makes the failing test from #3779 passing.

As a final step of package resolution, for each dependency we check whether any version satisfies all resolved version ranges. 

**Test plan**

Fixes an existing (failing) test: "unbound transitive dependencies should not conflict with top level dependency"

### DIFF
--- a/src/package-resolver.js
+++ b/src/package-resolver.js
@@ -228,6 +228,14 @@ export default class PackageResolver {
 
   getAllInfoForPackageName(name: string): Array<Manifest> {
     const patterns = this.patternsByPackage[name] || [];
+    return this.getAllInfoForPatterns(patterns);
+  }
+
+  /**
+   * Retrieve all the package info stored for a list of patterns.
+   */
+
+  getAllInfoForPatterns(patterns: string[]): Array<Manifest> {
     const infos = [];
     const seen = new Set();
 
@@ -284,6 +292,13 @@ export default class PackageResolver {
 
   collapseAllVersionsOfPackage(name: string, version: string): string {
     const patterns = this.dedupePatterns(this.patternsByPackage[name]);
+    return this.collapsePackageVersions(name, version, patterns);
+  }
+
+  /**
+   * Make all given patterns resolve to version.
+   */
+  collapsePackageVersions(name: string, version: string, patterns: string[]): string {
     const human = `${name}@${version}`;
 
     // get manifest that matches the version we're collapsing too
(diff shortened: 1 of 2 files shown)


src/package-resolver.js
```diff
@@ -530,10 +530,48 @@ export default class PackageResolver {
...
 
+  // for a given package, see if a single manifest can satisfy all ranges
+  optimizeResolutions(name: string) {
+    const patterns: Array<string> = this.patternsByPackage[name] || [];
+
+    // don't optimize things that already have a lockfile entry:
+    // https://github.com/yarnpkg/yarn/issues/79
+    const anyLocked = patterns.some(pattern => this.lockfile.getLocked(pattern));
+    if (anyLocked) {
+      return;
```
BYK: Why not exclude only the locked ones and still optimize the rest?

src/package-resolver.js
```diff
@@ -530,10 +530,48 @@ export default class PackageResolver {
...
+    }
+
+    const manifestByVersion: Map<string, Manifest> = new Map();
```
BYK: Looks like you just need `this.getAllInfoForPackageName(name)` above instead.

src/package-resolver.js
```diff
@@ -530,10 +530,48 @@ export default class PackageResolver {
...
+    patterns.forEach(p => {
+      const manifest = this.patterns[p];
+      manifestByVersion.set(manifest.version, manifest);
+    });
+
+    // reverse sort, so we'll find the maximum satisfying version
+    const availableVersions = [...manifestByVersion.keys()];
```
BYK: Why not `Array.from(manifestByVersion.keys())`. That said spread looks is faster in latest V8 versions so NVM: https://jsperf.com/set-iterator-vs-foreach/4

src/package-resolver.js
```diff
@@ -530,10 +530,48 @@ export default class PackageResolver {
...
+    availableVersions.sort(semver.rcompare);
+
+    const requiredRanges = patterns.map(p => normalizePattern(p).range);
```
BYK: I think you can leverage some semver secrets as follows:
```js
const fullRange = patterns.map(pattern => normalizePattern(p).range).join(' ');
const oneRingToRuleThemAll = semver.maxSatisfying([...manifestByVersion.keys()], fullRange);
if (oneRingToRuleThemAll) {
    this.collapseAllVersionsOfPackage(name, oneRingToRuleThemAll);
}
```

This should be more efficient than what the code below does. Also, the code below continuously resolves to new versions as it goes on since `return` in a `forEach` loop doesn't terminate the loop at all.

mxmul: Yep, that's a nice optimization.

src/package-resolver.js
mxmul: So I've made this change, but it causes some workspace related tests to fail ("install should link workspaces that refer each other", "install should not link workspaces that refer not compatible version of another workspace"). I don't quite understand why these are failing, but I've just disabled the optimization wherever `remote.type === "workspace"`. Open to other suggestions.

PR #4495, opened 2017-09-18
Fix: make lockfile parser tolerant to CRLF

BYK:
**Summary**

Lockfile parser fails when the lockfile had Windows-style line
endings with an "Invalid token" error. This is default on Windows
and, say when editing the file using vim, it is not obvious or
trivial that the file endings are in Windows-style. This patch
makes the parser tolareate potential "\r" chars before "\n" while
keeping the existing parsing logic the same.

**Test plan**

Added three new unit tests.

### DIFF
--- a/src/lockfile/parse.js
+++ b/src/lockfile/parse.js
@@ -1,12 +1,13 @@
 /* @flow */
 /* eslint quotes: 0 */
 
+import util from 'util';
+import invariant from 'invariant';
+import stripBOM from 'strip-bom';
+
 import {LOCKFILE_VERSION} from '../constants.js';
 import {MessageError} from '../errors.js';
 import map from '../util/map.js';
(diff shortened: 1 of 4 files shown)


src/lockfile/parse.js
```diff
@@ -155,7 +159,7 @@ function* tokenise(input: string): Iterator<Token> {
     }
 
     col += chop;
-    lastNewline = input[0] === '\n';
+    lastNewline = input[0] === '\n' || input[1] === '\n';
```
bestander: does input[0] matter when input[1] === '\n'?

BYK: I guess I can change this to `input[0] === '\n' || (input[0] === '\r' && input[1] === '\n')` to be precise?

Issue #4470, opened 2017-09-15
Confusing order of CLI flags


zemlanin:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
I've got this script in `package.json`:

```
    "build": "webpack --config a.config.js --config b.config.js",
```

In v0.27.5 (and prior) command `yarn build --no-progress --prod` called `webpack --config a.config.js --config b.config.js` with production NODE_ENV

Now, after update to `yarn@1.0.2`, the same `yarn build --no-progress --prod` calls `webpack --config a.config.js --config b.config.js "--no-progress" "--prod"`, passing through `--no-progress --prod` to webpack

I've tried to rearrange arguments in `yarn build --no-progress --prod` so that it worked as previously, but:

- `yarn run build --no-progress --prod` still calls `webpack --config a.config.js --config b.config.js "--no-progress" "--prod"`
- `yarn run --no-progress --prod build` _(maybe I needed set yarn flags _before_ script name)_ results in `error Command "--prod" not found.`
  - (flags reordering only changes "not found" command name)
- `yarn --no-progress --prod run build` results in `error No command specified.`
- and only `yarn --prod --no-progress run build` (alongside with `yarn --prod --no-progress build`) worked as expected

**If the current behavior is a bug, please provide the steps to reproduce.**

https://gist.github.com/zemlanin/67f3d1db3dc39997bf9aa164a358eb1b

```
$ git clone https://gist.github.com/67f3d1db3dc39997bf9aa164a358eb1b.git yarn-flags-order

$ cd yarn-flags-order

$ yarn echo --no-progress --prod
  yarn echo v1.0.2
  $ echo $NODE_ENV "--no-progress" "--prod"
  --no-progress --prod
  ✨  Done in 0.13s.

$ yarn --no-progress --prod echo
  yarn echo v1.0.2
  error No command specified.
  info Commands available from binary scripts: 
  info Project commands
     - echo
        echo $NODE_ENV
  ^C
  error No command specified.
  ✨  Done in 2.98s.

$ yarn --prod --no-progress echo
  yarn echo v1.0.2
  $ echo $NODE_ENV
  production
  ✨  Done in 0.12s.
```

**What is the expected behavior?**
- flags before double-dash (`--`) do not pass to the script
- order of the flags should be important (at least in _this_ case)

**Please mention your node.js, yarn and operating system version.**
```bash
$ node -v
v8.5.0
$ yarn -v
1.0.2
[...]

PR #4498, opened 2017-09-18
Fix: fix argument parsing and .yarrc argument passing

### LINKED ISSUE #4457: Emoji disabled

bug
The only way to see emoji is to explicitly pass `--emoji` to yarn. It would seem the line `emoji true` in .yarnrc is ignored (as are variations like `--emoji`, `no-emoji false`, etc).
```
rasmus ⇒ yarn config list        
yarn config v1.0.1
info yarn config
{ 'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'strict-ssl': true,
  'user-agent': 'yarn/1.0.1 npm/? node/v8.4.0 linux x64',
  emoji: true,
  lastUpdateCheck: 1505388775752,
  prefix: '/home/rasmus/.nvm/versions/node/v8.4.0' }
info npm config
{}

```

Emoji would be displayed since they are turned on.

yarn 1.0.1
node 8.4.1
ubuntu 16.04

BYK:
**Summary**
Fixes #4457, partially handles #4470. `.yarnrc` argument passing logic was not aligned with
how `commander` works. Even more, `commander`'s boolean logic was
also not used correctly. This patch fixes both, allowing use of
boolean type CLI options both ways (enable/disable) both from the
command line and `.yarnrc` file.

**Test plan**

Added new integration test for getting args from `.yarnrc`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ matrix:
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      RAMDISK_SIZE=320;
+      RAMDISK_SIZE=330;
       RAMDISK_SECTORS=$(( $RAMDISK_SIZE * 1024 * 1024 / 512 ));
       RAMDISK=$(hdiutil attach -nomount ram://$RAMDISK_SECTORS);
       newfs_hfs -v yarn_ramfs $RAMDISK;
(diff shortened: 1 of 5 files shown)


BYK:
>Commander also support --no-<flag>.

This also supports it.

>So yarn --prod 0 should be the same as yarn --no-prod.

I dare you to try it. I thought the same and no, it doesn't work that way. You need to explicitly define `--no-prod` in options for this to work and then `--prod` stops working but you get the value via `commander.prod`. I'm just sad that I spent hours to discover this. :cry:

Issue #4284, opened 2017-08-30
"yarn add package --dev" installs globally instead of locally


arhyth:
**Do you want to request a *feature* or report a *bug*?**
reporting a bug. 

**What is the current behavior?**
`yarn add package --dev` installs packages on my global cache directory

**If the current behavior is a bug, please provide the steps to reproduce.**
no special steps. just the routine `yarn add package --dev` call.  with the only difference being that my yarn msi installation is in a different directory because of sandboxing by my antivirus.

**What is the expected behavior?**
`yarn add package --dev` to install on directory it is called on the commandline.

**Please mention your node.js, yarn and operating system version.**
Yarn 0.27.5
Node 6.11.2
Windows 7 x64 Ultimate

PR #4502, opened 2017-09-19
Update: warn when used with not supported Node versions

### LINKED ISSUE #4501: Make the node requirements clear

feature

yarn will fail on unsupported versions of node

spit out some warning before failing

< ^4.8.0 and <= 5.7.0

see also https://github.com/yarnpkg/website/issues/646 for more information

BYK:
**Summary**

Fixes #4501, refs #4490, refs #4284. Yarn now warns when it
detects it is running in a Node version that is not fully
supported and warns the user about this. This is different than
the hard Node 4+ check in the entry file since in that case,
Yarn wouldn't run at all due to syntax incompatibilities. This
warning is to signal that users may encounter unexpected errors
but are allowed to use Yarn if they wish. It also adds a new
flag to suppress this warning: `--no-node-version-check`.

**Test plan**

Since we cannot add unsupported Node versions to our CI and
spoof the Node version internally, this has to be tested
manually, which I did.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -1,5 +1,17 @@
 /* @flow */
 
+import http from 'http';
+import net from 'net';
+import path from 'path';
+
+import commander from 'commander';
+import fs from 'fs';
+import invariant from 'invariant';
+import lockfile from 'proper-lockfile';
+import loudRejection from 'loud-rejection';
+import onDeath from 'death';
+import semver from 'semver';
+
 import {ConsoleReporter, JSONReporter} from '../reporters/index.js';
 import {registries, registryNames} from '../registries/index.js';
 import commands from './commands/index.js';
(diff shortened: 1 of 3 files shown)


SimenB:
You can mock the node version if you really want to.

https://github.com/SimenB/node-version-check/blob/d761da1b4fa2bac2845b4ba31c525817f4d31755/test/lt8.test.js#L4-L26

BYK:
@SimenB yes but that requires us to run Yarn from inside the same process that mocks the version which is not easy or may not be possible with the CLI entry point.

src/cli/index.js
```diff
@@ -98,6 +100,7 @@ export function main({
     'prepend the node executable dir to the PATH in scripts',
     boolify,
   );
+  commander.option('--no-node-version-check', 'do not warn when using a potentially unsupported Node version');
```
arcanis: That's a long command name :D

Haroenv: `--no-node-version-check-or-else-youll-be-fired`

mkj28:
Trying to understand why was node 7 dropped from support? Cannot find the reason in the documents linked here.

BYK:
@mkj28 I think I dropped it unintentionally since it is not listed anywhere and it is not LTS. We do not test anything with Node 7 so we do not commit to supporting it. This doesn't mean it is going to break.

PR #4519, opened 2017-09-22
compare current version against latest and not wanted version in outdated command

danez:
**Summary**
`current` and `wanted` version might be the same but `latest` is a new major version
and as current and wanted are compared against each other it results in most outdated entries being white instead of the proper color

Here a before/after screenshot:

bildschirmfoto 2017-09-22 um 13 41 56


BYK:
I don't have a strong opinion on this but I'd like to verify this behavior with @yarnpkg/core. Anyone willing to approve this patch?

@danez it would be great if you could add some tests for this btw.

Haroenv:
What about adding a colour key on the bottom here, it might be hard to tell those apart otherwise

kaylieEB:
Thanks @danez! The code change looks fine, but I agree with @Haroenv that it's not as intuitive that the color changes between major vs minor versions.

danez:
In this case it is also weird because all the changes in the screenshot are considered "major" (0.1.0 -> 0.2.0 or 1.0.0 > 2.0.0), though they still have different colors.

I guess the difference in the color would be:

 - red: You need to run `yarn upgrade pkg-name --latest` to update it and it is potentially breaking (according to semver)
 - yellow: It will be updated when running `yarn run upgrade` and it is potentially not breaking (according to semver)

Does that make sense?

For the tests I could mock the formatter and check which color is used for which dep. Does that sound okay?

danez:
I also just noticed that `upgrade-interactive` uses the same coloring scheme that `outdated` uses which seems confusing to me as flow is not in my range, similar to all other dependencies. So the most understandable difference between the colors for me is about is the upgrade in my current range (yellow) or not (red).

bildschirmfoto 2017-09-25 um 17 38 53

kaylieEB:
> I also noticed that in the above case it is also weird because all the changes in the screenshot are considered "major/breaking" (0.1.0 -> 0.2.0 or 1.0.0 > 2.0.0), though they still have different colors.

(0.1.0 -> 0.2.0) is not considered major/breaking, since it's a minor bump and should be backwards compatible

> I guess the difference in the color should be:
> 
> red
> You need to run yarn upgrade pkg-name --latest to update the package
> Potentially breaking (according to semver)
> yellow
> If ranges are used in pkg.json, the package will be updated when running yarn run upgrade
> Potentially not breaking (according to semver)

I think it's confusing to have a color mean two things (how to upgrade and potentially breaking / non-breaking--the latter especially feels inconclusive). Ideally it should use the same key as upgrade-interactive
```
 "<red>"    : Major Update backward-incompatible updates
 "<yellow>" : Minor Update backward-compatible features
 "<green>"  : Patch Update backward-compatible bug fixes
```

> For the tests I could mock the formatter and check which color is used for which dep. Does that sound okay?

Sounds good to me!

> I also just noticed that upgrade-interactive uses the same coloring scheme that outdated uses which seems confusing to me. 

I'm not sure I follow. The colors seem to correctly reflect the key in upgrade-interactive (major bumps being in red for backwards-incompatible change). The difference is that `flow-bin` is a minor bump, whereas the rest are major bumps.

danez:
@kaylieEB thanks for the clarification, seems I mixed some things up. I always thought that when the major version is 0 that semver defines that the minor version gets promoted to major, but that is not the case. The specification only says:
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

It seems i confused semver with the way how the caret operator works, but that is not part of the specification.

Will try to finish this today.

kaylieEB:
@danez ping! need any help?

danez:
I think it is done now, I also added the legend.

src/reporters/lang/en.js
```diff
@@ -114,6 +114,8 @@ const messages = {
   allDependenciesUpToDate: 'All of your dependencies are up to date.',
   legendColorsForUpgradeInteractive:
     'Color legend : \n $0    : Major Update backward-incompatible updates \n $1 : Minor Update backward-compatible features \n $2  : Patch Update backward-compatible bug fixes',
+  legendColorsForOutdated:
```
danez: This is the same as `legendColorsForUpgradeInteractive` should I merge them? If yes what to name them?

kaylieEB: yeah I'd prefer merging them. How about `legendColorsForVersionUpdates`?

danez: Okay changed :)

kaylieEB:
Thanks so much for your help @danez ! I'll merge once CI passes :)

danez:
Thank you :)

PR #4553, opened 2017-09-27
Ensure publish hooks are run at the same time.

### LINKED ISSUE #4515: Yarn publish is publishing state of package PRIOR to the prepublish script
This is not a feature request.  I believe it is a bug.

I have the following scripts in my package.json file:

> "scripts": {
>   "build": "tsc",
>   "prepublish": "yarn build"
> },

It appears that if I invoke "yarn publish" from the command line, the code that is published is the code existing in the package PRIOR to typescript compilation (tsc).  In other words, if I execute "yarn build" on the command line and let it complete, and then execute "yarn publish", the correct output is published.  But if I make a change to the underlying typescript and execute "yarn publish", I will see screen output indicating that yarn build is being called, and that tsc is being called, but the files that actually get published are the ones in the repo BEFORE the yarn build is called.

I would expect that explicitly executing "yarn build" followed by "yarn publish" should (with the scripts above) yield exactly the same published results as if I only explicitly executed "yarn publish".

Node: 6.10.3
Yarn: 1.0.2
OS: Windows 8.1 Enterprise

johnf:
Move publish, prepublish and prepare hooks to run before tar ball is
created.

Also switch the order of prepare to be consistent with
https://docs.npmjs.com/misc/scripts

**Summary**

Currently publish hooks are being run after tarball creation, this means that any output of this process e.g. processing by babel isn't included in the published artifacts.

Fixes #4515

**Test plan**

Create following package.json
``` json
{
  "name": "hubber-iot",
  "scripts": {
    "build": "echo foo > bar.txt",
    "prepublish": "yarn run build",
  },
  "version": "1.0.0"
}
```

Then ```yarn publish``` bar.txt should be included in the module


BYK:
LGTM. Deferring to @arcanis and @kaylieEB for approval and merge.

PS: I merged master into your branch which should clear the CI failure.

PR #4572, opened 2017-09-28
Fix: always prefer CLI options over env for --prod

### LINKED ISSUE #4557: --production=false flag not being honored on install
Bug

When passing `--production=false` with `NODE_ENV=production` devDependencies are not installed. Prior to v1.1.0 this worked correctly. 

`NODE_ENV=production yarn install --production=false`

devDependencies are installed despite `NODE_ENV` being set to `production`

Node: 6.10.3
Yarn: 1.1.0
OS: MacOS 10.12.6

BYK:
**Summary**

Fixes #4557. Also adds `yarn config current` that lists the current
configuration as JSON for testing purposes.

**Test plan**

Added integration tests.

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -5,6 +5,41 @@ import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import buildSubCommands from './_build-sub-commands.js';
 
+const CONFIG_KEYS = [
+  // 'reporter',
+  'rootModuleFolders',
+  'registryFolders',
+  'linkedModules',
+  // 'registries',
+  'cache',
+  'cwd',
+  'looseSemver',
+  'commandName',
+  'preferOffline',
+  'modulesFolder',
+  'globalFolder',
+  'linkFolder',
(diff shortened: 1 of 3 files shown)


src/cli/commands/config.js
```diff
@@ -5,6 +5,41 @@ import type {Reporter} from '../../reporters/index.js';
...
+  'networkTimeout',
+  'workspacesEnabled',
+  'pruneOfflineMirror',
+  'enableMetaFolder',
+  'enableLockfileVersions',
+  'linkFileDependencies',
+  'cacheFolder',
+  'tempFolder',
+  'production',
+];
```
arcanis: Couldn't we just iterate over `config`, and possibly blacklist fields if there's internal ones? A whitelist seems a burden to maintain.

BYK: We can but isn't it safer to have a fixed list? It also makes our job easier by simply using `JSON.stringify` and passing the whitelist. For a blacklist, we need to manually walk over the config object beforehand and copy all to a new object.

PR #4583, opened 2017-09-29
Fix commands when used in a workspace

### LINKED ISSUE #4328: yarn outdated not working w/workspaces

Bug

In a project work workspaces, `yarn outdated` claims everything is up to date. (The root project has no dependences). `yarn workspace api outdated` errors with the following:

```
yarn workspace v1.0.0
yarn outdated v1.0.0
error Command failed.
Exit code: 1
Command: /usr/local/Cellar/node/8.4.0/bin/node
Arguments: /usr/local/Cellar/yarn/1.0.0_1/libexec/lib/cli.js outdated
Directory: /Volumes/git/project/packages/api
Output:

```

I expect it to tell me the outdated modules.

yarn 1.0.0

### LINKED ISSUE #4334: "yarn remove" inside workspace fails

Attempting to remove a package from inside a workspace fails with:

```
error No lockfile in this directory. Run `yarn install` to generate one
```

```sh
$ cd packages/some-package
$ yarn remove some-dependency
yarn remove v1.0.1
error No lockfile in this directory. Run `yarn install` to generate one
```

The package should be removed and the root lockfile should be updated.

Node 8.4.0
Yarn 1.0.1
Mac OS 10.12

Possibly related to #4328.

### LINKED ISSUE #4348: Adding root dependency removes workspace packages from lockfile

In a workspace project:

```sh
```

**Workaround**

```sh
$ yarn install
```

Node: 8.4.0
Yarn: 1.0.1
OS: Mac OS 10.12

jgoz:
**Summary**
Fixes #4334 
Fixes #4328 
Fixes #4348
Closes #4584

Previously, `remove`, `outdated` and several related commands would fail when used inside of a workspace package, either directly or via `yarn workspaces <pkg> <cmd> [args]`. These failures were caused by incorrect lockfile resolution, specifically using `config.cwd` rather than `config.lockfileFolder`.

Also fixed:
- When using `remove` inside a workspace, `reinstall` would not clean up extraneous packages.
- When using `outdated`, `upgrade` or `upgrade-interactive` inside of a workspace root or child, the wrong dependencies were often listed and the lockfile was being saved incorrectly after completing an upgrade
- Adding a dev dependency to the workspace root would remove all workspace child dependencies from the lockfile (this PR replaces #4584)

**Test plan**

- Added relevant unit tests
- Existing tests pass
- Manually tested against [repro project](https://github.com/jgoz/yarn-workspace-issue)


src/package-linker.js
```diff
@@ -246,29 +246,15 @@ export default class PackageLinker {
...
-              possibleExtraneous.add(path.join(filepath, subfile));
-            }
-          } else {
-            possibleExtraneous.add(filepath);
-          }
-        }
-      }
+    // look for extraneous packages in workspaces too
+    if (this.config.lockfileFolder !== this.config.cwd) {
+      await this._readRegistryFolders(this.config.cwd, possibleExtraneous, scopedPaths);
```
arcanis: Shouldn't this be a `while` loop that would iterate on each workspace?

jgoz: Yes, I think that would make sense. Do you see any value including `cwd` in addition to `lockfileFolder` and all of the workspaces?

arcanis: I don't think it's necessary, all the node_modules should be guaranteed to be in these folders.

jgoz:
@arcanis Please note that the scope of this PR increased somewhat as I continued testing commands that could not be executed previously. I have updated the description accordingly.

In order to properly test the now-available commands, my other fix from #4584 had to be included here. @kaylieEB, I would greatly appreciate your insight on this PR because it deviates from the approach you took in #4278.

arcanis:
Closing because of #4630

PR #4584, opened 2017-09-29
Preserve workspace packages when adding to the root

### LINKED ISSUE #4348: Adding root dependency removes workspace packages from lockfile

jgoz:
**Summary**

Fixes #4348

Before this, adding a dev dependency to the workspace root would clear out workspace dependencies from `yarn.lock`. This is because `ignoreWorkspaces` was being set to `true` when executing `add` in the workspace root.

**Test plan**

- Added new test
- Existing tests pass
- Manually tested in [repro project](https://github.com/jgoz/yarn-workspace-issue)

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -54,6 +54,24 @@ test.concurrent("add with --dev shouldn't fail on the workspace root", async () 
   });
 });
 
+test.concurrent('adding to the workspace root should preserve workspace packages in lockfile', async () => {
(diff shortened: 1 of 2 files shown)


arcanis:
@kaylieEB would you mind reviewing this PR? I think it goes against what you did in #4278, I think you'll be more capable than us to make sure this commit doesn't make regression on the previous bug 😃

jgoz:
This is superseded by #4583.

Issue #2636, opened 2017-02-05
bundledDependencies don't get packed as of >= v0.19.0


selkhateeb:
**Do you want to request a *feature* or report a *bug*?**
*BUG*

**What is the current behavior?**
`yarn pack` does not package bundledDependecies in v0.19.0, v0.19.1 and v0.20.0

**If the current behavior is a bug, please provide the steps to reproduce.**
```
{
  "name": "test-bundled-deps",
  "version": "0.0.0",
  "dependencies": {
    "bcryptjs": "^2.0.0",
  },
  "bundledDependencies": [
    "bcryptjs",
  ]
}

```
then run `yarn` followed by `yarn pack`
check the tgz file. It will not contain the node_modules folder.

**What is the expected behavior?**
to work as in version 0.18.1
**Please mention your node.js, yarn and operating system version.**
```
$ node -v
v7.4.0
```
OS X El Capitan 10.11.6

kmdm:
Would seem to at least be caused by:-

 * 'node_modules' being added to FOLDERS_IGNORE d1d5343082cd5a3b2020962c0bb6f2d6fb1ac4f3
 *  Accessing of bundledDependencies incorrect in commands/pack.js
  - src/cli/commands/pack.js:  const {bundledDependencies, main, files: onlyFiles} = pkg;
  - src/util/normalize-manifest/fix.js:    info.bundleDependencies = info.bundledDependencies;
  - src/util/normalize-manifest/fix.js:    delete info.bundledDependencies;

Resolving those gives me node_modules/module in the packed .tgz but no subdirectories/files.

bestander:
A PR with a test is welcome

cameronhunter:
@bestander: I've added a PR with a failing test case (#4598)

paul-soporan:
Closed in https://github.com/yarnpkg/yarn/pull/5966.

PR #4598, opened 2017-09-30
Failing test-case for issue #2636

cameronhunter:
**Summary**
This is a failing test-case for issue #2636. It tests that bundled dependencies are not correctly being included in the archive created by `yarn pack`.

**Test plan**
Changing the behavior of `pack` to include the bundled dependencies should make this failing test-case pass.


BYK:
You should probably mark the test with `.skip` so it doesn' break the builds? :)

BYK:
Thanks a lot btw!

PR #4599, opened 2017-09-30
Fix: allow package without manifest to be removed (fixes #4310)

### LINKED ISSUE #4310: Packages installed w/o manifest cannot be removed via command line: 'yarn remove'
Bug

Cannot remove a package that was installed w/o manifest (cf. #3624)

Install a package w/o manifest, e.g.
``yarn add https://github.com/pbakaus/scroller.git``

Then try to remove it:
``yarn remove scroller``

And the system complains:

```
error Couldn't find a package.json file in "...\\node_modules\\scroller"
```

The package should have been removed

yarn v0.28.4, npm v5.3.0, Windows 7

philquinn:
**Summary**

Fixes #4310 

If you add a package and it does not have a package.json, the remove command fails and does not actually remove the package. Now it ignores if there is none (as it can't do anything with it anyway) and removes the package correctly.

**Test plan**

Added test for a node_modules with a package that has no manifest. Checks if removed correctly.


arcanis:
👍

Issue #4450, opened 2017-09-14
Remove *.gz from default .yarnclean


nkbt:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
`autoclean` command has `*.gz` in default `.yarnclean` file.
https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/autoclean.js#L56

Problem: some packages are coming with `gz` to reduce package size

I reckon it is quite a minor issue, but not obvious why required files are missing.

**If the current behavior is a bug, please provide the steps to reproduce.**
```sh
docker run -it mhart/alpine-node:8
# then
rm -rf /test
mkdir /test
cd /test
cat > package.json <<- EOM
{
  "name": "test",
  "version": "0.0.1",
  "dependencies": {
    "left-pad": "*"
  }
}
EOM
yarn install
yarn autoclean --init
yarn remove left-pad
yarn add @std/esm
find .
```

Result:
```sh
/test # find .
.
./package.json
./.yarnclean
./node_modules
./node_modules/.yarn-integrity
./node_modules/@std
./node_modules/@std/esm
./node_modules/@std/esm/package.json
./node_modules/@std/esm/LICENSE
./node_modules/@std/esm/index.js
./yarn.lock
/test #
```

Running same code but omitting `yarn autoclean --init` line:
```sh
/test # find .
.
./package.json
./node_modules
./node_modules/.yarn-integrity
./node_modules/@std
./node_modules/@std/esm
./node_modules/@std/esm/README.md
./node_modules/@std/esm/package.json
./node_modules/@std/esm/LICENSE
./node_modules/@std/esm/index.js
./node_modules/@std/esm/esm.js.gz
./yarn.lock
```

**What is the expected behavior?**
`gz` should stay

**Please mention your node.js, yarn and operating system version.**
```sh
#OS: mhart/alpine-node:8

/ # node -v
v8.5.0
/ # yarn -v
1.0.1
/ #
```

BYK:
Sounds fair. Would you like to submit a PR?

BYK:
That said, this would be a breaking change for people relying on this behavior.

nkbt:
Sure, will do a PR.

This is only a default value for cleanup. Just removing `*.gz` might be a one solution.

Alternatively (or in addition) - maybe worth giving some note to user if some files were cleaned up. Like basic stats. For example:

> X files matched by .yarnclean and were automatically removed, Y Mb space saved

[...]

PR #4601, opened 2017-10-01
ref(autoclean): remove *.gz from the default .yarnclean file

nbarray:
Hello!

**Summary**

I removed the *.gz from the default .yarnclean generated by the `yarn autoclean --init` command. 

**Test plan**

I tested by hand and launch `yarn test` and didn't find a test case for the autoclean --init option. If needed I could write a test case to check the default .yarnclean file content against the constant DEFAULT_FILTER.

I hope it resolve half the issue #4450 :)

### DIFF
--- a/src/cli/commands/autoclean.js
+++ b/src/cli/commands/autoclean.js
@@ -53,7 +53,6 @@ Gruntfile.js
 *.yml
 
 # misc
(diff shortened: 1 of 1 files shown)


nbarray:
;)

Issue #3393, opened 2017-05-12
~ in config should be interpreted as HOME 


bestander:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the expected behavior?**

```
yarn config set cache-folder ~/yarn-caches
```
Should create a folder in HOME/yarn-caches and use it for caches.
Currently ~ is not interpreted as HOME and yarn just creates folder named `~` in CWD of yarn execution.

This will be useful for multiple Yarn commands, so it would be nice to make this generic.

bestander:
Duplicate of https://github.com/yarnpkg/yarn/issues/2008

bestander:
Contributions are very much welcome

samjacobclift:
This seems to be fixed in later versions of yarn
```
➜  test yarn config set cache-folder ~/yarn-caches
yarn config v0.22.0
success Set "cache-folder" to "/Users/samclift/yarn-caches".
✨  Done in 0.05s.
```

or am I missing something :)

bestander:
It translates `~` to home folder on the way into .yarnrc.
In .yarnrc it will be saved as `/Users/samclift/yarn-caches` which can't be shared among multiple users

bestander:
I'd like ~ inside .yarnrc to be interpreted as home folder

samjacobclift:
Ahh I see

mjesun:
Hi, I'll be working on this.

bestander:
@mjesun, consider making it generic so that other config settings could use it.

mjesun:
Hi @bestander,

There is not much we can do here. [Tilde expansion](https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html) happens before we can intercept it; so when we read it there is no access to the original value. However, when you quote it (i.e., `yarn config set cache-folder '~/yarn-caches'`) it works.

bestander:
Yeah, we can cover this case.
I am more interested in ~ expansion at read time

mjesun:
Ok, sounds cool :)

PR #4605, opened 2017-10-01
chore(utils): replace last `expandPath` with `resolveWithHome`

viatsko:
**Summary**

Looking at two solutions introduced in #3393 and #3756, the first one doesn't support win32, while the second does, sticking with the second one more beneficial and supports a wider range of OS.

Removed the stuff introduced in #3393 keeping only #3756.

#3756 also introduced config file normalization, so probably second argument to getOption is obsolete, will discover that and submit another PR if that's the case.

**Test plan**

Modified tests appropriately.


viatsko:
^ can't reproduce issue from tests

src/registries/npm-registry.js
```diff
@@ -65,7 +65,7 @@ function normalizePath(val: mixed): ?string {
   }
 
   if (typeof val !== 'string') {
-    val = '' + (val: any);
+    val = String(val);
```
viatsko: Thanks for taking care of a fix! Didn't find time at the beginning of the week to dig deeper!

One question. Does it make sense to have this at all? In the end, this will go to `path.resolve`, maybe it makes sense to throw an error instead?

PR #4606, opened 2017-10-01
chore(tests): Make integration tests passing on Node 7

viatsko:
The change introduced in 96c215c1ce5944a6fe993ab0d9b13e6edfe65675 caused tests to fail on node 7 as react-scripts is not compatible with it, this skips test on node v7 which is used within yarn's Docker.

I excluded this test for node v7, here's the reasoning behind it:
```
Summary of all failing tests
 FAIL  __tests__/integration.js (182.903s)
  ● yarn add react-scripts@1.0.13

    expect(received).not.toMatch(expected)

    Expected value not to match:
      /^warning /gm
    Received:
      "warning You are using Node \"7.10.1\" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: \"^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0\""

      at __tests__/integration.js:82:1793
```
Node 7.x is no longer supported, so this should be completely fine.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -59,11 +59,7 @@ addTest('https://git@github.com/stevemao/left-pad.git'); // git url, with userna
 addTest('https://github.com/yarnpkg/yarn/releases/download/v0.18.1/yarn-v0.18.1.tar.gz'); // tarball
 addTest('https://github.com/bestander/chrome-app-livereload.git'); // no package.json
 addTest('bestander/chrome-app-livereload'); // no package.json, github, tarball
-// many peer dependencies, there shouldn't be any peerDep warnings
-addTest('react-scripts@1.0.13', {
-  strict: true,
-  args: ['--no-node-version-check'],
-});
+addTest('react-scripts@1.0.13', {strict: true}, ['--no-node-version-check']); // many peer dependencies, there shouldn't be any peerDep warnings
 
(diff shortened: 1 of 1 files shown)


__tests__/integration.js
```diff
@@ -60,7 +60,8 @@ addTest('https://github.com/yarnpkg/yarn/releases/download/v0.18.1/yarn-v0.18.1.
 addTest('https://github.com/bestander/chrome-app-livereload.git'); // no package.json
 addTest('bestander/chrome-app-livereload'); // no package.json, github, tarball
 // Only run `react-scripts` test on Node 6+
-if (parseInt(process.versions.node.split('.')[0], 10) >= 6) {
+const nodeMajorVersion = parseInt(process.versions.node.split('.')[0], 10);
```
BYK: We still support Node 4 AFAIK. I think a better option would be to add the `--no-node-version-check` option to yarn invocations.

BYK:
>Thanks a lot but I have an alternative which may be a better solution :)

Just to clarify, the alternative is what I mentioned in the inline comment.

viatsko:
OK, this is done. Altered function signature (`addTest`) and added `--ignore-engines` as well only for that test, since it's designed to check `peerDependencies`. Keeping other tests for `yarn add` as they are now.

PR #4607, opened 2017-10-01
Updated node to 8.x in Dockerfile.dev

viatsko:
The reasoning behind this:
1) As discovered in #4606 not all of the tests are passing on 7.x 
2) Node 7.x is no longer supported anyway

I also removed MAINTAINER field from Dockerfile as it's not required.


Daniel15:
Oops, I already did this and pushed the Docker image to the Docker Hub, but forgot to post a PR for it. This looks good to me.

PR #4571, opened 2017-09-28
Add ability to filter yarn cache (fixes #3200)

philquinn:
**Summary**

This was proposed in #3200. It adds the ability filter the cache list with multiple filters.

If it matches any of the filters it will add the module to the list. If it doesn't it just skips.

**Test plan**
Added tests for this new feature.

Matching with single filter:
screen shot 2017-09-28 at 14 14 34

Matching with different filters:
screen shot 2017-09-28 at 14 18 22

No match in the cache:
screen shot 2017-09-28 at 14 21 52

PR #4609, opened 2017-10-01
Add ability to filter yarn list with pattern

philquinn:
**Summary**

In PR #4571, list was highlighted to be a candidate for adding a pattern option to allow filtering.

This PR adds this option.

Usage:
`yarn list --pattern mypattern`

Still allows:
`yarn list exact-package --pattern mypattern`

but this is now deprecated.

**Test plan**

Updated current tests to use new flag.

### DIFF
--- a/__tests__/commands/list.js
+++ b/__tests__/commands/list.js
@@ -81,6 +81,18 @@ describe('list', () => {
     });
   });
 
+  test.concurrent('accepts a pattern', (): Promise<void> => {
+    return runList([], {pattern: 'is-plain-obj'}, 'one-arg', (config, reporter): ?Promise<void> => {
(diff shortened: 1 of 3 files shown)


arcanis:
Could it be possible to restore the original behavior of filtering through the args? Otherwise it's a breaking change, which would technically require a version major bump :(

arcanis:
To be clear I meant that ideally we should support both the args and the flag, and add a deprecation warning on the args. This way it becomes a minor, and we can remove the arg support another minor later.

philquinn:
@arcanis That's really fair. I added the args back in and a deprecation notice. Let me know if it needs to be reworded or anything else needs changing. Thanks for reviewing :)

PR #4610, opened 2017-10-01
Add ability to upgrade based on pattern

philquinn:
**Summary**

In PR #4571, upgrade was highlighted to be a candidate for adding a pattern option to allow filtering.

This can be used with the current args or by itself.

Usage:
`yarn upgrade --pattern mypattern`
`yarn upgrade diff-package --pattern mypattern`

**Test plan**

Added test that filter with pattern and with the current args together.

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -25,6 +25,8 @@ import {normalizePattern} from './util/normalize-pattern.js';
 
 type ResolverRegistryNames = $Keys<typeof registryResolvers>;
 
+const micromatch = require('micromatch');
+
 export default class PackageRequest {
   constructor(req: DependencyRequestPattern, resolver: PackageResolver) {
     this.parentRequest = req.parentRequest;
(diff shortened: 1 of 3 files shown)


src/package-request.js
```diff
@@ -351,7 +352,11 @@ export default class PackageRequest {
     // prevents us from having to query the metadata for all packages.
     if (filterByPatterns && filterByPatterns.length) {
       const filterByNames = filterByPatterns.map(pattern => normalizePattern(pattern).name);
-      depReqPatterns = depReqPatterns.filter(dep => filterByNames.indexOf(normalizePattern(dep.pattern).name) >= 0);
+      depReqPatterns = depReqPatterns.filter(
+        dep =>
+          filterByNames.indexOf(normalizePattern(dep.pattern).name) >= 0 ||
+          (flags && flags.pattern && normalizePattern(dep.pattern).name.includes(flags.pattern)),
```
arcanis: Shouldn't you use `micromatch` rather than `includes`, which will only search for the substring?

philquinn: Yeah you're right. Changed now.

PR #4613, opened 2017-10-03
fix(resolver): Only optimize patterns in flat mode

### LINKED ISSUE #4550: can't seem to install request@^2.82.0 and request-promise-native@^1.0.5 together
**Do you want to request a *feature* or report a *bug*?** bug

Having an issue installing request@^2.82.0 I suspect it's to do with webdriverio pinning request@~2.81.0

https://github.com/graingert/yarn-request-strange
(remove the yarn.lock file, and run yarn install)

 request is still outdated, and I get a peerdep warning on install.

 request is updated, and no peerdep warning, webdriverio gets a nested request@~2.81.0

node: v8.5.0
uname -a: Linux fester 4.13.0-11-generic #12-Ubuntu SMP Tue Sep 12 16:03:57 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
lsb_release -a: No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu Artful Aardvark (development branch)
Release:	17.10
Codename:	artful
yarn: 1.1.0

BYK:
**Summary**

Fixes #4550. The optimization introduced in #4488 should only
apply to flat installations since even if a single pattern can
satisfy all resolved versions, it is not guaranteed that it is
strict enough for resolving correctly for all patterns under all
circumstances.

**Test plan**

Manual verification.

### DIFF
--- a/src/package-resolver.js
+++ b/src/package-resolver.js
@@ -545,9 +545,11 @@ export default class PackageResolver {
       this.resolveToResolution(req);
     }
 
-    for (const dep of deps) {
-      const name = normalizePattern(dep.pattern).name;
-      this.optimizeResolutions(name);
+    if (isFlat) {
+      for (const dep of deps) {
+        const name = normalizePattern(dep.pattern).name;
+        this.optimizeResolutions(name);
+      }
     }
 
     activity.end();


BYK:
I'm not happy that we don't have a test case for this but I have tried so hard to create a minimal test case and I've failed. I also couldn't get tests work properly with the provided package.json file or even a reduced version of it. I think it is a combination of too many dependencies and our ad-hoc request cache so we need to look into that later.

PR #4614, opened 2017-10-03
fix(network): Allow `https-proxy false` to not use a proxy for HTTPS

### LINKED ISSUE #4546: yarn tries to access https over http-proxy which doesn't support https
This is a feature request. I've tested this with current yarn (1.1 and master).

I'm behind a corporate http proxy which won't allow https connections for security reasons (https directly must be allowed). (Security reasons involve tunneling arbitrary connections via https, see e.g. https://github.com/proxytunnel/proxytunnel.)

I'd like to be able to set https-proxy to null explicitly to prevent this (this is supported in current npm).

I've tried to create a pull request for this, https://github.com/yarnpkg/yarn/compare/master...TauPan:patch-1 but apparently I don't understand how options work. If I test this with my branch, yarn apparently still tries to access the http-proxy even if I set https-proxy to null.

rally25rs:
**Summary**

Fixes #4546. Previous behavior was that if `.npmrc` or `.yarnrc` contained and `http-proxy` but not an `https-proxy`, HTTPS requests would "fall back" to the `http-proxy`.

This is not always the desired behavior; See #4546 sometimes one protocol needs a proxy and the other does not.

This PR adds the ability to set the `https-proxy` to `false` which will cause Yarn to not use an HTTPS proxy (and not fall-back to the HTTP proxy). This allows the code to treat `undefined` as the setting not being specified (which will fall back to http-proxy) and `false` as the setting being specified, but set to false (which will not use a proxy).

**Test plan**

Tested manually on Windows with Fiddler.

I have no idea how to test this since requests go through the mock http requester when running tests. Suggestions welcome!

### DIFF
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -416,7 +416,6 @@ export default class RequestManager {
       proxy = this.httpsProxy;
     }
     if (proxy && typeof proxy === 'string') {
-      console.log('PROXY', proxy);
       params.proxy = String(proxy);
     }
 
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -96,8 +96,8 @@ export default class RequestManager {
   userAgent: string;
   reporter: Reporter;
   running: number;
-  httpsProxy: ?string;
-  httpProxy: ?string;
+  httpsProxy: ?string | ?boolean;
+  httpProxy: ?string | ?boolean;
   strictSSL: boolean;
   ca: ?Array<string>;
   cert: ?string;
@@ -118,8 +118,8 @@ export default class RequestManager {
     userAgent?: string,
     offline?: boolean,
     captureHar?: boolean,
-    httpProxy?: string,
-    httpsProxy?: string,
+    httpProxy?: string | boolean,
+    httpsProxy?: string | boolean,
     strictSSL?: boolean,
     ca?: Array<string>,
     cafile?: string,
@@ -146,7 +146,13 @@ export default class RequestManager {
     }
 
     if (opts.httpsProxy != null) {
-      this.httpsProxy = opts.httpsProxy;
+      // fallback to httpProxy if httpsProxy is undefined (not specified).
+      // setting httpsProxy to `false` will not use a proxy for HTTPS. Requested by #4546
+      if (this.httpsProxy === undefined) {
+        this.httpsProxy = opts.httpProxy;
+      } else {
+        this.httpsProxy = opts.httpsProxy;
+      }
     }
 
     if (opts.strictSSL !== null && typeof opts.strictSSL !== 'undefined') {
@@ -407,10 +413,11 @@ export default class RequestManager {
 
     let proxy = this.httpProxy;
     if (params.url.startsWith('https:')) {
-      proxy = this.httpsProxy || proxy;
+      proxy = this.httpsProxy;
     }
-    if (proxy) {
-      params.proxy = proxy;
+    if (proxy && typeof proxy === 'string') {
+      console.log('PROXY', proxy);
+      params.proxy = String(proxy);
     }
 
     if (this.ca != null) {
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -78,9 +78,9 @@ export default class RequestManager {
     this._requestModule = null;
     this.offlineQueue = [];
     this.captureHar = false;
-    this.httpsProxy = null;
+    this.httpsProxy = '';
     this.ca = null;
-    this.httpProxy = null;
+    this.httpProxy = '';
     this.strictSSL = true;
     this.userAgent = '';
     this.reporter = reporter;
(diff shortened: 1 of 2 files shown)


BYK:
>CircleCI shows a failing test:

This is a transient error due to network hiccups. I really need to fix that by making that network message an info instead of a warning.

rally25rs:
> Can we drop that and only accept false

Yeah that's what I did originally, since the current system already handled it correctly. I added a last-second commit to translate "null" to false just to handle that case. I'm fine with removing it though. I'll do that...

BYK:
I'll merge after that change and hopefully tag 1.2.0 with this patch included!

rally25rs:
Pushed 👍  :shipit:

src/util/request-manager.js
```diff
@@ -96,8 +96,8 @@ export default class RequestManager {
   userAgent: string;
   reporter: Reporter;
   running: number;
-  httpsProxy: ?string;
-  httpProxy: ?string;
+  httpsProxy: ?string | ?boolean;
```
BYK: I think this type should be `string | boolean` and the default value should be `''` instead of `null` for the sake of simplicity.

src/util/request-manager.js
```diff
@@ -407,10 +413,10 @@ export default class RequestManager {
 
     let proxy = this.httpProxy;
     if (params.url.startsWith('https:')) {
-      proxy = this.httpsProxy || proxy;
+      proxy = this.httpsProxy;
     }
-    if (proxy) {
-      params.proxy = proxy;
+    if (proxy && typeof proxy === 'string') {
+      params.proxy = String(proxy);
```
BYK: So we check the type and then cast it to a `String` again? I'm guessing this is due to some messed up Flow typing. Shall we fix it since we are here?

src/util/request-manager.js
```diff
@@ -146,7 +146,13 @@ export default class RequestManager {
     }
 
     if (opts.httpsProxy != null) {
-      this.httpsProxy = opts.httpsProxy;
+      // fallback to httpProxy if httpsProxy is undefined (not specified).
+      // setting httpsProxy to `false` will not use a proxy for HTTPS. Requested by #4546
+      if (this.httpsProxy === undefined) {
```
BYK: `this.httpsProxy` can never be `undefined` as far as I can trace this code. The constructor sets it to `null` and then `setOption` gets called either with `false` an empty string or a string containing the value. I am very confused at this point.

src/util/request-manager.js
rally25rs: Yes this was purely to make Flow happy, because `proxy` is `string | boolean` and it won't let you assign it to `parame.proxy` because that only takes `string`. However I don't think the `typeof` check was actually needed.

rally25rs:
@BYK I think I simplified it a bit. `''` now represents an unset proxy instead of `null`.
There are a couple `|| ''` left to handle the case where no proxy is sent at all to the constructor opts, which only really happens in some unit tests.

TauPan:
I'm back from a brief holiday and would like to test if this behaves as intended.

However I'm having a little trouble building yarn. After I've fetched packages with `npm install` (because the old yarn  can't access the network) and then run `yarn build-dist`, running `./dist/bin/yarn install` gives me the following error:
```
yarn install v1.1.0
error An unexpected error occurred: "Cannot find module \"./rx.lite\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/taupan/git/yarn/yarn-error.log".
```

There seems to be a package `rx@4.1.0` installed, as both npm and yarn confirm. A package `rx.lite` is not in the npm registry, so I can't install it via npm.

TauPan:
Ok, I just noticed this was apparently released in yarn-1.2.0.

With this .yarnrc:
```
# yarn lockfile v1


https-proxy false
```

And this config `yarn config list`
```
> yarn config list                                                                                                                          git:(master) 11:48
yarn config v1.2.0
info yarn config
{ 'version-tag-prefix': 'v',
  'version-git-tag': true,
  'version-git-sign': false,
  'version-git-message': 'v%s',
  'init-version': '1.0.0',
  'init-license': 'MIT',
  'save-prefix': '^',
  'ignore-scripts': false,
  'ignore-optional': false,
  'strict-ssl': true,
  'user-agent': 'yarn/1.2.0 npm/? node/v6.11.2 linux x64',
  'https-proxy': false }
info npm config
{ proxy: 'http://webproxy.dfn-cert.de:3128/',
  'https-proxy': null,
  'strict-ssl': true,
  registry: 'http://registry.npmjs.org/',
  progress: true }
Done in 0.11s.
```

I still get:
```
yarn install v1.2.0
info No lockfile found.
[2/5] Resolving packages...
info There appears to be trouble with your network connection. Retrying...
```
Here are my proxy environment vars:
```
> env|grep -i proxy                                                                                                                         git:(master) 11:50
HTTP_PROXY=http://webproxy.dfn-cert.de:3128
NO_PROXY=localhost, .dfn-cert.de, .cert.dfn.de, 127.0.0.1, ::1
SOCKS_PROXY=
auto_proxy=file:///afs/dfn-cert.de/webproxy.dfn-cert.de.pac
ftp_proxy=http://webproxy.dfn-cert.de:3128
gopher_proxy=
http_proxy=http://webproxy.dfn-cert.de:3128
https_proxy=
no_proxy=127.0.0.1, localhost, .dfn-cert.de, .cert.dfn.de
socks_proxy=
```

PR #4630, opened 2017-10-04
feat(workspaces): Add `--ignore-workspace-root-check` flag and improve usage

arcanis:
**Summary**

Supercedes #4583 - cc @kaylieEB and @jgoz 

- Implements a `--ignore-workspace-root-check` option that replaces the `--dev` requirement when adding a dependency at the root. We received a lot of feedback about this warning, which prompted this change. You can now add both production and development dependencies to your packages, as long as you also use the new flag. Hopefully it will make things easier to understand!

- Fixes the bug of #4583 where the extraneous files were not removed from the workspaces.

- Partially fixes `outdated`, which will now print the whole tree (including the workspace root).

- Running `yarn add` in the root won't remove workspaces dependencies from the lockfile anymore. 

**Details**

At first this PR was only about adding the `--ignore-workspace-root-check` flag, but then I needed to change some parts of the code that #4583 was already altering, so I had to spend some time thinking about it as well, and eventually integrated the relevant changes here.

First, a distinction has to be made:

- First we have the `package.json` file format. We have a `package.json` for each workspace.
- Then we have the `yarn.lock` file format. There's only a single lockfile that aggregate all workspaces.

Which now brings me to my point: the following bugs, fixed in #4278 and #4583:

- "When called at the project root level with workspaces, upgrade and upgrade-interactive should only change the root dev dependencies"
- "upgrade to workspace root preserves child dependencies"
- "upgrade to workspace child preserves root dependencies"

They are not bugs.

Now that might sounds weird, but it all makes sense if you think about how we store the resolved dependencies! Our lockfile format looks like this:

```
"foo@~1.0.0"
    version: 1.0.0
```

Think about a workspace root containing two workspaces, A and B. Both depends on `foo@~1.0.0`. Now what will happen when we run `yarn upgrade foo` in A? Should we only update A? But then, what will our lockfile look like? We can't have two entries with the same key, like this:

```
"foo@~1.0.0"
    version: 1.0.0

"foo@~1.0.0"
    version: 1.0.5
```

And here lies the problem: the lockfile format doesn't (yet) allow us to operate on part of the workspace tree. We always must deal with the whole workspace tree, because if we don't we'll end up corrupting the lockfile by "forgetting" entries (like what happened when we were just not reading the workspaces package.json files), or overriding them accidentally (like what would happen if we were to just override `foo@~1.0.0`: B would be surprisingly affected as well by this change because it happens to share the same pattern range).

**Test plan**

Added some tests from #4583, fixed a few others, everything should be green.


src/cli/commands/add.js
```diff
@@ -216,6 +207,7 @@ export function hasWrapper(commander: Object): boolean {
 
 export function setFlags(commander: Object) {
   commander.usage('add [packages ...] [flags]');
+  commander.option('-W, --ignore-workspace-root-check', 'required to run yarn add inside a workspace root');
```
BYK: Interesting short-hand flag name.

src/package-linker.js
```diff
@@ -244,31 +244,39 @@ export default class PackageLinker {
...
-
-      if (await fs.exists(loc)) {
-        const files = await fs.readdir(loc);
-        let filepath;
-        for (const file of files) {
-          filepath = path.join(loc, file);
-          if (file[0] === '@') {
+    const scopedPaths: Set<string> = new Set();
+
+    const findExtraneousFiles = async basePath => {
```
BYK: It may be better in the future to extract this method outside of the closure.

arcanis: Since it needs to changes `scopedPaths` and `extraneousFiles` and isn't meant to be reused anywhere else in the code I thought it was cleaner to keep it together with the code segment that uses it.

src/cli/commands/add.js
arcanis: Didn't think long about this one, but you make me remember it conflicts with `-W(arning)` of compilers 😞 oh well, should be fine for this iteration.

arcanis:
@jgoz Yep I missed those fixes :/ Right now I'll have to merge this PR as is (the new `yarn add` option is important for common use cases), would you mind opening a new PR with the things I missed? I'm sorry, I know it must be time-consuming 😞

jgoz:
@arcanis No problem! I can open a new PR that will look very similar to #4583 😉

edmorley:
Thank you both for all your work on this - I can't wait to start using yarn workspaces once the remaining papercuts are fixed 😃 

Looking at #4583 which this superseded, I see a few test-cases (and possibly fixes) that weren't migrated to this PR - was this intentional? eg:
* for `outdated`: https://github.com/yarnpkg/yarn/pull/4583/files#diff-b803c2733e818112a7578aac3382f13cR134
* for `remove`: https://github.com/yarnpkg/yarn/pull/4583/files#diff-314ee529e34b31b97ee889bb502f5b8fR319

jgoz:
@edmorley I plan to migrate those fixes to a new PR shortly, though `outdated` will work as @arcanis suggested in the description of this PR (i.e., it will operate across the whole workspace, both root and individual packages).

Issue #4393, opened 2017-09-11
hasHTTPCapability issue with bitbucket shortcut resolver and private repo


bertho-zero:
It's a bug.

If I have a private dependency like `"activities": "bitbucket:openagenda/activities"` in my package.json, and I run `yarn install --verbose` then I have this error:

```
[1/4] Resolving packages...
verbose 0.407 Performing "HEAD" request to "https://bitbucket.org/openagenda/es-node".
verbose 0.867 Request "https://bitbucket.org/openagenda/es-node" finished with status code 302.
verbose 0.873 Performing "GET" request to "https://bitbucket.org/openagenda/es-node.git/info/refs?service=git-upload-pack".
verbose 0.98 Request "https://bitbucket.org/openagenda/es-node.git/info/refs?service=git-upload-pack" finished with status code 401.
verbose 0.981 Error: Error connecting to repository. Please, check the url.
    at /home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:33269:15
    at Generator.next (<anonymous>)
    at step (/home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:92:30)
    at /home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:103:13
    at process._tickCallback (internal/process/next_tick.js:109:7)
error An unexpected error occurred: "Error connecting to repository. Please, check the url.".
info If you think this is a bug, please open a bug report with the information provided in "/home/bertho/OpenAgenda/cibul-node/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

How to force usage of ssh instead of https for a shortcut of github/bitbucket ?

**Please mention your node.js, yarn and operating system version.**
```
$ node -v
node 7.10.1
$ yarn -v
yarn 1.0.0
$ uname -a
Linux bertho-X751LB 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

bertho-zero:
This is a problem with `this.hasHTTPCapability` in `src/resolvers/exotics/hosted-git-resolver.js` that returns true when it should not.

BYK:
@bertho-zero why do you think this is a bug? BitBucket indeed seems like to be supporting HTTP?

This looks like a permission/credentials issue to me and the error message doesn't reflect that so may be we should update that instead?

[...]

PR #4633, opened 2017-10-04
fix(git): Fix Bitbucket exotic shortcut resolvers with private repositories

bertho-zero:
**Summary**

* fix hasHTTPCapability issue with bitbucket shortcut resolver and private repo (#4393)
  *bug with a private repo that used like `"module": "bitbucket:team/repo"`*
* fix setRefRemote issue with exotic shortcut resolvers and branch/tag/commit
  *bug with a repo that used like `"module": "bitbucket:team/repo#tag"`*

**Test plan**

If I have a private dependency like `"activities": "bitbucket:openagenda/activities"` in my package.json, and I run `yarn install --verbose` then I have this error:

```
[1/4] Resolving packages...
verbose 0.407 Performing "HEAD" request to "https://bitbucket.org/openagenda/es-node".
verbose 0.867 Request "https://bitbucket.org/openagenda/es-node" finished with status code 302.
verbose 0.873 Performing "GET" request to "https://bitbucket.org/openagenda/es-node.git/info/refs?service=git-upload-pack".
verbose 0.98 Request "https://bitbucket.org/openagenda/es-node.git/info/refs?service=git-upload-pack" finished with status code 401.
verbose 0.981 Error: Error connecting to repository. Please, check the url.
    at /home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:33269:15
    at step (/home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:92:30)
    at /home/bertho/.config/yarn/global/node_modules/yarn/lib/cli.js:103:13
error An unexpected error occurred: "Error connecting to repository. Please, check the url.".
info If you think this is a bug, please open a bug report with the information provided in "/home/bertho/OpenAgenda/cibul-node/yarn-error.log".
```

### DIFF
--- a/src/resolvers/exotics/bitbucket-resolver.js
+++ b/src/resolvers/exotics/bitbucket-resolver.js
@@ -7,20 +7,6 @@ export default class BitbucketResolver extends HostedGitResolver {
   static hostname = 'bitbucket.org';
   static protocol = 'bitbucket';
 
-  static isVersion(pattern: string): boolean {
-    // bitbucket proto
-    if (pattern.startsWith('bitbucket:')) {
-      return true;
-    }
-
-    // bitbucket shorthand
-    if (/^[^:@%/\s.-][^:@%/\s]*[/][^:@\s/%]+(?:#.*)?$/.test(pattern)) {
-      return true;
-    }
-
-    return false;
-  }
-
   static getTarballUrl(parts: ExplodedFragment, hash: string): string {
     return `https://${this.hostname}/${parts.user}/${parts.repo}/get/${hash}.tar.gz`;
   }
@@ -45,14 +31,16 @@ export default class BitbucketResolver extends HostedGitResolver {
   }
 
   async hasHTTPCapability(url: string): Promise<boolean> {
-    return (
-      (await this.config.requestManager.request({
-        url,
-        method: 'HEAD',
-        queue: this.resolver.fetchingQueue,
-        followRedirect: false,
-        rejectStatusCode: 302,
-      })) !== false
-    );
+    // We don't follow redirects and reject a 302 since this means BitBucket
+    // won't allow us to use the HTTP protocol for `git` access.
+    // Most probably a private repo and this 302 is to a login page.
+    const bitbucketHTTPSupport = await this.config.requestManager.request({
+      url,
+      method: 'HEAD',
+      queue: this.resolver.fetchingQueue,
+      followRedirect: false,
+      rejectStatusCode: 302,
+    });
+    return bitbucketHTTPSupport !== false;
   }
 }
--- a/src/resolvers/exotics/bitbucket-resolver.js
+++ b/src/resolvers/exotics/bitbucket-resolver.js
@@ -45,17 +45,14 @@ export default class BitbucketResolver extends HostedGitResolver {
   }
 
   async hasHTTPCapability(url: string): Promise<boolean> {
-    try {
-      const result = await this.config.requestManager.request({
+    return (
+      (await this.config.requestManager.request({
         url,
         method: 'HEAD',
         queue: this.resolver.fetchingQueue,
         followRedirect: false,
         rejectStatusCode: 302,
-      });
-      return result !== false;
-    } catch (e) {
-      return false;
-    }
+      })) !== false
+    );
   }
 }
--- a/src/resolvers/exotics/bitbucket-resolver.js
+++ b/src/resolvers/exotics/bitbucket-resolver.js
@@ -7,6 +7,20 @@ export default class BitbucketResolver extends HostedGitResolver {
   static hostname = 'bitbucket.org';
   static protocol = 'bitbucket';
 
+  static isVersion(pattern: string): boolean {
+    // bitbucket proto
+    if (pattern.startsWith('bitbucket:')) {
+      return true;
+    }
+
+    // bitbucket shorthand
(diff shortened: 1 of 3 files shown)


src/util/request-manager.js
```diff
@@ -395,7 +396,7 @@ export default class RequestManager {
           const errMsg = (body && body.message) || reporter.lang('requestError', params.url, res.statusCode);
           reject(new Error(errMsg));
         } else {
-          if (res.statusCode === 400 || res.statusCode === 404 || res.statusCode === 401) {
+          if (~[400, 401, 404].concat(params.rejectStatusCode || []).indexOf(res.statusCode)) {
```
BYK: Please do not use the vague `~[...].indexOf(...)` notation. Use `!== -1`.

src/util/git.js
```diff
@@ -364,7 +365,7 @@ export default class Git implements GitRefResolvingInterface {
   }
 
   async setRefRemote(): Promise<string> {
-    const stdout = await spawnGit(['ls-remote', '--tags', '--heads', this.gitUrl.repository]);
+    const stdout = await spawnGit(['ls-remote', '--tags', '--heads', this.gitUrl.repository.split('#')[0]]);
```
BYK: Why do we need this now?

src/util/git.js
```diff
@@ -61,6 +61,7 @@ export default class Git implements GitRefResolvingInterface {
   static npmUrlToGitUrl(npmUrl: string): GitUrl {
     // Expand shortened format first if needed
     npmUrl = npmUrl.replace(/^github:/, 'git+ssh://git@github.com/');
+    npmUrl = npmUrl.replace(/^bitbucket:/, 'git+ssh://git@bitbucket.org/');
```
BYK: You can just chain `.replace` calls instead of assigning twice.

src/resolvers/exotics/bitbucket-resolver.js
```diff
@@ -7,6 +7,20 @@ export default class BitbucketResolver extends HostedGitResolver {
   static hostname = 'bitbucket.org';
   static protocol = 'bitbucket';
 
+  static isVersion(pattern: string): boolean {
+    // bitbucket proto
+    if (pattern.startsWith('bitbucket:')) {
```
BYK: This is already covered by `protocol` above, right? Please use the already defined constant or modify it to not use a custom string.

src/resolvers/exotics/bitbucket-resolver.js
```diff
@@ -7,6 +7,20 @@ export default class BitbucketResolver extends HostedGitResolver {
...
+      return true;
+    }
+
+    // bitbucket shorthand
+    if (/^[^:@%/\s.-][^:@%/\s]*[/][^:@\s/%]+(?:#.*)?$/.test(pattern)) {
```
BYK: Can you move this outside of the function, make it ` const` and add some comments about this regular expression? It is not apparent what it is doing for someone not familiar with bitbucket.

src/resolvers/exotics/bitbucket-resolver.js
```diff
@@ -7,6 +7,20 @@ export default class BitbucketResolver extends HostedGitResolver {
...
+      return true;
+    }
+
+    return false;
```
BYK: You can probably combine all these into a single expression:
```
return pattern.startsWith(protocol) || BITBUCKET_PATTERN.test(pattern);
```

src/resolvers/exotics/bitbucket-resolver.js
```diff
@@ -29,4 +43,16 @@ export default class BitbucketResolver extends HostedGitResolver {
   static getHTTPFileUrl(parts: ExplodedFragment, filename: string, commit: string): string {
     return `https://${this.hostname}/${parts.user}/${parts.repo}/raw/${commit}/${filename}`;
   }
+
+  async hasHTTPCapability(url: string): Promise<boolean> {
+    return (
+      (await this.config.requestManager.request({
```
BYK: Storing the return value of this request in a nicely-named variable to give a hint to the reader about what this is and why are we doing things would be great. It would also make the `return` line more apparent:
```
// We don't follow redirects and reject a 302 since this means BitBucket won't allow us to use the HTTP protocol for `git` access. Most probably a private repo and this 302 is to a login page.
const bitbucketHTTPSupport = await ...;
return bitbucketHTTPSupport !== false;

src/resolvers/exotics/bitbucket-resolver.js
bertho-zero: I did not understand that it was the shortest shortcut, the one that should only work with github.

src/util/git.js
bertho-zero: I do not know how to explain it but yarn attempts to run this command, which does not work:
`git ls-remote --tags --heads ssh://git@bitbucket.org/openagenda/validators#v1.12.2`
*instead of `git ls-remote --tags --heads ssh://git@bitbucket.org/openagenda/validators`*
(from https://github.com/yarnpkg/yarn/blob/master/src/resolvers/exotics/hosted-git-resolver.js#L239)

I can not reproduce this bug outside my big repo which gathers fifty repo that tangle each other, and use a lot of `validators`. This is the only way I have found to effectively target this bug.

bertho-zero:
@BYK It sounds complicated to me to add an integration test with a private repo.

src/util/git.js
BYK: Alright, thanks for the explanation :)

PR #4636, opened 2017-10-05
Upgrade transient deps during upgrades.

### LINKED ISSUE #4476: Yarn Upgrade not changing yarn.lock since v1.0
BUG

nothing happens when i run yarn upgrade. No yarn-error.log is generated either. It happens for other people too.

> yarn upgrade v1.0.2
success All of your dependencies are up to date.

But they are not all up to date.

in the other project, i see this every time i run yarn upgrade

> success Saved lockfile.
success Saved 5 new dependencies.
├─ angular-auto-focus@1.0.4
├─ angular-bind-html-compile@1.4.1
├─ chiri@1.0.0
├─ html2canvas@0.4.1
└─ sass-quatch@1.0.3

(it's worth noting that those 5 packages are the only 5 that come up when i say yarn upgrade-interactive, and they are all direct git repo dependencies, not npm packages. the other project has no such dependencies)

but yarn.lock never changes

yarn upgrade also no longer installs dependencies when node_modules is empty.

I am just running yarn upgrade in my existing project.  It's private so I can't link to it. But it happens in multiple projects

yarn.lock should update along with dependencies

node: 8.5.0
yarn: 1.0.2
OS: Mac Sierra

Someone else has yarn 0.27.5 with the same project and it updates the dependencies. They are not just all up to date.

One other thing is that I am using an internal yarn registry (with artifactory), but it has never had an issue (and still works as of now) with yarn 0.27.5.

I'll try to post my package.json if that sheds any light on it

```js
{
  "name": "whatever",
  "version": "9.0.1",
  "description": "foo",
  "private": false,
  "module": "es/components.js",
  "publishConfig": {
    "registry": "some internal url"
  },
  "files": [
    "es/"
  ],
  "scripts": {
    "unimportant": null
  },
  "repository": {
    "type": "git",
    "url": "some url"
  },
  "dependencies": {
    "classnames": "^2.2.5",
    "lodash-es": "^4.17.4",
    "numeral": "^2.0.3",
    "react-addons-shallow-compare": "^15.5.2",
    "react-dates": "^12.1.1",
    "react-onclickoutside": "^6.4.0",
    "shallowequal": "^1.0.2"
  },
  "devDependencies": {
    "@storybook/addon-info": "^3.2.0",
    "@storybook/addon-knobs": "^3.2.0",
    "@storybook/addon-options": "^3.2.3",
    "@storybook/react": "^3.2.3",
    "babel-polyfill": "^6.16.0",
    "immutable": "^3.8.1",
    "moment": "^2.18.1",
    "react": "^15.5.4",
    "react-dom": "^15.5.4",
    "react-router-dom": "^4.1.1",
    "svg-sprite-loader": "^3.2.5"
  },
  "peerDependencies": {
    "babel-polyfill": "^6.16.0",
    "moment": "^2.18.1",
    "react": "^15.3.2",
    "react-dom": "^15.3.2"
  },
  "registry": "npm"
}
```

and i have a .yarnrc file pointing to an internal yarn registry (running on artifactory. it works with 0.27.5)

rally25rs:
**Summary**

Fixes #4476 

Due to the changes made to `upgrade` for v1, we inadvertently stopped upgrading transient dependencies.

This happend for 2 reasons:
* because the existing lockfile is left in-place, so transient deps were already at a locked version
* only direct dependencies that were actually out of date were upgraded (for performance reasons)

This change will now clear dependencies to be upgraded, recursively, from the lockfile. This lets transient deps get upgrade as well. In the case that all packages are being upgraded, the lockfile is just removed and rebuilt.

I also found a bug in upgrade in yarn v1.0 and fixed that while in here...

in Yarn v1.0 if you `yarn upgrade some-package` then it will possibly upgrade `some-package` if there is a new version, but the transient deps underneath `some-package` would not be upgraded (because of the entries in the lockfile, mentioned above). With this update those transient deps will be upgraded.

**Test plan**

Added 2 additional unit tests that check transient deps.


__tests__/commands/upgrade.js
```diff
@@ -31,6 +31,11 @@ const expectInstalledDevDependency = async (config, name, range, expectedVersion
   await _expectDependency('devDependencies', config, name, range, expectedVersion);
 };
 
+const expectInstalledTransientDependency = async (config, name, range, expectedVersion) => {
```
bdwain: i think transient is not the right word here. It should be transitive.

bdwain:
I'm running this locally and it seems to ignore my .yarnrc file in my project. it Points to an internal registry, but yarn upgrade --verbose logs calls to registry.yarnpkg.com. I'm not sure if it's related to running locally or not, but yarn install did not have the same issue.

__tests__/commands/upgrade.js
rally25rs: whoopsies 😆 thanks for noticing that.

rally25rs:
@bdwain is this .yarnrc registry thing a new error in this PR, or did it not work in v1.0 either? I'll try to set something up here to reproduce the error...

rally25rs:
Hmm, OK so yeah I see the same error as @bdwain if I have a `registry "http://localhost:4873/"` set in my `.yarnrc`

With Yarn v1.0:

```
$ yarn upgrade --verbose
...
verbose 0.401 Performing "GET" request to "http://localhost:4873/strip-ansi".
verbose 0.402 Performing "GET" request to "http://localhost:4873/array-union".
verbose 0.485 Request "http://localhost:4873/array-union" finished with status code 200.
verbose 0.49 Request "http://localhost:4873/strip-ansi" finished with status code 200.
...
```

With this PR:

```
$ yarn upgrade --verbose
...
verbose 0.752 Performing "GET" request to "https://registry.yarnpkg.com/strip-ansi".
verbose 0.753 Performing "GET" request to "https://registry.yarnpkg.com/array-union".
verbose 0.878 Request "https://registry.yarnpkg.com/array-union" finished with status code 200.
verbose 0.88 Request "https://registry.yarnpkg.com/strip-ansi" finished with status code 200.
...
```

that's very weird. I have no clue what I would have changed that would have affected that. Well, I'll start digging in...

arcanis:
> This change will now clear dependencies to be upgraded, recursively, from the lockfile.

I'm not convinced this is the right behavior. Let's say with have:

- A@1.0.0, depends on B@~1.0.0
- A@1.0.1, depends on B@~1.0.0
- B@1.0.0
- B@1.0.1
- Root depends on A@~1.0.0 and B@~1.0.0
- Lockfile is A@1.0.0 and B@1.0.0

If I run `yarn update A`, I don't expect B to also be updated to `1.0.1`.

rally25rs:
@arcanis I'm not sure how you would get round that. They would both share one `B@~1.0.0` entry in the lockfile.

rally25rs:
@arcanis hmm, interesting case. Maybe direct dependencies get excluded from that recursion?

rally25rs:
@arcanis I added a test for the case you mentioned. (also fixed merge conflicts and added some verbose logging to the upgrade process)

arcanis:
Thanks @rally25rs! I just merged #4654 which should ship with fixes to a few commands that include upgrade and upgrade-interactive, do you mind rebasing and checking if it makes your tests pass as well?

voxpelli:
Tried this PR on a larger project of mine and it correctly produced the same `yarn.lock` that was produced by removing the `yarn.lock` and starting over from scratch. 👌 Thanks!

rally25rs:
@voxpelli Thank you for taking the time to try this out!

@arcanis Merged master and resolved conflicts.

arcanis:
Thank you both!

bdwain:
thanks again!

vkrol:
@rally25rs if I run the upgrade command with the latest flag (for example yarn upgrade package-name --latest) then the transitive dependencies won’t be upgraded. I think that it is the wrong behavior. What do you think about it?

PR #4642, opened 2017-10-05
bug: added parent node_modules/.bin folders in PATH during run commands

### LINKED ISSUE #4640: Yarn unable to execute binaries installed in higher level node_modules/.bin
Given the following folder hierarchy: 

- packageA
  - node_modules
  - .bin
     - myprogram
  - packageB

`yarn run somecommand`

from inside **packageA/node_modules/packageB** fails.  The same command using npm works.

The problem showed up because **packageB** has a postinstall directive that executes **myprogram**.

I've determined the problem is because yarn doesn't traverse the directory tree looking for additional node_module directories to add to the path.  

I'll submit a PR shortly.

lucbelliveau:
**Summary**
fixes #4640 

Allows execution of "postinstall" commands in nested node_module folders when the binary  dependencies were installed in a higher level node_modules/.bin folder.  (Same as npm run)

**Test plan**
yarn run test run


hoodie:
@lucbelliveau this failed in in CI 11days ago - do you still have an eye on this?

lucbelliveau:
@hoodie I'll take a look

hoodie:
@lucbelliveau looks like you only have a linting issue there - you can pull from my fork if you like
https://github.com/hoodie/yarn

lucbelliveau:
@hoodie thanks; looks like I was fixing it when your message came in!

lucbelliveau:
@BYK Looks like the CI server ran out of disk space...

hoodie:
can't we just trigger that build to run again? @lucbelliveau

lucbelliveau:
@hoodie apparently closing and opening the PR rebuilds... underway now!

jmaher409:
Would it be possible to have this reviewed and merged soon? This is a blocker for our CI systems that need to test on all operating systems.

jmaher409:
@BYK can you explain your concerns with it? It seems like the lack of it breaks the expected hoisting mechanics.

arcanis:
I don't see this working. Take this case:

- Root depends on A
- A depends on B and C
- B depends on nothing
- C depends on nothing
- C provides the `foo` binary
- Let's say that Yarn x.y.z has no hoisting at all, so we have the following paths:
    - node_modules/A/node_modules/B
    - node_modules/A/node_modules/C
    - node_modules/A/node_modules/.bin/foo
  With your diff, B will be able to invoke `foo` correctly.
- Now let's say that Yarn u.v.w has a good hoisting mechanism. We now have the following paths:
    - node_modules/B
    - node_modules/C
    - node_modules/A/node_modules/.bin/foo
  Contrary to before, B will not be able to call `foo` anymore.

So this patch seems to require a standardized and frozen hoisting mechanism, which is not something we aim for.

lucbelliveau:
@arcanis In the scenario you suggest `C` should be a dependency of `B` ... in that case everything works as expected.  If `B` needs the binary provided by `C`, it is up to the developer to properly define the dependencies.

arcanis:
I agree with the concept, but it might be a bit too break-y for our taste. We'll close it for now - we have some plans to make constraints stricter, and this might fall under this umbrella. Thanks anyway for your work!

PR #4238, opened 2017-08-23
feat(configuration): add custom registry cli flag

lirantal:
**Summary**
implements #792 - addresses a command line flag --registry to specify a registry that overides yarn/npm configuration for installing dependencies.

PR #4643, opened 2017-10-05
fix(config): Don't override registry with `undefined` by default

BYK:
**Summary**

Follow up to #4238. We were always passing the `registry` key in
registry overrides but its value was `undefined` when an override
was not in place. `Object.assign` doesn't care about that though
so we were overriding the registry all the time, mostly with
`undefined`.

**Test plan**

Added new test case.


BYK:
@rally25rs thanks a lot for reporting it!

PR #4478, opened 2017-09-15
Fix: Resolve peerDependencies from all higher levels, not just root

BYK:
**Summary**

Fixes #4446, fixes #4433, fixes #2688, fixes #2387. Follow up to #3803. The fix in #3893 was
too aggressive, allowing only top-level dependencies to be used in
peer dependency resolution which was incorrect. This patch allows
resolving peer dependencies from the same or higher levels in the
dependency tree.

**Test plan**

Additional unit and integration tests.

PR #4644, opened 2017-10-05
chore(resolver): Minor improvements in resolver code and tests

BYK:
**Summary**

This is a follow up to #4484 and #4478 which improves the code
around those areas a bit and removes a now-unnecessary `while`
loop.

**Test plan**

Existing tests should pass.

### DIFF
--- a/src/package-reference.js
+++ b/src/package-reference.js
@@ -66,13 +66,7 @@ export default class PackageReference {
   addRequest(request: PackageRequest) {
     this.requests.push(request);
 
-    let requestLevel = -1;
-    let currentRequest = request;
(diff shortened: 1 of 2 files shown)


src/package-reference.js
```diff
@@ -66,13 +66,7 @@ export default class PackageReference {
   addRequest(request: PackageRequest) {
     this.requests.push(request);
 
-    let requestLevel = -1;
-    let currentRequest = request;
-    while (currentRequest && requestLevel < this.level) {
-      requestLevel += 1;
-      currentRequest = currentRequest.parentRequest;
-    }
-    this.level = requestLevel;
```
arcanis: Well spotted!

Issue #4469, opened 2017-09-15
Yarn outdated and upgrade-interactive should work in workspaces' packages also


marcofugaro:
Hey, I tried implementing Workspaces on [a repo](https://github.com/accurat/accurapp) and here is some feedback that you might be interested in.

Basically I am having trouble running `yarn outdated`, `yarn upgrade-interactive` or other similar commands in a workspaces project.

**What is the current behavior?**
`yarn outdated` or `yarn upgrade-interactive` run from the root folder work only for the root package.json.
If I run those commands in a `packages/*` folder the error `No lockfile in this directory. Run 'yarn install' to generate one.` comes up.

**What is the expected behavior?**
`yarn outdated` or `yarn upgrade-interactive` should also check workspaces folders when run from root folder.

**Please mention your node.js, yarn and operating system version.**
yarn v1.0.2 with workspaces flag enabled
node v8.5.0

This is related to #4442

SimenB:
It's the same thing if using the `workspace` command.

```sh-session
$ yarn workspace @podium/client outdated
yarn workspace v1.0.2
yarn outdated v1.0.2
error No lockfile in this directory. Run `yarn install` to generate one.
error Command failed.
Exit code: 1
Command: /Users/simbekkh/.nvm/versions/node/v8.4.0/bin/node
Arguments: /usr/local/Cellar/yarn/1.0.2/libexec/lib/cli.js outdated
Directory: /Users/simbekkh/repos/podium/packages/podium-client
Output:

info Visit https://yarnpkg.com/en/docs/cli/workspace for documentation about this command.
```

Also, https://yarnpkg.com/en/docs/cli/workspace 404s

SukantGujar:
This should work the other way round too, i.e., when running it from the root folder of the workspace, it should detect and show possible upgrades for all packages inside that workspace.

donaldpipowitch:
I created a similar issue for `$ yarn list`. I guess they are duplicates https://github.com/yarnpkg/yarn/issues/4494? I'll close my issue.

edmorley:
Hi! I think the fixes in #4654 might have solved this. Could you give yarn 1.2.0 a go and see if it still reproduces? :-)

donaldpipowitch:
Seems to work great!

ericclemmons:
@edmorley Confirmed #4654 (v1.2.0) fixes this!

marcofugaro:
Tested it, it now works fine

[...]

PR #4654, opened 2017-10-06
Fix: workspace support in several commands

### LINKED ISSUE #4328: yarn outdated not working w/workspaces

### LINKED ISSUE #4334: "yarn remove" inside workspace fails

### LINKED ISSUE #4348: Adding root dependency removes workspace packages from lockfile

jgoz:
**Summary**

Fixes #4334
Fixes #4328
Fixes #4348

(Follow-up to #4630, #4583).

This gets several commands working when used in a workspace-enabled solution:

- `outdated`:
  - when used in a workspace root or any child packages, it will show outdated packages across the entire workspace
  - added "Workspace" column to show where the outdated package is installed
- `remove`:
  - when used in a workspace root, removes given package from the root only
  - when used in a workspace child, removes given package from that child only
- `upgrade`:
  - when used in a workspace root, upgrades given package in root only
  - when used in a workspace child, upgrades given package in child only
- `upgrade-interactive`:
  - shows outdated packages across entire workspace, and can upgrade any/all
  - added "workspace" column to show where outdated package is installed

This is slightly different from how @arcanis described in #4630, specifically with `upgrade`. I thought it would be less confusing/frustrating for users, especially if child workspaces have conflicting versions of the same dependency. Rather than forcing them all to the same version, this allows for fine-grained control while offering a whole-workspace upgrade option with `upgrade-interactive`.

**Test plan**

- Added tests where appropriate
- Existing tests pass
- Screenshots of updated command output below:

`outdated` with "Workspace" column:
image

`upgrade-interactive` with "workspace" column:
image


JohnFultonian:
Thanks for your work on this, we just started using yarn workspaces recently and this will make our lives a lot easier.

arcanis:
> This is slightly different from how @arcanis described in #4630, specifically with upgrade. I thought it would be less confusing/frustrating for users, especially if child workspaces have conflicting versions of the same dependency.

Then how does it work if A and B both depend on C@~1.0.0, and you run `yarn upgrade C` from inside A? It will upgrade C in both A and B, right?

jgoz:
> Then how does it work if A and B both depend on C@~1.0.0, and you run yarn upgrade C from inside A? It will upgrade C in both A and B, right?

No, it will only upgrade C in A, and the lockfile will now have two entries for C. If you run `yarn upgrade C` in B, then they will once again converge.

This is a choice that was initially dictated by the complexity of implementing it the way you described it (where it upgrades C in both A and B when you run the command from just one workspace). There were a couple of issues to deal with here:
- What if C is a `dependency` in A and a `devDependency` in B? We would need to change the way `upgrade` works to be more like `upgrade-interactive`, where it runs install multiple times to ensure that difference is preserved.
- How would a user indicate that they _only_ want to upgrade C in A, leaving the version as ~1.0.0 in B? We would need to add a CLI flag or something to toggle this behaviour.

Then I thought about `add` and `remove`. Those only operate on a single workspace child/root, so having `upgrade` work this way too is arguably less confusing and more consistent. That it avoids bigger changes, refactoring in `install`, and extra flags is a bonus. And for upgrading C in both A and B, we can use `upgrade-interactive`.

arcanis:
i-ship-it-gif-16

vvo:
awesome work, thanks

ericclemmons:
I just tested #4469 with v1.2.0 and it works now! 👏

milesj:
@jgoz Awesome work on this PR. Such a life saver and it's been working flawlessly!

PR #4660, opened 2017-10-08
chore(repo): Add `yarn commit` for structured commit messages

BYK:
**Summary**

Adds commitizen and adopts the cz-conventional-changelog adapter for commit logs. Will be followed
up by commit message linters and automatic version number bumping in CI builds.

**Test plan**

Run `yarn install` and then `yarn commit -a` after making some changes to test the new awesomeness. This commit message was also generated via that command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,8 @@
     "babel-preset-node5": "^10.2.0",
     "babel-preset-stage-0": "^6.0.0",
     "babylon": "^6.5.0",
+    "commitizen": "^2.9.6",
+    "cz-conventional-changelog": "^2.0.0",
     "eslint": "^4.3.0",
     "eslint-config-fb-strict": "^20.1.0-delta.3",
     "eslint-plugin-babel": "^4.0.0",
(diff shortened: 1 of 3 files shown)


BYK:
The current adapter does not have the separate **summary** and **test plan** fields so I need to add them in a follow up. This is more like a proof of concept.

package.json
```diff
@@ -111,16 +113,26 @@
     "test-ci": "yarn build && yarn test-only",
     "test-only": "node --max_old_space_size=4096 node_modules/jest/bin/jest.js --verbose",
     "test-coverage": "node --max_old_space_size=4096 node_modules/jest/bin/jest.js --coverage --verbose",
-    "watch": "gulp watch"
+    "watch": "gulp watch",
+    "commit": "git-cz"
```
kaylieEB: nit, alphabetize and put this above?

BYK:
>I guess my only question is how do we encourage/enforce people to use the script command instead of the typical git commit workflow. Have you considered adding a git hook?

I did but you cannot distribute git hooks with the repo automatically. My plan is to introduce another linting step that checks your commit message and fail the CI if it doesn't conform. `standard-changelog` provides many tools around this.

>(I was reading about Husky) or at least updating https://github.com/yarnpkg/yarn/blob/master/CONTRIBUTING.md?

That's also on my mind, thanks for the reminder!

package.json
BYK: Will do in a follow up to avoid a rebuild for this.

PR #4669, opened 2017-10-09
fix(config): Relay exit code when using yarn-path from yarnrc

BYK:
**Summary**

Follow up to #4204. We forgot to relay the exit code of the
newly spawned yarn instance when using `yarn-path` which is
causing false negatives especially when using `yarn run`. This
patch relays the exit code of the spawned process.

**Test plan**

Added a new test that fails without the fix.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -250,30 +250,48 @@ test('yarnrc arguments', async () => {
   expect((await fs.stat(`${cwd}/yarn-cache`)).isDirectory()).toBe(true);
 });
 
-test('yarnrc binary path (js)', async () => {
-  const cwd = await makeTemp();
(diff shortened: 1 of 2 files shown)


__tests__/integration.js
```diff
@@ -250,30 +250,48 @@ test('yarnrc arguments', async () => {
...
+    await fs.writeFile(`${cwd}/override.js`, 'process.exit(1);');
+
+    let error = false;
+    try {
+      await runYarn([], {cwd});
+    } catch (err) {
+      error = true;
+    }
+
+    expect(error).toEqual(true);
```
bestander: where is the exit code test?

BYK: Well I was lazy. Incoming

PR #4679, opened 2017-10-10
Fixes how the exit code was forwarded when using the yarnrc path

arcanis:
**Summary**

Yarn doesn't correctly forward the exit code if the process is executed via `spawnp`.

**Test plan**

Will write tests in a separate PR, but I'd prefer to merge this one asap.


arcanis:
Thanks @kaylieEB! 👍

PR #4680, opened 2017-10-10
Adds an extra test for the yarn path feature

arcanis:
**Summary**

This is the missing test for #4679 :)


__tests__/integration.js
```diff
@@ -292,6 +292,24 @@ describe('yarnrc path', () => {
 
     expect(error).toEqual(123);
   });
+
+  test('sh file exit code', async () => {
```
BYK: Windows :(

PR #4681, opened 2017-10-10
Restore version if publish is aborted before publishing

### LINKED ISSUE #662: Aborting publish should not change version number
bug
When using yarn publish and aborting after entering a version number, the version number is not reset. As a consequence it seems impossible to publish the same version number "again" without reverting the change manually.

```
[tkloht@Tobias-Air ~/Code/react-component-scripts:master] yarn publish
yarn publish v0.15.1
[1/4] Bumping version...
info Current version: 1.4.1
question New version: 1.4.2
info New version: 1.4.2
[2/4] Logging in...
info npm username: tkloht
info npm username: tobias.kloht@gmail.com
error canceled
    at Interface.<anonymous> (/Users/tkloht/.nvm/versions/node/v6.2.1/lib/node_modules/yarnpkg/node_modules/read/lib/read.js:66:13)
    at emitNone (events.js:86:13)
    at Interface.emit (events.js:185:7)
    at Interface._ttyWrite (readline.js:707:16)
    at ReadStream.onkeypress (readline.js:115:10)
    at emitTwo (events.js:106:13)
    at ReadStream.emit (events.js:191:7)
    at emitKeys (internal/readline.js:385:14)
    at next (native)
    at ReadStream.onData (readline.js:939:36)
info Visit http://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

after this the version number is already changed:
```
diff --git a/package.json b/package.json
index e8764fc..c6d0fb4 100644
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tkloht/react-component-scripts",
-  "version": "1.4.1",
+  "version": "1.4.2",
```

This makes it impossible to publish version 1.4.2 in this example:

```
[tkloht@Tobias-Air ~/Code/react-component-scripts:master*] yarn publish
yarn publish v0.15.1
[1/4] Bumping version...
info Current version: 1.4.2
question New version: 1.4.2
error New version is the same as the current version.
info Visit http://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

When I abort a publish the version number in package.json should not have changed.

node: v6.2.1, yarn: 0.15.1, OSX 10.12

qfox:
**Summary**

Bug fix: the version in `package.jscon` would be updated even if the `publish` command failed prior to actually publishing.

Fixes #662

**Test plan**

This is a wip, need input on how to write a test for this. There currently appear to be no tests for `publish` to take as an example AND it needs to throw an error somewhere after writing the version.

### DIFF
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -130,7 +130,6 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   // save current version so we can restore it in case of an abort
   const oldVersion = await getVersion(config);
 
-  // sorry if this try/catch breaks your stack trace :'(
   try {
     return await _run(config, reporter, flags, args, pkg, dir);
   } catch(e) {
@@ -144,7 +143,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
       // this was a last ditch effort to restore the version. ignore errors here
     }
 
-    throw new Error(e);
+    throw e;
   }
 }
 async function _run(config: Config, reporter: Reporter, flags: Object, args: Array<string>, pkg:any, dir:string): Promise<void> {
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -4,7 +4,7 @@ import type {Reporter} from '../../reporters/index.js';
 import type Config from '../../config.js';
 import NpmRegistry from '../../registries/npm-registry.js';
 import {MessageError} from '../../errors.js';
-import {setVersion, setFlags as versionSetFlags} from './version.js';
+import {getVersion, setVersion, setFlags as versionSetFlags} from './version.js';
 import * as fs from '../../util/fs.js';
 import {pack} from './pack.js';
 import {getToken} from './login.js';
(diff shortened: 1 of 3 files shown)


src/cli/commands/publish.js
```diff
@@ -127,6 +127,27 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
+      // try to restore the version
+      // note that if it crashed on setting the version, the original
+      // version probably wasn't changed in the first place
+      reporter.step(1, 0, reporter.lang('restoringVersion'));
+      await setVersion(config, reporter, {newVersion: oldVersion}, [], false);
+    } catch (_) {
+      // this was a last ditch effort to restore the version. ignore errors here
+    }
+
+    throw new Error(e);
```
arcanis: You can just rethrow `e` :)

src/cli/commands/publish.js
```diff
@@ -127,6 +127,27 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     throw new MessageError(reporter.lang('unknownFolderOrTarball'));
   }
 
+  // save current version so we can restore it in case of an abort
+  const oldVersion = await getVersion(config);
+
+  // sorry if this try/catch breaks your stack trace :'(
```
arcanis: No need for this comment, rethrowing should keep the callstack more or less right

arcanis:
What would you think about reading the `package.json` file as a string before modifying it, and restoring it as-is the version steps fails? The problem I see with the current flow is that the file will still end up being modified if something bad happens (like keys might be moved, indentation lost, etc).

src/cli/commands/publish.js
qfox: Cause of habit, but you're right. In this case we don't want to change the stack trace at all.

qfox:
@arcanis I considered that. This approach reused mechanics already used in the flow (namely updating the version). Reading and restoring the entire package.json may introduce new vectors and I didn't want to introduce them. But of course we can just as easy do that...

arcanis:
I think it would be better - for example, one particular problem there is that the file will be restored with a `0.0.0` version if it didn't had one to begin with.

arcanis:
Btw to fix the linting error you should just have to run `yarn prettier` in the root of the repository :)

qfox:
This PR is flawed because it uses `setVersion` but doesn't take into account that the version bump may be git committed. See the issue for more details why a fix is non-trivial.

PR #4687, opened 2017-10-11
fix(resolver): Fix incorrect peer dependency res. from different trees

### LINKED ISSUE #4539: Theia builds are weirdly broken with npm 5.3.0 + yarn 1.1.0
Theia builds are failing with weird errors from tslint with npm 5.3.0 + yarn 1.1.0: https://travis-ci.org/theia-ide/theia/jobs/279443045

With yarn 1.0.2 it is fine: https://travis-ci.org/theia-ide/theia/jobs/279458465

BYK:
**Summary**

Fixes #4539. Yarn was resolving peer dependencies from the closest level where the peer dependency was requested
but it was not checking if the peer dependency was in the same subtree. This was causing incorrect
peer dependency resolutions and package duplication when  an unrelated subtree has a depedency
satisfying the required peer dependency at the same tree level.

**Test plan**

Added new install integration test that fails without the fix.


src/package-linker.js
```diff
@@ -446,6 +446,24 @@ export default class PackageLinker {
       }
       const ref = pkg._reference;
       invariant(ref, 'Package reference is missing');
+      // TODO: We are taking the "shortest" ref tree but there may be multiple ref trees with the same length
+      const refTree = ref.requests.map(req => req.parentNames).sort((arr1, arr2) => arr1.length - arr2.length)[0];
+      const refTreeLen = refTree.length;
```
arcanis: I'd just use `refTree.length`, it looks like a premature optim (same for `pkgParentsLen`)

src/package-linker.js
```diff
@@ -446,6 +446,24 @@ export default class PackageLinker {
       }
       const ref = pkg._reference;
       invariant(ref, 'Package reference is missing');
+      // TODO: We are taking the "shortest" ref tree but there may be multiple ref trees with the same length
```
arcanis: Is it a problem if we choose one of those at random?

BYK: I honestly don't know, hence the TODO.

src/package-linker.js
BYK: It was mostly for ease of reading for myself but sure.

BYK: Oh well, these were easier on the eyes in a previous version of the code. Now they are useless as you pointed out.

src/package-linker.js
```diff
@@ -459,16 +477,16 @@ export default class PackageLinker {
           if (!(peerPkgRef && peerPkgRef.patterns)) {
             continue;
           }
-          const levelDistance = ref.level - peerPkgRef.level;
-          if (levelDistance >= 0 && levelDistance < resolvedLevelDistance) {
+          const levelDistance = getLevelDistance(peerPkgRef);
+          if (isFinite(levelDistance) && levelDistance < resolvedLevelDistance) {
```
kaylieEB: Is it possible to not set `minDistance` to Infinity and just check for existence now?

src/package-linker.js
```diff
@@ -446,6 +446,24 @@ export default class PackageLinker {
...
+      const refTreeLen = refTree.length;
+
+      const getLevelDistance = pkgRef => {
+        let minDistance = Infinity;
+        for (const req of pkgRef.requests) {
+          const pkgParents = req.parentNames;
+          const pkgParentsLen = pkgParents.length;
+          const distance = refTreeLen - pkgParentsLen;
+
+          if (distance >= 0 && distance < minDistance && pkgParents.every((name, idx) => name === refTree[idx])) {
```
kaylieEB: Might be more less readable, but more efficient to use `!pakgParents.some` instead?

__tests__/commands/install/integration.js
```diff
@@ -55,6 +55,15 @@ test.concurrent('install should not hoist packages above their peer dependencies
   });
 });
 
+test.concurrent('install should resolve peer dependencies from same subtrees', async () => {
+  await runInstall({}, 'peer-dep-same-subtree', async (config): Promise<void> => {
+    expect(JSON.parse(await fs.readFile(`${config.cwd}/node_modules/d/node_modules/a/package.json`)).version).toEqual(
+      '1.1.0',
```
kaylieEB: shouldn't this equal `1.0.0` since d points to `file:../a-1.0.0` as a dependency? and the top level version of a equal `1.1.0`?

src/package-linker.js
```diff
@@ -446,6 +446,21 @@ export default class PackageLinker {
...
       invariant(ref, 'Package reference is missing');
+      // TODO: We are taking the "shortest" ref tree but there may be multiple ref trees with the same length
+      const refTree = ref.requests.map(req => req.parentNames).sort((arr1, arr2) => arr1.length - arr2.length)[0];
+
+      const getLevelDistance = pkgRef => {
+        let minDistance = Infinity;
+        for (const req of pkgRef.requests) {
+          const distance = refTree.length - req.parentNames.length;
+
+          if (distance >= 0 && distance < minDistance && req.parentNames.every((name, idx) => name === refTree[idx])) {
```
kaylieEB: It might be slightly less readable but more efficient to use `!some` to exit early if they don't match

__tests__/commands/install/integration.js
BYK: Yup, that's why the test was failing :)

src/package-linker.js
BYK: I think both `every` and `some` do early exits so don't think it would change perf.

src/package-linker.js
BYK: I wanted that too but then I'd need to add special casing into the `getLevelDistance` code above due to the `distance >= 0 && distance < minDistance` checks. Is it too confusing right now? :(

src/package-linker.js
kaylieEB: ugh, I keep forgetting that sorry!

src/package-linker.js
kaylieEB: Ah I see, hmm can't think of an alternative atm. Thanks for explaining. :)

PR #4689, opened 2017-10-11
Support `--peer` for install to install peer deps as well

### LINKED ISSUE #1503: Add a means to install package peer dependencies for development / testing
Feature

N/A

Provide a CLI command `yarn install --peer` which will install peer dependencies specified in `package.json`. That way development / testing can use the peers such as react/ng2/grunt.

qfox:
**Summary**

This fixes #1503
This also drops some legacy options for help output (which were moved to `add` and don't work at all under `install`)

**Test plan**

Have a package with `peerDepdendencies`, install with `yarn install --peer` and confirm that the packages listed under `peerDepdendencies` are also installed.
Test cases added.


qfox:
Wasn't sure how to add a test for mock packages. Would like some input on that. I'd prefer not to use actual packages as I only want to see that the peer packages were or weren't included.

arcanis:
I think I like it, but what do you think about making this a default behavior rather than something hidden behind a flag? Since it only affects the top-level package (which shouldn't have peer dependencies anyway, unless it's a plugin of something in which case it would make sense to install the peer dependencies), that shouldn't change a lot of things.

The only things I can see being affected would be people using `yarn link` (because then the linked plugin would use the official release instead of the one that linked it), but then they would just have to run `yarn link` in their plugin as well to setup the cross-link.

Would probably require a major bump, but I think it's worth being considered.

arcanis:
As for the tests, you can setup a fixture that contains sub-packages and reference them through the `file:` protocol.

src/cli/commands/install.js
```diff
@@ -917,13 +928,8 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
 
 export function setFlags(commander: Object) {
   commander.usage('install [flags]');
+  commander.option('--peer', 'explicitly install peer dependencies as well');
   commander.option('-g, --global', 'DEPRECATED');
-  commander.option('-S, --save', 'DEPRECATED - save package to your `dependencies`');
```
BYK: Why remove these? They are deprecated but removing them is a breaking change and we don't want to release a major version just yet.

src/cli/commands/install.js
```diff
@@ -934,16 +940,10 @@ export async function install(config: Config, reporter: Reporter, flags: Object,
...
-  if (flags.lockfile === false) {
-    lockfile = new Lockfile();
-  } else {
-    lockfile = await Lockfile.fromDirectory(config.lockfileFolder, reporter);
-  }
-
   if (args.length) {
-    const exampleArgs = args.slice();
+    let command = 'add';
+    let error = 'installCommandRenamed';
```
BYK: Is this change relevant to the new feature?

src/cli/commands/install.js
```diff
@@ -344,6 +351,9 @@ export class Install {
             pushDeps('dependencies', workspaceManifest, {hint: null, optional: false}, true);
             pushDeps('devDependencies', workspaceManifest, {hint: 'dev', optional: false}, !this.config.production);
             pushDeps('optionalDependencies', workspaceManifest, {hint: 'optional', optional: true}, true);
+            if (this.flags.includePeerDeps) {
```
BYK: @arcanis I feel like this would change the lockfile resolution based on the flag which is not something we want.

I feel like we should drop the if block and use the following line:
```
pushDeps('peerDependencies', workspaceManifest, {hint: 'peer', optional: false}, this.flags.includePeerDeps);
```

Even then, I feel like this will be a big change in the final resolution. What do you think?

qfox:
@arcanis 

> but what do you think about making this a default behavior

As you say yourself, the whole thing doesn't happen too often. The flag pleases a small subset of people who explicitly want to install the peer deps as well. As I understand it, this is normally not the case for most people. Making it a default behavior could actually be detrimental since you may be installing something local which is already available globally, and now you have two versions to keep to update.

If making this a default behavior may also affect other cases like `link` then that seems like all the more reasons not to make this a default behavior.

Additionally, if the use case is so minor then a major version bump seems definitely unwarranted. 

I think the flag is fine.

I'll have another look at the test.

src/cli/commands/install.js
qfox: That's a good point. Whether it's a valid point is probably going to be a plain decision by the product owner. 

Either we support this and running the command will invalidate the cache if used inconsistently (with and without the flag). The user may complain about the cache being broken, or it may not notice it at all. We could print a warning for this if that's desirable.

Or we don't support this feature. In that case we should mention a wontfix in the ticket, perhaps state the reason why, and close it.

I don't think it's worth a major version bump for making this a default behavior (see https://github.com/yarnpkg/yarn/pull/4689#issuecomment-336170821)

src/cli/commands/install.js
qfox: They aren't deprecated so much as simply invalid. They won't do anything anymore so why report them as deprecated? 

The `--peer` option was going to be re-used in light of this and the other options are only valid in `add`, unless I'm mistaken?

I left `-g` because I think that still makes sense with `install`. Although I think that one is also moot?

arcanis: Removing deprecated features doesn't require a major per semver. I don't have a problem removing them, since it's been pretty clear for quite a long time now that `yarn install` isn't meant to add dependencies.

src/cli/commands/install.js
qfox: No. Guess I'll remove the refactor.

src/cli/commands/install.js
qfox: Wait, I thought these options _did not work_ at all with `install`. That's why I removed their listing. If they still function then I stand corrected and they should remain until the feature itself is removed.

(Note that this only removes the `--help` output, not the actual features they once represented...)

qfox:
Updated with proper tests and sans the refactoring.

Still dropping the flags since I think they can't be used without add. So it's not a breaking change to remove the help messages, merely cleanup.

src/cli/commands/install.js
BYK: >Removing deprecated features doesn't require a major per semver. I don't have a problem removing them, since it's been pretty clear for quite a long time now that yarn install isn't meant to add dependencies.

I think this is still a hard-breaking change. I'll ask React guys to see what they do. I think they start deprecating in feature updates and only remove them in major releases which makes sense.

>Wait, I thought these options did not work at all with install. That's why I removed their listing. If they still function then I stand corrected and they should remain until the feature itself is removed.

The feature is always there. These are just legacy flags from NPM but AFAIK they do work.

src/cli/commands/install.js
arcanis: @BYK I think there is a precedent for this (namely, the `--flat` option). But that's something that would be solved if peer dependencies were installed by default rather than via a flag (which I still think would make sense).

src/cli/commands/install.js
arcanis: They don't work per-se, but generate a 'suggested command line'.

https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L951-L975

@qfox Can you remove their removal from this PR? I think we should focus on the peer dependency itself, removing the deprecated commands will be a separate PR :)

arcanis:
cc @qfox ? do you think you'll have the bandwidth to rebase this PR and make the few changes requested ?

qfox:
@arcanis I don't think I have the necessary bw to push this further. I remember from discussions in our team that this is a more complex and contentious feature to work on than I have time to flesh out.

If somebody else feels this is important and wants to finish where I left of then please feel free to do so. You can use my code, I'll keep the branch up.

I am closing this PR since I won't see myself working on it any further any time soon. If this feature is important to you please take my code, rebase it, apply the changes as suggested above, and polish it. Thanks :)

PR #4692, opened 2017-10-12
Only show `--ignore-scripts` for `install` command

### LINKED ISSUE #2735: `--ignore-scripts` is ignored
bug

 yarn [command] --ignore-scripts does not ignore scripts

`package.json`:
```
{
  "scripts" : {
    "prebuild": "echo prebuild",
    "build": "echo build"
  }
}
```

```
$ yarn run build --ignore-scripts
yarn run v0.20.3
$ echo prebuild
prebuild
$ echo build
build
✨  Done in 0.25s.
```
```
$ yarn run build --ignore-scripts
yarn run v0.20.3
$ echo build
build
warning Ignored scripts due to flag.
✨  Done in 0.25s.
```

| what | version |
| --- | --- |
|`node -v`| 6.9.1|
|`yarn -V`|0.20.3|
|`os-version`:| OS X 10.11.6|

qfox:
Fixes #2735

**Summary**

Moves the help line for `--ignore-scripts` from generic to just `install` command.

**Test plan**

Check output for `bin/yarn --help run` and `bin/yarn --help install` and confirm that the switch only shows up for `install`.

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -181,6 +181,16 @@ test.concurrent('should run add command with help option', async () => {
   expectHelpOutputAsSubcommand(stdout);
 });
 
+test.concurrent('should not show --ignore-scripts for `run` command', async () => {
+  const stdout = await execCommand('--help', ['run'], 'run-help');
+  expect(stdout.join('\n')).not.toContain('--ignore-scripts');
+});
+
+test.concurrent('should show --ignore-scripts for `install` command', async () => {
+  const stdout = await execCommand('--help', ['install'], 'run-help');
+  expect(stdout.join('\n')).toContain('--ignore-scripts');
+});
+
 test.concurrent('should run add command with h option', async () => {
   const stdout = await execCommand('add', ['-h'], 'run-help');
   expectHelpOutputAsSubcommand(stdout);
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -917,6 +917,7 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
 
 export function setFlags(commander: Object) {
   commander.usage('install [flags]');
+  commander.option('--ignore-scripts', "don't run lifecycle scripts");
   commander.option('-g, --global', 'DEPRECATED');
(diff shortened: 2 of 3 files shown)


arcanis:
Does it work if you call `yarn add foobar --ignore-scripts`? I'm not sure we have a test for this

qfox:
I just followed the ticket which explicitly mentioned this option only to work for `install`. 

Quoting yourself: 

> 'The documentation lists --ignore-scripts in the yarn install command. The yarn run command has no such option (it has no option at all, actually), so the behaviour you're looking for just doesn't exists."

qfox:
Updated with proper tests and sans the refactoring. 

Still dropping the flags since I think they can't be used without `add`.  So it's not a breaking change to remove the help messages, merely cleanup.

qfox:
Wrong ticket. Deleting previous comment.

arcanis:
There's a difference, since `yarn add` does make an installation, when `yarn run` only runs a script. Documentation could be clearer, but basically the idea is that some commands "inherit" their options from others when it makes sense.

qfox:
Okay. Which commands inherit `--ignore-scripts` and ought to? Which commands should output this in their help? I don't think it's `run` but apart from `add`, which others? I'll add them and we can merge this. We should add the help messages per command specifically. But that's a different change.

BYK:
>Okay. Which commands inherit --ignore-scripts and ought to? Which commands should output this in their help? I don't think it's run but apart from add, which others? I'll add them and we can merge this. We should add the help messages per command specifically. But that's a different change.

There is also `publish` I think.

arcanis:
From the top of my head, those are the likely candidates:

- `add`
- `global`
- `import`
- `remove`
- `upgrade-interactive`
- `upgrade`

arcanis:
ping @qfox ?

qfox:
I'm going to close this PR. If you feel this is helpful for you, feel free to pick it up from where I left it. I'll leave the branch in my account for you to continue on.

To repeat: closing because I won't work on this. Feel free to take what is there and finish it if this is important to you.

Issue #1538, opened 2016-10-28
Installing phantomjs-prebuilt fails due to no npm_config_tmp


joshkel:
**Do you want to request a _feature_ or report a _bug_?**

Bug (maybe - Yarn's behavior is not necessarily incorrect, but it's incompatible with NPM)

**What is the current behavior?**

`yarn add phantomjs` fails with the following error:

```
error /home/jkelley/t2/node_modules/phantomjs: Command failed.
Exit code: 1
Command: sh
Arguments: -c node install.js
Directory: /home/jkelley/test/node_modules/phantomjs
Output:
PhantomJS not found on PATH
Phantom installation failed [TypeError: Path must be a string. Received undefined] TypeError: Path must be a string. Received undefined
    at assertPath (path.js:8:11)
    at Object.posix.join (path.js:479:5)
    at findSuitableTempDirectory (/home/jkelley/t2/node_modules/phantomjs/install.js:127:30)
    at /home/jkelley/test/node_modules/phantomjs/install.js:476:19
    at nextTickCallback (/home/jkelley/t2/node_modules/kew/kew.js:47:28)
    at nextTickCallbackWith0Args (node.js:420:9)
    at process._tickCallback (node.js:349:13)
```

Investigation revealed that the problem is that phantomjs 2.1.7 [unconditionally expected](https://github.com/Medium/phantomjs/blob/v2.1.7/install.js#L121) the `npm_config_tmp` environment variable to be passed down by NPM. Yarn doesn't do this, so Yarn fails to install phantomjs in cases where NPM succeeds.

This is probably a bug in the phantomjs package (the [current phantomjs-prebuilt package](https://github.com/Medium/phantomjs/blob/e0d2e6188a324a1287586f06dc403e62321ed556/install.js#L108) fixes it), but phantomjs is a popular package and probably not the only package to make this mistake, so it would improve Yarn's compatibility with NPM to pass `npm_config_tmp`.

A simple workaround for phantomjs in particular is to set TMPDIR before calling yarn.

**If the current behavior is a bug, please provide the steps to reproduce.**

`yarn add phantomjs`

**What is the expected behavior?**

`yarn add phantomjs` works.

**Please mention your node.js, yarn and operating system version.**

Node 4.6.1, Yarn 0.16.1, Ubuntu 14.04

pgporada:
Is there any traction on this from yarn upstream?

kevingelion:
[...]

PR #4693, opened 2017-10-12
Allow ${HOME} in .npmrc to work in Windows

joshkel:
Windows doesn't set the HOME environment variable by default, but NPM has logic to set process.env.HOME based on the current OS home directory.  [See here.](https://github.com/npm/npm/blob/fb28e5868a9dbbe21a15f23fe8cf8b3703e8adf2/lib/config/defaults.js#L81)

Yarn doesn't do this, so an `.npmrc` config such as `prefix =
${HOME}/.npm-packages` that work for NPM will cause Yarn to refuse to run; it instead exits with the following error:

```
yarn install v1.2.0
error An unexpected error occurred: "Failed to replace env in config: ${HOME}".
info If you think this is a bug, please open a bug report with the information provided in "C:\\src\\yarn\\yarn-error.log".
```

This commit updates Yarn's behavior to be closer to NPM's.

Unlike NPM, this commit only sets HOME if it's not already set, to avoid potentially incompatible changes with existing Yarn users.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -19,7 +19,7 @@ import * as constants from '../constants.js';
 import * as network from '../util/network.js';
 import {MessageError} from '../errors.js';
 import Config from '../config.js';
-import {getRcConfigForCwd, getRcArgs} from '../rc.js';
+import {prepareEnv, getRcConfigForCwd, getRcArgs} from '../rc.js';
 import {spawnp, forkp} from '../util/child.js';
 import {version} from '../util/yarn-version.js';
 import handleSignals from '../util/signal-handler.js';
@@ -537,6 +537,7 @@ export function main({
 }
 
 async function start(): Promise<void> {
+  prepareEnv();
   const rc = getRcConfigForCwd(process.cwd());
   const yarnPath = rc['yarn-path'];
 
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -19,7 +19,7 @@ import * as constants from '../constants.js';
 import * as network from '../util/network.js';
 import {MessageError} from '../errors.js';
 import Config from '../config.js';
-import {prepareEnv, getRcConfigForCwd, getRcArgs} from '../rc.js';
+import {getRcConfigForCwd, getRcArgs} from '../rc.js';
 import {spawnp, forkp} from '../util/child.js';
 import {version} from '../util/yarn-version.js';
 import handleSignals from '../util/signal-handler.js';
@@ -537,7 +537,6 @@ export function main({
 }
 
 async function start(): Promise<void> {
-  prepareEnv();
   const rc = getRcConfigForCwd(process.cwd());
   const yarnPath = rc['yarn-path'];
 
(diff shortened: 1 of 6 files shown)


src/rc.js
```diff
@@ -10,6 +10,14 @@ import * as rcUtil from './util/rc.js';
 // Keys that will get resolved relative to the path of the rc file they belong to
 const PATH_KEYS = new Set(['yarn-path', 'cache-folder', 'global-folder', 'modules-folder', 'cwd']);
 
+// prepares environment variables for processing rc
+export function prepareEnv() {
+  // Allow ${HOME} in, e.g., .npmrc to work even on Windows.
+  if (rcUtil.home && !process.env.HOME) {
+    process.env.HOME = rcUtil.home;
```
arcanis: I don't like this. We shouldn't overwrite environment variables and, if we really must, we shouldn't do it in this file.

arcanis:
Why don't you just use `~`?

src/cli/index.js
```diff
@@ -537,6 +537,7 @@ export function main({
 }
 
 async function start(): Promise<void> {
+  prepareEnv();
```
BYK: I think this should be done in `getRcConfigForCwd` instead of changing the whole `process.env`. I don't think it is a good idea to patch/mutate `process.env`

joshkel: Normally, I'd agree. In this case, though, I figured that setting HOME only if it's not already set should be safe (the fact that NPM does it could be evidence for its safety?), and, since Yarn is promoted as a drop-in replacement for NPM, I figured there's benefit to doing what NPM does. (My thinking here is probably colored by my running into #1538.)

I'm happy to go either way, though, so I can change the PR this evening to use `getRcConfigForCwd`.

Thanks for the feedback!

joshkel:
> Why don't you just use `~`?

Because I didn't realize I could. 🙂 (I think I copied that config from an online source when I was first getting familiar with NPM and never thought about it again.) Thanks for the suggestion.

src/cli/index.js
joshkel: Upon further investigation, I don't think that `getRcConfigForCwd` is correct. That finds config files but doesn't process them. The environment variable changes instead need to be made within `NpmRegistry`. I just pushed an updated PR with these changes.

BYK:
@joshkel sorry for the wait didn't realize you've updated the PR. Will review it shortly.

src/util/env-replace.js
```diff
@@ -1,7 +1,9 @@
 /* @flow */
 const ENV_EXPR = /(\\*)\$\{([^}]+)\}/g;
 
-export default function envReplace(value: string, env: {[key: string]: ?string} = process.env): string {
+export type Env = {[key: string]: ?string};
```
BYK: ❤️

src/registries/npm-registry.js
```diff
@@ -206,11 +207,20 @@ export default class NpmRegistry extends Registry {
     return actuals;
   }
 
+  static getConfigEnv(env: Env = process.env): Env {
+    // To match NPM's behavior, HOME is always the user's home directory.
+    const overrideEnv = {
```
BYK: protip:
```
return {
  ...env,
  HOME: home,
};
```

BYK:
I'll let @arcanis do the final review and merge if there are no blockers.

PR #4708, opened 2017-10-13
fix(config): Always output JSON format in `yarn config list`

### LINKED ISSUE #4496: Inconsistent config strings
Continuation from #4457 

Potential bug.

Yarn config has different formatting for object keys that are computed / set vs defaults.

I'm not sure I understand if this is by design or not. If so, what is the use case?

screen shot 2017-09-17 at 5 15 06 pm

Keys should all be wrapped in single quotes or DEFAULTS should follow same formatting as `registry`, `lastUpdateCheck`, and user configured values do

Yarn: 1.0.2
Node: 8.1.4

sreeramjayan:
**Summary**

Fixes #4496.

Used `JSON.stringify` to ensure the output of yarn config list displays JSON.

**Test plan**

New integration test to validate JSON output.


arcanis:
Can you share a pic of before/after?

src/cli/commands/config.js
```diff
@@ -83,10 +83,10 @@ export const {run, setFlags, examples} = buildSubCommands('config', {
     }
 
     reporter.info(reporter.lang('configYarn'));
-    reporter.inspect(config.registries.yarn.config);
+    reporter.inspect(JSON.stringify(config.registries.yarn.config, null, 2));
```
BYK: I think you should use `reporter.log` at this point with the `{force: true}` option.

BYK:
Hey, thanks a lot for the PR!

A test case would be great if you can add it. A simple integration test in `__tests__/integration.js` would be sufficient to make sure the output is always valid JSON.

Also having a proper title for the PR would be great. We use the following format:
```
fix(config): Always output JSON format in `yarn config list`

**Summary**

Fixes #4496. Use `JSON.serialize` instead of `reporter.inspect` to ensure the output of the config object is always a valid JSON string.

**Test plan**

Added new integration test that checks JSON output.
```

BYK:
@sreeramjayan let me know if you need any help with the tests!

sreeramjayan:
@BYK I did make the requested changes but am facing an issue coming up with the test case.
https://github.com/sreeramjayan/yarn/commit/462afb2de70c90a85ba892eaa423c643068ef788

The issue seems to be that the output of running yarn config list isn't JSON since we get the results of the yarn and npm config.
https://github.com/sreeramjayan/yarn/blob/462afb2de70c90a85ba892eaa423c643068ef788/__tests__/integration.js#L420
 I was wondering how the output can be validated to be JSON.

BYK:
Sorry for the late response and thanks a lot for the updates!

>The issue seems to be that the output of running yarn config list isn't JSON since we get the results of the yarn and npm config.

I think you can get a "pure" JSON string if you run `yarn --silent config list` ;)

sreeramjayan:
@arcanis Here are the before and after screen shots. 
Before:
before

After:
after

As you mentioned the formatting is off when the string is serialized.

sreeramjayan:
@arcanis Here are the before and after screen shots. 
Before:
before


After:
after


As you mentioned the formatting is off when the string is serialized.

arcanis:

Issue #4388, opened 2017-09-11
File dependencies from transitive dependencies are stored incorrectly in lock file


joscha:
**What is the current behavior?**

When having a file dependency that has more file dependencies, these second-level dependencies are referenced in cache, e.g.:
```json
{
  name: "x"
  dependencies: {
    "a": "file:../../a"
  }
}
```
with:
```
{
  name: "a",
  dependencies: {
    "b": "../b"
  }
```
would result in a lock entry like:
```
 "a@file:../../a":
   dependencies:
     "b" "file:../../../../../Library/Caches/Yarn/v1/npm-b"
```
instead of (yarn 0.27.5):
```
 "a@file:../../a":
   dependencies:
     "b" "file:../b"
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Create two chains of dependencies, see: https://github.com/joscha/yarn-101-links especially https://github.com/joscha/yarn-101-links/blob/master/my-package/yarn.lock#L8

**What is the expected behavior?**
Have relative links only, so the whole `node_modules` folder can be gzipped.

**Please mention your node.js, yarn and operating system version.**
node: 8.4.0
yarn 1.0.1
OSX Sierra

code0100fun:
I am seeing this too but it appears that yarn 1.0.1 running on a linux box doesn't have a problem with an OSX `file:../../../../../Library/Caches/Yarn/v1/...` reference. Running install on linux does produce a yarn.lock diff where the caches have been updated to `file:../../../../../.cache/yarn/v1/...`. This doesn't break my CI but it seems like it will cause issues for some people.

ZeeCoder:
We have this issue too.
I don't think it broke anything so far, but yarn.lock changes tend to show up in every MR now.

BYK:
@ZeeCoder which version of Yarn are you using?

ZeeCoder:
1.0.2, had the same with 1.0.1

jan-j:
I'm having same issue with version 1.1.0

joscha:
Still happening on 1.2.1 - this is blocking any upgrade to 1.x - is there any way to escalate this issue?

joscha:
I created a PR with a failing test: https://github.com/yarnpkg/yarn/pull/4717 - still unsure on where the fix has to be made.

joscha:
is there anyone here that would be able to have a look at #4717 and maybe give me a hint on where the fix needs to be?

joscha:
[...]

PR #4717, opened 2017-10-16
test(integration): Adds assertion to exemplify wrong path resolution in lock file

joscha:
The added assertions make sure that the relative path generated in the lock file points to the
installation, not the yarn cache

This exemplifies the problem outlined in #4388:

1__yarn_test_--_--watch__node__and_integration_js_ _yarn


joscha:
Error seems to be produced here: https://github.com/yarnpkg/yarn/blob/4c38ca7771922d5d56dac91f4be6e69a4d8d7885/src/config.js#L412 where the cache path is used as a starting point to generate an absolute path

BYK:
Hi @joscha! Thanks for the PR! Obviously, we cannot merge this since it breaks CI :)

Are you planning to submit a fix for the problem? If not can we have this test split up and skip the failing part so we can keep the code and the CI?

joscha:
Hi @BYK - I opened the PR to illustrate the problem outlined in #4388 in a test. This is currently blocking us from upgrading from `0.27.5` to `1.x`, so I would love to provide a fix. Unfortunately, I dug around in the code and can't perfectly pinpoint where the fix will need to be made, as there are a few places where it could be, but I think something more general needs correction than a simple path fix, hence my comment https://github.com/yarnpkg/yarn/pull/4717#issuecomment-336775218 - do you know anyone I could have a short chat with to find out in which module the fix would live?

BYK:
@joscha - sorry for the late responses. Unfortunately, PRs are not the best place to have quick conversations anymore 😄 Come to our [Discord Channels](https://discord.gg/yarnpkg) anytime and ping me or anyone really to initiate a chat. I feel like this issue is somewhat related to #3453. 

I'll do some more digging and come back to you. Thanks a lot for sticking with us although us not responding very quickly :)

BYK:
@joscha I've pushed some changes to your branch and did some hacky things to fix the issue but I think we need a larger refactoring here.

@rally25rs @arcanis @kaylieEB  - Please have a look at the new code now. It also fixes #3453. I think this is super-hacky but maybe it is an "OK" start and we can start fixing these intermingled parts?

BYK:
Woo, I've broken so many other tests now :D

src/fetchers/copy-fetcher.js
```diff
@@ -5,11 +5,10 @@ import BaseFetcher from './base-fetcher.js';
 import * as fs from '../util/fs.js';
 
 export default class CopyFetcher extends BaseFetcher {
-  async _fetch(): Promise<FetchedOverride> {
-    await fs.copy(this.reference, this.dest, this.reporter);
```
rally25rs: I don't see this copy anywhere else, just removed. Was it not needed? Is something else performing the copy now?

src/config.js
```diff
@@ -395,6 +395,10 @@ export default class Config {
       return pkg.location;
     }
 
+    if (pkg.remote.type === 'copy') {
```
rally25rs: The couple `pkg.remote.type === 'copy'` checks feel like an implementation detail that should be left to the `remote`. (leaky abstraction?) Perhaps this module path creation should be part of the `remote` and not part of `config` (config feels like a weird place for it anyway).

src/fetchers/copy-fetcher.js
BYK: Since "file" dependencies are already on the disk, they are now copied at the linking stage, directly from the source instead of from the cache folder.

src/config.js
BYK: Totally agree. I think we need to extend the `remote` information to allow "directly available" packages. This may become useful for other things in the future but I don't know how directly. Right now it feels specific to `link:` and `file:` protocols so I was unsure about extending this data structure or not.

rally25rs: `link-resolver` has a type of `link` not `copy`, but `tarball-resolver` also has a type of `copy` so will also be affected by this change. Not sure if that is OK or not...

rally25rs: Looks like `tarball-resolver` can also report a type of `copy`. Not sure if that is OK or not...

```
    // set remote so it can be "fetched"
    pkgJson._remote = {
      type: 'copy',
      resolved: `${url}#${hash}`,
      hash,
      registry,
      reference: dest,
    };
```

joscha:
\o/

joscha:
@BYK any chance to get this in?

joscha:
@BYK is the general gist of the changes in this PR what we want to do? If yes, I can try and make tests go green.

BYK:
@joscha I guess so. I feel like this requires a bit more refactoring around how we use cache and assume it is there so not sure if it would be fun times for you to figure that out. That said if you really want to try, who am I to stop an enthusiastic soul? 😁

joscha:
> That said if you really want to try, who am I to stop an enthusiastic soul?

😆  haha, hence my question if the general approach is fine, because looking at the code right now, I couldn't definitely say it is. Are there user stories somewhere, e.g. `When I have a local file:-dependency, then I expect the files to be copied into the node modules folder and the yarn cache not to be populated`, etc.?

BYK:
>When I have a local file:-dependency, then I expect the files to be copied into the node modules folder and the yarn cache not to be populated, etc.?

I don't think there's a user scenario but since `file:` dependencies are local, unless you have used `yarn link` they should never have been copied to cache in the first place if you ask me. This is what I mean with larger refactoring. Right now, Yarn tries to load everything from the cache folder with the exception of links. That most probably needs to change but it didn't look too easy when I last tried.

joscha:
I suppose it might have been done for multiple runs of `install`, e.g. a
local dependency should probably stay stable if you just do another
`install` even if the folder of the target dependency has been touched,
e.g.:

X is v1
Y depends on file:..X with yarb.lock X=1
Install in Y (X in Y/node_modules is at 1)
X gets touched and bumped to 2
Install in Y (X in Y/node_modules should stay at 1)
Delete Y/node_modules
Install in Y (X is at 2, yarn.lock of Y is updated with X=2)

Which means that there needs to be a copy of X which isn't touched (unless
that one copy is the one in Y/node_modules?)
It's hard to say what is right without having a user story reasoning about
it.

BYK:
@joscha I feel like it is fine to sync/copy at all times. That would be a divergence from the old npm behavior but would align with the new npm behavior where `file:` is treated as `link:` since I think that's what the developers expect anyway. In that case, bypassing the cache makes sense. Even if we want to use the cache, we need to namespace these packages so they don't collide with things from other registries etc.

joscha:
This seems to be fixed in `1.5.1`. I propose to get the tests in, to ensure it won't be broken in the future by accident.

arcanis:
@joscha Do you mind rebasing the PR? I'll see to merge it!

joscha:
rebased @arcanis

joscha:
ah, I am not sure what I am seeing locally with `1.5.1` - seems this problem still exists in `master`...

joscha:
Just tested again locally. I am unsure what the hiccup was, but this problem still exists in 1.5.1 unfortunately.

tsdorsey:
Forgive me if this is unrelated but another consideration when tackling this issue is that yarn will convert absolute paths in package.json to relative ones in yarn.lock. In my use case, that breaks things (which I can expound upon if need be but I do not think relavant). I would expect that a local dependency (absolute or relative path) should not be altered by the yarn.lock generation process.

joscha:
@arcanis I fixed the merge commit up, you can now see the problem again.

PR #4720, opened 2017-10-16
docs(linker): Use breadcrumbs in peer dependency warnings

BYK:
**Summary**

@gaearon called Yarn out for better peer dependency error messages so this is the first step to make
them more useful and informative.

**Test plan**

Use a repo like `https://github.com/edmorley/yarn-unmet-peerdependency-testcase` to have peer dependency
warnings and observe they now show the path for the package missing the peer dependency. This should
allow people to investigate these errors in an easier fashion.

```
warning "neutrino-preset-airbnb-base > eslint-config-airbnb-base@12.0.2" has unmet peer dependency "eslint@^4.8.0".
warning "neutrino-preset-airbnb-base > eslint-plugin-import@2.7.0" has unmet peer dependency "eslint@2.x - 4.x".
```

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -468,7 +468,7 @@ export default class PackageLinker {
 
         let peerError = 'unmetPeer';
         let resolvedLevelDistance = Infinity;
-        let resolvedPeerPkgPattern;
+        let resolvedPeerPkg;
         for (const peerPkg of peerPkgs) {
           const peerPkgRef = peerPkg._reference;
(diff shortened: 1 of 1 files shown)


arcanis:
Looks good - have you checked if it works if you run "yarn" in the root of a package that expects peer dependencies?

BYK:
>Looks good - have you checked if it works if you run "yarn" in the root of a package that expects peer dependencies?

Good point. I'll check and come back

PR #4726, opened 2017-10-17
quickwin for failed fetch of missing fsevents

javatheist:
fsevents is an optional dependency. Although platforms such as linux ignore the failed fetch, I'm having a situation that Yarn still wants to resolve the version due to conflicting consumer plugins and eventually adds it as a resolution to package.json (and to the lock file).
If you want to re-run yarn install, Yarn insists on checking the fsevents entry at "resolutions" without having the information that it's just optional.
My approach is a bit more generic that just "is it fsevents" and simply checks whether the config holds a location for the current item. If falsy, pretend that it's optional.

**Summary**

**Test plan**

### DIFF
--- a/src/package-compatibility.js
+++ b/src/package-compatibility.js
@@ -106,7 +106,8 @@ export function checkOne(info: Manifest, config: Config, ignoreEngines: boolean)
     const ref = info._reference;
     invariant(ref, 'expected package reference');
 
-    if (ref.optional) {
+    // no location -> no previous fetch, possibly optional
+    if (ref.optional || !info.location) {
       ref.ignore = true;
       ref.incompatible = true;
 


BYK:
Thanks a lot for taking the time to address this! Looks like you have a good start but I'm a bit concerned about this change masking potential real issues. My question is why `fsevents` is not being marked as optional anyway?

Also, I think we need to support this change with tests. Paging @arcanis :)

javatheist:
> My question is why fsevents is not being marked as optional anyway?

Well, fsevents is marked as optional in the consumer plugin. When installing it for the first time, just the "warning info" appears and the installation completes. The problem is that fsevents is being used at two different consumer plugins which request different versions of fsevents. So, although fsevents is not being installed, the package.json resolutions section gets the fsevent entry with the expected version.
So, when I need to update something, yarn tries to resolve all dependencies being found in the resolutions section. As fsevents has not been installed formerly, there is no information that it has to be treated as optional. Thus, it throws an error at "checkOne". Simple as that :)
My feeling is that if location is null, it's a good indication that it was optional - there is no other possible workflow for location === null.
Another solution would be to mark optional plugins with a question mark at the resolutions section (ex.g. "fsevents?": "1.1.2") and parse it either at checkOne or filter it on fetching the info collection. Anyway, I'm not trying to be smart here. It's just my 2ct and I count on you guys :)

I hope I could add some value to this issue.

BYK:
Hi! Apologies for the very late response and thanks a lot for the detailed explanation.


>My feeling is that if location is null, it's a good indication that it was optional - there is no other possible workflow for location === null.

Since I don't know the code that well, I'm a bit concerned about this assumption (fact?). @arcanis @bestander @kaylieEB can you also check? With the comment in the code, I think this is fine. The only thing is we still need to resolve optional depenedencies and record them into the lockfile to be deterministic. I feel like this changes that and thus a breaking change (and also a change we cannot accept).

>Another solution would be to mark optional plugins with a question mark at the resolutions section (ex.g. "fsevents?": "1.1.2") and parse it either at checkOne or filter it on fetching the info collection. Anyway, I'm not trying to be smart here. It's just my 2ct and I count on you guys :)

Is this an issue with offline installs or selective version resolutions? Ping @kaylieEB in case it is only a resolutions issue.

kaylieEB:
hey @javatheist thanks for the PR!

I'm a little worried as well about skipping all checks for packages without location. A quick search through the code base seems to indicate that packages that don't have location just means it's not yet installed in node_modules, so it could be any type of dependency.

Did you try running install with `--ignore-platform` or `--ignore-engines` to see if you can get past the initial check errors?

> Yarn insists on checking the fsevents entry at "resolutions"

Do you mean at the package resolution step or is fsevents entered as a resolutions field of package.json?

arcanis:
I agree with both @BYK and @kaylieEB here, this assumption looks dangerous to me:

> My feeling is that if location is null, it's a good indication that it was optional - there is no other possible workflow for location === null

It might be true right now (and frankly I'm not entirely sure it is, the reference objects / info data are sometimes populated in quite unpredictable ways, especially with the asynchronous parallel loading), but it looks like the kind of thing that might not be true in the future.

javatheist:
Hey @all , sry for the late response, but - believe it or not - I was offline during my vacation :)

Please be assured that all your opinions and remarks sound very valid to me. As a matter of fact, this does not change the urge to fix this issue somehow. As there seem to be loads of question marks around, let me be even more specific by using a package.json file, where this issue arises. Pls find the file attached. I didn't change it in order to be as authentic as possible.

1) On executing "yarn", you will be asked to choose a fsevents version. On doing so, everything completes successfully. Now, you will see that the fsevents version is fixed at the bottom of the package.json [resolutions] section.

2) Now re-trigger "yarn" and it will fail to fetch packages with "error fsevents@1.1.2: The platform "linux" is incompatible with this module."

As far as I'm concerned, this happens due to the fact that yarn simply wants to check all package versions listed at [resolutions] and fsevents is not recognized as optional anymore. So if you (understandably) refrain from checking location === null, there should be another mechanism to check optional packages which are listed at the package.json [resolutions] section. That's why I proposed to use question marks such as "fsevents?": "1.1.2" there instead of location === null.

I hope I was able to clear it up now. In case you need more input, pls don't hesitate to shoot your questions. As always, thank you for your efforts!

BYK:
>As far as I'm concerned, this happens due to the fact that yarn simply wants to check all package versions listed at [resolutions] and fsevents is not recognized as optional anymore.

This part especially, but the whole response was very explanatory. Thanks a lot! I think now the ball is in @kaylieEB's court since she implemented the resolutions field and I feel like she can come up with a good solution to this issue :)

Thanks a lot again and I'm glad you were able to stay offline during your vacation! 😁

BYK:
@javatheist  - oh btw. we've changed our CI configuration and unfortunately, it requires a merge from master for things to pass in PRs. Would you mind doing a merge from the latest master? It should be a clean merge anyway :)

kaylieEB:
@javatheist One workaround you can use is to remove `flat: true` from your package.json. Technically flat mode is legacy version of resolutions, and that's what's causing it to prompt fsevents to be resolved and added to package.json.

javatheist:
@kaylieEB Thank you for your suggestion to remove the flat option. If flat mode is legacy, does it mean that yarn is completely dropping this feature? This would be horrible news for our project. which is depending on SystemJS and lazy/dynamic/imperative assets loading etc.

arcanis:
The flat mode isn't legacy per se (in that we haven't deprecated it and don't currently plan to do it), so don't be scared to use it. It definitely has some shortcomings that we would like to address later on, but if it happens it will be a separate independent feature.

Note that Yarn now ships with a feature called 'selective resolution override' (built by @kaylieEB, actually!) that I think can be a interesting replacement for the flat mode: instead of telling Yarn to flatten *everything*, you tell it to install everything normally, then to optionally flatten some dependencies by forcing your packages versions to resolve to a specific one. By doing this, you'll be able to avoid the annoying conflict resolution prompts for the modules you don't really care about (like `left-pad`, or `lodash`, or whatever), while still being able to select a unique version for your most important third-party dependencies (like `react`, `angular, ...).

kaylieEB:
@javatheist sorry if I caused any confusion, @arcanis got it on this one. I only meant that selective resolutions feature was written to be capable of doing what flat mode does but in a more granular fashion. So by removing all the `flatMode: true` option in the package.json and listing all of the resolutions that you want, you should be able to achieve the same effect and install all dependencies except `fsevents`.

arcanis:
Given that this PR was a bit controversial at the time and that we came up with workarounds, I'm going to close it for now. Thanks anyway for taking the time to help us, and feel free to open a new PR if you think this is something that we really should have!

Issue #2010, opened 2016-11-23
Error if running yarn with stderr redirected


pascalberger:
**What is the current behavior?**
Calling yarn on PowerShell with stderr redirected to a file results in an error message:

```
  TypeError: Expected `count` to be a positive finite number
      at module.exports (C:\Program Files (x86)\Yarn\node_modules\repeating\index.js:10:9)
      at ProgressBar.render (C:\Program Files (x86)\Yarn\lib\reporters\console\progress-bar.js:58:22)
      at ConsoleReporter.progress (C:\Program Files (x86)\Yarn\lib\reporters\console\console-reporter.js:388:9)
      at C:\Program Files (x86)\Yarn\lib\package-fetcher.js:123:36
      at Generator.next (<anonymous>)
      at step (C:\Program Files (x86)\Yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:17:30)
      at C:\Program Files (x86)\Yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:35:14
      at F (C:\Program Files (x86)\Yarn\node_modules\core-js\library\modules\_export.js:35:28)
      at C:\Program Files (x86)\Yarn\node_modules\babel-runtime\helpers\asyncToGenerator.js:14:12
      at PackageFetcher.init (C:\Program Files (x86)\Yarn\lib\package-fetcher.js:158:7)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
* Call `yarn install somepackage 2>c:\somefile.txt`

**What is the expected behavior?**
Yarn should also work if stderr is redirected to a file

**Please mention your node.js, yarn and operating system version.**
* Yarn version: 0.17.6
* Node version: 7.2.0
* Platform: win32 x64

pascalberger:
A workaround is using `yarn config set no-progress true` which is available since 0.19.1 (see #1962)

nwholloway:
@BYK PR #4742 wouldn't address this.

I was able to reproduce with the v0.17.6 code, but you don't get the reported error in current yarn.

This is because the commit 864b563074 switched to use `string.prototype.repeat`, and that is amenable to being passed `NaN`.

There is some tidy-up to do, as yarn is looking at process.stdout to see if it is a TTY, but then writing the progress bar to process.stderr (so terminal control codes get written to a file).

bulgariamitko:
I am also getting this error

`Arguments: 
  /usr/local/bin/node /usr/local/bin/yarn

[...]

PR #4697, opened 2017-10-12
Fixes progress display on non-color tty (#4647)

nwholloway:
If the output does not support color, then each render of the progress bar is added to a single line, which wraps over multiple lines.

As a fallback, a simple carriage return is used to move to the start of the line, and space characters to clear the line.

**Summary**

This is a minor cosmetic output on the formatting of the progress output.

I came across the issue raised when looking for problems to address for Hacktoberfest.

**Test plan**

I verified by running `yarn check` to provide output with the progress bar under the following conditions:
* Interactive use with color output.
* Interactive use with terminal not supporting color output
* Non-interactive use (output was piped)
 
The output from the 3 tests can be seen in the following screenshot.

image

This screenshot shows the non-color test with the active progress bar.

image

PR #4742, opened 2017-10-19
fix(reporters): Handle process.stdout.columns being zero

nwholloway:
**Summary**

My recent pull request was to improve the appearance of the progress bar for non-color terminals (PR #4697).

However, @skevy reported a RangeError when running with macOS 10.12, with Node 8.6.  This would have been caused by process.stdout.columns returning a negative number.

In this case, this just assumes a default width of 100 characters (as in spinner-progress.js).

**Test plan**

I have not been able to reproduce the condition where `process.tty.columns` returns a negative number, so have verified the logic by considering key cases, e.g., `undefined > 0`, `-1 > 0`.

### DIFF
--- a/src/reporters/console/util.js
+++ b/src/reporters/console/util.js
@@ -12,7 +12,10 @@ const CLEAR_RIGHT_OF_CURSOR = 1;
 export function clearLine(stdout: Stdout) {
   if (!supportsColor) {
     if (stdout instanceof tty.WriteStream) {
-      stdout.write(`\r${' '.repeat(stdout.columns - 1)}\r`);
+      if (stdout.columns > 0) {
+        stdout.write(`\r${' '.repeat(stdout.columns - 1)}`);
+      }
+      stdout.write(`\r`);
     }
     return;
   }


src/reporters/console/util.js
```diff
@@ -12,7 +12,8 @@ const CLEAR_RIGHT_OF_CURSOR = 1;
 export function clearLine(stdout: Stdout) {
   if (!supportsColor) {
     if (stdout instanceof tty.WriteStream) {
-      stdout.write(`\r${' '.repeat(stdout.columns - 1)}\r`);
+      const columns = stdout.columns > 0 ? stdout.columns : 100;
```
arcanis: I think rather than being negative, `stdout.columns` equals 0. Then when you make the -1, it becomes negative. Considering that, wouldn't it be better to just update the condition to just be `stdout instanceof tty.WriteStream && stdout.columns > 0`?

BYK: I think this `100` number should be a constant in `constants.js` file.

Also a comment above this line about why we have this check would be great.

nwholloway:
I'll take a look at incorporating changes to address comments.

Thanks for the feedback.

BYK:
@nwholloway let us know if you need any help or just when you are ready!

nwholloway:
@arcanis  Thanks for the comment pointing out the problem was the column being reported as 0, rather than -1 -- this makes much more sense, and is something I can replicate.

Instead of overwriting the line with an arbitrary number of spaces, I now skip the clearing of the line when the column width is unknown.  Generally, the next output would overwrite the short progress bar.

The screenshots show the progress bar output is short, and then the final output.

screenshot_2017-10-28_18-24-25

screenshot_2017-10-28_18-21-20

@BYK I did look at adding Jest tests for the behaviour.  I started with tests for the util.js methods `clearLine` and `toStartOfLine`, but the replacement of '\r' by '\n' in the snapshot means the sense of it is largely lost.

BYK:
This PR should also fix #2010.  @nwholloway are you willing to test if this PR actually fixes #2010 too? I'll try to find a way to write tests for this.

BYK:
>@BYK I did look at adding Jest tests for the behaviour. I started with tests for the util.js methods clearLine and toStartOfLine, but the replacement of '\r' by '\n' in the snapshot means the sense of it is largely lost.

😢  Alright, may be next time :)

PR #4750, opened 2017-10-20
fix(install): Rebuild native modules when node version changes

### LINKED ISSUE #756: Rebuilding native modules

Feature!

https://yarnpkg.com/en/docs/migrating-from-npm does not list a way to rebuild.

We (in [nteract](https://github.com/nteract/nteract)) end up needing to rebuild one native module, zeromq. Normally we run this:

```
npm rebuild zeromq --runtime=electron --target=1.4.5 --disturl=https://atom.io/download/atom-shell --build-from-source
```

Our use case is specifically with Electron; I'm guessing there are others that use `npm rebuild`.

```
$ node --version
v6.7.0
```

```
$ yarn --version
0.15.1
```

OS X 10.11.6

/cc @jdetle @captainsafia @lgeiger

romanschejbal:
**Summary**

Fixes #756. We have multiple versions of our app and each one uses a different version of node. 
Therefore we need to rebuild our `node-sass` module every time we move from one to another. 

This PR addresses that by saving the NODE version those artifacts were built with within the `.yarn-integrity` file and triggers forced scripts install (only if the node version is different ofc).

**Test plan**

```
1. Install Node.js 7.x
2. Add the node-sass dependency to the project via Yarn
3. Update Node.js to 8.x (new NODE_VERSION)
4. Run "yarn install" (you should see yarn downloading fresh scripts/binaries)
```

### DIFF
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -424,6 +424,12 @@ export class Install {
       return false;
     }
 
+    if (match.nodeVersionDoesntMatch) {
+      // node version doesn't match, force script installations
+      this.scripts.setForce(true);
+      return false;
+    }
+
     if (!patterns.length && !match.integrityFileMissing) {
       this.reporter.success(this.reporter.lang('nothingToInstall'));
       await this.createEmptyManifestFolders();
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -424,8 +424,8 @@ export class Install {
       return false;
     }
 
-    if (match.nodeVersionDoesntMatch) {
-      // node version doesn't match, force script installations
+    if (match.hardRefreshRequired) {
+      // e.g. node version doesn't match, force script installations
       this.scripts.setForce(true);
       return false;
     }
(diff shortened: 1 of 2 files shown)


romanschejbal:
@arcanis sure and done. Thanks for reviewing 👍  😊

aahoughton:
Maybe I'm something obvious, but I don't believe this solves *my* use-case for #756 -- build and package on one architecture, deploy and rebuild on another.  I'm not worried about NODE_VERSION changing.

"Bad idea," sure, but npm handles this well.

romanschejbal:
@aahoughton what if Yarn stored the arch. as well and _rebuild_ upon changes? Would that solve the issue you're having? Anyway, for now I'd wait for the next release and create a separate issue describing the different architectures problem (if that's still a thing) - feel free to ping me then and I'll give it a go.

Any thoughts on this ^ anyone - i.e. is that the right direction?

aahoughton:
That would solve my issue, yes -- I'd just need to insure that the appropriate files are built into my deb.  This is effectively the same logic as the version rebuild trigger, yes?

aahoughton:
poke.

Issue #4735, opened 2017-10-18
`upgrade-interactive --latest` suggests to downgrade dependency 


ipanasenko:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
when running `yarn upgrade-interactive --latest` it suggests to install version 1.0.1

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn add react-refetch@1.0.3-0
yarn upgrade-interactive --latest
```

**What is the expected behavior?**
I think it shouldn't suggest to downgrade 

**Please mention your node.js, yarn and operating system version.**
mac os high sierra
node 8.7.0
yarn 1.2.1

edmorley:
The most recent stable release of `react-refetch` is 1.0.1, with the `1.0.3-0` release being on the `beta` tag:
```
$ npm view react-refetch dist-tags
{ latest: '1.0.1', beta: '1.0.3-0' }
```

Whilst it seems logical for `upgrade-interactive --latest` to not suggest upgrading to the beta release (if one is on an older release), I agree that perhaps it should not **downgrade** if already on the higher version beta.

rally25rs:
I think this was my fault; I had some code in `getOutdatedPackages` that would ignore filtering out packages that actually needed an upgrade, and just return them all when not upgrading a specific package. I believe that code is no longer needed and can be removed. I'll make a PR...

PR #4756, opened 2017-10-21
Fix for #4735 Ignore `returnAllPackages` param in upgrade

kaylieEB:
**Summary**
Fixes https://github.com/yarnpkg/yarn/issues/4735

Upgrade-interactive uses upgrade's `getOutdated` method, which previously called package-request's `getOutdatedPackages` method with `returnAllPackages` when package name wasn't being specified. This was unnecessary since `getOutdatedPackages` [already checks for filter patterns](https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L356-L363).

It also inadvertently caused a bug where it would list packages that weren't outdated. For example, `react-refetch@1.0.3-0` was being correctly filtered [here](https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L405), but because all dependencies were being returned, it was added back as an outdated package [here](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade.js#L218).

**Test plan**
I've verified that the `returnAllPackages` param was only being used in upgrade file. All existing tests related to upgrade should pass. Will try and see if I can add more.

*BEFORE*

screen shot 2017-10-21 at 12 25 51 pm

*AFTER*

screen shot 2017-10-21 at 12 26 11 pm

### DIFF
--- a/src/cli/commands/upgrade.js
+++ b/src/cli/commands/upgrade.js
@@ -199,7 +199,6 @@ export async function getOutdated(
 ): Promise<Array<Dependency>> {
   const install = new Install(flags, config, reporter, lockfile);
   const outdatedFieldName = flags.latest ? 'latest' : 'wanted';
-  const updateAll = patterns.length === 0;
 
   // ensure scope is of the form `@scope/`
   const normalizeScope = function() {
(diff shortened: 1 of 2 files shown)


kaylieEB:
closing as a duplicate of https://github.com/yarnpkg/yarn/pull/4763/commits

Issue #3202, opened 2017-04-20
`yarn upgrade` breaks dependencies


chinesedfan:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

`yarn upgrade` updates the version of npmlog, but does not check its dependencies.

**If the current behavior is a bug, please provide the steps to reproduce.**

Suppose you have a very simple package.json file like following:
```
{
  "dependencies": {
    "npm": "^3.10.5"
  }
}
```
Run `yarn`. You will find the folder structure is like:
```
node_modules
├─npm@3.10.10
│  ├─npmlog@4.0.0
│  │  ├─gauge@2.6.0
```
Then `yarn upgrade`.
```
node_modules
├─npm@3.10.10
│  ├─npmlog@4.0.2 // it is updated, but it requires "gauge": "~2.7.1"
│  │  ├─gauge@2.6.0
```

**What is the expected behavior?**

Not only update the version of npmlog, but also inner dependencies like gauge.

**Please mention your node.js, yarn and operating system version.**

node v6.9.1, OSX 0.11.6

OmerHerera:
I think all these issues are the same, all of them talking about sub-dependencies not updated:
[3115](https://github.com/yarnpkg/yarn/issues/3115)
[3082](https://github.com/yarnpkg/yarn/issues/3082)
[2394](https://github.com/yarnpkg/yarn/issues/2394)

We need to find a solution.

Thanks

gossi:
A workaround is to `rm -rf node_modules/` and `yarn install` again. The workaround also may help once... I also did this:

`yarn install gauge -D`

however this forces you to `yarn upgrade gauge` EVERYTIME after using yarn... and if you forget it you can't go back to your regular workflow.

li-xinyang:
Re-install entire `node_modules` does solve the issue. I wonder why no everyone has this issue. Is this because it only affects certain packages?

prewk:
FYI: just deleting `node_modules/npm` followed by a `yarn install` is sufficient to make it work correctly again. No need to delete the whole `node_modules` dir.

bestander:
Can repro on 0.27.

arcanis:
The following testcase is completely broken. The dependency itself isn't upgraded until the node_modules is removed:

```
(
    mkdir -p pkg-a && cd pkg-a
[...]

PR #4757, opened 2017-10-23
fix(linker): Fix yarn removing linked deps during link stage

### LINKED ISSUE #3288: yarn overwrites symlinked dependencies

Bug

I have a development checkout of the package `@company/foo` which I made globally available through `yarn link`. I then linked that packages into my project `@company/project`.

I also have package `@company/bar`, which is linked identically into `@company/project`.

When I now run `yarn upgrade @company/bar` in the directory of `@company/project`, yarn will overwrite my local development checkout with data it pulled from the npm registry.

It should be reproducible with any combination of linked packages as outlined above. It seems to me that when yarn says "linking dependencies", it will simply pull some fixed version from the local cache and overwrite my local development snapshot.

The linked package is not overwritten. npm deals with this: https://github.com/npm/npm/commit/1619871ac0cc8839dc9962c78e736095976c1eb4

6.10.2, 0.23.3, Windows 10

### LINKED ISSUE #4603: `yarn install`1.1.0 overwrites symlinks created by `yarn link`
bug

When using `yarn link <local-package>` a symlink is created in `node_modules/`. After running `yarn install`, the symlink is removed with the public registry version of the local module.

https://github.com/javinor/yarn-removes-symlinks using yarn 1.1.0

symlinks should not be removed when running `yarn install` as happens in yarn v0.27.5

macOS Sierra
node 6.9.2
yarn 1.1.0 (doesn't happen with yarn 0.27.5)

### LINKED ISSUE #4635: `yarn add` is losing workspace symlinks
Using yarn 1.1.0, node 8.3.0:

```sh
ls node_modules # includes several symlinks from workspace modules
yarn add --dev jest
ls node_modules # symlinks are gone
```

There are no errors during the `yarn add`, it just seems to blow away workspace links.  Running `yarn` again restores them.

### LINKED ISSUE #4770: Running yarn link unlinks any modules linked via yarn link.
Yarn version 1.2.1.

Bug.

Running `yarn add [package-name]` unlinks any linked modules. Here is a screenshot showing this:

screen shot 2017-10-24 at 10 18 15 am

`cd app`
`ls -l node_modules | grep sequential` --> shows that the module is not linked
`yarn link sequential-ids`
`ls -l node_modules | grep sequential` --> shows that the module is now linked
`yarn add lodash`
`ls -l node_modules | grep sequential` --> shows that the module is now unlinked

`yarn add [package-name]` should never cause linked packages to become unlinked.

Node 8.6.0.

macOS High Sierra. This also happened on Sierra.

gandazgul:
**Summary**

 Actual fix: changed fs.readlink to fs.realpath when checking if a symlink is a linked dependency in package-linker.js This fixes yarn removing linked deps when installing or updating.

Fixes #3288, fixes #4770, fixes #4635, fixes #4603.

Potential fix for #3202.

**Test plan**

See https://github.com/yarnpkg/yarn/issues/3288#issuecomment-335955366 for repro steps.
See https://github.com/yarnpkg/yarn/issues/3288#issuecomment-338503103 for my explanation of the problem.

With a real world test scenario this works, but I'm unable to have it break from a unit test. I added a test in the integration suite but with the bug added back in it still passes because both generated paths are identical. I would like some help with the unit test.


gandazgul:
Note to myself, found a bug, yarn throws when a linked module's destination no longer exists, try catch?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "yarn",
   "installationMethod": "unknown",
-  "version": "1.2.1",
+  "version": "1.2.2",
```
bestander: this is unnecessary, the core team will bump the patch version when they are ready to release it

__tests__/fixtures/install/install-dont-overwrite-linked/.npmrc
```diff
@@ -0,0 +1 @@
+yarn-offline-mirror=./mirror-for-offline
```
bestander: this file is not needed for the test

src/package-linker.js
```diff
@@ -304,8 +304,15 @@ export default class PackageLinker {
       const stat = await fs.lstat(entryPath);
 
       if (stat.isSymbolicLink()) {
-        const packageName = entry;
-        linkTargets.set(packageName, await fs.readlink(entryPath));
+        try {
```
bestander: eslint errors on this.
Please run `yarn run prettier` to autoformat the code

__tests__/fixtures/install/install-dont-overwrite-linked/.npmrc
bestander: nevermind, I was wrong here

__tests__/commands/install/integration.js
```diff
@@ -865,6 +865,29 @@ test.concurrent('install will not overwrite files in symlinked scoped directorie
   );
 });
 
+test.concurrent('install will not overwrite linked dependencies', async (): Promise<void> => {
+  await runInstall(
+    {},
+    'install-dont-overwrite-linked',
+    async (config): Promise<void> => {
+      const dependencyPath = path.join(config.cwd, 'node_modules', 'fake-dependency');
```
bestander: I don't understand the test.
i would expect that you would run a link command and then install should not overwrite it?

bestander: A few tips.

0. You probably want to setup several folders for the test: a linked package.json and the linker package.json
1. here is how link command is used in tests https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/link.js#L35
2. after linking you can check that link exists
3. after that run install command and check that link was not destroyed
https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L93

package.json
gandazgul: I will revert this. Thank you.

src/package-linker.js
gandazgul: Will do, and will check with the linter before committing. Thanks.

__tests__/commands/install/integration.js
gandazgul: Thank you! Yeah I dont get this test either. I copied it from the other one that checks if scoped links are maintained, I suspect that one is faulty too. 

I'll update both with your suggestions here.

src/package-linker.js
gandazgul: Done. yarn lint completed.

gandazgul:
I re-wrote the tests with @bestander suggestions and tested them with the old code to make sure they break. Both tests for scoped and regular packages now fail correctly and pass correctly with the new code.

__tests__/commands/_helpers.js
```diff
@@ -113,11 +125,16 @@ export async function run<T, R>(
...
     const source = typeof name === 'string' ? name : name.source;
 
-    const dir = path.join(fixturesLoc, source);
-    cwd = await fs.makeTempDir(path.basename(dir));
-    await fs.copy(dir, cwd, reporter);
-    if (typeof name !== 'string') {
-      cwd = path.join(cwd, name.cwd);
+    // if source wasn't set then assume we were given a complete path
+    if (typeof source === 'undefined') {
+      cwd = typeof name !== 'string' ? name.cwd : await fs.makeTempDir();
```
BYK: These nested conditionals are a bit confusing. Do you think we can find a cleaner way to do this?

__tests__/commands/link.js
```diff
@@ -1,24 +1,12 @@
 /* @flow */
 
-import {run as buildRun} from './_helpers.js';
-import {run as link} from '../../src/cli/commands/link.js';
+import {runLink} from './_helpers.js';
```
BYK: Nice refactor.

src/package-linker.js
```diff
@@ -317,7 +322,13 @@ export default class PackageLinker {
 
           if (stat2.isSymbolicLink()) {
             const packageName = `${scopeName}/${entry2}`;
-            linkTargets.set(packageName, await fs.readlink(entryPath2));
+            try {
```
BYK: This repeated code kind of calls for a helper function.

BYK:
LGTM. @bestander approve and merge at will!

BYK:
>Fixes: #3288 and other reported issues relating to linked deps being removed.

Do we have a list for these?

src/package-linker.js
gandazgul: Yeah I was thinking about it. But I didn't want to confuse this too much. There's also a todo comment to merge the link targets logic here with logic in another part of the code. I can take a look after and make another PR.

gandazgul:
@BYK I found these: 

https://github.com/yarnpkg/yarn/issues/4770
https://github.com/yarnpkg/yarn/issues/4635
https://github.com/yarnpkg/yarn/issues/4603

__tests__/commands/_helpers.js
gandazgul: they are, if I take this one off which is unnecessary then flow complains. I will take a look in another PR see if I can can come up with a clearer logic and put more comments.

bestander:
Great job, @gandazgul!
Thanks for helping with review, @BYK

bertho-zero:
It works perfectly, npm has this problem since the beginning of version 3, thank you very much!

gandazgul:
@bestander thank you, happy to help. When do you think this will be released? My team is anxious to upgrade yarn to the latest version.

bestander:
@gandazgul it is due to be released any day soon according to recent release history.
In the mean time feel free to use nightly builds https://yarnpkg.com/en/docs/nightly

rdsedmundo:
This PR hasn't fixed this problem entirely. I'm still facing this issue.

NodeJS version: `10.0.0`
Yarn version: `1.6.0`

I do the `yarn link package` and I can confirm it works by executing `ls -l node_modules/package` and seeing the symlink pointing to the right folder. After I run `yarn install` though and execute the `ls -l node_modules/package` command again, the symlink is removed and it does use the package from what is on `package.json` again (it's a Git URL by the way, don't know if it makes difference).

Issue #1625, opened 2016-11-02
Yarn install Invalid protocol error


lambertkuang:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Running yarn on a new system gives me this error: 

```
Error: Invalid protocol: git+https:
      at Request.init (/usr/local/lib/node_modules/yarn/node_modules/request/request.js:459:31)
      at new Request (/usr/local/lib/node_modules/yarn/node_modules/request/request.js:128:8)
      at request (/usr/local/lib/node_modules/yarn/node_modules/request/index.js:54:10)
      at RequestManager.execute (/usr/local/lib/node_modules/yarn/lib/util/request-manager.js:327:17)
      at RequestManager.shiftQueue (/usr/local/lib/node_modules/yarn/lib/util/request-manager.js:353:10)
      at Promise (/usr/local/lib/node_modules/yarn/lib/util/request-manager.js:150:12)
      at RequestManager.request (/usr/local/lib/node_modules/yarn/lib/util/request-manager.js:148:21)
      at NpmRegistry.request (/usr/local/lib/node_modules/yarn/lib/registries/npm-registry.js:97:32)
      at TarballFetcher.fetchFromExternal (/usr/local/lib/node_modules/yarn/lib/fetchers/tarball-fetcher.js:176:21)
      at TarballFetcher._fetch (/usr/local/lib/node_modules/yarn/lib/fetchers/tarball-fetcher.js:218:19)
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Running yarn on a new system
**What is the expected behavior?**
Install successfully
**Please mention your node.js, yarn and operating system version.**
Node: 6.9.1
Yarn: 0.16.0
OS: CentOS 6.8

kittens:
Need more information to reproduce this like the `package.json` you're using.

lambertkuang:
```
{
  "name": "web",
  "version": "1.0.0",
  "description": "",
  "main": "app.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "node-dev app.js"
  },
  "repository": {
    "type": "git",
    "url": ""
  },
  "author": "",
  "bugs": {
    "url": ""
  },
  "homepage": "",
  "private": true,
  "dependencies": {
    "co": "^4.6.0",
    "cookie-parser": "^1.4.3",
    "lodash": "^4.15.0",
    "lokka": "^1.7.0",
[...]

PR #4759, opened 2017-10-23
fix(resolver): Add general support for git-over-protocol URLs

knuton:
**Summary**

Yarn can not handle the `git+https://` dependency format correctly, as described for various versions in #1625. The problem is present in Yarn 1.2.1.

A related problem for `git+ssh://` has been described in #573 and fixed in #3425.

This PR extends the solution from #3425 to use the Git fetcher for any [Git-over-protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols) URLs.

**Test plan**

Extended the `package-request` unit tests to verify that the correct remote type (git) is used for `git+https://`, while the tarball remote type continues to be used for regular HTTP(S) URLs.


knuton:
Nice, thanks for greenlighting.

Just realized that I checked in changes to `yarn.lock`. That's probably a mistake, shall I remove those?

BYK:
>Just realized that I checked in changes to yarn.lock. That's probably a mistake, shall I remove those?

I took care of that for you ;)

I feel like this logic should be merged with the one I worked on at #4657. What do you two think?

knuton:
Thanks @BYK!

Your question prompted me to browse the Yarn codebase for a moment. Now my feeling is that this check should happen earlier, that is `request.getLocked` should already be called with the correct `remoteType`.

I just looked at some traces and I'd bet that Yarn picks the wrong resolver for certain patterns. Concretely in the project that was affected by this, Yarn looks for a resolver for the pattern `multicast-dns@git+https://github.com/bnielsen1965/multicast-dns` and ends up using the npm registry resolver which calls `getLocked('tarball')`. My familiarity with Yarn internals is too slight to tell whether the problem is with the Git resolver not claiming this, or with a pattern of this shape even getting to that point.

PR #4763, opened 2017-10-23
fix(upgrade): Let outdated filter correctly when called from upgrade.

### LINKED ISSUE #4735: `upgrade-interactive --latest` suggests to downgrade dependency

rally25rs:
Fixes #4735 

Upgrade was calling outdated in a way that would not filter packages that did not need
an upgrade. The led to potential backdate of package if a beta version was in use. Yarn would
backdate to the "older" latest.

**Summary**

Some older upgrade code I had written had a code path to have `getOutdatedPackages()` return all packages, not just ones that were actually outdated. My belief had been that it wouldn't cause an issue because upgrading to the same version would just do nothing... but it turns out if you are on a beta version of a package, this would cause `upgrade --latest` to backdate you to the actual latest.

So for example if you were currently at beta `1.0.3-0` and the registry had `latest 1.0.1` then `yarn upgrade --latest` would backdate you to `1.0.1`.

This PR reinstates the filter that only returns versions if the current is less-than the latest, which would in the above example leave the user at `1.0.3-0`

**Test plan**

Added a unit test that would have failed previously, but passes with this PR change.


kaylieEB:
haha I had the same PR open! https://github.com/yarnpkg/yarn/pull/4756

src/package-request.js
```diff
@@ -404,7 +403,6 @@ export default class PackageRequest {
     const isDepOld = ({current, latest, wanted}) =>
       latest === 'exotic' || (latest !== 'exotic' && (semver.lt(current, wanted) || semver.lt(current, latest)));
```
kaylieEB: I thought this deserves a quick cleanup `latest === 'exotic' || (semver.lt(current, wanted) || semver.lt(current, latest));`

rally25rs: Oh geez I'm sorry... I should read the thread first huh :)
I'll review yours and kill mine???

kaylieEB: oh i don't mind it at all. Yours has a test so let's go with that! I can close mine :)

rally25rs: Actually, since mine has a unit test, should we keep mine? I'll pull in your change above too... Sorry for the churn... just physically got back from vacation, but my brain is still at the beach 🌴 ...

PR #4767, opened 2017-10-24
Fix wrong condition

ViceIce:
**Summary**
This PR fixes a wrong condition which cause the color fix  from  #3607  for #2724 to be no working as intended.

**Test plan**
This change should not influence any tests.


bestander:
Looks good, thanks!

PR #4779, opened 2017-10-25
feat(config): Allow JSON "//" comments in package.json resolutions

### LINKED ISSUE #4774: "//" keys should be ignored in "resolution" section of package.json

`"//"` keys (comments) in the `"resolution"` section are not ignored.

Steps to reproduce:
1. Add a key to the `"resolution"` section of a `package.json` like:

    ```
    {
      "resolution": {
        "//": "This should be a comment."
      }
    }
    ```

2. Run `yarn install`.

3. See warning:

    ```
    warning Resolution field "//" does not end with a valid package name and will be ignored
    ```

The key should be ignored. `"//"` keys are a standard way of commenting in `package.json` files in lieu of actual comments being available in JSON. (The `"resolution"` section in particular is a place where comments are necessary, so you can communicate why you're doing heavy-handed dependency tweaking.)

* Node v8.7.0
* Yarn v1.2.1
* macOS 10.12.6 (16G29)

rally25rs:
Fixes #4774

**Summary**

Previously package.json comments were being ignored for "dependencies",
"devDependencies", "optionalDependencies".

This change adds "resolutions" to the sections that will ignore
comments.

**Test Plan**

Added unit test to make sure warning is not printed for a comment in a
resolution.

### DIFF
--- a/src/util/normalize-manifest/fix.js
+++ b/src/util/normalize-manifest/fix.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import {DEPENDENCY_TYPES} from '../../constants';
+import {MANIFEST_FIELDS} from '../../constants';
 import type {Reporter} from '../../reporters/index.js';
 import {isValidLicense} from './util.js';
 import {normalizePerson, extractDescription} from './util.js';
@@ -308,7 +308,7 @@ export default (async function(
     }
   }
 
-  for (const dependencyType of ['resolutions', ...DEPENDENCY_TYPES]) {
+  for (const dependencyType of MANIFEST_FIELDS) {
     const dependencyList = info[dependencyType];
     if (dependencyList && typeof dependencyList === 'object') {
       delete dependencyList['//'];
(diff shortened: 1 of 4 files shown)


src/util/normalize-manifest/fix.js
```diff
@@ -308,7 +308,7 @@ export default (async function(
     }
   }
 
-  for (const dependencyType of DEPENDENCY_TYPES) {
+  for (const dependencyType of ['resolutions', ...DEPENDENCY_TYPES]) {
```
kaylieEB: what do you think about adding `resolutions` in the constants file 

and doing something like 
```
RESOLUTIONS = 'resolutions';
MANIFEST_FIELDS = [RESOLUTIONS, ...DEPENDENCY_TYPES];
```
then doing
```
for (const fields of MANIFEST_FIELDS) {
```
since resolution is not really a dependencyType? And we can add more fields to add comments to later if we want to?

rally25rs: Good call. I pushed up this change 👍

PR #4785, opened 2017-10-26
style(reporters): Simplify Windows bold fix check condition

BYK:
**Summary**

Used a single negation instead of double negation. Follow up to #4767.

**Test plan**

Existing tests


BYK:
>Fwiw, the current formatting looked more readable to me. I read it as "True if A is false or B is false", whereas I read the new one as "True if not A and B together", which is weird.

I think the intention was "Check if `TERM` doesn't start with `'xterm'` unless it is empty". I can see both ways so I can abandon this PR if you think the original version was clearer.

PR #4792, opened 2017-10-28
fix: Use wanted package repository URL and homepage from npm

### LINKED ISSUE #4650: yarn outdated does not show URLs

Bug

Running "yarn outdated" displays a lists of dependencies with no URLs

Not sure, yarn is installed with "brew install yarn", version is v1.1.0. Same v1.1.0 version on another machine works fine.

URLs would be visible.

macOS 10.13, Node.js v6.11., Yarn: v1.1.0

raido:
**Summary**

Fixes #4650.

Problem is that Verdaccio (and probably Sinopia2) both are affected as they do not return "repository" nor "homepage" in top level data structure.

**Test plan**

Added new test cases. Also, manually tested with and without https://github.com/verdaccio/verdaccio as registry.

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -349,3 +349,127 @@ describe('getPossibleConfigLocations', () => {
     );
   });
 });
+
+describe('checkOutdated functional test', () => {
+  const mockConfig = {
+    resolveConstraints(): string {
+      return '2.0.0';
+    },
+  };
+
+  test('homepage URL from top level', async () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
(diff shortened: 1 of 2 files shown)


BYK:
This looks like a good solution to me but I'd love to get @rally25rs' approval before merging.

Thanks a lot for the PR!

rally25rs:
Seems reasonable to me. Should be fall-back to the top-level config if it doesn't exist in the `versions`? Also, is there a way we can unit test this?

raido:
As I see overall the solution is approved, I’ll look into unit testing this.

Maybe another way around that it falls back to information in “versions” if not available in top level. This way it would return latest information usually and if required uses from “versions” which may be out of date if package X is far behind in a project and range does not allow major version upgrades.

raido:
@BYK I added unit test coverage + fallback logic, not sure if the best way to mock config for npm-resolver but it works.

PR #4798, opened 2017-10-29
Offline Install Fails due to Mismatch Between Github and Registry Version

### LINKED ISSUE #4710: Yarn Offline Install with GitHub Hash Fails
This is a bug.

I have a simple package.json with the following dependency:
```
"dependencies": {
    "mapbox-gl": "^0.22.0"
}
```
mapbox-gl in turn has the following problematic dependencies:
```
"dependencies": {
   [...]
    "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#de2ab007455aa2587c552694c68583f94c9f2747",
    "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#83b1a3e5837d785af582efd5ed1a212f2df6a4ae", 
}
```
With the usual offline .yarnrc:
```
yarn-offline-mirror "./offline-cache"
yarn-offline-mirror-pruning true
```

Then I run:
```
$ yarn install # works fine
$ rm -r node_modules
$ yarn install --offline
[1/4] Resolving packages...
error Can't make a request in offline mode ("https://github.com/mapbox/mapbox-gl-shaders")
```

I expect the offline-cache to contain two tgz archives containing the internal github packages. Instead I get the two hashes (de2ab007455aa2587c552694c68583f94c9f2747 & 83b1a3e5837d785af582efd5ed1a212f2df6a4ae) in the folder and `yarn install --offline` fails to find either package. I can manually change these hashes to tgz files and the yarn.lock with explicit version numbers to get it to work but this is far from ideal because it will reset when I add/remove/update a package.

Tried on Windows 7 / Debian, yarn v1.2.1, node v6.11.2

**Extra details**
Unlike #4266 and #4324, this problem disappears in later version where scoped packages are used (e.g. mapbox-gl: "^0.41.0"), but I need this earlier version of the project because of a parent dependency (plotly.js) requires it.

nwestbury:
**Summary**

Fixes #4710.

Some packages when requesting github hash versions of packages, have the 'github:' prefix whereas the package.json does not (e.g. mapbox/mapbox-gl-shaders in [mapbox-gl registry](https://registry.yarnpkg.com/mapbox-gl) vs [mapbox-gl package.json](https://raw.githubusercontent.com/mapbox/mapbox-gl-js/v0.22.1/package.json)). This mismatch in version causes it not to be found in the cache in offline mode.

**Proposed Solution**
This PR replaces the fetched dependencies to use the registry version (which is stored in the yarn.lock cache with the prefix) and ignores the package.json dependencies (no prefix). In most cases, I believe these dependencies *should* be identical but in rare cases like this one, they are different.

**Test plan**

`package.json`: 

    [...]
    "dependencies": {
        "mapbox-gl": "^0.22.0"
    }

`.yarnrc`:

    yarn-offline-mirror "./offline-cache"
    yarn-offline-mirror-pruning true

The following should now succeed with this PR:

    $ yarn install
    $ rm -r node_modules
    $ yarn install --offline

See the issue for more details.

Further tests to make sure this doesn't break anything else might be a good idea.


src/package-request.js
```diff
@@ -151,6 +151,14 @@ export default class PackageRequest {
   }
 
   /**
+   * Remove git protocol prefix in order to match the cache range name.
+   */
+
+  normalizeRegistryRange(range: string): string {
+    return range && range.replace(/^(bitbucket|github|gitlab):/, '');
```
BYK: I don't think we should remove the provider prefixes since they may be hosting completely unrelated repos. Moreover, even if we decide to do this, I think it should be done by the resolvers themselves so we don't repeat hard-coded protocols around. I also don't think regular expressions are a sustainable way to do this.

BYK:
@nwestbury I've merged from the latest master which should fix the build issues. Do you need anything from us? I've looked at the latest version of the patch but I'm not sure I understand why and how your fix works. Updating the summary with your diagnosis would really help :)

nwestbury:
@BYK Thanks for your response and feedback. Your enthusiasm and patience really help. Currently, I'm taking the npm registry response and replacing the dependencies instead of using the package.json (without github). I'm not convinced this is a satisfactory solution yet so feel free to disregard this change until I comment and update the description.

nwestbury:
@BYK What do you think of this approach? Does it make sense?

arcanis:
Hi @nwestbury! I'm really grateful for the work you put in this PR, but unfortunately the project moved forward since you submitted it and I fear it can be merged in this current state (we don't know if the problem is still there, or if the tests still pass). I'm going to close it, but please feel free to check if the problem still exists, and to rebase your work on top of the current master and open a new PR if it's the case. Again, sorry we didn't got around merging it sooner!

PR #4802, opened 2017-10-30
fix(cache): Adds a per-user cache fallback

arcanis:
**Summary**

When Yarn is being run with a different user than the one who created the cache folder, everything
might break (in particular if the cache folder has been created by the root user). This PR fixes this by using the user UID as part of the temporary folder name if possible. If not, it will fallback to the same directory than before.

**Test plan**

The change is small enough that it doesn't need an extra test. We already have a test that makes sure we correctly select a fallback, so adding a new one would only check that it has a specific name, which seems low-signal and prone to break.


src/constants.js
```diff
@@ -59,7 +59,11 @@ function getPreferredCacheDirectories(): Array<string> {
     preferredCacheDirectories.push(getDirectory('cache'));
   }
 
-  preferredCacheDirectories.push(path.join(os.tmpdir(), '.yarn-cache'));
+  if (process.getuid) {
+    preferredCacheDirectories.push(path.join(os.tmpdir(), `.yarn-cache-${process.getuid()}`));
+  }
```
BYK: At this point, it may be a better idea to make this an `if`/`else` block so we always use the user-specific temp fallback. I'm fine with this version too tho, so approving :)

PR #4803, opened 2017-10-30
docs(contributing): Add section about what to work on and triaging

BYK:
**Summary**

We have cleaned up our labels a bit and this PR adds some useful links for people to find what needs
attention from the community.

**Test plan**

Read and make sure the links and the text are accurate.


Daniel15:
This should be on the site rather than in the repo (https://yarnpkg.com/en/org/contributing) 😃

PR #4797, opened 2017-10-29
fix(info): Use version from `latest` dist-tag instead of the highest one

neonowy:
**Summary**

Fixes #3947. By default, package `version` was set by sorting all the versions and getting the highest one. Now it's provided via package `latest` dist-tag.

**Test plan**
##### Before:
```bash
yarn info ui-select | grep version: 
  version: '0.20.0',
```
##### After:
```bash
$ yarn info ui-select | grep version:
   version: '0.19.8',
```

I've also added a test case for it in `__tests__/commands/info.js`.
Although, I'm not sure if using a specific package like `ui-select` (from #3947) is a proper way to do it, because it'll break with the new release.

Could we add it to the `__tests__/fixtures/request-cache/`, so the package manifest won't change? Or do you have some other idea how can I write this test better?

PR #4804, opened 2017-10-30
fix(resolution): use registry latest if it satisfies requested semver range

### LINKED ISSUE #3560: Yarn installs wrong version

Report a *bug*

- put `"dependencies": { "css-tree": "^1.0.0-alpha17" }` in package.json
- exec `yarn`

`yarn.lock` is
```
css-tree@^1.0.0-alpha17:
  version "1.0.0-alpha9"
  resolved "https://registry.yarnpkg.com/css-tree/-/css-tree-1.0.0-alpha9.tgz#6ee3e397b569eef32419286c86e3e8b65fcaf3e3"
  dependencies:
    source-map "^0.5.3"
```

With NPM it's ok (NPM installs `1.0.0-alpha17`)

Resolve and install `1.0.0-alpha17` instead of `1.0.0-alpha9`

MacOS 10.12.4
Node: v8.0.0
NPM: 5.0.1
But this case reproduces and on earlier version of node and NPM

rally25rs:
**Summary**

Fixes #3560.
Related PR: #4797.

Mimic behavior in NPM; use the `latest` version in registry if it
satisfies the semver range requests.

Otherwise fallback to `semver.maxSatisfying()`

NPM source that specifies this behavior is: https://github.com/npm/npm/blob/d46015256941ddfff1463338e3e2f8f77624a1ff/lib/utils/pick-manifest-from-registry-metadata.js#L10-L14

I believe this is OK directly in the NPM resolver, as other resolvers may not have a "latest" tag or dist-tags in general (for example, I don't believe git-resolver has this metadata to check).

**Test Plan**

Added unit test to verify behavior.


__tests__/commands/add.js
```diff
@@ -986,3 +986,42 @@ test.concurrent('preserves unaffected bin links after adding to workspace packag
     expect(await fs.exists(`${config.cwd}/packages/workspace-2/node_modules/.bin/workspace-1`)).toEqual(true);
   });
 });
+
+test.concurrent('installs "latest" instead of maxSatisfying if it satisfies requested pattern', (): Promise<void> => {
+  /* eslint-disable */
```
BYK: Can you just disable the relevant warning with `/* eslint-disable rule-name */` instead of a blanket disable statement in case the enable part goes away?

Or ideally, we wouldn't need this at all since you can use https://git.io/ to get short GitHub URLs ;)

PR #4805, opened 2017-10-30
fix(git-spawn): Add custom Batch wrapper when using plink.exe (#4729)

### LINKED ISSUE #4729: Can't install package over git+ssh with non-standard port when using plink on Windows
Bug

A `plink: unknown option "-p"` error is shown.

```
λ yarn add git+ssh://git@github.com:22/substack/node-mkdirp.git
yarn add v1.2.1
info No lockfile found.
[1/4] Resolving packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads ssh://git@github.com:22/substack/node-mkdirp.git
Directory: C:\yarn-issue
Output:
plink: unknown option "-p"
```

Attempt to install a package over git+ssh, specifying a port number, on Windows with Git configured to use plink.

When researching other git+ssh issues, I found this command, which works: 
`yarn add git+ssh://git@github.com/substack/node-mkdirp.git`

But, when specifying a port number, it does not: 
`yarn add git+ssh://git@github.com:22/substack/node-mkdirp.git`

My company uses a private BitBucket server that runs over a non-standard port, so I'm not able to use Yarn with our private dependencies like I can with NPM, at least on Windows/plink.

Plink's documentation (https://the.earth.li/~sgtatham/putty/0.60/htmldoc/Chapter7.html) specifies that ports are specified by `-P`. If Yarn is using lowercase `-p` instead, that would explain the error.

Private dependencies on non-standard ports should be able to be installed with Windows/plink.

Node v6.9.5, Yarn 1.2.1, Windows 10

neonowy:
**Summary**

Fixes #4729.

The main issue is that we use `GIT_SSH_COMMAND` instead of `GIT_SSH`, as we have to pass `-batch` argument in order to use `ssh`/`plink` in the non-interactive mode. Unfortunately, Git converts `-p` to `-P` for `plink.exe` **only** when using `GIT_SSH`.

The only way I could find to get out of this tricky issue is to create a wrapper script responsible for the `-p`/`-P` conversion and use it when `plink.exe` is provided via `GIT_SSH`.

Similar issue was also described here: https://jira.atlassian.com/browse/SRCTREEWIN-5600?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

I'm not quite sure if the `bin/` directory is the best place for the wrapper script `yarn-plink.cmd`, would `scripts/` be a better choice?

**Test plan**

##### Before:
virtualbox_msedge_-_win10_30_10_2017_16_35_24

##### After:
virtualbox_msedge_-_win10_30_10_2017_16_31_42

I've also updated the test suite for `git-spawn.js` accordingly, but couldn't figure out how to test the `yarn-plink.cmd` wrapper script.

### DIFF
--- a/__tests__/util/git/git-spawn.js
+++ b/__tests__/util/git/git-spawn.js
@@ -42,6 +42,7 @@ describe('spawn', () => {
     process.env.GIT_SSH_COMMAND = '';
     // Test for case-sensitivity too (should be insensitive)
     const plinkPath = path.join('C:', 'pLink.EXE');
+    const plinkWrapperPath = path.join(__dirname, '..', '..', '..', 'bin', 'yarn-plink.cmd');
     process.env.GIT_SSH = plinkPath;
 
     const gitCall = runGit(['status']);
@@ -52,7 +53,7 @@ describe('spawn', () => {
     expect(gitCall[2].env).toMatchObject({
       GIT_ASKPASS: '',
       GIT_TERMINAL_PROMPT: 0,
-      GIT_SSH_COMMAND: `"${plinkPath}" -batch`,
+      GIT_SSH_COMMAND: `"${plinkWrapperPath}" "${plinkPath}" -batch`,
       ...process.env,
     });
   });
--- a/bin/yarn-plink.cmd
+++ b/bin/yarn-plink.cmd
@@ -1,0 +1,33 @@
+@echo off
+
+set PLINK_PATH="%~1"
+set PARAMS=""
+
+REM Shift params once at the beginning to exclude the first one
+shift
+
+REM Start of the loop
+:params_loop
+  if "%~1"=="" (
+    REM If we went through all the params, go to the end of the loop
+    goto after_params_loop
+  )
+
+  if "%~1"=="-p" (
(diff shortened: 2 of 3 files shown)


arcanis:
Unfortunately that won't work with the single-file releases, so I don't think we can merge it :(

neonowy:
@arcanis Oh, I didn't think of that. 😢  

But actually, I think I found another (even better) way of doing this! Apparently `GIT_SSH_VARIANT` variable was added to Git, so we can just set it properly and it should work.

BYK:
@neonowy are you sure it is not this line forcing everything to be lower case: https://github.com/yarnpkg/yarn/blob/6b185c23d8d7770a3dc7a9b872c18ad25a98a20c/src/util/git/git-spawn.js#L17

neonowy:
@BYK I've tested it and it doesn't make any difference. If I understand this code right, we're using `sshCommand` directly when setting `env.GIT_SSH_COMMAND` and `sshExecutable` is only used as a normalized key for `BATCH_MODE_ARGS` here.

neonowy:
Closing in favor of #4806.

PR #4810, opened 2017-10-30
fix(cli): Display warning when adding a package already in project

### LINKED ISSUE #3376: --dev/-D flag ignored if package is already listed under dependencies
Bug

If a package has already been added as a dependency, running `yarn add <pkg> -D` does not add it to `devDependencies`. The `-D` switch is ignored with no indication to the user that it's still only listed as a regular dependency.

```bash
yarn add <pkg>
cat package.json | less
yarn add <pkg> -D
cat package.json | less
```

I would expect the package to be either listed only under `devDependencies` or in both `dependencies` and `devDependencies`.

node v7.7.4
yarn 0.23.4
OS X 10.11.6

calvinhuang:
**Summary**

Fixes #3376 by displaying a warning to users when they try to add a package already listed in the manifest as a different dependency type. The behavior remains the same otherwise (package will remain in the original dependency type).

**Test plan**

New positive and negative tests for the warnings.


dawnmist:
This commit has broken `yarn upgrade` functionality. The upgrade command calls Add to do the actual upgrades, but because it is upgrading all dependency types at the same time and the Add script is validating against a single dependency type, any upgrades in 'devDependencies' get mistakenly flagged as invalid. The end result is that while yarn.lock gets updated, package.json does NOT get updated, and the upgrade gets left in an invalid state. See issue #4840.

calvinhuang:
I think I have a fix for this (by checking if `Add.config.commandName` is `"upgrade"` before printing the warning). I just to need to write some tests for it. Let me know if you see any problems with this fix: https://github.com/calvinhuang/yarn/commit/c12a7ffb0b2e8ff998f3efac1109125539b2603e

dawnmist:
Is it possible to check whether the commandName is "add" instead of is not "upgrade"? Upgrade-interactive also uses Add...I think locking the check to the command where it is desired rather than having to individually list all the exclusions is less likely to break.

Flygenring:
@dawnmist I'm really not sure that this breaks the `yarn upgrade` functionality at all; it warns that it's trying to do something wrong, but doesn't change the way the `add` command is called.

Judging from the code (https://github.com/yarnpkg/yarn/pull/4810/files#diff-52ee34799680d11ceefae45b77e6a72f) it would seem that `yarn add` will correctly add the dependency to the right section of the manifest based on what was specified by flags in the call (`this.flagToOrigin`) and thus the problem would be that `yarn upgrade` doesn't use the correct flags when calling `add`.

dawnmist:
@Flygenring `yarn upgrade` doesn't appear to pass the dependencies in one group at a time - it passes the entire set that need upgrading. Which means that it has multiple 'possible' flags in the same group - i.e. both dependencies and devDependencies are included in the same call to Add. It cannot set 'the right' flag, as it is upgrading multiple flags at once. This set of commits didn't take into account that Add was being used by Upgrade and Upgrade-Interactive, so assumed that the list of packages provided would always be of a single type.

The end result is that the package.json file does not get updated, though the yarn.lock does seem to. It leaves things in an inconsistent state. (I had a few other oddities occur at the same time - a bunch of packages with a "*" dependency suddenly decided to downgrade that dependency by a major version - but I couldn't be sure whether it was caused by this issue or was just a co-incidence. It completely ruined being able to compile though - it was the @types/react package that was downgraded from version 16 to version 15, resulting in the actual dependency of version 16 from our package.json file being installed plus every package with a "*" dependency on the react types installing their own version of the version 15 dependency. Mismatched type definitions = typescript cannot compile at all).

Flygenring:
@dawnmist True, there is a problem in how `yarn add` works or is being run, and I'm not arguing that there is not. It could be probably fixed by handling each group (or even each dependency) on its own and probably several other ways too.

Still I do believe that reporting the warning does not break anything, but merely reports the symptoms of something else being broken. From the code it definitely seems that it would make the exact same error when saving/updating dependencies in the manifest, only keeping quiet about it.

From looking a bit closer at the `add` code, it would seem that it is trying to properly handle this issue, but ultimately fails to do so?
https://github.com/calvinhuang/yarn/blob/3ed413cdda0f4d8f7b752ad4d857e36031b28a54/src/cli/commands/add.js#L182 (also line 190)

dawnmist:
@Flygenring The problem exists at line 199. https://github.com/calvinhuang/yarn/blob/3ed413cdda0f4d8f7b752ad4d857e36031b28a54/src/cli/commands/add.js#L199

The test for when to generate a warning does not take into account the resolutions that you pointed to, instead testing against an overall type that is guaranteed to be wrong during the upgrade/upgrade-interactive commands as used by upgrade/upgrade-interactive currently. Those commands could be altered to call Add multiple times (one for each possible package section) to work around the warning...but in the case of upgrades the user is not _installing_ new packages themselves so I don't believe that the warning is actually a valid warning to provide at all during upgrades. They are not attempting to change the package's section type, just to upgrade it in place.

Further testing today shows that the type issue I'd run into during the previous upgrade attempt isn't occurring anymore - so that part at least is not a part of this issue. I'm still not getting any of the upgrade information being written to the package.json file though, and I cannot myself see why a warning would block that from happening.

Flygenring:
@dawnmist I have to respectfully disagree; the test for whether or not a warning should be shown works exactly as it should (although it's slightly confusing to test `target` and print out `depType` in the warning, but at that point the two will have the same content).

That said, the test and warning has _nothing_ to do with the actual functionality of the `add` or `upgrade` commands as it _only_ tests if a warning should be shown, and if it should show a warning, it shows a warning and nothing more.

The problematic part is the way `add` is called from `upgrade` when `add` relies on the flag to determine which section to save/update the package information in in the manifest.  
Calling `add` once for each section of dependencies in the manifest with the proper section-flag should handle it just fine.

PR #4811, opened 2017-10-31
feat(cli): Disable only header/footer in nested commands, not all output

### LINKED ISSUE #4615: Nested yarn commands should still display all but version/timing output
In #3536, the handling of nested yarn commands was changed, such that the outer yarn invocation sets a `YARN_SILENT` environment variable, which causes the child yarn processes to behave the same as if `--silent` had been passed. The reasoning for this change was to reduce the noise from all of the `yarn <COMMAND> <VERSION>` and `Done in XXXs` output.

However the current implementation also suppresses useful output too, such as the output from `yarn outdated`.

An example use-case of when this is problematic, is when using the tool [oao](https://github.com/guigrpa/oao) (similar to yerna) to manage a monorepo's packages, and using oao's `outdated` command, which runs a child `yarn outdated` process for each package within the monorepo. 

Output before:
```
2017-10-03T13:16:19.232Z  main INFO  ┌── Outdated dependencies in neutrino [CREATED]
2017-10-03T13:16:19.509Z  yarn INFO  | yarn outdated v0.26.1
2017-10-03T13:16:20.208Z  yarn INFO  | Package   Current Wanted Latest     Package Type URL
2017-10-03T13:16:20.209Z  yarn INFO  | immutable 3.8.1   3.8.1  4.0.0-rc.4 dependencies https://facebook.github.com/immutable-js
2017-10-03T13:16:20.210Z  yarn INFO  | Done in 0.70s.
2017-10-03T13:16:20.241Z  yarn INFO  |
2017-10-03T13:16:20.245Z  main INFO  └── Outdated dependencies in neutrino [CLOSED]
2017-10-03T13:16:20.247Z  main INFO  ┌── Outdated dependencies in neutrino-middleware-banner [CREATED]
2017-10-03T13:16:20.518Z  yarn INFO  | yarn outdated v0.26.1
2017-10-03T13:16:20.857Z  yarn INFO  | Done in 0.34s.
2017-10-03T13:16:20.868Z  yarn INFO  |
2017-10-03T13:16:20.871Z  main INFO  └── Outdated dependencies in neutrino-middleware-banner [CLOSED]
```

Output after #3536:
```
2017-10-03T13:16:31.331Z  main INFO  ┌── Outdated dependencies in neutrino [CREATED]
2017-10-03T13:16:32.318Z  yarn INFO  |
2017-10-03T13:16:32.321Z  main INFO  └── Outdated dependencies in neutrino [CLOSED]
2017-10-03T13:16:32.323Z  main INFO  ┌── Outdated dependencies in neutrino-middleware-banner [CREATED]
2017-10-03T13:16:32.950Z  yarn INFO  |
2017-10-03T13:16:32.953Z  main INFO  └── Outdated dependencies in neutrino-middleware-banner [CLOSED]
```

Now oao could try and work around this breaking change by setting `YARN_SILENT=0` before creating the child process, however I don't think callers should have to know to do that. When would someone ever call `yarn outdated` and not want to see the "meat" (ie non header/footer) of the output?

This appears to have affected a number of other people too: #3671, #4423, #3767, #3771.

I think it might be best to:
* revert #3536 for now
* devise an alternate fix for that use case, that suppresses just the header/footer rows, and not all yarn output entirely

Thoughts? :-)

CC @kittens, @MalcolmDwyer, @borekb, @EddiG,

sth:
**Summary**

Fixes #4615. Disabling all Yarn output in nested commands with `YARN_SILENT` is a bit much, we usually want to see the output. This pull request introduces a new environment variable `YARN_WRAP_OUTPUT` that can be set to `0` to disable the header and footer Yarn normally displays.

Disabling the header/footer might also be useful in other situations, like other tools calling Yarn, so the `YARN_WRAP_OUTPUT` variable has general use.

**Test plan**

Existing integration tests.

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -75,7 +75,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
     if (cmds.length) {
       // Disable wrapper in executed commands
-      process.env.YARN_WRAPOUTPUT = '0';
+      process.env.YARN_WRAPOUTPUT = 'false';
       for (const [stage, cmd] of cmds) {
         // only tack on trailing arguments for default script, ignore for pre and post - #1595
         const cmdWithArgs = stage === action ? sh`${unquoted(cmd)} ${args}` : cmd;
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -75,7 +75,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
     if (cmds.length) {
       // Disable wrapper in executed commands
-      process.env.YARN_WRAPOUTPUT = 'false';
+      process.env.YARN_WRAP_OUTPUT = 'false';
       for (const [stage, cmd] of cmds) {
         // only tack on trailing arguments for default script, ignore for pre and post - #1595
         const cmdWithArgs = stage === action ? sh`${unquoted(cmd)} ${args}` : cmd;
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -74,8 +74,8 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
 
     if (cmds.length) {
-      // propagate YARN_SILENT env variable to executed commands
-      process.env.YARN_SILENT = '1';
+      // Disable wrapper in executed commands
+      process.env.YARN_WRAPPER = '0';
       for (const [stage, cmd] of cmds) {
         // only tack on trailing arguments for default script, ignore for pre and post - #1595
         const cmdWithArgs = stage === action ? sh`${unquoted(cmd)} ${args}` : cmd;
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -75,7 +75,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
     if (cmds.length) {
       // Disable wrapper in executed commands
-      process.env.YARN_WRAPPER = '0';
+      process.env.YARN_WRAPOUTPUT = '0';
       for (const [stage, cmd] of cmds) {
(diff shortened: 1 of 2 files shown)


src/cli/index.js
```diff
@@ -207,7 +207,8 @@ export function main({
   reporter.initPeakMemoryCounter();
 
   const config = new Config(reporter);
-  const outputWrapper = !commander.json && command.hasWrapper(commander, commander.args);
+  const outputWrapperEnv = process.env.YARN_WRAPOUTPUT === undefined || process.env.YARN_WRAPOUTPUT == '1';
```
BYK: I'd go with the following:

```
const disableOutputWrapper = process.env.YARN_WRAPOUTPUT === '0' || process.env.YARN_WRAPOUTPUT === 'false';
const shouldWrapOutput = !disableOutputWrapper && !commander.json && command.hasWrapper(commander, commander.args);

if (shouldWrapOutput) {
```

sth:
I moved the boolification of the environment variable into a separate function. This function can then also be used for other variables, like `YARN_SILENCE`, to make sure they all accept the same values as true/false.

The functionality of `boolifyWithDefault()` could get integrated directly into `boolify()´, but it seemed more clear to me as a separate function.

src/cli/commands/run.js
```diff
@@ -74,8 +74,8 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
 
     if (cmds.length) {
-      // propagate YARN_SILENT env variable to executed commands
-      process.env.YARN_SILENT = '1';
+      // Disable wrapper in executed commands
+      process.env.YARN_WRAP_OUTPUT = 'false';
```
BYK: Note to self: do not mutate `process.env` in any way.

edmorley:
Many thanks for this! :-)

PR #4815, opened 2017-10-31
Fixes scoped packages handling with offline installation

### LINKED ISSUE #4266: Issue with offline installation of @types/node
Bug. @BYK suggested I repost this as a new issue.

Offline installation of JSDOM fails on the dependency `@types/node`. This was noted in #825 but does not seem fully resolved.

```
sraman@sraman: bar$ yarn add jsdom@11.0.0 --offline
yarn add v0.27.5
info No lockfile found.
[1/4] Resolving packages...
error Couldn't find any versions for "@types/node" that matches "^6.0.46" in our cache. Possible versions: ""
```

It is definitely available in my cache.

```
sraman@sraman: bar$ yarn cache ls | grep "@types"
@types/node                                          6.0.78                      npm      https://registry.yarnpkg.com/@types/node/-/node-6.0.78.tgz#5d4a3f579c1524e01ee21bf474e6fba09198f470
@types/node                                          6.0.87                      npm      https://registry.yarnpkg.com/@types/node/-/node-6.0.87.tgz#5ab5774f8351a33a935099fa6be850aa0b0ad564
@types/node     
```

A similar failure with a different message in the latest nightly:

```
sraman@sraman: bin$ ./yarn add jsdom@11.0.0 --offline
yarn add v1.0.0-20170825.1555
warning No license field
error Couldn't find any versions for "@types/node" that matches "^6.0.46" in our cache (possible versions are ""). This is usually caused by a missing entry in the lockfile, running Yarn without the --offline flag may help fix this issue.
```

I expect JSDOM to be installed offline. Other cached packages work fine with offline installation:

```
sraman@sraman: bin$ ./yarn add yargs@3.27.0 --offline
yarn add v1.0.0-20170825.1555
warning No license field
[2/4] 🚚  Fetching packages...
success Saved lockfile.
success Saved 18 new dependencies.
├─ align-text@0.1.4
├─ camelcase@1.2.1
├─ center-align@0.1.3
├─ cliui@2.1.0
├─ decamelize@1.2.0
├─ invert-kv@1.0.0
├─ is-buffer@1.1.5
├─ kind-of@3.2.2
├─ lazy-cache@1.0.4
├─ lcid@1.0.0
├─ longest@1.0.1
├─ os-locale@1.4.0
├─ repeat-string@1.6.1
├─ right-align@0.1.3
├─ window-size@0.1.4
├─ wordwrap@0.0.2
├─ y18n@3.2.1
└─ yargs@3.27.0
warning No license field
✨  Done in 0.43s.
```

Node: v6.7.0
Tested with Yarn 0.27.5 and nightly  v1.0.0-20170825.1555

sharno:
**Summary**
Fixes #4266 
I tracked the problem and found out that yarn misses scoped packages from cache. As scoped packages are put inside a scope (folder) in the cache folder. And yarn just scans one level of the folders inside the cache for the metadata file.
So added code to determine if the folder doesn't have the metadata file, it should be a scope folder and packages inside it gets added to valid packages.

**Test plan**
I didn't write tests for this part. But help would be appreciated if I should do. Just tried installation of many scoped packages using the offline flag and they all worked well.

This is my first PR here. Feedback would be a lot appreciated


src/resolvers/registries/npm-resolver.js
```diff
@@ -107,8 +107,19 @@ export default class NpmResolver extends RegistryResolver {
...
       const files = await fs.readdir(cacheFolder);
-      const validFiles = [];
+      const scoped_files = {};
+      for (const scope of files) {
+        // no hidden files
+        if (scope[0] === '.') {
+          continue;
+        }
+
+        if (!await fs.exists(path.join(cacheFolder, scope, METADATA_FILENAME))) {
```
BYK: Using async exists is discouraged by Node folks since the file may not be there when you actually want to use it. I guess that's not very different from what is previously done but this now adds a lot more I/O during offline-cache init.

src/resolvers/registries/npm-resolver.js
```diff
@@ -107,8 +107,19 @@ export default class NpmResolver extends RegistryResolver {
 
     const files = await this.config.getCache('cachedPackages', async (): Promise<Array<string>> => {
       const files = await fs.readdir(cacheFolder);
-      const validFiles = [];
+      const scoped_files = {};
```
BYK: You should use a "real" `Map` object here ;)

src/resolvers/registries/npm-resolver.js
```diff
@@ -107,8 +107,19 @@ export default class NpmResolver extends RegistryResolver {
...
+      for (const scope of files) {
+        // no hidden files
+        if (scope[0] === '.') {
```
BYK: I think this check should be extended to only accept things starting with an `@` since that denotes a scope.

src/resolvers/registries/npm-resolver.js
```diff
@@ -107,8 +107,19 @@ export default class NpmResolver extends RegistryResolver {
...
+          continue;
+        }
+
+        if (!await fs.exists(path.join(cacheFolder, scope, METADATA_FILENAME))) {
+          scoped_files[scope] = await fs.readdir(path.join(cacheFolder, scope));
+        }
```
arcanis: Rather than checking if the file exists, you should instead check whether the folder name starts with a `@`, cf [this code](https://github.com/yarnpkg/yarn/blob/master/src/package-linker.js#L261)

src/resolvers/registries/npm-resolver.js
```diff
@@ -121,6 +132,20 @@ export default class NpmResolver extends RegistryResolver {
           validFiles.push(name);
         }
       }
+      for (const scope of Object.keys(scoped_files)) {
```
BYK: Why do two passes instead of doing all in the same loop? You know iterate on `files` twice and then do another loop over everything inside folders. You can definitely combine the two loops over the `files` part and avoid a separate loop for files inside the folder by extracting the core logic in the loop into a separate function.

src/resolvers/registries/npm-resolver.js
arcanis: I'd also prefer if you could iterate over the scope folder inside this loop, instead of storing the result inside `scoped_files` (which should be `scopedFiles` anyway :) and iterating over it later on.

arcanis:
Thanks a lot! Asked for a few changes, let me know if they make sense.

> I didn't write tests for this part. But help would be appreciated if I should do

Yep, tests would be great. Here are some pointers:

- I think those tests should go into `__tests__/commands/install/integration.js`
- You can run the test using `yarn test __tests__/commands/install/integration.js`
- In order to test this feature, you can run an classic install with a scoped package (via `runInstall`), then remove the generated `node_modules`, then run another install in offline mode (look for `new Install` in the file to see how it's done)

newren:
I have an alternative fix for offline mode with scoped packages at #4822, though I wasn't sure reading through #4266 if we're both truly suffering from the same problem.  My fix is a simple two-liner, and shouldn't have any performance implications, but it's possible it doesn't solve everything you need.  

(Just thought I'd look for similar issues and post a heads-up in case it's useful.)

sharno:
Thanks a lot @BYK and @arcanis for your feedback. Makes a lot of sense, indeed it was an inefficient solution.
I think @newren 's solution makes a lot of sense at #4822 .
I've tested it with the same scenario as #4266 and it partially worked (without throwing errors) by installing an older version jsdom@9.12.0 although I had all dependencies of jsdom@11.3.0 in cache. As the solution doesn't flatten the scoped packages in cache still. And the issue of reading scoped packages from cache still persists. (probably jsdom@9.12.0 didn't require scoped packages)

I might take a shot at flattening the cache folder for scoped packages if @newren wouldn't add it to his solution.

BYK:
>I might take a shot at flattening the cache folder for scoped packages if @newren wouldn't add it to his solution.

I'd try that :)

newren:
My PR was just about the offline mirror and avoiding file collisions there; I didn't intend to touch the cache folder or modify it.  That may change if we start running into whatever issues you are hitting here, but for now I'm happy with #4822 and maybe reading/testing any patches that get posted to this PR.

PR #4816, opened 2017-11-01
fix(cli): Don't warn when running on node 9

graingert:
**Summary**

Update the accepted semver range to not warn on unknown future versions of node.

**Test plan**

N/A since we don't want to add Node 9 into our build matrix just yet.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -13,7 +13,7 @@ export const DEPENDENCY_TYPES = ['devDependencies', 'dependencies', 'optionalDep
 export const RESOLUTIONS = 'resolutions';
 export const MANIFEST_FIELDS = [RESOLUTIONS, ...DEPENDENCY_TYPES];
 
-export const SUPPORTED_NODE_VERSIONS = '^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0';
+export const SUPPORTED_NODE_VERSIONS = '^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0';
 
 export const YARN_REGISTRY = 'https://registry.yarnpkg.com';
 


src/constants.js
```diff
@@ -13,7 +13,7 @@ export const DEPENDENCY_TYPES = ['devDependencies', 'dependencies', 'optionalDep
 export const RESOLUTIONS = 'resolutions';
 export const MANIFEST_FIELDS = [RESOLUTIONS, ...DEPENDENCY_TYPES];
 
-export const SUPPORTED_NODE_VERSIONS = '^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0';
+export const SUPPORTED_NODE_VERSIONS = '^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0 || ^9.0.0';
```
BYK: I think we can simply do `>= 8.0.0` for now to be more future-proof.

appveyor.yml
```diff
@@ -1,5 +1,6 @@
 environment:
   matrix:
+    - node_version: "9"
```
BYK: Since Node 9 will be "experimental" I don't think adding it to the build chain would be a good idea. Things may break or change randomly and that would cause churn with the team and reduce the signal/noise ratio for CI build results.

src/constants.js
graingert: Ok, but then we'd need to add that to CI

BYK: Not really. It is just to make the warning go away since we simply don't know.

appveyor.yml
SimenB: I'm curios, what do you mean with experimental?

mgol: Formally it's stable, not experimental. Experimental were RCs before the final. Node 8 is LTS but that doesn't mean Node 9 is not stable; it officially is.

SimenB: It's odd to test on node 5, which has been EOL for over a year, but not node current

PR #4822, opened 2017-11-01
fix(fetcher): offline mirror name collision w/ private registries and scopes

newren:
**Summary**

When using artifactory as an internal npm registry in conjunction with scoped packages, yarn install was failing (about 50% of the time) with messages of the form
    
      Hashes don't match when extracting file
      "https://artifactory.internal.site:443/artifactory/api/npm/external-mirror/@types/react/-/react-15.6.4.tgz".
      Expected "3bb57bd43183a05919ceb025a264287348f47e9d" but got
      "da39a3ee5e6b4b0d3255bfef95601890afd80709"
    
The problem was that yarn was writing both `@types/react-15.6.4.tgz` and `react-15.6.4.tgz` to the same location within the offline mirror due to making an assumption about no additional leading paths before the scope name in the URL -- an assumption which was true for registry.npmjs.org, but not for artifactory npm repositories.

**Test plan**

Added a new unit test.


newren:
The two failures in one of the travis builds look to me like unrelated test flakes; the code I touched shouldn't affect either of those two operations.  Is there a way to retrigger the build to see if it duplicates?

newren:
Thanks for the review!  I've tried to address each of your three issues, but I'm still a newbie to the web development side of programming, so if I missed anything or there's further changes you'd like to see, just let me know.

newren:
The failed builds look like flakes to me; my code didn't touch the codepaths of the tests that failed, plus some were just timeouts.  Is there a way to retrigger those tests or something?

newren:
Hey, @BYK was just curious if my changes still needed further fixes or if they're now to your liking.  (As mentioned above, the test failures look like unrelated flakes and I don't seem to have perms to retrigger the builds.)  If you're just busy, I understand; just pinging in case my update fell through the cracks.

BYK:
Woops, sorry for the silence. This fell off my radar. Doing another pass now.

src/fetchers/tarball-fetcher.js
```diff
@@ -17,6 +17,8 @@ const stream = require('stream');
 const gunzip = require('gunzip-maybe');
 const invariant = require('invariant');
 
+const RE_URL_NAME_MATCH = /\/(?:(@[^/]+)\/)?[^/]+\/-\/([^/]+)$/;
```
newren: I think this will fail for some packages.  For example, looking at some sample resolved urls in a random local yarn.lock, it appears that _most_ would match your regex, such as ones of the form

https://publish.internal.site:443/artifactory/api/npm/external-npm/abab/-/abab-1.0.3.tgz
or
https://publish.internal.site:443/artifactory/api/npm/external-npm/@types/react/-/react-15.0.38.tgz

However, there are also some of the form

https://artifactory.internal.site:443/artifactory/api/npm/internal-npm-release/@team/package/-/@team/package-1.2.3.tgz

It looks like all internally team-scoped packages appear to have an extra "/@team/" between the "/-/" and the tarballBasename.

BYK: That's quite weird and it was also what I feared. We can tune the regex to handle these too but ideally, we should not be extracting the scope information from the URL anyway.

Since changing the names of these files would invalidate any existing offline mirror, we need to find a migration path when doing this in a better way.

BYK: I've pushed a new change but I also need to extend and refactor the tests to cover this new scenario. Feel free to add them since I can't look at this patch until tomorrow.

PR #4832, opened 2017-11-02
ci(circle): Remove unnecessary `--` from test-ci invocations

BYK:
**Summary**

After Yarn 1.0, the `--` separator to pass arguments to run scripts is no longer necessary. We still
have it in our CircleCI config and this patch removes it since Circle CI now uses Yarn 1.1.0+

**Test plan**

CircleCI builds should pass without any warnings or errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
               ./scripts/set-dev-version.js
             fi;
             # Limit maxWorkers to 3 to avoid OOM on CircleCI
-            yarn test-ci -- -- --maxWorkers 3
+            yarn test-ci --maxWorkers 3
             yarn check-lockfile
   lint:
     <<: *defaults


arcanis:
You've removed two `--`, which seem to imply that one of them was required (and so the first one as well, because of the deprecation), are you sure it will work?

BYK:
>You've removed two --, which seem to imply that one of them was required (and so the first one as well, because of the deprecation)

It was because `test-ci` command itself runs `yarn test-only` so we needed the extra `--` which would be stripped by the enclosed `yarn run` call.

>, are you sure it will work?

Why not ask this question to CircleCI build logs? 😁

arcanis:
> It was because test-ci command itself runs yarn test-only so we needed the extra -- which would be stripped by the enclosed yarn run call.

Makes sense 👍

PR #4834, opened 2017-11-02
ci(circle): Run deploy tasks on master branch

arcanis:
**Summary**

CircleCI is not deploying the build artifacts anymore. This PR tries to fix that by using the `only` field to explicitly say that the deploy should run on master (not sure it will have an effect).

https://circleci.com/docs/2.0/workflows/#branch-level-job-execution

cc #4271

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,8 @@ workflows:
           filters:
             <<: *default_filters
             branches:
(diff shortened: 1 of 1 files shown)


arcanis:
Queued for 2h40m and still going ...

Daniel15:
Isn't deploy supposed to run for tags, not for master?

PR #4835, opened 2017-11-02
fix(resolutions): Warn instead of erroring if unmatched dependency matches resolutions

### LINKED ISSUE #4825: yarn check does not respect resolutions
*bug*

Issue: `backbone.marionette@1.8.8` states `backbone@1.0.0 - 1.1.2` as its dependency, but I need it to work with `backbone@1.2.3` and it actually **does** work with no issues, so yarn's v1.0 `resolutions` was a life-saver for me.
However `yarn check` still complains `error "backbone.marionette#backbone@1.0.0 - 1.1.2" doesn't satisfy found match of "backbone@1.2.3"`

The `package.json` is as follows
```json
{
  "dependencies": {
    "backbone": "1.2.3",
    "backbone.marionette": "1.8.8"
  },
  "resolutions": {
    "backbone": "1.2.3"
  }
}
```
`yarn check` does not complain about that mismatch, cause it's stated in `resolutions` field

 - yarn 1.2.1
 - node 6.11.2
 - MacOS

dreyks:
**Summary**

Fixes #4825. 

When running `yarn check` do not error if the unmatched dependency matches the override in `resolutions` field of `package.json`. Just warn about the mismatch

**Test plan**

New test case added.


BYK:
Hey @dreyks, thanks for the PR! Do you think you can make the test case isolated and smaller with custom packages you create so we don't need to check in all these files?

dreyks:
yup, will do

dreyks:
changed to packages that do not have other dependencies. will that do or we need something even smaller and hand-crafted?

also rebased onto current master

src/cli/commands/check.js
```diff
@@ -336,7 +337,20 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
           const foundHuman = `${humaniseLocation(path.dirname(depPkgLoc)).join('#')}@${depPkg.version}`;
           if (!semver.satisfies(depPkg.version, range, config.looseSemver)) {
             // module isn't correct semver
-            reportError('packageDontSatisfy', subHuman, foundHuman);
+            const resRange = mainPackageJson.resolutions && mainPackageJson.resolutions[name];
+            if (resRange) {
+              const resHuman = `${human}#${name}@${resRange}`;
+
+              if (semver.satisfies(depPkg.version, resRange, config.looseSemver)) {
+                reporter.warn(reporter.lang('packageDontSatisfy', subHuman, foundHuman));
```
kaylieEB: From the test case above, I find the warning above a little confusing. Would you mind using [incompatibleResolutionVersion](https://github.com/yarnpkg/yarn/blob/master/src/reporters/lang/en.js#L126) instead? 

Currently it reads `warning.*"pad-left#repeat-string@\\^1.5.4" doesn't satisfy found match of "repeat-string@1.4.0"` 

but it'd be better to say

`warning.*Resolution field (repeat-string@1.4.0) is incompatible with requested version (pad-left#repeat-string@\\^1.5.4)'`

dreyks: 👍
done.

rebased onto current master once again

kaylieEB:
@BYK this lgtm but it looks like I no longer have rights to merge (getting `This pull request can be automatically merged by project collaborators
Only those with write access to this repository can merge pull requests. You're not authorized to merge this pull request.`). Do you want to go ahead? :) Thanks!

BYK:
>@BYK this lgtm but it looks like I no longer have rights to merge 

No idea how that happened but I'll make sure to fix that instead. Sorry :(

dreyks:
do I have to squash it? any other actions from my side?

BYK:
@dreyks nope, but I'd appreciate if you can expand the summary part of your PR a bit so we can use it in the commit log. We use squash merges so you don't need to do anything else.

I'm just waiting for @kaylieEB to wake up and see if she can merge this and if not, why.

kaylieEB:
Done! Thanks @dreyks 🎉  and @BYK

Issue #2933, opened 2017-03-16
Upgrade to CircleCI 2.0


Daniel15:
This is just a tracking task so I don't forget 😄 

Last time I attempted this, I encountered an issue where CircleCI would say that the build succeeded even when it failed. Need to investigate why that happened and fix it.

WIP branch: https://github.com/yarnpkg/yarn/tree/circleci-2

charpeni:
The config from @ryanwohara seems a bit outdated, so I followed the latest documentation from CircleCI 2.0 if you want to take a look in #2947.

Daniel15:
You're right! :)

PR #4839, opened 2017-11-02
ci(circle): Fix deploy job so it only runs on version tags

BYK:
**Summary**

This PR fixes the deploy job to only run for version tags. It also renames it to "Publish" to convey
the actual task being performed since this job only publishes to NPM. Everything else is handled by
our webhooks.

Source: https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

**Test plan**

Builds on master should be fixed and when something is tagged, it should be deployed to NPM.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,10 +5,6 @@ defaults: &defaults
   docker:
     - image: yarnpkg/dev:latest
 
-default_filters: &default_filters
-  tags:
-    only: '/^v[0-9]\+\.[0-9]\+\.[0-9]\+$/'
-
 restore_node_modules: &restore_node_modules
   restore_cache:
     name: Restore node_modules cache
@@ -90,14 +86,14 @@ jobs:
           root: ~/project
           paths:
             - yarn
-  deploy:
+  publish:
     <<: *defaults
     steps:
       - attach_workspace:
           at: ~/project
       - *restore_node_modules
       - run:
-          name: Deploy
+          name: Publish
           command: |
             # Only NPM is handled here - All other release files are handled in a webhook.
             if [ "${CIRCLE_PROJECT_USERNAME}" == "yarnpkg" ]; then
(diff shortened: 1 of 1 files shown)


tusbar:
@BYK @arcanis that does not seem to be the issue. The original scripts from #2933 was fine. The issue seems to be with the RegExp.

```bash
$ node
> const r = /^v[0-9]\+\.[0-9]\+\.[0-9]\+$/
undefined
> 'v1.3.1'.match(r)
null
```

`/^v[0-9]+\.[0-9]+\.[0-9]+$/` should be better.

BYK:
Oh god. Well this is needed tho since now it will try to publish every version on master after #4834.

tusbar:
@BYK this will not work. Tags are opt-in in CircleCI v2. If you don’t specify a tag filter, they will not be run. See https://circleci.com/docs/2.0/configuration-reference/#tags.

> If neither only nor ignore are specified then the job is skipped for all tags.

BYK:
@tusbar I did this based on the information on https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

BYK:
@tusbar are you saying I need to add the same thing to all jobs as it was previously?

tusbar:
@BYK yes.

From your link:

> For a branch push unaffected by any filters, CircleCI runs the job.
> For a tag push unaffected by any filters, CircleCI skips the job.

The current state of this PR will run **ALL** tasks for branches (including `publish`) and **ONLY** `publish` for tags matching the regexp.

You want all tasks to be ran for tags matching the regexp, and all tasks except for `publish` for branches.

BYK:
@tusbar does it look good now?

tusbar:
Just tested it out: https://circleci.com/workflow-run/0f0b1315-d829-423b-a5b8-c230917736c1 works fine without the quotes as expected.

PR #4730, opened 2017-10-17
fix(linker): use lockfileFolder when creating bin links

jgoz:
**Summary**

Fixes #4706, fixes #4359, refs #4513. `this.config.cwd` was being used as the root for bin link paths, rather than `this.config.lockfileFolder`.

**Test plan**

- Added tests for `add` and `remove` commands (#4706)
- Added test for `install` command (#4359)

Issue #4543, opened 2017-09-25
[workspaces] linking binaries from the root in workspaces


akosyakov:
I wonder whether it would be possible to link binaries in all workspaces from dependencies declared in the root?

Right now to call binaries within a workspace one should redeclare it. It leads to trouble with keeping identical versions between workspaces.

bryzaguy:
+1 Yes this!

bryzaguy:
Basically create the symlinks in the .bin folder of each workspace project for the matching hoisted dependencies that project relies on.  I have to script this right now to keep everything working.

rtm:
@bryzaguy  But I thought that was already how it worked, assuming you are referring to the links within `node_modules/.bin`. At least if I understand the question correctly. Although at the moment I am running into some issue with this behavior; see https://stackoverflow.com/questions/46482966/yarn-putting-broken-links-into-node-modules-bin-in-workspace-repos.

Or is this issue requesting that links be placed in **all** packages even for dependencies only declared at the root? That seems questionable.

akosyakov:
> Or is this issue requesting that links be placed in all packages even for dependencies only declared at the root? That seems questionable.

I concern mostly about managing dev dependencies in one place.

BTW as a workaround one can run `yarn` in a workspace before running scripts. It links binaries then into the workspace.

wojtekmaj:
This would greatly improve install time, disk usage and performance in my use case. This is one feature that sadly stops me from using yarn in most of my projects I develop.

julien-f:
IMHO, if your binaries are used directly from your packages, they should be declared as dependencies of these package, not simply of the root.

I personally try to keep some locality, root dependencies are only used from the root (https://github.com/vatesfr/xen-orchestra/blob/master/package.json).

AviVahl:
This is a small feature that can go a long way in terms of usability.

I have a repo with several packages, where I'm using yarn workspaces for proper cross-package development.

Many of my CLI-based devDependecies (typescript, rimraf, mocha, karma, tslint) keep repeating in every package, and I would rather control their versions in a **single** place (the root `package.json`).

If I don't include them in every `package.json` (not just the root one), scripts in those packages aren't able to access them. I end up repeating these deps and their versions just so the binaries are accessible.

There are probably several ways to address this...

Linking the root `.bin` doesn't seem like a good idea, just for those cases where the specific package has a unique dependency (meaning one not moved to the root scope) that has a cli.

[...]

PR #4848, opened 2017-11-02
feat(run): Include the workspace root .bin in env path

rally25rs:
**Summary**

Partial resolution for #4543.

Previously, when running `yarn run` the env PATH would be set to look in node_modules/.bin, however, in workspaces the root workspace .bin path was not being included.

This PR adds the workspace root
node_modules/.bin path after the individual package's path.

This is generally needed because #4730 ensures bin links in a workspace will be at the workspace root. With this PR, you can now `yarn run` commands in an individual package again.

**Test plan**

Manually tested by adding a script that runs `echo $PATH`


BYK:
This is awesome! Let's add one more test, merge from master so we get the latest CI upgrades and I'm game!

rally25rs:
@BYK finally got around to fixing this test. All ✅ now 👍

__tests__/commands/run.js
```diff
@@ -133,3 +144,20 @@ test('adds quotes if args have spaces and quotes', (): Promise<void> => {
...
   });
 });
+
+test('adds workspace root node_modules/.bin to path when in a workspace', (): Promise<void> => {
+  return runRunInWorkspacePackage('packages/pkg1', ['env'], {}, 'workspace', (config, reporter): ?Promise<void> => {
+    const logEntry = reporter.getBuffer().find(entry => entry.type === 'log');
+    const logData = logEntry ? logEntry.data.toString() : '{}';
+    const parsed = logEntry ? JSON.parse(logData) : {};
+    let envPaths = [];
+    if (parsed.PATH) {
```
BYK: Please use [`path.delimiter`](https://nodejs.org/api/path.html#path_path_delimiter) instead. I'll go ahead and fix this myself and then merge tho :)

__tests__/commands/run.js
```diff
@@ -133,3 +144,20 @@ test('adds quotes if args have spaces and quotes', (): Promise<void> => {
...
+      envPaths = parsed.PATH.split(':');
+    } else if (parsed.Path) {
+      envPaths = parsed.Path.split(';');
+    }
+
+    expect(envPaths).toContain(path.join(config.cwd, 'node_modules/.bin'));
```
BYK: I expected this to fail on Windows since the path separator should be `\` there. Weird.

rally25rs: IIRC `path.join()` converts it auto-magically.

__tests__/commands/run.js
rally25rs: Oh cool, I didn't even know that was a thing. Thanks!

__tests__/commands/run.js
rally25rs: although... why didn't I do `path.join(config.cwd, 'node_modules', '.bin')`? 😕 ...I'm an idiot 😆

BYK: Wow, nice.

iansu:
Any idea when this is going to be released?

jonaskello:
I cannot get this to work. Is it included in 1.3.2?

simonbuchan:
I think there was some talk about putting the release tags a merged PR is included by, like they are in commits, but for now you can check in the footer of the commit message on the page for the merge commit, which is linked in the merge message above.

PR #4849, opened 2017-11-03
dont set proxy when proxy does not exist

yangli1990:
**Summary**

Request library uses proxies such as HTTP_PROXY and HTTPS_PROXY, hence by setting these proxies it should work. However, the previous change to add an empty proxy even when no proxy is provided to yarn config will override the behaviour of request using the environment variable proxies.

Regarding #4761 we should look at another way of doing this rather than disallowing HTTPS_PROXY/HTTP_PROXY settings?

**Test plan**


arcanis:
That seems reasonable. @rally25rs what do you think? Maybe instead of checking for falsy-ness we could instead check for strictly false values and only write an empty string in this case.

rally25rs:
Sorry I totally missed this PR.

The reason I had started always passing the proxy was because if you set `https_proxy false` in .yarncfg, then it should _not_ fall back to the env variable. If you don't pass a proxy (as done in this PR) then the underlying `request` library will use the env var.

 I ended up making a different PR for the fix before I noticed this one existed. Please see #4903 instead of this PR. In that PR I started checking explicitly for a `false` value.

Going to close this PR in favor of #4903.

PR #4852, opened 2017-11-03
test: Fix flaky tests after #4811 due to bad assumptions

BYK:
**Summary**

Tests started failing on Travis after #4811, somewhat randomly, due to the mexpecting the unwrapped
output. This PR fixes those expectations and moves normalize-manifest tests to snapshots since
that's easier than updating 40+ JSON files by hand.

**Test plan**

Tests should pass on all platforms and CI and locally.


Daniel15:
👍

PR #4857, opened 2017-11-04
ci(circle): Test all node versions on CircleCI

BYK:
**Summary**

This PR removes the custom Docker image we have and adds all node versions to CircleCI tests. The
aim is to levearge increased granularity and parallelism that the new CircleCI workflows provide.

**Test plan**

All steps on CircleCI should pass.


BYK:
I've contacted CircleCI and got access to their macOS stack too. The plan is, to remove our reliance on TravisCI completely by leveraging that plan. It should help us get much faster signal on PRs since the workflows make things quite efficient.

BYK:
@Daniel15 @arcanis you can now see that all build steps passed on https://circleci.com/gh/yarnpkg/yarn/tree/circle-power (only check the last 5).

After this PR lands, I'll change the repo configuration to look for those steps instead of `ci/circleci: test` step.

.circleci/config.yml
```diff
@@ -1,28 +1,47 @@
...
+  steps:
+    - *attach_workspace
+    - *restore_cache
+    - run:
+        name: Tests
+        command: |
+          if [ "$CIRCLE_BRANCH" == 'master' ]; then
+            ./scripts/set-dev-version.js
+          fi;
+          yarn build
```
BYK: We can possibly move this step into `install` and save a few more seconds per test run.

.circleci/config.yml
```diff
@@ -43,41 +62,33 @@ jobs:
...
-          at: ~/project
-      - *restore_node_modules
+      - *attach_workspace
+      - *restore_cache
       - run:
           name: Build distribution
           command: |
-            node -v
+            apt-get -y update && \
+            apt-get install -y --no-install-recommends \
```
Daniel15: This should really be in Docker images so we don't need to install the packages on every single build. That'll unnecessarily slow down the build and defeats the purpose of using Docker :/

I can make a Docker image per Node.js version if that'd be useful?

BYK: We don't need a docker image per node version since this is only needed for the build stage which only happens with Node 8. I don't like having to maintain custom docker images for trivial stuff like this but I'll add back a docker image for this step and update the PR.

BYK:
@Daniel15 all green: https://circleci.com/gh/yarnpkg/yarn/tree/circle-power

BYK:
@Daniel15 I think we can definitely do it. Filed #4863 so we don't forget.

PR #4860, opened 2017-11-05
feat(install): Add --update-checksums to cli install

### LINKED ISSUE #4817: Moving repositories without recreating yarn.lock?

Bug (I think!)

We're trying to move to a new internal npm repository (specifically, moving from Sinopia to Nexus). We've been using yarn extensively for all of our packages. So when moving, we have to do a search/replace of the repository-url in the yarn.lock file.
While not ideal, this is not a problem.

The problem is with the sha checksum of each package. The checksum includes (from the generated package.json file of each package in node_modules) the repository url. Meaning it is by definition different and yarn fails with:
`error <package_in_new_repository>.tgz: Bad hash. Expected "<old_checksum>" but got "<new_checksum>"`

Looking through the documentation and the issues I could not find any mention of a solution for this. Am I missing something? Is there a better solution than updating all the checksums in addition to the repository urls in `yarn.lock`?

1. install a private package through one private repository (Sinopia/Verdaccio/Nexus/etc.) and save it to package.json and yarn.lock
2. delete node_modules and yarn cache, change the repository url in `yarn.lock` to a different private repository
3. try to install the package again

If indeed I did not miss an existing option to do so, I think a `yarn --update-checksum` flag would be nice. I'd be willing to work out a PR.

imsnif:
**Summary**

Fixes #4817.
When the `--update-checksums` flag is set, yarn would know to ignore a checksum mismatch between `yarn.lock` and the repository, and instead update the yarn.lock file with the proper checksum(s).

**Test plan**

Added new tests.

To manually check this:
1. Change one or more of the package checksums in `yarn.lock`
2. Delete node_modules (optionally also run `yarn cache clean`)
3. Run `yarn` => checksum mismatch error will be received.
4. Run `yarn --update-checksums` => will install successfully and fix the damaged checksums in `yarn.lock`


imsnif:
Sure thing!
Linking the PR here for symmetry's sake: https://github.com/yarnpkg/website/pull/708

bestander:
Travis: `ENOSPC: no space left on device, open '/tmp/ramfs/jest_dx/jest-transform-cache-`
Let's wait for other CI to finish queue

bestander:
@imsnif, could you merge master into this PR?
It does not look like CI failures are affected by your change so let's rerun them

imsnif:
Merged: helped with Travis, but not so much with circleci or appveyor... networking issues?
The build also failed for me locally, but they were the same (different) failures that were in Travis, so I assumed some work was being done there.

PR #4864, opened 2017-11-06
ci(travis): Reduce the build matrix on TravisCI

BYK:
**Summary**

We now build and lint on CircleCI so reduce the number of tests on TravisCI and simplify the config.

**Test plan**

TravisCI tests should still pass and be less.


arcanis:
What tests are still being run on Travis, and can't we move them to CircleCI? Is it only OSX?

BYK:
>What tests are still being run on Travis, and can't we move them to CircleCI? Is it only OSX?

Only Node 8 and Node 6 but you're right, I'll remove those two and only leave OS X. I'll submit another PR that moves macOS tests to CircleCI too. They've recently given us access to their new macOS build stack.

PR #4866, opened 2017-11-06
test(util): Add tests for boolify() and boolifyWithDefault()

sth:
**Summary**

This adds tests for `boolify()` and `boolifyWithDefault()`. It also fixes a bug in `boolify()` discovered in the process: the number `0` should evaluate to `false`, not `true`. 

The functions are moved from cli/index.js to util/conversion.js since they can be generally useful.

**Test plan**

Adding tests is the main point.

### DIFF
--- a/__tests__/util/conversion.js
+++ b/__tests__/util/conversion.js
@@ -33,7 +33,7 @@ describe('boolifyWithDefault', () => {
     expect(boolifyWithDefault('', false)).toBe(false);
   });
 
-  test('should be consistent with boolify()', () => {
+  test('should be consistent with boolify() for non-empty values', () => {
     for (const value of [1, '1', true, 'true', 0, '0', false, 'false', 'randomstringlkjsaljdja', 649815]) {
       for (const defaultResult of [true, false]) {
         expect(boolifyWithDefault(value, defaultResult)).toBe(boolify(value));
--- a/__tests__/util/conversion.js
+++ b/__tests__/util/conversion.js
@@ -1,0 +1,43 @@
+/* @flow */
+
+import {boolify, boolifyWithDefault} from '../../src/util/conversion.js';
+
+describe('boolify', () => {
+  test('should recognize "true" values', () => {
+    expect(boolify('1')).toBe(true);
+    expect(boolify(1)).toBe(true);
+    expect(boolify('true')).toBe(true);
+    expect(boolify(true)).toBe(true);
+  });
+
+  test('should recognize "false" values', () => {
+    expect(boolify('0')).toBe(false);
+    expect(boolify(0)).toBe(false);
+    expect(boolify('false')).toBe(false);
+    expect(boolify(false)).toBe(false);
+  });
+
+  // Not sure if that's really a requirement, but it's how it currently works
+  test('should treat unrecognized values as "true"', () => {
+    expect(boolify('randomstringdfkjh')).toBe(true);
+  });
+});
+
+describe('boolifyWithDefault', () => {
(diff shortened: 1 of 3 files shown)


__tests__/util/conversion.js
```diff
@@ -0,0 +1,43 @@
...
+  test('should return the default for empty values', () => {
+    expect(boolifyWithDefault(undefined, true)).toBe(true);
+    expect(boolifyWithDefault(undefined, false)).toBe(false);
+    expect(boolifyWithDefault(null, true)).toBe(true);
+    expect(boolifyWithDefault(null, false)).toBe(false);
+    expect(boolifyWithDefault('', true)).toBe(true);
+    expect(boolifyWithDefault('', false)).toBe(false);
+  });
+
+  test('should be consistent with boolify()', () => {
```
BYK: Well, instead of generating these expectations, you can also just mock `boolify` and make sure it is being called? :)

src/util/conversion.js
```diff
@@ -0,0 +1,14 @@
+/* @flow */
+
+export function boolify(val: string | number | boolean): boolean {
+  const strVal = val.toString().toLowerCase();
+  return strVal !== 'false' && strVal !== '0';
```
BYK: How about we move these constants into `const FALSY_STRINGS = new Set(['false', '0']);` so we can just do `return FALSY_STRINGS.has(strVal.toString().toLowerCase())`?

This would also make it more extensible.

src/util/conversion.js
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export function boolifyWithDefault(val: ?(string | number | boolean), defaultResult: boolean): boolean {
+  if (val === undefined || val === null || val === '') {
+    return defaultResult;
```
BYK: Why not using a ternary?

src/cli/index.js
```diff
@@ -23,6 +23,7 @@ import {getRcConfigForCwd, getRcArgs} from '../rc.js';
 import {spawnp, forkp} from '../util/child.js';
 import {version} from '../util/yarn-version.js';
 import handleSignals from '../util/signal-handler.js';
+import {boolify, boolifyWithDefault} from '../util/conversion.js';
```
BYK: I'm not sure if `conversion` is a great module name. How about `user-input` or something like that? Or maybe `sanitization`?

__tests__/util/conversion.js
sth: Do we really care if it's called, as long as it gives correct results? Using a mock would complicate the test and mostly test that the function stays implemented the way it is currently implemented. I'm not sure that would really be preferable.

src/cli/index.js
sth: I don't really like any of those names, but can't come up with anything better.

I did use `conversion` because the functions convert things to boolean and maybe there might be other conversion functions some day that could also be placed in this module. `convert` would be the same but shorter.

If there is no good name and creating a new module for just these two functions is too much I could also add the functions to `util/misc.js`, but I wanted to avoid that.

sth:
Sorry for not coming back to this sooner.

Should I rebase the branch to current master? There are no conflicts when doing so, everything rebases cleanly.

arcanis:
Hey @sth - if you can rebase it would be great, because it would run the tests against master. Let me know when it's done and I'll see to merge your PR! 😃

sth:
@arcanis I rebased and the tests all passed :smile:

arcanis:
Thanks!

PR #4875, opened 2017-11-07
3733 clean cache folders

### LINKED ISSUE #3733: Caching on OS X
This is more of an enquiry, as I noticed yesterday when going through my disk usage that Yarn was taking up a fair bit of space for it's cache. This is not a problem in itself, but what I found odd is that there were seemingly two cache locations, so I was wondering if Yarn is using both, or if one of them is a legacy location that hasn't been cleaned up.

I found cache files both in my user directory under `~/.yarn-cache` as well as in`~/Library/Caches/Yarn.

I cleaned up both and from what it seems only the Library one has been recreated. Is this the one in use now? If so it might be a good idea to have yarn auto-clean the old location if there's still files present there.

sreeramjayan:
fix(cache clean): Clean all cache folders when cache clean is used.
**Summary**

Fixes #3733 
Currently when yarn cache clean is used it only retrieves the first cache folder and deletes it. The change is to retrieve "all" folders and then delete them.

**Test plan**
Making sure the existing test doesn't fail.


src/config.js
```diff
@@ -59,6 +59,7 @@ export type ConfigOptions = {
   registry?: ?string,
 
   updateChecksums?: boolean,
+  cacheFolders?: ?Array<string>,
```
arcanis: Why is `cacheFolders` here?

sreeramjayan:
@arcanis 
Thank you for reviewing the PR. I am unfamiliar with what the following lint errors mean.

> 316:       cacheRootFolder = cacheFolderQuery.folders[0];
                                              ^^^^^^^ property `folders`. Property not found in
316:       cacheRootFolder = cacheFolderQuery.folders[0];
                             ^^^^^^^^^^^^^^^^ object type

> 

Any help would be appreciated.

olingern:
ping @sreeramjayan do you still want to merge this? It would be great to have tests with this, too.

PR #4881, opened 2017-11-08
fix(util): Allow git tag and branch name with hash

### LINKED ISSUE #4880: Yarn fails to resolve git tag which contains # character

```
$ yarn add 'https://github.com/linagora/ical.js.git#v1.1.2-#279-require-twice-protection'
yarn add v1.3.2
warning No license field
[1/4] Resolving packages...
error Couldn't find match for "v1.1.2-" in "_c" for "https://github.com/linagora/ical.js.git".
```

But this works:

```
yarn add 'linagora/ical.js#v1.1.2-#279-require-twice-protection'                       
yarn add v1.3.2
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
└─ ical.js@1.1.2
warning No license field
Done in 14.02s.
```

Arch Linux:
```
node -v
v8.2.1
yarn -v
1.3.2
```

torifat:
**Summary**

In **Git** `#` is a valid character for `tag` and `branch`

```sh
git check-ref-format "tags/v1.1.2-#279-require-twice-protection" && echo "Valid tag" || echo "Invalid tag"
```

Fixes #4880

### DIFF
--- a/__tests__/util/version.js
+++ b/__tests__/util/version.js
@@ -1,0 +1,25 @@
+/* @flow */
+
+import {explodeHashedUrl} from '../../src/util/version.js';
+
+test('explodeHashedUrl', () => {
+  expect(explodeHashedUrl('https://github.com/yarnpkg/yarn.git#v1.2.3')).toEqual({
+    url: 'https://github.com/yarnpkg/yarn.git',
+    hash: 'v1.2.3',
+  });
+
+  expect(explodeHashedUrl('https://github.com/yarnpkg/yarn.git')).toEqual({
+    url: 'https://github.com/yarnpkg/yarn.git',
+    hash: '',
+  });
+
+  expect(explodeHashedUrl('https://github.com/yarnpkg/yarn.git#')).toEqual({
+    url: 'https://github.com/yarnpkg/yarn.git',
+    hash: '',
+  });
+
+  expect(explodeHashedUrl('https://github.com/yarnpkg/yarn.git#v1.2.007#secret-version')).toEqual({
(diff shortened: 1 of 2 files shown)


arcanis:
I'm a bit concerned by the ambiguity this change would add, which could cause issues in other parts of our stack that make the assumption that there can only be a single `#` character in a reference string. I'd personally be in favor of not supporting this syntax.

torifat:
I agree. We can keep the old function as it is and handle multiple `#` inside git resolver. What do you think?

heroandtn3:
@arcanis I don't agree, if yarn is supposed to support git tab, it must support all forms of valid tag.

torifat:
> I agree. We can keep the old function as it is and handle multiple # inside git resolver. What do you think?

@arcanis ☝️

arcanis:
Sorry @torifat, I missed your comment! Problem with that is that it breaks a very common assumption about URLs, so I'm not sure it would work :/

Is it common to have hashes inside a branch name? I understand the "it's valid git" argument, but it's quite inconvenient nevertheless.

BYK:
I think any legit `#` in the tag name should be URL encoded. Does that scenario work right now?

rally25rs:
I opened #5960 before realizing that this PR existed. I'll close mine since it's basically the same as this one.

My opinion is that unless we can prove that this does break something elsewhere in yarn, I don't see a reason to not support it. I don't think the branch/tag is used in a URL so I don't think it would need URL encoded. It's just passed through `git` commands or used when comparing the output of `git refs`.

PR #4883, opened 2017-11-08
feat(workspaces): Adds a `workspaces` command

arcanis:
**Summary**

Adds a command name `workspaces` that display useful data about the workspaces layout. Can be used for debugging, or for integrating other tools on top of Yarn itself. Currently only expose the two following information, but can be easily expanded if we need to add more fields:

- Workspace name
- Workspace location

**Test plan**

Added an automated test.

### DIFF
--- a/__tests__/commands/workspaces.js
+++ b/__tests__/commands/workspaces.js
@@ -1,0 +1,40 @@
+// @flow
+
+import {BufferReporter} from '../../src/reporters/index.js';
+import {run as workspace} from '../../src/cli/commands/workspaces.js';
+import * as reporters from '../../src/reporters/index.js';
+import Config from '../../src/config.js';
+import path from 'path';
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'workspace');
+
+async function runWorkspaces(
+  flags: Object,
+  args: Array<string>,
+  name: string,
+  checkSteps?: ?(config: Config, reporter: BufferReporter) => ?Promise<void>,
+): Promise<void> {
+  const cwd = path.join(fixturesLoc, name);
+  const reporter = new reporters.BufferReporter({stdout: null, stdin: null, isSilent: true});
+
+  const config = await Config.create({cwd}, reporter);
+
+  await workspace(config, reporter, flags, args);
+
+  if (checkSteps) {
+    await checkSteps(config, reporter);
+  }
+}
+
+test('workspaces should list the workspaces', (): Promise<void> => {
+  return runWorkspaces({}, ['info'], 'run-basic', (config, reporter) => {
+    expect(reporter.getBufferJson()).toEqual({
+      'workspace-1': {
+        location: 'packages/workspace-child-1',
+      },
+      'workspace-2': {
+        location: 'packages/workspace-child-2',
+      },
+    });
+  });
+});
--- a/__tests__/commands/workspaces.js
+++ b/__tests__/commands/workspaces.js
@@ -26,7 +26,7 @@ async function runWorkspaces(
   }
 }
 
-test('workspaces should list the workspaces', (): Promise<void> => {
+test('workspaces info should list the workspaces', (): Promise<void> => {
   return runWorkspaces({}, ['info'], 'run-basic', (config, reporter) => {
     expect(reporter.getBufferJson()).toEqual({
       'workspace-1': {
--- a/src/cli/commands/workspaces.js
+++ b/src/cli/commands/workspaces.js
@@ -35,7 +35,7 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
   reporter.log(JSON.stringify(publicData, null, 2), {force: true});
 }
 
-const {run, setFlags, examples} = buildSubCommands('cache', {
+const {run, setFlags, examples} = buildSubCommands('workspaces', {
   async info(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
     await info(config, reporter, flags, args);
   },
--- a/src/cli/commands/workspaces.js
+++ b/src/cli/commands/workspaces.js
@@ -1,0 +1,44 @@
+// @flow
+
+import type Config from '../../config.js';
+import {MessageError} from '../../errors.js';
+import type {Reporter} from '../../reporters/index.js';
+import buildSubCommands from './_build-sub-commands.js';
+
+const invariant = require('invariant');
+const path = require('path');
+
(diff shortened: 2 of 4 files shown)


arcanis:
screen shot 2017-11-08 at 2 22 02 pm

__tests__/commands/workspaces.js
```diff
@@ -0,0 +1,44 @@
...
+  } catch (err) {
+    throw new Error(`${err && err.stack}`);
+  }
+}
+
+test('workspaces should list the workspaces', (): Promise<void> => {
+  return runWorkspaces({}, [], 'run-basic', (config, reporter) => {
+    expect(reporter.getBufferJson()).toEqual({
+      'workspace-1': {
+        location: `${path.resolve(fixturesLoc)}/run-basic/packages/workspace-child-1`,
```
BYK: Assigning `path.resolve(fixturesLoc)` to a variable would make this easier to read.

__tests__/commands/workspaces.js
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    const config = await Config.create({cwd}, reporter);
+
+    await workspace(config, reporter, flags, args);
+
+    if (checkSteps) {
+      await checkSteps(config, reporter);
+    }
+  } catch (err) {
+    throw new Error(`${err && err.stack}`);
```
BYK: Why are we doing this instead of rethrowing the error or just removing the whole `try`/`catch` block?

src/cli/commands/workspaces.js
```diff
@@ -0,0 +1,36 @@
...
+
+import type Config from '../../config.js';
+import {MessageError} from '../../errors.js';
+import type {Reporter} from '../../reporters/index.js';
+
+const invariant = require('invariant');
+
+export function setFlags(commander: Object) {}
+
+export function hasWrapper(commander: Object, args: Array<string>): boolean {
```
BYK: Do we really need to define this function? I thought it defaulted to `true`.

src/cli/commands/workspaces.js
```diff
@@ -0,0 +1,36 @@
...
+
+export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
+  const {workspaceRootFolder} = config;
+
+  if (!workspaceRootFolder) {
+    throw new MessageError(reporter.lang('workspaceRootNotFound', config.cwd));
+  }
+
+  const manifest = await config.findManifest(workspaceRootFolder, false);
+  invariant(manifest && manifest.workspaces, 'We must find a manifest with a "workspaces" property');
```
BYK: _We need a manifest at a higher level with `workspaces` defined in it._

src/reporters/buffer-reporter.js
```diff
@@ -27,4 +27,12 @@ export default class BufferReporter extends JSONReporter {
   getBuffer(): Buffer {
     return this._buffer;
   }
+
+  getBufferText(): String {
+    return this._buffer.map(({data}) => String(data)).join(``);
+  }
+
+  getBufferJson(): any {
```
BYK: This name is misleading since the returned thing is not JSON. Also, I don't understand the need to go through casting things to a JSON string and back into an object. Why not use the `JSONReporter` at this point and just parse the whole thing?

arcanis: We already have `readJson` in a few places, which does not return json either ;) What would you suggest?

> Why not use the JSONReporter at this point and just parse the whole thing?

- The JSON reporter wraps what is written in irrelevant fields (`type` and `data`)
- The `data` field is a string, so it would have to be parsed anyway.

src/cli/commands/workspaces.js
arcanis: The wording doesn't matter much here since it's an invariant

src/cli/commands/workspaces.js
arcanis: I think @voxsim worked to standardize the commands a few months ago, and made this required in practice if not in theory. I'm fine with it, it feels less magic (even if the whole wrapper concept is a bit opaque).

__tests__/commands/workspaces.js
arcanis: Good question. I took this code from the `workspace` test, but I can't remember why I'm doing this.

arcanis:
> I think the code is good but I think we need to make this command called yarn workspaces list or yarn workspaces resolve so we can build new functionality like yarn workspaces run etc.

Btw I agree with this, I'll rename it to `workspaces info`. That being said, `yarn workspaces run` is actually already possible via `yarn workspace <pkg-name> run ...` (notice the singular `workspace`).

src/reporters/buffer-reporter.js
BYK: I'm sad because we don't have a good way of ingesting JSON output easily :(

__tests__/commands/workspaces.js
```diff
@@ -0,0 +1,40 @@
...
+  const config = await Config.create({cwd}, reporter);
+
+  await workspace(config, reporter, flags, args);
+
+  if (checkSteps) {
+    await checkSteps(config, reporter);
+  }
+}
+
+test('workspaces should list the workspaces', (): Promise<void> => {
```
BYK: Test name is misleading now. It should say `workspaces info should list the workspaces`

src/cli/commands/workspaces.js
```diff
@@ -0,0 +1,44 @@
...
+  for (const workspaceName of Object.keys(workspaces)) {
+    publicData[workspaceName] = {
+      location: path.relative(config.lockfileFolder, workspaces[workspaceName].loc),
+    };
+  }
+
+  reporter.log(JSON.stringify(publicData, null, 2), {force: true});
+}
+
+const {run, setFlags, examples} = buildSubCommands('cache', {
```
BYK: Did you mean `workspaces` instead if `cache`?

Since the tests missed this mistake, should we change them to be more like full integration tests?

arcanis: Not in this PR, but yeah, it's the main issue I have with these tests: they don't test the CLI (except in `__test__/integration.js`, but it spawns actual processes).

arcanis: Funnily, it works even with this mistake :p I guess this string is only used for the usage string.

rolandjitsu:
Note that this is still not documented, see [workspaces](https://yarnpkg.com/en/docs/cli/workspaces). But there may already be an issue opened for that.

PR #4887, opened 2017-11-08
ci(appveyor): Drop Node4 builds from AppVeyor

BYK:
**Summary**

Stops building for Node 4 on AppVeyor to save some time. Node 8.9.0 is the new LTS right now and
there's little value building for Node 4 anyway. It also updates the test script to use `yarn
test-only` like other CI builds.

**Test plan**

AppVeyor builds should pass.


BYK:
I'm abandoning this since AppVeyor builds keep failing. I'm guessing it has something to do with the command change but I've also realized that we would need Node4 for legacy builds on Windows.

See you in April 2018 when Node 4 support will be dropped by Node folks.

PR #4888, opened 2017-11-08
change registry from quote style to no-quote style

### LINKED ISSUE #4637: added config item is not compatible with the original config items
Yarn v1.1.0
Add config with:
`yarn config set registry https://registry.npm.taobao.org`
List config items:
image

The added `registry` item is not quoted while the original ones are quoted. But npm config's default style is no quote. I think coordinate with npm's will be better.

dc3671:
**Summary**

No.

Fix #4637 to make registry style more unified.

**Test plan**

Simple reformat change. No need to test I think.

### DIFF
--- a/src/registries/yarn-registry.js
+++ b/src/registries/yarn-registry.js
@@ -19,7 +19,7 @@ export const DEFAULTS = {
   version-git-message: 'v%s',
 
   init-version: '1.0.0',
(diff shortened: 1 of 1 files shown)


arcanis:
This PR is invalid and cannot be merged. The quotes are required because `-` is not a character allowed in identifiers.

PR #4891, opened 2017-11-09
feat(run): Show a "no bin scripts" or "no scripts available" message

maxchehab:
**Summary**

This adds the additional warning if there are no bin scripts or if there are no scripts available. If no scripts are available, scripts will not be listed.

I created this pull request when running `yarn run` without any scripts specified, yarn would try to ask the user to specify which script.

To view this feature in the CLI, run `yarn run` inside a project without any bin or scripts available.
For example:
```bash
$ yarn run
yarn run v1.3.2
error No command specified.
error There are no binary scripts available.
error There are no scripts specified inside package.json.
Done in 0.13s.
```

**Test plan**

This new functionality is tested with the `returns noScriptsAvailable and noBinAvailable with no bins and scripts` test.

This test checks if the implemented error messages are thrown.


__tests__/commands/run.js
```diff
@@ -133,3 +133,16 @@ test('adds quotes if args have spaces and quotes', (): Promise<void> => {
     expect(execCommand).toBeCalledWith(...args);
   });
 });
+
+test('returns noScriptsAvailable and noBinAvailable with no bins and scripts', (): Promise<void> => {
+  return runRun([], {}, 'no-scripts', (config, reporter): ?Promise<void> => {
+    const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
```
BYK: Please just use `const reporter` :)

__tests__/commands/run.js
```diff
@@ -133,3 +133,16 @@ test('adds quotes if args have spaces and quotes', (): Promise<void> => {
...
+
+    // Emulate run output
+    rprtr.error(rprtr.lang('commandNotSpecified'));
+    rprtr.error(rprtr.lang('noBinAvailable'));
+    rprtr.error(rprtr.lang('noScriptsAvailable'));
+
+    expect(reporter.getBuffer()).toEqual(rprtr.getBuffer());
```
BYK: You can skip all of the "emulate run output" part and do `expect(reporter.getBuffer()).toMatchSnapshot()`.

__tests__/commands/run.js
```diff
@@ -133,3 +133,16 @@ test('adds quotes if args have spaces and quotes', (): Promise<void> => {
...
   });
 });
+
+test('returns noScriptsAvailable and noBinAvailable with no bins and scripts', (): Promise<void> => {
+  return runRun([], {}, 'no-scripts', (config, reporter): ?Promise<void> => {
+    const rprtr = new reporters.BufferReporter({stdout: null, stdin: null});
+
+    // Emulate run output
+    rprtr.error(rprtr.lang('commandNotSpecified'));
+    rprtr.error(rprtr.lang('noBinAvailable'));
```
BYK: Can you split this test so we test each case individually? This way we know we don't say `noScriptsAvailable` when we have them but we don't have any binaries for instance.

maxchehab:
No problem! Glad I can contribute 😃

PR #4903, opened 2017-11-11
fix(request) Fix https_proxy env var

### LINKED ISSUE #4885: Yarn 1.3.2 incorrectly bypasses HTTP(S)_PROXY
*Bug*

1. Set HTTP_PROXY and HTTPS_PROXY environment variables to a valid proxy
2. `yarn install` attempts to access yarnpkg.com on port 443, ignoring the proxy settings

1. Yarn instead makes HTTPS requests via the proxy set in the HTTPS_PROXY environment variable. Yarn 1.2.1 does this correctly.

Node 4.8.5
Yarn 1.3.2
Ubuntu 16.04.1

rally25rs:
**Summary**

Fixes #4885. In #4761 I started always passing the proxy to the `request` library to prevent it from falling back to env vars HTTPS_PROXY and HTTP_PROXY. At the time I thought that Yarn would pick up these env vars through its config system and pass them down itself.

It turns out this was not true and caused #4885. Now handling an explicit `false` in the request-manager to allow it to override the `https_proxy` if set. Otherwise, resume allowing request library to pick up the env vars.

**Test plan**

Manually tested. Since the actual requests are proxies under unit test, this is difficult to unit test.

### DIFF
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -148,6 +148,8 @@ export default class RequestManager {
 
     if (opts.httpsProxy === '') {
       this.httpsProxy = opts.httpProxy || '';
+    } else if (opts.httpsProxy === false) {
+      this.httpsProxy = false;
     } else {
       this.httpsProxy = opts.httpsProxy || '';
     }
@@ -412,7 +414,14 @@ export default class RequestManager {
     if (params.url.startsWith('https:')) {
       proxy = this.httpsProxy;
     }
-    params.proxy = String(proxy);
+
+    if (proxy) {
+      params.proxy = String(proxy);
+    } else if (proxy === false) {
+      // passign empty string prevents the underlying library from falling back to the env vars.
+      // an explicit false in the yarn config should override the env var. See #4546.
+      params.proxy = '';
+    }
 
     if (this.ca != null) {
       params.ca = this.ca;
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -416,9 +416,11 @@ export default class RequestManager {
     }
 
     if (proxy) {
+      // if no proxy is set, do not pass a proxy down to request.
+      // the request library will internally check the HTTP_PROXY and HTTPS_PROXY env vars.
       params.proxy = String(proxy);
     } else if (proxy === false) {
-      // passign empty string prevents the underlying library from falling back to the env vars.
+      // passing empty string prevents the underlying library from falling back to the env vars.
       // an explicit false in the yarn config should override the env var. See #4546.
       params.proxy = '';
     }


src/util/request-manager.js
```diff
@@ -412,7 +414,14 @@ export default class RequestManager {
     if (params.url.startsWith('https:')) {
       proxy = this.httpsProxy;
     }
-    params.proxy = String(proxy);
+
+    if (proxy) {
+      params.proxy = String(proxy);
+    } else if (proxy === false) {
+      // passign empty string prevents the underlying library from falling back to the env vars.
```
timja-kainos: typo on passing

src/util/request-manager.js
```diff
@@ -412,7 +414,16 @@ export default class RequestManager {
...
+
+    if (proxy) {
+      // if no proxy is set, do not pass a proxy down to request.
+      // the request library will internally check the HTTP_PROXY and HTTPS_PROXY env vars.
+      params.proxy = String(proxy);
+    } else if (proxy === false) {
+      // passing empty string prevents the underlying library from falling back to the env vars.
+      // an explicit false in the yarn config should override the env var. See #4546.
+      params.proxy = '';
+    }
```
kaylieEB: what happens to other falsey variables like 0 or undefined? Maybe you should flip the order?
```
if (proxy === false) {
   params.proxy = '';
} else {
   params.proxy = String(proxy);
}
``` 
or if they're meant to be empty strings
```
params.proxy = proxy ? String(proxy) || '';
```

rally25rs: Based on some logic above:
```
    if (opts.httpsProxy === '') {
      this.httpsProxy = opts.httpProxy || '';
    } else if (opts.httpsProxy === false) {
      this.httpsProxy = false;
    } else {
      this.httpsProxy = opts.httpsProxy || '';
    }
```

so `httpsProxy` should be one of `''`, `false`, `'https://wherever`

In the case of `''` we don't want to set any `params.proxy` at all because the `request` library will check `hasOwnProperty('httpsProxy')` on the object we pass it. If that property exists at all (even if set to a falsey value) then it will not check the env vars.

So basically the logic here is:

```
if ''
  don't set a proxy (request will fall back to env vars)
if false
  set a proxy to '' (request will not use env vars, and not use any proxy)
if 'some non-falsy string'
  set a proxy to that string (request will use this as the proxy)
```

Hopefully that makes sense...

kaylieEB: I see. Thanks for explaining!

lennyburdette:
@kaylieEB @BYK Glad to see this fixed—thanks for doing this! Would either of you be able to say when this would be released? I haven't found anything on github or yarnpkg.com about Yarn's release schedule or cadence. Thanks!

Issue #4500, opened 2017-09-19
yarn global upgrade-interactive doesn't work


dhcmrlchtdj:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
`global upgrade` and `global upgrade-interactive` seem not work.

**If the current behavior is a bug, please provide the steps to reproduce.**

```
$ yarn global add prettier@1.6.1
$ yarn global upgrade-interactive --latest
```

**What is the expected behavior?**
Yarn should report that prettier can be upgraded to 1.7.

**Please mention your node.js, yarn and operating system version.**
node v8.5.0
yarn 1.0.2
macOS 10.12.6

publicarray:
This is probably due to #4443 and #4390

amannn:
Seems like this is by intention: https://github.com/yarnpkg/yarn/issues/4439#issuecomment-330325134

`yarn upgrade-interactive --latest` works for me.

dhcmrlchtdj:
> passing the -L/--latest flag will tell upgrade-interactive to ignore the package.json specified range and use the latest as specified by the registry

The `--latest` flag works for `update / upgrade-interactive` but not `global update`.

https://github.com/yarnpkg/yarn/blob/v1.0.2/src/cli/commands/global.js#L282
https://github.com/yarnpkg/yarn/blob/v1.0.2/src/cli/commands/upgrade-interactive.js#L15
https://github.com/yarnpkg/yarn/blob/v1.0.2/src/cli/commands/_build-sub-commands.js#L26

jmsmrgn:
@dhcmrlchtdj having same issue, flag not working with global

dwiyatci:
I too can confirm this bug on yarn v1.2.1.

`yarn global upgrade-interactive --latest` doesn't work,
`yarn upgrade-interactive --latest` does.

🤷‍♂️

ringrid:
I use yarn v1.3.2, same to me.

dertuerke:
same here
`yarn global upgrade-interactive` not working 

```
yarn global v1.3.2
error Outdated lockfile. Please run `yarn install` and try again.
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

PR #4915, opened 2017-11-14
Add latest flag to global upgrade interactive

kaylieEB:
**Summary**
Fixes https://github.com/yarnpkg/yarn/issues/4500#issuecomment-342616138.

The global subcommands currently only allow flags from the two places below to be passed from the CLI
https://github.com/yarnpkg/yarn/blob/8918b6cff99e037f3ab56523ebfde96e0c19cdae/src/cli/index.js#L65-L113 and https://github.com/yarnpkg/yarn/blob/8918b6cff99e037f3ab56523ebfde96e0c19cdae/src/cli/commands/global.js#L289

which meant [by the time it gets to upgrade interactive](https://github.com/yarnpkg/yarn/blob/8918b6cff99e037f3ab56523ebfde96e0c19cdae/src/cli/commands/upgrade-interactive.js#L37-L38), commander didn't correctly pass in `latest` as a valid key in the flags object. As a result, whenever `yarn global upgrade-interactive` was called, `--latest` was ignored.

**Test plan**
This is a pretty low risk change given that global modules don't need to read off package.json or a lockfile, but the issue did reveal some future tasks for a) cleaning up flags for each command and b) finding a better testing strategy for upgrade interactive as these tests are pretty sparse.

*BEFORE*
screen shot 2017-11-13 at 4 43 00 pm
*AFTER*
screen shot 2017-11-13 at 4 42 50 pm

### DIFF
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -287,4 +287,5 @@ export {run};
 export function setFlags(commander: Object) {
   _setFlags(commander);
   commander.option('--prefix <prefix>', 'bin prefix to use to install binaries');
+  commander.option('--latest', 'upgrade to the latest version of packages');
 }


arcanis:
> a) cleaning up flags for each command

Yep, the whole cli is too much complex for what we're doing. I'd love to have the time someday to rework it to simply use yargs or something, and stop reimplementing the world on top of commander :)

PR #4921, opened 2017-11-14
Fix for broken renaming of scoped packages

Ambroos:
**Summary**

I noticed that trying to rename a scoped package fails. After some investigation it turns out the regex in exotics/registry-resolver did not know what to do with scoped packages like "npm:@babel/register", or "npm:@babel/register@7.0.0-beta.32".

I simply had to add an `@?` for the `@` right before the name capture group so the `@` wouldn't result in everything after it being thrown into the version capture group (leaving name empty). That was the only thing required to get things to work.

I've added a test for it to the existing `install renamed packages` test, it seemed to make sense. As test package I used `@turf/helpers` since it'd been used as test package in other tests as well.

Context: From Babel 7 onwards packages are moved to the @babel scope. I was experimenting with upgrading one of our internal projects to Babel 7. The package `babel-register` is used as an optional peer dependency in other projects (like webpack, for automatically handling config.babel.js) via interpret. By adding @babel/register renamed to babel-register I could use the new version without having to wait for other package authors to update their packages.

**Test plan**

Before this PR:
```
Ambroos@Mac:~/Dev/temp $ yarn add --dev babel-register@npm:@babel/register@7.0.0-beta.32
yarn add v1.3.2
info No lockfile found.
error Received malformed response from registry for undefined. The registry may be down.
```

After:
```
Ambroos@Mac:~/Dev/temp $ alias yarn="node ~/Dev/yarn/lib/cli/index.js"
Ambroos@Mac:~/Dev/temp $ yarn add --dev babel-register@npm:@babel/register@7.0.0-beta.32
yarn add v1.3.2
info No lockfile found.
[2/4] 🚚  Fetching packages...
warning " > babel-register@7.0.0-beta.32" has unmet peer dependency "@babel/core@7.0.0-beta.32".
success Saved lockfile.
success Saved 20 new dependencies.
├─ babel-register@7.0.0-beta.32
...
└─ source-map@0.5.7
✨  Done in 1.56s.
Ambroos@Mac:~/Dev/player/temp $```


Ambroos:
There seems to have been a problem with the test runners, as the tests should all pass. I can't rerun tests and don't want to make unnecessary changes. Could someone re-run them?

waleedahmed3045:
Can the builds be triggered again for this PR? As it was a circleci issue. 
image

https://discuss.circleci.com/t/unable-to-spin-up-environment/18109

Ambroos:
I have rebased my work on master, tests seem to be (mostly) OK now.

Gudahtt:
It looks like this has been fixed already, in this PR: https://github.com/yarnpkg/yarn/pull/5229

Issue #3373, opened 2017-05-10
Possibility to add warning when running yarn in directory containing node_modules but missing yarn.lock


martinek:
Would it be possible to add some kind of warning when running yarn inside directory where `node_modules` already exist, but there is no `yarn.lock` file?

The only use-case for this should be when migrating from npm to yarn. I realized yarn updated all my `node_modules` only after I ran `yarn` command and then it was too late to search for `yarn import`.

It would be nice to get warning, that running `yarn` or `yarn install` without `yarn.lock` will result in `node_modules` being downloaded and updated.

I realize this experience is depending on environment, but the destructive nature of such simple mistake could warrant some kind of protection against it.

Alternatively, running `yarn` command in this specific case, could by default fallback to `yarn import` command instead of `yarn install`.

Also `yarn import` command could be mentioned on https://yarnpkg.com/lang/en/docs/migrating-from-npm/ page.

bestander:
That sounds like a nice idea.
Send a PR.

rubenazevedoff:
So, whenever you're running yarn, but there is a node_modules in their but no yarn.lock, should give a warning?
Tomorrow morning i will PR this :)

bestander:
If there is a node_modules then there should be a package.json so it should
be fine.

rubenazevedoff:
But this can only happen when you're making the import  right?

bestander:
Yeah or if you don't care that preserve existing node_modules and just
start using yarn.

jscoobyced:
Was this PR submitted? Should the warning be blocking? or should it prompt to "Continue? [Y/n]" ?

arcanis:
No PR has been submitted as far as I know. The warning should not be blocking, we don't want to prevent people from using Yarn :) I think a prompt defaulting to Yes would be fine.

msramalho:
Me and a group of two more are going to address this issue over the next week :smile:

PR #4923, opened 2017-11-14
Fixed issue #3373

msramalho:
### Summary

With this pull request, we present a solution for issue [#3373](https://github.com/yarnpkg/yarn/issues/3373). 

**Motivation:** Besides being a good first issue to start understanding yarn, this PR is useful for users transitioning from [npm](https://www.npmjs.com/) to yarn by preventing unwanted changes on the :file_folder:`node_modules`. 
**Solves:** Issue [#3373](https://github.com/yarnpkg/yarn/issues/3373). According to which users should be warned when they run `yarn` on a project that has a :file_folder:`node_modules` and does not have a :lock: `yarn.lock` file. 
### Test plan
The test environment was a project with a `node_modules` folder and no `yarn.lock` file.

After running `yarn` we present a question to the user:
```
>yarn
yarn install v1.3.2
info No lockfile found.
question
  There is already a node_modules folder and no yarn.lock.
  This probably means you are transitioning from npm to yarn.
  Proceeding means the node_modules folder will be overwritten by yarn.
  Do you want to proceed? (yes/no):
```
When the answer is `yes` (or a valid equivalent) yarn proceeds normally.

When the answer is `no` (or a valid equivalent) yarn exits normally. 

We tested with and without the `yarn.lock` file and it worked every time.

### DIFF
--- a/src/lockfile/index.js
+++ b/src/lockfile/index.js
@@ -128,7 +128,7 @@ export default class Lockfile {
       const nodeModulesLoc = path.join(dir, NODE_MODULES_FOLDER);
       if (await fs.exists(nodeModulesLoc)) {
         if (!await reporter.questionAffirm(reporter.lang('lockfileModulesConflict'))) {
-          reporter.footer();
+          reporter.footer(false);
           throw new MessageError(reporter.lang('operationAborted'));
         }
       }


msramalho:
Hi @rally25rs, I will take this into account, thank you for the relevant information. The PR will be updated somewhere over the next week!

msramalho:
Ok, since there were some problems with my fork I am going to ground zero it and redo the work, also taking into account your suggestion!

PR #4931, opened 2017-11-15
chore(lint): Rename file to .eslintrc.json

mc10:
**Summary**

.eslintrc is deprecated:
https://eslint.org/docs/user-guide/configuring#configuration-file-formats

**Test plan**

ESLint should work the same as before.


BYK:
Awesome, thanks!

PR #4907, opened 2017-11-12
Check whether a git repo is up-to-date for upgrade-interactive

mertkahyaoglu:
Fixes #4896

PR #4948, opened 2017-11-17
test(integration): Fix left-pad version used in a test with snapshots

BYK:
**Summary**

We had a test using left-pad without pinning its version and since now there's a new version of
left-pad, the snapshots are failing. This PR fixes the issue by pinning the version of left-pad in
the test.

**Test plan**

Tests should pass.


mertkahyaoglu:
That's great. I was wondering why my snapshot test was failing. #4907

PR #4956, opened 2017-11-18
fix(pack): make globs match dotfiles

rhendric:
**Summary**

This makes globs in the files array in package.json and in .*ignore files match dotfiles, consistent with npm's behavior and how git processes .gitignore lines.

**Test plan**

See automated test.


arcanis:
Thanks!

Issue #4929, opened 2017-11-14
Invalid string length


jycouet:
Hello, 

I'm using yarn `v1.3.2` and I have a bug.

**What is the current behavior?**
I'm running yarn from js like this:
```js    
    ({ stdout, stderr } = await exec('yarn install --ignore-scripts  --ignore-engines', {
      cwd: tmpDir,
      shell: true,
      env: { NODE_ENV: 'dev', PATH: process.env.PATH },
    }));
    logger.debug(`yarn stdout:\n${stdout}`);
    logger.debug(`yarn stderr:\n${stderr}`);
```
And I often get: `"stderr": "error An unexpected error occurred: \"Invalid string length\".\n"`

Here is my trace:
```
Trace: 
  RangeError: Invalid string length
      at tokenise (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59559:16)
      at tokenise.next (<anonymous>)
      at Parser.next (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59657:30)
      at Parser.parse (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59737:14)
      at Parser.parse (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59761:28)
      at parse (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59850:17)
      at module.exports.exports.default (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:59443:96)
      at D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:1976:63
      at Generator.next (<anonymous>)
      at step (D:\Dev\jycouet_renovate\node_modules\yarn\lib\cli.js:92:30)
```

You you can see the package.json that I have (nothing really special?):
```json
{
  "name": "vsts_tmpReno",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@angular/animations": "4.4.6",
    "@angular/cdk": "2.0.0-beta.12",
    "@angular/common": "4.4.6",
    "@angular/compiler": "4.4.6",
    "@angular/core": "4.4.6",
    "@angular/flex-layout": "2.0.0-rc.1",
    "@angular/forms": "4.4.6",
    "@angular/http": "4.4.6",
    "@angular/material": "2.0.0-beta.12",
    "@angular/platform-browser": "4.4.6",
[...]

PR #4965, opened 2017-11-20
refactor(lockfile): remove string concatenation from parsing code

rhendric:
Replace all instances of building strings incrementally using += with more efficient methods.

**Summary**

This is motivated by #4929, as one cause of "Invalid string length" is trying to concatenate with too large of a string, and the stack trace there points to a string concatenation operation in this function. It's probably not the root cause of #4929, so I wouldn't say this should close that issue until we understand what in the lockfile was causing runaway string concatenation. But it does address the proximate issue and helps with general code health and probably performance as well, if only in a small way.

**Test plan**

Automated tests should cover regressions. Manually review code to ensure no lingering string concatenation.

### DIFF
--- a/src/lockfile/parse.js
+++ b/src/lockfile/parse.js
@@ -69,44 +69,43 @@ function* tokenise(input: string): Iterator<Token> {
     } else if (input[0] === '#') {
       chop++;
 
-      let val = '';
-      while (input[chop] !== '\n') {
-        val += input[chop];
-        chop++;
+      let nextNewline = input.indexOf('\n', chop);
+      if (nextNewline === -1) {
+        nextNewline = input.length;
       }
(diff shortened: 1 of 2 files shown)


src/lockfile/parse.js
```diff
@@ -69,44 +69,43 @@ function* tokenise(input: string): Iterator<Token> {
...
+          chop = indentSize;
+          yield buildToken(TOKEN_TYPES.indent, indentSize / 2);
         }
       } else {
         chop++;
       }
     } else if (input[0] === '"') {
-      let val = '';
-
-      for (let i = 0; ; i++) {
```
Gudahtt: Huh. It looks like this would be an infinite loop, in the case where the closing quote was not present.

Issue #3882, opened 2017-07-10
Workspaces - Is it possible to exclude some node_modules from beeing hoisted to the root node_modules folder?


thomasfr:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Please mention your node.js, yarn and operating system version.**
yarn --version: 0.27.5
node -v: v6.11.0

We are currently using `hoist` feature from lerna. I tried to move to yarn workspaces and it works way faster! Awesome, thank you!
But i also need the ability (sadly) to exclude some node_modules from not beeing hoisted to the root node_modules folder but be physically installed in the package i configured it as dependency. (The same way the `--no-hoist` arguments are working for lerna. )
The reason is, sadly, that some dependencies we use, are trying to require some dependencies with an absolute path from `packages/<package>/node_modules` directly. So i have to exclude some.
Is there a way to configure some sort of exclude glob for workspaces?

BYK:
@thomaschaaf is this a feature request? :) (since you haven't mentioned that in the first question)

BYK:
//cc @bestander

thomasfr:
@BYK If its not yet possible (maybe it is not documented?) than it is a feature request.

bestander:
Would a symlink to the hoisted location work?

thomasfr:
lerna installs it directly when using `--no-hoist`. But i think a symlink should also work, i can try that out later. I will let you know

seansfkelley:
Another use-case for disabling hoisting is that it helps prevent devs from creating accidentally-malformed packages.

I worked on a Lerna repo that had hoisting enabled. A number of the packages were erroneously using hoisted dependencies they had not specified because the code compiled/ran fine due to the way in which Node searches for modules -- namely, by traversing up the file hierarchy until it finds something in any ancestor's `node_modules` with the right name. In turn, this meant that packages would be published broken, because consumers wouldn't have the complete, true list of dependencies at install time.

For our use-case, hoisting bought us nothing while introducing more pitfalls, so I would love a way to  disable hoisting on any given monorepo -- indeed, I would actually argue that no-hoist should be the default for the reasons stated above and by the principle of least surprise.

bestander:
All good points here.
We can discuss whether the no-hoist should be default later.
For now I can't dedicate enough time to work on this feature and help is needed.

[...]

Issue #86, opened 2016-06-29
`kpm install` then `kpm check` reports errors


shayne:
```
$ kpm install
fbkpm install v0.4.0
info Read lockfile fbkpm.lock
[1/5] ⏩  Rehydrating dependency graph...
[2/5] 📦  Fetching packages...
[3/5] ✅  Checking package compatibility...
[4/5] 🔗  Linking dependencies...
[5/5] 📃  Running install scripts...
✨  Done in 25.80s. Peak memory usage 205.83MB.
```

```
$ kpm check
fbkpm check v0.4.0
error Module not installed: node_modules/mkdirp/node_modules/minimist
error Module not installed: node_modules/escodegen/node_modules/source-map
error Module not installed: node_modules/jsbn
error Module not installed: node_modules/tweetnacl
error Module not installed: node_modules/jodid25519
error Module not installed: node_modules/ecc-jsbn
```

PR #4979, opened 2017-11-22
nohoist baseline implementation

connectdotz:
**Summary**

this is the implementation of [RFC #86](https://github.com/yarnpkg/rfcs/pull/86) for #3882.

To prevent scope creep, this PR addressed only the issues directly impacting "install and run react-native in yarn workspaces with nohoist", which included:

- `install`: including nohoist config, matching and flat tree generation. Many commands utilize the same tree generation, if there is no conflicting assumption, hopefully they _should_ just work...
- `why`: bug fixes and nohoist enhancement, it will now report all locations of the packages with the full dependency tree in one go. this can help self-diagnosis if nohoist went wrong.
- `add`: replaced manual seeding with Manifest update and let install do the rest.  The code change is not ideal as I tried not to change too much in this PR. Suggest later to refactor to make it cleaner, more like _remove.js_ ... 
- also added many tests and fixtures

for more details please reference the [RFC](https://github.com/yarnpkg/rfcs/pull/86).

note: 
- lots of files in this PR, most of them are tests and fixtures.
- I was not able to get a clean test run on my machine due to timeouts. Running the failed tests in isolation all passed, hopefully, CI will be fine... 
- nohoist did introduce a fundamental assumption shift in the system, which I've seen yarn commands yield unexpected results when operating in the nohoist workspace. I've only fixed those mentioned above, many might still exist. I would anticipate more follow-up PRs to address these issues as they come... maybe consider keeping nohoist in a separate branch or a beta channel to let it bake for a while...?

**Test plan**

**for "react-native init"**

create a yarn workspaces and add the following config to the root package.json
```
// project root package.json

"workspaces": {
   "packages": ["packages/*"],
    "nohoist": ["**/react-native", "**/react-native/**"]
}
```
then install react-native following the react-native instruction. It should just work....

you can also move the nohoist config to the workspace's package.json, if it is a private package:
```
// workspace package.json

"workspaces": {
    "nohoist": ["react-native", "react-native/**"]
}
```
it should produce the same result

**for "create-react-native-app"**

additional issues for this scenario:

1. create-react-native-app does not use yarn consistently in the workspaces setting, I submitted a PR ([create-react-native-app/pull/503](https://github.com/react-community/create-react-native-app/pull/503)) to address this, hopefully, it will be merged in soon. 
1. the react-native-scripts and Expo don't play well in workspaces, therefore, I had to add them to the nohoist list as well: 
```
// project root package.json

"workspaces": {
    "packages": ["packages/*"],
    "nohoist": [
        "**/react-native", "**/react-native/**",
        "**/react-native-scripts", "**/react-native-scripts/**",
        "**/expo", "**/expo/**"
    ]
```
then you should be able to create and run the default expo app as documented.


jmaher409:
@connectdotz thanks for making this change. This was a major issue for us with chromedriver and electron-chromedriver. Is there any timeline for this being merged to master?

achingbrain:
This PR is exactly what my team needs - we're building a lerna/yarn workspaces-based monorepo that contains shared libraries, web apps and an electron app and so have some weird requirements around which packages should have some or all of their dependencies hoisted.

It'd be great if this feature could be merged.

connectdotz:
@aracarie these are all the changes we discussed in RFC, the ball is in your court now...

connectdotz:
@arcanis, any progress on the review of this PR? is there something I should address to get this moving? we have been running with the local yarn build for almost a month, just wondering when we can use the official build...

burabure:
sorry to drop in here, but just wanted to say we reeeeeeaaaaaaaaaaaaally need this. We have a super hacky patch now to avoid the issues but it's suboptimal at best.

connectdotz:
ping @arcanis @bestander

bestander:
Acknowledged.
Will have a look asap

__tests__/commands/add.js
```diff
@@ -42,20 +42,20 @@ const runAdd = buildRun.bind(
   },
 );
 
-test.concurrent('add without --ignore-workspace-root-check should fail on the workspace root', async () => {
-  await runInstall({}, 'simple-worktree', async (config, reporter): Promise<void> => {
+test.concurrent('add without --ignore-workspace-root-check should fail on the workspace root', (): Promise<void> => {
```
bestander: any reason for switching from async to promises?
Afaik async is used throughout the code and it would be better to keep it as is

__tests__/config.js
```diff
@@ -42,3 +43,185 @@ test('getOption does not change empty-string when resolve=false', async () => {
...
+    manifest.private = false;
+    expect(getNohoist(config.getWorkspaces(manifest, false))).toBeUndefined();
+    expect(mockReporter.numberOfCalls()).toEqual(1);
+    expect(mockReporter.findCalls('workspacesNohoistRequirePrivatePackages').length).toEqual(1);
+  });
+});
+
+class MockReporter extends BufferReporter {
+  lang = jest.fn();
+  // lang(key: LanguageKeys, ...args: Array<mixed>): string {
```
bestander: can be removed?

__tests__/package-hoister.js
```diff
@@ -311,3 +334,509 @@ test('will hoist packages under subdirectories when they cannot hoist to root',
   expect(result).toContainPackage('c@1.0.0', atPath('a', 'node_modules', 'c'));
   expect(result).toContainPackage('d@1.0.0', atPath('a', 'node_modules', 'd'));
 });
+
+describe('nohoist', () => {
+  function _showResults(results: HoistManifestTuples) {
```
bestander: function not used?

src/types.js
```diff
@@ -129,7 +134,7 @@ export type Manifest = {
   files?: Array<string>,
   main?: string,
 
-  workspaces?: Array<string>,
+  workspaces?: Array<string> | WorkspacesConfig,
```
bestander: any reason to keep the old structure?
Maybe use WorkspacesConfig for hoist cases too?

src/reporters/lang/en.js
```diff
@@ -183,7 +183,12 @@ const messages = {
     'Running this command will add the dependency to the workspace root rather than workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).',
   workspacesRequirePrivateProjects: 'Workspaces can only be enabled in private projects',
   workspacesDisabled:
-    'Your project root defines workspaces but the feature is disabled in your Yarn config. Please check "workspaces-experimental" in your .yarnrc file.',
+    'Your project root defines wokkspaces but the feature is disabled in your Yarn config. Please check "workspaces-experimental" in your .yarnrc file.',
```
bestander: workspaces

src/config.js
```diff
@@ -321,6 +322,7 @@ export default class Config {
       this._cacheRootFolder = String(cacheRootFolder);
     }
     this.workspacesEnabled = this.getOption('workspaces-experimental') !== false;
+    this.workspacesNohoistEnabled = this.getOption('workspaces-nohoist-experimental') !== false;
```
bestander: After thinking about this I think we don't really need to hide this feature behind a flag.
Workspaces is still behind "experimental" flag :)

src/types.js
connectdotz: this is merely for backward compatibility so people who didn't need the new feature doesn't need to be forced to update their package.json... Internally the code is all converted to use WorkspaceConfig.

src/config.js
connectdotz: I am happy to remove it but it was requested specifically by @arcanis cited `workspaces-experimental` will be removed soon, just want to make sure we are all on the same page...

src/types.js
bestander: Ah, ok, let's keep it backwards compat, I missed that this is read from package.json

src/config.js
bestander: ok, let's keep it

connectdotz:
@bestander thx for your prompt review, the fixes are all in.

bestander:
Could you rebase?
it shows src/config.js conflicting

connectdotz:
@bestander sorry didn't see your comment until this morning, merge is done now.

src/package-hoister.js
```diff
@@ -63,10 +78,14 @@ export default class PackageHoister {
...
     this.levelQueue = [];
     this.tree = new Map();
+
+    this.nohoistResolver = new NohoistResolver(config, resolver);
   }
 
   resolver: PackageResolver;
   config: Config;
+  nohoistResolver: NohoistResolver;
+  // _rootNohoist: ?Array<string>;
```
arcanis: Commented code

arcanis:
Thanks a lot @connectdotz, it looks like a great addition to Yarn!

Could you detail a bit the design of the feature as a comment here, for future reference? For example, I wonder if the extra classes are required: I would have thought about nohoist has a special case of the regular hoister rather than an extension (schematically, I probably would have thought that we could have implemented it simply by making a function return false, so I'm curious about what required all the logic changes).

src/cli/commands/install.js
```diff
@@ -316,9 +316,11 @@ export class Install {
         }
       };
 
-      pushDeps('dependencies', projectManifestJson, {hint: null, optional: false}, true);
-      pushDeps('devDependencies', projectManifestJson, {hint: 'dev', optional: false}, !this.config.production);
-      pushDeps('optionalDependencies', projectManifestJson, {hint: 'optional', optional: true}, true);
+      if (cwdIsRoot) {
```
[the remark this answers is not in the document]
connectdotz: short answer: if not disabled, the workspace's dependencies will be hoisted to the root, by-passing nohoist check in `package-hoister`. 

longer answer: in the production `package_hoister` implementation, all the patterns passed in (such as from the pushDeps here) are assumed to reside on the root level. I need to either change that assumption in `package_hoister` or change the code here to only push the true root level dependency. I choose the later merely due to it has less code change. 

additionally, I also updated the test case "can add nohoist pacakge from workspace" in `__tests__/commands/add.js` to safeguard and illustrate this particular use case. You can test the behavior by comment out the `if(cwdIsRoot){` and the last 3 lines in the test should fail.

connectdotz:
@arcanis if I understand your question correctly, you are wondering why creating a separate class  `NohoistResolver` instead of just a simple method `canHoist(pkg)` in `package-hoister.js`? 

Conceptually, we need to answer 2 questions:
1. can this package be hoisted
1. if not, where should the package live

To determine if a package can be hoisted in a deep-nohoist environment turns out to be quite a challenge. For example, imagine a nohoist pattern with `workspace-a/**/react-native/**`, we can't just look at the package "shallowly" because some "in-between" packages might be hoisted while their dependencies might match the nohoist pattern in the future iterations. Before the nohoist,  `PackageHoister` didn't need to care about the package dependency "path" so to speak since everything was hoisted, so it only does shallow iteration and only care about the "destination" or the "to" and not the source or the "from". Therefore, we had to introduce a few new concept/artifacts such as the originalPath ("from") for the nohoist pattern match to work as expected (This turns out to be quite useful later for the "why" command, a bonus), and how to match deep path in a shallow iteration used by `PackageHoister` etc. In short the logic of `canHoist()` is not as straight-forward as we all hoped...

If the package can't be hoisted, where should it go? We need to resolve the "highest-hoisted-point" ...  fortunately, this part is not too complicated, but nevertheless, a distinct functionality from the `canHoist()` described above.

It is entirely true that while the logic is non-trivial, it doesn't necessarily warrant a sperate class... I can certainly add these methods to the existing `PackageHoister` class, but I thought the complexity and functionality difference is significant enough that it would be better to group them into a separate unit for easier bug fix and code changes going forward... 

Let me know if this didn't answer your question...

arcanis:
I see, thanks. I'll need to think a bit about this before merging, but it should hopefully be in the not-the-next-one-but-the-one-after release :)

Pranit-Harekar:
Please merge this PR ASAP. We are waiting for this PR to be merged. Thanks!

bestander:
Pinged @arcanis.
@connectdotz would you be so kind to rebase to speed up the merge process?

connectdotz:
shoot, maybe speak too soon... I did a merge but you mentioned rebase. If rebase is the policy, I will need to redo this...

bestander:
All is fine, thanks for this work over the 3 months, @connectdotz, huge change!

bestander:
Want to write a blog post about this feature https://yarnpkg.com/blog/?

MrLoh:
Awesom that this has landed. Thanks for all the work. What does that mean now, when will it become available?

bestander:
If you want to start using it now then feel free to use this bundle https://9293-49970642-gh.circle-artifacts.com/0/yarnpkg/yarn-1.4.1-20180128.1637.js

Otherwise we'll need to wait for 1.4.2, it should take a week or so for the next official release, depends on the core team workload

connectdotz:
@bestander it's great to see this finally in! 👍 

> Want to write a blog post about this feature https://yarnpkg.com/blog/?

ok, will do. Will also publish the react-native repository we used to test nohoist, so there is a working example/demo and people can play with it themselves...

k15a:
@bestander Hey, I don't want to be annoying but as it seems that many people including myself would love to use this feature officially I would like to know if there are any plans for a next release?

Thank you for the hard work on this feature and Yarn!

arcanis:
You can use right now by installing a [nightly build](https://yarnpkg.com/en/docs/nightly). That would actually help us making sure everything is ready for prime time! :)

And the official release shouldn't be far, I just haven't took the time to make a decent changelog yet. By the end of next week I'd say.

razagill:
@connectdotz awesome work!!
I'm testing this with the nightly in a yarn workspace managed by lernajs. I think we need to open an issue there as well to make use of this? Because currently I'm getting this error on running `lerna bootstrap`
```
TypeError: packageConfigs.some is not a function
```
I'm guessing this has to do with `workspaces` now being an object instead of an array.

k15a:
@razagill This is already merged to Lerna in https://github.com/lerna/lerna/pull/1254. Might not be released as well.

razagill:
@k15a nice!

darthtrevino:
So I'm running into some issue with the nightly build. If I have a configuration that looks like this, then I would expect that vr client's dependencies are not hoisted. 
```
"workspaces": {
    "packages": [
        "packages/schema",
        "packages/server",
        "packages/common",
        "packages/clients/*"
    ],
    "nohoist": ["packages/clients/vr/**"]
}
```

What I'm seeing is that the dependencies of the vrclient project remain hoisted. I've tried a few variants: 
* Changing the nohoist spec to "**/packages/clients/vr**"
* Adding `"nohoist": "**"` to the vr client (per rfc text, I think this should work)
* Adding `{ "workspaces": { "nohoist": "**" } }` to the vr client

What does work is nohoisting specific dependencies, (e.g **/react-native/**). This seems like an antipattern, because you may want hoisting throughout the rest of your codebase. For example, I may want to use redux in my react-native project so I nohoist it, but then it's going to be nohoisted for my web project as well, which may not be desirable. 

The common use cases here are either a react-native based project, where the bundling system expects your dependencies to exist in the directory of the package or the an aws lambda function where you're uploading a function script and a node_modules directory. 

I realize that there's already been an RFC and everything, but would it make sense to support a streamlined configuration that works at the package-level instead of the dependency level?

```
{
  "packages": [
    "packages/clients/*",
    "packages/lambdas/"*",
  ],
  "nohoist": [
     "packages/clients/native",
     "packages/clients/vr",
     "packages/lambdas/*"
  ] 
}
```

Additionally: There's an issue when trying to set nohoist a package as a dependency in my workspace. For example, my vr project depends on a common project. If I add the common project to the nohoist spec then it remains hoisted. I'm not sure if this is an issue with local dependencies or scoped package-names (everything in my project is scoped).

connectdotz:
Glad people are trying nohoist. My bad that I should have published the examples and a more comprehensive introduction earlier (was waiting for 1.4.2 official release...), I think they would have spared some of your troubles... Here they are:

- [yarn-nohoist-examples](https://github.com/connectdotz/yarn-nohoist-examples), you should be able to run these with the nightly build or compare with your own projects. Many of your use cases should be covered there. 
- [nohoist introduction](https://github.com/connectdotz/yarn-nohoist-examples/blob/master/nohoist.md), it needs a little bit of polish before submitting to yarn blog, but you can preview it here for now.

relferreira:
I've been using the bundle` yarn-1.4.1-20180128.1637.js` for a while, and it's working pretty well. Thanks @connectdotz

dannycochran:
@arcanis @connectdotz above it was mentioned this might get into 1.4.2, is that still the case? Is there a known timeline for 1.4.2?

arcanis:
Yep, I planned to review a few PR and release the next version tomorrow! In
the meantime, feel free to use the nightly, it would help us making sure
everything is a-ok!

k15a:
@arcanis I've tested our projects with the nightly build and everything works great. The only thing is that it takes a lot of time to install if you don't hoist anything. Maybe there is something Yarn could cache in the future or even have an option which hoists the dependencies but links them back.

Thanks for this great feature.

arcanis:
1.5.0 is here! Thanks again for your hard work, @connectdotz 🎉

natew:
Want to voice thanks for implementing this as well as note my initial install time was a bit high as well. For a medium(ish?) monorepo:

image

connectdotz:
hi,

glad this helps. A few things you can help to isolate/improve the perf issue:
1. compare the performance without nohoist. it will help us isolate if the slowness is specifically caused by nohoist. 
1. try yarn install with `--link-duplicates` and compare the perf. The flag instructs yarn to hard-link, instead of copying, duplicate files. Not only it reduces the footprint of your repo,  I wonder if it has some perf advantage as well, if the slowness is caused by file-system related ops (?)... However, there was a bug that just got fixed(#5442), you will need to use the nightly build to avoid that issue.

Issue #2840, opened 2017-03-06
Yarn does not remove old package name from yarn.lock when install from git repo


kokororin:
yarn version 0.21.3

**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When install new package which is the same name as the old one from git repo, the old package in yarn.lock does not disappear.

```bash
npm init -y
yarn add https://github.com/FF-Mercurial/ug
yarn add https://github.com/kokororin/ug
```

see yarn.lock:
```yaml
"ug@https://github.com/FF-Mercurial/ug":
  version "1.0.0"
  resolved "https://github.com/FF-Mercurial/ug#5a0400908d4a2b3ad8ce6d9a4c42fd3d3640983c"
  dependencies:
    escodegen "^1.6.1"
    esprima "^2.5.0"
    uglify-js "^2.4.24"

"ug@https://github.com/kokororin/ug":
  version "1.0.0"
  resolved "https://github.com/kokororin/ug#d38db29b6ce10037d235875c3b5b56b97a36877e"
  dependencies:
    escodegen "^1.6.1"
    esprima "^2.5.0"
    uglify-js "^2.4.24"
```

**What is the expected behavior?**
The old package in yarn.lock should be removed.

**Please mention your node.js, yarn and operating system version.**
```bash
$ ver
Microsoft Windows [版本 6.1.7601]

$ node -v && yarn --version
v7.5.0
0.21.3
```

sentience:
Seeing the same issue when adding from a git repo a package that was previously installed from the NPM package repository: yarn.lock contains two entries for the same dependency.

BYK:
Just confirmed that this is still valid for latest master as of today (1.2.1+)

james-rabbit:
I see that a fix was submitted for this, but I still have a similar issue with `link` dependencies.
What is the current status of this bug?

Lonli-Lokli:
Any update? It's really annoying to have multiple git deps in repo!

kittimiyo:
Just saw this happen with yarn version 1.14.0

PR #4989, opened 2017-11-24
Fix for issue #2840

davidreis97:
**Summary**

This is a fix for issue #2840 .

It fixes the issue by comparing the name of the git repo with the entries on the lockfile, instead of comparing the whole git url (which wouldn't match with the version already on the lockfile, since the url would be different).

**Test plan**

We tested with multiple git repositories, including the ones that were given as an example on the issue page.


kaylieEB:
@davidreis97 Thanks for the PR! can you add a test case for this?

davidreis97:
Hi @kaylieEB , I just added a test case in the last commit.

src/cli/commands/install.js
```diff
@@ -234,14 +237,20 @@ export class Install {
...
-      // can't extract a package name from this
-      if (getExoticResolver(pattern)) {
+      if (GitResolver.isVersion(pattern)) {
+        const pathname = urlParse(pattern).pathname;
+
+        if (pathname) {
+          const name = util.removeSuffix(pathname.split('/')[2], '.git');
+          excludeNames.push(name);
+        }
+      } else if (getExoticResolver(pattern)) {
```
kaylieEB: Can you instead try using [guessName util](https://github.com/yarnpkg/yarn/blob/8918da15c36e793c8f5ed7be24197a7b5afc89d4/src/util/guess-name.js), so something like
```
if (getExoticResolver(pattern)) {
   excludeNames.push(guessName(pattern))
}
```
This way we avoid git-specific logic in the install command file and other exotic resolvers and hosted git resolvers can't take advantage of this fix too :)

__tests__/commands/add.js
```diff
@@ -609,6 +609,27 @@ test.concurrent('upgrade scenario 2 (with sub dependencies)', (): Promise<void>
   });
 });
 
+test.concurrent('install another fork of an existing package', (): Promise<void> => {
+  // When installing a package with the same name as an existing one but from a different repo,
+  // the old one should be replaced with the new one in the lock file.
+  const firstSource = 'https://github.com/davidreis97/example-yarn-package.git';
+  const secondSource = 'https://github.com/yarnpkg/example-yarn-package.git';
```
kaylieEB: To avoid making requests in the ci, can you add the tarball cache to `__tests__/fixtures/request-cache/GET/github.com`

PR #4994, opened 2017-11-25
fix(install): use node-gyp from homebrew npm

chrmoritz:
**Summary**

With this `yarn` will be able to discover and use the `node-gyp` from the homebrew installed `npm` on macOS instead of falling back to globally installing `node-gyp` every time a native addon needs to be compiled from source.

Homebrew installs a clean copy of `npm` inside a `libexec` folder together with `node`. Previously yarn didn't look there when trying to locate `node-gyp` and the `globally install node-gyp`-fallback would be used every time when building native addons with a yarn version from homebrew.
This PR adds the `libexec` path of `node-gyp` from homebrew to the `node-gyp` search paths of yarn, making it possible to compile native addons using the homebrew npm provided `node-gyp` with yarn without relying on this fallback.

**Test plan**

This can't be tested outside a homebrew environment (with node installed by homebrew, not by a node version manager like nvm).

To manually test this on a mac with homebrew installed you can install a test build of this PR (uploaded to github releases on my fork) by installing [this gist `yarn` formula with](https://gist.github.com/chrmoritz/a572e5efca056d4dcff43ba376740bed):

```
brew reinstall https://gist.githubusercontent.com/chrmoritz/a572e5efca056d4dcff43ba376740bed/raw/21ef12adb4b9efc33c298ed29744dfc89299f936/yarn.rb
```

After this you can delete your global `node-gyp` and install any native addon (with `yarn config set build-from-source` set to skip prebuild / node-pre-gyp) and verify that no new global `node-gyp` is installed. You could also look at the verbose install output to verify, that the fallback isn't used this time. You can also verify with the current homebrew version of yarn, that the global node-gyp fallback is used every time when building native addons from source until now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,7 +117,6 @@ jobs:
           command: |
             brew uninstall --ignore-dependencies node
             HOMEBREW_NO_AUTO_UPDATE=1 brew install node@8
-            yarn global add node-gyp
       - *attach_workspace
       - *test_build
       - *test_run
@@ -130,7 +129,6 @@ jobs:
             brew uninstall --ignore-dependencies node
             HOMEBREW_NO_AUTO_UPDATE=1 brew install node@6
             brew link --force node@6
-            yarn global add node-gyp
       - *attach_workspace
       - *test_build
       - *test_run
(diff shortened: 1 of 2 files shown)


chrmoritz:
Could someone please review this PR? (friendly ping @BYK)

What else do I need to get this PR merged?

arcanis:
It makes sense and I guess the line you removed actually test that the feature works, so I'm gonna go ahead and merge it. Thanks for your contribution!

PR #4999, opened 2017-11-27
fix(add): Make semver flags compatible with versioned requests

### LINKED ISSUE #4978: yarn add --tilde does not install newer patch versions
Bug 

The --tilde flag does not install the latest patch version matching the version I specify. It does not put a tilde in my package.json, either.

For instance, take the package moment. Currently, moment's versions include 2.18.0 and 2.18.1. As such, if I run `yarn add moment@2.18.0 --tilde`, I expect to install version 2.18.1, and to add this line to my package.json file: `"moment": "~2.18.0"`.

Instead, yarn installs version 2.18.0, and adds this line to my package.json file: `"moment": "2.18.0".

See above.

I expect yarn to install the latest patch version of the package I include.

yarn 1.3.2, node 6.9.4, Mac OS X High Sierra

kaylieEB:
**Summary**

Fixes #4978

Previously, there was logic in add command that [uses user-specified range out of the box](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/add.js#L71-L72), resulting in the issue above. For example, if the user ran `yarn add left-pad@1.0.0` the pattern parser skipped `--exact` or `--tilde flags`, as well as saved prefix config options.

This change updates it so that if the user specifies a valid numeric version, then these flags still have an effect. These cases don't cover `yarn add left-pad@~1.0.0` or `yarn add left-pad@^1.0.0` since I think the confusion only arises from when the user enters `exact` pattern (just the number).

**Test plan**

Added a test case that fails on master

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -147,6 +147,16 @@ test.concurrent('install with --optional flag', (): Promise<void> => {
   });
 });
 
+test.concurrent('install with --tilde flag', (): Promise<void> => {
+  return runAdd(['isarray@2.0.1'], {tilde: true}, 'add-with-flag', async config => {
+    const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+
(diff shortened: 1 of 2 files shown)


BYK:
+1 to @rally25rs  - thanks for the review too :)

Issue #4983, opened 2017-11-22
On macosx "Do you want the application "node" to accept incoming network connections?" popup causing `yarn install --mutex network` to hang


greyepoxy:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
On macosx running `yarn install --mutex network` results in a popup "Do you want the application "node" to accept incoming network connections?" which hangs the process.
![image](https://user-images.githubusercontent.com/11470527/33145440-3acb82ee-cf75-11e7-915a-694d746a6e8c.png)

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Ensure firewall is on System Preferences -> Security or Security & Privacy -> Firewall
1. Ensure Node is not on the firewall exception list, check in Firewall Options
1. `yarn init` a new project
1. `yarn install --mutex network`

**What is the expected behavior?**
No popup to occur. It doesn't matter what the user clicks (Deny vs Allow) yarn starts a webserver to serve as a mutex and does not actually need any incoming network connections. Once either button is clicked yarn continues and finished successfully.

**Please mention your node.js, yarn and operating system version.**
`yarn --version` -> 1.3.2
`node --version` -> 8.9.1
mac osx -> 10.12.6

greyepoxy:
I believe the fix is to update https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L288
to pass in host: 'localhost' in addition to the port

kaylieEB:
@greyepoxy thanks for filing the issue! Would you like the submit a PR?
Alternatively, you can also set your firewall settings to automatically accept connections like below and it won't trigger the prompt

<img width="668" alt="screen shot 2017-11-27 at 8 27 59 am" src="https://user-images.githubusercontent.com/18429494/33277798-75f7e628-d34e-11e7-87d1-ffed8611c4ce.png">

greyepoxy:
Thanks @kaylieEB! yeah I can submit a PR

Unfortunately we are seeing the issue on our CI machines in some cases and we don't have sudo access so cannot automate the setting of the exception.

PR #5006, opened 2017-11-28
fix: connectionOptions passes in localhost as its host to prevent pop up on MacOsx

greyepoxy:
**Summary**
This is to fix https://github.com/yarnpkg/yarn/issues/4983 to prevent the firewall popup on Mac OSX.

**Test plan**
Ran the following tasks
```bash
yarn run test
yarn run lint
```

Also validated that the popup no longer occurs by setting the alias
`alias yarn="node /path/to/my/yarn/lib/cli/index.js"`
and running `yarn install --mutex network`

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -286,6 +286,7 @@ export function main({
     return new Promise((resolve, reject) => {
       const connectionOptions = {
         port: +mutexPort || constants.SINGLE_INSTANCE_PORT,
+        host: 'localhost',
       };
 
       function startServer() {


src/cli/index.js
```diff
@@ -286,6 +286,7 @@ export function main({
     return new Promise((resolve, reject) => {
       const connectionOptions = {
         port: +mutexPort || constants.SINGLE_INSTANCE_PORT,
+        host: 'localhost',
```
kaylieEB: It's interesting that it didn't work before given that https://nodejs.org/api/http.html#http_http_request_options_callback (and by extension http.get) defaults to `localhost`?

greyepoxy: Yeah I think for the http.get call below this does not matter but for the [server.listen](https://nodejs.org/api/http.html#http_server_listen) call it does.

Issue #4769, opened 2017-10-24
Add license text to yarn licenses list


crobi:
**Do you want to request a *feature* or report a *bug*?**

A feature

**What is the current behavior?**

It is difficult to write a script that consumes the output of `yarn licenses list` and creates a customized disclaimer (same content as `yarn licenses generate-disclaimer`, but with a custom format)

- `yarn licenses list --json --no-progress` is not documented.
- `yarn licenses list` does not include the license text or a link to the license file.

**What is the expected behavior?**

Improve `yarn licenses list --json --no-progress` to include the license text, using the same mechanism to get the text as `yarn licenses generate-disclaimer` uses.

**Please mention your node.js, yarn and operating system version.**

```
$ yarn --version
1.2.1
$ node --version
v8.7.0
```

MartaTorgal:
Me and a group of two more are going to address this issue over the next week 😄

PR #5011, opened 2017-11-28
Issue 4769- license disclaimer in JSON output 

msramalho:
### Summary

With this pull request, we present a solution for issue [#4769](https://github.com/yarnpkg/yarn/issues/4769). 

**Motivation:** Besides being a good beginner issue to start understanding yarn, this PR is useful for users trying to consume the json output of `yarn licenses list --json --no-progress` or `yarn licenses list --json` by including the license disclaimer text in the JSON output. 
**Solves:** Issue [#4769](https://github.com/yarnpkg/yarn/issues/4769). According to which users should also receive the license text in the licenses listing in JSON format. 
### Test plan
The changes were simple enough and it was tested on yarn itself, returning the license text where present and Undefined otherwise.

### DIFF
--- a/src/cli/commands/licenses.js
+++ b/src/cli/commands/licenses.js
@@ -49,11 +49,9 @@ async function getManifests(config: Config, flags: Object): Promise<Array<Manife
 
 async function list(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const manifests: Array<Manifest> = await getManifests(config, flags);
-//TODO: invoke a function to get the licenseText, see function generateDisclaimer below and extract a new function from there
   if (flags.json) {
     const body = [];
-
-    for (const {name, version, license, repository, homepage, author} of manifests) {
+    for (const {name, version, license, licenseText, repository, homepage, author} of manifests) {
       const url = repository ? repository.url : homepage;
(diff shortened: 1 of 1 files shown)


src/cli/commands/licenses.js
```diff
@@ -49,25 +49,23 @@ async function getManifests(config: Config, flags: Object): Promise<Array<Manife
...
         license || 'Unknown',
+        licenseText || 'Unknown',
         url || 'Unknown',
         vendorUrl || 'Unknown',
         vendorName || 'Unknown',
       ]);
     }
-
-    reporter.table(['Name', 'Version', 'License', 'URL', 'VendorUrl', 'VendorName'], body);
+    reporter.table(['Name', 'Version', 'License', 'Disclaimer', 'URL', 'VendorUrl', 'VendorName'], body);
```
Gudahtt: `License Text` would be more appropriate here than `Disclaimer`. Disclaimers are typically one component of the license text.

Gudahtt:
The branch where this change was made seems to be untested at the moment.

There are a couple of tests for the `licences` command in `__tests__/commands/licenses.js`, but none that use the `--json` flag. It would be pretty straightforward to add another test using `--json` I think, following the same pattern as the `list` test (using a Jest snapshot).

PR #5012, opened 2017-11-28
Fix transient bin links overriding direct ones

### LINKED ISSUE #4937: Sub dependency .bin symlinks can override top level dependency .bin symlink

Bug

Sub dependencies .bin symlinks override top level dependencies .bin symlinks in some cases (seems to be depending on which top level package is first when sorted alphabetically). Likely introduced in https://github.com/yarnpkg/yarn/pull/3310

Example: https://github.com/anders-advisa/yarn-symlink-bug

Dependencies explicitly entered into `package.json` should have precedence over sub dependencies.

node: 8.9.1
yarn: 1.3.2

AndersAstrand:
**Summary**

This fixes #4937

The problem still exists when doing `yarn add` in an already installed module however, but that is unaffected by this patch as far as I can tell, and all tests pass.

This is my very first foray into the yarn code base, so please feel free to comment on anything in here. However petty it might seem to you!

**Test plan**

Run the test suite.

### DIFF
--- a/__tests__/commands/install/bin-links.js
+++ b/__tests__/commands/install/bin-links.js
@@ -145,6 +145,17 @@ test('first dep is installed when same level and reference count and one is a de
   });
 });
 
+// Scenario: Transitive dependency having bin link with a name that's conflicting with that of a direct dependency.
+// Behavior: a-dep and b-dep is linked in node_modules/.bin rather than c-dep and d-dep
+test('direct dependency is linked when bin name conflicts with transitive dependency', (): Promise<void> => {
+  return runInstall({binLinks: true}, 'install-bin-links-conflicting-names', async config => {
+    const stdout1 = await execCommand(config.cwd, ['node_modules', '.bin', 'binlink1'], []);
+    const stdout2 = await execCommand(config.cwd, ['node_modules', '.bin', 'binlink2'], []);
+    expect(stdout1[0]).toEqual('direct a-dep');
+    expect(stdout2[0]).toEqual('direct b-dep');
+  });
+});
+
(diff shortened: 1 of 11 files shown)


rally25rs:
Unfortunately I can't push a commit to this, but I pushed a commit to my fork if Yarn that would fix this issue.

@AndersAstrand have a look at https://github.com/yarnpkg/yarn/commit/c471ccf605d3175425dee6bb3a18a6e29c9d6809 and see what you think. We could either incorporate those changes to this PR, or I could PR my branch and close this one (I'd hate to hijack your PR). Either way, that you for writing the test case for this, it really helps! 👍

AndersAstrand:
Thanks a lot for your comments!

Transitive, got it, I read transient somewhere I think and just went with it :).

You are, of course, correct about the `z-dep` thing, and here I thought I was so clever when I made sure they were cross referencing the same bin names.

I suspected my proposed fix was way too naive to actually work, but a green test run have a tendency to instill confidence :).

Feel free to just close this pull request or hi-jack it completely. I'm only happy if i could be of help in getting this bug fixed since it's what's stopping us from switching to yarn.

rally25rs:
> Transitive, got it, I read transient somewhere I think and just went with it :).

Trust me, I do the same thing myself all the time 😆 

> Feel free to just close this pull request or hi-jack it completely.

Closing this PR and opening #5016 to replace it (it's the same as this PR with 1 additional commit)

> I'm only happy if I could be of any help in getting this bug fixed since it's what's stopping us from switching to yarn.

And we thank you for your contribution! 🎉  The more community members we can get involved, the better this project will be 👍

PR #5016, opened 2017-11-30
Fix transient symlinks overriding direct ones v2

### LINKED ISSUE #4937: Sub dependency .bin symlinks can override top level dependency .bin symlink

rally25rs:
**Summary**

This PR is a replacement for #5012 with a different fix.

fixes #4937

Added a sort to the top level bin links to ensure that direct dependencies are linked last. Since bin links are overwritten if they already exist, this will ensure that direct deps bin links will overwrite transitive ones.

**Test plan**

Run the test suite.


src/package-linker.js
```diff
@@ -436,17 +436,34 @@ export default class PackageLinker {
...
+      const bIsDirect = bHoistManifest.isDirectRequire;
+
+      if (aIsDirect && !bIsDirect) {
+        return 1;
+      }
+      if (!aIsDirect && bIsDirect) {
+        return -1;
+      }
+      return 0;
+    });
```
arcanis: What do you think about splitting the array rather than doing a full sort? The algorithmic complexity would be better:

```
let transientBins = [], topLevelBins = [];
for (let ... of Array.from(linksToCreate.values())) {
    if (isDirectRequire) {
        topLevelBins.push();
    } else {
        transientBins.push();
    }
}

return [... transientBins, ... topLevelBins];
```

rally25rs: I like that. The code is easier to read.

AndersAstrand:
This bug is still present in master (fb6511c9) and is blocking us from switching to yarn. Is there anything I can do to help move this PR forward?

rally25rs:
@arcanis think we can merge this?

arcanis:
Totally! This will be in the next release, hopefully today or tomorrow 👍

PR #5018, opened 2017-11-30
test(jest): Upgrade jest to latest available version

BYK:
**Summary**

Upgrades Jest to latest available version. It also fixes the following issues with tests:
 - Fix a few issues regarding cache tests
 - Speed up a fixture used in cache tests
 - Fix mocking of `process.platform` in tests
 - Fix flow types with the update of Jest flow definitions

**Test plan**

All tests should pass on all platforms and CI.


rally25rs:
Should there be an updated Flow typing file for this new version? I would expect https://github.com/yarnpkg/yarn/blob/master/flow-typed/npm/jest_v20.x.x_with_custom_matchers.js to be replaced with a new v21 file.

BYK:
Test failures should be fixed with #5076.

Issue #1123, opened 2016-10-16
extra color for bolded text in the output


e-cloud:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
the installed/changed dependencies in the output are bolded without color escaped
![image](https://cloud.githubusercontent.com/assets/5255354/19416661/256d107c-93c9-11e6-8045-92369bfde960.png)

**What is the expected behavior?**
give some color to the bolded text, perhaps cyan :).

**Please mention your node.js, yarn and operating system version.**
node v6.8.0
yarn v0.15.1
ubuntu 16.04

samccone:
@e-cloud thanks for the issue. this would make for a great first issue.

your contributions are welcome here!

akshbn:
I would like to take this up, if no one else is working on it.

olingern:
@akshbn @samccone I'm knee deep in the tree reporter, would it make sense for me to take this up? Should be super simple and could be a small addition to #1463

akshbn:
@olingern  i managed to change the format from bold to color,ie, the content gets colored but its no more bold. still haven't dived deep enough to retain the format to bold n assign it a color. If you are able to do this, please go ahead. If not, do tell and i shall send a PR.

olingern:
@akshbn Not sure when #726 will merge, so [here's an example](https://github.com/yarnpkg/yarn/blob/master/src/reporters/console/console-reporter.js#L209) of coloring output.

akshbn:
@olingern thanks man! will check it out.

olingern:
I wanted to pick back up on this since #1463 was merged, and I wanted to understand the reasoning behind bolding before changing the color, since it's closely related to tree output that I worked with.

In the [latest stable release](https://github.com/yarnpkg/yarn/blob/0.16-stable/src/cli/commands/ls.js#L75), there's a condition that evaluates to remove bold, but I'm unsure of what triggers this scenario (other than the caret pre-pended to the version in the unbolded item). 

<img width="245" alt="screen shot 2016-11-07 at 5 33 56 pm" src="https://cloud.githubusercontent.com/assets/1470297/20078904/68035c68-a510-11e6-9d9f-6491b23e90d3.png">

pinging @kittens as it looks like this landed in [this commit](https://github.com/yarnpkg/yarn/commit/b74f878dc3e76a22723f39421376f61de553e990).

bestander:
I think it does not display second level dependencies anymore, looks like got broken

[...]

PR #5023, opened 2017-12-01
Fixed issue #1123 extra color for bolded text in the output

Koreris:
**Summary**

This pull request presents a solution for issue [#1123](https://github.com/yarnpkg/yarn/issues/1123), adding the suggested color for the bolded input.

**Motivation**

This issue was picked because as we didn't have much experience, this seemed like a good first issue so we could have some entry way point. 
This isn't a substantial bug fix or feature request, it merely adds the convenience of distinguishing better which dependencies were updated, by bolding them and giving them a different color in the tree that shows up after `yarn add` is used.

**Test plan**

We tested the developed code by installing different versions of "lodash" by doing `yarn add lodash@1.2.0` for an exemple, but we kept changing the versions so we could see the tree with the altered dependencies. We also tested with another package, standard-version `yarn add standard-version@4.2.0`, following a similar procedure. We were only able to test on Windows so far.

The result can be seen in the following images

capturar

image   

**EDIT:** By distraction I had created a pull request in my own fork of the repo, however as I noticed it I fixed it by creating this one.


Koreris:
It seems to me that some of the tests are failing because the tests are prepared for a different color (null), and so the snapshot doesn't match the received, as seen here, even though the trees are virtually the same except for the intended new feature of having colored and bolded dependencies.

capturar
capturar2

ClaudiaRaquelGuedes:
ANSWER PLEASE!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

bestander:
@Koreris, nice first PR! 
Thanks for sending it in

bestander:
@Koreris, can you update the snapshots `./node_modules/.bin/jest -u`?

bestander:
This seems fine to me, I don't have a strong opinion on the bold/not bold usage in the formatter.
@olingern, you have been looking into this feature quite a lot, what do you think?

Koreris:
@bestander I have been looking for that file but maybe I'm looking in the wrong folder? Because I don't seem to find it
image

bestander:
Maybe it is OS specific.

Try
`node node_modules/jest/bin/jest.js -u`

ClaudiaRaquelGuedes:
Me and @Koreris found the file but don't know exactly what to do with it.

image

bestander:
You need to run command `node node_modules/jest/bin/jest.js -u` from CMD when you cd to yarn folder.

rally25rs:
Hi @Koreris thanks for the contribution!

I'd like to ask you to move the color change to a different spot in the code. Having it in `console-reporter` causes all output using the `tree()` function to be in cyan. If you do a global search for `.tree(` you can find it used a couple other places. For example the commands `yarn list` and `yarn licenses list` are now all cyan.

A bit of a hint is that in this code block:

```javascript
    const output = ({name, children, hint, color}, titlePrefix, childrenPrefix) => {
      const formatter = this.format.bold;
      const out = getFormattedOutput({
        prefix: titlePrefix,
        hint,
        color: 'cyan',
        name,
        formatter,
      });
      this.stdout.write(out);

      if (children && children.length) {
        recurseTree(sortTrees(children), childrenPrefix, output);
      }
    };
```

the `color` variable is no longer used. I'm actually really surprised that Flow or ESLint doesn't throw an error on that. 😕

Anyway, if you backtrack through the code a bit, you can find that the `add` command builds its info to send to the `ConsoleReporter.tree()` function over in `src/cli/commands/list.js` `buildTree()`

```
    let color = 'bold';

    ...

    if (info.originalKey !== info.key || opts.reqDepth === 0) {
      // was hoisted
      color = null;
    }

    ...

    if (topLevel || nextDepthIsValid || showAll) {
      treesByKey[info.key] = {
        name: `${info.pkg.name}@${info.pkg.version}`,
        children,
        hint,
        color,
        depth,
      };
    }

    ...

        if (!hoistedByKey[`${info.key}#${pkg.name}`] && (nextChildDepthIsValid || showAll)) {
          children.push({
            name: pattern,
            color: 'dim',
            shadow: true,
          });
        }

```

So `bold`, `dim` and "normal" (`null`) are all used here. The original issue looks like they wanted some color added to "bold" so I think you could just change `let color = 'bold';` to `let color = 'cyan';`

-----

There seems to be a bigger issue in what the output shows when doing a `yarn add`; it looks like it is only showing the top level and hoisted dependencies, which is a bit weird. This check:

```
    if (info.originalKey !== info.key || opts.reqDepth === 0) {
      // was hoisted
      color = null;
    }
```

is causing all the output for `yarn add` to be normal text because `add` passes `reqDepth=0` to this function.

`yarn list` does not, and so shows all 3 colors (bold, normal, dim). But I think that may be a bigger issue that you can ignore for this PR.

-----

For the unit tests, look in `__tests__/commands/__snapshots__/list.js.snap` and you can see what objects the test is expecting. It has the color in it:

```
        Object {
          "children": Array [],
          "color": "bold",   <-- here
          "depth": 0,
          "hint": null,
          "name": "a@1.0.2",
        },
```

so when it was changed to `'cyan'` these comparisons no longer match. You would just need to change the color in the snapshot file to match the changed color.

You can run `yarn run test` to run the tests locally so you can see if they pass before pushing commits.

Hope this helps! Thanks again!

Koreris:
@rally25rs Thanks for the feedback, it's much appreciated! I'll work on it as soon as I can and fix those problems.

rally25rs:
It also looks like somewhere along the way we got a `package-lock.json` file committed, which should not exist. That would probably be from an `npm install` but we should be using `yarn` not `npm` ;)

olingern:
Late to the party, sorry about that.

> This seems fine to me, I don't have a strong opinion on the bold/not bold usage in the formatter.
@olingern, you have been looking into this feature quite a lot, what do you think?

Ergonomically, I think if colors were a consistent shade of white / grey and dimmed as the progressed down the tree -- that would be nice. Off the top of my head, I'm not sure how many shades we have access to through standard terminal colors / chalk.

The only downside I see to including a color outside of the current white / grey scheme, is the overhead / noise that may happen when looking through the list.

Otherwise, it's definitely an improvement :)

olingern:
+1 to @rally25rs comment. We'll need to remove the `package-lock.json` and rebase master over this branch. Should make review easier

arcanis:
I'm going to close this PR for now to have a better overview of the ones currently active, but feel free to reopen it once you've been able to work on @rally25rs's comment! And thanks for your involvement into making Yarn even better 😄

Issue #5017, opened 2017-11-30
add local repository with branch


alpha-and-omega:
Requesting a feature.

Please make it possible to add local repository providing a branch:
yarn add file:../repo/sitory#branch
fails
yarn add file:../repo/sitory@version
fails

rally25rs:
#5026 pushes this a step closer. It will allow a local git repo of the form `file:../my-project.git` (the tailing `.git` is the key to having Yarn process it as a Git repo instead of a plain filesystem structure).

From there is should be a simple change to have it check for local un-pushed branches.

alpha-and-omega:
Why file-git-fix is not a part of a master since 15 december 2017 ?

arcanis:
Because I just merged it 🙂 Will be part of the next release.

PR #5026, opened 2017-12-01
feat(resolution) Local .git repositories are now locally cloned

### LINKED ISSUE #3670: Local packages named by *.git extension are handled as remote sources instead local ones.
This is a bug report.

In package.json:
`"myBackendModule": "file:../myBackendModule.git"`
```
C:\Git\myFrontendModule.git>yarn
yarn install v0.24.5
[1/4] Resolving packages...
error Command failed.
Exit code: 128
Command: git
Arguments: clone file:../myBackendModule.git C:\Users\christoph\AppData\Local\Yarn\cache\v1\.tmp\534d3fc1a8a74d07d486ff305cbd21f0
Directory: C:\Git\myFrontendModule.git
Output:
Cloning into 'C:\Users\christoph\AppData\Local\Yarn\cache\v1\.tmp\534d3fc1a8a74d07d486ff305cbd21f0'...
ssh: Could not resolve hostname file: Name or service not known

and the repository exists.
```

Renaming the folder to myBackendModule.got, it works. But on next checkout, it fails again, because the repos are named *.git.
Link a local package which name has git as extension.
In our enterprise, the git extension is default to show that the package comes from our git repository.

If the path is declared by "file" or if it's a relative path without server name, dont use the remote method but the local file access.

>node --version
v4.6.1
> npm upgrade --global yarn
> yarn --version
yarn v0.24.5

rafaeldamasceno:
**Summary**
This fixes #3670.
Verifies that the url parser returns a `file:` url, and stops using the added `file://` prefix when invoking git commands. Enables local git cloning even with `.git` extensions.

**Test plan**

All tests should pass.


src/util/git.js
```diff
@@ -114,6 +115,17 @@ export default class Git implements GitRefResolvingInterface {
...
+      return {
+        hostname: parsed.hostname,
+        protocol: parsed.protocol,
+        repository: parsed.path,
+      };
+    }
+
     return {
       hostname: parsed.hostname || null,
       protocol: parsed.protocol || 'file:',
```
rally25rs: This can now use `FILE_PROTOCOL`

src/util/git.js
```diff
@@ -114,6 +115,17 @@ export default class Git implements GitRefResolvingInterface {
       };
     }
 
+    // npm local packages are specified as file:, but url parser interprets them as using the file protocol.
+    // This changes the behavior so that git doesn't see this as a hostname, but as a file path.
+    // See #3670.
+    if (parsed.protocol === FILE_PROTOCOL && !parsed.hostname && parsed.path && parsed.port === null) {
+      return {
```
rally25rs: I think the only real difference between this returned object and the one below it is the `repository`.

Could this be simplified to something like 

```
function isLocalDir(parsed) {
  return parsed.protocol === FILE_PROTOCOL && !parsed.hostname && parsed.path && parsed.port === null;
}
```
```
const repository = isLocalDir(parsed) ? parsed.path : url.format({...parsed, hash: ''});
return {
  hostname: parsed.hostname || null,
  protocol: parsed.protocol || FILE_PROTOCOL,
  repository,
};
```

src/util/git.js
rally25rs: Do you think there is a way we can write a test for this? It doesn't look like there were any unit tests for local git directories using `file:./somewhere/.git` in fact, I didn't even realize that would work until I looked at this PR. It would be really nice to have some tests around that so we don't accidentally break it in the future.

rally25rs:
Do you think there is a way we can write a test for this? It doesn't look like there were any unit tests for local git directories using `file:./somewhere/.git` in fact, I didn't even realize that would work until I looked at this PR. It would be really nice to have some tests around that so we don't accidentally break it in the future.

rally25rs:
I have a hunch that wherever `git ls-remote` is being run, we need to use `git show-ref` instead for local repositories. I think that would possibly implement #5017 as well, since that is how Yarn finds the SHA hash that lines up with a branch, but `ls-remote` will only list ones that have actually been pushed.

src/util/git.js
rafaeldamasceno: Where would I go about declaring the isLocalDir function?

rafaeldamasceno:
I've added a single test and I'm waiting for your feedback on the code review. Is there any other kind of test I could add?

About fixing #5017: we are very limited on our knowledge of JS, git and the yarn codebase, so I don't think we'd be able to it right now, as we are behind schedule on some other projects and can't really afford to spend more time trying to understand it.

src/util/git.js
rafaeldamasceno: I'm hitting a lot of errors in flowcheck with incompatible types trying to implement it like you suggested. Not quite sure of what I can do, as most of them point to `parsed` attributes as being undefined and incompatible with the expected types. This doesn't allow me to do a simple return in `isLocalDir()` or to assign the correct path to `repository`.

rafaeldamasceno: No matter how I do it, I can't use the ternary operator because Flow doesn't really like it.

```
125:     const repository: string = isLocalDir(parsed) ? parsed.path : url.format({...parsed, hash: ''});
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ undefined. This type is incompatible with
125:     const repository: string = isLocalDir(parsed) ? parsed.path : url.format({...parsed, hash: ''});
                           ^^^^^^ string
```

Of course, it doesn't complain if repository isn't defined as `string`, but then I can't use it inside the return value because the `GitUrl` type specifies `repository` as `string`. I'm not sure how you'd want to do it because without the ternary operator, it's not as simple as a single `const repository` and then we're back to separate checks.

rally25rs: Oh yeah, Flow likes to yell at me too 😆 

I'm guessing that is because `parsed.path` could be `undefined`. I was just trying to simplify the code a bit, but if Flow is going to fight you, then don't worry about it.

rally25rs:
If I try this in a project using the dependency:

```
  "dependencies": {
    "yarn-scripts-test": "file:../yarn-scripts-test.git"
  },
```

Then I get this error:

```
yarn install v1.3.2
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads ../yarn-scripts-test.git
Directory: /Users/jvalore/Projects/yarn-test
Output:
fatal: '../yarn-scripts-test.git' does not appear to be a git repository

and the repository exists.
```

Is this still a known issue? Or is this working for you?

rafaeldamasceno:
I have seen that behavior but it was before I started working on the issue. Right now, as of the latest commit in this branch, I get no such thing. 

Here's the sample I'm using: [testsuite.zip](https://github.com/yarnpkg/yarn/files/1543784/testsuite.zip)

I run `yarn run build` on my local yarn source, then this is how I test it:

```
Rafael@FIXO-PC MINGW64 ~/Desktop
$ ls test test_module.git/
test:
package.json

test_module.git/:
package.json  somecontent

Rafael@FIXO-PC MINGW64 ~/Desktop
$ cd test

Rafael@FIXO-PC MINGW64 ~/Desktop/test
$ node ../yarn/lib/cli/index.js
yarn install v1.3.2
info No lockfile found.
warning test@0.0.1: No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Saved lockfile.
Done in 0.55s.

Rafael@FIXO-PC MINGW64 ~/Desktop/test
$ find .
.
./node_modules
./node_modules/.yarn-integrity
./node_modules/testmodule
./node_modules/testmodule/package.json
./node_modules/testmodule/somecontent
./package.json
./yarn.lock
```

If I use master, this is what I get:
```
Rafael@FIXO-PC MINGW64 ~/Desktop/test
$ yarn
yarn install v1.3.2
info No lockfile found.
warning test@0.0.1: No license field
[1/4] Resolving packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads file://../test_module.git
Directory: C:\Users\Rafael\Desktop\test
Output:
fatal: 'C:/Program Files/Git/test_module.git' does not appear to be a git repository

and the repository exists.
```

rally25rs:
Hey @rafaeldamasceno I opened a PR against your branch https://github.com/rafaeldamasceno/yarn/pull/1 that makes a couple tweaks, and implements the stuff I was talking about with `show-ref`.

Could you try it out and see if it still works as you would expect? If so, then merging that PR into your branch _should_ make the combination show up here in this PR.

We really appreciate your effort in putting this PR together! 👍

rafaeldamasceno:
@rally25rs, your PR seems to break things again :( Using the same test environment as before:
```
> node ..\yarn\lib\cli\index.js
yarn install v1.3.2
info No lockfile found.
warning test@0.0.1: No license field
[1/4] Resolving packages...
error Couldn't find the binary git
```

It seems you misunderstood the original issue: the problem wasn't cloning local repositories. The problem was that if the repository directory ended **_specifically_** in `.git`, it was automatically assumed to be a remote repository. Our current fix simply fixes this. I'm not even sure that it clones them or simply imports them as any other non git module.

I've accepted the PR and, in the meanwhile, removed the part where you remove the trailing `.git`. Then, it works as expected. However, I'm not sure this is the behavior you're expecting. Using `npm install`, it symlinks the repository directory (the full `test_module.git`) as `test_module` which the expected name of the module.

rally25rs:
OHHH! So your local directory is actually named something like `~/Projects/my-project.git`? Yeah I can see where that could cause an issue where I put in the trim of `.git`. I should probably chek for a directory existance with the .git first then trim it if it doesn't exist.

What I was thinking is that  you have a local git repo `~/Projects/my-project` and want to use it as a git repo, not just a filesystem copy or symlink, and the package.json entry

```
"my-project": "file:../my-project.git"
```

will feed `../my-project.git" to the `git-resolver`. Without the `.git` extension, it uses the `file-resolver` and just flat copies or symlinks the local dir, not using `git`.

I trimmed the `.git` so that it would run the `git` command from the `../my-project` directory. I wasn't thinking about if the directory actually ended with `.git`. Sorry... I'll try to find a way to fix that or make it work in both cases...

rally25rs:
You know what, I'm starting to wonder about this entire feature...

The original issue mentions having a dependency specified as:

```
"myBackendModule": "file:../myBackendModule.git"
```

I'm wondering if this should be run through the `FileResolver` or the `GitResolver`. In Yarn currently it goes through the `GitResolver` because it ends with a trailing `.git`, so we try to `git clone` it.

However if it was passed to the regular `FileResolver` then I assume it would work... it's unclear from the original issue...

Perhaps the real fix to this is that `file:` should always route to the `FileResolver` even if it ends with a trailing `.git`. But I have no idea if there was actual intent by whoever wrote this originally to support cloning from local "bare" git repos (unpushed repos).

#5017 asks for us to support what amounts to bare repos or at least be able to get unpushed branches from a local repo. This is actually possible through the stuff already committed to this PR, but I'm now really wondering if `file:` protocol should _maybe_ use the FileResolver _or_ the GitResolver. It seems like supporting a local bare repo would best be done through a special separate protocol, something like `"my-project": "git+file:../my-project"`

-----

@yarnpkg/core any thoughts on this are appreciated! Was `GitResolver` really intended to handle local repos? Or is it a bug that `file:` dependencies can route into it if they end with `.git`?

In other words, in current Yarn v1.3.2:

* "file:../my-project" uses FileResolver
* "file:../my-project.git" uses GitResolver

Intentional or bug?

rafaeldamasceno:
That's also what I was talking about: I never fully understood if it was supposed to be locally cloned or not, but the issue was about it not being able to add the dependency at all because of conflicting indicators `file:` and `.git`. To be fair, I'm not sure it currently does locally `git clone` them or the returned type simply breaks `GitResolver` and makes it move on to the next one. Did anyone else run them using the test suite in https://github.com/yarnpkg/yarn/pull/5026#issuecomment-350369167? I actually think it does end up using `FileResolver` because there's no repository info on the generated `node_modules/test_module/`.

rally25rs:
Dug through the code a bit further and I'm pretty sure that what it's doing now (with the changes we have made for this PR) is that it runs through the GitResolver (because of the trailing .git). Then Yarn runs `git archive` in the directory to get the archive at a certain hash as a tar file, and extracts it into the Yarn cache. Then it copies from the cache to the project's `node_modules/` just like a normal remote git repo's code.

Unless someone else from Yarn core team has any better ideas, I'm thinking what we should do is make `file:` always pass through FileResolver. Then add a new `git+file:` protocol that does what it's doing now; treat a path as a git repo.

@rafaeldamasceno if you want to help work that out, you are welcome to, or I could probably whip it up in a couple hours tomorrow and submit another PR for you to try out. Either way is cool with me.

Sorry again for all the confusion. The title of this PR read "Local .git repositories are now locally cloned" so I just came into it assuming the goal was to get `file:` deps cloning from the local dir. My fault for not reading through all the issue comments. 😞

rally25rs:
@rafaeldamasceno sorry to go around this circle again, but https://github.com/rafaeldamasceno/yarn/pull/3
I apparently copy/paste added a new test but didn't actually make sure it passed before committing it. 😞 

With the previous PR's changes, does this behave as you would expect it to for your `file:` dependency? I had set up a unit test to reproduce that case and it passed, so 🤞 it should work for you.

rally25rs:
OK now the CI tests are only failing for the same reason as `master` so this should be ready to merge. Since I've contributed work to this PR now, adding @BYK as a requested reviewer.

rafaeldamasceno:
Everything seems good here 😄 

I can confirm that now the behavior is exactly as stated in https://github.com/yarnpkg/yarn/pull/5026#issuecomment-350576887

I have noticed, though, that with FileResolver, it seems to copy hidden files and not keep them hidden (.git folder), but seeing that has nothing to do with all this, I'd say it's ready to be merged. Thanks for all the help 🎉

rally25rs:
**Hold off on a merge**

I just discovered that this is a valid git url: ` "my-proj": "csb-ric/my-proj.git"` that should hit github.com, not treat it as a local file.

There are way too many undocumented URI possibilities 😢

rally25rs:
@rafaeldamasceno I submitted a new PR to you that merges `master` into this branch to fix the CI tests, and fixed a potential parsing error in the `git` output.

https://github.com/rafaeldamasceno/yarn/pull/4

Thanks!

rally25rs:
@rafaeldamasceno sorry to keep doing this to you, but apparently I left a bad test in the last PR. Not sure how I managed to do that... 😞 
https://github.com/rafaeldamasceno/yarn/pull/5

rally25rs:
Yay tests finally pass. @BYK what do you think? 👍 / 👎 ?

The TL;DR version of this change is:

* `"some/repo.git"` will be resolved with GithubResolver and turned into an SSH github URL (no change from previous behavior)

* `"file:../test.git"` will be resolved with FileResolver as a plain directory (not processed as a git repo, this is a change from previous behavior, it used to go through GitResolver incorrectly)

* `"git+file:../test#branch3"` will be resolved with GitResolver as a local git repo (so a branch or hash or tag can be specified just like a remote git repo, this is new functionality)

arcanis:
Seems like Jest had a concurrency issue. Sounds unrelated to this PR.

PR #5033, opened 2017-12-04
feat(help) Add command descriptions to commander output (#1119)

### LINKED ISSUE #1119: yarn help COMMAD does not output command description
When typing the following for any `COMMAND`

```
$ yarn help COMMAND
```

the output is `usage` and `options`, but there is no `description` about what the command actually does.

One would expect to see a command description. For example:

```
$ yarn help install

    Install all the dependencies listed within `package.json` in the local `node_modules` folder

    Usage: `yarn install [flags]`

    Options:

        ...
```

node.js version `6.7.0`
yarn version `0.15.1`

Marko50:
Resolves  #1119

**Summary**

#1119 states when typing help for any command,  the output is usage and options, but there is no description about what the command actually does.

This pull request makes use of the setFlags function and the commander.description method to add and print the description of every documented command.

**Test plan**

yarn run test-only run tests for listing cached packages, removing all packages, removing a specific package by name


rally25rs:
Thanks for the contribution. Could you merge the latest `master` branch into your branch please. I think that will make the CI tests pass again.

Also, I noticed some of the help descriptions start with a capital letter and some don't, and some end with a period and some don't. Could you make them the same?

Thanks!

excelsior97:
I did that but test-macos-node8 is still failing, any ideas why?

Also changed the help descriptions to be more consistent, aswell as some minor changes to pass the lint test

Issue #816, opened 2016-10-12
SHA1 -> SHA256


jorangreef:
I have done a quick check but it looks like Yarn is using SHA1 for checksumming?

If so, I think it would be a good time to switch to SHA256.

ljharb:
it's worth noting that SHA1 is achievable in POSIX, but SHA256 is much harder to [make work](https://github.com/creationix/nvm/blob/master/nvm.sh#L1061-L1087) cross-platform.

Daniel15:
>  SHA256 is much harder to make work cross-platform.

Just need to use PowerShell everywhere, it has it built-in (`$hash = (Get-FileHash -Path $file -Algorithm SHA256).Hash`) 😛 

Node's crypto library supports SHA256, I think we'd use that.
```js
> var crypto = require('crypto')
undefined
> crypto.createHash('sha256').update('Hello World').digest('hex');
'a591a6d40bf420404a011733cfb7b190d62c65bf0bcda32b57b277d9ad9f146e'
```

Daniel15:
Looks like there's three hashing functions in the codebase:

["hash" in src/util/misc](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/util/misc.js#L6), uses SHA256
["HashStream" in src/util/crypto](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/util/crypto.js#L15), uses SHA1
["hash" in src/util/crypto](https://github.com/yarnpkg/yarn/blob/0f1db96846badbbe3b43c470c05196eb29437fb4/src/util/crypto.js#L7), takes the algorithm as an argument and defaults to MD5

`hash` in src/util/misc and `hash` in src/util/crypto should probably be unified and they should all use SHA256, I think we'd need to think about backwards compatibility though.

kittens:
Thanks for the issue! The shasums we use are given to us by the npm registry so there's not much we can do about this unfortunately. 😢

Daniel15:
@kittens - Can we compute our own hash once the file has been downloaded, and store that in the local cache?

graingert:
might as well jump straight to SHA3

bestander:
Now it is time to reconsider the issue.
A few problems to solve first:
- forwards compatibility / migration path
- speed concerns
- should we just ignore sha1 coming from npm or use it anyway?

Maybe sha2/3 could be an additional field?

[...]

PR #5042, opened 2017-12-05
feat(checksums): add integrity field with sha512 authentication to yarn.lock

imsnif:
**Summary**
Continuing the discussion in: https://github.com/yarnpkg/yarn/issues/816
This turned out to be a feature that touched quite a few parts of the code base. The diff is a little big, mostly because this required (minor) changes in a lot of the tests and fixtures (since we've added a new field to yarn.lock).

I tried to adjust the commits to "tell a story", to make this PR easier to follow. That being said, if anything is unclear, I would be happy to explain or adjust it to make it more understandable (I have made an effort to do so in the code as much as possible).

Since this is not a trivial change, what guided me through it is the will to concentrate on it as much as possible without changing other things. Thus for example, I did not change the `validationStream` in other non-tarball fetchers to use the `ssri` module, even though it creates an inconsistency, so as not to distract from the feature at hand. I think this PR does introduce some wanted changes (such as the stream example) but I believe they are all very minor, and would be best fixed in additional refactoring PRs (that I or someone else will perform once this is approved and merged).

**Test plan**
I added both integration and unit tests where appropriate.
Please note though: since there are no tests for `publish` (unless I missed them!) I did not add tests, because I felt it would be out of scope. I tested this manually with `verdaccio` and it worked. I would urge the reviewer to do the same to make sure I was not mistaken!


imsnif:
Regarding CI issues: The issues on previous versions of node in appveyor seem to be mostly due to different behaviour of the Buffer object. These can be trivially remedied and I'll get to them in the next few days.
Regarding the failed circleci builds: I'm not quite sure what the issue is here. It seems not to recognize the `ssri` module. Is there some sort of caching involved? Any tips on debugging this?

zkat:
I recommend you make sure your tests also check that this behaves if you hand it an `integrity` field with `sha384-deadbeef`, `sha3-c0ffee`, and `madeupalgorithm-abad1dea`, and that things still _work_ in those cases.

I also recommend you add tests to make sure `?option` syntax is handled reasonably well: `sha1-deadbeef?foo` and `sha512-deadbeefc0ffee?key=value1234!@#` should both be ok.

imsnif:
Thanks @zkat ! I'll add those test cases (the `madeupchecksum` scenario would even require a small code adjustment to get a proper error message).
Would love to hear more comments.

imsnif:
Hey @zkat - I added the test cases (and made the relevant code changes) you mentioned. If you have the time and feel like commenting, I'd very much appreciate it!

@BYK - the CircleCI builds for node6 and node8 are failing. I think this is because they cache `node_modules` and I added the `ssri` module which they thus do not see. I do not have permission to clear the cache. If you have time to take a look, it would be great! I fixed the other issues that made some builds fail previously.

BYK:
@imsnif thanks a lot for the PR and heads up about the cache issues. I've submitted #5060 to fix the invalid cache setup. You can merge from that and see if it fixes your builds. Or wait until it lands on master and merge from there.

Right now, as long as Linux tests pass, I'm okay until right before merging.

imsnif:
Thanks @BYK ! I think I'll wait with merging (either from #5060 or from master) until we finish discussing this, seeing as you're okay with it. I have a feeling it will make reviewing this easier. :)

BYK:
@imsnif merged to master. Merge from there and your builds should pass. I'll review the code later. I've also pinged a few more people since this is a fairly large change. Thanks a lot for all the effort you've put into this.

imsnif:
@BYK @arcanis @bestander - Hello friends! I hope you don't mind me making this public ping regarding the review progress. I'd very much like to continue working on this and see it merged. Hoping we'll find time soon!

bestander:
@imsnif, thanks for pinging, this is a right thing to do, no need to apologize.
Could you rebase the PR to merge conflicts

imsnif:
Hey @bestander , thanks!
I rebased and pushed (as you can see) there are some non-trivial conflicts (namely the offline mirror feature makes some assumptions that contradict some of my assumptions when making this PR). :) It's no biggy though, I started working on it and hoping to update and have this fixed tonight or tomorrow. Will keep this thread posted.

imsnif:
Hey @bestander - thanks again for taking a look at this!

So the rebase is done and I made the necessary fixes. Most of the stuff is pretty straightforward (namely adding the `integrity` field to more fixtures), but there is one tricky part.

Here: https://github.com/imsnif/yarn/blob/master/src/resolvers/registries/npm-resolver.js#L195
For convenience (and future generations) the relevant code:

```javascript
if (this.config.packBuiltPackages && shrunk.prebuiltVariants && shrunk._remote) {
  const prebuiltVariants = shrunk.prebuiltVariants;
  const prebuiltName = getPlatformSpecificPackageFilename(shrunk);
  const offlineMirrorPath = this.config.getOfflineMirrorPath();
  if (prebuiltVariants[prebuiltName] && offlineMirrorPath) {
    const filename = path.join(offlineMirrorPath, 'prebuilt', prebuiltName + '.tgz');
    const {_remote} = shrunk;
    if (_remote && (await fs.exists(filename))) {
      _remote.reference = `file:${filename}`;
      _remote.hash = prebuiltVariants[prebuiltName];
      _remote.integrity = ssri.fromHex(_remote.hash, 'sha1').toString();
    }
  }
}
```

Here, I added the line starting with `_remote.integrity` so that in case of an offline mirror the integrity check will not fail.
The gotcha here is that this means an offline mirror will *always* use `sha1` encryption. I think this is both not terrible (because if someone changes files on your HD they can also change the checksum) and easily fixed (we simply need to use the integrity field as the key to the offline packages where it is available).
I didn't go into the fix in this PR though because its scope is already quite big. I wanted to add the minimum possible changes to make it work and make these refactors/adjustments in later PRs.

The other thing is (in the same code snippet) - I had to make the if in that same block directed at the `_remote` object directly rather than at `shrunk._remote`. This makes little sense to me, but it was in order to make flow happy. :)
Flow and I are not good friends yet, so if there is a better way to do it I'd love to be corrected!

Would love to hear your thoughts. What else do we need to do in order to get this merged?

src/resolvers/registries/npm-resolver.js
```diff
@@ -187,21 +188,35 @@ export default class NpmResolver extends RegistryResolver {
...
       }
+    }
+    if (shrunk && shrunk._remote && shrunk._remote.integrity) {
+      // if the integrity field does not exist, it needs to be created
       return shrunk;
     }
 
-    const info: ?Manifest = await this.resolveRequest();
+    const desiredVersion = shrunk && shrunk.version ? shrunk.version : null;
+    const info: ?Manifest = await this.resolveRequest(desiredVersion);
```
arcanis: Why do we need to resolve to a different version than before?

src/lockfile/index.js
```diff
@@ -102,6 +124,20 @@ export default class Lockfile {
...
+  // if true, we're parsing an old yarn file and need to update integrity fields
+  entriesExistWithoutIntegrity(): boolean {
+    if (!this.cache) {
+      return false;
+    }
+    return Object.keys(this.cache).some(key => {
+      if (this.cache && this.cache[key] && !/^.*@(file:|http)/.test(key)) {
+        return !this.cache[key].integrity;
+      } else {
+        return false; // some entries are not supposed to have integrity, eg. files and web links
```
arcanis: Why's that?

src/fetchers/tarball-fetcher.js
```diff
@@ -65,58 +67,56 @@ export default class TarballFetcher extends BaseFetcher {
...
+    let supportedIntegrity;
+    try {
+      supportedIntegrity = this._supportedIntegrity();
+    } catch (e) {
+      if (this.config.updateChecksums) {
+        // it's okay that we do not have any supported integrity, because we need to update it!
+        supportedIntegrity = {integrity: null, algorithms: ['sha1']};
+      } else {
+        reject(
+          new SecurityError(
```
arcanis: I think we should throw this `SecurityError` rather than calling `reject` - `reject` is supposed to be useful for async code only (like in the `finish` callback). It doesn't make sense to report an error but still return something.

arcanis:
Thanks a lot for your work on this! I think we're pretty close now from being able to merge it. I put a few questions here and there to better understand some decisions but overall it looks solid and well covered.

> This makes little sense to me, but it was in order to make flow happy. :)

It's because Flow thinks that the `fs.exists` call might make `shrunk._remote` null, which makes your check useless (from Flow's perspective). When you bind it to a variable before checking it, Flow knows for sure that its value won't be changed by function calls side effects.

src/resolvers/registries/npm-resolver.js
imsnif: Before this feature, we would resolve the cached tar file if we had one: https://github.com/yarnpkg/yarn/blob/615ee94b0d88dfbd728d03114aa031cd11e838e1/src/resolvers/registries/npm-resolver.js#L196
Here, we add the condition that said cached file has to have an integrity field: https://github.com/yarnpkg/yarn/pull/5042/files#diff-3aebe0cb09fdaf76a6f7303534a3ee26R200
If it doesn't, we want to fetch the package again from the npm repository in order to get that field from there to update the lockfile for future checksum checks. When we do this, we want to make sure we get the exact same version as specified in the lockfile and not (as the method would without adding the `desiredVersion` argument) a version that satisfies the semver range

src/fetchers/tarball-fetcher.js
imsnif: In theory I completely agree. The problem is that this stream pipeline is created inside a `Promise` (eg. https://github.com/imsnif/yarn/blob/a9e7ef26ea2ab534a160a2fbdf892a02368bdd24/src/fetchers/tarball-fetcher.js#L142). Since this is a synchronous error (as you say, we have enough information to throw this error before we return the `Promise`), it might be better to check it and throw appropriately before creating the pipeline using the `createExtractor` method (or better yet, create the pipeline before returning the `Promise` and then throw to our heart's content).
I haven't experimented with this, but my initial gut feeling tells me this would either involve duplicating a lot of logic (since we also have integrity checks that we need to perform and reject inside the pipeline) or a not-so-small refactoring of the tarball fetcher. I am very much in favour of the latter and might even volunteer for it if you'd like, but I feel it might be out of the scope of this PR. I hope you agree?
As for the former option (duplicating logic), I have a feeling it will make the code less coherent. I think the current solution, while not ideal, is the lesser evil at the moment... what do you think?

src/resolvers/registries/npm-resolver.js
imsnif: I'll start with the short and honest answer: because it would require a great deal more changes than this already monstrous and hard-to-swallow PR. :)
Right now, what this PR does is create compatibility with the npm integrity feature. It mirrors the integrity from the npm repository (and creates it when publishing to npm through yarn). All of the integrity checks are performed against fields that at some point were in a (npm) registry json file. This whole feature is built to support that. For everything else, we still have the `hash` field which - while not perfect - can keep serving more than just backwards compatibility for now. This feature does give a good infrastructure to eventually replace the `hash` field (maybe in a new lockfile version?) though. When it does that, it can replace all the other roles of the `hash` field such as being a checksum for non-registry packages.
I hope we see eye to eye on this one?

src/lockfile/index.js
imsnif: I'll start with the short and honest answer: because it would require a great deal more changes than this already monstrous and hard-to-swallow PR. :)
Right now, what this PR does is create compatibility with the npm integrity feature. It mirrors the integrity from the npm repository (and creates it when publishing to npm through yarn). All of the integrity checks are performed against fields that at some point were in a (npm) registry json file. This whole feature is built to support that. For everything else, we still have the `hash` field which - while not perfect - can keep serving more than just backwards compatibility for now. This feature does give a good infrastructure to eventually replace the `hash` field (maybe in a new lockfile version?) though. When it does that, it can replace all the other roles of the `hash` field such as being a checksum for non-registry packages.
I hope we see eye to eye on this one?

imsnif:
Hey @arcanis  - thanks for your input. I very much appreciate it! I responded inline to your comments.

Also, thanks for the `flow` explanation. It definitely makes sense. I'm slowly getting to understanding it better. :)

src/resolvers/registries/npm-resolver.js
arcanis: Makes sense 👍

src/fetchers/tarball-fetcher.js
arcanis: I'm not sure I understand - if we throw instead of rejecting, since we are in a promise constructor, the promise will be automatically rejected. The only difference I can see would be that [these lines](https://github.com/imsnif/yarn/blob/a9e7ef26ea2ab534a160a2fbdf892a02368bdd24/src/fetchers/tarball-fetcher.js#L144-L146) would not be executed. Am I missing something?

src/lockfile/index.js
arcanis: That sounds fair.

src/fetchers/tarball-fetcher.js
imsnif: This was my misreading of the code, you are of course correct! For some reason I thought there were some async things before it and that the pipeline was being passed around. I tested it locally with throwing and it looks great.
Just pushed a fix and so far the CI seems happy.

rally25rs:
It appears that the `yarn.lock` in the `master` branch now contains the `integrity` field
https://github.com/yarnpkg/yarn/blob/master/yarn.lock#L8

but since yarn doesn't create that field without this PR merged, anyone else who runs `yarn install` in the project ends up with a large `yarn.lock` diff of these integrity lines being deleted.

Was that change to the `master` `yarn.lock` in error or intentional?

imsnif:
ping @BYK , @arcanis - when one of you has time, would be great to give this one final push. The TLDR of what neess to be done IMO is:
1. Go over the code with a pair of fresh eyes
2. Manually test `yarn publish` and backwards compatibility (both of them I've done but I'd love to be double checked).

Thanks!

arcanis:
The last remaining test throwing is caused by a flakiness when retrieving archives from github on AppVeyor (will probably end up skipping this test on Windows). This looks good to merge!

HUGE thank to @imsnif for your work on this! Sorry it took so long, but we're finally ready to merge this in master! Thanks to @BYK too for helping getting the PR in shape. Really great work 😃👍

imsnif:
Thanks @arcanis and @BYK! Great to see this merged.
Been a fun ride. :)

PR #5044, opened 2017-12-05
Implementation of small feature suggested in issue #624

### LINKED ISSUE #624: Should yarn run ignore .cmd files for binary scripts on Windows?
Example for v0.15.1:

cmd_2016-10-11_17-01-43

tiagolascasas:
fixes #624 

**Summary**

This pull request aims to fix issue #624 . It was suggested on that issue's discussion that when running the `run` command on Windows only `.cmd` scripts should be considered, and that these should show up without the extension when displayed and allow the user to write the name of the command without writing the extension.

**Test plan**

The code has undergone successful manual tests and passes the lint tests. An example of the new usage of the `run` command on Windows (using `flow` as an arbitrary script) can be the following:

```
E:\Git\yarn\bin>yarn run
yarn run v1.3.2
error No command specified.
info Commands available from binary scripts: acorn, atob, babylon, browserslist, commitizen, errno, escodegen, esgenerate, eslint, esparse, esvalidate, flow, git-cz, gulp, gunzip-maybe, handlebars, jest-runtime, jest, js-yaml, jsesc, jsinspect, json5, loose-envify, miller-rabin, mkdirp, prettier, regjsparser, rimraf, sane, semver, sha.js, shjs, sshpk-conv, sshpk-sign, sshpk-verify, strip-indent, strip-json-comments, uglifyjs, user-home, uuid, webpack, which
info Project commands
   - build
      gulp build
   - build-bundle
      node ./scripts/build-webpack.js
   - build-chocolatey
      powershell ./scripts/build-chocolatey.ps1
   - build-deb
      ./scripts/build-deb.sh
   - build-dist
      bash ./scripts/build-dist.sh
   - build-win-installer
      scripts\build-windows-installer.bat
   - commit
      git-cz
   - dupe-check
      yarn jsinspect ./src
   - lint
      eslint . && flow check
   - prettier
      eslint src __tests__ --fix
   - release-branch
      ./scripts/release-branch.sh
   - test
      yarn lint && yarn test-only
   - test-coverage
      node --max_old_space_size=4096 node_modules/jest/bin/jest.js --coverage --verbose
   - test-only
      node --max_old_space_size=4096 node_modules/jest/bin/jest.js --verbose
   - watch
      gulp watch
question Which command would you like to run?: flow
$ E:\Git\yarn\node_modules\.bin\flow.cmd
Launching Flow server for E:\Git\yarn
Spawned flow server (pid=6504)
... (rest of the output doesn't matter)
```

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -8,7 +8,6 @@ import {registries} from '../../resolvers/index.js';
 import * as fs from '../../util/fs.js';
 import map from '../../util/map.js';
 
-const os = require('os');
 const leven = require('leven');
 const path = require('path');
 const {quoteForShell, sh, unquoted} = require('puka');
@@ -31,18 +30,17 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     if (!visitedBinFolders.has(binFolder)) {
       if (await fs.exists(binFolder)) {
         for (const name of await fs.readdir(binFolder)) {
-		  //consider only .cmd scripts on Windows and show them without the extension - #624
-		  if (process.platform === 'win32') {
-			  if (name.indexOf(".cmd") !== -1) {
-				let strippedName = name.substring(0, name.indexOf(".cmd"));
-			  	binCommands.push(strippedName);
-				scripts[name] = quoteForShell(path.join(binFolder, name));
-			}
-		  }
-		  else {
-	          binCommands.push(name);
-	          scripts[name] = quoteForShell(path.join(binFolder, name));
-	  	  }
+          //consider only .cmd scripts on Windows and show them without the extension - #624
+          if (process.platform === 'win32') {
+            if (name.indexOf('.cmd') !== -1) {
+              const strippedName = name.substring(0, name.indexOf('.cmd'));
+              binCommands.push(strippedName);
+              scripts[name] = quoteForShell(path.join(binFolder, name));
+            }
+          } else {
+            binCommands.push(name);
+            scripts[name] = quoteForShell(path.join(binFolder, name));
+          }
         }
       }
       visitedBinFolders.add(binFolder);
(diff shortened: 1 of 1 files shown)


src/cli/commands/run.js
```diff
@@ -30,8 +30,17 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     if (!visitedBinFolders.has(binFolder)) {
       if (await fs.exists(binFolder)) {
         for (const name of await fs.readdir(binFolder)) {
-          binCommands.push(name);
-          scripts[name] = quoteForShell(path.join(binFolder, name));
+          //consider only .cmd scripts on Windows and show them without the extension - #624
+          if (process.platform === 'win32') {
+            if (name.indexOf('.cmd') !== -1) {
```
rally25rs: Might be more "correct" to make sure that the name ends with ".cmd". This just checks that it is anywhere in the string (so "test.cmd.bak" would match).

Maybe use `name.match(/\.cmd$/)` instead?

src/cli/commands/run.js
```diff
@@ -30,8 +30,17 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
+              const strippedName = name.substring(0, name.indexOf('.cmd'));
```
rally25rs: Same potential issue as the above comment. Should check that ".cmd" is the end of the string.

tiagolascasas:
@rally25rs You are absolutely right, that was a bit silly. I pushed a fixed version just now.

rally25rs:
It looks like several unit tests are failing on Windows/AppVeyor. From a quick glance, it looks like some of the test fixtures only provide a bin without a `.cmd` version.

For example: `__tests__/fixtures/run/no-scripts/node_modules/.bin/` contains `cat-names` but not a `cat-names.cmd` so the `run` command output has changed and causes a test to fail.

You can run the tests locally using `yarn run test` (be warned, Jest is terribly slow on Windows. It takes something like 8 minutes to run on my Surface Pro 3 i7)

tiagolascasas:
I have tried to open up an exception in the parsing of the arguments solely for `cat-names`, and the tests now pass (locally, at least). It is a very crude approach and I am not sure if it is the proper way of doing this, but I can't see any other option other than changing `cat-names` itself...

rally25rs:
Could be a bit tricky... my first guess was to just add a `cat-names.cmd` to the directory too, since on Windows there would be both. However that might make the tests fail un Unix and Mac, since `.cmd` files normally wouldn't be there.

It might be a good idea to add the `cat-names.cmd` file to the test fixtures where needed, then add some logic not _not_ list `.cmd` bins when not on Windows.

olingern:
@rally25rs @tiagolascasas I'd like to close this unless you all feel that this feature adds a lot of value for windows users. I'll comment over at #624 as well.

tiagolascasas:
@olingern The conclusion I reached back when I tried to do this was that it was more trouble than it is worth. It's not that simple to determine what is a valid script for Windows and what isn't. I'm ok with closing this.

olingern:
@tiagolascasas ok, sorry things didn't work out :bowing_man:

PR #5046, opened 2017-12-05
Include bins when using link: dependency

kjbekkelund:
**Summary**

When using `link:` dependencies `bin`s are not included when installing because of the special-cased "mock response" for symlinked packages: https://github.com/yarnpkg/yarn/blob/master/src/package-fetcher.js#L28-L31

**Test plan**

I've added a test and a check before updating the pkg.


arcanis:
The original design behind the `link:` protocol is that it creates a symlink and that's it (it currently tries to read a package.json file, but that could actually be considered a bug). I'm can see why you'd like this feature, but I'm not sure we should change the `link:` protocol semantic :/

cc @byk, wdyt?

kjbekkelund:
Ah. My change was based on how Lerna does it, https://github.com/lerna/lerna/blob/7b3738eaed168150005cdd73966b207ecbe33c91/src/PackageUtilities.js#L253-L261.

I can definitely see the point about not wanting to change the `link:` semantics, though.

kjbekkelund:
Looks like npm5 `file:` dependencies also symlink binaries, btw:

```
~/dev/explore/npm5-file-bin/foo$ ls -l node_modules/.bin 
total 0
lrwxr-xr-x  1 kim  staff    11B 15 Jan 15:14 bar@ -> ../bar/bin.js
```

BYK:
I'm okay with what this patch is fixing (which would yield to potentially inconsistent linking of binaries from `link:` packages) but not with how it does this. I think the "proper" way to do this is to try reading a manifest file at fetching stage and gracefully drop to the "mock" one. We should also plan to fix how we special case `link:` protocol everywhere (like in linking) if poossible.

kjbekkelund:
I agree. I'll see if I've got time to PR a better fix to this. Closing this PR.

PR #5059, opened 2017-12-08
fix(linking): Handle missing package when using --ignore-optional

### LINKED ISSUE #4876: --ignore-optional is broken
Bug!

```bash
$ yarn add webpack webpack-dev-server
...
Done in 6.34s.
$ node_modules/.bin/webpack-dev-server 
No configuration file found and no entry configured via CLI option.
When using the CLI you need to provide at least two arguments: entry and output.
A configuration file could be named 'webpack.config.js' in the current directory.
Use --help to display the CLI options.
# This indicates the module installed properly.
$ rm -rf node_modules
$ yarn install --ignore-optional
$ node_modules/.bin/webpack-dev-server 
module.js:538
    throw err;
    ^

Error: Cannot find module 'fs.realpath'
    at Function.Module._resolveFilename (module.js:536:15)
    at Function.Module._load (module.js:466:25)
    at Module.require (module.js:579:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/benesch/Desktop/t/node_modules/glob/glob.js:44:10)
    at Module._compile (module.js:635:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
```

`yarn install --ignore-optional` should install a working set of dependencies. It appears a transitive dep is somehow causing `fs.realpath` to look like it's optional, when it is in fact required.

macOS Sierra 10.12.6

```
$ node --version
v8.9.0
$ yarn --version
1.3.2
```

/cc @arcanis @BYK I suspect this might be related to #3976.

### LINKED ISSUE #5054: Dependency is missing when using yarn --ignore-optional
bug

`yarn --ignore-optional ` doesn't install all (none optional) dependencies 

Related to https://github.com/yarnpkg/yarn/pull/4448

Steps to reproduce:
package.json:
`{ "dependencies": { "some-package": "^4.1.0", "yargs": "^9.0.1" } }`

now some-package has this as optionalDependency:
`"yargs": "^7.0.2"`

the result of yarn --ignore-optional is missing package: is-fullwidth-code-point
which dependency path is: yargs -> string-width -> is-fullwidth-code-point

Tested on:
node 6.11.2
yarn 1.3.2

### LINKED ISSUE #5152: Doesn't install correct dependencies for 1.3.2
Bug

Installation of packages fails when installing in production, Yarn 1.3.2 errors with:

`Error: Cannot find module 'are-we-there-yet'`

This does *not* happen with Yarn 0.18.1, where the installs complete successfully.

These are the relevant dependencies for a particular package that fails:

```
"dependencies": {
    "@meanie/angular-analytics": "^2.0.0",
    "@meanie/angular-api": "^3.0.1",
    "@meanie/angular-convert": "^2.0.1",
    "@meanie/angular-duplicate-requests-filter": "^2.0.0",
    "@meanie/angular-focus": "^2.0.0",
    "@meanie/angular-key-codes": "^2.0.0",
    "@meanie/angular-log": "^2.0.0",
    "@meanie/angular-modal": "^3.0.0",
    "@meanie/angular-storage": "^2.0.1",
    "@meanie/angular-store": "^2.0.1",
    "@meanie/angular-url": "^2.0.1",
    "angular": "1.6.8",
    "angular-animate": "1.6.8",
    "angular-cookies": "1.6.8",
    "angular-messages": "1.6.8",
    "angular-sanitize": "1.6.8",
    "angular-touch": "1.6.8",
    "angular-ui-router": "^0.4.2",
    "fastclick": "^1.0.6",
    "marked": "^0.3.9",
    "meanie-angular-form-controls": "^1.4.5",
    "moment": "^2.20.1",
    "ng-infinite-scroll": "^1.3.0",
    "raven-js": "^3.21.0"
  },
  "devDependencies": {
    "babel-polyfill": "^6.23.0",
    "babel-preset-es2015": "^6.24.0",
    "browser-sync": "^2.18.6",
    "chalk": "^1.1.3",
    "del": "^2.2.2",
    "git-rev-sync": "^1.8.0",
    "gulp": "git+https://github.com/gulpjs/gulp#4.0",
    "gulp-angular-templatecache": "^2.0.0",
    "gulp-autoprefixer": "^3.1.1",
    "gulp-babel": "^6.1.2",
    "gulp-concat": "^2.6.1",
    "gulp-csso": "^3.0.0",
    "gulp-file": "^0.4.0",
    "gulp-htmlclean": "^2.7.16",
    "gulp-inject": "^4.3.0",
    "gulp-ng-annotate": "^2.1.0",
    "gulp-ng-constant": "^1.1.0",
    "gulp-plumber": "^1.1.0",
    "gulp-preprocess": "^2.0.0",
    "gulp-remove-empty-lines": "^0.0.8",
    "gulp-remove-html-comments": "^1.0.1",
    "gulp-rename": "^1.2.2",
    "gulp-sass": "^3.1.0",
    "gulp-sourcemaps": "^2.4.1",
    "gulp-uglify": "^2.1.2",
    "gulp-wrapper": "^1.0.0",
    "http-proxy-middleware": "^0.17.4",
    "merge-stream": "^1.0.1",
    "npm-run-all": "^4.0.2",
    "mime": "^1.3.4",
    "netlify-cli": "^1.2.2",
    "replace-in-file": "^2.5.0",
    "rimraf": "^2.6.1",
    "s3-cli": "^0.13.0",
    "serve-static": "^1.12.1",
    "yargs": "^10.0.3"
  }
```

For the Yarn install to work.

Node.JS 8.9.3, Yarn 1.3.2, the build was failing on Netlify VM's.

Ref #761

kjbekkelund:
**Summary**

Closes #5054, closes #4876, closes #5152.

Currently when running `yarn --ignore-optional` required dependencies can be marked as optional because they exist in the tree of `optionalDependencies` of one of the dependencies (but it's required by some other non-optional dependency).

For example: `once` depends on `wrappy`, but `wrappy` is also in the chain from `fsevents` (`glob` -> `inflight` -> `wrappy`) which is an optional dependency of `chokidar`. So if the first `wrappy` comes from an "optional chain" when yarn processes it, it ends up being marked as `_reference.optional = true`, and the package hoister therefore doesn't mark it as required.

When running the released version of yarn with https://github.com/spalger/reproduce-issues/blob/master/yarn-ignores-non-optional-dependencies you'll see

```
~/dev/reproduce-issues/yarn-ignores-non-optional-dependencies
$ yarn --ignore-optional
yarn install v1.3.2-20171204.1856
[2/4] 🚚  Fetching packages...
✨  Done in 0.17s.

~/dev/reproduce-issues/yarn-ignores-non-optional-dependencies
$ yarn check --verify-tree
yarn check v1.3.2-20171204.1856
error "once#wrappy" not installed
error An unexpected error occurred: "Found 1 errors.".
```

This fix checks whether or not the parent is marked as required, and if so, marking the dependency as required _unless_ it's listed in the parent's `optionalDependencies`.

**Test plan**

I tested this implementation with both https://github.com/spalger/reproduce-issues/blob/master/yarn-ignores-non-optional-dependencies and https://github.com/yarnpkg/yarn/issues/4876, and in both cases `yarn check --verify-tree` succeeded after running `yarn --ignore-optional`.

Also added automated tests.


kaylieEB:
Woot! Thanks @kjbekkelund, do you feel comfortable trying to resolve this bug? Otherwise I can take a look it. Let me know :)

kjbekkelund:
@kaylieEB I've been exploring a bit, but I haven't had much progress yet. Would be great if you could take a look.

kjbekkelund:
I think it's related to `resolveToExistingVersion`. When it resolves for `once` it switches `optional` from `true` to `false`, but it doesn't walk its dependencies.

So it feels like there's a need to walk the dependencies in `addOptional` (https://github.com/yarnpkg/yarn/blob/9c2bbca92140b6e360dd566d31658d7af1fef8d5/src/package-reference.js#L115) and switch it for them too.

Thoughts?

kjbekkelund:
Ah, or maybe `_propagateRequired` is the location for a fix? https://github.com/yarnpkg/yarn/blob/ef8185b015c13f10c602fb695f91f391271d1f5c/src/package-hoister.js#L221

However, that depends on that `_reference.optional` value again, which is `true` instead of `false` here for `wrappy`

kjbekkelund:
@kaylieEB Looks like I got it running.

The problem is that you need to know if it was listed in `optionalDependencies` when deciding whether or not to make them required. To fix this I re-used the `hint` field, which is already used in this way in e.g. https://github.com/yarnpkg/yarn/blob/2065988f9667df2d349756ae040879e2bce6cf7a/src/cli/commands/install.js#L319-L321

Let me know if there's a better way to handle this.

kjbekkelund:
With the latest fix, this should also handle https://github.com/yarnpkg/yarn/issues/4876

kjbekkelund:
The tests are failing because of an unrelated change: `content-type-parser` just changed name to `whatwg-mimetype`, but `jsdom` hasn't pushed an update yet (https://github.com/tmpvar/jsdom/blob/224d28614bb8bb5f012c61545edd3767583d3d10/package.json#L25).

src/package-hoister.js
```diff
@@ -242,7 +242,17 @@ export default class PackageHoister {
           continue;
         }
 
-        const isMarkedAsOptional = depinfo.pkg._reference && depinfo.pkg._reference.optional && this.ignoreOptional;
+        const depRef = depinfo.pkg._reference;
+
+        let isMarkedAsOptional = depRef && depRef.optional && this.ignoreOptional;
```
BYK: We should be able to reduce this to:
```
let isMarkedAsOptional = depRef && depRef.optional && this.ignoreOptional && !(info.isRequired && depRef.hint !== 'optional');
```

What do you think? The comment you added should stay though, it is great!

src/package-request.js
```diff
@@ -52,6 +53,7 @@ export default class PackageRequest {
   config: Config;
   registry: ResolverRegistryNames;
   optional: boolean;
+  hint: ?string;
```
BYK: Would you mind changing this to an enum both to avoid unnecessary string usage all over the app and to avoid potential typing errors.

My suggestion is to add something like the following into constants.js and use that:
```
const REQUEST_HINTS = Object.freeze({
    dev: Symbol('REQUEST_HINT_DEV'),
    optional: Symbol('REQUEST_HINT_OPTIONAL'),
});
```

src/package-hoister.js
```diff
@@ -242,7 +242,17 @@ export default class PackageHoister {
...
+
+        // If it's marked as optional, but the parent is required and the
+        // dependency was not listed in `optionalDependencies`, then we mark the
+        // dependency as required.
+        if (isMarkedAsOptional && info.isRequired && depRef && depRef.hint !== 'optional') {
```
BYK: Can we not use `depRef.isRequired` instead of this? Or may be add `depRef.isOptional`?

kjbekkelund: I tried that at first, but it fails with "deeper optionals". E.g.

```
foo (in `optionalDependencies`) -> bar -> baz
```

Here both `bar` and `baz` will have `_reference.isOptional === true`. However, as this code doesn't mark the reference when performing this logic (only setting `depinfo.isRequired = true`), `baz` will stay `optional` as `bar._reference.optional === true`. If we changed the logic to also do a `depRef.setOptional(false)` in addition to setting `depinfo.isRequired` this approach would work. That was my initial fix, but I wasn't sure about doing that work from the package hoister. Thoughts?

kjbekkelund:
@BYK I ended up doing:

```
export type RequestHint = 'dev' | 'optional' | 'resolution' | 'workspaces';
```

as this field seems to be used for a couple things. That's likely a sign this should be cleaned up(?) Given https://github.com/yarnpkg/yarn/blob/19c8cd5b85b837278e74b42a94c4b5bb1b28fc35/src/cli/commands/outdated.js#L33 it definitely feels like the intention was to codify what type of dependency it is.

Thoughts? I'd of course prefer to not extend the scope of this PR ;)

kjbekkelund:
@BYK Just back from my holidays, so I thought I'd ping you about my suggestion above.

BYK:
@kjbekkelund I think the PR is fine but I have realized this is masking another issue which makes this part of the code as complex as it is right now which is marking all transient dependencies of optional dependencies as optional too, opening the door for potentially corrupt installations for optional dependencies.

Also we apparently already have a test case for this behavior: https://github.com/yarnpkg/yarn/blob/3e1c3a7ed27e017dcbe7c2a189de37fd705ff97f/__tests__/commands/install/integration.js#L956-L967

May be instead of adding all these new files, you can fix that test to cover transient dependencies (or just extend it).

What do you think?

kjbekkelund:
@BYK Finally had some time to get this done. Are you okey with this approach? Is the `fs.readdir` in the test okey, or do you want me to list separate `fs.exists` assertions? (as this ends up mentioning the integrity file)

src/cli/commands/install.js
```diff
@@ -282,7 +282,12 @@ export class Install {
         }
       }
 
-      const pushDeps = (depType, manifest: Object, {hint, optional}, isUsed) => {
+      const pushDeps = (
+        depType,
+        manifest: Object,
+        {hint, optional}: {hint: ?constants.RequestHint, optional: boolean},
```
BYK: Why make `hint` optional here?

src/constants.js
```diff
@@ -124,3 +124,5 @@ export const VERSION_COLOR_SCHEME: {[key: string]: VersionColor} = {
 };
 
 export type VersionColor = 'red' | 'yellow' | 'green' | 'white';
+
+export type RequestHint = 'dev' | 'optional' | 'resolution' | 'workspaces';
```
BYK: Don't see other hints being used anywhere else. Do we need anything other than `'dev'` and `'optional'` at the moment?

BYK:
I'll merge this since it fixes an important issue but I'd love to get answers to my final comments and maybe get a follow-up to address them.

src/constants.js
kjbekkelund: `resolution` is used in: https://github.com/yarnpkg/yarn/blob/5ef113f7db1201f48e22498374023ebe1e61b069/src/cli/commands/install.js#L281

and `workspaces` in: https://github.com/yarnpkg/yarn/blob/5ef113f7db1201f48e22498374023ebe1e61b069/src/cli/commands/install.js#L379

src/cli/commands/install.js
kjbekkelund: Because of https://github.com/yarnpkg/yarn/blob/5ef113f7db1201f48e22498374023ebe1e61b069/src/cli/commands/install.js#L327

Not sure if `constants.RequestHint | null` or `?constants.RequestHint` is preferred(?) Or maybe add a `production` hint and always require `hint`?

PR #5060, opened 2017-12-08
ci(circle): Fix cache key setup for proper node_modules sharing

BYK:
**Summary**

We had a tiered cache key setup for some reason (probably remnant of the pre-macOS builds config) which was breaking macOS builds when a new dependency was introduced due to common install was done on a Docker machine and cached with a key including the architecture. This patch changes that and ties everything to a single cache key.

**Test plan**

CircleCI builds should pass without issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,9 +18,7 @@ restore_cache: &restore_cache
   restore_cache:
     name: Restore node_modules cache
     keys:
-      - v1-node-{{ arch }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
-      - v1-node-{{ arch }}-{{ .Branch }}-
-      - v1-node-{{ arch }}-
+      - v1-node-{{ .Branch }}-{{ checksum "yarn.lock" }}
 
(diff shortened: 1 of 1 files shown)


rally25rs:
So we don't have to worry about the architecture any more? I would think it might make a difference between Mac and Linux for any native/build deps.

BYK:
>So we don't have to worry about the architecture any more?

We didn't do that already. The way we setup CircleCI right now is to have single, unified install step and then share the installed resources. This is not completely "safe" as you point out but we cab get away with it since we don't have any platform-specific dependencies and we build JS files that are platform agnostic. Doing it this way saves us time.

>I would think it might make a difference between Mac and Linux for any native/build deps.

Again, it may and if we ever pull in this kind of dependencies, we would need to separate the workflows into Linux and macOS which would be costlier and slower.

Daniel15:
This should be fine... We'll need to change the build and release process if we ever add native dependencies. Since we currently bundle Yarn as a single .js file, it doesn't take native dependencies into account at all.

Issue #5050, opened 2017-12-06
Sort options and commands in the help output


BYK:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

Currently, yarn has a seemingly random order when listing available commands and options. 

**What is the expected behavior?**

Sorting them alphabetically or some other consistent and easy to discover way would probably help people trying to read the help output. (cc @mjesun)

**Please mention your node.js, yarn and operating system version.**

Yarn 1.3.2

prabakarviji:
@BYK But it seems like commands list is already sorted whereas options are unsorted.

BYK:
Well, then only options. And also having a test case for both?

Koreris:
I created a pull request that solves this issue, feedback is welcome anyway :)

PR #5069, opened 2017-12-09
Fix for #5050 Sort options and commands in the help output

Koreris:
**Summary**

This is a feature that aims to make the use of the help command easier by sorting the options alphabetically when they're displayed on the screen.

**Test plan**

Using the command `yarn help` was enough to test the changes and see if they were being applied.

Before    
image

After  
image


arcanis:
The tests are not passing, can you try rebasing and updating the snapshots (`yarn test -u`)?

PR #5076, opened 2017-12-11
test(integration): Fix react-scripts test due to unexpected warning

BYK:
**Summary**

We don't expect any warnings when installing high-profile and well-tested packages like
react-scripts so we have an integration test that enforces 0 warnings. Unfortunately a package got renamed and started producing a warning about this so the tests started failing. This PR relaxes the test to only fail on peer dependency warnings.

**Test plan**

All builds and tests should pass.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -12,7 +12,7 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000;
 
 const path = require('path');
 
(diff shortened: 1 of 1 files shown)


BYK:
@rally25rs I like that suggestion. If we start to do this more in the future, I think investing into that makes a lot of sense. My hope was to introduce the "strict mode" some time and force no warnings at all but this recent problem made me less hopeful about that.

I'd say a rename is not serious enough to "warn" people. This should probably show up as an info but not sure if packages can control that.

PR #5080, opened 2017-12-12
fix(build): Use `tar --strip-components=1` instead of `--strip`

rubicks:
**Summary**

`tar ... --strip 1` is incorrect; should be `tar ... --strip-components=1`

* https://www.freebsd.org/cgi/man.cgi?tar(1)
* https://linux.die.net/man/1/tar

**Test plan**

Manually verified. Also builds should still succeed.

### DIFF
--- a/scripts/build-deb.sh
+++ b/scripts/build-deb.sh
@@ -36,7 +36,7 @@ PACKAGE_TMPDIR_ABSOLUTE=$(readlink -f $PACKAGE_TMPDIR)
 # Create Linux package structure
 mkdir -p $PACKAGE_TMPDIR/usr/share/yarn/
 mkdir -p $PACKAGE_TMPDIR/usr/share/doc/yarn/
-tar zxf $TARBALL_NAME -C $PACKAGE_TMPDIR/usr/share/yarn/ --strip 1
+tar zxf $TARBALL_NAME -C $PACKAGE_TMPDIR/usr/share/yarn/ --strip-components=1
 cp resources/debian/copyright $PACKAGE_TMPDIR/usr/share/doc/yarn/copyright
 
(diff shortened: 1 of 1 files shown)


BYK:
@Daniel15 can you look at this? Not sure if this is actually a problem since if it was, we would have broken builds for a very long time.

Daniel15:
Huh, that's strange, since the current version works fine. 😕 Maybe we don't need this option at all?

Did you confirm that the resulting package works as expected?

rubicks:
@Daniel15 sorry, was that directed at me?

Daniel15:
@rubicks Yeah. Are there any differences between the new package (after this change) vs the packages that were generated before this change?

rubicks:
@Daniel15 , none that affected the viability of the package as an installable artifact. I'm going to dig into it further now.

rubicks:
@Daniel15 , because `dpkg-deb` won't show me the contents of the debian package...

```
$ dpkg-deb -c artifacts/yarn_1.3.2_all.deb
dpkg-deb: error: archive 'artifacts/yarn_1.3.2_all.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up
```
...I'm unable to report any meaningful differences between `master` and `patch-1` with respect to this artifact.

Based on some little experiments I've run, it looks like both `tar` and `bsdtar` silently accept `--strip` to mean `--strip-components`. I don't know where to go from here with this issue, but I do feel like opening a ticket for that debian package weirdness.

arcanis:
I agree with the principle, the man seems clear, however I prefer not to merge this PR. We know that `--strip` is working on our pipeline, so switching to `--strip-components` seem to have no immediate benefit and would have a small risk of breaking something. Going to close this PR, but feel free to discuss it.

PR #5088, opened 2017-12-13
fix(resolution) Eliminate "missing peerDep" warning when dep exists at root level.

### LINKED ISSUE #4743: Workspaces complain about missing hoisted peer dependencies

peer deps that are hoisted to the repo root are seen as missing by each workspace

Add two workspaces with a Peer dependency and dev dependency on `react` (or anything in npm). Running `yarn` results in "[Workspaces package] has unmet peer dependency" warning

You can try it out with: https://github.com/jquense/react-widgets and would see complaints about react, react-dom, moment, and globalize being missing.

It not to warn

Yarn 1.2.0, node 8.5 and Mac Sierra

rally25rs:
fixes #4743

**Summary**

Previously, the logic in `PackageResolver` would `return` if a pattern was already resolved (ignore duplicate patterns). However, the check for `peerDependencies` needs to check all references.

This was causing an issue where if there was a peerDep on `depZ`, and `depZ` was resolved from a deeply nested transitive dep first, possibly `> depA > depB > depC > depZ`. And also `depZ` was directly included from package.json `> depZ` that second shallow `PackageReference` would not be tracked.

peerDep resolution would then _only_ see the reference from `> depA > depB > depC > depZ` which may be too far down the tree to satisfy the dependency, thus a missing peerDep warning was issued.

Fortunately the underlying data structure seems to already support the concept of multiple`PackageReferences` for a dependency, so I changed the PackageResolver logic to always call `PackageRequest .find()` which already had its own duplication check, so now adds all`PackageReferences` to the data structure. This causes the peerDep resolution process to check all included references to the dependency.

**Additionally**, this exposed a problem with `yarn why` where if a package with the exact same pattern was added at the root package.json, and by transitive deps, `yarn why` would duplicate `Specified in "dependencies"` output, like:

```
info Reasons this module exists
   - "b#caniuse-api" depends on it
   - "b#caniuse-api#browserslist" depends on it
   - Specified in "dependencies"
   - Specified in "dependencies"
```

This was fixed by making a unique set of dependency patterns in the `why` command code instead of allowing duplicates.

**Test plan**

Added tests for both of the above bugs.


arcanis:
Awesome 👍

PR #5104, opened 2017-12-18
fix broken link in CONTRIBUTING.md

gnujoow:
previous link is linked to 404page.
so I fixed it. 😄


gnujoow:
README.md has same issue in contribution section.
I will make a new pr to fix both.

PR #4981, opened 2017-11-22
fix broken link

seongahjo:
**Summary**

fix broken link for contributing

**Test plan**

PR #5105, opened 2017-12-18
fix(document): broken link to contributing page

gnujoow:
previous link is linked to 404page.
so I fixed it. 😄

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,1 @@
-Please see https://yarnpkg.com/org/contributing/
+Please see https://yarnpkg.com/en/org/contributing/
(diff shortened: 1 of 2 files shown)


gnujoow:
@rally25rs Thanks for reply!
 ive just checked https://github.com/yarnpkg/yarn/pull/4981 which is same pr to this problem.

I think If the page has not been translated yet, it redirected to 404 page. And the page https://yarnpkg.com/org/contributing/  isn't translated in Korean. I will translate it 😄.

but the thing is if the page isn't translated into someone's language, the problem will happen again.
Personally It would be nicer link contribution page to english page like other links in this documentation.
(only link to contribution doesn't have language url)

thanks :)

Fer0x:
I have same problem with `accept-language:ru,en-US;q=0.9,en;q=0.8,uk;q=0.7,pl;q=0.6`.

Maybe there should be redirect with default language if translation is not found? Or there is bug in parsing of  `Accept-Language`?

BYK:
So I think we just need to fix https://github.com/yarnpkg/website/blob/712dae9fe1f54881eb5fca8256f208edae4192ac/_redirects#L48-L60 instead of the link here.

Gonna close this PR out but anyone willing to work on this on the website repo?

PR #5108, opened 2017-12-19
chore(cli): don't ask questions during publish when --non-interactive is specified (#5002)

### LINKED ISSUE #5002: yarn publish --non-interactive still asks questions
yarn 1.3.2
node 9.2.0

# steps to reproduce

```console
$ mkdir hello
$ yarn init
$ yarn publish --non-interactive
```

# actual

yarn stops with:

```
yarn publish v1.3.2
[1/4] Bumping version...
info Current version: 1.0.0
question New version: 
```

# expected

An error message stating that `--non-interactive` must be used with `--new-version` ?

# notes

yarn `--non-interactive` also stops if no npm token is found.

iansu:
**Summary**

Running `publish` with `--non-interactive` should not prompt for input and should fail if required args like `--new-version` are not provided.

Fixes #5002 

**Test plan**

* Run `yarn publish --non-interactive` and see error message about not providing `--new-version`
* Delete NPM token, run `yarn publish --non-interactive --new-version=1.0.1` and see error message about not being able to ask for login due to `--non-interactive` flag

### DIFF
--- a/__tests__/commands/version.js
+++ b/__tests__/commands/version.js
@@ -50,6 +50,24 @@ test('run version with no arguments, --new-version flag where version is same as
   });
 });
 
+test('run version with --non-interactive and --new-version should succeed', (): Promise<void> => {
+  return runRun([], {nonInteractive: true, newVersion}, 'no-args', async (config, reporter): ?Promise<void> => {
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+
+    expect(pkg.version).toEqual(newVersion);
+  });
+});
+
+test('run version with --non-interactive and without --new-version should fail', async (): Promise<void> => {
+  let thrown = false;
+  try {
+    await runRun([], {nonInteractive: true}, 'no-args');
+  } catch (e) {
+    thrown = true;
+  }
+  expect(thrown).toEqual(true);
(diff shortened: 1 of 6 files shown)


gandazgul:
What if instead of requiring --new-version it publishes as is? I'm still using npm publish because I can't use yarn publish from CI.

castarco:
Maybe the branch of this PR should be rebased to avoid having a broken commit. Having broken commits is a problem when someone tries to do a `git bisect` search.

bestander:
Thanks, @iansu, very much appreciated!

bestander:
@castarco, we squash commits when merge, so it should not be a problem

bestander:
@gandazgul, I think there is a separate issue open for your point.
I actually would like to have an ability to bump version and publish as separate commands

rafambbr:
yarn publish --registry=http://mynexus:8081/repository/yarn-repo/ --access public --non-interactive --new-version 0.0.4

janosh:
There should be a non-interactive mode that doesn't require passing in a new version. What if the version was already bumped beforehand?

merceyz:
Create a new issue, please don't bump a three year old PR

PR #5112, opened 2017-12-19
fix(fetching): Moved "link:" logic to LinkFetcher. Fix bug: bin links for "link:" type deps

rally25rs:
**Summary**

Previously there was logic specific to the "link:" type in `PackageFetcher`. This violates
single-responsibity and open-closed principles. This was moved to a new `LinkFetcher` class to align with the other `*Fetcher` classes that house logic specific to dep types.

There was also a bug where no `Manifest` (package.json) was loaded for "link:" deps, which led to bin links not being created, engines ignored, etc. The Manifest is now loaded from the linked dependency if it exists, so these bugs no longer happen.

**Test plan**

Added a test to make sure that .bin links in a "link:" type dep are created now.

### DIFF
--- a/src/fetchers/link-fetcher.js
+++ b/src/fetchers/link-fetcher.js
@@ -1,0 +1,48 @@
+/* @flow */
+
+import type {PackageRemote, FetchedMetadata, Manifest} from '../types.js';
+import type Config from '../config.js';
+import type {RegistryNames} from '../registries/index.js';
+import * as fs from '../util/fs.js';
+
+export default class LinkFetcher {
+  constructor(dest: string, remote: PackageRemote, config: Config) {
+    this.config = config;
+    this.dest = dest;
+    this.reference = remote.reference;
+    this.registry = remote.registry;
+  }
+
+  config: Config;
+  dest: string;
+  registry: RegistryNames;
+  reference: string;
+
+  setupMirrorFromCache(): Promise<?string> {
+    return Promise.resolve();
+  }
+
+  async fetch(): Promise<FetchedMetadata> {
+    const sourceExists = await fs.exists(this.reference);
+    let pkg: Manifest = {_uid: '', name: '', version: '0.0.0'};
+
+    if (sourceExists) {
+      try {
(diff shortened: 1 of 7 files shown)


rally25rs:
Before accepting/merging this, we should see how it jives with the comments in #4341 🤔

src/fetchers/link-fetcher.js
```diff
@@ -0,0 +1,48 @@
+/* @flow */
+
+import type {PackageRemote, FetchedMetadata, Manifest} from '../types.js';
+import type Config from '../config.js';
+import type {RegistryNames} from '../registries/index.js';
+import * as fs from '../util/fs.js';
+
+export default class LinkFetcher {
```
kaylieEB: Can you extend from the BaseFetcher so that you only need to override the fetch method?

rally25rs: Good call! I'll make that change...

src/fetchers/link-fetcher.js
```diff
@@ -0,0 +1,31 @@
...
+export default class LinkFetcher extends BaseFetcher {
+  async fetch(defaultManifest: ?Object): Promise<FetchedMetadata> {
+    const sourceExists = await fs.exists(this.reference);
+    let pkg: Manifest = {_uid: '', name: '', version: '0.0.0'};
+
+    if (sourceExists) {
+      try {
+        pkg = await this.config.readManifest(this.reference, this.registry);
+      } catch (ex) {
+        // pkg will remain the fake Manifest
```
BYK: Can we limit the types of exceptions that we are ignoring here like
```js
catch (err) {
  if (err.code !== 'ENOENT') {
    throw err;
  }
}
```

Also a better explanation regarding why it is okay to ignore this error would be great.

the-spyke:
Any update on this?

anomiex:
We ran into #8290, and looking around I find this patch that seems to do exactly what's needed but it has been stalled on a nitpick since 2018.

Issue #4248, opened 2017-08-24
Support npm config "script-shell"


aikar:
NPM 5.1 added a much longed feature to set the "script-shell" to be used for npm run commands.

This allowed our team to resolve the issue where scripts ran by npm run did not execute in git bash on windows, even though the command was being executed by git bash.

running `npm config set script-shell bash` resolves the issue under NPM, but does not work for yarn.

Please add support for this config so that when yarn has to invoke a command, it respects the config for what shell to use instead of cmd.exe.

aikar:
Any update on this?

arcanis:
Our attention is focused on other issues atm, but please feel free to open a PR and we'll gladly review it :)

aikar:
closed via a2a054438fe91f9860dd06cc4ee7b8004ad7efa0

carsonpowers:
Need this :<

kaleidawave:
Is this present in yarn 2?

`shell-scripts` doesn't seem to exist under the configuration section

arcanis:
No. Yarn 2 implements its own bash-like shell, which is portable across systems and thus doesn't require to set a custom interpreter (using a custom one would make your code non-portable, hence why we don't provide a setting to change it).

kaleidawave:
Ah okay. Having a problems on windows with `yarn 2.4.2`. Running a script via `yarn workspace .. run` with a `cp` command in the script fails saying it is not present...?

arcanis:
`cp` isn't a shell, it's a command. Regardless of the shells, you need to have the underlying binaries present (except for a few binaries implemented as builtins, like `echo`).

kaleidawave:
Yep I know `cp` isn't a shell 😂 I have had a problem in the past with npm and a script with `cp` in. Then I switched npms `shell-scripts` to use powershell (which has cp command as a alias) instead of command prompt and the script worked fine.

I believe this is the same issue. I think my yarn installation is correct so it should have the *underlying binaries* present? Currently the script is works fine when in git-bash so using that as workaround for the moment :)

Issue #4951, opened 2017-11-17
yarn scripts file matches in process.argv doesn't match result in shell


sixinli:
**report a *bug***

**What is the current behavior?**
yarn scripts with file regex doesn't match folders properly
**If the current behavior is a bug, please provide the steps to reproduce.**

repro at: https://github.com/sixinli/testYarn

```
➜  testYarn git:(master) ✗ yarn test
yarn run v1.3.2
$ ./node_modules/test-log/bin/test-log folder1/src/**/*.txt
2 [ 'folder1/src/inner-src/inner-src1.txt',
  'folder1/src/inner-src/inner-src2.txt' ]
✨  Done in 0.18s.

➜  testYarn git:(master) ✗ ./node_modules/test-log/bin/test-log folder1/src/**/*.txt
4 [ 'folder1/src/inner-src/inner-src1.txt',
  'folder1/src/inner-src/inner-src2.txt',
  'folder1/src/src1.txt',
  'folder1/src/src2.txt' ]
```

`test-log` is basically doing `console.log(process.argv.slice(2).length, process.argv.slice(2));`

**What is the expected behavior?**
expected the list of files that matches to be the same

**Please mention your node.js, yarn and operating system version.**
node: v8.9.1
yarn: 1.3.2
MacOS: 10.12.6

rally25rs:
I get the same 2 files any way I run it:

```
~/Projects/testYarn : yarn run test
yarn run v1.3.2
$ ./node_modules/test-log/bin/test-log folder1/src/**/*.txt
2 [ 'folder1/src/inner-src/inner-src1.txt',
  'folder1/src/inner-src/inner-src2.txt' ]
✨  Done in 0.14s.

~/Projects/testYarn : test_packages/test-log/bin/test-log folder1/src/**/*.txt
2 [ 'folder1/src/inner-src/inner-src1.txt',
  'folder1/src/inner-src/inner-src2.txt' ]

~/Projects/testYarn : ./node_modules/test-log/bin/test-log folder1/src/**/*.txt
2 [ 'folder1/src/inner-src/inner-src1.txt',
  'folder1/src/inner-src/inner-src2.txt' ]
~/Projects/testYarn (sli/yarn_file_regex_test) :
```

The 2 files are also what OSX (Sierra 10.12.4)'s `ls` command sees for that glob pattern:

```
~/Projects/testYarn : ls folder1/src/**/*.txt
[...]

PR #5133, opened 2017-12-28
 feat(cli): allow running scripts with custom shell (#4248)

yunyu:
**Summary**

This fixes issue #4248, based off of [what NPM does](https://github.com/npm/npm/blob/07d2296b10e3d8d6f079eba3a61f0258501d7161/lib/utils/lifecycle.js#L264).

**Test plan**

I'm not actually sure how to write a test suite for this (as testing any alternate shell would involve additional installations), but it passes all of the current tests. Any feedback/advice would be appreciated!

Here's how I am manually confirming that the change works (see the `echo $SHELL` output):

image

The Windows terminal emulator (which I am using) doesn't support emojis, which causes the weird characters before the Done message. However, MinTTY should.

### DIFF
--- a/__tests__/commands/run.js
+++ b/__tests__/commands/run.js
@@ -37,6 +37,18 @@ const runRunInWorkspacePackage = function(cwd, ...args): Promise<void> {
     return retVal;
   })(...args);
 };
+const runRunWithCustomShell = function(customShell, ...args): Promise<void> {
+  return buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, config, reporter): Promise<void> => {
+    const yarnRegistry = config.registries.yarn;
+    const originalCustomShell = yarnRegistry.config['script-shell'];
+    yarnRegistry.config['script-shell'] = customShell;
+    const retVal = run(config, reporter, flags, args);
+    retVal.then(() => {
+      yarnRegistry.config['script-shell'] = originalCustomShell;
+    });
+    return retVal;
+  })(...args);
+};
 
 test('lists all available commands with no arguments', (): Promise<void> => {
   return runRun([], {}, 'no-args', (config, reporter): ?Promise<void> => {
@@ -63,7 +75,7 @@ test('lists all available commands with no arguments', (): Promise<void> => {
 test('runs script containing spaces', (): Promise<void> => {
   return runRun(['build'], {}, 'spaces', async (config): ?Promise<void> => {
     const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
-    // The command get's called with a space appended
+    // The command gets called with a space appended
     const args = ['build', config, pkg.scripts.build, config.cwd];
 
     expect(execCommand).toBeCalledWith(...args);
@@ -165,5 +177,15 @@ test('adds workspace root node_modules/.bin to path when in a workspace', (): Pr
 
     expect(envPaths).toContain(path.join(config.cwd, 'node_modules', '.bin'));
     expect(envPaths).toContain(path.join(config.cwd, 'packages', 'pkg1', 'node_modules', '.bin'));
+  });
+});
+
+test('runs script with custom script-shell', (): Promise<void> => {
+  return runRunWithCustomShell('/usr/bin/dummy', ['start'], {}, 'script-shell', async (config): ?Promise<void> => {
+    const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
+    // The command gets called with the provided customShell
+    const args = ['start', config, pkg.scripts.start, config.cwd, '/usr/bin/dummy'];
+
+    expect(execCommand).toBeCalledWith(...args);
   });
 });
--- a/__tests__/fixtures/run/script-shell/package.json
+++ b/__tests__/fixtures/run/script-shell/package.json
@@ -1,0 +1,6 @@
+{
+  "license": "MIT",
+  "scripts": {
+    "start": "echo $SHELL"
+  }
+}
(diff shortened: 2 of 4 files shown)


rally25rs:
I didn't know NPM could do this. It's, IMO, a dangerous idea to change shells. People's package install scripts are tested to work in `sh`, and things like globing patterns differ between shells, especially `bash` vs `zsh`, so install scripts might start failing if someone is using a different shell.

Take a look at #4951 for an example of a glob pattern that changes between shells.

yunyu:
This is really only needed on Windows, where lots of scripts assume *sh (see the original issue report) and use features that aren't available in cmd.exe. I do agree with your rationale, however, NPM uses the same lifecycle execution code for package installation and `npm run-script`. I personally think that as this setting is a bit hidden/out-of-the-way, the people who would be using it tend to know what they are doing, but I'm not opposed to modifying the PR.

yunyu:
@rally25rs I made the requested change and manually tested the behavior again. I'm wondering if  `__tests__\run.js` would the appropriate place to add a test? (Sorry, I'm not particularly familiar with Jest).

rally25rs:
> I'm wondering if __tests__\run.js would the appropriate place to add a test

`__tests__\commands\run.js` contains all the current tests for the `yarn run` command, so that would be the correct place to add a test.

Thanks!

yunyu:
After some wrangling, I managed to add the test. Just for reference, NPM only tests the passed arguments as well: https://github.com/npm/npm/blob/latest/test/tap/run-script.js#L319

aikar:
@rally25rs a main use for this is to override: https://github.com/yarnpkg/yarn/pull/5133/files#diff-939cc0d319704d4f0d98baf9167ba850R191

My team mandates windows users use bash, and even with cross-env, we still had issues with some scripts on windows.

this setting fixed the issues with `npm config set script-shell bash` (not even hard coding path, just use whatever is in ENV). but did not fix it for yarn due to this missing support.

glad to see this finally land, though too late now that my entire team went to linux heh.

BYK:
We may need to revert this change since it essentially reverts #4031 which fixed an issue on Windows. We should have added tests for this but haven't back then so now it came back to bite us.

@yunyu what do you think?

aikar:
@byk why do you think it reverts it? Users have to opt into this, and never binaries still work under git bash just fine?

BYK:
@aikar - this PR changes the execution command for all cases, regardless of them using this new option or not: https://github.com/yarnpkg/yarn/pull/5133/files#diff-939cc0d319704d4f0d98baf9167ba850R218

yunyu:
Sorry, completely wasn't aware of that being an issue earlier. Reverting is fine, I can likely submit a PR that doesn't include this issue later. From what it looks like, { spawn: true } does something slightly different from what NPM/this commit is doing, so that might have something to do with it: https://github.com/nodejs/node/blob/master/lib/child_process.js#L477

Reading the linked issue, I think this it would go away if this were revised with the child_process args logic.

aikar:
@byk but if a shell isn't set, it would fall into the same windows specific check. The desire here is to explicitly opt out of that code. If shell isn't set, it looks like it behaves the same?

yunyu:
@BYK I can't seem to reproduce that issue on my computer:

aikar:
When I say opt out, I meant to opt out of the cmd.exe stuff. Shouldn't it be the user's problem if they set it to a shell that does not handle windows commands correctly?

BYK:
@aikar - Although the code looked identical, using the `shell:true` option tit something differently and fixed the issue. I don't think you and I are talking about the same thing. I'm not talking about anything on the user-side. There literally was a bug that was fixed with using `shell:true` and this patch reverts that fix. If you look at the previous code there, before the `shell:true` fix, it looks very much like the current code, sans the custom shell feature this PR introduced.

@yunyu 
>I think there's a likely chance that fixCmdWinSlashes fixed it.

I think this is the reason why it is not an issue right now since this is what changed in-between that may affect running commands.

>There actually is an issue replaced to sh backslash handling if script-shell is set, however (I didn't encounter this as I didn't know this was a feature):

I don't think to assume a sh-compatible shell is a good idea on Windows. PowerShell is a common shell replacement for instance. All that said, this is why I think this feature is not a great idea. It adds complexity without covering important edge cases. It is a band-aid solution to this problem that will likely create more issues.

The docs for `child_process.exect` says the following about the `shell` option:
>shell <string> Shell to execute the command with. Default: '/bin/sh' on UNIX, process.env.ComSpec on Windows. See Shell Requirements and Default Windows Shell.

And the "Default Windows Shell` section states the following:
>Although Microsoft specifies %COMSPEC% must contain the path to 'cmd.exe' in the root environment, child processes are not always subject to the same requirement. Thus, in child_process functions where a shell can be spawned, 'cmd.exe' is used as a fallback if process.env.ComSpec is unavailable.

So essentially, this whole feature can be replaced by simply defining `$COMSPEC` before running yarn scripts and I think that is a better way to fix this since I'm guessing this feature is not popular amongst macOS or Linux users. What do you think?

aikar:
@BYK Well if somethings still off, seems rather trivial to simply special case 'shell: true' to the previous behavior.

From my understanding, script-shell should be overriding the shell used, and not use comspec if script-shell is set.

So this PR satisfies that situation correctly.

if script-shell isn't set, then shell shouldn't be changed?

https://github.com/yarnpkg/yarn/pull/5133/files#diff-fd065890a349a1c304804a9a9dafe02dR84

yunyu:
@BYK the main problem with setting the shell value in opts is that Node assumes shells on Windows are CMD-compatible, and that shells on *nix are sh compatible: https://github.com/nodejs/node/blob/master/lib/child_process.js#L477

This results in Git Bash/WSL not working, which are the primary use cases for this feature.

aikar:
My team members on Windows used 'npm set script-shell bash', and work specifically out of Git Bash/WSL, so the intent was to stay within the bash ecosystem and not ever jump back through cmd.exe.

Based on @yunyu 's response, looks like everythings working correctly and this PR is fine?

yunyu:
The only problem right now is described in https://github.com/yarnpkg/yarn/pull/5133#issuecomment-355576423, but that doesn't break anything that was previously working without a shell set. For compatibility with NPM and targeting the most common use case,, I personally think that assuming a sh-compatible shell is the best approach, but I'm open to discussion. Should I file a new issue for that?

PR #5138, opened 2017-12-30
style(cli): Use ellipsis instead of three dots

Simek:
**Summary**

Small CLI formatting/cosmetic change. Use ellipsis instead of three dots in console messages.

_Example:_
alt text

### DIFF
--- a/__tests__/reporters/__snapshots__/console-reporter.js.snap
+++ b/__tests__/reporters/__snapshots__/console-reporter.js.snap
@@ -97,7 +97,7 @@ Object {
 exports[`ConsoleReporter.step 1`] = `
 Object {
   "stderr": "",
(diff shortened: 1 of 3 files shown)


KidkArolis:
I kinda personally prefer `...`, it's more .. ASCII?

What's your reasoning for making this change?

Simek:
Reasoning was based on typography and overall presentation aspects.
I'd thought there will be no issue regarding encoding or charset since emoticons are used.

arcanis:
I prefer using the ascii set as much as possible. Emoji and colors help parse the output more easily, an ellipsis not so much (and it can mess with other tools, for example if you're trying to grep it, etc).

Going to close this one if you don't mind, but thanks anyway for the contribution!

Issue #5134, opened 2017-12-28
Missing global dependency generates unhelpful message


BurtHarris:
**Do you want to request a *feature* or report a *bug*?**
Bug - Unhelpful error message

**What is the current behavior?**

Commands like **yarn global list** are failing on my system with the message `error "C:\\code\\antlr4ts\\tool" doesn't exist.`   This is of my own doing (I recognize the missing directory, and had intentionally deleted it) but it wasn't clear how to recover.   More context on what **yarn** was doing would have helped.

I tracked it down to  the file **C:\Users\Burt_\AppData\Local\Yarn\config\global\package.json**, which contained:
```
{
  "dependencies": {
    "antlr4ts-cli": "file:\\code\\antlr4ts\\tool",
    "gulp-cli": "^2.0.0",
    "lerna": "^2.5.1",
    "rimraf": "^2.6.2",
    "typescript": "^2.6.2"
  }
}
```
**Workaround**

Once I knew what package was involved, **yarn global remove antlr4ts-cli** fixed it up.   

**Stack Trace**
```
verbose 0.669 Error: "C:\\code\\antlr4ts\\tool" doesn't exist.
    at MessageError (C:\Program Files (x86)\Yarn\lib\cli.js:139:5)
    at C:\Program Files (x86)\Yarn\lib\cli.js:48701:15
    at next (native)
    at step (C:\Program Files (x86)\Yarn\lib\cli.js:92:30)
    at C:\Program Files (x86)\Yarn\lib\cli.js:103:13
```

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
Best would be to recover automatically from this using behavior like **yarn global remove**

**Please mention your node.js, yarn and operating system version.**
os: Microsoft Windows [Version 10.0.16299.125]
node: v6.12.2
yarn: 1.3.2

rally25rs:
It seems reasonable that on error, we could have a message that says something like "The ___ package appears to be missing. It may have been deleted. You can remove this from Yarn's like of global packages with the command "yarn global remove ___"`

Or someone propose better wording...

iansu:
This seems to be unique to `file://` URLs where the file pointed to is missing. This can happen with local and global packages so that makes the error message a bit trickier. Maybe just something like `<dir> doesn't exist. It may have been deleted. You can remove it with the command "yarn remove <package name>"`?

BurtHarris:
Either message would have made things easier...

PR #5145, opened 2018-01-02
feat(cli): improve error message when file resolver can't find file (#5134)

iansu:
**Summary**

Improve the error message when the file resolver can't find the specified file.

**Test plan**

* Create a `package.json` with a dependency that points to a non-existent file (`file://noexist`)
* Run `yarn`
* Observe that the error message says what the problem is and provides a command to resolve the issue


arcanis:
Thanks!

Issue #4451, opened 2017-09-14
yarn does not honor authentication settings in .npmrc


carlosduclos:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
Yarn does not honor .npmrc
**If the current behavior is a bug, please provide the steps to reproduce.**
We require authentication for one of our repositories and we used to do this by specifying the authentication in .npmrc. This worked up to 0.28.4 but broke in 1.0.0

**What is the expected behavior?**
Honoring the authentication settings in .npmrc

**Please mention your node.js, yarn and operating system version.**
It happens after upgrading yarn to 1.0.0/1.0.1 (have tried both versions). Regardless of OS and nodejs version.

BYK:
This should be fixed in 1.0.2 which was released yesterday. Can you give it a spin?

beardedtim:
@BYK I am experiencing the same issues as above and as mentioned at the end of #4157 . I tried running `1.0.2` and failed with `404`. Running on `0.27.5` works as expected.

KidkArolis:
Could you paste your npmrc file here so we could try to reproduce (obviously edit out secrets or other private info)

beardedtim:
@KidkArolis while trying to get my working dir as clean as possible, it somehow fixed the issue? 🙃 

Steps I took:

* Delete `yarn.lock` and any `yarn-error.log` files
* Delete `node_modules/`
* `brew link yarn` and `yarn --version` to ensure I was at `1.0.2`
* `yarn`

That seems to have worked. I am going to chalk this up as a user error on mine and my team's end.

stieg:
@beardedtim What about if you now blow away `node_modules` but keep `yarn.lock` and then run `yarn install`?  Does it all work then?  This is in part what #4157 tracks but may be related to the larger issue at hand?

kmoe:
I am experiencing this issue (404 on private scoped repo, whichever registry I use) and have tried yarn versions 0.27.5, 1.0.1, and 1.0.2.

I have tried all the workarounds discussed in https://github.com/yarnpkg/yarn/issues/4157. Happy to post logs if it will help.

beardedtim:
@stieg I will try that later today when I get a few moment around lunch and post results. I believe I still got the 404 error but I've tried so many things at this point, who knows!

solkaz:
Has anyone found a solution to this?

perlun:
FWIW: I tried with:

[...]

Issue #4186, opened 2017-08-16
Chore: Add end-to-end tests for private packages


BYK:
We don't have e2e coverage for private packages which sometimes results in bugs going unnoticed until we make a release. We should have at least one smoke test for a very simple private package on NPM and private repo on GitHub or Bitbucket.

onurtemizkan:
I can take this one if it's ok

BYK:
@onurtemizkan go ahead and take it! The main issue here is who will own the test-related private packages? Looks like we can create a private repo under yarnpkg org and probably do the same with scoped packages on NPM but I'd rather hear what @bestander and @Daniel15 has to say about this first. Also may be @cpojer .

bestander:
Yes, I think full e2e with private npm repo is the way to go.
Whether it lives in yarnpkg it does not matter as long as we are fine with
security.

Daniel15:
I thought we already had a task for this?

BYK:
@Daniel15 - I couldn't find it. Feel free to close this or the other as a duplicate if you bump into it.

bestander:
There was one but I did not create it, so can't find it.

avindra:
Relevant breaking issue right now with private registry libs completely failing to install

https://github.com/yarnpkg/yarn/issues/4157

PR #5162, opened 2018-01-06
Fix scoped pkg auth when using npm private pkgs via npm login --scope

### LINKED ISSUE #4157: Yarn writes registry.yarnpkg.com to lockfile when private scoped packages is from registry.npmjs.org
Bug

I have something like the following in my ~/.npmrc file:
```
@mycompany:registry=https://registry.npmjs.org/
//registry.npmjs.org/:_authToken=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
```
This is necessary so I can login to my private repo and get the packages I need to build locally. 

When I add a new package like @mycompany/foo I would expect an entry in the `yarn.lock` file to look something like this:
```
"@mycompany/foo@^1.1.0":
  version "1.1.0"
  resolved "https://registry.npmjs.org/@mycompany/foo/-/foo-1.1.0.tgz#103034ba565c288eb781b1fd1065c9f05dfb90f9"
```
Yeah... not so much.  Instead it looks like this:
```
"@mycompany/foo@^1.1.0":
  version "1.1.0"
  resolved "https://registry.yarnpkg.com/@mycompany/foo/-/foo-1.1.0.tgz#103034ba565c288eb781b1fd1065c9f05dfb90f9"
```
Uncool yarn.  This causes my CI to break because it will try and dig up the packages at `registry.yarnpkg.com`.  Since this is a private registry the login will fail (since the registry is hosted on registry.npmjs.org) and thus it will cause other systems to fail to install packages.

Here is how to reproduce:

1. Setup a private repo on npm with some namespace
2. Set your .npmrc file to something like the one I provided above (except you replace the scope `mycompany` with your own scope, and set a valid auth token of course), and run `yarn add <your scoped package>`.  You can do this by doing `npm login --registry=https://registry.npmjs.org --scope=@<your scope name>`.  I would use yarn login but that doesn't work :(
3. Add a package from your scope using `yarn add`
4. Push a trivial update to that package in your scope.
5. Tell yarn to upgrade the package.

You will see that the URL gets changed in the lock file.

node: 8.2.1
yarn: 0.28.4
OS: Fedora 26

KidkArolis:
Fixes #4157, #4451

**Summary**

This attempts to squash one of the last remaining auth bugs in yarn (🤞). There currently exists an edge case where if you login to the registry using `npm login --scope=@foo`, the contents of .npmrc become 

```
@foo:registry=https://registry.npmjs.org/
//registry.npmjs.org/:_authToken=xxxxxx
```

Installing pkgs from @foo however does not work. That's because the request URL is being made to `https://registry.yarnpkg.com` and `isRequestToRegistry` returns false because the args to it are `isRequestToRegistry('https://registry.yarnpkg.com/@foo/pkg', 'https://registry.npmjs.org')`.

This PR attempts to remove `isRequestToRegistry` entirely, because I believe at this stage it is no longer needed. The only purpose of this function is to make sure that auth tokens for registry X are only sent with requests to registry X. This is now ensured by `getAuth` function which looks at the request url or package identifier and decides what token to use.

Before this PR, `getAuth('http://Y)` would sometimes return the auth token for `registry.npmjs.org` since that's the default registry. Instead, it now returns no token since the request url does not match the registry.

Let me know if I can clarify something further.

**Test plan**

I've tested this manually using my own npm private account. It would be great to get that e2e test going as discussed here: https://github.com/yarnpkg/yarn/issues/4186. I could try working on that next if there's interest, but I'd do that in a separate PR.

I've also added a new unit test for this edge case `npm login --scope=@foo`. In other words, all existing tests pass (with a small exception - see `should add authorization header with token for custom registries with a scoped package` test) + I've added a new test to cover for the edge case in question.

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -286,6 +286,18 @@ describe('request', () => {
     const requestParams = createRegistry(config).request(url);
     expect(requestParams.headers.authorization).toBe('Bearer testScopedAuthToken');
   });
+
+  test('should add authorization header if pathname does not match but custom-host-suffix is used', () => {
+    const url = 'https://some.other.registry/tarball/path/@testScope%2fyarn.tgz';
+    const config = {
+      '//some.other.registry/some/path/:_authToken': 'testScopedAuthToken',
+      '@testScope:registry': 'https://some.other.registry/some/path/',
+      'custom-host-suffix': 'some.other.registry',
+    };
+
+    const requestParams = createRegistry(config).request(url);
+    expect(requestParams.headers.authorization).toBe('Bearer testScopedAuthToken');
+  });
 });
 
 const packageIdents = [
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -197,7 +197,7 @@ describe('request', () => {
 
     npmRegistry.config = {
       '//some.other.registry/:_authToken': 'testScopedAuthToken',
-      '@testScope:registry': '//some.other.registry/',
+      '@testScope:registry': 'https://some.other.registry/',
     };
     npmRegistry.request(url);
 
@@ -205,61 +205,23 @@ describe('request', () => {
 
     expect(requestParams.headers.authorization).toBe('Bearer testScopedAuthToken');
   });
-});
 
-describe('isRequestToRegistry functional test', () => {
-  test('request to registry url matching', () => {
+  test('should add authorization header with token for default registry when using npm login --scope=@foo', () => {
     const testCwd = '.';
     const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
     const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
 
-    const validRegistryUrls = [
-      ['http://foo.bar:80/foo/bar/baz', 'http://foo.bar/foo/'],
-      ['http://foo.bar:80/foo/bar/baz', 'https://foo.bar/foo/'],
-      ['http://foo.bar/foo/bar/baz', 'http://foo.bar/foo/'],
-      ['http://foo.bar/foo/00000000-1111-4444-8888-000000000000/baz', 'http://foo.bar/foo/'],
-      ['https://foo.bar:443/foo/bar/baz', 'https://foo.bar/foo/'],
-      ['http://foo.bar/foo/bar/baz', 'https://foo.bar:443/foo/'],
-      ['https://foo.bar/foo/bar/baz', 'https://foo.bar:443/foo/'],
-      ['HTTP://xn--xample-hva.com:80/foo/bar/baz', 'http://êxample.com/foo/bar/baz'],
-    ];
-
-    const invalidRegistryUrls = [
-      ['https://wrong.thing/foo/bar/baz', 'https://foo.bar/foo/'],
-      ['https://foo.bar:1337/foo/bar/baz', 'https://foo.bar/foo/'],
-    ];
-
-    validRegistryUrls.forEach(([requestUrl, registryUrl]) =>
-      expect(npmRegistry.isRequestToRegistry(requestUrl, registryUrl)).toBe(true),
-    );
-    invalidRegistryUrls.forEach(([requestUrl, registryUrl]) =>
-      expect(npmRegistry.isRequestToRegistry(requestUrl, registryUrl)).toBe(false),
-    );
-  });
-
-  test('isRequestToRegistry with custom host prefix', () => {
-    const testCwd = '.';
-    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
-    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+    const url = 'https://npmjs.registry.org/@foo%2fyarn.tgz';
 
     npmRegistry.config = {
-      'custom-host-suffix': 'some.host.org',
+      '//npmjs.registry.org/:_authToken': 'testScopedAuthToken',
+      '@foo:registry': 'https://npmjs.registry.org/',
     };
+    npmRegistry.request(url);
 
-    expect(npmRegistry.isRequestToRegistry('http://pkgs.host.com:80/foo/bar/baz', 'http://pkgs.host.com/bar/baz')).toBe(
-      false,
-    );
+    const requestParams = mockRequestManager.request.mock.calls[0][0];
 
-    npmRegistry.config = {
-      'custom-host-suffix': 'pkgs.host.com',
-    };
-
-    expect(npmRegistry.isRequestToRegistry('http://pkgs.host.com:80/foo/bar/baz', 'http://pkgs.host.com/bar/baz')).toBe(
-      true,
-    );
-    expect(npmRegistry.isRequestToRegistry('http://pkgs.host.com:80/foo/bar/baz', '//pkgs.host.com/bar/baz')).toBe(
-      true,
-    );
+    expect(requestParams.headers.authorization).toBe('Bearer testScopedAuthToken');
   });
 });
 
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -72,6 +72,22 @@ function createMocks(): Object {
 }
 
 describe('request', () => {
+  // a helper function for creating an instance of npm registry,
+  // making requests and inspecting request parameters
+  function createRegistry(config: Object): Object {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter);
+    npmRegistry.config = config;
+    return {
+      request(url: string): Object {
+        npmRegistry.request(url);
+        const requestParams = mockRequestManager.request.mock.calls[0][0];
+        return requestParams;
+      },
+    };
+  }
+
   test('should call requestManager.request with url', () => {
     const testCwd = '.';
     const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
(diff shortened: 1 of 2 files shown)


__tests__/registries/npm-registry.js
```diff
@@ -197,69 +197,31 @@ describe('request', () => {
 
     npmRegistry.config = {
       '//some.other.registry/:_authToken': 'testScopedAuthToken',
-      '@testScope:registry': '//some.other.registry/',
+      '@testScope:registry': 'https://some.other.registry/',
```
KidkArolis: By default, `npm login` populates the `.npmrc` file with the full url, e.g. 

```
@scope:registry=https://some.other.registry/
```

and not

```
@scope:registry=//some.other.registry/
```

I made sure this is not a regression by editing away the `https://` in my `~/.npmrc` and running `npm install` which gives me `npm ERR! Only absolute URLs are supported`

__tests__/registries/npm-registry.js
```diff
@@ -197,69 +197,31 @@ describe('request', () => {
...
+      '@foo:registry': 'https://npmjs.registry.org/',
     };
+    npmRegistry.request(url);
 
-    expect(npmRegistry.isRequestToRegistry('http://pkgs.host.com:80/foo/bar/baz', 'http://pkgs.host.com/bar/baz')).toBe(
-      false,
-    );
-
-    npmRegistry.config = {
-      'custom-host-suffix': 'pkgs.host.com',
```
KidkArolis: The `custom-host-suffix` feature is no longer necessary. If the request is being made to a registry that needs a token - the token will get sent. If the requested registry does no have a token, no token is sent.

Though, looking at this test a bit more now, I might have misunderstood. If someone is using a registry with a full path, e.g. `http://pkgs.host.com/bar/baz` instead of just `http://pkgs.host.com/`, removal of this "custom-host-suffix" feature might indeed introduce a regression.

But why not just specify the registry as `http://pkgs.host.com/`? You basically trust that server.

src/registries/npm-registry.js
```diff
@@ -133,10 +113,8 @@ export default class NpmRegistry extends Registry {
       opts.headers,
     );
 
-    const isToRegistry = this.isRequestToRegistry(requestUrl, registry);
-
     // this.token must be checked to account for publish requests on non-scopped packages
-    if (this.token || (isToRegistry && (alwaysAuth || this.isScopedPackage(packageIdent)))) {
+    if (this.token || alwaysAuth || this.isScopedPackage(packageIdent)) {
```
KidkArolis: The key change is basically:

We stop checking if request isToRegistry (which is a strange and slightly complicated thing to understand, of course request is being made to a registry, what else would it be made to?)

And instead we make use of `this.getAuth` to figure out if this specific request needs a token or not.

In fact, we could push these 3 conditions (`this.token`, `alwaysAuth`, `scopedPackage`) to `getAuth` making this method much cleaner. `getAuth` takes a url and figures out if authorization is required or not. Should I do that?

src/registries/npm-registry.js
```diff
@@ -271,6 +249,7 @@ export default class NpmRegistry extends Registry {
       if (registry) {
         return String(registry);
       }
+      return '';
```
KidkArolis: This is to make sure that if URL is with a protocol, e.g. `https://github.com`, we don't "assume" this is `registry.npmjs.org`. This change is making `getRegistry` stricter. Since this method is used only for authentication and scope resolving purposes, we basically make sure to be strict and only return `registry.npmjs.org` if request is really to that registry.

src/registries/npm-registry.js
```diff
@@ -290,6 +269,11 @@ export default class NpmRegistry extends Registry {
     }
 
     const baseRegistry = this.getRegistry(packageIdent);
+
+    if (!baseRegistry) {
+      return '';
+    }
```
KidkArolis: Now that we more strictly check what registry the request is for (e.g. don't conclude a request to `https://github.com` is to known registry), we can conclude that non registry requests don't need auth. This is basically the replacement for the `isRequestToRegistry`.

src/registries/npm-registry.js
```diff
@@ -346,4 +330,15 @@ export default class NpmRegistry extends Registry {
...
+  getAvailableRegistries(): Array<string> {
+    const availableRegistries = super.getAvailableRegistries();
+    if (availableRegistries.indexOf(YARN_REGISTRY) === -1) {
+      availableRegistries.push(YARN_REGISTRY);
+    }
+    if (availableRegistries.indexOf(DEFAULT_REGISTRY) === -1) {
+      availableRegistries.push(DEFAULT_REGISTRY);
+    }
+    return availableRegistries;
+  }
```
KidkArolis: `DEFAULT_REGISTRY` and `YARN_REGISTRY` are both available registries even if not mentioned in any of the config. Make sure this is reflected as we now rely on this when extracting registry from a request url.

In theory only `DEFAULT_REGISTRY` should suffice, but I wanted to make sure that if someone sets their default registry for some reason to be the yarn registry, things should work the same.

__tests__/registries/npm-registry.js
bestander: History shows @soda0289 adding it in https://github.com/yarnpkg/yarn/pull/3987.
Do you think the original issue is still fixed?

bestander:
@KidkArolis thank a lot for a PR and your annotations!
I'd leave this for @BYK @arcanis to review it.

Pleas ping here if review gets stuck at any point

TheLudd:
I installed the build version and after that the error in https://github.com/yarnpkg/yarn/issues/4451#issuecomment-355248563 was no longer reproducible. I.e. the bug seems to be solved for me. :+1:

__tests__/registries/npm-registry.js
KidkArolis: The `custom-host-suffix` feature is the only point in this PR that I'm unsure about.

This PR preserves the feature where the authorization token is only sent if it's applicable to that registry (which was one of the intentions of #3987). But I don't look at pathnames.

Fix #3987 made sure that pathname and not just host is also checked when deciding if auth header should be used. But I'm not convinced that's needed. The reasoning is briefly described here: https://github.com/yarnpkg/website/pull/504/files by @lumaxis, but I need to understand that use case better. Are you around to discuss this @lumaxis? In particular how can I setup this VSTS registry to test yarn?

I will dive deeper on this will get back here with my conclusions.

KidkArolis:
Currently, multiple functions are used to decide if a given request URL/pkg requires authorization. This leads to subtle issues like the on #4451.

The key point of this PR is to _simplify_ the authorization decision by making sure that `getAuth(packageIdent)` alone decides whether auth should be used for a particular pkg/url.

The only blocker to this PR currently is figuring out whether `pathname` matters when comparing a URL against a configured registry. E.g. if someone's registry in `.npmrc` is `@foo:registry=https://myregistry.com/with/path`, but requests go to `https://myregistry.com/some/hash/pkg.tar.gz` - the logic in this PR will not send the authorization token. In master branch, however, a token would be sent if `custom-host-suffix` is provided in the config file.

__tests__/registries/npm-registry.js
lumaxis: Hey @KidkArolis, thanks for the ping!
Unfortunately, I haven't really been keeping up with the `custom-host-suffix` feature because for unrelated reasons, we're still using NPM in the projects I'm currently working on.
Looking at one of our internal VSTS package feeds though, it appears that they "fixed" my original issue by simply adding both URLs with the slightly different paths to the snippet they have you copy to your `~/.npmrc`:

```
//team-name.pkgs.visualstudio.com/_packaging/ProjectName/npm/registry/:_authToken=$TOKEN
//team-name.pkgs.visualstudio.com/_packaging/ProjectName/npm/:_authToken=$TOKEN
```

My work in #3231 was really only picking up an existing solution proposed in #2507. Overall, I think my specific problem with VSTS would still be solved if the `custom-host-suffix` feature were removed. I remember however that other people who were using other private package feeds were looking forward to the feature too. I don't know if they actually ended up using it though.

Does this help you at all?

KidkArolis: Thanks for the info!

My conclusion is then that one could still use VSTS if the `~/.npmrc` looked something like:

```
registry=https://team-name.pkgs.visualstudio.com/
//team-name.pkgs.visualstudio.com/:_authToken=$TOKEN
```

or if using scopes

```
@foo:registry=https://team-name.pkgs.visualstudio.com/
//team-name.pkgs.visualstudio.com/:_authToken=$TOKEN
```

With config like that, authorization headers would be sent correctly. So it's good to know this would work for that VSTS case.

Next, I'll look into how exactly `npm` handles the case when registry config contains a path, I found the relevant bit of code: https://github.com/npm/npm/blob/5e426a78ca02d0044f8dd26e0c5f881217081cbd/lib/config/fetch-opts.js#L46-L67. But haven't had a chance to look mode deeply yet.

KidkArolis:
Wow, this is quite deep. Took me a couple hours to write this comment during which I realized twice some issues with this PR. But I think I've figured some things out, just need a break ;).

### tl;dr

**This PR is not yet ready to be rolled out. I realized that removing `isPackageToRegistry` is not good, because `request()` takes both `packageName` and `requestUrl`. Both of those need to be properly checked when considering authorization and currently I only look at `packageName`.**

In the meantime, I've split all the changes into 4 independently reviewable commits if you fancy a look.

### The longer version

Thing is, this `custom-host-suffix` feature was a little odd and I was considering removing it. It all has to do with one edge case where a package registry includes a path **and** tarballs are at a different path.

An example of a registry without any path:

- `https://registry.npmjs.org/` - root
- `https://registry.npmjs.org/@pkg/name` - meta
- `https://registry.npmjs.org/@pkg/name/tarball.tar.gz` - tarball

An example of a registry with a path (e.g. Artifactory, Nexus, VSTS):

- `https://artifactory.example.com/artifactory/api/npm/registry/` - root
- `https://artifactory.example.com/artifactory/api/npm/registry/@pkg/name` - meta
- `https://artifactory.example.com/artifactory/api/packages/@pkg/name.tar.gz` - tarball

The issue that `custom-host-suffix` is trying to solve is to say, even though the registry root is `https://artifactory.example.com/artifactory/api/npm/registry/`, let's configure `custom-host-suffix` as `artifactory.example.com`. This tells yarn to ignore the path difference in `isRequestToRegistry` check. This then proceeds to `getAuth` which uses `pkgIdent` to determine what auth token to use. 

I was gonna explain why I thought `custom-host-suffix` is no longer needed, but been typing this comment for too long.

### Conclusion

In conclusion, I'll revisit this PR to make sure:

1. The bug it's trying to fix is fixed.
2. Yarn is still secure when it comes to installing pkgs from authed meta with unauthed tarballs (e.g. private npm registry package, but tarball on public cdn).
3. Registries with paths (like artifactory, nexus, vsts) are supported with `custom-host-suffix` for backwards compatibility.
4. Those same registries are supported without needing `custom-host-suffix`, but instead using some standard existing config. I'm thinking something like setting `@foo:registry=artifactory/npm` and also `@foo-artefacts:registry=artifactory/npm`. We could then deprecate `custom-host-suffix` if we wanted to. But not sure anyone cares about this much detail at this point.

Oh how I wish we could simplify this code further, it seems quite brittle and has been worked on a lot over the years. I think the main reason this might be so complicated is because request can be called in all of the following patterns:

```js
request('@scope/pkg', {})
request('@scope/pkg', {}, '@scope/pkg')
request('https://registry/@scope/pkg.tar.gz', {})
request('https://registry/@scope/pkg.tar.gz', {}, '@scope/pkg')
```

But I think they're all needed:

1. Making requests to outside of registry (e.g. tarballs)
2. Making requests to registry in context of a pkg (e.g. meta)
3. Making requests to registry out of pkg context (e.g. non pkg related registry APIs)

Anyway, something to think about..

### What about npm?

Haven't looked deeply at their auth code yet. But looks like folks at npm are also dealing with similar issues. Here is an issue with recent comments about artifactory installs not working: https://github.com/npm/npm/issues/16528.

awkaiser:
@KidkArolis thank you so much for your attention to detail on this issue 🍻

lumaxis:
Amazing work @KidkArolis ! Your comment is a perfect summary of the problem and will be a tremendous help for anyone trying to understand why we added `custom-host-suffix` in the first place 🙏

KidkArolis:
I've opened a new PR with a much smaller change to the code (and bigger change to the tests) - https://github.com/yarnpkg/yarn/pull/5216.

Closing this for now as it changed the code too much in a direction that wasn't fully correct.

PR #5213, opened 2018-01-13
feat: better error when package is not found

GAumala:
**Summary**

When a package is not found during installation, show which package is requiring it (if any) in the
error message. If there is no parent request, throw the same error as before.

Add new localized string 'requiredPackageNotFoundRegistry' which is the same 'packageNotFoundRegistry', but includes the parent package pattern. 
Add isMessageError boolean value to MessageError class to distinguish
between errors thrown by yarn from errors thrown by yarn dependencies.

**Test plan**

Use a workspace with at least 2 packages `foo` and `bar`. `bar` depends on `foo`. If you manually bump the package version of `foo`, running the install command will fail because the npm registry doesn't have the latest `foo` version that was just created. 

**Error shown (Before)**

screenshot from 2018-01-13 17-42-27

**Error shown (After)**

screenshot from 2018-01-13 17-42-46

To implement this I decided to put a try catch statement in `PackageRequest`'s `findVersionOnRegistry` method. I think the error is actually thrown by the `request` module but I decided to catch the error here because I still have access to the reference to the parent request. I didn't want to hide any `MessageError` errors that may be thrown in the underlying function so I added a `isMessageError`  attribute to that class so that I can identify such errors. 

Please let me know if there is a better way to implement this feature.

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -111,7 +111,18 @@ export default class PackageRequest {
 
     const Resolver = this.getRegistryResolver();
     const resolver = new Resolver(this, name, range);
-    return resolver.resolve();
+    try {
+      return await resolver.resolve();
+    } catch (err) {
+      // if it is not an error thrown by yarn and it has a parent request,
+      // thow a more readable error
+      if (!(err instanceof MessageError) && this.parentRequest && this.parentRequest.pattern) {
+        throw new MessageError(
+          this.reporter.lang('requiredPackageNotFoundRegistry', pattern, this.parentRequest.pattern, this.registry),
+        );
+      }
+      throw err;
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/errors.js
```diff
@@ -4,9 +4,11 @@ export class MessageError extends Error {
   constructor(msg: string, code?: string) {
     super(msg);
     this.code = code;
+    this.isMessageError = true;
```
arcanis: Can we use an `instanceof` check instead?

GAumala: Yeah I was thinking about using `instanceof`, but what if subclasses of `MessageError` are thrown? Should I check for all of them or only `MessageError`?

arcanis: Only MessageError is fine. If they're subclass, the instanceof will work as expected and will return true, even if they are from a derived class.

arcanis:
Thanks!

Issue #5206, opened 2018-01-10
`yarn add --cwd invalid-dir` should warn/fail


thatmarvin:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
`yarn add --cwd invalid-dir bluebird` installs bluebird to the root:

```
yarn add v1.3.2
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
success Saved 1 new dependency.
└─ bluebird@3.5.1
✨  Done in 0.30s.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Just `yarn add --cwd invalid-dir bluebird` (and `invalid-dir` not exist, of course)

**What is the expected behavior?**
Throw an error, or at least show a warning.

For reference, on npm 5.6.0, `npm --prefix invalid-dir bluebird` creates `invalid-dir` and installs it there (which I also think is unexpected behavior), but at least it throws a warning:

```
$ npm i --prefix invalid-dir bluebird
npm WARN saveError ENOENT: no such file or directory, open '/Users/marvin/devel/yarn-add-cwd-demo/invalid-dir/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/Users/marvin/devel/yarn-add-cwd-demo/invalid-dir/package.json'
npm WARN invalid-dir No description
npm WARN invalid-dir No repository field.
npm WARN invalid-dir No README data
npm WARN invalid-dir No license field.

+ bluebird@3.5.1
added 1 package in 0.734s
```

**Please mention your node.js, yarn and operating system version.**
node v8.9.4, yarn 1.3.4, macOS High Sierra 10.13.2

maciej-ka:
Hi, I'm on it.

witalewski:
Is this issue solved? If not, I'd like to tackle this

arcanis:
It should have been fixed in #5222 (1.5+). Still having this issue?

witalewski:
I'm still having it with yarn 1.13 if I the path passed to --cwd doesn't begin with '/':

```yarn on  master [!?] is 📦 v1.13.0-0 via ⬢ v8.11.4 took 2s
➜ node $PWD/lib/cli/index.js add --cwd /nonexistent bluebird
yarn add v1.13.0-0
[...]

PR #5222, opened 2018-01-15
feat(cli): error on missing workspace directory (#5206)

maciej-ka:
Add an error when the working directory doesn't exist (most probably when `--cwd` option points to the wrong directory)

As a side effect, this PR detects typo in one of the tests. Path "non-packages" should be "non-package". That test was passing because the yarn was creating target directory.

**Test plan**

`yarn add bluebird --cwd /wrong-path`
current behavior: creates `/wrong-path` directory
expected: Error message


arcanis:
Thanks! :+1:

PR #5226, opened 2018-01-15
feat(cli): sort by version when asking user to select a package version (#5220)

### LINKED ISSUE #5220: Incorrect order of package versions

🐞 

If I specify incorrect package version inside `package.json` and run `yarn` I'm presented with a following:
image
Order of versions doesn't make any sense and it's very easy to press enter thinking you select the latest version and then spend a lot of time debugging.

1. Add package with an incorrect version to `package.json`. `@types/react` in my case.
2. Run yarn

Versions sorted according to semantic version spec.

```
$ yarn --version
1.3.2
$ node --version
v9.3.0
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1114
```

This bug is pretty easy to fix it's enough to change this line:
https://github.com/yarnpkg/yarn/blob/dc7057688fd0e8a60653b99bf7a78939a35d7fb4/src/resolvers/registries/npm-resolver.js#L70

Instead of `Array. reverse` it should use [rsort](https://github.com/npm/node-semver/blame/master/semver.js#L586) from `semver` which is [already required](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/registries/npm-resolver.js#L17) in this file.

iansu:
**Summary**

When prompting the user to select a package version to install sort the options by version number.

Closes #5220

**Test plan**

* Create a package and add `"@types/react": "15.4.0"` (or another non-existent package version) to `package.json`
* Run `yarn` and when prompted to select a package version observe that the list is sorted by version (according to semver rules) in descending order

### DIFF
--- a/src/resolvers/registries/npm-resolver.js
+++ b/src/resolvers/registries/npm-resolver.js
@@ -67,7 +67,7 @@ export default class NpmResolver extends RegistryResolver {
           name: 'package',
           type: 'list',
           message: config.reporter.lang('chooseVersionFromList', body.name),
-          choices: Object.keys(body.versions).reverse(),
+          choices: Object.keys(body.versions).sort(semver.rcompare),
           pageSize,
         },
       ]);
--- a/src/resolvers/registries/npm-resolver.js
+++ b/src/resolvers/registries/npm-resolver.js
@@ -67,7 +67,7 @@ export default class NpmResolver extends RegistryResolver {
           name: 'package',
           type: 'list',
           message: config.reporter.lang('chooseVersionFromList', body.name),
-          choices: Object.keys(body.versions).sort(semver.rcompare),
+          choices: (semver: Object).rsort(Object.keys(body.versions)),
           pageSize,
         },
       ]);


iansu:
Looks like there are some flow errors that weren't showing up in my editor for some reason. The code still works but I'll fix the errors.

src/resolvers/registries/npm-resolver.js
```diff
@@ -67,7 +67,7 @@ export default class NpmResolver extends RegistryResolver {
           name: 'package',
           type: 'list',
           message: config.reporter.lang('chooseVersionFromList', body.name),
-          choices: Object.keys(body.versions).reverse(),
+          choices: (semver: Object).rsort(Object.keys(body.versions)),
```
iansu: I have no idea why this `(semver: Object)` syntax is needed but without it I get a flow error about property `rsort` not being found in module `semver`. It's weird because on line 49 we call `semver.satisfies` with no issues. 🤷‍♂️

TLadd: The Object annotation is necessary because the flow-typed definition being used for the semver package is old. It [contains a definition for satisfies](https://github.com/yarnpkg/yarn/blob/master/flow-typed/npm/semver_v5.1.x.js#L63), but does not have one for rsort. 

The [latest definition](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/semver_v5.1.x/flow_v0.25.x-/semver_v5.1.x.js#L144) does contain rsort, so you could probably get rid of the Object annotation by updating.

iansu: That makes sense. I was actually looking for the type definitions but I looked in the semver package itself. I didn't realize all these flow definitions were stored externally.

iansu: I tried updating the flow definitions for semver and it broke a lot of stuff. Leaving this as is for now.

PR #5227, opened 2018-01-16
feat(upgrade, add): Separately log added/upgraded dependencies

kaylieEB:
Implementation of feature https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-show-updated-packages-only.md

When adding or upgrading dependencies, group them into directly required ones vs.
transitive ones so that version shown is clearer to the end user.

https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-show-updated-packages-only.md

**Summary**

screen shot 2018-01-15 at 4 12 30 pm

**Test plan**

Added a test in upgrade.js


__tests__/index.js
```diff
@@ -61,19 +61,19 @@ async function execCommand(
...
+  expect(lastLines[6]).toContain('Done');
 }
 
 function expectAddSuccessfullOutputWithNoLockFile(stdout, pkg) {
-  const lastLines = stdout.slice(stdout.length - 4);
+  const lastLines = stdout.slice(stdout.length - 7);
   expect(lastLines[0]).not.toEqual('success Saved lockfile.');
   expect(lastLines[1]).toEqual('success Saved 1 new dependency.');
-  expect(lastLines[2]).toContain(pkg);
-  expect(lastLines[3]).toContain('Done');
```
rally25rs: Wondering if converting this test to a "snapshot" would be easier than trying to maintain this "index of line of output" approach...

src/cli/commands/add.js
```diff
@@ -155,9 +155,27 @@ export class Add extends Install {
...
+
+    const resolverPatterns = new Set();
+    for (const pattern of patterns) {
+      const {version, name} = this.resolver.getResolvedPattern(pattern) || {};
+      resolverPatterns.add(`${name}@${version}`);
+    }
+    const directRequireDependencies = trees.filter(({name}) => resolverPatterns.has(name));
+
+    this.reporter.info(this.reporter.lang('directDependencies'));
+    this.reporter.tree('newDependencies', directRequireDependencies);
```
rally25rs: I think we might want different keys instead of reusing `'newDependencies'` for the 2 calls to `.tree(`.  It doesn't matter for the console-reporter, but if you `--json` to use the json-reporter, this key could be used to differentiate the two trees if parsing the json output. Maybe change this one to `'newDirectDependencies'`.

kaylieEB: ah, that's what it was used for! Will fix :)

__tests__/index.js
kaylieEB: So I had to make the tests not run concurrently (otherwise the snapshots would match random stdout depending on the order the commands ran) and still slice the output (remove yarn version and the time it took to complete the command), but I think the change is still worthwhile for developer experience.

albybarber:
34966402-3d81da90-fa10-11e7-955e-3dddcfeb0519_png__640x1768_

Thoughts?

PR #5255, opened 2018-01-19
Add npm metadata to installed packages (#5240)

### LINKED ISSUE #5240: "npm install" deletes everything from a Yarn-produced tree

Hard to say. An integration issue where neither of the parties is directly responsible but the result is confusing to everyone.

Running `npm install <something>` over a Yarn-produced tree corrupts it, removing everything Yarn has previously installed.

```
npm --version # 5.6.0
yarn --version # 1.4.0

yarn init -y
yarn add react-scripts

# ./node_modules/.bin/react-scripts exists

npm install --save react

# ./node_modules/.bin/react-scripts is deleted with all other package content
```

I’d expect that this either works, or somehow fails gracefully. I don’t have a specific proposal. To Create React App users, this is extremely confusing because their project gets created with Yarn (if Yarn exists on the system) but then they follow the docs of some library that tells them to `npm install`, and that blows away the whole tree.

I reached out to @zkat about this out of band, and the response is that npm verifies integrity using metadata in fetched `package.json`. If Yarn wrote that metadata by default, perhaps npm could work with Yarn trees without this issue.

I want to stress I don’t blame either party for this. I understand both have reasons for this behavior. But is highly confusing to users of both packagers, and IMO hurts the overall ecosystem. I wish we could find a way to solve or work around this; even if it’s npm refusing to run `install` when `yarn.lock` exists, or Yarn writing additional metadata that npm requires.

I’m filing this in the Yarn repository because the npm repository is too flooded with issues, and the maintainers have indicated it is unlikely to get a response there.

macOS Sierra
Node 8.6.0
Yarn 1.4.0 (and earlier versions)

iansu:
**Summary**

**Work in Progress**

Running `npm install` in a yarn-based project results in npm deleting all packages installed by yarn. It sounds like we can prevent this by adding a couple of pieces of metadata to the `package.json` of installed packages.

Closes #5240 

**Test plan**

1. Run `yarn add react-scripts`
2. Run `npm install react`
3. npm should not delete (almost) everything in `node_modules`

### DIFF
--- a/src/package-fetcher.js
+++ b/src/package-fetcher.js
@@ -24,6 +24,11 @@ async function fetchOne(ref: PackageReference, config: Config): Promise<FetchedM
   const dest = config.generateHardModulePath(ref);
 
   const remote = ref.remote;
+
+  if (ref.patterns && ref.patterns[0]) {
+    remote.pattern = ref.patterns[0];
+  }
+
   // Mock metedata for symlinked dependencies
   if (remote.type === 'link') {
     const mockPkg: Manifest = {_uid: '', name: '', version: '0.0.0'};
--- a/src/fetchers/base-fetcher.js
+++ b/src/fetchers/base-fetcher.js
@@ -77,6 +77,20 @@ export default class BaseFetcher {
         ),
       );
 
+      await fs.writeFile(
+        path.join(this.dest, constants.NODE_PACKAGE_JSON),
+        JSON.stringify(
+          {
+            ...pkg,
+            _from: this.remote.pattern,
+            _resolved: this.remote.reference,
+            _shasum: hash,
+          },
+          null,
+          '  ',
+        ),
+      );
+
       return {
         hash,
         dest: this.dest,
--- a/src/types.js
+++ b/src/types.js
@@ -41,6 +41,7 @@ export type PackageRemote = {
   resolved?: ?string,
   hash: ?string,
   packageName?: string,
+  pattern?: string,
 };
 
 // `dependencies` field in package info


iansu:
This is very much work in progress. There are still some outstanding issues but I wanted to start getting feedback from everyone.

Issues:
1. With git and tarball dependencies the package name is prepended to `_from`. For example, `_from: aproba@github:iarna/aproba#v1.2.0`. This doesn't happen with registry dependencies. If this is a problem it should be fairly easy to strip the package name.
2. When installing a tarball dependency that comes from the yarn cache `_resolved` points to the cache, not the registry
3. This doesn't actually seem to work yet. This could be because of point 1 above, it could be because yarn points to its own registry (`https://registry.yarnpkg.com/react-scripts/-/react-scripts-1.1.0.tgz`) instead of npm or it could be something else.

src/package-fetcher.js
```diff
@@ -24,6 +24,11 @@ async function fetchOne(ref: PackageReference, config: Config): Promise<FetchedM
   const dest = config.generateHardModulePath(ref);
 
   const remote = ref.remote;
+
+  if (ref.patterns && ref.patterns[0]) {
+    remote.pattern = ref.patterns[0];
```
iansu: I don't know if this is an appropriate place to add this information but `remote` is one of the only things passed to the fetcher. I initially had it in `defaultManifest` but that didn't seem right either.

ifeltsweet:
Any updates on this?

merceyz:
Closing as npm 7 stopped mutating `package.json` and thus no longer adds these fields https://github.com/npm/rfcs/blob/latest/implemented/0013-no-package-json-_fields.md

PR #5293, opened 2018-01-30
feat(install): Implement a very basic hook system

arcanis:
**Disclaimer:** This is NOT a public feature, we reserve the right to change it as we wish, including between patch releases.

---

This PR adds a very basic and undocumented hook system. I plan to use it internally to get better
stats on how Yarn performs, and how much time is spent on the linking step.

**Test plan**

I tested this PR by adding the following code to the beginning of a custom build:

```js
let trace = name => {
    return async fn => {
        let x = Date.now();
        let r = await fn();
        console.log(`step ${name}: ${Date.now() - x}`);
        return r;
    };
};

global.experimentalYarnHooks = {

    resolveStep: trace('resolve'),
    fetchStep: trace('fetch'),
    linkStep: trace('link'),
    buildStep: trace('build'),

};
```

More comprehensive tests are on their way.

### DIFF
--- a/src/util/hooks.js
+++ b/src/util/hooks.js
@@ -1,0 +1,21 @@
+/* @flow */
+
+const YARN_HOOKS_KEY = 'experimentalYarnHooks';
+
+export function callThroughHook<T>(type: string, fn: () => T): T {
+  if (typeof global === 'undefined') {
+    return fn();
+  }
+
+  if (typeof global[YARN_HOOKS_KEY] !== 'object' || !global[YARN_HOOKS_KEY]) {
+    return fn();
+  }
+
+  const hook: (() => T) => T = global[YARN_HOOKS_KEY][type];
+
+  if (!hook) {
+    return fn();
+  }
+
+  return hook(fn);
+}
--- a/src/util/hooks.js
+++ b/src/util/hooks.js
@@ -1,8 +1,10 @@
 /* @flow */
 
+export type YarnHook = 'resolveStep' | 'fetchStep' | 'linkStep' | 'buildStep';
+
 const YARN_HOOKS_KEY = 'experimentalYarnHooks';
 
-export function callThroughHook<T>(type: string, fn: () => T): T {
+export function callThroughHook<T>(type: YarnHook, fn: () => T): T {
   if (typeof global === 'undefined') {
     return fn();
   }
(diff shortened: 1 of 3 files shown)


src/util/hooks.js
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+
+export function callThroughHook<T>(type: string, fn: () => T): T {
```
BYK: Let's name this `triggerHook` or `callHook` and use an enum or `Symbol`s for hook names.

src/util/hooks.js
```diff
@@ -0,0 +1,19 @@
...
+  if (typeof global === 'undefined') {
```
BYK: Maybe we should move these checks outside of the function call and make them static? Also, instead of a callback interface, shouldn't we promote a `Promise` based interface? That would encourage `async` and `await` usage and would also simplify the code:
```js
const hooks = typeof global !== 'undefined' && global.yarnHooks || {};
const defaultHook = () => Promise.resolve();

export function triggerHook<T>(type): Promise<T> {
    return (hooks[type] || defaultHook)();
};
```

BYK:
Also, let's add at least one smoke test to make sure this actually works.

src/util/hooks.js
arcanis: We can't, otherwise the hooks will have no way to know when the actions start and end (unless I create one extra hook for each action to mark the end).

src/util/hooks.js
arcanis: I prefer `callThroughHook` because it underlines that the hook has the responsibility of actually calling the callback.

I improved the Flow typing of the `type` parameter to be an enumeration that only accepts valid field. Symbols aren't possible because the hooks will be setup before Yarn is even required.

src/util/hooks.js
```diff
@@ -0,0 +1,23 @@
+/* @flow */
+
+export type YarnHook = 'resolveStep' | 'fetchStep' | 'linkStep' | 'buildStep';
```
BYK: I would remove the `Step`in the next iteration since it doesn't add any value.

arcanis: It was like this at the beginning, then I considered that we might want to have a higher granularity of hooks later on. For example a hook that would be called once for each package we fetch.

Issue #5273, opened 2018-01-24
Yarn cache clean does not handle several package name at once


ThomasVuillaume:
**Do you want to request a *feature* or report a *bug*?**
A feature

**What is the current behavior?**
When you yan "yarn cache clean module1 module2 module3 ... moduleN", only module1 is cleared from cache.

**What is the expected behavior?**
Have it to work the same way as "yarn add" or "yarn remove", and be able to clear several modules from cache a once.

**Please mention your node.js, yarn and operating system version.**
Yarn : 1.2.1
Node: v6.10.1
OS: Windows 7

lankaapura:
@rally25rs I've sent a PR. Let me know if there any changes needed.

ThomasVuillaume:
Nice ! thanks :+1:

PR #5297, opened 2018-01-31
'Cache clean' with multiple packages at once

lankaapura:
This PR fixes https://github.com/yarnpkg/yarn/issues/5273


bestander:
Isn't it awesome?
Thanks for the contribution!

PR #5314, opened 2018-02-03
Ability to store a built package in offline mirror

bestander:
**Summary**

Packages with install scripts will be saved into a .tgz file into offline mirror if `experimental-pack-script-packages-in-mirror` is set in .yarnrc.

Next time you run `yarn install` instead of unpacking the original .tgz package and running scripts yarn unpacks the previously built one and does not run any scripts.

Good for CI systems, slow machines and when you don't trust packages to do run shady download scripts.

Of course it adds platform specific suffixes to the .tgz file name, so that you can safely share the prebuilt offline mirror folder with other OS and Node versions.

Follow up: blog post and docs.

**Test plan**

Added tests for all common cases: rerunning install, switching Node version etc.


src/package-install-scripts.js
```diff
@@ -292,6 +303,50 @@ export default class PackageInstallScripts {
...
+        // TODO ! .yarn-integrity should contain prebuiltPackages array now
+        pkg.prebuiltVariants = pkg.prebuiltVariants || {};
+        pkg.prebuiltVariants[filename] = hash;
+      }
+    }
     set.end();
+
+  }
+
+  static getPrebuiltName(pkg: Manifest): string {
```
bestander: todo: put in a separate file

src/package-install-scripts.js
```diff
@@ -292,6 +303,50 @@ export default class PackageInstallScripts {
...
+          console.log("building", pkg.name)
+          const validateStream = new crypto.HashStream();
+          stream
+          .pipe(validateStream)
+          .pipe(fs2.createWriteStream(filePath))
+          .on('error', reject)
+          .on('close', () => resolve(validateStream.getHash()));
+        });
+        // TODO ! don't save artifacts in .yarn-integrity, it is part of the package now
+        // TODO ! .yarn-integrity should contain prebuiltPackages array now
```
bestander: todo: add tests for this

src/resolvers/registries/npm-resolver.js
```diff
@@ -176,6 +177,14 @@ export default class NpmResolver extends RegistryResolver {
     // lockfile
     const shrunk = this.request.getLocked('tarball');
     if (shrunk) {
+      if (shrunk.prebuiltVariants && shrunk._remote) {
+        const prebuiltName = PackageInstallScripts.getPrebuiltName(shrunk);
+        if (shrunk.prebuiltVariants[prebuiltName]) {
+          const filename = this.config.getOfflineMirrorPath(prebuiltName + '.tgz');
```
bestander: todo: verify that file exists

__tests__/commands/install/integration.js
```diff
@@ -145,22 +143,6 @@ test('reading a lockfile should not optimize it', async () => {
   });
 });
```
bestander: moved offline mirror tests out

__tests__/commands/install/offline-mirror.js
```diff
@@ -0,0 +1,318 @@
...
+    await reinstall.init();
+
+    tgzFiles = await fs.readdir(path.join(config.cwd, 'mirror-for-offline', 'prebuilt'));
+    expect(await fs.exists(path.join(config.cwd, 'node_modules', 'dep-a', 'module-a-build.log'))).toEqual(true);
+    expect(await fs.exists(path.join(config.cwd, 'module-a-build.log'))).toEqual(false);
+    expect(tgzFiles.length).toBe(2);
+  });
+});
+
+test('creates the file in the mirror when fetching a git repository', async () => {
```
bestander: following tests are copied from integration.js file

src/util/package-name-utils.js
```diff
@@ -0,0 +1,13 @@
+/* @flow */
+
+export function getPlatformSpecificPackageFilename(pkg: {name: string, version: string}): string {
+  // TODO support hash for all subdependencies that have installs scripts
```
arcanis: This will be problematic, yeah :/ Something I had in mind was that maybe we could implement nested lockfiles to solve this issue. Given a nested lockfile, we could guarantee that a package would always be built against a specific set of dependencies, and we could cache it more efficiently. Wdyt?

src/util/package-name-utils.js
```diff
@@ -0,0 +1,13 @@
...
+  const normaliseScope = name => (name[0] === '@' ? name.substr(1).replace('/', '-') : name);
+  const suffix = getSystemParams();
+  return `${normaliseScope(pkg.name)}-v${pkg.version}-${suffix}`;
```
arcanis: Nit: normalize

src/config.js
```diff
@@ -328,6 +331,7 @@ export default class Config {
     this.enableMetaFolder = Boolean(this.getOption('enable-meta-folder'));
     this.enableLockfileVersions = Boolean(this.getOption('yarn-enable-lockfile-versions'));
     this.linkFileDependencies = Boolean(this.getOption('yarn-link-file-dependencies'));
+    this.packBuiltPackages = Boolean(this.getOption('pack-built-packages'));
```
arcanis: Shouldn't the option be called `experimental-build-mirror` or something like this, to clearly state that it's experimental?

src/package-install-scripts.js
```diff
@@ -280,15 +292,55 @@ export default class PackageInstallScripts {
...
-        const beforeFiles = beforeFilesMap.get(loc);
-        invariant(beforeFiles, 'files before installation should always be recorded');
-        await this.saveBuildArtifacts(loc, pkg, beforeFiles, set.spinners[0]);
+    // generate built package as prebuilt one for offline mirror
+    if (this.config.packBuiltPackages) {
+      for (const pkg of pkgs) {
+        if (this.packageCanBeInstalled(pkg)) {
+          const offlineMirrorPath = this.config.getOfflineMirrorPath();
+          if (!offlineMirrorPath) {
+            break;
```
arcanis: If the packBuiltPackages option is set but no offline mirror has been set, the `saveBuildArtifacts` function won't be called. Is it expected?

src/package-install-scripts.js
```diff
@@ -280,15 +292,55 @@ export default class PackageInstallScripts {
...
+          const ref = pkg._reference;
+          invariant(ref, 'expected reference');
+          const loc = this.config.generateHardModulePath(ref);
+          const pkgConfig = await Config.create(
+            {
+              cwd: loc,
+            },
+            this.reporter,
+          );
+          const stream = await pack(pkgConfig, loc);
```
arcanis: Does `pack` also includes the build artifacts? :o

src/util/package-name-utils.js
bestander: I think nested lockfile is an overkill and may be confusing considering that we don't support nested lockfiles in general.
We might generate something in between - a lock-like file that will only have `dependencies` section because we don't really need "version" and "resolved" fields as they are already defined in the parent yarn.lock

bestander: On the other hand having that file inside the .tgz file requires us to unzip it before checking whether the archive works for us, we need to encode all dependencies into the file suffix and hashing them seems a reasonable way to shorten the suffix

bestander: Anyway, we'll discuss it in V2 of this feature, we need to see how it will be used by the users if it will be :)

src/util/package-name-utils.js
bestander: done

src/config.js
bestander: Yeah, makes sense.
How about `experimental-pack-script-packages-in-mirror`?

src/package-install-scripts.js
bestander: Good point!
Not inteded

src/package-install-scripts.js
bestander: I pass `loc` which from local `node_modules` with built artifacts, pack just generates a .tgz from a folder

bestander: I'll name loc better to be clear

bestander:
@arcanis, ready for another round

src/package-install-scripts.js
```diff
@@ -3,12 +3,17 @@
...
-import type Config from './config.js';
+import Config from './config.js';
 import type {ReporterSetSpinner} from './reporters/types.js';
 import executeLifecycleScript from './util/execute-lifecycle-script.js';
+import * as crypto from './util/crypto.js';
 import * as fs from './util/fs.js';
+import {getPlatformSpecificPackageFilename} from './util/package-name-utils.js';
+import {pack} from './cli/commands/pack.js';
 
+const fs2 = require('fs');
```
arcanis: I find that a bit confusing - maybe createWriteStream should be simply reexported from `util/fs` (or maybe we should use `fs-extra` which takes care of exporting promises when we need them).

src/package-install-scripts.js
```diff
@@ -280,15 +292,52 @@ export default class PackageInstallScripts {
...
+            const validateStream = new crypto.HashStream();
+            stream
+              .pipe(validateStream)
+              .pipe(fs2.createWriteStream(prebuiltPath))
+              .on('error', reject)
+              .on('close', () => resolve(validateStream.getHash()));
+          });
+          pkg.prebuiltVariants = pkg.prebuiltVariants || {};
+          pkg.prebuiltVariants[prebuiltFilename] = hash;
+          // this.artifacts[`${pkg.name}@${pkg.version}`] = [prebuiltFilename];
```
arcanis: Can we remove this line if it isn't used?

src/resolvers/registries/npm-resolver.js
```diff
@@ -180,6 +181,18 @@ export default class NpmResolver extends RegistryResolver {
     // lockfile
     const shrunk = this.request.getLocked('tarball');
     if (shrunk) {
+      if (this.config.packBuiltPackages && shrunk.prebuiltVariants && shrunk._remote) {
+        const prebuiltVariants = shrunk.prebuiltVariants;
+        const prebuiltName = getPlatformSpecificPackageFilename(shrunk);
+        const offlineMirrorPath = this.config.getOfflineMirrorPath();
+        if (prebuiltVariants[prebuiltName] && offlineMirrorPath) {
+          const filename = path.join(offlineMirrorPath, 'prebuilt', prebuiltName + '.tgz');
+          if (shrunk._remote && (await fs.exists(filename))) {
```
arcanis: Fwiw the Node documentation recommends using `fs.existsSync` instead of `fs.exists` to prevent race condition as much as possible (`fs.exists` is actually deprecated).

src/package-install-scripts.js
```diff
@@ -280,15 +292,52 @@ export default class PackageInstallScripts {
...
-        const beforeFiles = beforeFilesMap.get(loc);
-        invariant(beforeFiles, 'files before installation should always be recorded');
-        await this.saveBuildArtifacts(loc, pkg, beforeFiles, set.spinners[0]);
+    // generate built package as prebuilt one for offline mirror
+    const offlineMirrorPath = this.config.getOfflineMirrorPath();
+    if (this.config.packBuiltPackages && offlineMirrorPath) {
+      for (const pkg of pkgs) {
+        if (this.packageCanBeInstalled(pkg)) {
+          let prebuiltPath = path.join(offlineMirrorPath, 'prebuilt');
+          if (!await fs.exists(prebuiltPath)) {
```
arcanis: We don't have to check it exists since we're using `mkdirp` which will do it for us, right?

src/package-install-scripts.js
bestander: Doh, debugging code

src/package-install-scripts.js
bestander: Yeah, fs2 is a poor choice for naming this variable.
Let's do it like in other places the Node.js fs module should be `fs` but out promise wrappers should be named `fsUtil`.

A year ago there was a proposal to reexport all exports from Node.js `fs` module there but we decided not to do it because it means we would have to manually list all functions of Node.js `fs`. Could be just extra trouble to worry when Node.js introduces a new function or breaks API

src/resolvers/registries/npm-resolver.js
bestander: We have 59 places where this wrapper is used.
If we migrate away from it we should do it in a separate PR, right?

bestander:
@arcanis, ready for another round

src/package-install-scripts.js
arcanis: `fs-extra` automatically exports all the `fs` Node functions in such a way that they automatically support promises. Using it would allow us to let someone else care about keeping the list updated! :) Ok, not urgent, let's keep this for later

hoodie:
Question: does this depend on the engine being used? because some dependencies need to be rebuilt if installed for node or electron

bestander:
@hoodie, currently platform, architecture and modules (Node libs version) are taken into account
https://github.com/yarnpkg/yarn/blob/master/src/util/package-name-utils.js#L12

Feel free to send a PR to add another parameter.

hoodie:
@bestander that seems to be thorough enough, thank you :D

lxe:
I absolutely love this feature. This adds a "supported" way to bring hermetic builds across platforms and environments.

tommilligan-plutoflume:
For those wishing to enable this feature from a clean yarn installation, simply to speed up local development rebuilds - add the following in your project root:

**.yarnrc**
```
yarn-offline-mirror "./yarn-offline-cache"
yarn-offline-mirror-pruning true
experimental-pack-script-packages-in-mirror true
```
**.gitignore**
```
/yarn-offline-cache
```

Loving this feature already!

bazyli-brzoska:
We're having issues with this feature, as it saves references to prebuilt packages to `yarn.lock` and those change, I think based on the architecture of the computer (?). It caused `yarn install`s to fail in some scenarios for us, but we didn't had time to debug fully, and thus had to disable it.

theandrewlane:
Looks like ```yarn-offline-mirror-pruning true``` deletes```./<yarn-offline-cache>/prebuilt/*```, is this as designed?

bestander:
No :)
Could you send a PR to fix that if you have time?

theandrewlane:
@bestander - if you can confirm that the ```yarn-offline-mirror-pruning true``` config option is deleting the contents of /prebuilt for you, and shouldn't be, i'll submit a PR :)

bestander:
Please do and thank you!

mikaello:
@theandrewlane I have the same problem, that `yarn-offline-mirror-pruning true` removes the content of `/prebuilt`

theandrewlane:
@mikaello: Cool, thanks for verifying!

Created #5614 to address this issue, PR coming this week :)

PR #5318, opened 2018-02-05
fix(cli): fail when `yarn add`ing Github package with no version (#5292)

### LINKED ISSUE #5292: Yarn installs package without version and creates breaking error

Reporting a bug

If I run this
```
$ yarn add github:Studio-42/elFinder
```
It will successfully install the package, but it will add this to the lockfile:
```
"elFinder@github:Studio-42/elFinder":
  version ""
  resolved "https://codeload.github.com/Studio-42/elFinder/tar.gz/12383f05a30d61bf3ead4f94dbfe0a3a3073bfa9"
```
This is because the `package.json` file in the GitHub repo is missing the `version` field.

Next, I'll run this
```
$ yarn add some-other-package
```
And get this error:
```
yarn add v1.3.2
[1/4] Resolving packages...
error Package "elFinder@" doesn't have a "version".
```
Because of the empty version in the lockfile.

See above steps.

Don't install the package that is missing the version; do what NPM does when I try to install the same package:
```
$ npm install --save github:Studio-42/elFinder
npm ERR! Can't install github:Studio-42/elFinder#12383f05a30d61bf3ead4f94dbfe0a3a3073bfa9: Missing package version

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Joshua\AppData\Roaming\npm-cache\_logs\2018-01-30T01_29_17_779Z-debug.log
```

This prevents the lock file from being corrupted and preventing additional commands from being run.

Node.js v9.4.0
Yarn 1.3.2
Windows 10.0.16299

**Additional Info**
I should mention that both #1602 and #4052 seem to concern the same issue, but both of them were closed without addressing what I believe is the actual problem: the fact that yarn will install an invalid package in the first place and thus corrupt the lockfile.

mhchen:
**Summary**

Running `yarn add` to install from a Github repo that has a package.json with no version succeeds, but then all further `yarn` operations fail due to failing checks for `version`s in dependencies. As well, per npm, the install method should not succeed at all. This PR causes `yarn add` to fail if `version` is not properly formatted as a semver.

Fixes #5292 

**Test plan**

Running:

```
$ yarn add github:Studio-42/elFinder
```

Produces the following output:
```
yarn add v1.4.1
warning ../../package.json: No license field
[2/5] Resolving packages...
error Can't add "elFinder": invalid package version undefined.
```


mhchen:
@bestander please review at your convenience. Thank you!

src/util/git.js
```diff
@@ -473,7 +473,7 @@ export default class Git implements GitRefResolvingInterface {
     });
     if (!resolvedResult) {
       throw new MessageError(
-        this.reporter.lang('couldntFindMatch', version, Object.keys(refs).join(','), this.gitUrl.repository),
+        this.reporter.lang('couldntFindMatch', version, Array.from(refs.keys()).join(','), this.gitUrl.repository),
```
mhchen: `refs` is a `Map` which always returns an empty array with `Object.keys()`

__tests__/commands/add.js
```diff
@@ -114,6 +114,16 @@ test.concurrent('install from github', async () => {
...
 });
 
+test.concurrent('install from github with invalid version should fail', async () => {
+  let message = '';
+  try {
+    await runAdd(['yarnpkg/example-yarn-package#invalid-package-json-version'], {}, 'install-github');
+  } catch (err) {
+    message = err.message;
+  }
+  expect(message).toEqual(expect.stringContaining('invalid package version'));
```
mhchen: Would be preferable to use `.rejects` and assert on the error message using `toThrow` or the like, but it looks like this is still problematic: https://github.com/facebook/jest/issues/3601

JoshuaCWebDeveloper:
Thank you @mhchen !

bestander:
@mhchen great work! Thank you for following it through and doing tests.

PR #5320, opened 2018-02-05
Horizontal rule fixed in README

liamross:
**Summary**

The horizontal rule had no empty space between it and the previous paragraph, and as such was rendered as '---' in the markdown, rather than the intended horizontal rule.

This is not a substantial feature pull request.

Motivation: it will look better.

**Test plan**

Not relevant.


bestander:
Thanks :)

Issue #5319, opened 2018-02-05
When building sourcemaps with `yarn run build --sourcemaps`, add sourceRoot


edeustace:
**Do you want to request a *feature* or report a *bug*?**

bug - I think

**What is the current behavior?**
running `yarn run build --sourcemaps` generates `*.map` files, but these files are missing a `sourceRoot` property.

Adding `sourceRoot` means that you can set a breakpoint in the code in the `src/` dir, launch `node --inspect-brk`, attach your favorite debugger and the breakpoint will be hit. If you don't set this then you have to configure your debugger to map the path of the *.map file to the src/*.js file.

> It may be that this is desired behavior, if that's the case it'd be great if there was a developer build, which one can run and have `sourceRoot` set so that one can start debugging the code with no configuration.

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
git clone git@github.com:yarnpkg/yarn.git
cd yarn
yarn install
yarn run build --sourcemaps
# using ripgrep but any grep tool will do (or just view a *.map file in your favorite editor).
rg -i "sourceRoot" lib/*.map 

# !! output is empty
```

**What is the expected behavior?**
ripgrep output showing that `sourceRoot` is set correctly on the *.map files.

**Please mention your node.js, yarn and operating system version.**
node v9.4.0, yarn 1.3.2, OS X 10.13.2

This commit fixes it: 
https://github.com/PieLabs/yarn/commit/d6ed0a25fa408ea417f6f540e29e88e28b2eb32a

rally25rs:
Thanks for finding a resolution to the issue.
Could you submit the above linked fix as a PR and mention this issue?

edeustace:
There you go @rally25rs

PR #5323, opened 2018-02-06
fix(build): add sourceRoot to sourcemaps

edeustace:
**Summary**
Add `sourceRoot` to sourcemaps config in the gulpfile.js. Fixes: #5319.

Without this a developer who wants to debug the yarn src would have to manually configure a mapping between the lib/**/*.map files and the src/**/*.js files so that the breakpoints get hit. With this fix this manual configuration step is not necessary.

**Test plan**

```shell
yarn run build --sourcemaps
rg -i "sourceRoot" lib/*.map # should get some output from ripgrep listing sourceRoots in .map files.
```


arcanis:
Thanks!

PR #5329, opened 2018-02-07
fix(workspace) Workspace pass args 5301

### LINKED ISSUE #5301: yarn workspace command doesn't forward flags to commands
bug
`yarn workspace {name} {command} --anything` does not forward the flags to `{command}`

This is due to the implementation in `src/commands/workspace.js`:

```

  ...

  const [workspaceName, ...rest] = args;

  ...
  try {
    await child.spawn(NODE_BIN_PATH, [YARN_BIN_PATH, ...rest], {
```

`rest` is the remaining parameters, but does not include `flags`

Example, make a monorepo/workspace project with a package named `foo`.

Run `yarn workspace foo version --new-version 0.1.0`

The `workspace` command executes `yarn version` in the `foo` package without the `--new-version` flag.

Flags should be forwarded to command

yarn v1.4.1 (current master)

rally25rs:
fixes #5301

**Summary**

Previously the `yarn workspace {workspace_name} {command} {args...}` command would only forward plain arguments (words) to the child command process. Flags or any arguments starting with `-` were not forwarded.

This was due to the way that commander.js parsed those args into separate "buckets" of arguments. Now all command line parameters including flags will be passed to the child process.

**Test plan**

Added a unit test for argument passing. The new functionality relies on commander.js putting all raw arguments in `flags.rawArgs` (what we take in as `args` is actually the commander.js instance). Since the unit tests don't pass through commander.js, I had to pass `rawArgs` as a "flag" in the unit tests, so it looks a little weird. I left a comment in the code to try to explain it since we don't follow that pattern elsewhere.


bestander:
Thank you, @rally25rs!

PR #5336, opened 2018-02-08
RFC: feat(config): Use more conventional paths for config and data

wbinnssmith:
This implements:

* Supporting user-defined environment variables adhering to the [XDG
Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html)
to override yarn's defaults
* Supporting equivalent environment variables on Windows such as
%LOCALAPPDATA%
* More conventional defaults for these locations according to the
operating system.
* Support for the user defined config dir in the `.yarnrc` lookup path
* Storing global modules in a data-oriented location rather than a
config location (seeing this is actually what motivated this PR)

Concerns:

* Existing Windows config locations will break. This probably need to be
addressed with a migration path and/or a breaking change
* A few notes included in comments (will highlight these with inline GH
comments)
* Unclear test status as master builds appears to fail on my laptop as
well.

Really interested in your feedback. I know [this has been attempted
before](https://github.com/yarnpkg/yarn/pull/3674/files) -- cc @kelseasy
-- and I'd really like to get this in!

### DIFF
--- a/src/util/user-dirs.js
+++ b/src/util/user-dirs.js
@@ -1,0 +1,45 @@
+/* @flow */
+
+const path = require('path');
+const userHome = require('./user-home-dir').default;
+
+export function getDataDir(): string {
+  if (process.platform === 'win32') {
+    return path.join(getLocalAppDataDir(), 'Data');
+  } else if (process.env.XDG_DATA_HOME) {
+    return path.join(process.env.XDG_DATA_HOME, 'yarn');
+  } else {
+    // This could arguably be ~/Library/Application Support/Yarn on Macs,
+    // but that feels unintuitive for a cli tool
+
+    // Instead, always use the typical location of XDG_DATA_HOME
+    return path.join(userHome, '.local', 'share', 'yarn');
+  }
+}
+
+export function getCacheDir(): string {
+  if (process.platform === 'win32') {
+    // process.env.TEMP also exists, but most apps put caches here
+    return path.join(getLocalAppDataDir(), 'Cache');
+  } else if (process.env.XDG_CACHE_HOME) {
+    return path.join(process.env.XDG_CACHE_HOME, 'yarn');
+  } else if (process.platform === 'darwin') {
+    return path.join(userHome, 'Library', 'Caches', 'Yarn');
+  } else {
+    return path.join(userHome, '.cache', 'yarn');
+  }
+}
+
+export function getConfigDir(): string {
+  if (process.platform === 'win32') {
+    return path.join(getLocalAppDataDir(), 'Config');
+  } else if (process.env.XDG_CONFIG_HOME) {
+    return path.join(process.env.XDG_CONFIG_HOME, 'yarn');
+  } else {
+    return path.join(userHome, '.config', 'yarn');
+  }
+}
+
+function getLocalAppDataDir(): string {
+  return path.join(process.env.LOCALAPPDATA || path.join(userHome, 'AppData', 'Local'), 'Yarn');
+}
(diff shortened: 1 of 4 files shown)


wbinnssmith:
cc @Daniel15 I feel like you'd have some insight here, especially re: Windows paths

src/util/user-dirs.js
```diff
@@ -0,0 +1,45 @@
...
+    // but that feels unintuitive for a cli tool
+
+    // Instead, always use the typical location of XDG_DATA_HOME
+    return path.join(userHome, '.local', 'share', 'yarn');
+  }
+}
+
+export function getCacheDir(): string {
+  if (process.platform === 'win32') {
+    // process.env.TEMP also exists, but most apps put caches here
```
wbinnssmith: @Daniel15 do you know about this?

Daniel15:
I can take  a look once I'm back from vacation.

src/util/user-dirs.js
```diff
@@ -0,0 +1,45 @@
...
+  if (process.platform === 'win32') {
+    return path.join(getLocalAppDataDir(), 'Data');
+  } else if (process.env.XDG_DATA_HOME) {
+    return path.join(process.env.XDG_DATA_HOME, 'yarn');
+  } else {
+    // This could arguably be ~/Library/Application Support/Yarn on Macs,
+    // but that feels unintuitive for a cli tool
+
+    // Instead, always use the typical location of XDG_DATA_HOME
+    return path.join(userHome, '.local', 'share', 'yarn');
```
bestander: I have a concern that this changes current setup people have, may cause confusion.

Before it was `~/.config/yarn/global`, now it will be `~/.local/share/yarn/global`.
Do we really need this change?

wbinnssmith: Honestly this is my main reason for the PR 😝 

I thought it was really, really weird to be storing a ton of binaries in the user's *configuration* directory. The kind of stuff we put here is a perfect fit with XDG_DATA_HOME, which on many machines defaults to this location.

wbinnssmith:
> But I would keep the old paths if they unless there is a stronger reason to migrate.

Yeah, I definitely understand. Personally I'd like to discuss a migration path (partly why this PR is RFC). Maybe a quick stat to see if this directory exists and either move its contents or fall back to it?

kelseasy:
The idea behind the change is that it's easier for people using mounts to take advantage of the xdg spec (like having your flash drive mounted on the cache directory) to get a working setup out of the box without having to rely on creating symlinks.

The approach used in other package managers is usually to check if the legacy directory exists and use it if it does. You can also have a symlink for a few releases to give people time to adapt.

Thanks for picking this up @wbinnssmith!

bestander:
Looks good, let's merge it and discuss the migration path in the next PR

src/util/user-dirs.js
Daniel15: TEMP is for temp files (anything you don't care about any more after some period of time while it's being used), whereas I think we want to keep the Yarn cache around.

src/util/user-dirs.js
```diff
@@ -0,0 +1,54 @@
...
+    return FALLBACK_CACHE_DIR;
+  }
+}
+
+export function getConfigDir(): string {
+  if (process.platform === 'win32') {
+    // Use our prior fallback. Some day this could be
+    // return path.join(WIN32_APPDATA_DIR, 'Config')
+    const WIN32_APPDATA_DIR = getLocalAppDataDir();
+    return WIN32_APPDATA_DIR == null ? FALLBACK_CONFIG_DIR : path.join(WIN32_APPDATA_DIR, 'Config');
```
Daniel15: If we want to be the "most correct" here, I think the config should go in AppData, while the cache should go in LocalAppData. The difference is that LocalAppData is for anything that should **not** be saved to a roaming profile (ie. synchronised to a server), while regular (roaming) AppData is anything that could/should be synchronised.

Issue #2172, opened 2016-12-07
Engines check broken for nightly builds


swernerx:
We have a package which has the following `engines` field:

```
"engines": {
  "node": ">=6.0.0",
  "yarn": ">=0.18.0",
  "npm": ">=4.0.0"
}
```

in `package.json`. As there is not yet a stable "msi" package for Windows users for installation we thought of installing the latest nightly. Unfortunately yarn does not seem to be happy with the installed version. Possibly this is because the version check is somehow flawed.

Output from installing:

```
yarn install v0.18.0-20161123.1451
[1/4] Resolving packages...
[2/4] Fetching packages...
error advanced-boilerplate@0.14.7: The engine "yarn" is incompatible with this module. Expected version ">=0.18.0".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

This seems to be a bug as `v0.18.0-20161123.1451` should be successfully matched by `>=0.18.0`.

Daniel15:
> As there is not yet a stable "msi" package for Windows users for installation

https://yarnpkg.com/latest.msi will get you the latest stable build.

swernerx:
@Daniel15 thanks, problem is it seems the latest stable differes widely depending on where you look. Homebrew/Git is at 0.18.0 while npm/msi is on 0.17.x.

Daniel15:
0.18.0 is essentially a release candidate.

swernerx:
aha - interesting - It would be probably helpful to follow semver ;)

Daniel15:
Yeah, https://github.com/yarnpkg/yarn/issues/2161 tracks that. There's some
tooling changes that need to be done.

gsklee:
@bestander close

On the original issue that `"yarn": ">=0.18.0"` couldn't match `0.18.0-20161123.1451`, can't do anything here because this is exactly how [`node-semver`](https://github.com/npm/node-semver) works; eg. if you require a `react` version `>=15`, you will not get version `16.0.0-alpha.13`. You should bring this issue to `node-semver` instead.

bestander:
Actually Yarn decided to break semver here after multiple requests from the community.
This should be implemented in 0.25

bdwain:
@bestander was this actually implemented in 0.25? I'm experiencing this issue with the nightly build for 1.4.1 when trying to install https://github.com/sebastian-software/rollup-plugin-rebase (which has

bestander:
[...]

PR #5352, opened 2018-02-12
break semver for engine check by allowing prereleases

bdwain:
**Summary**
Addresses #2172 
When trying to run a prerelease version of yarn, packages that specify versions of yarn in their engine requirements in package.json cause installation to fail. When running prerelease versions of yarn, you most likely don't want this engine check to fail because you have already signed up for potentially unexpected behavior by using a prerelease version.

**Test plan**
[rollup-plugin-rebase](https://github.com/sebastian-software/rollup-plugin-rebase/blob/master/package.json) has an engine requirement of `>=1.0.0`

Install nightly build (1.4.1-20180211.2236)
`yarn add rollup-plugin-rebase`  (fails engine check)

alias local copy of yarn (1.4.1-20180208.2355)
`yarn add rollup-plugin-rebase`  (success)

### DIFF
--- a/src/package-compatibility.js
+++ b/src/package-compatibility.js
@@ -6,6 +6,7 @@ import {MessageError} from './errors.js';
 import map from './util/map.js';
 import {entries} from './util/misc.js';
 import {version as yarnVersion} from './util/yarn-version.js';
+import {satisfiesWithPreleases} from './util/semver.js';
 
 const invariant = require('invariant');
 const semver = require('semver');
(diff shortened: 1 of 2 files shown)


src/package-compatibility.js
```diff
@@ -70,6 +71,10 @@ export function testEngine(name: string, range: string, versions: Versions, loos
     return true;
   }
 
+  if (name === 'yarn' && satisfiesWithPreleases(actual, range, looseSemver)) {
```
bdwain: I could have made the check on line 70 use satisifiesWithPrerelease and not bothered with the yarn specific check, but it felt safer to just do yarn for now.

bdwain:
I can also add a note to the documentation for the engines section. It currently says 

> The engines specify versions of clients that must be used with your package. This checks against process.versions as well as the current version of yarn.

bestander:
Awesome, @bdwain!
If you could send a PR for documentation change it will be great, just ping me

bdwain:
I made https://github.com/yarnpkg/website/pull/779 @bestander

PR #1083, opened 2016-10-15
Enable gzip compression

kdzwinel:
**Summary**

Gzip compression wasn't enabled causing yarn to download twice the data npm does. Fixes #945 .

**Test plan**
`yarn add yo`

Before - 9.9MB
before

After - 5.3MB
after

PR #5369, opened 2018-02-16
Enable curl `--compressed` option to speed up request

PeterDaveHello:
This will let `curl` automatically to send `Accept-Encoding` header to include the compression methods such like `deflate`, `gzip`, etc, to request compressed data from server, and automatically decompress the data if it's compressed, I believe that this is very like(pretty the same) what commit 0829a745ae5d6948d939d0c181bbb7bb2cd1686a / PR #1083 did, will save the time and bandwidth on transferring data.

The output and other behavior should be the same as `curl` will automatically handle the process to add header and to decompress.

My first PR to this project, let me know if I need to revise/improve anything, thanks a lot!

### DIFF
--- a/scripts/install-latest.sh
+++ b/scripts/install-latest.sh
@@ -5,4 +5,4 @@ reset="\e[0m"
 red="\e[0;31m"
 
 printf "${red}This script has moved to https://yarnpkg.com/install.sh, please update your URL!$reset\n"
-curl -o- -L https://yarnpkg.com/install.sh | bash -s -- "$@"
+curl --compressed -o- -L https://yarnpkg.com/install.sh | bash -s -- "$@"
--- a/scripts/update-homebrew.sh
+++ b/scripts/update-homebrew.sh
@@ -12,7 +12,7 @@ fi;
 # number from Yarn site
 if [ -z "$YARN_VERSION" ]; then
   echo 'Getting Yarn version from https://yarnpkg.com/latest-version'
-  version=`curl --fail https://yarnpkg.com/latest-version`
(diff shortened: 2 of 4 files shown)


arcanis:
Thanks! Technically this is correct, but I think we'll keep the command line as it is - I prefer to keep the one-liner to stay short and simple so that users can quickly parse it mentally and understand it. Sounds good?

PeterDaveHello:
What about to move all parameters in a variable? Though I think it's still very short here, it has obvious advantage, and don't think most of the users need to read this source code behind yarn, for those who do trace source code, if they don't recognize this parameter yet, it's also a good opportunity to learn that, as it's not a complex parameter, very easy to understand :)

arcanis:
As I see it there's only two cases:

- Either the users will type the command themselves, in which case the perf gain is counterbalanced by the extra characters

- Or the users will copy/paste the command, in which case they won't even notice the argument.

Not sure we gain a lot in either case.

arcanis:
I'm going to close this PR, but thanks anyway to have took the time to make this proposal! 🙂

PeterDaveHello:
hmmm, I guess most people will copy and paste it and in most cases the users don't need to do it manually? The changes are in the scripts/codes level only but not the user manual, so I'd help this is acceptable, thanks :)

arcanis:
Oh wait, I completely misread the PR, I remembered seeing the change in the readme but I must have been mistaken 😧 

Yeah it totally makes sense, let's merge this!

PeterDaveHello:
Thanks!

PR #5384, opened 2018-02-21
Remove unnecessary console output when running test cases

sth:
**Summary**

Running the tests did, for some test cases, show yarn command output like `[1/4] Resolving packages...`. This pull request captures the command output instead of printing it to the console, the same way as this is done in many other test cases.

Unnecessary command output during testing makes the tests noisy. We are mostly only interested in test success or failure messages, not all the details of all the commands that run internally. Even *if* the command output were interesting, it usually should be captured and tested in the test, not written to console.

**Test plan**

The tests still pass, just with less random console output.

### DIFF
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -1140,7 +1140,7 @@ test.concurrent('installs "latest" instead of maxSatisfying if no requested patt
 });
 
 describe('nohoist', () => {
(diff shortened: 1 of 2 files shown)


arcanis:
Makes sense, not sure why those tests specifically used a `ConsoleReporter` in the first place 🤔

PR #5385, opened 2018-02-21
Fix test for `yarn licenses generate-disclaimer` output

sth:
**Summary**

There is [a test][testcase] that attempts to test that license files are show correctly for `yarn licenses generate-disclaimer`. The test currently just prints the license files to the console and doesn't verify what was printed. This pull request modifies the test to not show unwanted output during testing and to actually test that the output is correct.

**Implementation**

The normal method of capturing output isn't sufficient to capture the output of `yarn licenses generate-disclaimer`, since that command [intentionally][1] uses `console.log()` directly. Because of this a special helper function is used that mocks `console.log()` and lets the test verify what was written.

**Test plan**

The `licenses.js.snap`, which is used to verify the commands output, now contains the license text we expect to be printed. The test passes using this .snap file.

**Additional change**

I also renamed the "lists all licenses of the dependencies with the --json argument" test in the same file because it doesn't seem to involve any --json argument.

  [testcase]: https://github.com/yarnpkg/yarn/blob/1a4e9ff4128b3aefe6b57e77a3456813dea6f28a/__tests__/commands/licenses.js#L28
  [1]: https://github.com/yarnpkg/yarn/blob/1a4e9ff4128b3aefe6b57e77a3456813dea6f28a/src/cli/commands/licenses.js#L137

### DIFF
--- a/__tests__/commands/__snapshots__/licenses.js.snap
+++ b/__tests__/commands/__snapshots__/licenses.js.snap
@@ -13,3 +13,35 @@ exports[`should generate disclaimer on demand 1`] = `
 {\\"type\\":\\"warning\\",\\"data\\":\\"package.json: No license field\\"}
 "
 `;
+
+exports[`should generate disclaimer on demand 2`] = `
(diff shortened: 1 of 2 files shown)


arcanis:
Sounds good 👍

PR #5387, opened 2018-02-21
adjusted windows path so that bin folder has a backslash on the end o…

derekellison:
…f the path value, as per windows path requirements

**Summary**
this is a bug fix for the windows environment path

fixes windows path bug

**Test plan**

tested via adjusting windows system path on local machine - screenshot should not be needed for such a simple change.


arcanis:
Thanks!

PR #5388, opened 2018-02-21
Implements "yarn node"

arcanis:
**Summary**

This PR implements a new command called `yarn node`. It is meant to call Node with the correct environment to run the application. The main difference with running Node directly is that Yarn will automatically "cd" to the right directory.

**Test plan**

I manually ran the following commands and checked their outputs:

```
yarn node
yarn node -p 42
yarn node -- -p 42
yarn node script.js
```

I haven't written unit tests for this feature because it is relatively small and self-contained, and because I'm currently designing a new architecture for our tests, which will include tests for this feature. I prefer to spend time on this 🙂

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -172,10 +172,18 @@ export function main({
   const command = commands[commandName];
 
   let warnAboutRunDashDash = false;
-  // we are using "yarn <script> -abc" or "yarn run <script> -abc", we want -abc to be script options, not yarn options
-  if (command === commands.run || command === commands.create) {
+  // we are using "yarn <script> -abc", "yarn run <script> -abc", or "yarn node -abc", we want -abc
+  // to be script options, not yarn options
+  const PROXY_COMMANDS = new Set([`run`, `create`, `node`]);
+  if (PROXY_COMMANDS.has(commandName)) {
     if (endArgs.length === 0) {
-      endArgs = ['--', ...args.splice(1)];
+      // the "run" and "create" command take one argument that we want to parse as usual (the
+      // script/package name), hence the splice(1)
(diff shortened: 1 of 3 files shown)


src/cli/index.js
```diff
@@ -172,10 +172,17 @@ export function main({
   const command = commands[commandName];
 
   let warnAboutRunDashDash = false;
-  // we are using "yarn <script> -abc" or "yarn run <script> -abc", we want -abc to be script options, not yarn options
-  if (command === commands.run || command === commands.create) {
+  // we are using "yarn <script> -abc", "yarn run <script> -abc", or "yarn node -abc", we want -abc to
+  // be script options, not yarn options
+  if (command === commands.run || command === commands.create || command === commands.node) {
```
BYK: Looks like it is time to create `ARG_PROXY_COMMANDS = new Set('run', 'create', 'node');` or add a new field to all commands, `proxyArgs: boolean` to determine this behavior.

arcanis:
> This is essentially equivalent to having a script inside package.json making use of the $NODE environment variable we have set, right?

Yup! The difference being that it would give us more opportunities to do extra things on the environment if we want to (for example we could imagine logging the require calls to analyze which dependencies are not being used, or stuff like this).

colepatrickturner:
Could this be used to colocate specific node versions within packages?

arcanis:
I'm not sure I understand, can you be more specific? :D

colepatrickturner:
Ah sorry what I mean is that I’ve recently seen some modules include the “node” binary as a dependency. Im wondering if this could be used to run that directly. 

Cheers!

PR #5389, opened 2018-02-21
Improves "yarn workspaces info"

arcanis:
**Summary**

Running `yarn workspaces info` will now yield information to reconstruct the workspace dependency tree. This information can then be feed to other build systems.

**Test plan**

Updated the tests to include the new fields

### DIFF
--- a/src/cli/commands/workspaces.js
+++ b/src/cli/commands/workspaces.js
@@ -4,9 +4,11 @@ import type Config from '../../config.js';
 import {MessageError} from '../../errors.js';
 import type {Reporter} from '../../reporters/index.js';
 import buildSubCommands from './_build-sub-commands.js';
+import {DEPENDENCY_TYPES} from '../../constants.js';
 
 const invariant = require('invariant');
 const path = require('path');
+const semver = require('semver');
 
 export function hasWrapper(commander: Object, args: Array<string>): boolean {
   return true;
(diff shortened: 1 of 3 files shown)


src/cli/commands/workspaces.js
```diff
@@ -27,8 +29,30 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
   const publicData = {};
 
   for (const workspaceName of Object.keys(workspaces)) {
+    const {loc, manifest} = workspaces[workspaceName];
+
+    const workspaceDependencies = new Set();
+    const mismatchedWorkspaceDependencies = new Set();
+
+    for (const dependencyType of DEPENDENCY_TYPES) {
```
BYK: Isn't it better to just omit this before the loop?

src/cli/commands/workspaces.js
```diff
@@ -27,8 +29,30 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
...
+      if (dependencyType !== 'peerDependencies') {
+        for (const dependencyName of Object.keys(manifest[dependencyType] || {})) {
+          if (Object.prototype.hasOwnProperty.call(workspaces, dependencyName)) {
```
BYK: Don't we have a shortcut for `Object.prototype.hasOwnProperty.call` here? Do we _really_ need it?

src/cli/commands/workspaces.js
```diff
@@ -27,8 +29,30 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
...
+            const request = manifest[dependencyType][dependencyName];
```
BYK: `expectedRange`, `requestedRange`? Something that wouldn't confuse me with an actual HTTP request or the `request` module?

src/cli/commands/workspaces.js
arcanis: What do you mean? Duplicate the DEPENDENCY_TYPES array, but without peerDependencies?

src/cli/commands/workspaces.js
arcanis: I don't think we have, there's multiple matches for this function in the codebase.

BYK: Ah, we should make that happen. That said I'm not sure if we need to be so defensive about this. We should be able to use `Object.hasOwnProperty()` I guess? Or some variant of `Object.keys()` that only returns own properties?

src/cli/commands/workspaces.js
BYK: Yes. It seems wasteful and less readable this way to check for this specific value in every iteration.

src/cli/commands/workspaces.js
arcanis: `Object.keys` only returns own properties, but we're calling it on a different object from the one we're dereferencing. So for example, a malicious (or a bit weird) user could add a dependency called `__proto__` to one of its workspaces, and without this check we would then access `workspaces.__proto__`.

src/cli/commands/workspaces.js
arcanis: Yeah, maybe we could have a `DIRECT_DEPENDENCY_TYPES` array.

PR #5400, opened 2018-02-24
Add message for integritySystemParamsDontMatch

wabain:
**Summary**

As of the current stable version, yarn fails with an internal error when `yarn check --integrity` is run in an environment where the Node version is different from the one listed in the integrity file. This happens because a language key is missing:

```
$ node --version
v9.2.0

$ yarn check --integrity
yarn check v1.3.2
success Folder in sync.
✨  Done in 0.12s.

$ nvm use stable
Now using node v9.6.1 (npm v5.6.0)

$ yarn check --integrity
yarn check v1.3.2
error An unexpected error occurred: "Unknown language key integrityNodeDoesntMatch".
info If you think this is a bug, please open a bug report with the information provided in "/Users/wabain/code/yarn/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

The `integrityNodeDoesntMatch` error was just replaced with `integritySystemParamsDontMatch` in eab88b8268f15851b817fafe554b653a301a6747, but the error is otherwise the same on master (understandably, since the error only surfaces when running the check command, which is kind of tangential).

This PR adds the missing language key. It also tweaks the error message for when a language key is missing. In this context, "Unknown language key" read like something which could be a problem with some kind of manifest in the user's workspace.

**Test plan**

Added a test which runs "yarn check --integrity" after changing the `systemParams` field in the integrity file.

### DIFF
--- a/__tests__/commands/check.js
+++ b/__tests__/commands/check.js
@@ -279,6 +279,28 @@ test.concurrent('--integrity should fail if integrity file have different linked
   );
 });
 
+test.concurrent('--integrity should fail if integrity file has different systemParams', async (): Promise<void> => {
+  await runInstall(
+    {},
+    path.join('..', 'check', 'integrity-lock-check'),
+    async (config, reporter, install, getStdout): Promise<void> => {
+      const integrityFilePath = path.join(config.cwd, 'node_modules', '.yarn-integrity');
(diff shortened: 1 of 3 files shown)


arcanis:
Looks good. Thank you!

PR #5410, opened 2018-02-26
test(info): Fix failing `yarn info <package> readme` tests

BYK:
**Summary**

`yarn info` tests were broken due to some registry changes. This patch updates the tests to use an
older package, left-pad to get master green.

**Test plan**

Fixes tests. All builds should pass.


cpojer:
Always bet on left-pad.

Daniel15:
Thanks! I'll cherry-pick this on top of v1.5.0 so we can actually finish releasing that version 😛

Daniel15:
Also, it doesn't seem ideal that a test relies on actual production data rather than a mock data fixture... Could we change that?

PR #5411, opened 2018-02-26
yarn link: Use "package" consistently

perlun:
Previously, we are using the word "module" and "package" somewhat interchangeably.

However, [the web page for `yarn link`](https://yarnpkg.com/lang/en/docs/cli/link/) talks about linking a "package" consistently.  The same goes for [`yarn unlink`](https://yarnpkg.com/en/docs/cli/unlink); it doesn't even mention the word module.

I believe we should change these strings in accordance to the naming already in use at the web page; the way it is now is a bit confusing and inconsistent to our users.


arcanis:
Thanks 🙂

Issue #5407, opened 2018-02-25
Subdirectory's .gitignore ignores root directory files when packaging project


padarom:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When packaging a project via yarn (for publishing, or just using `yarn pack`), .gitignore globs from subdirectories extend to the root directory.

The file `logs/.gitignore` below contains:
```gitignore
*
!.gitignore
```
For git, this means to ignore every file _within the `logs` directory_, except `.gitignore`. Yarn interprets it as _ignore every file_ (except the ones that are always included). As such, the `src` directory is not included in the package:

```
λ tar -tvf yarn-issue-5407-v1.0.0.tgz
drwxr-xr-x  0 0      0           0 25 Feb 16:29 package
-rw-r--r--  0 0      0         142 23 Jan 18:24 package/README.md
drwxr-xr-x  0 0      0           0 24 Feb 20:47 package/logs
-rw-r--r--  0 0      0        1755 24 Feb 23:13 package/package.json
-rw-r--r--  0 0      0          14 25 Jan 17:08 package/logs/.gitignore
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Clone this repository: http://github.com/padarom/yarn-issue-5407
2. Run `yarn pack`
3. Run `tar -tvf yarn-issue-5407-v1.0.0.tgz`

**What is the expected behavior?**
```
λ tar -tvf yarn-issue-5407-v1.0.0.tgz
drwxr-xr-x  0 0      0           0 25 Feb 16:29 package
-rw-r--r--  0 0      0        1877 24 Feb 20:11 package/.gitignore
-rw-r--r--  0 0      0         142 23 Jan 18:24 package/README.md
drwxr-xr-x  0 0      0           0 24 Feb 20:47 package/logs
-rw-r--r--  0 0      0        1755 24 Feb 23:13 package/package.json
drwxr-xr-x  0 0      0           0 24 Feb 20:11 package/src
-rw-r--r--  0 0      0          14 25 Jan 17:08 package/logs/.gitignore
-rw-r--r--  0 0      0           0 24 Feb 20:11 package/src/index.js
-rw-r--r--  0 0      0           0 24 Feb 20:11 package/src/a.js
-rw-r--r--  0 0      0           0 24 Feb 20:11 package/src/b.js
```

**Please mention your node.js, yarn and operating system version.**
Node: v9.4.0
Yarn: 1.5.0
[...]

Issue #2603, opened 2017-01-31
.*ignore rules in subdirectories erroneously apply to parent directories


cspotcode:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

`.*ignore` rules specified within subdirectories apply to *all* directories, whereas they should only apply to the directory and subdirectories of where they were declared.

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
#!/usr/bin/env bash
git init ; yarn init -y
mkdir dir
# Tell git, npm, and yarn to ignore /dir/foo
echo "foo" > dir/.gitignore
git add dir/.gitignore ; git commit -m "initial commit"

# should ignore this file
touch dir/foo 
# should *not* ignore this file
touch foo

# Create NPM and yarn packages to that we can see what gets ignored
npm pack
yarn pack --filename yarn-pack.tgz

echo GIT
# git sees "foo" and ignores "dir/foo"
git status

echo NPM
# npm does the same
tar -tf *-1.0.0.tgz

echo YARN
# Notice that yarn ignored both "foo" and "dir/foo"
tar -tf yarn-pack.tgz
```

**What is the expected behavior?**

I expect yarn's `.*ignore` behavior to match npm and git.

**Please mention your node.js, yarn and operating system version.**

node v6.4.0
yarn v0.19.1
macOS Sierra v10.12.3

PR #5414, opened 2018-02-26
Restrict .gitignore to their own subdirectories

padarom:
**Summary**
Fixed the bug from #5407 and #2603.

The function `matchesFilter` returned `true` if the file's relative path or its basename matched the given regular expression. I adjusted this to only test the basename for a match if the `base` option is not set (or set to `.`).

**Test plan**
I added the test `pack should include files only ignored in other directories`.

Two tests were failing on my local build, but they were also failing before making the changes. All other tests were green.


arcanis:
Thanks!

PR #4809, opened 2017-10-30
Allow insecure URLs with warnings

sth:
This makes the handling of insecure repository URLs more consistent, according to issue #4307. 

Some insecure URLs were allowed (plain HTTP) while others did raise different kinds of `SecurityError`s, With this pull requests all insecure URLs report a warning, but don't raise hard errors. See issue #4307 for details.

Issue #4307, opened 2017-09-05
util/git.js:secureGitUrl() treats insecure URLs inconsistently


sth:
**Setup**
Original idea of the function [secureGitUrl()][1] function: Ensure access to the repository is secure, by either specifying a git hash or by using HTTPS. When an unencrypted URL is specified, check if we can use a equivalent HTTPS URL instead. Otherwise raise a `SecurityError`.

**Inconsistency**
In reality the function allows plain HTTP repositories without a hash. The only additional check is that the repository has to exist (this check doesn't really add much):
https://github.com/yarnpkg/yarn/blob/67dfc7af80bad0034a1f23e9ef4da8c696e134c7/src/util/git.js#L143-L154

With this plain HTTP being allowed it doesn't make much sense that other insecure URLs like `git:` are rejected.

**Possible Solutions**
- Raise `SecurityError` on `http:` URLs again. This would be a breaking change and I don't really think this is viable. HTTP URLs have been allowed [since v0.20.0](https://github.com/yarnpkg/yarn/commit/8231aecd41a2fa6a01bfff757358aebe1a72d2f5).
- Remove the other `SecurityError`s. If plain HTTP is good enough, other unencrypted URLs should also work.
- Don't raise errors but issue warnings for HTTP and other insecure URLs.

**Addendum: Inaccessible repositories**
The function also checks in some cases if the given repository exists and is accessible and raises a `SecurityError` if not. This isn't helpful since a missing/inaccessible repository isn't a security problem. Downloading the missing repository will fail later anyway, with a more useful message.

 [1]: https://github.com/yarnpkg/yarn/blob/de7cefc3120a7246c507616a17517b45452d6ff3/src/util/git.js#L128

sth:
dda2be0 would implement warnings instead of raising errors and removes the unnecessary `reposExists()` checks.

BYK:
@sth - on which branch this commit is on? Can you submit a PR?

PR #5424, opened 2018-02-28
Make secureGitUrl() warn for insecure Urls

sth:
This is an updated version of #4809. It is rebased to current master and has some new tests added that verify the behavior of the modified `secureGitUrl()` function.

**Summary**
`secureGitUrl()` tries to make sure that we use secure URLs. It currently behaves inconsistently, like allowing plain http download without any warnings or raising `SecurityError` if a repository doesn't exist. See #4307 for details.

Main changes in behavior with this pull request:
- Insecure `http:` and `git:` URLs are handled consistently:
    - Plain `http:` URLs without a commit hash now report a warning. Previously they were silently accepted.
    - Plain `git:` URLs without a commit hash now no longer raise a SecurityError. Instead they report a warning.
- Non-existing or otherwise inaccessible repositories now no longer raise `SecurityError`. Previously this happend in some cases, leading to very confusing error messages.

For more context/rational see #4307 and the test cases in this pull request.


arcanis:
A bit late, but this looks good! Thanks @sth! 🙂

Issue #5429, opened 2018-02-28
Private scoped repositories authentication issue (different dependency url from tarball url)


karl-run:
**What is the current behavior?**

Using a scoped private repository, in this case bintray.com, yarn will fail to authenticate (HTTP 401).

**If the current behavior is a bug, please provide the steps to reproduce.**
Given a `.npmrc` file:
```
@myprivatescope:registry=https://api.bintray.com/npm/mycompany/npm/
//api.bintray.com/npm/mycompany/npm/:_authToken="blablaObscurifiedBearerToken"
//api.bintray.com/npm/mycompany/npm/:username=mycompany
//api.bintray.com/npm/mycompany/npm/:email=email@example.com
//api.bintray.com/npm/mycompany/npm/:always-auth=true
```

Adding a privately scoped package with yarn:
`yarn add @myprivatescope/cool-internal-module`

1. Yarn loads the `.npmrc` file correctly.
2. Yarn successfully makes the initial request:
```
verbose 1.73 Request "https://api.bintray.com/npm/mycompany/npm/@myprivatescope%2fcool-internal-module"
finished with status code 200.
```
This request request contains the following `versions/1.0.0/dist` object:

```
"dist": {
  "tarball": "https://dl.bintray.com//mycompany/npm/@myprivatescope/page-header/-/@myprivatescope/cool-internal-module-1.0.0.tgz",
},
```

Notice that the host on the initial request is `api.bintray.com` and the host on the tarball is `dl.bintray.com`.

That means, when in [`npm-registry.js:174`](https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L174), this if statement:
`if (requestParts.host === registryParts.host && requestParts.path.startsWith(registryParts.path)) {`,
more specifically this conditional: `requestParts.host === registryParts.host`, will be false, because it's comparing `requestParts.host` (api.bintray.com) against `registryParts.host` (dl.bintray.com), which obviously are different.

**What is the expected behavior?**

It should authenticate correctly.

**Please mention your node.js, yarn and operating system version.**
```
$ node --version
v8.9.4
$ yarn --version
1.5.0
$ uname -a
MINGW64_NT-10.0 ILDBERG 2.9.0(0.318/5/3) 2018-01-12 23:37 x86_64 Msys
```
Windows 10 Pro, 1709

karl-run:
Obviously expanding the `.npmrc` file to:
```diff
[...]

PR #5430, opened 2018-02-28
Only compare root domains to determine authentication

karl-run:
Possible fix for #5429, if comparing root domains is sufficient.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -85,6 +85,10 @@ function urlParts(requestUrl: string): UrlParts {
   return {host, path};
 }
 
+function rootDomain(host: string): string {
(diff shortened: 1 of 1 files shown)


arcanis:
I'm not sure I like this ... we can imagine a registry service using subdomains to split their users instances :/

karl-run:
I do agree, even though that sounds like a very rare case. Any idea on how npm handles this?

karl-run:
Also this implementation won't work on domains such as co.uk.

Anyway, would it be sufficient to first check for the specific host, and then also check on the root domain? That would work in both cases.

karl-run:
On further inspection this is garbage.

PR #5436, opened 2018-03-01
Makes CircleCI run the pkg-tests testsuite

arcanis:
**Summary**

This PR makes CircleCI run a test instance for the pkg-tests testsuite. It's currently only run on a Node 8 instance, but this can be improved later on.

**Test plan**

CircleCI


arcanis:
Would be a worthy goal :) There's still some things that will have to be improved (like git support, etc), but that would be great.

Issue #5421, opened 2018-02-27
Workspaces nohoist not working with --link-duplicates


k15a:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Yarn doesn't install dependencies.

**If the current behavior is a bug, please provide the steps to reproduce.**

I've got a bug in our private repo and tried to reproduce it in a new repo. I got a similar result but not exactly the same. The outcome was the same that yarn wasn't able to install the dependencies.
I think the root of the problem is the same.

In both cases, I am using the new nohoist feature and the `--link-duplicates` flag.

[Reproduction](https://github.com/k15a/yarn-nohoist-link-duplicates) / [Log](https://github.com/k15a/yarn-nohoist-link-duplicates/blob/master/yarn-error.log): Just clone the repo and run `yarn install --link-duplicates`

The inital bug I tried to reproduce is:
```
Error: ENOENT: no such file or directory, lstat '/Users/konstantin/Development/frontend/projects/atoms/node_modules/@monorepo/styling/node_modules/jest'
```

**What is the expected behavior?**

Yarn should link all duplicate dependencies in a workspace with nohoist enabled.

**Please mention your node.js, yarn and operating system version.**

Node: 8.6.0
Yarn: 1.5.1
OS: MacOS High Sierra 10.13.2

k15a:
I've posted the error log of the inital bug here: https://gist.github.com/k15a/6ca1a7582b612cb1bbab2883e54862a6

connectdotz:
it's indeed a problem with the combination of nohoist + link-duplicates. I can reproduce this error. Will dig deeper to see how we can fix it.

PR #5442, opened 2018-03-03
fixed linkDuplicates for symlinks

connectdotz:
**Summary**
addressing #5421.

**src**:
1. package-linker: fixed a bug that should have skipped hardlinking the symlinks. While this is not a nohoist specific bug, the fact that linked modules can live in multiple locations surfaced this bug. 
1. Config: separated error from warning in extracting workspace config so we can turn off the nohoist via `workspaces-nohoist-experimental` independent of nohoist config in package.json. 

**\_\_tests\_\_**:
1. _helpers: added ability in runner to disable nohoist in Config. This enable tests easily switch between hoist/nohoist without changing the fixture's package.json. 
1. tests and fixtures: added related tests in `__tests__/commands/install/integration-deduping.js` and corresponding fixtures.

**Test plan**

1. all existing and new tests should pass
1. the [sample project](https://github.com/k15a/yarn-nohoist-link-duplicates) in #5421 should pass.


connectdotz:
the test-macos-node6 and test-macos-node8 have been stuck over 20 hours... looks like the other PR have suffered the similar fate for 4 days... can someone from the yarn core team take a look of the task queue...?

connectdotz:
good, tests finally unstuck. @bestander @rally25rs, I didn't see reviewer assigned for this, just wondering if we can get some traction

bestander:
Love it! Thank you, @connectdotz.

dannycochran:
@arcanis is there a planned 1.5.2 release? If so, can we include this change? It would make a world of difference for folks using workspaces.

arcanis:
It should happen soonish but I can't give you an exact date. In the meantime, the nightlies are the recommended way to get features already merged in the trunk:

https://yarnpkg.com/en/docs/nightly

dannycochran:
@arcanis -- Thank you for the response. I'm aware of the nightly build option but it's hard to prescribe that solution for a team of engineers.

It would be nice if there could be a release cadence for yarn, maybe every 2 weeks? I believe that's what npm does. I say this just as a friendly suggestion and understanding that you all are probably quite busy :-)

PR #5470, opened 2018-03-07
fix(linking): Dont integrity check build artifacts

### LINKED ISSUE #932: Native packages rebuilt every time

Bug.

It seems that all native packages are rebuilt every time yarn is asked to either add a new package or just install the currently locked.

1. Add some native packages.

```
yarn add leveldown bcrypt
```
1. Run yarn again and observe that both of the packages will be rebuilt for no reason.

```
yarn
```

The same happens when adding a completely unrelated packages which, as far as I can tell, cannot affect the native packages in any way.

```
yarn add co
```

Native packages should not be rebuilt if there's no reason to do that. Note that #248 seems pretty similar.

Node.js 6.7.0
Yarn 0.15.1
Ubuntu 12.04

rally25rs:
fixes #932 

**Summary**

If an install script modified a file that existed in the package, then yarn would view the file as stale on every command, mark the package as "fresh", and re-run the install scripts.

With this change, known build artifacts from the integrity file are checked, and if this package file is a
build artifact, it's size and timestamp are not checked.

In addition it was discovered that on Linux with node 8's `fs.copyFile()` timestamps are not preserved (they are on Win and OSX). This would cause yarn to always re-copy all files for all packages when doing a `yarn add` or `yarn remove`. Yarn will now detect this by checking the first file copy's timestamps, and if the first file is wrong then timestamps will be corrected using `fs.futimes()` for all file copies.

**Test plan**

Added test in `__tests__/commands/add.js`


bestander:
I think it is consistent with Yarn's promises.
Install scripts are evil from package manager's point of view, they can do anything outside of its control anyway. 
So probably we could instruct Yarn to ignore the artifacts and not delete them.

bestander:
Are test failures related?

rally25rs:
The CircleCI test is related. I'll take a look at that. 🤔 
Strange that they pass on Appveyor (the one that is failing is an unrelated timeout that we see often).

BYK:
@rally25rs - I think the problem stems from here: https://github.com/yarnpkg/yarn/blob/5ef113f7db1201f48e22498374023ebe1e61b069/src/cli/commands/add.js#L152-L164

The `bailout` function in `install.js` returns with `true` when the integrity has not changed whereas the one from `add` always returns `false`, no matter what. I think this triggers the regular lifecycle events for all modules causing `postinstall` script to be run, thus changing the file output.

So it is not the copying, it is the postinstall script running and modifying it.

rally25rs:
Hmm, interesting. Then why the difference between node versions?

rally25rs:
@BYK so get this... on linux with node 8, when files are copied from the cache to node_modules, the timestamps are updated. yarn decides the files are different and marks the reference fresh.
This only seems to happen on linux node 8 though.

image

Look at the stat timestamps on the copy in node_modules... all 3 timestamps are changed to the Change Time.

This is the same commands run in node 6:

image

This is going to cause a lot of churn in yarn on linux with node 8. It's going to re-copy every file every time. 😿

rally25rs:
_sigh_ it looks like calling `futimes` makes things fail on AppVeyor (windows).

> Error: EPERM: operation not permitted, futimes

how were we getting around this lack of `futimes` before? In master it is always called in older node versions https://github.com/yarnpkg/yarn/blob/master/src/util/fs.js#L69

rally25rs:
🎉 tests pass. The one failing on CircleCI node6 is a random timeout and not related. It passed in a couple of the previous runs.

It seems that on OSX and Linux, you can open a file with 'read' access and still edit the timestamps with `fs.futimes()` but on windows you have to open the file with 'write' access... good times... 🙄

BYK:
>Our fallback when fs.copyFile doesn't exist does a fs.futimes call to fix the timestamps. I think we might need to add that to the modern fs.copyFile implementation too.

I tried this and it caused a significant performance drop due to the extra file I/O. `fs.copyFile` was supposed to keep the timestamps with the original file. I really wouldn't want to do this :(

BYK:
Seems like this is a bug in Node itself and explains why I haven't seen the issue on Windows: https://github.com/nodejs/node/issues/15793#issuecomment-351230606 (and you haven't seen it on Windows and OS X).

Maybe we can limit the extra operation to Linux only which is the fastest anyway so the performance hit won't be too much of an issue? Alternatively, we may look at the creation time instead of modification time which seems to be consistent across platforms? (need to verify this last part before acting)

rally25rs:
I had a thought to just copy 1 file, and after that first copy check the timestamps. If they changed, set a flag to do the `futimes` on all subsequent copies.

-----

however, this is still broken is some weird way. Somehow when copying files to the cache, after the copy completes the file still doesn't exist. This seems to happen if you have a `file:` dependency to a directory that is also a git repo.

```
verbose 2.163 Error: EACCES: permission denied, open '/Users/jvalore/Library/Caches/Yarn/v1/npm-cxcore-common-ui-0.1.2-9d4812bc-23b7-4d80-8253-121c0da71c33-1520603428055/.git/objects/03/9813c7ac8f77e7e8a849f7002cc37b0a75d6aa'
error An unexpected error occurred: "EACCES: permission denied, open '/Users/jvalore/Library/Caches/Yarn/v1/npm-cxcore-common-ui-0.1.2-9d4812bc-23b7-4d80-8253-121c0da71c33-1520603428055/.git/objects/03/9813c7ac8f77e7e8a849f7002cc37b0a75d6aa'".
info If you think this is a bug, please open a bug report with the information provided in "/Users/jvalore/Projects/cxcore-system-ui/yarn-error.log".
```

This error comes from trying to `fs.open` a file _after_ `fs.copyFile` finished.

-----

Eventually I think we need to just stop using timestamps as part of the check, but I'm not sure what else to do. Getting a checksum or hash for every file is probably much slower.

rally25rs:
@byk I did a bunch more work on this and fixed it up a bit. There are a lot of edge cases (read my code comments) so the code looks ugly, but...

On the first file copy, check the timestamp to see if it's wrong. If it is already corect, then disable the calls to `futimes`. Because of the way file copy batching works (multiple in parallel), it actually does the check about 10 times, but at least its not every file.

I discovered that my error above was because I was opening the file in write/append mode to call futimes, because you have to on Windows. However if the file is read-only (as is the case with .git objects files) then that fails. On OSX you can call futimes on a file that is open in read-mode, so I try to re-open the file in read mode.

Performance hit... quantifiable but not bad. The first run (`yarndev`) is this branch. This is in a project with >1,400 packages totaling >25,000 files, so it is quite a large project. Also this is node 8.10 so `fileCopy` is in use, and on OSX so `futimes` is not called for every file.

```
~/Projects/cxcore-system-ui (master) 🐒   rm -rf node_modules/

~/Projects/cxcore-system-ui (master) 🐒   yarndev
yarn install v1.6.0-0
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
[5/5] 📃  Building fresh packages...
✨  Done in 18.20s.

~/Projects/cxcore-system-ui (master) 🐒   rm -rf node_modules/

~/Projects/cxcore-system-ui (master) 🐒   yarn
yarn install v1.5.1
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
[5/5] 📃  Building fresh packages...
✨  Done in 17.72s.
```

Performance on Ubuntu with node 8, where `futimes` is needed. I didn't have the same project to test, so I just used `jest` because it is pretty big... (518 packages, 7,800 files)

```
$ yarn install
this branch: 5.3s
yarn 1.5.1: 4.8s
```

so noticeably slower, but subsequent adds won't re-copy files...

```
$ yarn add {another package}
this branch: 3.6s
yarn 1.5.1: 5.1s
```

So you take a perf hit on initial install to save time on later adds.

whatcha think? 👍 / 👎

src/util/fs.js
```diff
@@ -58,15 +60,23 @@ const futimes: (fd: number, atime: number, mtime: number) => Promise<void> = pro
 const copyFile: (src: string, dest: string, flags: number, data: CopyFileAction) => Promise<void> = fs.copyFile
   ? // Don't use `promisify` to avoid passing  the last, argument `data`, to the native method
     (src, dest, flags, data) =>
-      new Promise((resolve, reject) => fs.copyFile(src, dest, flags, err => (err ? reject(err) : resolve(err))))
+      new Promise((resolve, reject) =>
+        fs.copyFile(src, dest, flags, err => {
+          if (err) {
+            reject(err);
+          } else {
+            fixTimes(0, dest, data).then(() => resolve(err)).catch(ex => reject(ex));
```
arcanis: Is the `err` in `resolve(err)` is a mistake?

arcanis: I'm not sure I understand why the first parameter is `0`. The fd `0` is stdin, right?

src/util/fs.js
```diff
@@ -162,6 +172,56 @@ export const fileDatesEqual = (a: Date, b: Date) => {
...
 
+// This ensured the timestamps are preserved from the file in the cache to the file copied to node_modules.
+// These timestamps are checked to see if files have changed and need recopied, so preserving them is important.
+// There are some weird cases here:
+// * On linux, fs.copyFile does not preserve timestamps, but does on OSX and Win.
+// * On windows, you must open a file with write permissions to call `fs.futimes`.
+//   On OSX you can open with read permissions and still call `fs.futimes`.
+//   We first try to open the file in write mode because that works across all OSs.
+//   However if the file is read-only (not even the owner has write perms) then that will fail.
+//   We can try to reopen in read mode, which will work on OSX.
```
arcanis: It seems a bit convoluted. Would it make sense to use a third-party package for this?

src/util/fs.js
rally25rs: It was `0` just because it's `falsy`. In the case of not using `copyFile` we already have a file open that we can use, instead of having to re-open the file. If the `fd` is falsey, then `fixTimes` opens the file. I was going to pass `undefined` but then Flow gets mad. I'll see if I can make it more understandable...

src/util/fs.js
rally25rs: is there a 3rd party package that will handle this? I agree that it's ugly. Maybe I can move it to it's own package.

src/util/fs.js
rally25rs: Maybe. I just preserved what the original code did:

```
new Promise((resolve, reject) => fs.copyFile(src, dest, flags, err => (err ? reject(err) : resolve(err))))
```

The code I was replacing would `resolve(err)` too. I didn't want to make the assumption that it wasn't intentional. :shrug:

arcanis: That's ... weird. I think we can just fix this and check that everything is still green.

arcanis: I think you should be able to use `null`, then in the declaration of `fixTime` just use `fd: ?number` (instead of `: number`). Flow should then accept `=== null` without complaining.

src/util/fs.js
arcanis: Let's keep it there for now 👍

rally25rs:
@arcanis I did a bit of refactoring and moved all the stuff in `fs.js` that had OS or Node version specific stuff in it into a new `fs-normalized.js` file. This at least puts all that stuff in one place instead of making `fs.js` so messy, and should put it in a better position for future removal. I opened #5554 to capture some additional work that could be done to replace this file with a different library. I just don't have time to actually test out a couple libraries across 3 OS's right now.

Hopefully this at least improves the situation...

arcanis:
LGTM, I'll fix the conflict monday 👍

src/util/fs-normalized.js
```diff
@@ -0,0 +1,170 @@
...
+  if (doOpen) {
+    try {
+      openfd = await open(dest, 'a', data.mode);
+    } catch (er) {
+      // file is likely read-only
+      try {
+        openfd = await open(dest, 'r', data.mode);
+      } catch (err) {
+        // We can't even open this file for reading.
+        return;
```
james-rabbit: This is catching `er` and `err` without logging or throwing anything here.
Issues with this may be difficult to debug.

src/util/fs-normalized.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  try {
+    if (openfd) {
+      await futimes(openfd, data.atime, data.mtime);
+    }
+  } catch (er) {
+    // If `futimes` throws an exception, we probably have a case of a read-only file on Windows.
+    // In this case we can just return. The incorrect timestamp will just cause that file to be recopied
```
james-rabbit: Can we verify that the file is read-only and throw or log this `er` if that is not the case?

src/util/fs-normalized.js
rally25rs: There shouldn't be any issues if it fails. We just won't correct the timestamps (because we can't) and yarn will behave the same way it does now. The comment at the top of the function tries to explain this, but maybe I didn't do a great job of explaining it.

src/util/fs-normalized.js
rally25rs: There is no logger passed down to the `fs` modules, so I didn't have any easy way of logging these issues. I was trying to avoid even bigger changes by passing the reporter down another level. I also didn't want `fixTimes()` to ever cause a yarn install to fail. If we can't fix the time stamps, it will effect performance a bit, but won't really break anything. It'll just work like it does currently where sometimes modules might get rebuilt every time (#932 ) but that's better than failing the install. If I can bubble up an error without failing the install, I'll try to add a line to the verbose log. Otherwise I'd rather not restructure the code any more than I already have.

rally25rs:
@BYK could you re-review so we can unblock merging? thanks!

bestander:
@rally25rs, could you fix the merge conflict?
I'll ping the team in chat.

rally25rs:
@bestander yep, working on it now...

src/util/fs-normalized.js
james-rabbit: It is not critical, but if this code is not working on some machine it will look like a fairly big slowdown without a good way to debug it.

bestander:
So we see 3% perf reduction for first installs but we get a significant win on consequent installs which I think is a more common case.
Let's merge it

src/util/fs-normalized.js
bestander: @rally25rs, I agree with @james-rabbit, it makes sense to put verbose logs that `open -x` failed for the future when we might be debugging why install on one OS is slower than the other.
Would you add it?

bestander:
@rally25rs, this is a huge win! Thank you.
Let's merge, just one nit.

arcanis:
🎉

Haroenv:
I just waited four times for already-installed native modules today, so ready for this PR !

PR #5478, opened 2018-03-07
fix(mutex): Exit with a descriptive error when mutex port is in use

BYK:
**Summary**

Currently, if the network mutex port is used by another server which responds with something that
Yarn doesn't expect, it crashes with a non-descript error like a "JSON parse error". This patch
exits gracefully with a descriptive error message when Yarn cannot communicate properly with the
assumed Yarn mutex lead.

**Test plan**

Added new integration test.


__tests__/integration.js
```diff
@@ -151,6 +154,40 @@ test('--mutex network', async () => {
...
+  let mutexError;
+  try {
+    await runYarn(['--mutex', `network:${port}`, 'run', 'test', '100'], {cwd});
+  } catch (error) {
+    mutexError = error;
+  } finally {
+    server.close();
+  }
+
+  expect(mutexError).toBeDefined();
```
BYK: I guess this line is now obsolete with the `invariant` statement below? (added invariant later to silence flow)

src/cli/index.js
```diff
@@ -378,8 +378,16 @@ export function main({
           });
 
           response.on('end', () => {
-            const {cwd, pid} = JSON.parse(Buffer.concat(buffers).toString());
-            reporter.warn(reporter.lang('waitingNamedInstance', pid, cwd));
+            try {
+              const {cwd, pid} = JSON.parse(Buffer.concat(buffers).toString());
+              reporter.warn(reporter.lang('waitingNamedInstance', pid, cwd));
+            } catch (error) {
+              console.error(
```
BYK: I should probably use `reporter.error` here, right?

__tests__/integration.js
arcanis: Yeah but I'd still keep it to make it explicit that we want to test it.

Issue #5457, opened 2018-03-06
`yarn run` has changed behaviour, and generates a `yarn-error.log` file for exit code 1


vdh:
Recently, `yarn run` has changed behaviour and has started outputting an error message and generating a `yarn-error.log` file when a non-zero exit code occurs. Previously it would just exit with code 1. This was useful for running tests or linters, where an exit code of 1 is as expected outcome, and helps to automatically stop a build system when tests/linting fails.

**Do you want to request a *feature* or report a *bug*?**
Report a bug.

**What is the current behavior?**
Running scripts from `package.json` that return exit code 1 will output an error message and generate a `yarn-error.log` file, e.g.:
```
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c eslint --ext js,jsx --max-warnings 0 .
Directory: /Users/tim/Documents/foo
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "/Users/tim/Documents/foo/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Run a script that exits with code 1, e.g. ESLint:
```json
{
  "scripts": {
    "lint": "eslint --ext js,jsx --max-warnings 0 ."
  }
}
```

**What is the expected behavior?**
Previously, `yarn run` would just exit with code 1.

1. I can't find the release information or documentation about this change in behaviour. Does anyone know what version introduced this change?
2. Is there a way to flag that exit code 1 for a script *is* an expected outcome?

**Please mention your node.js, yarn and operating system version.**
node 9.6.1, yarn 1.5.1, macOS 10.13.3

sth:
See also #5451

Ermish:
I'm getting this same issue. 

node 9.8.0, yarn 1.5.1, macOS 10.13.3

2gnc:
the same problem... node 8.10.0 yarn 1.5.1 Ubuntu 16.04

Biuni:
Same issue. 
- **Node** v8.9.1
- **Yarn** v1.5.1
- **Windows** v10.0.16299.309

mjstelly:
Same:
Yarn version:  1.5.1
Node version: 9.2.1
Platform:  Darwin x64

```
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c cucumber-js
```

[...]

Issue #5451, opened 2018-03-05
ProcessTermError instance isn't an instance of ProcessTermError


jeremejevs:
**Do you want to request a *feature* or report a *bug*?**

This is a bug.

**What is the current behavior?**

Running `yarn -s run false` with the following `package.json`:

```json
{ "scripts": { "false": "false" } }
```

Results in the following output:

```
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c false
Directory: /my/cwd
Output:
".
```

**What is the expected behavior?**

The expected output:

```
error Command failed with exit code 1.
```

**Please mention your node.js, yarn and operating system version.**

Node.js: 9.7.1
Yarn: 1.5.1
OS: Linux 4.15.6

**Extra info.**

I'm using `--silent` for brevity. I've inspected the compiled sources, inserted some logs around [this catch](https://github.com/yarnpkg/yarn/blob/615ee94b0d88dfbd728d03114aa031cd11e838e1/src/util/execute-lifecycle-script.js#L291-L299), and narrowed it down to this fact:

```js
new ProcessTermError('whatever') instanceof ProcessTermError === false
```

Appears to be some sort of an issue with Babel/Webpack configuration, and I'm assuming that other uses of `instanceof` are affected by this too.

jeremejevs:
Fixed by #5497.

PR #5497, opened 2018-03-10
Fix yarn behaviour when scripts are failing

sth:
**Summary**
Fix for #5451, #5457.

The test case in the first commit fails for pre-node5 builds, in the way described in the links issues. The problem are classes deriving from `Error`, which are problematic in ES5 ([babel][babel-Error]). Adding `babel-plugin-transform-builtin-extend` fixes the issue in this case.

**Why now?**
The error surfaced because the [current yarn build published on npm][v1.5.1] is a pre-node5 build while for example [v1.3.1][v1.3.1] seems to have been created by the more modern build configuration. This can be seen by checking if code like "class A extends B" got compiled to ES5 or not.

 [babel-Error]: https://babeljs.io/docs/usage/caveats/#classes
 [v1.3.1]: https://registry.npmjs.org/yarn/-/yarn-1.3.1.tgz
 [v1.5.1]: https://registry.npmjs.org/yarn/-/yarn-1.5.1.tgz

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -19,7 +19,8 @@
      "plugins": [
        ["array-includes"],
        ["transform-inline-imports-commonjs"],
-       ["transform-runtime", { "polyfill": true, "regenerator": false }]
+       ["transform-runtime", { "polyfill": true, "regenerator": false }],
+       ["transform-builtin-extend", { "globals": ["Error"] }]
      ]
    },
    "test": {
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -398,6 +398,30 @@ test('yarn run <script> <strings that need escaping>', async () => {
   expect(stdout.toString().trim()).toEqual(JSON.stringify(trickyStrings));
 });
 
+test('yarn run <failing script>', async () => {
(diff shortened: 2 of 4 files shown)


arcanis:
> The error surfaced because the current yarn build published on npm is a pre-node5 build while for example v1.3.1 seems to have been created by the more modern build configuration.

That looks weird :/ cc @byk @Daniel15 did we change something recently?

rally25rs:
wow, nice find. 👏  I never would have guessed that `err instanceof MessageError` wouldn't work. I'm curious what happened in the build process, but either way, it looks like this babel plugin fixes it.

Daniel15:
> did we change something recently?

I'm not sure... I don't think we've changed the build config recently though.

Daniel15:
It looks like we haven't changed the npm publishing code in quite a while: https://github.com/yarnpkg/yarn/blame/master/scripts/update-npm.sh. Maybe the build config changed, but I'm not sure.

gyammy:
Still don't get how to fix the problem sorry. 
How and where should I add `babel-plugin-transform-builtin-extend`?

PR #5498, opened 2018-03-10
fix(cli): adjust colors for unstable version updates

### LINKED ISSUE #4942: yarn-outdated should not call minor updates backward-compatible when major is 0
I called `yarn outdated` in one of my projects and obtained this output:

screenshot_2017-11-16_16-53-20

It is claiming that upgrading `typedoc` would be a minor update with backward-compatible features, but the package is still in major version 0. By the definition of semantic versioning, a zero for the major version is used for less stable APIs, and bumping the `y` in `0.y.z` can be a breaking change. I believe that in these cases, the tool should instead label the package in red.

simonkberg:
**Summary**

Fixes #4942.

The yarn CLI uses colors to indicate backwards compatibility for version updates in the `yarn outdated` and `yarn upgrade-interactive` commands. These can be confusing and also misleading for updates to unstable package versions (0.x.y & 0.0.x), as it doesn't match either the semver spec nor how both yarn and npm handle updates to these versions using caret-ranges. 

This change special-cases the diff calculation, along with the color representation, for updates to *unstable* package versions, to more accurately indicate the level of backwards compatibility.

For `0.x.y` versions, increments to the *minor* version number (`y`) will be considered a *major* version change, and increments to the *patch* version number (`x`), will be considered a *minor* version change.

For `0.0.x` versions, increments to *any* version number will be considered a *major* version change.

This matches the behavior of how both yarn & npm[1] treats updates to packages defined with a caret-range, and anecdotally how most package authors treat updates in these version ranges. The semver *spec* is less specific regarding this, and states[2] that *any* version change to "Major version zero" versions could be breaking.

References:
[1] https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004
[2] https://semver.org/#spec-item-4

**Test plan**

Added unit tests.

**Example output for `yarn outdated`**

Before:
screenshot from 2018-03-10 18-37-55

After:
screenshot from 2018-03-10 18-38-18

### DIFF
--- a/__tests__/util/semver.js
+++ b/__tests__/util/semver.js
@@ -1,8 +1,7 @@
 /* @flow */
 
-import {satisfiesWithPrereleases} from '../../src/util/semver.js';
-
-const semver = require('semver');
+import semver from 'semver';
+import {satisfiesWithPrereleases, diffWithUnstable} from '../../src/util/semver.js';
 
 describe('satisfiesWithPrereleases', () => {
   it('matches the behavior of node-semver for non-prerelease versions', () => {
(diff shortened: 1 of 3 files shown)


arcanis:
Very nice, thanks!

BYK:
Shall we send a PR to `node-semver` to get this in?

PR #5499, opened 2018-03-10
yarn search

maticzav:
## Yarn search (Emma)

- This PR implements live package search in Yarn CLI.

Check `emma-cli` for usage example (https://github.com/maticzav/emma-cli)

### TODO:
 - [x] Trigger package install in the end (`yarn add`).
 - [x] First line is repeated two times (not being removed).

screen shot 2018-03-10 at 19 19 52


cpojer:
Thanks for adding this. It seems that it makes Yarn significantly larger – do you think there is any way we could align dependencies to reduce the size impact to something like 100k?

BYK:
The main dependency here I think is `algoliasearch`. The full version of it seems to be around 190kb and the lite version is about 117kb: https://github.com/algolia/algoliasearch-client-javascript/blob/develop/dist/algoliasearchLite.js

If we can only use that, I think we can ship this pretty efficiently. What do you all think?

BYK:
ping @maticzav

BYK:
pingx2 @maticzav

maticzav:
Hi 👋,
Sorry for the late reply. I've been extremely busy lately. I will check out the issue during the weekend and update the PR afterwards.

PR #5504, opened 2018-03-12
fix(config): reorder flags passed to workspace (#5496)

### LINKED ISSUE #5496: Flags in `.yarnrc` can cause `yarn workspace` commands to fail.
Report a bug.

Running a `yarn workspace` command while a flag is in the `.yarnrc` causes an error.

1. Create a repo with workspaces.
2. Create a `.yarnrc` with a flag included.
3. Run `yarn workspace WORKSPACE_NAME add PACKAGE_NAME`.

Here's a demo repo: https://github.com/wtgtybhertgeghgtwtg/yarnrc-workspace-flag
Run `yarn workspace package-one add react`.

The flag should be parsed and respected as it is for non-`workspace` commands.

```
Yarn version: 
  1.5.1

Node version: 
  9.8.0

Platform: 
  linux x64
```

rally25rs:
fixes #5496

**Summary**

If flags are specified in .yarnrc then they will be inserted between the
yarn command and the remaining parameters. In the case of workspace,
this means that if `--*.emoji false` is in .yarnrc then the args passed
to the workspace command will be `[yarn workspace --emoji false
workspace_name, workspace_command]` which causes the workspace command
to fail (it attempts to use `--emoji` as the workspace name and run
command `false`).

This change will take flags from the rawArgs array that yarn is aware of
and move them to the end of the params.

My first thought was to just ignore all `--*` flags in `.yarnrc` when running the `workspace` command. Since a child `yarn <command>` process is started, that child process would load its own `.yarnrc` and then use the flags correctly. However, flags like `--*.emoji false` should be used by both `yarn workspace` and by the `yarn <command>` child process.

The solution in this PR isn't perfect, but I think it will cover the majority of cases.

**Test Plan**

Added test to `__tests/commands/workspace.js`. This required initializing commander.js in that test suite, since `workspace` now uses it to determine which passed flags are yarn-related.

### DIFF
--- a/__tests__/commands/workspace.js
+++ b/__tests__/commands/workspace.js
@@ -8,11 +8,18 @@ import * as reporters from '../../src/reporters/index.js';
 import Config from '../../src/config.js';
 import path from 'path';
 import {NODE_BIN_PATH, YARN_BIN_PATH} from '../../src/constants';
+import commander from 'commander';
+import {boolify} from '../../src/util/conversion.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 90000;
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'workspace');
 const spawn: $FlowFixMe = require('../../src/util/child').spawn;
+
+// The normal test builder doesn't use commander.js for argument parsing,
+// but the workspace command requires it.
+commander.option('--emoji [bool]', 'enable emoji in output', boolify);
+commander.option('--verbose', 'enable verbose output', boolify);
 
 beforeEach(() => spawn.mockClear());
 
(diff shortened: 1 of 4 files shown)


arcanis:
This seems a bit convoluted. Wouldn't it be simpler to just change cli/index.js to push the command line arguments before the `workspace` command?

rally25rs:
@acranic I originally was trying not to add another special case to index.js, but also it would still need to figure out at what point to push the flags into the args array.

For the "add" command it's after the workspace command:

```
yarn workspace package_name add left-pad

-can be-

yarn workspace package_name add --emoji false left-pad
```

but for the `run` command it's after the command to be run

```
yarn workspace package_name run env --silent

-can not be-

yarn workspace package_name run --emoji false env --silent

-needs to be-

yarn workspace package_name run env --emoji false --silent
```

so either way you would have to have either multiple special cases, or start parsing the parameters individually to find the ones that are part of the command and ones that are flags.

I don't really like either approach. They all feel dirty to me. It would be nice to find a way to make commander.js parse this for us, but I don't think there is an easy way to do that either.

rally25rs:
closing this PR in favor of the much simpler #5534 (@arcanis )

PR #5534, opened 2018-03-17
fix(config): Don't pass .yarnrc generated flags to workspace child command

### LINKED ISSUE #5496: Flags in `.yarnrc` can cause `yarn workspace` commands to fail.

rally25rs:
fixes #5496

**Summary**

Previously, if flags were defined in .yarnrc then they would be inserted into the rawArgs array of
commander.js. The yarn workspace command would then attempt to pass these flags to the child process that it spawns. They were being passed out-of-order and causing the child command to fail to
properly parse it's command line args.

This change prevents those flags from being passed at all.
The yarn child process will read .yarnrc and inject the flags into it's own argument array anyway,
so passing them was just duplicating the flags.

**Test Plan**

No additional tests, existing tests pass.

### DIFF
--- a/__tests__/commands/workspace.js
+++ b/__tests__/commands/workspace.js
@@ -39,11 +39,11 @@ async function runWorkspace(
 }
 
 // The unit tests don't use commander.js for argument parsing.
-// `rawArgs` is normally passed by commander.js so we just simulate it in the tests.
+// `originalArgs` is normally passed by index.js so we just simulate it in the tests.
 
 test('workspace run command', (): Promise<void> => {
-  const rawArgs = ['/path/to/node', '/path/to/yarn', 'workspace', 'workspace-1', 'run', 'script'];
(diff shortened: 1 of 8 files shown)


arcanis:
That looks fine to me! Do you mind adding a new test to make sure we don't get regression later on?

Issue #5443, opened 2018-03-03
Do not write entire contents of yarn.lock to yarn-error.log


xPaw:
**Do you want to request a *feature* or report a *bug*?**

Feature?

**What is the current behavior?**
1. Run anything that produces an error, e.g. `yarn thiscommanddoesnotexist`
2. `yarn-error.log` is written with the entire contents of `yarn.lock` copied to it.

**What is the expected behavior?**
I'm not sure how useful it is to just copy both `package.json` and `yarn.lock` into the error log, especially in the javascript ecosystem where there are an insane amount of (sub-)dependencies.

**Please mention your node.js, yarn and operating system version.**
Node 9, Yarn 1.5.1, Windows 10.

hulkish:
+1

rally25rs:
What does it hurt having it there? I assume the intent is that you could just provide your error log when asking for assistance, instead of 3 files.

xPaw:
The problem I currently have is that yarn-error.log is created in a somewhat trigger-happy approach, and I haven't really seen many (if at all) issues where the lock file was relevant.

And it puts the stacktrace *below* the yarn.lock, so you have to scroll down for the actual relevant information.

hulkish:
I find it really noisy and unhelpful to put the yarn.lock into the error log, personally.

paul-soporan:
Closing as fixed in v2, where we don't generate `yarn-error.log` files anymore.

https://yarnpkg.com/getting-started/migration

PR #5564, opened 2018-03-23
Write stacktrace in yarn-error.log sooner

xPaw:
**Summary**

In my opinion lockfile shouldn't be written at all into the error file (#5443) but this should be a simpler solution to that.

The stack trace is more important than the entire contents of the lock file, so I've moved it above it. As it's rather unnecessary to scroll to the end of the file to see the actual error.

**Test plan**

Run any command that produces an error (e.g. non existing command) to produce an error file.


arcanis:
Yep, I agree. Thanks!

Issue #5250, opened 2018-01-18
Resolutions interfere with the integrity check for `yarn install`


edyburn:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using `resolutions`, `yarn install` doesn't bail out early if the installation is already up-to-date.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create a package, and specify a resolution ([example repo](https://github.com/edyburn/yarn-resolutions-integrity-example)). Run `yarn install` in the package. Run `yarn install` again, which will go through all the installation steps instead of bailing.

With some tracing, I found that the integrity check fails with `LOCKFILE_DONT_MATCH` because the resolution pattern exists in the lockfile on disk, but not in the cached lockfile. Apparently, the resolution pattern [is getting deleted from the cached lockfile by the `PackageResolver`](https://github.com/yarnpkg/yarn/blob/v1.4.0/src/package-resolver.js#L630) before the integrity check. Removing the linked line from the `PackageResolver` fixes the integrity check issue, and the tests pass. However, I haven't yet figured out the original reason for removing the resolution pattern from the lockfile, so maybe the conditional on the preceding line should be changed instead?

**What is the expected behavior?**
When using `resolutions`, `yarn install` should bail out early if the installation is already up-to-date.

**Please mention your node.js, yarn and operating system version.**
yarn 1.3.2 (and 1.4.0)
node 9.3.0
linux 4.14.13 x86_64

dreyks:
yes! can confirm. was just going to file the same issue.

kaylieEB:
Hi @edyburn , I can't see to repro
```integrity-example.git⑂master 🦊  git clone https://github.com/edyburn/yarn-resolutions-
Cloning into 'yarn-resolutions-integrity-example'...
remote: Counting objects: 10, done.
remote: Compressing objects: 100% (7/7), done.
Unpacking objects: 100% (10/10), done.
remote: Total 10 (delta 1), reused 10 (delta 1), pack-reused 0
~/playground/testing ⑂master* 🦊  cd yarn-resolutions-integrity-example/
~/playground/testing/yarn-resolutions-integrity-example ⑂master 🦊  yarn
yarn install v1.4.1
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 1.02s.
[...]

Issue #4778, opened 2017-10-25
`yarn install --frozen-lockfile` exits with error, but `yarn install` does nothing afterwards


kompot:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
1. Run `yarn install` with package.json present and no `yarn.lock` present.
2. Immediately after run `yarn install --frozen-lockfile` 
3. Errors states that `error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile.`
4. Run `yarn install` again - no changes are made to `yarn.lock`

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/kompot/yarn-frozen-lockfile-bug

**What is the expected behavior?**
`yarn install` should fix `yarn.lock` so that it's up-to-date OR `yarn --frozen-lockfile` should not fail

**Please mention your node.js, yarn and operating system version.**
macOS 10.12.6
yarn 1.2.1
node 6.11

rally25rs:
From a quick glance, it looks like for some reason Yarn is reporting that `babel-runtime@^6.26.0` is a missing pattern in the lockfile, but that pattern does exist:

```
babel-runtime@6.26.0, babel-runtime@6.x.x, babel-runtime@^6.11.6, babel-runtime@^6.18.0, babel-runtime@^6.22.0, babel-runtime@^6.23.0, babel-runtime@^6.26.0, babel-runtime@^6.5.0, babel-runtime@^6.9.2:
  version "6.26.0"
  resolved "https://registry.yarnpkg.com/babel-runtime/-/babel-runtime-6.26.0.tgz#965c7058668e82b55d7bfe04ff2337bc8b5647fe"
  dependencies:
    core-js "^2.4.0"
    regenerator-runtime "^0.11.0"
```

I'll continue digging in ...

rally25rs:
More research notes:

It looks like what happens is that:

* the lockfile is loaded and includes `babel-runtime@^6.26.0`

* PackageResolver.resolveToResolution() removes a lot of entries from the lockfile, including `babel-runtime@^6.26.0`
```
  resolveToResolution(req: DependencyRequestPattern): ?DependencyRequestPattern {
    const {parentNames, pattern} = req;

    if (!parentNames || this.flat) {
      return req;
    }

    const resolution = this.resolutionMap.find(pattern, parentNames);

    if (resolution) {
      const resolutionManifest = this.getResolvedPattern(resolution);

[...]

PR #5572, opened 2018-03-25
do not remove pattern from lockfile if it is overridden by resolution

MhMadHamster:
**Summary**
This pr addresses issue #5250 and #4778 . Currently if package.json contains resolutions, pattern which is overriden by resolution will be removed from lockfile, which causing integrity-check fail and as following new lockfile being generated. So i just removed code which deletes the pattern and since resolutions test are passing i assuming that part of code should not be there or some test cases are missing. Issue #4778 was fixed with #4793, so i reverted the code from that pr.

**Test plan**

Added a new test case in resolutions which will fail in a current master.


__tests__/commands/install/resolutions.js
```diff
@@ -92,3 +95,14 @@ test.concurrent('install with nested resolutions using flat mode', (): Promise<v
...
+test.concurrent('install with resolutions does not write new lockfile if existing one satisfied', (): Promise<void> => {
+  return runInstall(
+    {},
+    {source: 'resolutions', cwd: 'install-with-resolutions-dont-write-new-lockfile-if-satisfied'},
+    async (config): Promise<void> => {
+      const lockfile = await fs.readFile(path.join(config.cwd, 'yarn.lock'));
+      expect(lockfile.indexOf('foobar')).toBeGreaterThanOrEqual(0);
+    },
+  );
+});
```
arcanis: Can you rather test that we correctly bailout (ie `install with resolution settings should correctly bailout during the integrity check`)?

MhMadHamster: done

PR #5581, opened 2018-03-27
Remove leading spaces of `yarn run`'s answer

pd4d10:
**Summary**

If user's answer contains leading spaces, the command will fail. Like:

```sh
> yarn run
...
question Which command would you like to run?:  start
error An unexpected error occurred: "Command \"\" not found.".
```

This PR add a `trim` operation to remove leading spaces

**Test plan**


waleedahmed3045:
@pd4d10 its a good idea but i think its better to have strict check on the input of the user. I would prefer if error is more descriptive and specifies the full entry not just empty spaces :)

arcanis:
I think it's fine - this syntax is only used from the command line and not by script, so from a DevX perspective it's better if we can be more tolerant on such mistakes.

PR #5603, opened 2018-03-31
adding resolutions should cause lockfile regeneration

MhMadHamster:
**Summary**
After PR #5572 if you already have package.json with some dependencies and generated lockfile, after  adding resolution and running `yarn install` lockfile wont be updated and package overriden by resolution will not be removed

**Test plan**
added new test which checks that after adding resolution, redundant package removed, maybe it is better to check the lockfile itself?

before:
image

after:
image


__tests__/commands/install/resolutions.js
```diff
@@ -115,3 +115,26 @@ test.concurrent('install with resolution settings should correctly bailout durin
...
+test.concurrent('adding resolutions after install should cause lockfile regeneration on second install', (): Promise<
+  void,
+> => {
+  return runInstall(
+    {},
+    {source: 'resolutions', cwd: 'adding-resolutions-should-casue-lockfile-regeneration'},
+    async (config, reporter): Promise<void> => {
+      const packageJson = await fs.readFile(path.join(config.cwd, 'package.json'));
+      // create new package.json with resolutions which override e/left-pad version
+      const newPackageJson = packageJson.replace(/}\n}/g, '  },\n"resolutions": {\n    "e/left-pad": "1.1.1"\n  }\n}');
```
BYK: Can you use `JSON.parse`, modify the parsed object and then use `JSON.stringify` to write it back instead of using `.replace` on the JSON string?

__tests__/commands/install/resolutions.js
```diff
@@ -115,3 +115,26 @@ test.concurrent('install with resolution settings should correctly bailout durin
...
+      // write new package.json
+      await fs.writeFile(path.join(config.cwd, 'package.json'), newPackageJson);
+      // run install again
+      const reinstall = new Install({}, config, reporter, await Lockfile.fromDirectory(config.cwd));
+      await reinstall.init();
+      // don't expect left-pad in e/node_modules since it's now being replaced by single dependency left-pad@1.1.1
+      expect(await fs.exists(path.join(config.cwd, 'node_modules', 'e', 'node_modules', 'left-pad', 'index.js'))).toBe(
```
BYK: I'd also verify that the lockfile got updated here.

MhMadHamster:
tests failed because of the new version of left-pad, so snapshot doesn't match

src/package-resolver.js
```diff
@@ -447,6 +447,12 @@ export default class PackageResolver {
     return maxValidRangeManifest;
   }
 
+  getPackageRemoteResolved(manifest: Manifest): ?string {
+    const packageReference: ?PackageReference = manifest._reference;
+    const packageRemote: ?PackageRemote = packageReference ? packageReference.remote : null;
+    return packageRemote ? packageRemote.resolved : null;
```
kaylieEB: I don't _think_ you need to do all the null checks given the types and invariant check... Also I'm seeing that this can be more of a helper/validation logic with what you have below that could later be moved to integrity checker, i.e. `if there is a lockfile version, does the lockfile version match the resolution's resolved version?` 

```
// in resolution-map.js
export const shouldUpdateLockfile = (lockfileEntry: ?LockManifest, resolutionEntry: ?PackageReference) => {
    if (!lockfileEntry || !resolutionEntry) {
        return false;
    }

   return lockfileEntry.resolved !== resolutionEntry.remote.resolved;
}

// then call below;
if (shouldUpdateLockfile(lockManifest, resolutionManifest._reference)) {
   this.lockfile.removePattern(pattern);
}
```
maybe something like this would work? wdyt?

MhMadHamster: sure, this looks more readable

PR #5605, opened 2018-04-01
fix(upgrade): No longer warn when upgrading a devDependency

### LINKED ISSUE #4840: yarn upgrade - no update and spurious warning about existing entry in package.json for devDependencies
yarn 1.3.0

Bug (https://github.com/yarnpkg/yarn/issues/1458 is *NOT* fixed)

If there exists a devDependency when upgrading a package, yarn moans about this because it seems to want to add a new dependency. But doesn't do anything.

```
yarn init
yarn add @types/node -D
yarn upgrade @types/node

```
issues
`warning "@types/node" is already in "devDependencies". Please remove existing entry first before adding it to "dependencies".
`

update the devDependency (only) without a warning

node 9, yarn 1.3.2, Windows 10

rally25rs:
fixes #4840 

**Summary**

Previously the upgrade command would call add with the list of packages to upgrade to. No "--dev"
flag would be carried over and this list would contain all dependencies. As a result, Add would
report a warning `{package} is already in "devDependencies". Please remove existing entry first
before adding it to "dependencies".`

I am using the existing `config.commandName` to decide whether or not to display the warning.

I also updated the jest snapshots, because I was tired of the warning about 40-some obsolete snapshots, so the snapshot changes here aren't relevant to the functionality, just cleaning up.

**Test plan**

Added regression test in `__tests__/commands/upgrade.js`.

Upgrade tests now set `config.commandName` which is an existing property, but is not normally set by tests. I wanted to set it for all tests in `_helpers.js` but we don't actually specify the command we are going to test when we build a test runner.


rally25rs:
Seems that the "obsolete" snapshots are actually needed for CI environment. I'll redo this PR without those snapshot changes...

rally25rs:
retrying with no snapshot changes as #5606

Issue #5579, opened 2018-03-27
Too broad packages picked up when using nohoist option


maoueh:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
I started to play with the `nohoist` option on one of our internal project which has a structure like:

```
├── modules
│   ├── client
│   ├── domain
│   └── frontend
│       ├── common
│       └── mobile
```

When adding the `nohoist` option, either in the child projet `modules/frontend/mobile` or at the top level, it appears that on a second `yarn install` call, Yarn is picking up too much things probably looking inside `node_modules` where it should not leading to this series of messages:

```
yarn install v1.5.1
warning Missing name in workspace at "/private/tmp/yarn-1.5.1-too-broad-packages-with-nohoist/packages/frontend/mobile/node_modules/metro-source-map/src", ignoring.
warning Missing name in workspace at "/private/tmp/yarn-1.5.1-too-broad-packages-with-nohoist/packages/frontend/mobile/node_modules/metro/src/ModuleGraph/node-haste", ignoring.
warning Missing name in workspace at "/private/tmp/yarn-1.5.1-too-broad-packages-with-nohoist/packages/frontend/mobile/node_modules/metro/src/Server/symbolicate", ignoring.
error An unexpected error occurred: "There are more than one workspace with name \"nw-pre-gyp-module-test\"".
info If you think this is a bug, please open a bug report with the information provided in "/private/tmp/yarn-1.5.1-too-broad-packages-with-nohoist/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
git clone https://github.com/maoueh/yarn-1.5.1-too-broad-packages-with-nohoist.git
cd yarn-1.5.1-too-broad-packages-with-nohoist.git
yarn install # This one succeed
yarn install # This one fails with the above messages
```

**What is the expected behavior?**
No warning and no error on the second yarn install

**Possible workaround?**
Narrowing the `workspaces/packages` option resolves the issue. Here's what I used in my project:

```
  "workspaces": {
    "packages": [
      "modules/client",
      "modules/domain",
      "modules/frontend/*",
      "modules/tools/*"
    ],
    "nohoist": ["**/react-native*", "**/react-native*/**"]
  },
```

**Please mention your node.js, yarn and operating system version.**

 * Yarn 1.5.1
 * Node 8.4.0
[...]

PR #5607, opened 2018-04-01
fix workspace packages ignore-pattern

connectdotz:
**Summary**

address #5579 related to #3986/#3996: 
in order to prevent greedy workspaces.packages patterns to match unnecessary packages, #3996 added the `node_modules/*/package.json` pattern to exclude packages under the workspace package's node_modules folder. While sufficient for most use cases, it does not work for "deep" path such as "node_modules/X/Y/package.json". This PR replaced the shallow wildcard with globstar "**" so it will ignore **all** packages under the workspace package's node_modules folder.

- Config.js: replace the shallow ignore pattern with a deep pattern.
- execute-lifecycle-script: address when process.mainModule is undefined
- tests: the rest are all tests.

**Test plan**

both new and old tests should pass

### DIFF
--- a/__tests__/commands/install/workspaces-install.js
+++ b/__tests__/commands/install/workspaces-install.js
@@ -243,6 +243,29 @@ test.concurrent('install should ignore node_modules in workspaces when used with
   });
 });
 
+describe('install should ignore deep node_modules in workspaces', () => {
+  test('without nohoist', (): Promise<void> => {
+    return runInstall(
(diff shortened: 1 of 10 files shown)


connectdotz:
ping @arslanarshad31 @bestander for comments

connectdotz:
'yarn test` were failing

arcanis:
I see, probably a Flow update

PR #5608, opened 2018-04-01
remove .vscode settings from repository

connectdotz:
**Summary**

looks like a .vscode/settings file has been accidentally checked in a few days ago (#5572). This file is only related to vscode and most vscode dev would probably prefer their own settings ;-). We should not need to keep it in the repo.

- removed settings.json 
- added `.vscode/` to .gitignore to prevent accidental checkin in the future.

**Test plan**

there is no code change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ test/fixtures/**/.fbkpm
 .idea
 .yarn-meta
 /packages/lockfile/index.js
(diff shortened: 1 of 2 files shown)


connectdotz:
ping @arcanis @kaylieEB

arcanis:
Woops, thanks! :)

PR #5625, opened 2018-04-05
feat: added a `--size` flag to `yarn why`

Aghassi:
**Summary**
Added a flag to `yarn why` that moves the size reporting to opt in by default. While size information is useful, it isn't necessary for most of the time when looking up `why` I have a dependency.

I was motivated to make this change because the information provided by `yarn why` by default is meant to be "why do I have this dependency, where did it come from?". Most of the time, I find I don't care much for the size information. This also will make tools that parse `json` output from `why` have to account for ever so slightly less code.

On larger data sets, I expect this to speed up the execution of `yarn why` because it will avoid doing deep size calculations.

**Test plan**

I have writtent tests for what I could think of. However, I realize they are pretty weak and would love some guidance on making them better. And yes, there is some copy paste in there as this is my first time working with flow so I wanted an easy way to at least get the change working. I expect code review to weed out the duplicate and bad code.
In addition, the contribution guide didn't say the best way to actively test yarn. I assume it is linking it globally? Not sure. To be totally fair, I haven't actually run the command with this flag. All tests and code have been written and run through the test harness. I would also love some guidance (and would be happy to contribute it) on how to actively test the changes made.

### DIFF
--- a/src/cli/commands/why.js
+++ b/src/cli/commands/why.js
@@ -135,10 +135,18 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
   const query = await cleanQuery(config, args[0]);
 
-  reporter.step(1, 4, reporter.lang('whyStart', args[0]), emoji.get('thinking_face'));
+  // Be default there are 3 steps to this task
+  let total = 3;
+
+  // If the user wants size information, we enable the fourth step
+  if (flags.size) {
+    total = 4;
+  }
+
+  reporter.step(1, total, reporter.lang('whyStart', args[0]), emoji.get('thinking_face'));
 
   // init
-  reporter.step(2, 4, reporter.lang('whyInitGraph'), emoji.get('truck'));
+  reporter.step(2, total, reporter.lang('whyInitGraph'), emoji.get('truck'));
   const lockfile = await Lockfile.fromDirectory(config.lockfileFolder, reporter);
   const install = new Install(flags, config, reporter, lockfile);
   const {requests: depRequests, patterns, workspaceLayout} = await install.fetchRequestFromCwd();
@@ -150,7 +158,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   const hoisted = await install.linker.getFlatHoistedTree(patterns);
 
   // finding
-  reporter.step(3, 4, reporter.lang('whyFinding'), emoji.get('mag'));
+  reporter.step(3, total, reporter.lang('whyFinding'), emoji.get('mag'));
 
   const matches = queryWhy(query, hoisted);
 
@@ -159,7 +167,13 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     return;
   }
 
-  const processMatch = async (match: HoistManifestTuple) => {
+  /**
+   * A method that handles all the matches for yarn why
+   * @param {*} match A `HoistManifestTuple` that denotes a match
+   * @param {*} flags Any flags provided by the calling of `yarn why`. 
+   * If `size` is provided, this method will also calculate the direct and transitive size.
+   */
+  const processMatch = async (match: HoistManifestTuple, flags: Object) => {
     const [, matchInfo] = match;
     const matchRef = matchInfo.pkg._reference;
     invariant(matchRef, 'expected reference');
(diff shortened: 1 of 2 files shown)


src/cli/commands/why.js
```diff
@@ -199,20 +213,28 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
     let transitiveSizes = [];
-    try {
-      packageSize = await getPackageSize(match);
-    } catch (e) {}
+
+    // If we are asked to get the size of the matches, we attempt to gather size information
+    if (flags.size) {
+      try {
+        packageSize = await getPackageSize(match);
+      } catch (e) {}
```
BYK: Never ever do blanket try-catch blocks like this, please. Why are we swallowing the error here and below?

src/cli/commands/why.js
```diff
@@ -135,10 +135,18 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
   const query = await cleanQuery(config, args[0]);
 
-  reporter.step(1, 4, reporter.lang('whyStart', args[0]), emoji.get('thinking_face'));
+  // Be default there are 3 steps to this task
+  let total = 3;
+
+  // If the user wants size information, we enable the fourth step
+  if (flags.size) {
```
BYK: I think instead of doing this, you can have an array: `const steps = [];`

Start pushing the steps as functions to that:
```js
steps.push({msg: reporter.lang('whyStart', args[0]), emoji.get('thinking_face'), action: () => {}});
steps.push({msg: reporter.lang('whyInitGraph'), emoji.get('truck'), action: () => {
  const lockfile = await Lockfile.fromDirectory(config.lockfileFolder, reporter);
  ...
}});

...

if (flags.size) {
  steps.push(...);
}

for (const i = 0; i < steps.length; i++) {
  const {msg, action} = steps[i];
  reporter.step(i+1, steps.length, msg);
  action();
}
```

BYK:
>Let's turn this on by default since turning this off by default would be a backward-incompatible change and would require a major version bump. We can turn it off in 2.0.0

If you are concerned about this for your own use case, I think you can just add `--why.with-size false` to your `.yarnrc` file and live happily ever after :)

src/cli/commands/why.js
Aghassi: I hate to give this answer, but it was like that when I got here (look at the diff).

What would be a better way to handle this error? Gracefully or do we want to hard exit?

BYK: Ah, sorry you're right. I'm guessing it was to protect against `ENOENT` or something but now I need to dig into commit logs and PR discussions. You can leave these as they are :)

Aghassi:
@BYK I have addressed all comments except for the "turn this on be default". I wanted to clarify, what is the default behavior that you would like to see?

arcanis:
Not sure I find this flag a good addition.

- It makes the command line interface slightly more complex - I prefer tools to keep a small API, and Yarn is already bigger than I'm comfortable with
- The JSON output is something that I want to revisit later, it would be better if you didn't have to parse this at all
- We have ways to make size calculation almost instantaneous, so this shouldn't be an issue

Aghassi:
@arcanis That's fair, do you have a suggested root forward with this on how to maybe improve or change this PR? Otheriwse, it is ok to close it if the plan is to revisit the feature at a different time.

src/cli/commands/why.js
```diff
@@ -135,10 +136,28 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
 
   const query = await cleanQuery(config, args[0]);
 
-  reporter.step(1, 4, reporter.lang('whyStart', args[0]), emoji.get('thinking_face'));
+  // Steps that will be displayed during the run
+  // These always run so we store them in array so we can keep the length of steps should we add one after these
+  const steps = [
+    {msg: reporter.lang('whyStart', args[0]), emoji: emoji.get('thinking_face'), action: () => {}},
```
BYK: #nit: since you have separated the `emoji` part, you can simply pass the name of the emoji `emoji: 'thinking_face'` and use `emoji.get(emoji)` below.

src/cli/commands/why.js
```diff
@@ -255,9 +288,14 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
   };
 
-  reporter.step(4, 4, reporter.lang('whyCalculating'), emoji.get('aerial_tramway'));
+  // Only show fourth step if the size flag is passed
+  if (flags.withSize) {
+    steps.push({msg: reporter.lang('whyCalculating'), emoji: emoji.get('aerial_tramway'), action: () => {}});
```
BYK: You need to do `steps.push()` for this before your first `displayStep()` call for it to show the right number of steps on the console.

BYK:
>@BYK I have addressed all comments except for the "turn this on be default". I wanted to clarify, what is the default behavior that you would like to see?

All I want is `yarn why` to produce the exact same output with and without this patch.

I'll leave managing of this issue to @arcanis since he has raised some good points.

PR #5626, opened 2018-04-06
chore(deps): update dependency validate-npm-package-license to ^3.0.3

qkdreyer:
**Summary**

Updating validate-npm-package-license in order to update its dependencies :
- spdx-correct
- spdx-license-ids

This is allowing us to use newer licences such as EUPL-1.2.

https://github.com/kemitchell/validate-npm-package-license.js/compare/v3.0.1...v3.0.3
https://github.com/jslicense/spdx-correct.js/compare/v1.0.2...v3.0.0
https://github.com/shinnn/spdx-license-ids/compare/v1.2.2...v3.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "tar-stream": "^1.5.2",
     "uuid": "^3.0.1",
     "v8-compile-cache": "^1.1.0",
-    "validate-npm-package-license": "^3.0.1",
+    "validate-npm-package-license": "^3.0.3",
     "yn": "^2.0.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -5822,19 +5822,27 @@ spawn-sync@^1.0.15:
     concat-stream "^1.4.7"
     os-shim "^0.1.2"
 
-spdx-correct@~1.0.0:
-  version "1.0.2"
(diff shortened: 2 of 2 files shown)


qkdreyer:
I've already looked at the ci/circleci build output, and it fails because of :

```
#!/bin/bash -eo pipefail
yarn install --frozen-lockfile
yarn install v1.1.0
[2/5] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
Exited with code 1
```

From my understanding, we may not run ci/circleci tests if we're updating the yarn.lock file since it's running with the frozen-lockfile option.

BYK:
>From my understanding, we may not run ci/circleci tests if we're updating the yarn.lock file since it's running with the frozen-lockfile option.

Don't think I understand what you mean. Can you elaborate?

The issue is (the reason why CI fails) [the entry in `yarn.lock`](https://github.com/yarnpkg/yarn/pull/5626/files#diff-8ee2343978836a779dc9f8d6b794c3b2R6382) not matching the entry in the package.json file. One says `@^3.0.1` and the other says `@^3.0.3` so the failure is legit.

qkdreyer:
Could you re-run the test-macos-node6/8 tests that were canceled ?

qkdreyer:
Nevermind, I've force pushed to retrigger it. All checks have passed. Could you merge this please ?
I could also add a few tests but don't know where to add them.

PR #5632, opened 2018-04-10
fix(build): Stop scripts from trying to read user input during install

### LINKED ISSUE #976: Bug: unable to install semantic-ui
[Semantic-UI](https://github.com/Semantic-Org/Semantic-UI) installation hangs up with the following output:

```
$ yarn add semantic-ui 
yarn add v0.15.1
warning No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
warning fsevents@1.0.14: The platform "linux" is incompatible with this module.
info "fsevents@1.0.14" is an optional dependency and failed compatibility check. Excluding it from installation.
warning lodash@1.0.2: The engine "rhino" appears to be invalid.
[3/4] Linking dependencies...
[1/1] ⠐ semantic-ui:     at AnonymousObservable.tryCatcher (/home/bserdiuk/coding/test/node_modules/rx-lite/rx.lite.js:63:31)
[-/1] ⠐ waiting...
[-/1] ⠈ waiting...
[-/1] ⠈ waiting...
[-/1] ⠈ waiting...

```

the package installs fine.

Ubuntu 16.04.1 LTS
Node v6.7.0
Yarn v0.15.1
## 

P.S. Semantic-UI has [a huge installation script](https://github.com/Semantic-Org/Semantic-UI/blob/master/tasks/install.js). I guess it relies on some undocumented NPM logic. If I can gather more information for you, tell me how, I'd be happy to help.

iarna:
**Summary**

Fixes #976.

@byk mentioned to me that this was a problem you all had been having and while doing research for npm running lifecycle scripts in parallel I came across this solution and thought I'd share it with y'all:

This is a bug fix to the problem where lifecycle scripts that try to go interactive hang yarn installs with no indication as to why.

This sets stdin to /dev/null and, on unix-like systems, detaches the process from the terminal making it unable to read /dev/tty.

Previously with the spinner enabled was impossible to give input to an interactive script on stdin, because a pipe was kept open from the main yarn process to the child process but it was never written to.  Interactive scripts could previously use /dev/tty on unix-likes to bypass this, however the fact that scripts are run in parallel means that if two scripts go interactive in this manner that they'll step on each other.

This does not change the behavior if the spinner is disabled. Also added a simple integration test with a blocking install script that times out without the patch.

**Test plan**

Try running `yarn add semantic-ui`. Hangs without the patch, fails with the patch.


BYK:
Ignore the AppVeyor failure until we merge #5638.

BYK:
I have one concern: `yarn run x` uses the same function and in that scenario, I think allowing stdin is something we want. If you agree, I'll add a new flag to this function, `interactive` or `allowInteractive` to disable this new behavior for `yarn run`.

arcanis:
I don't think we should change `yarn run`'s default behavior - it would be quite unexpected if programs that expect an stdin suddenly stopped working. Isn't the original issue only about the spinner, ie scripts executed during install?

BYK:
>I don't think we should change yarn run's default behavior - it would be quite unexpected if programs that expect an stdin suddenly stopped working. Isn't the original issue only about the spinner, ie scripts executed during install?

Yes to both. I'll look into the `spinner` option and see if it is used by anything else. If not I'll rename it and switch the behavior to `'inherit'` everything like we currently do.

__tests__/commands/version.js
```diff
@@ -72,28 +72,58 @@ test('run version and make sure all lifecycle steps are executed', (): Promise<v
...
     const pkg = await fs.readJson(path.join(config.cwd, 'package.json'));
 
-    const preversionLifecycle = ['preversion', config, pkg.scripts.preversion, config.cwd];
-    const postversionLifecycle = ['postversion', config, pkg.scripts.postversion, config.cwd];
+    const preversionLifecycle = {
+      stage: 'preversion',
+      config,
+      cmd: pkg.scripts.preversion,
+      cwd: config.cwd,
+      isInteractive: false,
```
BYK: @iarna  - I think we need to mark `postX` and `preX` scripts on the main package as interactive. What do you think?

/cc @bestander @arcanis

src/cli/commands/run.js
```diff
@@ -62,30 +62,37 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
         const cmdWithArgs = stage === action ? sh`${unquoted(cmd)} ${args}` : cmd;
         const customShell = config.getOption('script-shell');
         if (customShell) {
-          await execCommand(stage, config, cmdWithArgs, config.cwd, String(customShell));
+          await execCommand({
+            stage,
+            config,
+            cmd: cmdWithArgs,
+            cwd: config.cwd,
+            isInteractive,
```
BYK: If we decide to make `postX` and `preX` scripts interactive, then this line should simply become `isInteractive: true` and all other changes above should be removed.

arcanis: I think the pre/post scripts can be left as non-interactive, this should be fine

src/util/child.js
```diff
@@ -61,7 +61,7 @@ type ProcessFn = (
 export function spawn(
   program: string,
   args: Array<string>,
-  opts?: child_process$spawnOpts & {process?: ProcessFn} = {},
+  opts?: child_process$spawnOpts & {detached?: boolean, process?: ProcessFn} = {},
```
arcanis: Why do we need to declared `detached`? Is it missing in the flow definition for `child_process$spawnOpts`? In that case, maybe we should make a PR to add it

BYK: Yes to both :)

BYK:
@iarna  - updated as you suggested. Once you give me the go, I'll merge it.

Let us know if you want someone from @yarnpkg/core to give it a second look before merging tho.

Thanks again for submitting this!

src/cli/commands/run.js
```diff
@@ -83,9 +83,16 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
...
+            stage,
+            config,
+            cmd: cmdWithArgs,
+            cwd: config.cwd,
+            isInteractive: true,
+            customShell: String(customShell),
+          });
         } else {
-          await execCommand(stage, config, cmdWithArgs, config.cwd);
+          await execCommand({stage, config, cmd: cmdWithArgs, cwd: config.cwd, isInteractive: true});
```
kaylieEB: Thanks for cleaning this up with named parameters! Can we do
```
await execCommand({
            stage,
            config,
            cmd: cmdWithArgs,
            cwd: config.cwd,
            isInteractive: true,
            customShell: customShell && String(customShell)
        });
```
and remove the conditional now? :)

src/util/execute-lifecycle-script.js
```diff
@@ -202,28 +206,15 @@ export async function executeLifecycleScript(
...
-        );
-      }
-    };
-  }
+  // if we aren't interactive then pipe everything to the terminal
+  const stdio = isInteractive ? 'inherit' : undefined;
+  // if this is not interactive, run child processes detached so they can't hang
+  // trying to read from /dev/tty but not on Windows, because Windows opens
+  // a new console window and there is no /dev/tty anyway
+  const detached = !isInteractive && process.platform !== 'win32';
```
kaylieEB: I find it slightly more readable this way, wdyt?
```
// By default (non-interactive), pipe everything to the terminal and run child process detached
// as long as it's not Windows (since windows does not have /dev/tty)
let stdio;
let detached = process.platform !== 'win32';

if (isInteractive) {
   stdio = 'inherit';
   detached = false;
}
```

BYK: Don't really like using `let` there but I also agree that your version expresses the intent more clearly. I'll see if I can find a middle ground :)

BYK: I ended up using your version since it is clearly more readable.

src/package-install-scripts.js
```diff
@@ -279,18 +301,15 @@ export default class PackageInstallScripts {
       workQueue.add(pkg);
     }
 
+    const set = this.reporter.activitySet(
+      installablePkgs,
+      Math.min(installablePkgs, this.config.childConcurrency, workQueue.size),
```
BYK: Also fixed a minor issue here when you only had 1 install script but you ended up seeing a bunch of `[1/-] waiting...` messages.

iarna:
@byk looks good!

BYK:
I've merged this since the single build failure was due to a flaky test on master. I'll fix that separately.

TallOrderDev:
Is this live now? Running into the issue and not sure if I just need to upgrade yarn or not (Perfect timing if it is live!)

arcanis:
It has been merged, but not yet released. You're free to use the nightly
builds, which should be stable enough anyway (we merge only if the tests
pass) :)

https://yarnpkg.com/en/docs/nightly

PR #5640, opened 2018-04-10
feat(version): Ability to disable commit hooks on versioning

faazshift:
**Summary**

Last year I contributed the ability to disable commit hooks when versioning with npm (See [this commit](https://github.com/npm/npm/commit/2dec601c6d5a576751d50efbcf76eaef4deff31e#diff-2992be03e25e9f38300b7ce175e4f4e1)). It makes sense to add the same functionality to yarn.

The original motivation for this feature is based on a pre-commit hook that included linting and testing a large code base, taking ~20 seconds to run. When that hook is run on the preceding commits, it's unnecessary and time-taking to do so on versioning. Because of that, it's nice to have the option to disable that git hook.

**Test plan**

Added lower level integration tests.

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -55,6 +55,10 @@ export async function setVersion(
     return Promise.resolve();
   }
 
+  function isCommitHooksDisabled(): boolean {
+    return flags.commitHooks === false || config.getOption('version-commit-hooks') === false;
+  }
+
   if (pkg.scripts) {
     // inherit `scripts` from manifest
     Object.assign(scripts, pkg.scripts);
@@ -131,14 +135,6 @@ export async function setVersion(
     return () => Promise.resolve();
   }
 
-  function buildCommitArgs(args): Array<string> {
-    args = args || ['commit'];
-    if (!flags.commitHooks || !config.getOption('version-commit-hooks')) {
-      args.push('-n');
-    }
-    return args;
-  }
-
   return async function(): Promise<void> {
     invariant(newVersion, 'expected version');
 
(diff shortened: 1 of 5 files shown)


src/cli/commands/version.js
```diff
@@ -130,6 +131,14 @@ export async function setVersion(
     return () => Promise.resolve();
   }
 
+  function buildCommitArgs(args): Array<string> {
```
BYK: I don't think abstracting this into a separate function adds much value unless you use it in a unit test.

src/cli/commands/version.js
```diff
@@ -130,6 +131,14 @@ export async function setVersion(
...
+    args = args || ['commit'];
```
BYK: Don't think allowing an empty args value is useful. It adds a new code path to be tested and currently this code path is not even being used. I'd rather us to enforce an array argument through proper flow definitions.

src/cli/commands/version.js
```diff
@@ -130,6 +131,14 @@ export async function setVersion(
...
+    if (!flags.commitHooks || !config.getOption('version-commit-hooks')) {
```
BYK: I think this part would be more useful as a separate function: `isCommitHooksDisabled()`. Then you'd use it like as follows:
```
// create git commit
await spawnGit(['commit', ...(isCommitHooksDisabled ? ['-n'] : []), '-m', message], {cwd: gitRoot});
```

faazshift:
@BYK Thanks for the feedback. I've made the requested changes and added test cases.

Issue #5660, opened 2018-04-12
Bug: upgrade with pattern upgrades all packages


swrobel:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
`yarn upgrade --pattern` seemingly ignores the pattern and upgrades all.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
$ yarn outdated
yarn outdated v1.5.1
info Color legend : 
 "<red>"    : Major Update backward-incompatible updates 
 "<yellow>" : Minor Update backward-compatible features 
 "<green>"  : Patch Update backward-compatible bug fixes
Package                               Current Wanted Latest Package Type    URL                                                        
@fortawesome/fontawesome-free-brands  5.0.9   5.0.10 5.0.10 dependencies    https://fontawesome.com                                    
@fortawesome/fontawesome-free-regular 5.0.9   5.0.10 5.0.10 dependencies    https://fontawesome.com                                    
@fortawesome/fontawesome-free-solid   5.0.9   5.0.10 5.0.10 dependencies    https://fontawesome.com                                    
bootstrap                             4.0.0   4.1.0  4.1.0  dependencies    https://getbootstrap.com  

$ yarn upgrade --pattern fontawesome
...
info Direct dependencies
├─ @fortawesome/fontawesome-free-brands@5.0.10
├─ @fortawesome/fontawesome-free-regular@5.0.10
├─ @fortawesome/fontawesome-free-solid@5.0.10
├─ bootstrap@4.1.0
...

$ yarn outdated | grep bootstrap
<no output>
```

@philquinn, am I understanding this feature incorrectly? https://github.com/yarnpkg/yarn/pull/4610

**What is the expected behavior?**
`--pattern` should restrict upgraded packages to only those that match the pattern

**Please mention your node.js, yarn and operating system version.**
Node 9.11.1
yarn 1.5.1
macOS 10.13.4

rally25rs:
The worst part is we even have a unit test for `--pattern`, but it was poorly written and passes when all packages upgrade 😞 

Thanks for reporting this. I'll try to track down when/why this broke.

philquinn:
Hey @swrobel , I can reproduce this. It appears the part which filters on the pattern is unreachable unless a package is also passed. So doing `yarn upgrade some-package-name --pattern fontawesome` would get what you want I think. But thats a bug. It should work without the package just the same. I will try do a fix for this as soon as I can.

[...]

PR #5662, opened 2018-04-12
Fix for pattern not working with upgrade

philquinn:
**Summary**

Fix for #5660

Before it needed to have package argument to enter the block to filter the deps. Now it will do it if either a package is given or a --pattern is given.

**Test plan**

Corrected my incorrectly created tests.
Also `is-negative-zero` would not update as there is no new version which would be found in a caret. So changed that to `array-union` which has a version `1.0.2`


rally25rs:
Thanks for the quick fix! 👍

PR #5663, opened 2018-04-13
Focused Workspaces RFC Implementation

bdwain:
UPDATE: This is complete and ready for final review.

---

I've been working on an implementation of the [focused workspaces RFC](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-focused-workspaces.md). It's not complete yet, but it mostly works so I was hoping to get some feedback as i try to get the last few things working to make sure my implementation makes sense. 

All of the details of what it does and the edge cases are in the RFC link above, so I won't explain any of that here. I was just hoping to get general high level feedback on my approach. Most importantly, I was hoping for feedback on was how I used the hoisted tree result (from package-hoister) to calculate which packages need to be shallowly installed inside the target workspace and where. I'll comment inline to point out where that is.

FYI I haven't written any unit tests yet. I'm beginning work on those now. But I wanted to post this early to get feedback in case there are any major issues with my implementation.

Sorry in advance if there's a better place to do this (like a development chat or something). This is my first big yarn change so I wasn't sure what the normal process was. 

Thanks!

cc @bestander


src/package-request.js
```diff
@@ -181,14 +181,20 @@ export default class PackageRequest {
...
-  findVersionInfo(): Promise<Manifest> {
+  async findVersionInfo(): Promise<Manifest> {
     const exoticResolver = getExoticResolver(this.pattern);
     if (exoticResolver) {
       return this.findExoticVersionInfo(exoticResolver, this.pattern);
     } else if (WorkspaceResolver.isWorkspace(this.pattern, this.resolver.workspaceLayout)) {
       invariant(this.resolver.workspaceLayout, 'expected workspaceLayout');
       const resolver = new WorkspaceResolver(this, this.pattern, this.resolver.workspaceLayout);
-      return resolver.resolve();
+      let manifest;
```
bdwain: workspaces were never downloaded from the registry before before, so i had to add that here.

src/package-hoister.js
```diff
@@ -649,6 +658,81 @@ export default class PackageHoister {
     }
   }
 
+  markShallowWorkspaceEntries(){
```
bdwain: this is where the main logic is happening.

src/package-hoister.js
```diff
@@ -649,6 +658,81 @@ export default class PackageHoister {
...
+
+    entries.forEach(([key, info]) => {
+      const splitPath = key.split('#');
+
+      //mark the workspace and any un-hoisted dependencies it has for shallow installation
+      const isShallowDependency = dependentWorkspaces.some(w => {
+        if(splitPath[0] !== w){ //entry is not related to the workspace
+          return false;
+        }
+        if(!splitPath[1]){ //entry is the workspace
```
bdwain: all sibling workspaces are shallow dependencies

src/package-hoister.js
```diff
@@ -649,6 +658,81 @@ export default class PackageHoister {
...
+          return true;
+        }
+        //don't bother marking dev dependencies for shallow installation
```
bdwain: Looks for any dependencies the sibling workspaces already install shallowly (due to version conflicts with the hoisted version) and marks those for shallow installation.

src/package-hoister.js
```diff
@@ -649,6 +658,81 @@ export default class PackageHoister {
...
+        const pkg = this.tree.get(w).pkg;
+        return !pkg.devDependencies || !(splitPath[1] in pkg.devDependencies);
+      });
+
+      if(isShallowDependency){
+        info.shallowPaths = [''];
+        return;
+      }
+
```
bdwain: Finally, it looks for packages that are normally hoisted but need to be shallowly installed underneath a sibling workspace because the target has its own version, and marks those for shallow installation (but not just directly under node_modules like the other ones)

src/package-hoister.js
```diff
@@ -670,17 +754,31 @@ export default class PackageHoister {
...
 
+      let shallowLocs = [];
       if (this.config.modulesFolder) {
         // remove the first part which will be the folder name and replace it with a
         // hardcoded modules folder
         parts.splice(0, 1, this.config.modulesFolder);
       } else {
         // first part will be the registry-specific module folder
         parts.splice(0, 0, this.config.lockfileFolder);
+        if(info.shallowPaths){
```
bdwain: add any packages marked for shallow installation to the main output tree

src/package-reference.js
```diff
@@ -55,15 +55,15 @@ export default class PackageReference {
...
+  locations: Array<string>;
   resolver: PackageResolver;
 
   setFresh(fresh: boolean) {
     this.fresh = fresh;
   }
 
-  setLocation(loc: string): string {
-    return (this.location = loc);
+  addLocation(loc: string): string {
```
bdwain: a package can have more than one location. For example, 3 workspaces depend on foo@2.0.0, and 2 workspaces depend on foo@1.0.0. foo@1.0.0 will be installed under both workspaces to avoid the conflict with the hoisted v2.

This was causing a bug where .bin links pointing to foo@1.0.0 would all point to a single copy of it, rather than the one installed for that package.

I also think if foo@1.0.0 has a postinstall script, it would only have run on one instance, but I never verified that. It shouldn't do that either way now though.

src/resolvers/contextual/workspace-resolver.js
```diff
@@ -19,19 +19,36 @@ export default class WorkspaceResolver extends BaseResolver {
...
     const registry = manifest._registry;
     invariant(registry, 'expected reference');
-
-    manifest._remote = {
+    let hash = '', registryRemote;
+    if(downloadedManifest && manifest.version === downloadedManifest.version){
+      registryRemote = downloadedManifest._remote;
+      hash = registryRemote.hash;
+      //override any local changes to manifest
+      Object.keys(manifest).forEach(k => k.startsWith('_') || delete manifest[k]);
```
bdwain: this was to avoid weird bugs where the local package.json differed from the remote one (for the same version).

bdwain:
@bestander any thoughts on this implementation? (Or do you have anyone else you think should weigh in?). I'm getting close to having something that is complete and would like to know sooner rather than later if there are any major implementation issues. 

Thanks!

bestander:
@bdwain, great job taking it to a working solution!

Half of it is a no-brainer, half of it need thinking like changing package-reference interface.

I’ve scrolled through the code, are all these changes necessary for the feature or is it some refactoring that could be on its own?

bdwain:
Thanks @bestander. I think all of the changes are necessary. They were all done to fix one edge case or another. Can you comment on the specific ones you’re concerned about? 

 For the package reference interface, I changed it to multiple locations because without that, the top level symlink of a workspace and the shallow installation would have the same location, and so any .bin links pointing to that package would both point to the location (which ended up being the one whose path ended up last alphabetically).  This also caused similar issues with not running post install scripts on all of the installations. 

this also fixed what I feel was incorrect behavior from before in the scenario where a package was installed in multiple locations (if it couldn’t be hoisted due to another version being hoisted). In that case, all of
The unhoisted versions would point to the same location, so for example, workspace b’s bin link for foo might point to workspace a’s installation of foo becuse they had the same location. And similar issues for post install scripts. 

I looked around and feel pretty confident that the change caused no unintended side effects. But if I misssd one id be happy to address it. But I do think the change is needed now since focused workspaces make it more likely to have multiple installations of a package and for those installations to be different.

package.json
```diff
@@ -117,6 +117,7 @@
     "release-branch": "./scripts/release-branch.sh",
     "test": "yarn lint && yarn test-only",
     "test-only": "node --max_old_space_size=4096 node_modules/jest/bin/jest.js --verbose",
+    "test-only-debug": "node --inspect-brk --max_old_space_size=4096 node_modules/jest/bin/jest.js --runInBand --verbose",
```
bdwain: i can remove this if needed but i found it very useful

src/resolvers/contextual/workspace-resolver.js
```diff
@@ -19,19 +19,40 @@ export default class WorkspaceResolver extends BaseResolver {
...
     invariant(registry, 'expected reference');
-
-    manifest._remote = {
+    let hash = '';
+    let registryRemote;
+    if (downloadedManifest && manifest.version === downloadedManifest.version) {
+      registryRemote = downloadedManifest._remote;
+      invariant(registryRemote, 'missing remote info');
+      hash = registryRemote.hash;
+      //override any local changes to manifest
```
bdwain: this is related to the comment [here](https://github.com/yarnpkg/rfcs/pull/92#issuecomment-375432881).

bdwain:
also FYI i have all old unit tests passing, so besides addressing any PR comments, the only thing left is to add in new unit tests for the focus functionality

bdwain:
I think I'm going to need to create a test repo with some workspaces in it for some new unit tests. I'll create it under my account for now, but would it make sense to move to the yarnpkg account so other people can edit it easily in the future if needed?

__tests__/commands/check.js
```diff
@@ -33,6 +33,16 @@ test.concurrent('--verify-tree should report wrong version', async (): Promise<v
   expect(thrown).toEqual(true);
 });
 
+test.concurrent('--verify-tree should work from a workspace cwd', async (): Promise<void> => {
```
bdwain: i made the change in check.js so that this test would pass. I needed it to work correctly for the new focus tests to work (or I could have avoided the check at the end of the install, but that felt wrong)

bdwain:
@bestander i think this is just about ready to go. The unit tests are all there. I did also make some changes to the main logic to fix some edge cases I missed in [this commit](https://github.com/yarnpkg/yarn/pull/5663/commits/e1e2dde176b34f03a7045bcb9de06d57aa517380) if you want to see the differences.

I am seeing one weird issue where on the [node 4 CI builds](https://circleci.com/gh/yarnpkg/yarn/15183?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), i get some errors trying to use es6 methods like Array.includes, but i notice instances of it being used in the same file that do not cause issues, which is confusing. I'm going to switch it to use indexOf for now, but it seems like i should not need to.

I also am a bit confused on one last test that's failing only in the [appveyor CI build](https://ci.appveyor.com/project/kittens/yarn/build/5330?fullLog=true), which seems to run on windows. It's my new [.bin link test](https://github.com/yarnpkg/yarn/pull/5663/files#diff-7ebb84a2ed1277abe1992607945bbd3eR203), but it passes on my local machine (a mac). I'll keep looking to see why, but if you have any suggestions I'd appreciate it. It's a slow debugging process since I need to commit to test it.

Thanks!

bdwain:
@bestander the issues i mentioned above are fixed. So I've got nothing left to do except address PR comments.

bdwain:
also if it's helpful, i can transfer [my example repo](https://github.com/bdwain/example-yarn-workspaces) and the associated npm packages (which I needed to make for the tests I added) to yarn so people can edit them if needed in the future.

bestander:
Thanks Bryan.
Sorry I’ve been phased out for a week, will review shortly

package.json
bestander: Looks all right to me

bdwain:
Sounds great! Thanks! I can work on a blog post. Should I just make a pr to https://github.com/yarnpkg/website/tree/master/_posts ?

One other question is what will the process be to release this once all 3 of those steps are done? It's not a breaking change so can it just go out in  v 1.7 or whatever the next minor release is?

arcanis:
Hey @bdwain! Thanks for submitting the PR! I'll look at it tomorrow and come back to you 👍

arcanis:
> Sounds great! Thanks! I can work on a blog post. Should I just make a pr to https://github.com/yarnpkg/website/tree/master/_posts ?

Yep, that would be perfect!

> One other question is what will the process be to release this once all 3 of those steps are done? It's not a breaking change so can it just go out in v 1.7 or whatever the next minor release is?

If there's no breaking change it should be fine in a minor :)

bdwain:
Hi @arcanis Please let me know if there's anything else you need before you look at this. I was holding off on finalizing the blog post until it was merged because if anything changes I didn't want to have to change it. 

Thanks!

bestander:
@bdwain, great job on this, sorry for a delay, we are all heads down working on specific projects.
Let me take the responsibility and merge it, I don't expect anyone to be extremely opposed to the changes.
Clicking merge button!

bestander:
Is blog post ready? I'll be happy to spread the word

bdwain:
thanks! It's not ready yet. I'll try to have it ready in the next day or so.

arcanis:
Noticed what I think is a bug, I opened #5967 to discuss it 🙂

brownbl1:
@bdwain Will a blog post for this still be written or has this stalled? I'm having a hard time discerning whether this has actually been released/published or not. Thanks!

bdwain:
yup it was released already! https://yarnpkg.com/blog/2018/05/18/focused-workspaces/

bdwain:
I see. I think focused workspaces was attempting to solve a different problem. But [yarn plug'n'play](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-plug-an-play.md) seems like it may help you solve your issues.

RDeluxe:
I'm in the same boat, I thought this could solve our problems with package isolation while creating docker images for example.

PR #5593, opened 2018-03-28
fix(owner): Fix the currently broken `yarn owner` command

rubycut:
**Summary**

* fixing currently broken owner command
* doesn't ask for npm username and password for list command
* adding basic test #3893 

**Test plan**

Yarn now correctly displays the list of package owners. Also added automated tests.

PR #5667, opened 2018-04-13
chore(request): Add an `unfiltered` option to getting full responses from npm

BYK:
**Summary**

Follow up to #5593. Looks like we use an npm-specific request header to reduce the bandwidth. That said, for some commands, we need the full responses, which was achieved by setting the `Accept: application/json` header manually. This patch introduces a more descriptive request option, `unfiltered` to standardize this usage.

See https://github.com/npm/npm-registry-client#requests for more info.

**Test plan**

Existing tests should pass.


arcanis:
Seems like it breaks the `yarn info` command

PR #5711, opened 2018-04-21
fix(publish): Fixes the path parameter when specifying a dir to publish.

### LINKED ISSUE #5678: Yarn publish ignores the folder parameter
Yarn publish task always publishes the current directory, even if the `folder` parameter is given (see [docs](https://yarnpkg.com/lang/en/docs/cli/publish/#toc-yarn-publish-folder)).

I.e., Running
`yarn publish dist/lib --skip-git-tag --new-version 0.0.1-dev`
will publish the whole current directory instead of `dist/lib` only.

Workaround would be to run publish from the directory that is to be published, but according to the docs this functionality is supported and additionally, this is how npm currently works.

Looking in the source code, it seems that the `dir` param in the `pack` method is never used:
https://github.com/yarnpkg/yarn/blob/4acfe3902a010d1673c777757dc31355f2b4fd8b/src/cli/commands/pack.js#L135-L140

rally25rs:
**Summary**

When specifying a directory as the first parameter to `publish` yarn was ignoring it and just using
`config.cwd`. This fix checks to see if the passed parameter is to a directory and if it is it sets
`config.cwd` to that directory, so that manifest loading and lifecycle script execution occur
correctly in that specified directory.

fixes #5678

**Test plan**

Added a test to `__tests__/commands/publish` that passes a path as the first parameter.


src/cli/commands/publish.js
```diff
@@ -118,6 +119,22 @@ async function publish(config: Config, pkg: any, flags: Object, dir: string): Pr
...
+    throw new MessageError(reporter.lang('tooManyArguments', 1));
+  }
+  if (!await fs.exists(dir)) {
+    throw new MessageError(reporter.lang('unknownFolderOrTarball'));
+  }
+
+  const stat = await fs.lstat(dir);
+  let publishPath = dir;
+  if (stat.isDirectory()) {
+    config.cwd = path.resolve(dir);
```
arcanis: I'm not fond of changing the `cwd` outside of the entry point, but since `publish` is a very specific feature ... Ideally I'd prefer if we were passing the publish path around.

PR #5712, opened 2018-04-22
fix(publish): Run pre and postpack lifecycle scripts on publish

### LINKED ISSUE #5707: yarn publish does not run prepack/postpack

Bug

```
$ npm publish --tag next

> pkg@6.2.5-2 prepublishOnly .
> echo prepublishOnly

prepublishOnly

> pkg@6.2.5-2 prepack .
> echo prepack

prepack

> pkg@6.2.5-2 postpack .
> echo postpack

postpack
+ pkg@6.2.5-2
$ yarn publish --tag next
yarn publish v1.6.0
[1/4] Bumping version...
info Current version: 6.2.5-3
question New version:
info Proceeding with current version: 6.2.5-3
[2/4] Logging in...
[3/4] Publishing...
$ echo prepublishOnly
prepublishOnly
success Published.
[4/4] Revoking token...
info Not revoking login token, specified via config file.
✨  Done in 6.99s.
```

Set `scripts.prepack` in `package.json`, called during `yarn pack` but not `yarn publish`.

Should call the scripts like npm does.

Node: 8.11.1
Yarn: 1.6.0
OS: MacOS High Sierra

rally25rs:
**Summary**

npm runs the pre and post pack scripts on publish, but yarn was not. Added these scripts.

fixes #5707

**Test plan**

Added a test that regex's the output to look for scripts being run.
If you look at the test, the regex is kinda weird, but I did it that way to make sure the log statements were in the correct order in the output.


src/cli/commands/publish.js
```diff
@@ -44,6 +44,7 @@ async function publish(config: Config, pkg: any, flags: Object, dir: string): Pr
   await config.executeLifecycleScript('prepublish');
   await config.executeLifecycleScript('prepare');
   await config.executeLifecycleScript('prepublishOnly');
+  await config.executeLifecycleScript('prepack');
```
arcanis: Shouldn't those lifecycles be part of the `pack` command rather than `publish`?

rally25rs: They already are, but they are in pack's `run()` function which is not called directly by `publish`. I considered moving them into the commonly called `pack()` function, but `postpack` shouldn't be executed until the output stream is written, so it makes refactoring the scripts to be called in a common spot a bit more complicated. I took the easier route here.

BYK: Shall we put that explanation in the code as a comment?

PR #5722, opened 2018-04-24
fix(console-reporter): Fix issue where tree logger was ignoring --silent/-s flag

### LINKED ISSUE #5721: --silent / -s not working for `upgrade`
Bug

Outputs a lot of text in the terminal

`yarn upgrade --silent`

Skip Yarn console logs. Only other types of logs (script output) should be printed. This is the behavior that `yarn upgrade --help` says it should have.

- Node: v9.11.1
- Yarn: 1.6.0
- macOS: 10.13.4

jxom:
**Summary**

This PR fixes #5721 where `upgrade` was not being silent when the `--silent`/`-s` flags were provided. It still printed the tree.  🖨🌳  

**Test plan**

Added two new test cases.

Manual CLI tests:

Before:
```
> yarn-local upgrade --silent
└─ left-pad@1.3.0
└─ left-pad@1.3.0
```

After:
```
> yarn-local upgrade --silent

```

### DIFF
--- a/__tests__/reporters/_mock.js
+++ b/__tests__/reporters/_mock.js
@@ -16,7 +16,7 @@ type MockCallback = (reporter: Reporter, opts: Object) => ?Promise<void>;
 export default function<T>(
   Reporter: Function,
   interceptor: Interceptor<T>,
-  prepare?: (reporter: Reporter) => any,
+  prepare?: ?(reporter: Reporter) => any,
   opts?: Object,
 ): (callback: MockCallback) => Promise<T> {
   return async function(callback: MockCallback): * {
--- a/__tests__/reporters/_mock.js
+++ b/__tests__/reporters/_mock.js
@@ -17,6 +17,7 @@ export default function<T>(
   Reporter: Function,
   interceptor: Interceptor<T>,
   prepare?: (reporter: Reporter) => any,
+  opts?: Object,
 ): (callback: MockCallback) => Promise<T> {
   return async function(callback: MockCallback): * {
     const data: MockData = {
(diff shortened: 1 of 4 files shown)


jxom:
cc/ @rally25rs

rally25rs:
To add a test for `silent` you'll have to add a way to pass options into `__tests__/reporters/_mock.js` so that you can pass down a `silent: true` flag to the reporter constructor.

I can lend a hand with that if needed.

jxom:
@BYK @rally25rs - can you take a look at those changes?

__tests__/reporters/_mock.js
```diff
@@ -16,7 +16,8 @@ type MockCallback = (reporter: Reporter, opts: Object) => ?Promise<void>;
 export default function<T>(
   Reporter: Function,
   interceptor: Interceptor<T>,
-  prepare?: (reporter: Reporter) => any,
+  prepare?: ?(reporter: Reporter) => any,
+  opts?: Object,
```
BYK: nit: this can be `extraOpts` and then you can avoid the `opts` -> `newOpts` rename.

__tests__/reporters/console-reporter.js
```diff
@@ -266,3 +266,41 @@ test('close', async () => {
     }),
   ).toMatchSnapshot();
 });
+
+test('ConsoleReporter.log is silent when isSilent is true', async () => {
```
BYK: Drop the 'ConsoleReporter.' prefix please :)

BYK: Scratch that, just realized the whole file does this.

__tests__/reporters/_mock.js
jxom: Yeah, I was thinking that too. But thought it would have been preferred to keep the argument named as `opts`. Let me know if you want to change though 😄

BYK: Nope, already merged :D

PR #5724, opened 2018-04-24
Updated contributing urls so they're not pointing to 404.

rafcioki:
**Summary**

The changes doesn't affect the code anyhow so I don't think the mandatory fields here need to be filled, it's just a change fixing 404 urls in readme and contributing files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,1 @@
-Please see https://yarnpkg.com/org/contributing/
+Please see https://yarnpkg.com/lang/en/org/contributing/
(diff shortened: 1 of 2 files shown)


rally25rs:
The yarnpkg.com site is supposed to automatically redirect you to the appropriate language. If I navigate to the original url https://yarnpkg.com/org/contributing/ then I get automatically redirected to https://yarnpkg.com/en/org/contributing/

I don't think we should need this change, and the site should be changed to redirect to `/en` when it can't detect the language.

According to https://github.com/yarnpkg/website/issues/367 this seems like it should have been fixed on the website.

BYK:
Please see: https://github.com/yarnpkg/website/issues/385

Issue #5611, opened 2018-04-03
yarn remove doesn't warn about workspace root like yarn add does (add -W flag)


rally25rs:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**

Currently if you try to `yarn add` in a workspace root, you get an error:

```
~/Projects/yarn-test 🐒   yarn add copy
yarn add v1.5.1
error An unexpected error occurred: "Running this command will add the dependency to the workspace root rather than workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).".
info If you think this is a bug, please open a bug report with the information provided in "/Users/jvalore/Projects/yarn-test/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

but `yarn remove` doesn't:

```
~/Projects/yarn-test 🐒   yarn remove copy
yarn remove v1.5.1
[1/2] Removing module copy...
[2/2] Regenerating lockfile and installing missing dependencies...
warning Resolution field "assert-plus@0.2.0" is incompatible with requested version "assert-plus@1.0.0"
warning Resolution field "assert-plus@0.2.0" is incompatible with requested version "assert-plus@^1.0.0"
warning Resolution field "assert-plus@0.2.0" is incompatible with requested version "assert-plus@^1.0.0"
warning Resolution field "assert-plus@0.2.0" is incompatible with requested version "assert-plus@^1.0.0"
warning Resolution field "assert-plus@0.2.0" is incompatible with requested version "assert-plus@^1.0.0"
success Uninstalled packages.
✨  Done in 0.78s.
```

**What is the expected behavior?**

`yarn remove` should have the same `-W` flag and behavior as `yarn add`.

**Please mention your node.js, yarn and operating system version.**

yarn 1.5.1

malcolmsgroves:
I am going to start working on this

jxom:
Hey @rally25rs, I'd love to tackle this.

PR #5737, opened 2018-04-26
Add warning when attempting to remove dependency from workspace root

jxom:
**Summary**

This PR addresses issue #5611. 

This PR basically adds a warning when trying to remove a dependency from a workspace root. When specifying the `-W`/`--ignore-workspace-root-check` it will remove the dependency from the workspace root.

**Test plan**

Before:
```
❯ yarn remove left-pad
yarn remove v1.6.0
[1/2] Removing module left-pad...
[2/2] Regenerating lockfile and installing missing dependencies...
success Uninstalled packages.
```

After:
```
❯ yarn-local remove left-pad
yarn remove v1.6.0
error Running this command will remove the dependency from the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
info Visit https://yarnpkg.com/en/docs/cli/remove for documentation about this command.
```

```
❯ yarn-local remove left-pad -W
yarn remove v1.6.0
[1/2] 🗑  Removing module left-pad...
[2/2] 📃  Regenerating lockfile and installing missing dependencies...
success Uninstalled packages.
```

### DIFF
--- a/__tests__/commands/remove.js
+++ b/__tests__/commands/remove.js
@@ -47,6 +47,20 @@ test.concurrent('throws error when package is not found', (): Promise<void> => {
   });
 });
 
+test.concurrent('remove without --ignore-workspace-root-check should fail on the workspace root', async () => {
+  await runInstall({}, 'workspaces-install-basic', async (config, reporter): Promise<void> => {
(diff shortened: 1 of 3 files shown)


arcanis:
Sounds good to me - thanks!

PR #5739, opened 2018-04-26
feat(bin): Implement an optional parameter support to `yarn bin`

arcanis:
**Summary**

This allows to get the path to a binary without making assumptions regarding the layout of the
`node_modules` folder.

**Test plan**

Added a test.

### DIFF
--- a/__tests__/commands/bin.js
+++ b/__tests__/commands/bin.js
@@ -11,13 +11,13 @@ const runBin = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, co
   return run(config, reporter, flags, args);
 });
 
-test('run bin with no arguments should return the folder where are stored the binaries', (): Promise<void> => {
+test('running bin without arguments should return the folder where the binaries are stored', (): Promise<void> => {
   return runBin([], {}, '../install/install-production-bin', (config, reporter): ?Promise<void> => {
     expect(reporter.getBufferText()).toMatch(/[\\\/]node_modules[\\\/]\.bin[\\\/]?$/);
   });
 });
 
-test('run bin with an arguments should return the location of the binary', (): Promise<void> => {
+test('running bin with a binary name as the argument should return its full path', (): Promise<void> => {
   return runBin(['rimraf'], {}, '../install/install-production-bin', (config, reporter): ?Promise<void> => {
     expect(reporter.getBufferText()).toMatch(/[\\\/]node_modules[\\\/]\.bin[\\\/]rimraf$/);
   });
--- a/__tests__/commands/bin.js
+++ b/__tests__/commands/bin.js
@@ -1,0 +1,24 @@
+/* @flow */
+
+import {run as buildRun} from './_helpers.js';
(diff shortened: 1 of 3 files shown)


__tests__/commands/bin.js
```diff
@@ -0,0 +1,24 @@
...
+import {run} from '../../src/cli/commands/bin.js';
+
+const path = require('path');
+
+const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'bin');
+const runBin = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, config, reporter): Promise<void> => {
+  return run(config, reporter, flags, args);
+});
+
+test('run bin with no arguments should return the folder where are stored the binaries', (): Promise<void> => {
```
BYK: `running bin without arguments should return the folder where the binaries are stored`

__tests__/commands/bin.js
```diff
@@ -0,0 +1,24 @@
...
+  return runBin([], {}, '../install/install-production-bin', (config, reporter): ?Promise<void> => {
+    expect(reporter.getBufferText()).toMatch(/[\\\/]node_modules[\\\/]\.bin[\\\/]?$/);
+  });
+});
+
+test('run bin with an arguments should return the location of the binary', (): Promise<void> => {
```
BYK: `running bin with a binary name as the argument should return its full path`

src/cli/commands/bin.js
```diff
@@ -16,6 +16,12 @@ export function setFlags(commander: Object) {
 
 export function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const binFolder = path.join(config.cwd, config.registries[RegistryYarn.registry].folder, '.bin');
-  reporter.log(binFolder, {force: true});
+  if (args.length === 0) {
+    reporter.log(binFolder, {force: true});
+  } else {
+    for (const arg of args) {
```
BYK: I think we should limit the argument length to 1 to keep the API surface smaller and open to future additions.

I also think we should ensure the binary exists before returning its path. Just calling `fs.realpath` on the value before returning it would probably be better for this and to ensure consistent output.

arcanis:
Updated. I prefer not to use `realpath` for now, since symlinks might be important (especially for things such as peer dependencies etc).

src/cli/commands/bin.js
```diff
@@ -16,6 +17,16 @@ export function setFlags(commander: Object) {
...
-  reporter.log(binFolder, {force: true});
+  if (args.length === 0) {
+    reporter.log(binFolder, {force: true});
+  } else {
+    const binName = args[0];
+    const finalPath = path.normalize(`${binFolder}/${binName}`);
+    if (fs.existsSync(finalPath)) {
+      reporter.log(finalPath, {force: true});
+    } else {
+      reporter.error(reporter.lang('packageBinaryNotFound', binName));
```
BYK: You need a new test to cover this case now.

BYK:
Your test fails btw.

PR #5740, opened 2018-04-26
feat(cli): Add a new --into option to `yarn node` and `yarn run`

arcanis:
**Summary**

This allow them to see their cwd changed to something different from the project directory.

Possible breaking change: Might break scripts running `yarn run something --into foo` if the `something` binary uses an option called `into`. The alternative would be to define `--into` as a global option, and require users to put it before the command name (`yarn --into /tmp node`), which doesn't sound very readable.

**Test plan**

Added an integration test.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -131,14 +131,14 @@ describe('production', () => {
 });
 
 test('yarn node should start Node', async () => {
-  const [stdoutOutput, _] = await runYarn(['node', '-p', '42']);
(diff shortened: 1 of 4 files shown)


BYK:
>The alternative would be to define --into as a global option, and require users to put it before the command name (yarn --into /tmp node), which doesn't sound very readable.

How about `yarn run --into x <cmd>`? Otherwise this looks like a breaking change that I want to avoid.

arcanis:
Withdrawing the PR for now

Issue #4297, opened 2017-09-02
.bin folder missing when using `--modules-folder`


kilianc:
**Do you want to request a *feature* or report a *bug*?**
BUG

**What is the current behavior?**
```
ENV NODE_ENV production
ENV NODE_PATH /opt/app-node-modules
ENV PATH $PATH:/opt/app-node-modules/.bin

RUN yarn --no-emoji --modules-folder $NODE_PATH --pure-lockfile
```

`.bin` is nowhere to be found. 

**What is the expected behavior?**

Behavior (as per yarn@0.24.4 and node@7.10.0) should be `.bin` folder at `/opt/app-node-modules/.bin`

**Please mention your node.js, yarn and operating system version.**

* node@8.4.0
* yarn@0.27.5

P.S. I see this behavior bumping our docker images to 8.4.0 from 7.10.0

BYK:
This looks like a duplicate of #3724. Please confirm :)

schmunk42:
I have the same issue, looks like duplicate of #3724

kilianc:
Any update on this, I have been stuck on `yarn@0.24.6` for months :(

crgt5252:
Any update on this?  We're also blocked by this.

EECOLOR:
@BYK yes, it's a duplicate of #3724

BYK:
Closing in favor of #3724.

PR #5741, opened 2018-04-26
fix(package-linker): Fix issue with .bin folder with --modules-folder flag

### LINKED ISSUE #3724: custom node_modules folder breaks .bin since yarn 0.25.4
In our setting we install all our yarn packages in a different folder then the default node_modules folder with the --modules-folder setting.
e.g.:
```
yarn --modules-folder vendor/node_modules
```

The current behaviour (version 0.24.*)  will install them in vendor/node_modules/.bin
After upgrading to version 0.25.4 it installs them incorrectly in node_modules/.bin, the packages are still installed in vendor/node_modules and the symlinks in node_modules/.bin still work.
However it seems some of them are missing, as I can't find uglifyjs binary for example in my project in that folder.

Install the packages as version 0.24.* did by installing them in vendor/node_modules/.bin

node.js v7.10.0
yarn: 0.24.5/0.25.4

voondo:
**Summary**

This PR tries to fix #3724 and #4297 where `install` was not installing binary symlinks to the correct location
when the `--modules-folder` flag was provided.

**Test plan**

Manual CLI tests:

Before:
```
> yarn --modules-folder /var/foo
> (ls /var/foo/.bin 2>1 > /dev/null && echo "found") || echo "not found"
not found
```

After:
```
> yarn --modules-folder /var/foo
> (ls /var/foo/.bin 2>1 > /dev/null && echo "found") || echo "not found"
found
```

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -436,7 +436,9 @@ export default class PackageLinker {
         topLevelDependencies,
         async ([dest, {pkg}]) => {
           if (pkg._reference && pkg._reference.location && pkg.bin && Object.keys(pkg.bin).length) {
-            const binLoc = path.join(this.config.lockfileFolder, this.config.getFolder(pkg));
+            const binLoc = this.config.modulesFolder
+              ? this.config.modulesFolder
+              : path.join(this.config.lockfileFolder, this.config.getFolder(pkg));
             await this.linkSelfDependencies(pkg, dest, binLoc);
             tickBin();
           }
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -26,7 +26,9 @@ export const runInstall = run.bind(
     const install = new Install(flags, config, reporter, lockfile);
     await install.init();
     await check(config, reporter, {}, []);
-    await check(config, reporter, {verifyTree: true}, []);
+    if (!flags.modulesFolder) { // check verify tree being broken with a custom modules folder
+      await check(config, reporter, {verifyTree: true}, []);
+    }
     return install;
   },
   [],
@@ -84,6 +86,7 @@ export function makeConfigFromDirectory(cwd: string, reporter: Reporter, flags: 
   return Config.create(
     {
       binLinks: !!flags.binLinks,
+      modulesFolder: flags.modulesFolder,
       cwd,
       globalFolder: flags.globalFolder || path.join(cwd, '.yarn-global'),
       cacheFolder: flags.cacheFolder || path.join(cwd, '.yarn-cache'),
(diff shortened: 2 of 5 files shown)


src/package-linker.js
```diff
@@ -436,7 +436,12 @@ export default class PackageLinker {
         topLevelDependencies,
         async ([dest, {pkg}]) => {
           if (pkg._reference && pkg._reference.location && pkg.bin && Object.keys(pkg.bin).length) {
-            const binLoc = path.join(this.config.lockfileFolder, this.config.getFolder(pkg));
+            let binLoc;
```
BYK: Prefer `const`:
```js
const binLoc = this.config.modulesFolder ? this.config.modulesFolder : path.join(this.config.lockfileFolder, this.config.getFolder(pkg));
```

voondo: done

voondo: eslint is unhappy because `Line 439 exceeds the maximum line length of 120`. How should I write this line without `if` and `let` ?

voondo:
Ok here is the latest state of my work but I encountered so many issues with `--modules-folder` I had to give up.

It contains a test as requested but I can't make it work (altough the manual CLI test works).
It also fixes a problem with `yarn --modules-folder /var/foo run [...]` (need also a test).
There are still some issues with `yarn check --modules-folder /var/foo --verify-tree`

TL;DR `--modules-folder` is currently too broken with yarn even with this patch, don't use it.

BYK:
@voondo how can we help you to get this merged?

voondo:
@BYK honestly I don't know. After digging a bit, I seen so many problems with this option.
By the way it has almost no test in the current code base (which could explain why it is broken).
Indeed, the only related test is `uses the modulesFolder configuration option as the first part of the path` in `__tests__/package-hoister.js`. From what I understand, it is far from sufficient.

On the other hand, in `src/`, we have dozens of occurrences of `path.join(cwd, 'node_modules'`, `path.join(config.cwd, registry.folder)`, `path.join(this.config.lockfileFolder, constants.NODE_MODULES_FOLDER)` or diferents variations which seems to refer to the same thing.

Therefore, the amount of work seems huge (especially for me who do not know well the secrets of yarn). 

Feel free to close this pull request but maybe it could help someone more courageous than me to get the work done ;)

arcanis:
Yeah, `--modules-folder` is kind of a relicate from the time where Yarn was trying to be bower-compatible, iirc, so it's not as battle-proof as we'd like. I sometime wonder whether there's value in keeping it 😕

kilianc:
@aracarie without `--modules-folder` docker workflows with mounted local volumes are broken. You mount your local dir that sometimes contains `node_modules` (standardjs, or other deps used by your IDE) and you want to use a different path in your container to avoid conflict with compiled deps.

The alternative is to `mv node_modules` after running `yarn` which is pretty hacky.

voondo:
The use with docker was also my use case. I finally "solved" it by automatically running yarn when the container is starting in development. But I agree it is hacky and would be better to have a working `--modules-foler`.
@arcanis Until it is fixed, maybe you could disable this option to not mislead us (or at least a warning)

kilianc:
Aside from being slower to re-run `yarn` at every cold start of the container it leads to all sort of side-effects in the opposite way, where now you have on your machine the deps used by the container.

At this moment, this is the best monkey-patch I found:

```dockerfile
RUN yarn --no-emoji --pure-lockfile

# https://github.com/yarnpkg/yarn/issues/4297
# https://github.com/yarnpkg/yarn/pull/5741
RUN mv ./node_modules $NODE_PATH
```

Issue #1305, opened 2016-10-20
official debian packages


marado:
**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behavior?**
No package on official Debian repositories.

**What is the expected behavior?**
Package on official Debian repositories.

**Description**

At #275 debian packages were created, and there are yarn-maintained repositories to allow users to install yarn via apt. However, it would be better still if people could install yarn directly from debian repositories.

For that to happen, an [ITP](https://www.debian.org/devel/wnpp/) should be created. I believe this will end up happening anyway, but it would be advantageous if the debian package was actually maintained by someone from upstream (I mean, whomever is currently maintaining the yarn debian repository).

This issue is not being submitted simply as a "feature request", but also to know if this is something yarn people are willing to do in the future, or if someone else should work on it.

Daniel15:
Hi @marado! 

> whomever is currently maintaining the yarn debian repository)

That's me! 😄 

Getting it into Debian would require some more work. The current Debian package is "good enough" for people to use and works well, but I don't think it's quite ready for inclusion in Debian as they have pretty strict quality requirements for packages. For one, the current package bundles all dependencies into the package itself. Debian requires dependencies to be in their own separate packages (for example, see the [npm package in Ubuntu](http://packages.ubuntu.com/xenial/npm)). Additionally, we currently don't have manpages/docs or a proper changelog for Yarn.

It's not something we'll do in the short term, but maybe in the long term!

Thanks for your interest.

simevo:
Here is the RFP https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843021

simevo:
We're making progress on this one !

The introduction with [this commit](https://github.com/yarnpkg/yarn/commit/c6bd30b4b01c2e5422d82b2568ab5f45f79350bc) of the new dependency on the inquirer module makes the matter harder because we need to package that first...

Next there's waiting for gulp to get into debian and finally implement the actual build in the curiously named makefile [debian/rules](https://anonscm.debian.org/git/pkg-javascript/node-yarnpkg.git/tree/debian/rules).

For detailed a status update see https://bugs.debian.org/843021.

Daniel15:
@simevo - Why `node-yarnpkg` rather than just `yarnpkg`? npm's package is just `npm` rather than `node-npm`, right?

simevo:
[...]

PR #5742, opened 2018-04-26
fix #5729: replace dependency on deepequal with deep-equal

### LINKED ISSUE #5729: deepequal
Hi a new dependency on deepequal was recently introduced (https://github.com/yarnpkg/yarn/commit/eab88b8268f15851b817fafe554b653a301a6747).
This is an immature module (v0.0.1) with ~300 weekly downloads and whose github repo has seen no activity since on 2014-01-31.
Is there any reason why you can't rely on deep-equal (which has 2M+ weekly downloads) ?
Thanks

simevo:
**Summary**

In eab88b8 you introduced a new dependency on [deepequal](https://yarnpkg.com/en/package/deepequal). This is an immature module (v0.0.1) with 1.1k downloads in the last 30 days and whose github repo has seen no activity since on 2014-01-31.

I propose to replace that with [deep-equal](https://yarnpkg.com/en/package/deep-equal), which has had 11.6m downloads in the last 30 days.

This is not substantial, the two should be functionally equivalent for what you need.

**Motivation**: In creating a native Debian package for yarn (#1305), each new dependency is a pain because it has to packaged separately. We try to package only Node.js modules which make sense and are not duplicates of modules we already have (and we do have deep-equal, as [node-deep-equal](https://tracker.debian.org/pkg/node-deep-equal)).

This PR was requested by @bestander.

**Test plan**

I run `yarn && yarn build && yarn test-only` with the `/usr/bin/yarn` pointing to the local yarn binary and it worked, and all tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "commander": "^2.9.0",
     "death": "^1.0.0",
     "debug": "^3.0.0",
-    "deepequal": "^0.0.1",
+    "deep-equal": "^1.0.1",
     "detect-indent": "^5.0.0",
     "dnscache": "^1.0.1",
     "glob": "^7.1.1",
--- a/src/cli/commands/install.js
+++ b/src/cli/commands/install.js
@@ -30,7 +30,7 @@ import WorkspaceLayout from '../../workspace-layout.js';
 import ResolutionMap from '../../resolution-map.js';
 import guessName from '../../util/guess-name';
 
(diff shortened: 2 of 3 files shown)


arcanis:
I'm good with the change itself, but the lockfile you've commited is bogus - it should only remove deepequal and add deep-equal.

arcanis:
Thanks!

PR #5743, opened 2018-04-26
Adds the directory that contains Node to the PATH

arcanis:
**Summary**

This diff orders Yarn to run the lifecycle scripts with a PATH configured to offer in priority the same `node` binary used to run Yarn itself. This change makes Yarn more coherent since we were previously doing this for node-gyp.

**Test plan**

Added a test to pkg-tests

### DIFF
--- a/packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
+++ b/packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
@@ -173,3 +173,15 @@ exports.readJson = async function readJson(source: string): Promise<any> {
     throw new Error(`Invalid json file (${source})`);
   }
 };
+
+exports.chmod = function chmod(target: string, mod: number): Promise<void> {
+  return fs.chmod(target, mod);
+};
+
+exports.makeFakeBinary = async function(
+  target: string,
+  {output = `Fake binary`, exitCode = 1}: {|output: string, exitCode: number|} = {},
+): Promise<void> {
+  await exports.writeFile(target, `#!/bin/sh\necho "${output}"\nexit ${exitCode}\n`);
+  await exports.chmod(target, 0o755);
+};
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -141,7 +141,11 @@ export async function makeEnv(
   // split up the path
   const envPath = env[constants.ENV_PATH_KEY];
   const pathParts = envPath ? envPath.split(path.delimiter) : [];
-
+  
+  // Include the directory that contains node so that we can guarantee that the scripts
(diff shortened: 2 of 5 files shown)


packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
```diff
@@ -173,3 +173,15 @@ exports.readJson = async function readJson(source: string): Promise<any> {
     throw new Error(`Invalid json file (${source})`);
   }
 };
+
+exports.chmod = function chmod(target: string, mod: number): Promise<void> {
+  return fs.chmod(target, mod);
```
BYK: Why is the return type for this, is a promise?

packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
```diff
@@ -173,3 +173,15 @@ exports.readJson = async function readJson(source: string): Promise<any> {
...
+};
+
+exports.makeFakeBinary = async function(
+  target: string,
+  {output = `Fake binary`, exitCode = 1}: {|output: string, exitCode: number|} = {},
+): Promise<void> {
+  await exports.writeFile(target, `#!/bin/sh\necho "${output}"\nexit ${exitCode}\n`);
```
BYK: You should probably escape `output` properly.

src/util/execute-lifecycle-script.js
```diff
@@ -142,6 +142,10 @@ export async function makeEnv(
   const envPath = env[constants.ENV_PATH_KEY];
   const pathParts = envPath ? envPath.split(path.delimiter) : [];
 
+  // Include the directory that contains node so that we can guarantee that the scripts
+  // will always run with the exact same Node release than the one use to run Yarn
+  pathParts.unshift(path.dirname(process.execPath));
```
BYK: We should probably assign `path.dirname(process.execPath)` to a variable at this point since it is calculated over and over below.

packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
arcanis: I think it's fine. These utilities aren't used with user input or anything really complex, so I prefer to keep them simple.

packages/pkg-tests/pkg-tests-core/sources/utils/fs.js
arcanis:
```
const fs = require('fs-extra');
```

fs-extra exposes the fs functions but make them promise-ready.

src/util/execute-lifecycle-script.js
arcanis: Yep, but I already plan to refactor this part in a followup!

Issue #5654, opened 2018-04-11
Competing lockfiles create poor UX


jmorrell:
NB: I'm creating this issue in the yarn repo, but this is actually a shared issue between yarn and npm.

With the release of npm 5 back in May, the Node ecosystem now has two lockfile-based package managers. This was overall a win for users, and it has been good to see competition in this space.

However now that there are two competing lockfile formats this can create a new problem for users, especially beginning developers.

When npm 5 came out, Heroku added a [new failure if an application was submitted with both npm and yarn lockfiles](https://github.com/heroku/heroku-buildpack-nodejs/blob/master/lib/failure.sh#L77-L94). In the past few months **this has become the most common reason Node builds fail on Heroku** and these failures account for ~10-12% of all Node build failures on the platform. **Thousands of developers hit this every month**.

It's surprisingly easy to end up with both in your repository. Even as an experienced developer I've found myself running the wrong tool for a specific project and having to catch myself before commiting. For an inexperienced developer building their first server / react / angular project who might not understand what a package manager, lockfile, or git repo is, this can be highly confusing.

Neither tool will give a warning or error if the other lockfile exists:

```
❯ ls *lock*   
ls: *lock*: No such file or directory
                                                                                                                                                         
❯ npm --version
5.8.0
                                                                                                                                                         
❯ yarn --version
1.5.1

❯ npm install
npm notice created a lockfile as package-lock.json. You should commit this file.

added 659 packages from 437 contributors in 10.553s
                                                                                                                                                         
❯ yarn install  
yarn install v1.5.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 7.67s.
                                                                                                                                                         
❯ ls *lock*          
package-lock.json yarn.lock
```

This is likely especially true for Yarn users where most of the documentation on the web instructs users to `npm install`.  Users who copy + paste commands from docs or Stack Overflow are likely to end up here.

I've spoken to @zkat and @iarna at npm and @arcanis on yarn, and all agreed that this is an issue that should be addressed, but there was not full agreement on how. Ideally this issue prompts discussion and both tools can agree on how we can help users here.

I've compiled a list of potential mitigations that have been suggested to me:

#### Do nothing

[...]

Issue #5419, opened 2018-02-27
extraneous files always checks/cleans node_modules, ignoring --modules-folder specified folder


rally25rs:
**Do you want to request a *feature* or report a *bug*?**
🐞 

**What is the current behavior?**
Yarn checks for and cleans extraneous files in `node_modules`.

When `yarn install --modules-folder somewhere_else` is run, the root `node_modules` dir is still cleaned instead of `somewhere_else`

**If the current behavior is a bug, please provide the steps to reproduce.**

```
~/Projects/yarn-test 🐒   yarn init -y
yarn init v1.3.2
warning The yes flag has been set. This will automatically answer yes to all questions which may have security implications.
success Saved package.json
✨  Done in 0.04s.

~/Projects/yarn-test 🐒   yarn add left-pad
yarn add v1.3.2
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
success Saved 1 new dependency.
└─ left-pad@1.2.0
✨  Done in 0.35s.

~/Projects/yarn-test 🐒   ls node_modules/
left-pad

~/Projects/yarn-test 🐒   yarn install --modules-folder ./dist/node_modules
yarn install v1.3.2
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
✨  Done in 0.13s.

~/Projects/yarn-test 🐒   ls node_modules/
<empty>
```

**What is the expected behavior?**

yarn should use the folder specified by `--modules-folder` when checking directories for extraneous files. The previous `node_modules` dir should not be deleted.

**Please mention your node.js, yarn and operating system version.**

yarn v1.3.2

**Research notes**

`PackageLinker` loops over `this.config.registryFolders` to use as folders to check for extraneous files:

```
    const findExtraneousFiles = async basePath => {
      for (const folder of this.config.registryFolders) {
```

however the folder specified by `modulesFolder` is not added to that list.

In `config.js`:

```
  _init(opts: ConfigOptions) {
    ...
    if (this.modulesFolder) {
      this.rootModuleFolders.push(this.modulesFolder);
    }

    ...

    for (const key of Object.keys(registries)) {
      ...

      this.registryFolders.push(registry.folder);
[...]

PR #5745, opened 2018-04-27
feat(import): convert package-lock.json to yarn.lock

imsnif:
This is the first step to addressing: https://github.com/yarnpkg/yarn/issues/5654

## Overview (TLDR)
At the moment, one is able to `yarn import` using `node_modules` as the dependency-tree state to create a `yarn.lock` file.
The feature created by this PR allows one to `yarn import` using `package-lock.json` as the dependency-tree state. When one enters the `yarn import` command, if a `package-lock.json` file exists, yarn would attempt to import from it. If not, it would import from `node_modules` normally.

## How does this work?
A logical dependency tree is created from the `package-lock.json` file, then yarn walks through the dependencies as normal, using the logical tree to determine the package details. Often, this tree will not include the full information needed by yarn (eg. package `shasum`), and for this reason the package manifest is downloaded from the registry. This makes the process a little lengthy, but not terribly so. I deem this acceptable, seeing as this is not an everyday process. Would love to hear your thoughts if you disagree.

I opted not to fall back on (or search initially for) the manifest in node_modules, because I felt this would be error prone and introduce non-trivial complexity (eg. corrupted node_modules).

## Limitations
Rarely, a `yarn.lock` would not be 100% logically identical to the `package-lock.json` from which it was created. In all incidents I encountered, this happens because of built-in limitations in `yarn`.
An example is two different packages with the same semver range string (eg. `^2.0.0`) being resolved in `package-lock.json` to two different versions (eg. `2.0.1` and `2.0.2`).
Another example is bundled dependencies.

I encourage the readers to test things for themselves (if there is interest, I have a script that logically compares yarn/npm trees - I can clean it up and post it somewhere).

## Open Questions / Issues
1. This PR only adds a `package-lock.json` import feature and does not address the full concerns raised in the issue linked above. I feel this is a good first step, after which we can continue discussing whether `yarn install` should issue a warning/error if it finds a `package-lock.json` file, and whether `import` should delete the `package-lock.json` file after successfully importing. As well as the myriad of other concerns raised there.
2. The careful reader would notice that there is a slight discrepancy in the import types. By nature when importing from `node_modules`, the user would end up with a fully installed `node_modules` folder consistent with their `yarn.lock` file. This is not the case when importing from `package-lock.json`. One must `yarn install` after `yarn import`ing from a `package-lock.json`. In my eyes, this is a desired behaviour for the conversion (which I see more as a 'surgical fix') even if it is not such a desired behaviour for the `import` command. I chose to give more weight to the former, but am of course open to different approaches. 
3. Continuing point 2, an empty `node_modules` folder is created when importing from `package-lock.json` as mentioned here: https://github.com/yarnpkg/yarn/issues/5419

## Tests
In the tests for this feature I copied the relevant cases from the existing import tests. I did not (as I initially intended) add tests that make a logical comparison of the created `yarn.lock` against the originating `package-lock.json`.
The existing cases use file snapshots, and I felt this would be more coherent upon failure. Would love to hear differing opinions though.


imsnif:
@iarna, @zkat - if you'd be willing to provide your input I'd very much appreciate it.

zkat:
Mostly bluebird, through `cacache` and `pacote` and such, yeah:

```
148K    node_modules/readable-stream
184K    node_modules/cacache
196K    node_modules/pacote
208K    node_modules/tar
216K    node_modules/smart-buffer
228K    node_modules/iconv-lite/encodings/tables
256K    node_modules/bluebird/js/release
288K    node_modules/es6-promise/dist
340K    node_modules/iconv-lite/encodings
376K    node_modules/socks
380K    node_modules/es6-promise
396K    node_modules/iconv-lite
432K    node_modules/bluebird/js/browser
688K    node_modules/bluebird/js
704K    node_modules/bluebird
6.1M    node_modules
6.1M    total
```

And you really don't need pacote here: you should be able to get the package manifest using Yarn's built in whatever-the-thing-is. You should be able to get by with only adding `npm-logical-tree` (which will only add ~28K or so to the unbundled version. It's basically negligible.)

imsnif:
I basically agree with everything.
My original intent in using `pacote` was to make it easier to get a generic "logical manifest tree" from a package.json+any-]lockfile combination eventually. Maybe I was getting ahead of myself. :)
In any case, I only noticed the increased build size after submitting this - so yeah, it's definitely not justifiable. I'll whip up a version using the resolvers instead of pacote in the next few days.

While I do this - are there any other unrelated concerns?

zkat:
(for the record, the reason pacote is so big is because you are literally pulling in npm's entire networking, caching, and package resolution layer. You're basically pulling in half the npm installer minus the tree calculation and reading logic. pacote does a lot of heavy lifting for being a tiny api :P)

src/cli/commands/import.js
```diff
@@ -276,12 +307,45 @@ export class Import extends Install {
     this.resolver = new ImportPackageResolver(this.config, this.lockfile);
     this.linker = new PackageLinker(config, this.resolver);
   }
-
+  createLogicalDependencyTree(packageJson: ?string, packageLock: ?string): LogicalDependencyTree {
+    try {
+      if (!packageJson || !packageLock) {
+        throw new Error('files missing');
```
rally25rs: this string should probably come from `reporter.lang` and maybe report what files are expected.

Actually, since their existence is already checked around line 337-ish, do these even need to be 'nullable' types? Although I imagine Flow complains if you change `?string` to just `string`

src/cli/commands/import.js
```diff
@@ -276,12 +307,45 @@ export class Import extends Install {
...
       throw new MessageError(this.reporter.lang('lockfileExists'));
     }
-    await verifyTreeCheck(this.config, this.reporter, {}, []);
+    const {packageJson, packageLock} = await this.getExternalLockfileContents();
+    const importSource =
+      packageJson && packageLock && semver.satisfies(nodeVersion, '>=5.0.0') ? 'package-lock.json' : 'node_modules';
+    if (importSource === 'package-lock.json') {
+      this.reporter.info(this.reporter.lang('importPackageLock'));
+      const tree = this.createLogicalDependencyTree(packageJson, packageLock);
+      if (this.resolver instanceof ImportPackageResolver) {
```
rally25rs: Is there a known case where `resolver` isn't `ImportPackageResolver`? If so, what would happen? Since the package-lock tree won't be set on the resolver, should this check maybe be part of `packageJson && packageLock && semver.satisfies(nodeVersion, '>=5.0.0') ? 'package-lock.json' : 'node_modules';` to prevent it from taking this code path? (if resolver is not instanceof ImportPackageResolver then we build the logical tree and don't use it)

src/cli/commands/import.js
```diff
@@ -150,28 +153,51 @@ class ImportResolver extends BaseResolver {
...
+    return manifest;
+  }
+
   async resolve(): Promise<Manifest> {
-    const {name} = normalizePattern(this.pattern);
-    let cwd = this.getCwd();
-    while (!path.relative(this.config.cwd, cwd).startsWith('..')) {
-      const loc = path.join(cwd, 'node_modules', name);
-      const info = await this.config.getCache(`import-resolver-${loc}`, () => this.resolveLocation(loc));
+    if (this.request instanceof ImportPackageRequest && this.request.fixedVersionPattern) {
```
rally25rs: Could we refactor these cases into separate functions for readability/clarity? something like

```
async resolve(): Promise<Manifest> {
  if (this.request instanceof ImportPackageRequest && this.request.fixedVersionPattern) {
    return await this._resolveFromFixedVersions();
  } else {
    return await this._resolveFromNodeModules();
  }
}
```

imsnif: Yes, I like it!

src/cli/commands/import.js
imsnif: Yeah, this is some `flow` acrobatics. I don't like it either.
I thought not to bother with `reporter.lang` in this case because it will be caught by the below try/catch block by definition. I'll try to work out a better way.

src/cli/commands/import.js
imsnif: More `flow` acrobatics, I'm afraid. :/ I've seen this done elsewhere in the file though. I'm not sure if adding that check below will satisfy `flow`, but can give it a shot.

src/cli/commands/import.js
arcanis: Change `createLogicalDependencyTree` to only accept non-nullable types, then use `invariant` calls before calling it at line  340 to instruct flow that they cannot be null. Something like this:

```
invariant(packageJson, 'packageJson should exist');
invariant(packageLock, 'packageJson should exist');
```

That being said, the condition at line 336 should be enough - maybe the fact that you're storing this inside a variable is throwing Flow off. Maybe you can try to split it?

```
let importSource = 'node_modules';
if (packageJson && packageLock && semver.satisfies(nodeVersion, '>=5.0.0')) {
    importSource = 'package-lock.js';
}
```

src/cli/commands/import.js
arcanis: Just use a [typecasting expression](https://flow.org/en/docs/types/casting/), or an `invariant` 🙂

imsnif:
Thanks for the input @rally25rs + @arcanis!
I'll batch these up with any other changes @rally25rs has from testing this.

rally25rs:
@imsnif it looks like this package causes an error:

```
  "dependencies": {
    "watchify": "3.2.1"
  }
```

```
yarn import v1.6.0
info found npm package-lock.json, converting to yarn.lock
warning watchify > browserify > glob > minimatch@1.0.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning watchify > browserify > glob > graceful-fs@3.0.11: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
warning watchify > browserify > glob > graceful-fs > natives@1.1.3: This module relies on Node.js's internals and will break at some point. Do not use it, and update to graceful-fs@4.x.
error An unexpected error occurred: "should have a resolved reference".
info If you think this is a bug, please open a bug report with the information provided in "yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/import for documentation about this command.
```

Using npm v5.8.0

In package-lock.json it looks like a bunch of packages depend on `safe-buffer@5.1.1` but the only entry for that package at that version is

```
        "safe-buffer": {
          "version": "5.1.1",
          "bundled": true
        },
```

It looks like `fsevents` bundles `safe-buffer`, and then all the other non-bundled packages under `fsevents` that require `safe-buffer` just use it's bundled version.

BYK:
Yarn doesn't handle bundledDependencies properly, can this be the reason?

Theory: since yarn doesn't handle bundled dependencies, it expects these packages in the cache but they are not there so it fails.

imsnif:
Nice find!

@rally25rs, @BYK - So after some digging:
This also happens when importing the same package from `node_modules`.
It's a bug that is due to this line: https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L208
Briefly, this is due to some assumptions made by `import` about version ranges and fixed versions: In the above method, the `PackageRequest` gets the highest semver range matching its *pattern* but that range does not satisfy the fixedVersion - which is not a range in this case, so it finds no manifest.
This happens in this case because there is another `safe-buffer` package here that matches this range: `5.1.2`.

I fixed it locally for this test-case by overriding the `resolveToExistingVersion` method in `import`. Will work on it today and should have this patched up in a few hours (will also include the flow fixes).

imsnif:
I fixed some of the unnecessary conditionals after the above comments from @rally25rs and @arcanis. I went in a bit of a different direction when creating the dependency tree. In my eyes, this is clearer. What do you think?

I also refactored the `resolve` method in `ImportResolver` as @rally25rs suggested.

Finally, I issued a fix for @rally25rs's bug discussed above. Since this bug happened both when importing `package-lock.json` and `node_modules`, I also added tests for both cases. In the fixtures, I mocked it down to a smaller dependency tree than the full `watchify` module. Manually I tested importing from an npm installation of `watchify@3.2.1` and it works.
The bug is a little involved, so I'll try to explain it as best I can:

It essentially has to do with `patterns` in the `resolver` being overridden.

So say we have this pattern in `resolver.patterns`:
```
{
  'depA@^1.0.0': Manifest {
     name: 'depA',
     version: '1.0.4'
   }
}
```
When we get to `PackageRequest.find`, we look through the resolved patterns to see whether the manifest that we have matches a version we already have in `resolver.patterns`: https://github.com/yarnpkg/yarn/blob/05080da817ec33359615ac5f35ab9d51793e5d50/src/package-request.js#L239
Now we get the request `depA@^1.0.4` pointing at the same manifest in our patterns (version `1.0.4`). We add it to a delayed queue and sort it out later in order not to create duplication: https://github.com/yarnpkg/yarn/blob/05080da817ec33359615ac5f35ab9d51793e5d50/src/package-request.js#L244-L247
The problem is, that we make the assumption that said `Manifest` will still be in `resolver.patterns` when we sort it out later here: https://github.com/yarnpkg/yarn/blob/05080da817ec33359615ac5f35ab9d51793e5d50/src/package-request.js#L208
In the normal course of an installation, this assumption works. When importing from a state of fixed versions, we can sometimes have an identical pattern resolving to a different manifest (one that does not satisfy the queued range). In this case, eg. `depA@^1.0.0 ===> version: 1.0.3`. And thus, we would throw here: https://github.com/yarnpkg/yarn/blob/05080da817ec33359615ac5f35ab9d51793e5d50/src/package-request.js#L209
I fixed this by overriding the `resolveToExisting` `PackageRequest` method for `ImportPackageRequest`. It essentially does the same as the overridden original method, but if it does not find a resolved pattern, it resolves itself to its own manifest.

Would love to hear additional comments on these fixes or other issues.

rally25rs:
With the latest changes, I get some big differences (like half the packages are missing) if I `yarn install` and compare that yarn.lock against one that is the result of `npm i && yarn import`.

PR #5065, opened 2017-12-08
Remove skip updating symlinks

rafaeldamasceno:
**Summary**

Remove the existing skip of updating symlink present in `cli/commands/global.js`. Fixes #4941.

**Test plan**

As detailed in #4941.

PR #5751, opened 2018-04-28
Remove logic to skip updating of existing symlinks

### LINKED ISSUE #4941: Global binary path symlink not updated on package upgrade
Bug

Yarn does not update changed global package binary paths when upgrading a package.

I changed the binary path in my `package.json` from:
```javascript
// v0.2.10
"bin": {
  "dodgem": "index.js"
}
```
to:
```javascript
// v0.2.12
"bin": {
  "dodgem": "src/app.js"
}
```

After upgrading the package, the symlink to the binary is not updated to reflect the new path in `package.json`.

**Steps to reproduce**
```sh
$ yarn global add dodgem@v0.2.10
success Installed "dodgem@0.2.10" with binaries:
      - dodgem
✨  Done in 10.39s.


$ dodgem --version
0.2.10


$ readlink /usr/local/bin/dodgem
../../../Users/jamie/.config/yarn/global/node_modules/dodgem/index.js


$ yarn global upgrade dodgem@latest
success Saved lockfile.
└─ dodgem@0.2.12
✨  Done in 1.89s.


$ dodgem --version
fish: Unknown command 'dodgem'


$ readlink /usr/local/bin/dodgem
../../../Users/jamie/.config/yarn/global/node_modules/dodgem/index.js
```

Uninstalling and re-installing the package shows the correct symlink:
```sh
$ yarn global remove dodgem
success Uninstalled packages.
✨  Done in 1.67s.


$ yarn global add dodgem
success Installed "dodgem@0.2.12" with binaries:
      - dodgem
✨  Done in 11.32s.


$ readlink /usr/local/bin/dodgem
../../../Users/jamie/.config/yarn/global/node_modules/dodgem/src/app.js
```

For the symlink to be updated to match the changed binary path.

###### Environment Versions
```
Node   8.9.1
Yarn   1.3.2
macOS  10.13.1
```

wolfib:
**Summary**

This removes the current logic which skips the updating of existing symlinks. Fixes #4941

The PR supersedes #5065 by adding tests.

**Test plan**

Added an automated test which updates the 'dodgem' package for which this bug was originally reported and checks whether the symlink is updated.

There might be a simpler way to test this, or maybe there is another package (with fewer dependencies) which could be used for testing.


wolfib:
It would be great if someone could point me in the right direction to get the ci tests to pass. 

The tests pass locally on Mac OS and on a fresh Ubuntu installation. I did not manage to understand what is wrong from the error messages alone.  Is there a way to recreate the circleCI environment so that I can reproduce the errors?

arcanis:
Closing/reopening to rerun the tests

BYK:
It says `Error: Couldn't find package "chalk@^2.3.0" required by "dodgem@v0.2.10" on the "npm" registry.` which means you may want to add the generated cache files into your commit or use a smaller dependency as you have suggested.

Btw. thank you very much for working on this!

wolfib:
I created an NPM package called `dummy-for-testing-changed-path`. It has no dependencies, but a binary path which changes between versions. This package has no other function and is only used for testing that yarn handles binary symlinks correctly.

Initially, the test was not only checking the existence of the binary path after the version upgrade, but also using `fs.realpath()` to make sure that the path after the upgrade is different from the one before the upgrade. But because of the way symlinks work in Windows, this part of the test was failing for the Windows build.

Therefore the test now only checks that after the version upgrade the binary path still exists. This passes on Linux/MacOS as well as Windows and fails for the unpatched yarn.

Issue #5048, opened 2017-12-06
Yarn ignores no_proxy environment variable when proxy property set


KimTaehee:
**Do you want to request a *feature* or report a *bug*?**
Both

**What is the current behavior?**
If `proxy` and `http-proxy` are set, `NO_PROXY` and `no_proxy` environment variables are ignored.

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Set `proxy` and `https-proxy` config and can find from `yarn config list`
```
...
info npm config
{ proxy: 'http://some.proxy.host:8080/',
  'https-proxy': 'http://some.proxy.host:8080/',
...
```
2. Set `NO_PROXY` or `no_proxy` environment variable like `$ no_proxy="some.url.private.repo"`
3. Run `yarn install`
4. `proxy` and `https-proxy` work but `no_proxy` environment variable is ignored.

**What is the expected behavior?**
`NO_PROXY` or `no_proxy` environment variable is applied. 

**Please mention your node.js, yarn and operating system version.**
Ubuntu 16.04
yarn 1.3.2

rally25rs:
Yes. This happens because the env vars `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` and handled by the [Request](https://github.com/request/request) library.

However, `request` ignores the env vars if a proxy is passed to it. Yarn finds a `proxy` in its config and sends it to `request` which causes it to ignore the `NO_PROXY` env var.

There is some discussion of this behavior here https://github.com/request/request/pull/1195 and it looks like NPM tackled the same issue at some point.

A workaround would be to set your proxies as env vars instead of in yarn config.

KimTaehee:
Thanks for your answer.
But it's weird to users and They can expect `no_proxy` env var works. I think argument like `--no-proxy <hosts>` or notice about this workaround can solve this problem. As https://github.com/yarnpkg/yarn/pull/1564/ did.

wvandeneede:
npm (5.8.0-next.0) just merged the no-proxy setting in .npmrc: https://github.com/npm/npm/pull/19157
Something similar possible with yarn?

rally25rs:
tagging this as a feature request to add a `no-proxy` config setting. Anyone want to put together a PR for this?

glausmichael:
Please review my pull request to add no-proxy support for config file: https://github.com/yarnpkg/yarn/pull/5757

Azaradel:
[...]

PR #5757, opened 2018-04-30
feat(request): add noProxy setting for configuration (#5048)

glausmichael:
**Summary**

No

Proxy exclusion is currently not possible in configuration file. If proxy is defined in configuration file, no proxy exclusion can therefore be specified at the moment, because environment variable is not taken into account in this case.

**Test plan**

Added test 'RequestManager.request with no_proxy option' to check both http and https behavior with proxy exclusion.
Tested with corporate proxy by calling yarn install with local repository.
I had to adapt the test 'RequestManager.execute timeout error with maxRetryAttempts=1' because the string port is not a valid port for a url.

### DIFF
--- a/__tests__/util/request-manager.js
+++ b/__tests__/util/request-manager.js
@@ -9,6 +9,7 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;
 
 const net = require('net');
 const https = require('https');
+const http = require('http');
 const path = require('path');
 
 test('RequestManager.request with cafile', async () => {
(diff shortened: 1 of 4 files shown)


arcanis:
I got this problem a few days ago, so I sympathize 😄 

A workaround is to set `YARN_PROXY= YARN_HTTPS_PROXY=` in the environment. It might also work with the yarnrc with something like this (haven't actually tested this one):

```
env:
  YARN_PROXY ""
  YARN_HTTPS_PROXY ""
```

As for your PR, I think ideally the best would be to support `--no-proxy` and `--no-https-proxy`, but I think Commander doesn't support it well enough :/

glausmichael:
To clarify, this PR should support adding a configuration like `no-proxy="127.0.0.1,mydomain.com"` in .yarnrc file as it is requested by https://github.com/yarnpkg/yarn/issues/5048 as I understood. 

My change just ensures that configuration is loaded from configuration file and checks if the target url is covered by this no-proxy setting. The new file proxy-exclusion.js is mainly the same implementation as the implementation in the underlying library is using to exclude a url from proxy. (https://github.com/request/request/blob/master/lib/getProxyFromURI.js)

Because request library just has a "proxy" config option but no "no-proxy" option, I see the only way to suppress proxy by passing an empty string to the request library. (as it was already done before if proxy was set to false in yarn)

This code is needed also in case of additional command line options I think.

dicko2:
Any progress on this?

paztis:
what is the current status ?
On my side I wait for it since a long time.
Due to a corporate nexus registry we have to define the proxy / https-proxy to null to let it be accessible, but all the postinstall fetchs always fail (node-sass, puppeteer, ...)

We don't like to ply with env var as we're on ayarn workspace where a lot of sub commands will try to access to the nexus.

paztis:
Any news ? 
If this is rejected perhaps you can change the status ?

cobolbaby:
Any news ?

PR #5771, opened 2018-05-02
Show error message to user when publishing fails

sth:
**Summary**

When publishing a package with `yarn publish` fails, yarn catches the error and shows "Couldn't publish package.". That's not very useful to diagnose the problem.

With this change, it additionally shows the message of the caught error, for example for a truncated `.tgz` file:

> error Couldn't publish package: "https://registry.yarnpkg.com/somepackage: unexpected end of file"

**Test plan**

No actual functionality changed, only the error message got expanded. The tests still pass.

### DIFF
--- a/src/cli/commands/publish.js
+++ b/src/cli/commands/publish.js
@@ -114,7 +114,7 @@ async function publish(config: Config, pkg: any, flags: Object, dir: string): Pr
       body: root,
     });
   } catch (error) {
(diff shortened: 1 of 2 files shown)


arcanis:
lgtm 👍

edmorley:
Thank you for this - we've lately had a number of strange failures in CI that were impossible to debug with the previous lack of error message! :-)

PR #5788, opened 2018-05-05
Excluded shebang for lockfile package

### LINKED ISSUE #5719: Make @yarnpkg/lockfile a proper Node package

Bug

One `index.js` file for all.

`@yarnpkg/lockfile` doesn't advertise itself as a CLI tool, but it contains a shebang. The expected behavior would be providing a proper entry point for using as a `package` (so it doesn't require special treatment from bundlers) and a separate `bin` file handling CLI scenarios.

the-spyke:
**Summary**

Fixes #5719.

**Test plan**

1. Run `yarn run build-dist`.
2. Check that `artifacts/yarn-${version}.js` still has a shebang.
3. Check that `packages/lockfile/index.js` has no shebang.

### DIFF
--- a/scripts/build-webpack.js
+++ b/scripts/build-webpack.js
@@ -40,6 +40,7 @@ const compiler = webpack({
     new webpack.BannerPlugin({
       banner: '#!/usr/bin/env node',
       raw: true,
+      exclude: /lockfile/
     }),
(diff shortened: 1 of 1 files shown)


arcanis:
Thanks!

PR #5792, opened 2018-05-06
Fix #5517: Backlashes in output fail for escape sequences in Windows

### LINKED ISSUE #5517: Backslashes in output from yarn fail when they match standard escape sequences in Windows
The error reporting seems to have an escaping issue with folder names that start with 'n' on Windows. For example, when printing an error from the ```yarn start``` command, it outputs the directory :

```
Directory: C:\root\name
```
as:
```
Directory: C:\\root\\
ame
```

This is on Windows 10 and with yarn v1.5.1

antrancs:
This is my very first PR. So if there's any suggestions or if I'm missing anything, please advise 😀

**Summary**
This is the fix for [#5517](https://github.com/yarnpkg/yarn/issues/5517) where backlashes in the output in Windows are broken if the directory path contains "\n". The **stringifyLangArgs** method has been modified so that "\n" will only be replaced by EOL if it is not preceded by a "\\". 

**Test plan**
The test project path is C:\\Users\\An\\Documents\\Projects\\oss\\name\
Before
```
error An unexpected error occurred: "C:\\Users\\An\\Documents\\Projects\\oss\ 
ame\\package.json: Unexpected token U in JSON at position 139". 
```

After 
```
error An unexpected error occurred: "C:\\Users\\An\\Documents\\Projects\\oss\\name\\package.json: Unexpected token U in JSON at position 139". 
```

Also added automated tests.

### DIFF
--- a/__tests__/reporters/base-reporter.js
+++ b/__tests__/reporters/base-reporter.js
@@ -1,7 +1,8 @@
 /* @flow */
 /* eslint yarn-internal/warn-language: 0 */
 
-import BaseReporter from '../../src/reporters/base-reporter.js';
+import BaseReporter, {stringifyLangArgs} from '../../src/reporters/base-reporter.js';
+import {EOL} from 'os';
 
 test('BaseReporter.getTotalTime', () => {
   const reporter = new BaseReporter();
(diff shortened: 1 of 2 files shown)


BYK:
If you can add a test case for it, I think the PR is good to go :)

Thanks a lot!

antrancs:
@BYK : Thanks for the review :D. I just updated function and added tests for it.

Issue #3932, opened 2017-07-13
Yarn publish command not reading password from .yarnrc


bchoddny:
**Do you want to request a *feature* or report a *bug*?**
Can yarn publish step read password from .yarnrc file? I would like to publish my package as part of my continuous integration process
**What is the current behavior?**
yarn doesn't read password from .yarnrc file, but it does reads username. Is there any other alternative since the continuous integration process cannot work with interaction.

**What is the expected behavior?**
Read credentials from yarnrc file
**Please mention your node.js, yarn and operating system version.**
yarn 0.27.2
node 6.9.4
mac 10.11.3

Daniel15:
Looking at the code, it seems like `_username` and `_password` should work: https://github.com/yarnpkg/yarn/blob/3bfa1e3d6303c58aed9660e1a2af3a6d6f030bce/src/registries/npm-registry.js#L224-L243

You could try an auth token in `_authToken` instead. Log in to npm or Yarn then copy the auth token it saves to the `.yarnrc` or `.npmrc`.

bchoddny:
Either password or _password didn't work.

How do I get the authentication token, can you please provide me some steps. I tried the command:
"yarn login", but it didn't provide any authentication token?

Daniel15:
Try `npm login`, enter your username and password, and then look in `~/.npmrc`. I don't think Yarn does auth the same way, but it does support npm auth tokens. Using an auth token is more secure as it does not expose your password (so there's no possibility for your password to leak)

gutenye:
I have same problem, after add `_authToken` to `~/.yarnrc` got this error:

```
error An unexpected error occurred: "Unknown token 8:1 in /Users/guten/.yarnrc".
```

akloeber:
@gutenye I ran into this problem as well. Because of the underscore it needs to be defined with surrounding quotes like that:
```
"_authToken" <TOKEN>
``` 
Or you can set it via `yarn config set _authToken <TOKEN>` which does it for you.

akloeber:
I have the same problem as well (`_authToken` in this case).

After doing some debugging it looks like `NpmRegistry#config` only contains values from `.npmrc` when `NpmRegistry#getAuth` is called. So, it only works if `_authToken`, `_auth`, `_password` etc. are defined in `.npmrc`. 

[...]

Issue #3683, opened 2017-06-20
Extract loadConfig from YarnRegistry/NpmRegistry responsibility (reading .npmrc/.yarnrc configs)


bestander:
**Do you want to request a *feature* or report a *bug*?**
code improvement

YarnRegistry extends NpmRegistry.
YarnRegistry reads .yarnrc into config object and NpmRegistry reads .npmrc into config object.

In some cases YarnRegistry is used for requests in other NpmRegistry.
This makes it confusing to track down when settings are used from .npmrc and when from .yarnrc and this causes issues like https://github.com/yarnpkg/yarn/issues/3612

bestander:
BTW if you want to make Yarn work better with .npmrc speak up and send a PR.

yyjazsf:
I think:  when use yarn
`project's  .yarnrc` > `project's .npmrc` > `global .yarnrc` > `global .npmrc`

yyjazsf:
This is also worth considering.
yarn only use .yarnc.
npm only use .npmrc
if someone use yarn, they should use yarn's file.  It's no friendly

paul-soporan:
Closing as in v2 we've decided to only read `.yarnrc.yml` files.

https://yarnpkg.com/getting-started/migration

PR #5812, opened 2018-05-12
Read config from ~/.yarnrc

otaku:
Given the following:
```
yarn config set @scope:registry https://registry.npmjs.org/
yarn config set //registry.npmjs.org/:_authToken xxxx-xxxx-xxxx-xxxx-xxxxxxxxx
```
I expect to be able to be able to access the private packages on the registry.

I believe this relates to #3932 as well as a few others.

**Explanation**: `YarnRegistry` extends `NpmRegistry` however `NpmRegistry.getScopedOption` does not try to read from `YarnRegistry.config` so it is unable to find `//registry.npmjs.org/:_authToken` resulting in errors.

### DIFF
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -298,8 +298,7 @@ export default class NpmRegistry extends Registry {
     }
 
     for (const scope of [this.getScope(packageIdent), '']) {
-      const registry =
-        this.getScopedOption(scope, 'registry') || this.registries.yarn.getScopedOption(scope, 'registry');
+      const registry = this.getScopedOption(scope, 'registry');
       if (registry) {
         return String(registry);
       }
@@ -347,7 +346,8 @@ export default class NpmRegistry extends Registry {
   }
 
   getScopedOption(scope: string, option: string): mixed {
-    return this.getOption(scope + (scope ? ':' : '') + option);
+    const key = scope + (scope ? ':' : '') + option;
+    return this.getOption(key) || this.registries.yarn.getOption(key);
   }
 
   getRegistryOption(registry: string, option: string): mixed {
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -346,24 +346,7 @@ export default class NpmRegistry extends Registry {
   }
 
   getScopedOption(scope: string, option: string): mixed {
-    const key = scope + (scope ? ':' : '') + option;
-
-    let val = this.getOption(key);
-
-    if (typeof val === 'undefined') {
-      const registries = Object.keys(this.registries);
-      for (const registryName of registries) {
-        if (this.registries[registryName].getOption) {
-          val = this.registries[registryName].getOption(key);
-        }
-
-        if (typeof val !== 'undefined') {
-          break;
-        }
-      }
-    }
-
-    return val;
+    return this.getOption(scope + (scope ? ':' : '') + option);
   }
 
   getRegistryOption(registry: string, option: string): mixed {
--- a/src/registries/npm-registry.js
+++ b/src/registries/npm-registry.js
@@ -347,7 +347,23 @@ export default class NpmRegistry extends Registry {
 
   getScopedOption(scope: string, option: string): mixed {
     const key = scope + (scope ? ':' : '') + option;
-    return this.getOption(key) || this.registries.yarn.getOption(key);
+
+    let val = this.getOption(key);
+
+    if (typeof val === 'undefined') {
+      const registries = Object.keys(this.registries);
+      for (const registryName of registries) {
+        if (this.registries[registryName].getOption) {
+          val = this.registries[registryName].getOption(key);
+        }
+
+        if (typeof val !== 'undefined') {
+          break;
+        }
+      }
+    }
+
+    return val;
   }
 
   getRegistryOption(registry: string, option: string): mixed {
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -658,7 +658,6 @@ describe('request', () => {
     };
     const registry = createRegistry({});
     registry.registries.yarn = createRegistry(testCase.config, 'yarn');
-    registry.registries.yarn.registries.npm = registry;
     testCase.requests.forEach(req => {
       const desc =
         `with request url ${req.url}${req.pkg ? ` in context of package ${req.pkg}` : ''} ` +
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -2,6 +2,8 @@
 
 import {resolve, join as pathJoin} from 'path';
 
+import type {RegistryNames} from '../../src/registries/index.js';
+import {registries} from '../../src/registries/index.js';
 import NpmRegistry from '../../src/registries/npm-registry.js';
 import {BufferReporter} from '../../src/reporters/index.js';
 import homeDir, {home} from '../../src/util/user-home-dir.js';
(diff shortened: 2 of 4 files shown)


src/registries/npm-registry.js
```diff
@@ -298,8 +298,7 @@ export default class NpmRegistry extends Registry {
     }
 
     for (const scope of [this.getScope(packageIdent), '']) {
-      const registry =
-        this.getScopedOption(scope, 'registry') || this.registries.yarn.getScopedOption(scope, 'registry');
+      const registry = this.getScopedOption(scope, 'registry');
```
otaku: `YarnRegistry.getsScopedOption` does not exist.

src/registries/npm-registry.js
```diff
@@ -298,8 +298,7 @@ export default class NpmRegistry extends Registry {
     }
 
     for (const scope of [this.getScope(packageIdent), '']) {
-      const registry =
-        this.getScopedOption(scope, 'registry') || this.registries.yarn.getScopedOption(scope, 'registry');
```
otaku: `YarnRegistry.getsScopedOption` does not exist.

BYK:
@arcanis @rally25rs @imsnif  can you take a look?

@otaku thanks a lot for the fix! Do you think you can add tests for the fix?

src/registries/npm-registry.js
```diff
@@ -347,7 +346,24 @@ export default class NpmRegistry extends Registry {
...
+          val = this.registries[registryName].getOption(key);
+        }
+
+        if (typeof val !== 'undefined') {
+          break;
+        }
+      }
+    }
+
+    return val;
```
imsnif: Do you think it will be possible to swap this for an implementation of `YarnRegistry.getScopedOption`?

otaku: I believe that would turn into endless recursion.

`NpmRegistry.getScopedOption` would loop through all registries looking for `getScopedOption` and call it, however if it doesn't exist in that registry, it in turn would do the same thing.

imsnif: Sorry, I'll clarify my meaning. How aboutif we implement `YarnRegistry.getScopedOption` as something like:
```javascript
  getScopedOption(scope: string, option: string): mixed {
    return this.getOption(scope + (scope ? ':' : '') + option);
}
```

otaku: Technically, it already implements `getScopedOption` as it is an extension of `NpmRegistry`.  I forgot about this fact; I have to undo one of my changes.

I could look into moving all the logic into `BaseRegistry` and remove them from both `YarnRegistry` and `NpmRegistry` if that's the desired goal?

imsnif: I think we can start by adding a test for this and then playing around to see what works best. It'll be much easier that way. Makes sense?

otaku:
Alternatively, instead of `YarnRegistry` checking `NpmRegistry` and `NpmRegistry` checking `YarnRegistry`.

I've modified `BaseRegistry.getOption` to include an optional argument boolean `searchRegistries` with a default value of true which will search other registries for the same key.  The `searchRegistries` argument so other registries may call `super.getOption`.  I've also moved the `npmMap` and turned into `aliases` in `BaseRegistry`.

This is a slightly cleaner implementation as it moves everything into `BaseRegistry`.

Same tests pass for this: https://github.com/otaku/yarn/compare/master...otaku:registry-fix2

Let me know if I should merge these changes into this PR.

imsnif:
Hey @otaku - first: thanks for all the rapid work on this!

I reverted your `src` changes locally and the new test-case you added still passes. A brief look tells me maybe it needs to be adjusted a little? Feel like taking a look?
Once we have a test that consistently fails without the changes and passes with them, it'll be significantly easier to decide where these changes should live (base, `YarnRegistry`, `NpmRegistry`, etc.)

As for tests - there indeed don't seem to be unit tests for `YarnRegistry`. While it is tested in the integration tests (and also in some other tests), I agree that it's Not Good(tm). iirc we plan on merging the two paths soon, so that will solve this particular issue.

arcanis:
Just want to mention that the `BaseRegistry` / `YarnRegistry` / `NpmRegistry` classes are a nonsense that we want to eventually merge into a single implementation. Just make sure that any change you make there won't make this more difficult 🙂

otaku:
Fixed the test case.  I had it running against `YarnRegistry` instead of `NpmRegistry`.  Given that you're moving to a single implementation, I went ahead and merged my other changes in.

imsnif:
@otaku - let me know if you want a hand fixing those tests.

otaku:
All good to go now!  The recursive getOption broke everything.

imsnif:
Hey @otaku - thanks again for your work on this.

The issue I find with this implementation (and please correct me if I missed something) is that it causes `.yarnrc` and `.npmrc` to be merged in a rather confusing way. Essentially, if I have a config key in one of them, it will be used.
Sometimes (like in the case of the registry here) `.npmrc` will take precedence, and at other times `.yarnrc` will take precedence (I think in most if not all other cases).
It's confusing to us maintainers and thus it will certainly be confusing for users.
I tried to find an easy way to fix this (starting out in my initial comments about implementing `getScopedOption` in the `YarnRegistry`), but I think it does not make sense the way things are in the `registries` right now.

I feel it would be much better, maintainability-wise and usability-wise to wait for the bigger refactor @arcanis talked about above. Where we can decide how the two config files are merged exactly and write it out explicitly. Barring that, I think a temporary solution that would simplify rather than add complexity is addressing this issue: https://github.com/yarnpkg/yarn/issues/3683
Doing that would also bring us a step closer to merging these three classes. If you'd want to go in that direction and implement it yourself, I'd be happy to review and merge that in a separate PR.  What do you think? Makes sense?

otaku:
@imsnif You're not wrong.  It's actually confusing even trying to follow the code.  Currently, you have `resolvers` which figures out which `registry` to read from.  Given the `registry` that it choice, the following occurs:

#### Currently in `master`:
* **NPMRegistry**
  * `.npmrc`
* **YarnRegistry**:
  * `.yarnrc`
  * aliases from `.npmrc`
  * `.npmrc`
  *  defaults (an object)
#### This PR:
* **NPMRegistry**
  * `.npmrc`
  * `.yarnrc`
  * Aliases, following above order
* **YarnRegistry**:
  * `.yarnrc`
  * `.npmrc`
  * Aliases, following above order
  *  defaults (an object)

This is the temporary solution that  I came up with for now, otherwise I could change this PR into a larger refactor of one `registry` with multiple configuration sources.

If you'd like a PR for the larger refactor I need a few answers:
* How does yarnpkg.com registry work?
* When should yarnpkg.com receive registry.npmjs.org tokens?
* Why not use config settings to set a registry for certain packages so that it will directly use registry.npmjs.org and not have to pass tokens to yarnpkg.com? (assuming yarnpkg.com is a proxy)
* What's the actual priority of config files?
* How should the configuration be merged together?

imsnif:
Hey, @otaku - do you feel like picking this up?

otaku:
I currently do not have time due to work @imsnif

imsnif:
Thanks for your work and effort up until now @otaku!

tregusti:
Was this added/fixed in some other PR? I would like to use it =)

jonaskello:
This would be useful to me too, was it ever merged?

codepunkt:
Same here. I find the combination of `NPMRegistry` and `YarnRegistry` and their handling of `.npmrc`/`.yarnrc` a major source of confusion. Either document the exact behavior, including a description of the methods using `.npmrc` information and those using `.yarnrc` information, or do something along the lines of this pull request.

It's literally impossible to configure yarn to talk to an internal private registry (including authentication and over SSL) simply by reading the docs. I have no clue what information goes where, all i know is that information configured in `.yarnrc` is not read in cases where it should be in order to work.

I cannot possibly know which of these configuration tidbits (`registry`, `_auth`, `strict-ssl`, `always-auth` and more) has to be in which configuration file. Factor in a mix of local (project) and global (user dir) configuration and i'm basically screwed.

Help.

arcanis:
The system is completely overhauled in [the v2](https://github.com/yarnpkg/berry/). Yarn will only read `.yarnrc` files, and the configuration settings can be queried via `yarn config -v` for an easier discoverability. Contributions welcome to move faster towards a release!

PR #5825, opened 2018-05-15
fix(config): allow .npmrc config to supersede yarn defaults

thoov:
**Summary**

Running `yarn config get registry` within a project that contains a .npmrc which overrides the registry config key is not respected and instead the default yarn registry option is returned. Conceptually if you look at the code for `getOption` it first checks to see if the config exists within yarn and if not found it will check for that config within npm's config. However, the problem arises as the config property that is built up gets merged with yarns default values. This means that any config key that exists as a yarn default cannot be overwritten via .npmrc. This change differentiates yarn default configs and configs specified via .yarnrc. This allows the proper config fall back.

**Test plan**

* mkdir /tmp/foobar
* cd /tmp/foobar
* echo 'registry=http://example.com' > .npmrc
* yarn config get registry

Returns `https://registry.yarnpkg.com/`

* yarn-local config get registry

Returns `http://example.com/`

### DIFF
--- a/__tests__/registries/yarn-registry.js
+++ b/__tests__/registries/yarn-registry.js
@@ -1,0 +1,100 @@
+/* @flow */
+
+import YarnRegistry from '../../src/registries/yarn-registry.js';
+import {BufferReporter} from '../../src/reporters/index.js';
+import {YARN_REGISTRY} from '../../src/constants.js';
+
+function createMocks(): Object {
+  const mockRequestManager = {
+    request: jest.fn(),
+  };
+
+  const mockRegistries = {
+    npm: {
+      getOption(key): mixed {
+        return this.config[key];
+      },
+    },
+    yarn: {},
+  };
+
+  const reporter = new BufferReporter({verbose: true});
+
+  return {
+    mockRequestManager,
+    mockRegistries,
+    reporter,
+  };
+}
+
+describe('yarn registry getOption', () => {
+  test('that getOption falls back to npm if the key is defined as a npm config and also is a yarn default', () => {
+    const {mockRequestManager, mockRegistries, reporter} = createMocks();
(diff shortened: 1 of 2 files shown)


thoov:
cc @stefanpenner

BYK:
#5812 also seems related.

imsnif:
Hey @thoov - thanks very much for this PR!
While it looks great, as @BYK mentioned, we've actually recently had a very similar patch that made us understand we'd like the fix to come in a different place. For details: https://github.com/yarnpkg/yarn/pull/5812

I think right now @otaku is working on it, but if not and you wish to go for it yourself - I'd be happy to merge that future PR (and/or provide clarifications beforehand as needed). How does that sound?

stefanpenner:
@imsnif specifically this comment https://github.com/yarnpkg/yarn/pull/5812#issuecomment-389636258 ?

imsnif:
@stefanpenner - yep.

thoov:
@imsnif Sorry I stepped away from this for a few days. I would gladly help out on implementing what you have described in https://github.com/yarnpkg/yarn/pull/5812#issuecomment-389636258. Is someone working on that?

imsnif:
No worries, @thoov - to the best of my knowledge no-one is working on this specifically. However, there is a general issue about merging the two registries here: https://github.com/yarnpkg/yarn/issues/5873

@arcanis - would it help out if we do as described here: https://github.com/yarnpkg/yarn/pull/5812#issuecomment-389636258 ?

arcanis:
Yeah, I think it would be a good thing. I wonder if we should whitelist the options and only keep the ones that we are sure are safe, since they might have different type or interpretation depending on the package manager.

imsnif:
Hey @thoov - what do you think? Still feel like working on this?

PR #5826, opened 2018-05-15
Fix #5816: Trailing slash breaks workspace features when yarn run inside workspace folder

### LINKED ISSUE #5816: Trailing slash breaks workspace features when yarn run inside workspace folder

Bug.

When the `workspaces` definition contains a trailing slash, running e. g. `yarn workspaces info` in the root works just as expected. But when you `cd` into the workspace folder and run `yarn workspaces info`, it fails because it cannot detect that that folder is a workspace.

1. Create a new project directory.
2. Create a subdirectory called `package`.
3. `yarn init` inside `package/`.
4. `yarn init` in the root.
5. Add `package/` (with trailing slash) as a workspace to the root `package.json`.
    ```
    ...
    "workspaces": {
      "packages": [
        "package/"
      ],
    },
    ...
    ```
6. Run `yarn workspaces info` in the root. => Outputs workspace list correctly.
7. Run `yarn workspaces info` in `package/`. => Fails with
    ```
    yarn workspaces v1.6.0
    error Cannot find the root of your workspace - are you sure you're currently in a workspace?
    info Visit https://yarnpkg.com/en/docs/cli/workspaces for documentation about this command.
    ```

Running `yarn workspaces info` should output the workspaces when run both inside root and `package/`.

If this behavior is intended, the documentation could reflect the reason why.

Node v8.11.1
Yarn v1.6.0
Ubuntu 16.04 LTS

Y0hy0h:
**Summary**
Attempt to fix #5816.

**Test plan**
I've added the use case from #5816 as a test case.

## Feedback needed
I'm very unsure about this solution. The test case is too general, it does not directly test the root problem, just one of its symptoms. Also, the replacement of the trailing slash does not seem very robust.

Can I improve this?

### DIFF
--- a/__tests__/commands/workspaces.js
+++ b/__tests__/commands/workspaces.js
@@ -11,10 +11,10 @@ const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'workspace');
 async function runWorkspaces(
   flags: Object,
   args: Array<string>,
-  name: string,
+  cwdSubfolders: string[],
   checkSteps?: ?(config: Config, reporter: BufferReporter) => ?Promise<void>,
 ): Promise<void> {
-  const cwd = path.join(fixturesLoc, name);
+  const cwd = path.join(fixturesLoc, ...cwdSubfolders);
   const reporter = new reporters.BufferReporter({stdout: null, stdin: null, isSilent: true});
 
   const config = await Config.create({cwd}, reporter);
@@ -27,7 +27,7 @@ async function runWorkspaces(
 }
 
 test('workspaces info should list the workspaces', (): Promise<void> => {
-  return runWorkspaces({}, ['info'], 'run-basic', (config, reporter) => {
+  return runWorkspaces({}, ['info'], ['run-basic'], (config, reporter) => {
     expect(reporter.getBufferJson()).toEqual({
       'workspace-1': {
         location: 'packages/workspace-child-1',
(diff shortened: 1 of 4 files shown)


src/config.js
```diff
@@ -633,7 +633,15 @@ export default class Config {
       const ws = extractWorkspaces(manifest);
       if (ws && ws.packages) {
         const relativePath = path.relative(current, initial);
-        if (relativePath === '' || micromatch([relativePath], ws.packages).length > 0) {
+        // Make packages trailing-slash-agnostic
+        const sanitizedPackages = ws.packages.map(packageGlob => {
+          if (packageGlob.endsWith('/')) {
+            packageGlob = packageGlob.slice(0, -1);
+          }
+          return packageGlob + '{,/}';
```
arcanis: Why not just replace this function by `return packageGlob.replace(/\/?$/, '')`? It would ensure that no pattern would be ended with a trailing slash, which seems what we want?

__tests__/commands/workspaces.js
```diff
@@ -42,3 +42,15 @@ test('workspaces info should list the workspaces', (): Promise<void> => {
     });
   });
 });
+
+test('workspaces info should list the workspaces inside workspace folder', (): Promise<void> => {
```
arcanis: I don't understand the test name

Y0hy0h:
Thanks for your feedback! Especially the shorter replacement statement is nice. :)

Issue #5800, opened 2018-05-09
Hi i'm color Color blindness


Loyide:
i'm color blindess so when use yarn outdated 

I can't distinguish the difference between green and yellow

```
 yarn outdated
yarn outdated v1.6.0
info Color legend :
 "<red>"    : Major Update backward-incompatible updates
 "<yellow>" : Minor Update backward-compatible features
 "<green>"  : Patch Update backward-compatible bug fixes
```

rally25rs:
Hi @Loyide 👋 Pardon my lack of knowledge with this particular accessibility concern, but is there a better color combination that you can recommend? We use [chalk.js](https://github.com/chalk/chalk) so the other color options can be viewed there.

I also believe the "yellow" and "green"  come from the color settings of your terminal, so changing color settings there might be another more universal approach that would cover other applications that use colored output.

In this particular case the colors are just a secondary indicator of which digit in the semver (major, minor, patch) is different, and that information is also communicated by displaying the actual semver ranges themselves, so I hope this hasn't prevented you from using `upgrade-interactive`.

I'll mark this as a feature request to improve accessibility. Thanks for your feedback!

Loyide:
For people like me, red, yellow and blue are much better combinations than
 red, yellow, and green combinations.

So I would like to recommend the blue family!

For your information
There are many kinds of color blindness. 
Some of them can not distinguish between green and yellow.
So the colors of the traffic lights can be confusing.

Thank you for your feedback!
Have a great day!

sth:
I feel green intuitively conveys "should cause no problems when upgrading", much more so than blue. So isn't green in general the better color for this?

How the colors actually look is very much dependent on the terminal you use (light or dark background, color scheme). Usually that can be configured by the user (though I don't know if it is possible on windows). Also there are different kinds of color blindness where different pairs of color are hard to distinguish.

Wouldn't it be a much better solution for the user to configure their terminal with a color scheme that is accessible by them? That would not only fix their problems for this yarn command, but for all terminal programs.

BYK:
>Wouldn't it be a much better solution for the user to configure their terminal with a color scheme that is accessible by them? That would not only fix their problems for this yarn command, but for all terminal programs.

[...]

PR #5829, opened 2018-05-16
feat(cli): Change color on console for better accessibility

ohbarye:
Hello maintainers. I addressed https://github.com/yarnpkg/yarn/issues/5800.

**Summary**

For better accessibility.

According to https://github.com/yarnpkg/yarn/issues/5800#issuecomment-387946810, blue is the one with which we should replace green.

> For people like me, red, yellow and blue are much better combinations than red, yellow, and green combinations.

Closes: https://github.com/yarnpkg/yarn/issues/5800

**Test plan**

- [x] `yarn run lint`
- [x] `yarn run test`
- [x] Build with `yarn run build` and use the built to check the change 👇 

before:

image

image

after:

image

📝 The repository is my own test repository, so please do not take care.

### DIFF
--- a/__tests__/commands/outdated.js
+++ b/__tests__/commands/outdated.js
@@ -68,7 +68,7 @@ test.concurrent('works with no arguments', (): Promise<void> => {
     const json: Object = JSON.parse(out);
 
     expect(json.data.body.length).toBe(1);
-    expect(reporter.format.green).toHaveBeenCalledWith('left-pad');
+    expect(reporter.format.blue).toHaveBeenCalledWith('left-pad');
   });
 });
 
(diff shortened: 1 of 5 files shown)


sth:
I feel green intuitively conveys "should cause no problems when upgrading", much more so than blue. So isn't green in general the better color for this?

How the colors actually look is very much dependent on the terminal you use (light or dark background, color scheme). Usually that can be configured by the user (though I don't know if it is possible on windows). Also there are different kinds of color blindness where different pairs of color are hard to distinguish.

Wouldn't it be a much better solution for the user to configure their terminal with a color scheme that is accessible by them? That would not only fix their problems for this yarn command, but for all terminal programs.

PR #5830, opened 2018-05-16
Remove unused constants

ohbarye:
**Summary**

I just found that those constants are not used anywhere, then just removed them.

- `SELF_UPDATE_TARBALL_URL` : it may be remain when removing `self-update`
- `SELF_UPDATE_DOWNLOAD_FOLDER` : it may be remain when removing `self-update`
- `ACCESS_FILENAME` : this seems to have never been used

Please feel free to close this if you won't need this kind of nitpicking pull request. :P

**Test plan**

Let me put search results.

```shell
~/.g/g/o/yarn ❯❯❯ ag (SELF_UPDATE_TARBALL_URL|SELF_UPDATE_DOWNLOAD_FOLDER|ACCESS_FILENAME) .                                                        ⏎ remove-unused-constants ✱
src/constants.js
26:export const SELF_UPDATE_TARBALL_URL = 'https://yarnpkg.com/latest.tar.gz';
27:export const SELF_UPDATE_DOWNLOAD_FOLDER = 'updates';
89:export const ACCESS_FILENAME = '.yarn-access';
~/.g/g/o/yarn ❯❯❯
```

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -23,8 +23,6 @@ export const YARN_INSTALLER_SH = 'https://yarnpkg.com/install.sh';
 export const YARN_INSTALLER_MSI = 'https://yarnpkg.com/latest.msi';
 
 export const SELF_UPDATE_VERSION_URL = 'https://yarnpkg.com/latest-version';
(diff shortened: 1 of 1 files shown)


ohbarye:
Thanks for your review!

PR #5844, opened 2018-05-21
Add badge to display install size

styfle:
**Summary**
This adds a badge to the `README.md` file to display the install size

This is a quick way to see that [yarn](https://packagephobia.now.sh/result?p=yarn) is much smaller than it's competition such as [npm](https://packagephobia.now.sh/result?p=npm).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 <p align="center">
   <a href="https://circleci.com/gh/yarnpkg/yarn"><img alt="Circle Status" src="https://circleci.com/gh/yarnpkg/yarn.svg?style=shield&circle-token=5f0a78473b0f440afb218bf2b82323cc6b3cb43f"></a>
   <a href="https://ci.appveyor.com/project/kittens/yarn/branch/master"><img alt="Appveyor Status" src="https://ci.appveyor.com/api/projects/status/0xdv8chwe2kmk463?svg=true"></a>
+  <a href="https://packagephobia.now.sh/result?p=yarn"><img alt="Install Size" src="https://packagephobia.now.sh/badge?p=yarn"></a>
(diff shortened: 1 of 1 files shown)


rally25rs:
I'm assuming this is fetching the size if installing via npm? Doesn't seem relevant to all installs, and also I suspect not many people check our readme file when looking to install yarn instead of yarnpkg.com

It's a simple addition, so I'm not really opposed to it I guess, I'm just not sure how much value it provides.

Daniel15:
I'm always suspicious of `.now.sh` links as they're not likely to stick around forever.

If you do want to add installation size, I'd prefer something that gets the size of the GitHub release artifact, rather than the npm size, as they may vary slightly. It seems like there's a very slight discrepancy - The badge in this pull request says `4.14 MB`, but the latest release (https://github.com/yarnpkg/yarn/releases/tag/v1.6.0) says 3.97 MB for the JS file, and 920 KB for the .tar.gz file.

styfle:
Thanks for the responses 🙏 

> I'm assuming this is fetching the size if installing via npm? 

Correct, Package Phobia is just running `npm install --save yarn` and reporting at the size on disk.

> I'm always suspicious of .now.sh links as they're not likely to stick around forever

I could purchase a domain name if that builds trust with users. I ran a [twitter poll](https://mobile.twitter.com/styfle/status/988406249334169600) a month ago and it seems that most people agreed **.com** is more trustworthy, I just didn't want to break existing links by changing the domain name.

> latest release (v1.6.0 (release)) says 3.97 MB for the JS file, and 920 KB for the .tar.gz file

Oh good point 🤔

I forgot that the artifacts are already on the GitHub Releases page (which makes sense since you don't want yarn to depend on npm) :octocat: 

I think it's safe to say that this PR is no longer needed 😅

PR #5845, opened 2018-05-21
chore: Better error handling and tests in TarballFetcher

sth:
**Summary**

Currently `TarballFetcher.fetch()` doesn't notice when a truncated tar is extracted, the extracted directory will simply be incomplete. This pull request updates tar-stream to the newest version to detect these errors and show a useful message to the user.

**Test plan**

Tests are added that show that the current code didn't work correctly, and that the updated code does.


sth:
The buildsize increased due to the updated dependencies of `tar-stream`.

BYK:
Looks really nice, thanks! Pinged @rally25rs and @imsnif for an more detailed review but overall it looks very clean and complete with tests.

imsnif:
Hey @sth - is this ready for another review now?

sth:
@imsnif @byk Yes, I think it's ready.

sth:
@BYK The requested changes have been addressed, could you have another look?

BYK:
@sth - oh sorry, I didn't realize I was blocking.

Looks good to me overall but seems like we ignore errors from the gzip decompression. Am I right?

sth:
@BYK: Yes, those errors are currently ignored. I have plans to fix that next, but decided to take it one step at a time (there are also quite a few other places that could use test cases for errors and might not work correctly when something goes wrong).

sth:
To elaborate a bit of my last comment: Decompression errors are ignored, but that's not something new in this pull request. The old code also ignores them. The error handler was in a different place but it didn't catch decompression errors there either, since since error events don't bubble through pipes.

PR #5858, opened 2018-05-23
Support for semver hash in hosted git dependency

lenkan:
**Summary**

npm supports referencing hosted github dependencies in the format `github:user/repo#semver:^1.6.0`. This is currently not supported by yarn as it resolves the hash into `#semver/^1.6.0`. Note the added `/` instead of the `:`.

Thus, yarn outputs the error: "Couldn't find match for "semver/^1.6.0" in ...."

**Motivation**
Given an empty yarn package:

```
// package.json 
{
  "name": "tjosan",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "yarn": "github:yarnpkg/yarn#semver:^1.6.0"
  }
}
```

When running yarn:

```
yarn
```

It errors with:

```
error Couldn't find match for "semver/^1.6.0" in ....
```

Whereas `npm` resolved the dependency without errors.

**Test plan**

Given an empty yarn package:

```
// package.json 
{
  "name": "tjosan",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "yarn": "github:yarnpkg/yarn#semver:^1.6.0"
  }
}
```

When running yarn:

```
yarn
```

It should resolve the dependencies:

```
yarn install v1.6.0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Saved lockfile.
Done in x.xxs.
```

### DIFF
--- a/__tests__/commands/install/bin-links.js
+++ b/__tests__/commands/install/bin-links.js
@@ -166,6 +166,15 @@ test('Only top level (after hoisting) bin links should be linked', (): Promise<v
   });
 });
 
+// fixes https://github.com/yarnpkg/yarn/issues/5876
+test('can use link protocol to install a package that would not be found via node module resolution', (): Promise<
+  void,
+> => {
+  return runInstall({binLinks: true}, {source: 'install-link-siblings', cwd: '/bar'}, async config => {
+    expect(await linkAt(config, 'node_modules', '.bin', 'standard')).toEqual('../standard/bin/cmd.js');
+  });
+});
+
 describe('with nohoist', () => {
   // address https://github.com/yarnpkg/yarn/issues/5487
   test('nohoist bin should be linked to its own local module', (): Promise<void> => {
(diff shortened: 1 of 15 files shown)


lenkan:
It seems like older node doesn't support the negative lookahead regex, I will look into it if this is something you want.

BYK:
Thanks a lot for the patch! I'll let @rally25rs and @imsnif look into this since they have worked on similar pieces recently.

>It seems like older node doesn't support the negative lookahead regex, I will look into it if this is something you want.

Tests seem to be failing even on Node 10 so we need to get those fixed before we can merge or even review this.

kruncher:
I too ran into this issue today when attempting to use yarn in a project.

lenkan:
I tried to figure out an easy way to do it without the negative lookbehind regex. But I couldn't understand the existing method at all with the different splits on strings. From the tests, all I could gather was that we should find the `user`, `repo` and `hash` from a git reference string. So I thought, isn't that just to:

1. Strip hash
2. Strip anything leading up to the user name.
3. Split the remaining parts on `/` where the last bit is the repo, and the second last is the user?

All the initial tests passed, I added a bunch of new test cases. Did I miss any test cases? I can't see what some of the previous splits were achieving.

PR #5859, opened 2018-05-23
Support MSYS{2} bash in bin/yarn

akrieger:
**Summary**
Add a case in `bin/yarn` to run `cygpath` on the yarn base dir if the environment is MSYS.

Automatic path translation happens differently, or perhaps conditionally, depending on whether msys detects the target binary is a native Win32 application or not. While bash from Git for Windows seems to treat `winpty` as a Win32 binary, the version of `winpty` which is compatible with MSYS2 does not trigger the path conversion.

**Test plan**
`yarn` executed from within bash within MSYS2 correct executes, instead of failing with a path
related error. `yarn` continues to execute correctly from within Git Bash.

```
// Changes were made to system install to test more easily.
$ which yarn
/c/Program Files (x86)/Yarn/bin/yarn
akrieger@AKRIEGER-701 /c/Users/akrieger
$ yarn
yarn install v1.6.0
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
```

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -5,6 +5,7 @@ basedir=$(dirname "$(readlink "$0" || echo "$argv0")")
 case "$(uname -s)" in
   Linux) basedir=$(dirname "$(readlink -f "$0" || echo "$argv0")");;
   *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
+  *MSYS*) basedir=`cygpath -w "$basedir"`;;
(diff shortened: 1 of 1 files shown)


BYK:
Looks and sounds good to me but calling our expert, @Daniel15, on this before merging.

PR #5879, opened 2018-05-25
fix bin links for link: protocol (fixes #5876)

### LINKED ISSUE #5876: TypeError: Reduce of empty array with no initial value
**What is the current behavior?** Installing dependencies fails with `error An unexpected error occurred: "Reduce of empty array with no initial value".`

This is the simplest reproduction I've been able to create. The result should look something like this: https://github.com/spalger/reproduce-issues/tree/master/yarn-reduce-of-empty-array-without-initial-value

error log: https://github.com/spalger/reproduce-issues/blob/master/yarn-reduce-of-empty-array-without-initial-value/bar/yarn-error.log

```sh
# create two packages
mkdir foo bar

# install `tree-kill` dependency in `foo`
cd foo && yarn init -y && yarn add tree-kill

# link to `foo` from bar
cd ../bar && yarn init -y && yarn add link:../foo
```

`yarn add` in `bar` should fail.

**What is the expected behavior?** `yarn add` should not fail with internal error

`node -v`: v8.11.0
`yarn -v`: 1.7.0

bdwain:
**Summary**
Fix #5876 . There was an error in the new bin link calculation logic when the bin folder was a symlink. The code would only look for a target installation of the package to be linked relative to the realpath of the bin folder, when it should also consider targets relative to the symlinked location of the bin folder and choose the closest overall location. Otherwise, it may fail to link.

**Test plan**

Added a new unit test which fails without this change. Based on [this reproduction](https://github.com/spalger/reproduce-issues/tree/master/yarn-reduce-of-empty-array-without-initial-value)

### DIFF
--- a/__tests__/commands/install/bin-links.js
+++ b/__tests__/commands/install/bin-links.js
@@ -166,6 +166,15 @@ test('Only top level (after hoisting) bin links should be linked', (): Promise<v
   });
 });
 
+// fixes https://github.com/yarnpkg/yarn/issues/5876
+test('can use link protocol to install a package that would not be found via node module resolution', (): Promise<
+  void,
+> => {
+  return runInstall({binLinks: true}, {source: 'install-link-siblings', cwd: '/bar'}, async config => {
+    expect(await linkAt(config, 'node_modules', '.bin', 'standard')).toEqual('../standard/bin/cmd.js');
+  });
+});
+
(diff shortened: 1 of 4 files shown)


bdwain:
fixed lint errors

bestander:
Thanks, @bdwain!

Issue #5863, opened 2018-05-24
changelog for 1.7.0?


lingxuan630:
I have got a  update version (1.7.0) for yarn this morning. But I found nothing change except `yarn.lock`(https://github.com/yarnpkg/yarn/compare/v1.7.0...master). What happen? And I can not find any changelog for 1.7.0.

bdwain:
don't you want to compare 1.7.0 to 1.6.0? Also, the [yarn twitter](https://twitter.com/yarnpkg/status/999404494906712066) said there would be a changelog tomorrow.

emilyaviva:
Any word on that changelog? "Release a stable version with no changelog" seems like a sub-optimal way to run a major open-source project.

dobesv:
https://github.com/yarnpkg/yarn/compare/v1.6.0...v1.7.0

emilyaviva:
@dobesv That's not a changelog.

dobesv:
Yeah, I should have included a comment along the lines of "The changelog wasn't posted yet but you can get a sense of it from the commit log."

I believe they said on Twitter they will release a ChangeLog in the next day or two.

ecbrodie:
@dobesv Excuse me if I'm asking this with a minimal amount of context, but I'm curious as to why Yarn does not continuously maintain a CHANGELOG in its source repo, as per https://keepachangelog.com/en/1.0.0/

LitoMore:
Why not release with release note?

rodrigobdz:
@ecbrodie I'm also asking myself, why would they release a new version without having a changelog first.

BYK:
Thanks for all the comments. There's definitely some more automation work that needs to be done regarding releases and changelog and we didn't have enough time to invest there yet.

I'll publish the changelog shortly.

bertrandjamin:
It seems that you are using the `conventional-changelog` git commit standard. 

By this way a changelog can be easily generated via [standard-version](https://github.com/conventional-changelog/standard-version). 

:)

emilyaviva:
So, any word on the Yarn 1.7.0 changelog?

BYK:
Published: https://github.com/yarnpkg/yarn/releases/tag/v1.7.0

PR #5882, opened 2018-05-25
Add new feature

LitoMore:
Related to #5863


benjamingr:
This is borderline abusive.

Some people wrote a package manager you are enjoying for free and they made a mistake. 

It would be great if you would not use the opportunity to make fun at their expense and acted more constructively.

LitoMore:
@benjamingr Yes, that's right. Thank you for your point :)

PR #5884, opened 2018-05-25
fix for --offline\

dylang:
**Summary**

When we read from the offline cache directory, we cache that file listing using `his.config.getCache`, however we had just one static key even though we need to read from different directories for scoped modules.

This means if it looks first in a scoped directory, it won't find the cache of non-scoped modules, or visa versa.

**Test plan**

I didn't see a specific test for this. I'm happy to add one if required, just point me to where.

Many thanks!

### DIFF
--- a/src/resolvers/registries/npm-resolver.js
+++ b/src/resolvers/registries/npm-resolver.js
@@ -110,7 +110,7 @@ export default class NpmResolver extends RegistryResolver {
     invariant(this.config.cacheFolder, 'expected packages root');
     const cacheFolder = path.join(this.config.cacheFolder, scope ? `${NPM_REGISTRY_ID}-${scope}` : '');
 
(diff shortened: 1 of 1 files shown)


arcanis:
We'll need a test for this one. They are typically put into the `__tests__` directory of the repository.

arcanis:
@dylang do you have repro steps for this issue? To see whether we can craft a good test

dylang:
@arcanis Great point, I'll work on getting a repo case.

dylang:
@aracarie I've added repo steps and more details to the ticket description. I also learned that having a lockfile works around this bug, so I'm no longer blocked by this issue, but still might want to fix the issue.

The funny thing is I'm doing all of this to avoid issues we're getting with `--mutex` not always working. Because I thought `--offline` was always broken when using scoped+non-scoped, this seemed like a more serious issue, but because it's only broken when there is no lockfile it doesn't feel as serious.

Issue #5864, opened 2018-05-24
install --focus tries to install workspace dependencies from npm


jacobp100:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behavior?**

I have a workspace setup with workspaces `app`, `web` and `core`. `app` and `web` depend on `core`

Running `yarn install` from `app` installs all dependencies, including workspace dependencies

Running `yarn install --focus` from `app` attempts to install `web` and `core` from npm.

**What is the expected behavior?**

It shouldn't be looking to npm at all.

**Please mention your node.js, yarn and operating system version.**

Yarn v1.7.0
Node v8.9.0

bdwain:
Hi @jacobp100. If your packages are not on the npm registry. focus won't really work for you. The point is to pull down the pre-built version from a remote location. If they aren't stored in a remote location, focus can't really do anything different.

Out of curiosity, what behavior were you expecting from a focused install versus an unfocused one? Maybe the docs can be improved to make it more clear what focus can and can't do.

jacobp100:
That makes sense! React native doesn't work too well with the symlinks, so I have a script to basically install the `core` module as if it was downloaded from npm (no symlinks etc). The script is pretty gnarly, so I was hoping I could get rid of it with this.

bdwain:
ah i see. Have you seen [nohoist](https://yarnpkg.com/blog/2018/02/15/nohoist/)? It was made specifically with react-native in mind and solves a lot of their symlink issues.

jacobp100:
Yea that worked for all but the sibling workspace dependency. Thanks anyway! I’ll close because it looked like I just misunderstood this feature!

jonaskello:
I got a similar problem where I created a new package and ran `yarn install --foucs`. It tried to install the package I was working on from npm. Since the package was brand new it had not yet been published to npm. I think it make sense that the dependencies has to be on the registry, but should the package your working on have to be published too?

bdwain:
Hi @jonaskello. Can you provide a a failing example? It should not request the package you are focusing on from npm. Only other packages in your project.

jonaskello:
Actually, this was in a private monorepo which has scoped packages that publishes to a private on-premise registry (running verdaccio). But I suspect it would be the same if it would publish to to the public npm repository, I can try and see if I can reproduce in that case. Anyway, once I published the package once the error went away :-). Currently I'm instead stuck on the error in #5881.

bdwain:
[...]

PR #5893, opened 2018-05-26
fix #5881: allow folder name of focused workspace to not match package name

### LINKED ISSUE #5881: --install focus gives error "targetHoistManifest missing"
bug

```
$ yarn install --focus
yarn install v1.7.0
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.1.3: The platform "win32" is incompatible with this module.
info "fsevents@1.1.3" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
error An unexpected error occurred: "targetHoistManifest missing".
info If you think this is a bug, please open a bug report with the information provided in "XXX\yarn-error.log".
```

Unfortunately, this is in a private monorepo.

To install dependencies without errors.

Windows 10

```
$ node --version
v8.9.1
$ yarn --version
1.7.0
```

bdwain:
**Summary**
fix #5881. Previously, the focused workspace logic determined the target workspace name by looking at the folder it was contained in. This was a mistake and it should have looked at the package.json name field.

I also moved the focus property to the config object rather than just as a flag for the install command. This made it easier to read the target name once rather than everywhere it was needed. 

**Test plan**

Added a unit test focusing on a scoped package. It also demonstrates that the focused workspace is not requested from npm, like mentioned in #5864.


src/config.js
```diff
@@ -229,6 +234,17 @@ export default class Config {
...
+    // using focus in a workspace root is not allowed
+    if (this.focus && (!this.workspaceRootFolder || this.cwd === this.workspaceRootFolder)) {
+      throw new MessageError(this.reporter.lang('workspacesFocusRootCheck'));
+    }
+
+    if (this.focus) {
+      const focusedWorkspaceManifest = await this.findManifest(this.cwd, false);
+      invariant(focusedWorkspaceManifest, `missing manifest at ${this.cwd}`);
+      this.focusedWorkspaceName = focusedWorkspaceManifest.name;
+    }
```
arcanis: You should use `readRootManifest`

src/package-hoister.js
```diff
@@ -666,9 +666,9 @@ export default class PackageHoister {
   }
 
   markShallowWorkspaceEntries() {
-    const targetWorkspace = path.basename(this.config.cwd);
```
arcanis: Why do you need the `focusedWorkspaceName` instead of just using `readRootManifest` here?

src/config.js
bdwain: that hardcodes isRoot to true for normalizeManifest though. Isn't the root in this case the root of the whole project, not the individual workspace that is being focused on? 

Maybe it would make more sense to say readManfiest and pass false for isRoot? Or does it not make a difference for this?

src/package-hoister.js
bdwain: I suppose we could read the manifest everywhere we need the focused package's name. It seemed simpler and faster to do it once up front because otherwise we'd need to do it like 4-5 times.

bdwain:
@arcanis the test failures seem to be occurring in master as well. I don't think they are related to my PR.

 FAIL  __tests__/commands/info.js
  ✓ without arguments and in directory containing a valid package file (465ms)
  ✓ with first argument "." and in directory containing a valid package file
  ✓ with one argument shows info about the package with specified name
  ✓ with one argument does not contain readme field (1ms)
  ✓ with two arguments and second argument "readme" shows readme string
  ✓ with two arguments and second argument "version" shows `latest` version
  ✓ with two arguments and second argument as a simple field
  ✓ with two arguments and second argument as "."-separated field path
  ✓ with two arguments and second argument as a non-existing field
  ✓ with two arguments and second argument path containing non-existing field (1ms)
  ✕ reports error on invalid package names (218ms)
  ✓ reports error with too many arguments

  ● reports error on invalid package names

    https://registry.yarnpkg.com/YARN.invalid.package.name.YARN: Not found

bdwain:
@arcanis FYI ill be out of town the next 4 days but can address any comments you have when i get back.

bdwain:
fixed the merge conflicts

arcanis:
LGTM, thanks!

bdwain:
thanks!

PR #5898, opened 2018-05-28
Update Debian package to "conflict" with cmdtest

Daniel15:
**Summary**

The `cmdtest` package also installs a `yarn` command, and people often get confused by this (see #2821). If we specify that the Yarn package conflicts with the cmdtest package, installing Yarn will automatically uninstall cmdtest.

**Test plan**

Will test the package built by CircleCI (since I don't have a functional dev environment right now).

### DIFF
--- a/resources/debian/control.in
+++ b/resources/debian/control.in
@@ -1,7 +1,7 @@
 Package: yarn
 Version: $VERSION-1
 Recommends: nodejs
-Conflicts: nodejs (<< 4.0.0)
+Conflicts: nodejs (<< 4.0.0), cmdtest
 Section: devel
 Priority: optional
 Architecture: all


Daniel15:
So it looks like with this change, `dpkg` will throw an error if you try to install `yarn` when `cmdtest` is installed:
```
root@f64c87a6ad83:/tmp# dpkg -i yarn_1.8.0-0_all.deb
Selecting previously unselected package yarn.
dpkg: regarding yarn_1.8.0-0_all.deb containing yarn:
 yarn conflicts with cmdtest
  cmdtest (version 0.22-1) is present and installed.

dpkg: error processing archive yarn_1.8.0-0_all.deb (--install):
 conflicting packages - not installing yarn
Errors were encountered while processing:
 yarn_1.8.0-0_all.deb
```

I'll try to see if there's a way to automatically remove it.

benjaoming:
> I'll try to see if there's a way to automatically remove it.

Removing a package silently from the user's system? I think that's impossible but for good reasons :)

But certain interfaces will suggest it in a cleaner way IIRC. Try `gdebi`.

I think the above `dpki` error message sounds just fine, nothing to worry about.. people will just have to remove it, that's less confusing than to face a `yarn` command that isn't Yarn after having installed it.

Daniel15:
> I think the above dpki error message sounds just fine, nothing to worry about.. people will just have to remove it, that's less confusing than to face a yarn command that isn't Yarn after having installed it.

That makes sense. I think I agree with you.

benjaoming:
The lesser evil :)

benjaoming:
This is a good intermediate solution - but I just found by reading the [Debian ITP for yarnpkg](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843021) that `cmdtest` is somewhat used.

The possibly best option would be to

1. Rename the package to `yarnpkg` to prepare for the official Debian package by the same name.
2. Rename the binary in the Debian package to `yarnpkg` and not have `yarn` at all in the main package.
3. Add a new package `yarnpkg-legacy`, same as `nodejs-legacy` and include the `yarn` binary there.

That way, someone can install `yarnpkg` and invoke `yarnpkg` without all these silly collisions. Most people will want the `yarn` binary/shortcut and it's easy to create manually or through the `yarnpkg-legacy` package.

Please refer to the [Debian ITP for yarnpkg](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843021) before making comments on the proposal.. it is quite detailed.

PR #5915, opened 2018-05-31
fix(info): report empty npm response with `infoFail` again

sth:
**Summary**

#5903 added support for exceptions thrown by the npm request, but changed handling of empty responses. They are now an invariant violation instead of an expected error reported as `infoFail`.

But empty responses [can still happen][1]. I assume they should continue to be handled like they were before #5903. This pull request makes this change.

**Test plan**

Tests should still pass and hopefully more reliably than before.

 [1]: https://circleci.com/gh/yarnpkg/yarn/16689


arcanis:
So now 404 are either empty or an error? 🙁

PR #5925, opened 2018-06-02
Replace Chocolatey nodejs dependency with nodejs-lts.

ihalila:
**Summary**

This PR replaces the Chocolatey package's [nodejs.install](https://chocolatey.org/packages/nodejs.install) dependency with [nodejs-lts](https://chocolatey.org/packages/nodejs-lts). Since Yarn doesn't require the latest version of NodeJS depending on the LTS release should cause less situations where a user's other tools are not compatible with the NodeJS version installed via Chocolatey.

The nodejs.install package's documentation recommends depending on nodejs-lts, and some people have requested this change on the Chocolatey discussion page.

**Test plan**

I verified that the Chocolatey package is buildable and installs, I don't expect this to cause issues since the change is quite trivial.

### DIFF
--- a/resources/win-chocolatey/yarn.nuspec
+++ b/resources/win-chocolatey/yarn.nuspec
@@ -36,7 +36,7 @@ utilization.
 		<bugTrackerUrl>https://github.com/yarnpkg/yarn/issues</bugTrackerUrl>
 		<releaseNotes>https://github.com/yarnpkg/yarn/releases/tag/v$version$</releaseNotes>
 		<dependencies>
-			<dependency id="nodejs.install" version="6.0" />
+			<dependency id="nodejs-lts" version="6.0" />
 		</dependencies>
 	</metadata>
 


ihalila:
The test failures on test-linux-node6 don't seem related to my changes: 
```
Command failed: /root/project/yarn/bin/yarn add react-scripts@1.0.13 --cache-folder /tmp/d-11852-133-1v41lsa.ses0mgqfr/cache --no-node-version-check --ignore-engines
error An unexpected error occurred: "https://registry.yarnpkg.com/es6-weak-map/-/es6-weak-map-2.0.2.tgz: Parse Error". 
```

resources/win-chocolatey/yarn.nuspec
```diff
@@ -36,7 +36,7 @@ utilization.
 		<bugTrackerUrl>https://github.com/yarnpkg/yarn/issues</bugTrackerUrl>
 		<releaseNotes>https://github.com/yarnpkg/yarn/releases/tag/v$version$</releaseNotes>
 		<dependencies>
-			<dependency id="nodejs.install" version="6.0" />
+			<dependency id="nodejs-lts" version="6.0" />
```
BYK: We should also bump this to 8, right?

Daniel15:
Sorry I missed this. Seems fine to me. I wish we could specify that **either** nodejs.install **or** nodejs-lts is fine.

PR #5927, opened 2018-06-02
fix(deps): change cmd-shim to @zkochan/cmd-shim

SevenOutman:
**Summary**
Update devDependency `cmd-shim` with issues to a patched version

As described in [its PR#29](https://github.com/npm/cmd-shim/pull/29), it contains [a piece of code](https://github.com/npm/cmd-shim/blob/master/index.js#L166) that can cause error in strict mode.
```javascript
function chmodShim (to, cb) {
  var then = times(2, cb, cb)
  fs.chmod(to, 0755, then) // SyntaxError: Invalid number
  fs.chmod(to + ".cmd", 0755, then)
}
```
There was actually NO octal number literal syntax before ES2015 adding support with `0o` prefix, and `0` prefix always throw errors in strict mode.
[Refer to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Converting_mistakes_into_errors) for more details.

`cmd-shim` has been [out of maintenance for 2 years](https://github.com/npm/cmd-shim/commits/master), and it's [only been tested out on early Node (<= 0.10)](https://github.com/npm/cmd-shim/blob/master/.travis.yml). So I update `yarn`'s dependency on it to [a fork version](https://github.com/npm/cmd-shim) in which this issue is fixed.

**Test plan**

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,11 +6,11 @@
   "preferGlobal": true,
   "description": "📦🐈 Fast, reliable, and secure dependency management.",
   "dependencies": {
+    "@zkochan/cmd-shim": "^2.2.4",
     "babel-runtime": "^6.26.0",
     "bytes": "^3.0.0",
     "camelcase": "^4.0.0",
     "chalk": "^2.1.0",
-    "cmd-shim": "^2.0.1",
     "commander": "^2.9.0",
     "death": "^1.0.0",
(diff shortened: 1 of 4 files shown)


arcanis:
Thanks!

PR #5934, opened 2018-06-05
fix(sshpk): Upgrade request and sshpk to latest versions

### LINKED ISSUE #5477: Replace use of deprecated/unsafe new Buffer constructor

`yarn` uses deprecated `new Buffer()` constructor and causes deprecation warnings when run with `NODE_PENDING_DEPRECATION=1`.

```
$ ag '\bBuffer\('
src/registries/npm-registry.js
340:        const pw = new Buffer(String(password), 'base64').toString();
341:        return 'Basic ' + new Buffer(String(username) + ':' + pw).toString('base64');

src/util/fs.js
835:const cr = new Buffer('\r', 'utf8')[0];
836:const lf = new Buffer('\n', 'utf8')[0];
```

`yarn` should not use deprecated/unsafe `Buffer` constructor. According to the deprecation warning `new Buffer()` should be replaced with one of `Buffer.alloc()`, `Buffer.allocUnsafe()` or `Buffer.from()`; the [`safe-buffer`](https://yarn.pm/safe-buffer) package is another option.

```
$ node -v
v8.9.4

$ yarn -v
1.5.1

$ uname -a
Linux 4.15.6-300.fc27.x86_64 #1 SMP Mon Feb 26 18:43:03 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

BYK:
**Summary**

Fixes #5477. Upgrades `request` ve `sshpk` modules to latest versions to avoid deprecated `Buffer` usages.

**Test plan**

Install new dependencies, build and run Yarn with Node 10. Make sure there are no `Buffer` warnings.


BYK:
I keep seeing `(node:3088) ExperimentalWarning: The fs.promises API is experimental` errors now but not sure if related to this patch.

simonkberg:
Are you sure this fixes #5477 completely? Since sshpk depends on asn1 which hasn't been patched yet: https://github.com/joyent/node-asn1/pull/30.

BYK:
@simonkberg good point. I haven't seen the warning myself and that maybe because the code path is not being used by yarn (or only in my tests).

I'll keep an eye on that patch too!

jacobq:
Another check you can do is search/grep for `new Buffer` in the stand-alone JS. Looks like it's much improved (thanks for your work!): down to 26 matches, some of which are just comments:
https://nightly.yarnpkg.com/yarn-1.10.0-20180719.1544.js

mkutny:
@jacobq , as for lodash - seems that they use gulp only as a dev dependency and it's quite recent:
https://github.com/yarnpkg/yarn/blob/11eb9bb72d804a5860049ce8d01f0c189659e2f7/package.json#L81

Am missing something?

jacobq:
@mkutny Oops, you're right, but there is still ancient version of lodash (1.x) being brought in somehow. It doesn't look like yarn's gulp is the problem but rather `globule#lodash@1.0.2`. I look will look again later when I have a minute & update my comment.

wopian:
@mkutny `gulp^3.9.0` is not a recent version of Gulp - it was released over 2 years ago (`3.9.1`). Latest is `4.0.0` back in Jan, which does not have the offending package (`globule`) in its chain as `vinyl-fs` no longer depends on `glob-watcher`

jacobq:
BTW, it looks like we may still have a problem. Of the 14 string matches for `new Buffer(` calls in [`yarn-1.10.0-20180817.0955.js`](https://nightly.yarnpkg.com/yarn-1.10.0-20180817.0955.js) the first 9 and last 2 are OK.

In module "471" there are 2 cases where it is not being used as a fall-back for newer methods. I believe this code corresponds to [`chardet@0.4.2`](https://github.com/runk/node-chardet/blob/v0.4.2/index.js#L95-L109) ([fixed in v0.7.0](https://github.com/runk/node-chardet/commit/6fa5c9b2686507af50f2e904a0de6720fc731674)), which is blocked by https://github.com/commitizen/cz-cli/issues/552. However, since `commitizen` is only a dev dependency I don't think it should be showing up in the standalone JS output.  

In module "546", which I believe corresponds to [`node-http-signature`](https://github.com/joyent/node-http-signature), there is another. This is blocked until https://github.com/joyent/node-http-signature/pull/67 gets merged.



```
$ yarn why chardet
yarn why v1.9.4
[1/4] Why do we have the module "chardet"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "chardet@0.4.2"
info Reasons this module exists
   - "external-editor" depends on it
   - Hoisted from "external-editor#chardet"
Done in 1.77s.
```

PR #5941, opened 2018-06-06
fix(autoclean): Fixes autoclean when used with workspaces

arcanis:
**Summary**

The current autoclean implementation doesn't take workspaces into account. It causes an error when running `yarn install` from a different directory, and some files not to be cleaned because only the top-level directory will.

**Test plan**

Added a test.


src/cli/commands/autoclean.js
```diff
@@ -67,7 +68,7 @@ export async function clean(
   removedFiles: number,
   removedSize: number,
 }> {
-  const loc = path.join(config.cwd, CLEAN_FILENAME);
+  const loc = path.join(config.lockfileFolder, CLEAN_FILENAME);
```
BYK: nit: seems like you're fixing another issue here and not adding tests for it ;)

arcanis: Test would be a pain, the command would have to run in a subfolder of the fixutre. The `run` helper doesn't work well with this use case.

PR #5957, opened 2018-06-08
Handle truncated gzip data in TarballFetcher

sth:
**Summary**

Currently `TarballFetcher.fetch()` doesn't handle truncated gzip data correctly. Extraction hangs and no error is reported. This pull request updates `gunzip-maybe` and it's dependencies to fix the issue, and improves the reported error message.

**Test plan**

The pull request adds a new test cases which shows the current problem and verifies the solution.


src/fetchers/tarball-fetcher.js
```diff
@@ -80,6 +80,9 @@ export default class TarballFetcher extends BaseFetcher {
     untarStream.on('error', err => {
       reject(new MessageError(this.config.reporter.lang('errorExtractingTarball', err.message, tarballPath)));
     });
+    extractorStream.on('error', err => {
+      reject(new MessageError(this.config.reporter.lang('errorDecompressingGzip', err.message, tarballPath)));
+    });
```
sth: This error handler is not strictly necessary: the `extractorStream` is returned from the function and there are only two callers. Both attach an error handler on their own which will catch this error event.

But this is not very easy to see and more complicated than necessary. The error handling is much more obviously correct if it is done directly here where the stream is created.

sth:
Closing pull request for now since I want to add some more changes that aren't quite ready yet.

PR #5966, opened 2018-06-10
Allow bundleDependencies to be respected during `pack`

thoov:
**Summary**

Yarn pack does not include bundleDependencies during `yarn pack`. This was due to 3 issues:

1. The normalized property is `bundleDependencies` and not `bundledDependencies`. This was causing the destructuring to always be undefined even if bundledDependencies or bundleDependencies were set in package.json.

2. The way the filtering logic is constructed excludes node_modules from the 'search space'. This leads the filtering to attempt to find dependencies but cant because they were excluded.

3. The logic was only set up to find the dependency but not that dependency's dependency tree. This would lead to an incomplete pack of the needed dependencies.

Fixing problem 1: was a simple property name change which you can see here: https://github.com/yarnpkg/yarn/commit/8081ccd29817a4fb0d02add8e5f34136baa8776e

For fixing problems 2 & 3: We take the specified bundleDependencies and construct the list of their dependency graphs. Once we have the list we figure out all of the needed files for each one of those (sub)dependencies. You can see that change here: https://github.com/yarnpkg/yarn/commit/54830e4ab4d6dab19f4ba9bb9530162fe3d9c6f1

Related Issue: https://github.com/yarnpkg/yarn/issues/2636

**Test plan**

I ran this in comparison to `npm pack` on a few local projects of various sizes and compared the outputs and performance (right now there is also an issue with bundleDependecies and npm so we will need to use npm 5.3.0 as that is the last working version).

The outputs look correct but I do notice a perf hit compared to npm pack. This perf slowness does not seem to be related to these changes directly but on the library that created the tar. On larger projects that I tested it would take npm pack about 15 seconds and for yarn pack it was 45 seconds.


arcanis:
Pack-related tests seem to fail.

Btw, I'm a bit curious, what are you using `bundleDependencies` for?

stefanpenner:
> Btw, I'm a bit curious, what are you using bundleDependencies for?

A deployable tarball.

thoov:
@arcanis Tests should now be passing (except for appveyor but they seem to be failing before my change). As @stefanpenner said bundleDependencies are needed for us to create deployable tarball that is completely self contained.

stefanpenner:
@thoov mind comparing appveyer before/after and confirming its not worse then before? That should help the yarn maintainers a tad.

thoov:
@stefanpenner @arcanis Ok now tests are fully passing

thoov:
@arcanis Could you provide a review or do you know who would be best to take a look at this PR?

byF:
Can anyone review and approve this please?

src/cli/commands/pack.js
```diff
@@ -65,10 +66,17 @@ export async function packTarball(
...
 
-  // include bundledDependencies
-  if (bundledDependencies) {
-    const folder = config.getFolder(pkg);
-    filters = ignoreLinesToRegex(bundledDependencies.map((name): string => `!${folder}/${name}`), '.');
+  // include bundleDependencies
+  let bundleDependenciesFiles = [];
+  if (bundleDependencies) {
+    for (const dependency of bundleDependencies) {
+      const dependencyList = depsFor(dependency, config.cwd);
```
arcanis: What is this `depsFor` used for? Why can't we just walk the `node_modules/${dependency}` tree?

arcanis:
Hey, sorry for the delay, I'd like to know more about the `depsFor` function. It seems to come from a package without much usage, that itself uses dependencies without much usage. Unless there's a good reason to use it, I'd prefer to be careful about the dependencies we add to the project.

src/cli/commands/pack.js
thoov: `depsFor` simply returns a list of all of the dependencies for a given package so in a sense it is "walking the tree". I agree with your concerns about not adding unnecessary extra packages to the project. Do you know if yarn has a built in util for getting this info? If not I think it is safe to bring this package in.

thoov:
Thanks for the comments @arcanis. I left a response inline to your concern above.

src/cli/commands/pack.js
arcanis: What I don't understand is: wouldn't something like this work without `depsFor` (pseudo code)?

```
const FOLDERS_IGNORE_NO_NODE_MODULES = new Set(FOLDERS_IGNORE);
FOLDERS_IGNORE_NO_NODE_MODULES.delete('node_modules');

for (const dependency of bundleDependencies) {
  const depPath = getPathForDependency(dependency);

  const filesForBundledDep = await fs.walk(depPath, null, new Set(FOLDERS_IGNORE_NO_NODE_MODULES));
  bundleDependenciesFiles = bundleDependenciesFiles.concat(filesForBundledDep);
}
```

Since `walk` would then traverse the whole `node_modules` tree and pack them all.

thoov: @arcanis The issue is that this pseudo code doesn't actually walk the dependency graph for a dependency and only walks the directory structure. If we have a dependency `a` which has `b` and `c` as dependencies we get this tree structure:

node_modules/a
node_modules/b
node_modules/c

And `const depPath = getPathForDependency(dependency);` would only walk `node_modules/a` and miss the other ones. Thats why I use `depsFor` as I need more than the "top" level bundled dependencies. Let me know if this make sense or if there is something I am missing.

arcanis: I see, that makes sense. I'm not fond of adding a new dep, but I don't see an easy replacement at the moment, so we can start with that.

arcanis:
Merged. Thanks!

PR #5979, opened 2018-06-12
fix(pack): higher priority to .npmignore over .gitignore on windows

pieh:
**Summary**

This fixes `yarn pack` not ignoring `.gitignore` when `.npmignore` is avaialable on Windows. Initial fix was done in https://github.com/yarnpkg/yarn/pull/3538, but it used string concatenation to join paths which make it unfortunately fail for Windows which uses different path separator than other systems. It was reported as not fixed in https://github.com/yarnpkg/yarn/issues/685#issuecomment-329895660 (OS: Windows) in initial issue after issue was closed as fixed.

**Test plan**

First I tested if tarball created by `yarn pack` is correct (so it ignores `.gitignore` when I have `.npmignore`). 

To demonstrate here, I added extra console output here https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/pack.js#L89 logging `dotIgnoreFiles` and here's before / after:

**Before:**

```
D:\dev\yarn-test>yarn pack
yarn pack v1.7.0
dotIgnoreFiles [ { relative: '.gitignore',
    basename: '.gitignore',
    absolute: 'D:\\dev\\yarn-test\\.gitignore',
    mtime: 1528836250698 },
  { relative: '.npmignore',
    basename: '.npmignore',
    absolute: 'D:\\dev\\yarn-test\\.npmignore',
    mtime: 1528836254147 } ]
success Wrote tarball to "D:\\dev\\yarn-test\\yarn-test-v1.0.0.tgz".
Done in 0.45s.
```

**After:**

```
D:\dev\yarn-test>yarn pack
yarn pack v1.7.0
dotIgnoreFiles [ { relative: '.npmignore',
    basename: '.npmignore',
    absolute: 'D:\\dev\\yarn-test\\.npmignore',
    mtime: 1528836254147 } ]
success Wrote tarball to "D:\\dev\\yarn-test\\yarn-test-v1.0.0.tgz".
Done in 0.06s.
```

`.gitignore` is correctly omitted if `.npmignore` is present

### DIFF
--- a/src/util/filter.js
+++ b/src/util/filter.js
@@ -160,7 +160,7 @@ export function filterOverridenGitignores(files: WalkFiles): WalkFiles {
     } else {
       //don't include .gitignore if .npmignore or .yarnignore are present
       const dir = path.dirname(file.absolute);
-      const higherPriorityIgnoreFilePaths = [`${dir}/${IGNORE_FILENAMES[0]}`, `${dir}/${IGNORE_FILENAMES[1]}`];
+      const higherPriorityIgnoreFilePaths = [path.join(dir, IGNORE_FILENAMES[0]), path.join(dir, IGNORE_FILENAMES[1])];
       const hasHigherPriorityFiles = files.find(file => higherPriorityIgnoreFilePaths.indexOf(file.absolute) > -1);
       if (!hasHigherPriorityFiles) {
         return [...acc, file];


hulkish:
LGTM

arcanis:
Yep, it would be nice to have a new test that would ensure that the .npmignore always has precedence over .gitignore

PR #5970, opened 2018-06-11
fix(resolution): Improve dependency resolution performance

rally25rs:
##DO NOT MERGE

**Summary**

Instead of continuing to re-resolve a package in the dependency tree that has already been resolved,
add a request to the existing references requests array. This speeds up very deep/complex dep trees
as seen in workspaces where the workspace modules all depend on each other. Previously the building
of the tree was leading to out of memory errors or just never finishing.

fixes #5726 #5628

**Test plan**

Existing tests still pass. (watch for regressions in peerDep checks, resolutions, nohoist as these features may be affected by changing the way the deps are resolved. i'm making the assumption that these features have sufficient test coverage already.)

For manual testing, see the .zip file in a comment in #5726 for a project that would never finish or eventually out-of-memory error.

PR #5982, opened 2018-06-13
fix(resolution): Improve dependency resolution performance (comparison of #5970)

hulkish:
This is meant as a WIP pr to compare with #5970


src/config.js
```diff
@@ -279,9 +276,11 @@ export default class Config {
       });
 
       this.registries[key] = registry;
-      this.registryFolders.push(registry.folder);
+      if (this.registryFolders.indexOf(registry.folder) === -1) {
```
hulkish: this fixes an issue i was seeing where value of `registryFolders` was `['node_modules', 'node_modules']` - which I think was causing double lookups.

sibelius:
@arcanis any change to get this or the related pull request reviewed?

PR #6007, opened 2018-06-18
Implements --no-default-rc and --use-yarnrc

arcanis:
**Summary**

This PR implements the `--no-default-rc` and `--use-yarnrc` options.

  - `--no-default-rc` disable the rcfile lookup
  - `--use-yarnrc <path>` loads a specified yarnrc file 

This allow to make more deterministic installs, since the exact rc files used are now known to the caller. I haven't implemented `--use-npmrc` as every settings should be configurable from a yarnrc file anyway.

**Test plan**

Added tests.


arcanis:
AppVeyor is timeouting on the publish command, unrelated.

StephanBijzitter:
Little late to the party, but this pull request has added a feature that I did not yet know I needed. Thanks, `--use-npmrc` will prove to be very useful!

clintwood:
@arcanis, @bestander please see issue #6174 

I dug into this a bit and from _my limited understanding of the code base_ it seems that the new options `enableDefaultRc` and  `extraneousYarnrcFiles` are not being propagated from within [updateCwd(...)](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/global.js#L46-L56).

Manually changing the code to look like this seemed to correct the issue for me (but I have no idea if this is the correct fix):
```js
async function updateCwd(config: Config): Promise<void> {
  await fs.mkdirp(config.globalFolder);

  await config.init({
    cwd: config.globalFolder,
    binLinks: true,
    globalFolder: config.globalFolder,
    cacheFolder: config._cacheRootFolder,
    linkFolder: config.linkFolder,
    // added missing config
    enableDefaultRc: config.enableDefaultRc,
    extraneousYarnrcFiles: config.extraneousYarnrcFiles,
  });
}
```

greysteil:
@clintwood - I've just run into the same issue on Dependabot (where we create our own config, and weren't passing through a value for `enableDefaultRc`). I think a better fix than the one you suggested is to set the default value for `enableDefaultRc` to true, which I've done in #6181.

clintwood:
@greyepoxy - That was the other option I thought of, however, it looks like the missing config settings are set to [default values on initialisation if not set by this PR's command line switches by `commander`](https://github.com/greysteil/yarn/blob/master/src/cli/index.js#L497-L498).  So _not_ passing them through and defaulting them in the `Config` class may cause these command line options to be ignored when actually set!?

BTW just skimming through the tests it seemed this specific issue did not have any test cases.

fenduru:
@arcanis I believe this PR should be reverted until #6174 is resolved, as it introduced a regression that I suspect will affect most enterprise users.

jonaskello:
+1 for `--use-npmrc`

Issue #5723, opened 2018-04-24
Yarn will not upgrade ember-source package from 3.1.0 to 3.1.1


RavenHursT:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
Changing version of `ember-source` from `3.1.0` to `3.1.1` in `package.json` deps and running `yarn install` updates lock file to `3.1.1`, but actual package in `node_modules` remains `3.1.0`

**If the current behavior is a bug, please provide the steps to reproduce.**

* create a new yarn project w/ `yarn init`
* create `package.json` w/ `"ember-source": "3.1.0"` under deps
* run `yarn install`
* `cat node_modules/ember-source/package.json` 
* version remains `3.1.0` (expected)
* manually change `ember-source` to `3.1.1` in `package.json`
* `yarn install`
* Output from `install` (notice, lock file updated, but not packages were installed/upgraded)
```
yarn install v1.6.0
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 0.55s.
➜  derp rg ember-source yarn.lock
119:ember-source@3.1.1:
121:  resolved "https://registry.yarnpkg.com/ember-source/-/ember-source-3.1.1.tgz#9cf95e8a6d7568d60b8eda2aeda17ac8944e654b"
```
* Search `yarn.lock` for `ember-source`, version is correctly recorded as `3.1.1`
* `cat node_modules/ember-source/package.json` 
* Actual module version remains `3.1.0` 🐛 🐞 

**What is the expected behavior?**
`ember-source` is now `3.1.1`

**Please mention your node.js, yarn and operating system version.**
```
$ node --version
v9.11.1
$ yarn --version
1.5.1
```
OSX 10.13.4

*Note:* Only workaround we could find was to nuke `node_modules` and reinstall deps

rally25rs:
Wow, how did this break? 😞 confirmed on 1.6.0.

> Only workaround we could find was to nuke node_modules and reinstall deps

you could also do `yarn install --force` and it will force re-copy all the files. I'll dig in and see if I can figure out what's going on...

rally25rs:
Debugging notes:

The file `size` and `mtime` are exactly the same between versions 3.1.0 and 3.1.1

```
[...]

PR #6010, opened 2018-06-19
fix(util/fs): use file content instead of mtime to determine equality

rufman:
**Summary**
npm 6 locks mtime of all files in the published tarball to be October
26th 1985, which means mtime is no longer changed when a package
is published with an npm version above 6.

See: https://npm.community/t/tarball-content-timestamps-locked-to-1985/149

Fixes: #5723


arcanis:
I'm sad to see this optimization go, but we won't need it anyway soon 😇

Issue #5726, opened 2018-04-25
Yarn install never ends with (complex ?) workspaces 


no-more:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When launching `yarn install` in workspace directory it never ends(waited all night). But if I rollback to v1.3.2 it completes in 42s.

**If the current behavior is a bug, please provide the steps to reproduce.**
The project is on a private repo so I can't provide a link.

**What is the expected behavior?**
Should end within minutes.

**Please mention your node.js, yarn and operating system version.**
node: v9.11.1
Yarn: 1.6.0, 1.5.1, 1.3.2
Os: Ubuntu v9.11.1 or windows 10

I'm not sure if it could be related but the packages are referenced as `
    "package-a": "git+ssh://git@bitbucket.org/companyname/package-a.git#0.1.1-beta.81",`

no-more:
I created a similar public repo on github:

[nomore-test-workspace.zip](https://github.com/yarnpkg/yarn/files/1945711/nomore-test-workspace.zip)

I have the same issue, with version 1.3.2 installation completes in less than 30s (cache empty).
But with version 1.6.0 it does not seems to end.

rally25rs:
Trying to dig into this a bit but debugging all the recursion is tricky. The change that causes this to slow is somewhere between v1.3.2 and v1.4.0.

The code seems to just be cycling between `package-request` and `pacakge-resolver` recursively building up the entire tree, which gets very deep by the time you build `test-m` in the attached repro example. If you remove `test-g` through `test-m` then the install will finish, but you can also see it print out that it is resolving `zone.js` 3 or 4 times. It seems to be exponentially worse as you add back in the packages (`test-f` depends on `test-e` so all the resolves that happened for `test-e` happen again).

Just thinking through it, I suspect this might have something to do with `resolutions` because yarn always has to build the entire tree, just incase you had a resolution that said `test-m/**/test-g/test-f/**/test-b/test-a: "v1.2.3"` so the entire resolution tree building recursive thing can't be exited just because a lib has been resolved before. Just a thought anyway... no concrete evidence yet...

rally25rs:
I bisected the performance change back to [this commit] (https://github.com/yarnpkg/yarn/commit/6adbd472129b5410436ddefbd92093868bd98e49) which strangely is mine, yet I don't remember making these specific code changes when fixing peerDep warnings 😕 

[...]

PR #6012, opened 2018-06-20
fix(workspaces): Allow patterns with non-semver ranges to match workspace packages

### LINKED ISSUE #3973: Cannot `yarn add` in workspace package if depending on a local package

_bug_

```
$ yarn add whatever
yarn add v0.27.5
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@scope%2fb: Not found".
info If you think this is a bug, please open a bug report with the information provided in "/private/tmp/yarn-issue/packages/a/yarn-error.log".
```

```
git clone https://github.com/aymericbeaumet/yarn-workspace-add-package.git /tmp/yarn-issue
cd /tmp/yarn-issue/packages/a
yarn add whatever
```

- The `@scope/b` should be found as it exists in the workspace
- The package `whatever` should be added as a dependency of the package

node 6.10.2
yarn 0.27.5
osx 10.12.5

### LINKED ISSUE #4878: Yarn workspace fails to add local package as dependency

Bug. (could be expected behavior)

When trying to install a local package using `yarn workspace <workspace-name> add <package-name>` adding the package fails. It however succeeds when specifying the exact version of the local package.

Example repo: https://github.com/kohlikohl/yarn-workspace-install-local-package

❌**Failure case**
```
$ yarn workspace @scope/a add @scope/b
yarn workspace v1.3.2
yarn add v1.3.2
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@scope%2fb: Not found".
info If you think this is a bug, please open a bug report with the information provided in "C:\\dev\\playground\\yarn-workspace-add-package\\packages\\a\\yarn-error.log".
error Command failed.
Exit code: 1
Command: C:\Program Files\nodejs\node.exe
Arguments: C:\Users\maknoll\AppData\Roaming\nvm\v8.6.0\node_modules\yarn\lib\cli.js add @scope/b
Directory: C:\dev\playground\yarn-workspace-add-package\packages\a
Output:

```

✔️ **Working case** (version of package is specified)
```
$ yarn workspace @scope/a add @scope/b@1.0.0
yarn workspace v1.3.2
yarn add v1.3.2
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
success Saved lockfile.
success Saved 2 new dependencies.
├─ @scope/a@1.0.0
└─ @scope/b@1.0.0
Done in 0.17s.
Done in 0.69s.

```

When running `yarn workspace @scope/a add @scope/b` and not specifying a version, I would expect yarn to resolve `@scope/b` to the locally available version and add it as dependency to `@scope/a`'s package.json.

I can, however, see that installing the package from the registry by default is a nice functionality. In which case it might be good to add a  flag that forces a local install. `yarn workspace @scope/a add @scope/b --local` 🤔

Yarn version: 
  1.3.2

Node version: 
  8.6.0

Platform: 
  win32 x64 (windows 10)

rally25rs:
**Summary**

Previously for a requested dependency pattern to match a package in a workspace, it's name had to
match and so did the semver range. If a request specified a non-semver range (like a github url)
then it would not match, causing the resolver to attempt to resolve the package on the registry
instead of use the workspace package. This change will match a workspace package just on it's name
if no valid semver range was found on the requested pattern.

This also fixes the command `yarn workspace @scope/foo add @scope/bar` where `@scope/bar` is a workspace package. The semver range of the request is empty (no version was specified in the command), so it used to not match the package and instead try to find it on the registry. Now it will properly match the workspace package name.

fixes #4878
fixes #3973
affects #5726

**Test plan**

Added new test file for `workspace-layout`.

### DIFF
--- a/__tests__/workspace-layout.js
+++ b/__tests__/workspace-layout.js
@@ -1,0 +1,57 @@
+/* @flow */
+
+import WorkspaceLayout from '../src/workspace-layout.js';
+import Config from '../src/config.js';
+import NoopReporter from '../src/reporters/noop-reporter';
+import type {WorkspacesManifestMap} from '../src/types';
+
+const workspaceMap: WorkspacesManifestMap = Object.freeze({
+  pkgA: {
+    loc: './pkgA',
+    manifest: {
+      _uid: '1',
+      name: 'pkgA',
+      version: '1.2.3',
+    },
+  },
+  '@yarn/pkgB': {
+    loc: './pkgB',
+    manifest: {
+      _uid: '1',
+      name: '@yarn/pkgB',
+      version: '1.2.3',
+    },
+  },
+});
+
+describe('WorkspaceLayout.getManifestByPattern()', () => {
+  test('returns null when workspace does not contain a package with requested name', () => {
+    const layout = new WorkspaceLayout(workspaceMap, new Config(new NoopReporter()));
+    const result = layout.getManifestByPattern('does-not-exist');
+    expect(result).toBe(null);
+  });
+
+  test('returns null when workspace contains a package with requested name but version does not match', () => {
+    const layout = new WorkspaceLayout(workspaceMap, new Config(new NoopReporter()));
+    const result = layout.getManifestByPattern('pkgA@^9.0.0');
+    expect(result).toBe(null);
+  });
+
+  test('returns manifest when workspace contains a package with requested name and version matches', () => {
+    const layout = new WorkspaceLayout(workspaceMap, new Config(new NoopReporter()));
+    const result = layout.getManifestByPattern('pkgA@^1.0.0');
+    expect(result).toBe(workspaceMap.pkgA);
+  });
+
+  test('returns null when workspace contains a package with requested name and version is not a semver range', () => {
+    const layout = new WorkspaceLayout(workspaceMap, new Config(new NoopReporter()));
+    const result = layout.getManifestByPattern('pkgA@git+https://git@github.com/yarnpkg/test-package.git');
+    expect(result).toBe(workspaceMap.pkgA);
+  });
+
+  test('returns manifest when workspace contains a scoped package with requested name and version matches', () => {
+    const layout = new WorkspaceLayout(workspaceMap, new Config(new NoopReporter()));
+    const result = layout.getManifestByPattern('@yarn/pkgB@^1.0.0');
+    expect(result).toBe(workspaceMap['@yarn/pkgB']);
+  });
+});
--- a/src/workspace-layout.js
+++ b/src/workspace-layout.js
@@ -23,9 +23,25 @@ export default class WorkspaceLayout {
   getManifestByPattern(pattern: string): ?{loc: string, manifest: Manifest} {
     const {name, range} = normalizePattern(pattern);
     const workspace = this.getWorkspaceManifest(name);
-    if (!workspace || !semver.satisfies(workspace.manifest.version, range, this.config.looseSemver)) {
+    if (!workspace) {
+      // the dependency specifies a package name that does not exist in this workspace.
       return null;
     }
-    return workspace;
+    if (!semver.validRange(range)) {
+      // if the dependency specifies a range that isn't valid semver.
+      // It is probably an "exotic" pattern like a github url,
+      // or has no range at all as in the case of running the command
+      //   yarn workspace @yarnpkg/pkgB add @yarnpkg/pkgA
+      // For these, we just match on the package name and allow the workspace package to be used.
+      return workspace;
+    }
+    if (semver.satisfies(workspace.manifest.version, range, this.config.looseSemver)) {
+      // the dependency specifies a package name that exists in this workspace
+      // and the workspace package's version matches the requested range. Use the workspace package.
+      return workspace;
+    }
+    // the dependency specifies a package name that exists in this workspace,
+    // but the workspace package is a version that does not match the requeted range.
+    return null;
   }
 }


arcanis:
I don't know how I feel about this change. I have a few remarks:

  - How often will someone add a non-npm dependency of a project and expect it to be fetched from the workspace? Why not use a version, then?

  - The current algorithm is quite simple: if it matches, it's linked. If it doesn't, it will fetch it from wherever it is requested. This PR adds a significant edge case which I think might be quite confusing and unexpected (to me, at least).

  - The `yarn workspace @scope/a add @scope/b` problem should be resolved by using the workspaces versions before resolving to the npm cache (so `@scope/b` would, by default, resolve to the version specified in the local workspace). It seems a different problem.

rally25rs:
@arcanis 

> How often will someone add a non-npm dependency of a project

in https://github.com/yarnpkg/yarn/issues/5726#issuecomment-397785446 @no-more mentions that he is using github refs because it is a set of private packages. That's the only case I've seen.

eropple:
> How often will someone add a non-npm dependency of a project and expect it to be fetched from the workspace? Why not use a version, then?

Hi. I just found this and it bit me really hard. Eventually packages are pushed to a private NPM registry, but during development I use a Yarn workspace to wire them up. My typical convention is to use versions with `-wip` at the end, i.e. `0.1.0-wip`, for internal development; packages are then renamed over time, i.e. `0.1.0-alpha.1`, `0.1.0-rc.1`, etc. before being released as `0.1.0`.

Yarn throwing a rod at this is a really big bummer. I'd hope that the PR under consideration gets some love.

kjs3:
I think this is exactly what my team needs. We publish to some private repos but want to have everything together during dev.

bendemboski:
I'm in exactly the same situation as @kjs3 and the absence of this is currently blocking me from using workspaces, as I haven't found a workaround...

bendemboski:
> I didn't think much about it at the time, but if they are private repos then they will probably never be published to the registry, so it might be fine to use a version or *.

@rally25rs I'm not quite sure how this would work...say I have a workspace with `project-a` that depends on unpublished `dep-b` (also in the workspace), pulled from a git URL. If I modify `project-a`'s `package.json` to specify `"dep-b": "*"` then it works fine in the workspace. But if I then try to use `project-a` *not* in a workspace (like in CI), it can't resolve `dep-b` with version `*` because it's un-published. I even tried adding `"dep-b": "<git url>"` to `resolutions` in `project-a`'s `package.json`, but no dice -- outside the workspace I get

```
error An unexpected error occurred: "https://registry.yarnpkg.com/dep-b: Not found".
```

Am I missing something? I'm happy to specify `*` as the version, but need it to still work outside of workspaces (because I really don't want to have to modify my `package.json` when working in a workspace and always remember not to check it in).

kjs3:
@bendemboski: Not exactly a completely open-ended `"*"` but I talk about that [here in 6027](https://github.com/yarnpkg/yarn/issues/6027). 
A token in the GitHub URL allows it to work anywhere.

rally25rs:
@bestander IIRC this issue is what I tried to PM you on Discord about.

bestander:
I agree that git dependencies are a common thing and we should address this.

I think resolution field would work but it is too hacky and not obvious, matching for semver validity makes more sense to me

arcanis:
Yep, but you have no way to know the version of the remote dependency until you bring it to the disk (at least not without querying the network, which is not ideal since - on top of the obvious problems with `--offline` - it would either require to be resolved each and every time, or to be persisted somewhere, like in the lockfile).

batusai513:
@arcanis @rally25rs Any updates on this one?

mdgbayly:
Not fully versed in all the subtleties of the issue and the proposed solution but it feels to me like there is an inconsistency in how yarn is handling non-standard semver ranges like git urls when it comes to workspace packages.

e.g. outside of workspaces, we can add versions that resolve to git urls like this to two different packages and yarn will resolve/dedupe this to most recent package.

```
component-a/package.json
"my-package": "myorg/my-package#^2.0.11"

component-b/package.json
"my-package": "myorg/my-package#^2.0.10"
```
So it was extremely confusing when we tried to use `my-package` as a workspace package but it never gets linked.

toddbluhm:
Would really like to see this or a similar solution get merged. My company would like to move to `yarn`  and `workspaces` is one of the main features we want to use, but unfortunately this issue is holding us up. All of our packages are privately connected via git urls + tokens and we do not publish to any private or public npm registry. So without this,`workspaces` will not work for us and is stopping us from migrating over.

rally25rs:
I just ran smack into this issue on a new project. I thought it might be a decent idea to have a monorepo that consists of git submodules for each package.

Each package is maintained by a separate dev team, so in order to let them have their own Github PR and issues tracking, the workspace projects are separate github repos.

The problem becomes these teams obviously want to have their own CI and test process, but to do that the dependencies would need to be resolved from github URLs when it's run outside the monorepo (which doesn't work because of this issue with semver resolution).

Otherwise the CI server would have to checkout the entire monorepo and all subprojects, then flip one of them to the PR branch in order to be built/tested.

kwelch:
I believe I ran into this by trying to have an alpha version as one of the workspace values and then in a different workspace linking it as `*`. I was unable to install the package. 

However, when i switched `*` to  the version in the other package it worked as expected. Hopefully that helps someone out there.

ericwooley:
Just wanted to bump this.

https://github.com/yarnpkg/yarn/issues/3973 has been an issue literally for years, and this PR has been open for (less years).

However this is solved, it needs to be solved at some point.

We aren't using workspaces to publish packages, we are just using them to break up main app into modules, so we don't have to do a 20 minute rebuild every time.

Because of this issue, we have to manually add packages and versions to a workspace package.json, and it's super tedious, and lowers my teams confidence in yarn workspaces.

I'm currently considering publishing empty packages to work around this. Which is just silly.

musicin3d:
Why hasn't this been reviewed yet? It seems like there is no way to depend on another package in the same monorepo/workspace without publishing. Am I missing something?

PR #6017, opened 2018-06-21
(WIP) yarn exits with an error when a dependency has no latest tag

### LINKED ISSUE #5955: yarn upgrade-interactive fails with Invalid version: undefined error if a dependency has only prerelease versions
A bug

When running yarn upgrade-interactive and a dependency does not have a "latest" version then it fails with an error "Invalid version: undefined"

* add a dependency that does not have a "latest" tag (only prerelease versions published with tag "beta")
* run yarn upgrade-interactive

The command should run successfully and show the upgradeable versions of all packages.

If the package in question has a later prerelease version, that version should be suggested to be upgraded to.

Arguments: 
  C:\Program Files (x86)\node\node.exe C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\bin\yarn.js upgrade-interactive

Yarn version: 
  1.7.0

Node version: 
  8.11.1

Platform: 
  win32 x64

Trace: 
  TypeError: Invalid Version: undefined
      at new SemVer (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2465:11)
      at SemVer.module.exports.SemVer.compare (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2528:13)
      at compare (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2752:31)
      at Function.lt (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2786:10)
      at isDepOld (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:36817:134)
      at Array.filter (<anonymous>)
      at Function.<anonymous> (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:36822:19)
      at step (C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:98:30)
      at C:\Users\jens.elstner\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:109:13

kpbl:
**Summary**
When a dependency package has no latest tag the commands **yarn upgrade** and **yarn upgrade-interactive** will throw the following error:
An unexpected error occurred: "Invalid Version: undefined".

This is because in method **getOutdatedPackages** of the class **PackageRequest** latest is undefined when no latest tag is found and the call to semver.lt(current, latest) throws an error.

I simply checked if latest is undefined before the call to semver.lt(current, latest) .

fixes #5955 

**Test plan**
I did not know how to write a test for this because i have found no public package that did not has a latest tag. Any advice would be great.

### DIFF
--- a/src/package-request.js
+++ b/src/package-request.js
@@ -440,7 +440,7 @@ export default class PackageRequest {
 
     // Make sure to always output `exotic` versions to be compatible with npm
     const isDepOld = ({current, latest, wanted}) =>
-      latest === 'exotic' || (semver.lt(current, wanted) || semver.lt(current, latest));
+      latest === undefined || latest === 'exotic' || (semver.lt(current, wanted) || semver.lt(current, latest));
     const orderByName = (depA, depB) => depA.name.localeCompare(depB.name);
     return deps.filter(isDepOld).sort(orderByName);
   }


arcanis:
> I did not know how to write a test for this because i have found no public package that did not has a latest tag. 

How did it happen? Private registry?

Also, I think the fix would rather be the following (to avoid so that if there is no latest, then the dependency isn't marked as out of date rather than the opposite).

```js
const isDepOld = ({current, latest, wanted}) =>
      latest === 'exotic' || (semver.lt(current, wanted) || latest === undefined && semver.lt(current, latest));
```

Issue #5810, opened 2018-05-11
Unmet peer deps with workspaces


sky87:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
In workspaces, `peerDependencies` that are also present as `devDependencies` are reported as unmet.

**If the current behavior is a bug, please provide the steps to reproduce.**
Clone https://github.com/sky87/unmet-peer-deps-issue, run

```
$ yarn
yarn install v1.6.0
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "workspace-aggregator-b6d53714-1712-40de-af51-af5d0487ed3e > a@1.0.0" has unmet peer dependency "react@^16.0.0".
warning "workspace-aggregator-b6d53714-1712-40de-af51-af5d0487ed3e > a@1.0.0" has unmet peer dependency "react-dom@^16.0.0".
[4/4] Building fresh packages...
success Saved lockfile.
Done in 4.49s.
```

**What is the expected behavior?**
Yarn should not issue a warning. 

Notice that without workspaces everything works as expected.
```
$ rm -rf node_modules &&  mv package.json _package.json &&  cd packages/a && yarn
yarn install v1.6.0
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
Done in 2.04s.
```

**Please mention your node.js, yarn and operating system version.**
```
$ yarn -v
1.6.0
$ node -v
v9.3.0
$ ver
Microsoft Windows [Version 10.0.16299.431]
```

waynebloss:
I don't think I'm using the workspace feature, but am getting the following error despite the fact that `babel-runtime@6.26.0` is installed. I think it's because I'm using `devDependencies` as well.

### warning "react-intl-cra > babel-preset-react-app@3.1.1" has unmet peer dependency "babel-runtime@^6.23.0".

### My packages:
```json
{
  "dependencies": {
    "babel-polyfill": "^6.26.0",
    "classnames": "^2.2.5",
    "eventemitter3": "^3.1.0",
    "history": "^4.7.2",
    "invariant": "^2.2.4",
    "lodash.debounce": "^4.0.8",
    "path-to-regexp": "2.2.1",
    "prop-types": "15.6.1",
    "query-string": "5.1.1",
    "react": "16.4.0",
    "react-bootstrap": "0.32.1",
[...]

PR #6026, opened 2018-06-24
fix: false negative unmet peer warnings

### LINKED ISSUE #5831: Bug: Yarn install in workspace root reporting false-negative unmet peer dependency warnings
Bug

Consider if you have:

1. `~/repos/foo-app`:
```json
{
  "private": true,
  "name": "@foo/app",
  "version": "1.0.0",
  "dependencies": {
    "@foo/bar": "1.0.0",
    "@foo/baz": "1.5.3",
    "@foo/widget": "2.0.0"
  }
}
```

2. `~/repos/foo-monorepo`:
```json
{
  "private": true,
  "workspaces": [
    "packages/*"
  ],
  "devDependencies": {
    "@foo/widget": "2.0.0"
  }
}
```

With 2 packages:
```json
{
  "name": "@foo/bar",
  "version": "1.0.0"
}
```
```json
{
  "name": "@foo/baz",
  "version": "1.5.3"
}
```

3. `~/repos/foo-widget`:
```json
{
  "name": "@foo/widget",
  "version": "2.0.0",
  "peerDependencies": {
    "@foo/bar": ">=0.5.4",
    "@foo/baz": ">=-0.6.2"
  }
}
```

Then, you run:
```bash
$ cd ~/repos/foo-modules
$ rm -rf node_modules packages/*/node_modules yarn.lock
$ yarn
```

Yarn will report these warnings:
```bash
warning " > @foo/widget@2.0.0" has unmet peer dependency "@foo/bar@>=0.5.4".
warning " > @foo/widget@2.0.0" has unmet peer dependency "@foo/baz@>=0.6.2".
```

After verifying everything is symlinked properly...you realize these warnings are **false-negatives**.

See above

There should be no warnings per this use case.

```bash
$ yarn versions v1.6.0
{ yarn: '1.6.0',
  http_parser: '2.8.0',
  node: '8.11.1',
  v8: '6.2.414.50',
  uv: '1.19.1',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  modules: '57',
  nghttp2: '1.25.0',
  openssl: '1.0.2o',
  icu: '60.1',
  unicode: '10.0',
  cldr: '32.0',
  tz: '2017c' }
✨  Done in 0.04s.
```

hulkish:
Fixes #5831, #5810

### DIFF
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -566,7 +566,7 @@ export default class PackageLinker {
       const ref = pkg._reference;
       invariant(ref, 'Package reference is missing');
       // TODO: We are taking the "shortest" ref tree but there may be multiple ref trees with the same length
-      const refTree = ref.requests.map(req => req.parentNames).sort((arr1, arr2) => arr1.length - arr2.length)[0];
+      const {parentNames: refTree = []} = ref.requests.find(ref => ref.parentNames.length) || {};
 
       const getLevelDistance = pkgRef => {
         let minDistance = Infinity;


hulkish:
@sky87 @waynebloss can u guys give this a try?

hulkish:
these tests succeed locally for me, not sure whats going on.

hulkish:
@nomore

__tests__/integration.js
```diff
@@ -232,7 +232,9 @@ describe('--registry option', () => {
       await runYarn(['add', 'is-array', '--registry', registry], {cwd});
     } catch (err) {
       const stdoutOutput = err.message;
-      expect(stdoutOutput.toString()).toMatch(/getaddrinfo ENOTFOUND example-registry-doesnt-exist\.com/g);
+      expect(stdoutOutput.toString()).toMatch(
+        /getaddrinfo ENOTFOUND example-registry-doesnt-exist\.com|example-registry-doesnt-exist.com\/is-array: connect ECONNREFUSED/,
```
hulkish: This was necessary in order for tests to pass locally, for me. Yet, it is still valid because it can depend on several network related events. In my case, it's probably my ISP.

rally25rs: to elaborate, my ISP's DNS servers will return a canned IP address for any invalid lookups that directs you to some static search page that they have set up, so you get an ECONNREFUSED instead of the ENOTFOUND.

src/package-linker.js
```diff
@@ -566,7 +566,7 @@ export default class PackageLinker {
       const ref = pkg._reference;
       invariant(ref, 'Package reference is missing');
       // TODO: We are taking the "shortest" ref tree but there may be multiple ref trees with the same length
-      const refTree = ref.requests.map(req => req.parentNames).sort((arr1, arr2) => arr1.length - arr2.length)[0];
```
arcanis: I don't think this is correct. The sort is there to have a consistent iteration order across yarn runs. If it works without the sorting, it doesn't mean that the bug is fixed, just that it doesn't appear on whatever random order the data have been layout.

rally25rs:
IMO this is testing a weird case. The scenario in this test case is:

* workspace contains a project named `test-a`
* workspace aggregator depends on `test-h`
* `test-h` has a peerDep on `test-a`
* yarn says: `"test-h" has unmet peer dependency "test-a"`

which I would argue is correct. Workspace aggregator doesn't _really_ have a dependency on `nomore-test-a`, it just has a workspace package named that.

I guess this could be a debate over whether or not the root workspace package actually "contains" it's workspace packages as a dependency. To me it feels weird to have a workspace root dependency that comes from outside the workspace, that has a peerDep on a package in the workspace.

rally25rs:
In the case of #5810 I would think that no peerDep warning would show because nothing is depending on `a`, except that is does look like the "workspace aggregator" depends on it and so it is being checked for peerDeps, which seems like the opposite of the case above where the workspace package `test-a` isn't satisfying the peerDep.... So maybe the workspace root _does_ "depend" on workspace packages?

IMO The overall intent of the workspace design isn't really clear here... I suspect this is the result of an implementation detail of the aggregator pretending to be a package that depends on all the workspace packages, more than design intent, but I could be wrong...

hulkish:
@rally25rs keep in mind the emphasis on **transient dependencies**.

src/package-linker.js
hulkish: yes, you're correct that `[].every(() => false) === true`. However the issue is not with `req.parentNames`. The`[<value>].every((name, idx) => name === refTree[idx])

hulkish:
@rally25rs 
> To me it feels weird to have a workspace root dependency that comes from outside the workspace, that has a peerDep on a package in the workspace.

Not at all. For example, we use a monorepo but have several legacy packages that are in individual repos. So, this is definitely a valid use case.

src/package-linker.js
NilSet: Am I correct that the goal here is to skip refs with parentNames.length === 0? If so you should just filter those ones out and then sort.

PR #6033, opened 2018-06-25
fix(resolutions): Warn instead of erroring if unmatched dependency matches resolutions with glob support

ramasilveyra:
**Summary**

PR #4835 fixed `yarn check` while using simple resolutions keys (like `repeat-string`).

This PR fixes `yarn check` while using resolutions keys with globs (like `**/repeat-string`) or dep trees (like `pad-left/repeat-string`).

**Test plan**

New test case added.


arcanis:
Good for me! Thanks! :)

PR #6039, opened 2018-06-27
Install should not link a workspace if the version is compatible but pinned down

tajo:
**Situation**

We use workspaces to keep and develop our applications and libraries together in one repository. 
Everything is using the latest versions and kept in sync so yarn can symlink all workspace dependencies. Lerna is used to automate publishing and package.json updates.

**Problem**

Sometimes we want to introduce a major change in one of our libraries (`my-lib`) but we don't have time to go through all applications that consume it and update them as well. In that case, we can manually pin down the version of `my-lib` in all places where it's being referenced as a dependency. 

We would simply change `"my-lib": "^x.y.z"` to `"my-lib": "x.y.z"` and proceed with major changes in `my-lib`. Once we merge those changes and run `lerna publish`, the new workspace version of `my-lib` will not be compatible with `"my-lib": "x.y.z"` anymore and yarn will stop symlinking `my-lib`. **The problem** is that until we publish the new version of `my-lib`, yarn still keeps everything symlinked since versions are compatible. That's problematic because we like to run our global test suite (all applications + libraries) even when changing a single library. However, in this case, `my-lib` changes are breaking and tests run before the new version is published, so everything is still symlinked and tests can't pass.

This PR adds a test that demonstrates this. Imagine this file structure:

```
repository
  ├─ packages
  │   └─ left-pad
  │     └─ package.json
  └─ package.json
```

`repository/package.json`:

```json
{
  "name": "my-project",
  "private": true,
  "dependencies": {
    "left-pad": "1.1.2"
  },
  "workspaces": [
    "packages/*"
  ]
}
```

`repository/packages/left-pad/package.json`:

```json
{
  "name": "left-pad",
  "version": "1.1.2"
}
```

After running `yarn install`, `left-pad` dependency in `my-project` will be symlinked to `repository/packages/left-pad`. However, since the version is pinned down, it would make sense to **remove the symlink** and download version `1.1.2` from npm. By pinning down `left-pad`, we declared that we **want to use that specific version and changes in our workspace should not be propagated**.

**Solution**

If workspace version is pinned down, don't symlink it but download it from npm. 

*I'm happy to contribute if you give me some hints. So far I just created this failing test.*

**Workarounds**

After you pin down the `left-pad` version in `/package.json`, you can also manually update `/packages/left-pad/package.json - version` to non-existent future version `1.1.3`.  That will force yarn to also remove the symlink. However, this creates a minor problem with lerna since it will want to bump the version again and you will end up with a "hole" in your versions.


arcanis:
> If workspace version is pinned down, don't symlink it but download it from npm.

I'd be against such a change, it would be a really unexpected behavior imo.

tajo:
> it would be a really unexpected behavior imo.

Assuming usage of some auto-publishing mechanism (with tool as lerna), how do you deal with case like that? Is there some elegant way how to force yarn to not link a specific package so you can introduce breaking changes without updating all consumers of such package?

Using npm version seems like a safe thing since even the smallest local change will make the pinned version not compatible and lerna will not auto-update pinned versions (which makes sense too). Remove `^` = freeze the dependency / prevent any updates.

sejoker:
> I'd be against such a change, it would be a really unexpected behavior imo.

Even if a result of this discussion will be not to change a default yarn behavior, I think yarn should allow to explicitly unlink specific package in the workspace.

arcanis:
I believe using `npm:left-pad@1.1.2` should circumvent the workspaces.

arcanis:
I don't think we should do this, so I'm going to close this to clear the queue, but feel free to discuss it.

Issue #6057, opened 2018-07-02
Yarn cannot resolve relative paths in .yarnrc correctly


dereli:
**Do you want to request a *feature* or report a *bug*?** Bug

**What is the current behavior?**
If *cafile* variable in the .yarnrc file is relative, the path to file is resolved relative to current working directory, instead it should be relative to the path of the .yarnrc file.

**If the current behavior is a bug, please provide the steps to reproduce.**

https://gist.github.com/dereli/948095221b45fa2e623070a03b1184bb

**What is the expected behavior?**
Yarn should be able to find correct CAFile resolving the path relative to .yarnrc.

**Please mention your node.js, yarn and operating system version.**
$ yarn -v
1.7.0
$ node -v
v8.9.4
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.5
BuildVersion:	17F77

dereli:
@arcanis thanks for the information.

I spent some time on this, but those changes don't solve the problem immediately.

As you suggested, I added the `cafile` key to the lists. This triggers a `path.resolve` with correct params; however, just right after path resolution, it tries to `fs.mkdirp` and it fails with an `EEXIST` error. This makes sense for the rest of the options but not for the `cafile` option.

paul-soporan:
Closing as this has been fixed in v2.

https://yarnpkg.com/getting-started/migration

PR #6071, opened 2018-07-08
fix(config): relative cafile path works correctly

dereli:
**Summary**

When the cafile variable in the .yarnrc file is relative, the path to
file was resolved relative to current working directory not to the path
of the .yarnrc file.

Fixes: #6057

**Test plan**

Tests are added.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -501,6 +501,34 @@ test('relative cache folder', async () => {
   expect(await fs.realpath(path.dirname(stdoutOutput.toString()))).toEqual(await fs.realpath(`${base}/foo`));
 });
 
+test('relative cache folder path is a file', async () => {
+  const base = await makeTemp();
+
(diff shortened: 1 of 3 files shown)


arcanis:
I unfortunately don't think this work. It will silently tolerate a directory settings (like `cache-folder`) to be a file or, even worse, not to exist at all (if one of its ancestors is a file). It would be better to have a `FOLDER_SETTINGS = ['cache-folder', ...]` whitelist that would decide whether to `mkdirp` or not.

dereli:
@arcanis Removed the try catch and introduced a new list of relative folder keys (`RELATIVE_KEYS_MAKE_DIR`), before mkdirp, we now check if the key is a folder.

Also, although the cases you mentioned didn't fail but not silently after update, I added tests for those two cases to make sure they always fail properly.

dereli:
@arcanis can you take a look here?

PR #6093, opened 2018-07-12
Add yarn.ps1, fixes #6092 (bad Ctrl C behavior)

### LINKED ISSUE #6092: Ctrl C doesn't work, cmd is no longer maintained, yarn.cmd and yarnpkg.cmd should be ps1
Per https://github.com/Microsoft/console/issues/217 - originally created when I wanted to be able to `Ctrl-C`  things like `yarn` and `gulp` without being confirmed - Microsoft strongly recommend against creating new `.cmd` files.

From Microsoft (and echoed repeatedly)

> If you're writing Windows Command-Line scripts today, they should be PowerShell scripts wherever possible

Moving `yarn.cmd` and `yarnpkg.cmd` (which are tiny) to `.ps1` would allow users to cancel them without having a silly confirmation prompot, and probably also avoid other unforeseen drama due to `cmd` being unmaintained.

mikemaccana:
… #6092

**Summary**

Not a substantial pull request. `cmd` is unmaintained, according to its authors at Microsoft. One of the issues is that the termination behavior (Ctrl C) for `cmd` scripts cannot be modifed, but there are likely other issues with `cmd` being unmaintained. 

**Test plan**

All Windows `yarn` commands still work, with no effect on speed and output.

### DIFF
--- a/bin/yarn.ps1
+++ b/bin/yarn.ps1
@@ -1,0 +1,6 @@
+function Get-ScriptDirectory {
+	$Invocation = (Get-Variable MyInvocation -Scope 1).Value
+	Split-Path $Invocation.MyCommand.Path
+}
+
+node "$(Get-ScriptDirectory)/yarn.js" $args


arcanis:
Can you detail what you mean with "the termination behavior (Ctrl C) for cmd scripts cannot be modifed"? What exactly doesn't work?

mikemaccana:
Hi Arcanis! You can see the discussion with Microsoft re `cmd` termination behavior in the linked bug. Short version is that `cmd` requires Ctrl C twice, and everyone on the `cmd` and console team at Microsoft firmly recommend new softwre not be made using `cmd` scripts.

arcanis:
What do you think about https://github.com/Microsoft/console/issues/217#issuecomment-404969224? It seems like a valid concern.

>It seems to me that cmd scripts are generally used in the place of exe symlinks and shebang scripts, which PowerShell is not a good replacement for - if you switch out yarn.cmd for yarn.ps1, that's fine if you're already in PowerShell, but if you're in cmd, or some other shell, or using ShellExecute or some scripting language's equivalent, you have to write pwsh -c yarn, so then why not get rid of the script altogether and do node yarn.js?

mikemaccana:
@arcanis Comment is accurate but doesn't affect anything: 

 - Yes, if I was calling yarn.js from node, I'd call `node yarn.js` and not use `yarn.ps1`.
 - In the exact same way, right now, I'd call `node yarn.js` and not use `yarn.cmd`.

The purpose of the `ps1` or `cmd` file is a quick alias to be run from the shell. The default shell in Windows is Powershell (see `Win X`). Powershell is also the only maintained shell, according to the `cmd` authors.

parkovski:
I'm the one with the concerned comment. I just did an experiment and found that PowerShell will pick a .ps1 file before a .cmd, so rather than replace the script, can you just add the PowerShell script? That should give you working ^C behavior in PowerShell without breaking cmd.

mikemaccana:
cmd wouldn't be broken.

kumarharsh:
@mikemaccana quick question - are you proposing that the `.cmd` file still be retained?

I followed the [discussion here](https://github.com/Microsoft/console/issues/217) and came to this issue. I created a `yarn.ps1` file in my yarn installation folder and then ran `yarn`. It works very well! Ctrl+C doesn't ask the question `Terminate batch job (Y/N)?`. So I imagined that we could remove the `.cmd` file and the `.ps1` file would be enough.

But if I remove the `yarn.cmd` file, then the `yarn.ps1` script fails - it opens up `yarn.js` in my text editor instead! This is a slightly weird side-effect.

mikemaccana:
@kumarharsh not sure what the purpose of the `cmd` file would be. `yarn.ps` will help default command line users, node can run the `yarn.js` itself. Only reason I could see if for users who explicitly launch `cmd` as an interactive terminal in 2018 - most of those are probably doing that out of ignorance so education might be better. 

Are you using the `ps1` from this commit, or a file you made?

kumarharsh:
I created a `ps1` file with the contents of this PR. It's weird that removing the cmd file stops the ps1 file from working - I'm using pwsh as my default shell fwiw.

arcanis:
> Only reason I could see if for users who explicitly launch cmd as an interactive terminal in 2018 - most of those are probably doing that out of ignorance so education might be better.

Fwiw I hate* powershell even more than I hate cmd, so I tend to use cmd at home. I doubt I'm the only one, so I'd say it's safer to keep both.

(*) Hate is a strong word tbh, it's just that I never got the time to learn the Powershell syntaxes, and I mostly use WSL nowadays anyway, so it's not a priority.

Gudahtt:
Keeping `yarn.cmd` is definitely the safer option.

A common method of determining whether a file is executable in Windows is to compare its file extension with the extensions listed in the `PathExt` environment variable. `PS1` is typically not in that list, but `CMD` is.

`node-which` is one example of a library that uses this method. If `yarn.cmd` was removed, then anything that uses `node-which` directly to find the `yarn` executable would instead find `yarn.js` (because `.JS` is typically in `PathExt`, after `.CMD`). If they then tried to execute `yarn.js` directly on Windows, it would blow up. Libraries like `cross-spawn` are smart enough to also read the shebang from `yarn.js` and execute it with `node` instead, but... a shebang is not conventional in Windows, and most libraries aren't as smart as `cross-spawn`. It's conceivable that removing `yarn.cmd` would break things, even putting `cmd.exe` interactive terminal users aside

mikemaccana:
As requested, I've updated the PR and left yarn.cmd for any people still running cmd.

mikemaccana:
Relevant discussion on npm: https://github.com/npm/rfcs/pull/20

arcanis:
Perfect, thanks!

bin/yarn.ps1
```diff
@@ -0,0 +1,6 @@
+function Get-ScriptDirectory {
+	$Invocation = (Get-Variable MyInvocation -Scope 1).Value
+	Split-Path $Invocation.MyCommand.Path
+}
```
felixfbecker: Is there any difference between the output of this function and $PSScriptRoot?

Gudahtt: `$PSScriptRoot` may not have been considered because it was introduced in PowerShell 3. It would not work for users of PowerShell <3.

mikemaccana: As the author `$PSScriptRoot` wasn't considered because I just learnt about it from @felixfbecker. Want me to update the PR?

felixfbecker: If `$PSScriptRoot` was introduced in PS 3 that would mean it is not supported in Windows 7 - I don't know what the support policy is here from yarn / NodeJS?

waynebloss:
I use `cmd` because I don't like `powershell`, but also because `&&` doesn't work in `powershell` [1] and there are a lot of `package.json` files out there that utilize this very handy trick to run 2 different commands from the same `scripts` entry.

[1] https://github.com/npm/npm/issues/4040#issuecomment-209034612

felixfbecker:
@waynebloss npm always uses `cmd` for scripts, even when running them from PowerShell.

waynebloss:
@felixfbecker OK, thanks. I'm not sure if that was always the case or not - I've been using Node  since very early on in Windows and well, regardless of what I do with `powershell` I always have problems :)

Issue #5991, opened 2018-06-14
Running `yarn link` multiple times return non clear message


JulioJu:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
When we trigger `yarn link` several times, in different folders, it's always the first folder who is used. The warning is not very clear. The second time, we have only a non clear message.
```
warning There's already a package called "package_name" registered.
Done in 0.34s.
``` 

As we could see, we have only a Warning. And we read the "Done" world. Therefore we could believe than this finished with success (thanks the word "Done"), and not deduce than this command has strictely no effect (we have simply a warning, not an error).

This behaviour was introduced by https://github.com/yarnpkg/yarn/pull/1793 from @ducks. It was asked by @TheLarkInn  in https://github.com/yarnpkg/yarn/issues/955. 

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
#!/usr/bin/env bash
# -*- coding: UTF8 -*-

PS4='demonstration@demonstration:\w $ '

set -exu

rm -Rf dependencyWeWantNot dependencyWeWant project
mkdir dependencyWeWantNot
cd dependencyWeWantNot
git clone https://github.com/isaacs/rimraf
cd rimraf
yarn link
cd ../../

mkdir dependencyWeWant
cd dependencyWeWant
git clone https://github.com/isaacs/rimraf
cd rimraf
yarn link # the output show "done" and only a warning
echo $?  # output is 0, not an error code
cd ../../

mkdir project
cd project
yarn init -y
yarn add rimraf
yarn link rimraf
cd node_modules/rimraf
# output is the folder dependencyWeWantNot and not dependencyWeWant,
pwd -P
```

**What is the expected behavior?**
* Maybe it could be cool to have a more clear output.

The second `yarn link` could return instead 
```
warning There's already a package called "rimraf" registered. 
This command could not have the desired effect.
If you have performed this command in another folder with the same name, 
it this other one who are still linked. Trigger `yarn unlink` before.'.
```
or shorter
```
warning There's already a package called "rimraf" registered. 
This command has had no effect.
```

* Or otherwise, an explanation about this side effect could be provided at https://yarnpkg.com/lang/en/docs/cli/link/.

* Or otherwise a new parameter could be added `--force` who force the new `yarn link` to be performed.

* Or otherwise a new parameter `--fail` could be introduced than perform 
```javascript
throw new MessageError(reporter.lang('linkCollision', name));
```
[...]

PR #6106, opened 2018-07-14
Clarified yarn link Warning

irrationalRock:
**Summary**
This PR fixes https://github.com/yarnpkg/yarn/issues/5991.

The message for linkCollision is now:
```
There's already a package called $0 registered. This command has had no effect. If this command was run in another folder with the same name,  the other folder is still linked. Please run yarn unlink in the other folder if you want to register this folder.
```

**Motivation**
As mentioned in the issue, the current warning for `yarn link` is vague and should be more specific.

**Test plan**
When running `yarn link`, it produces this warning now:
```
warning There's already a package called "rimraf" registered. This command has had no effect. If this command was run in another folder with the same name, the other folder is still linked. Please run yarn unlink in the other folder if you want to register this folder.
```

### DIFF
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -167,7 +167,8 @@ const messages = {
 
   binLinkCollision:
     "There's already a linked binary called $0 in your global Yarn bin. Could not link this package's $0 bin entry.",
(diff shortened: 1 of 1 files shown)


arcanis:
Thanks!

PR #6131, opened 2018-07-20
fix(hosted-git-resolver): ensure prepare script is executed for hosted git repo

### LINKED ISSUE #5235: Lifecycle scripts for hosted git and local dependencies not run on install
Bug

When a hosted git dependency is installed, lifecycle scripts are not run (e.g. `prepare`)

Run `yarn add user/repo` or `yarn add github:user/repo`. No lifecycle scripts will be run.

`prepare` should be run on install (like with npm@5)

Node.js 9.4.0
Yarn 1.3.2
MacOS 10.13

I've done some investigation as to *why* this broken. This is what I've found:

https://github.com/yarnpkg/yarn/pull/3553 added the correct behavior, but only for git fetching, meaning the dependency was *actually* fetched via git. 

Because Yarn [recognizes certain git hosts](https://github.com/yarnpkg/yarn/blob/8918da15c36e793c8f5ed7be24197a7b5afc89d4/src/resolvers/index.js#L49-L53), Yarn will install these dependencies via [HTTP/tarball method instead of git](https://github.com/yarnpkg/yarn/blob/543f649d355bb5b3b0b56e1c73c6e3ca93a299c8/src/resolvers/exotics/hosted-git-resolver.js#L227) (presumably for performance). So if this logic is triggered (which will happen with dependencies formatted like `repo/user` or `github:repo/user`, then the appropriate lifecycle scripts won't be run.

Related issues:
- https://github.com/yarnpkg/rfcs/pull/70
- https://github.com/yarnpkg/yarn/issues/1671
- https://github.com/yarnpkg/yarn/issues/2815
- https://github.com/yarnpkg/yarn/issues/2875
- https://github.com/yarnpkg/yarn/issues/3911
- https://github.com/yarnpkg/yarn/issues/4973
- https://github.com/yarnpkg/yarn/issues/5047

I'd be interested in making a PR to resolve this, but figured I'd make an issue first. I'm thinking that the logic for running lifecycle scripts in [`src/fetchers/git-fetcher.js`](https://github.com/yarnpkg/yarn/blob/543f649d355bb5b3b0b56e1c73c6e3ca93a299c8/src/fetchers/git-fetcher.js) should probably be moved into the more generic [`src/fetchers/base-fetcher`](https://github.com/yarnpkg/yarn/blob/543f649d355bb5b3b0b56e1c73c6e3ca93a299c8/src/fetchers/base-fetcher.js), since per https://github.com/yarnpkg/yarn/issues/3911 lifecycle scripts should be run for non-git dependencies as well.

3cp:
**Summary**

Use git-fetcher, not tarball-fetcher to deal with hosted git repo. tarball-fetcher has no knowledge about the difference between normal npm package tarball, and tarball provided by hosted git (which is NOT a proper npm package).

Obviously this fix has performance penalty, but it's necessary to support prepare script.

Another possible solution is to add git-tarball-resolver (extending tarball-resolver), but I don't understand the code base enough to be able to extract hasPrepareScript and fetchFromInstallAndPack out from git-resolver.

closes #5235

**Test plan**

Tested my apps with github public repo, bitbucket public and private repos, all can now trigger prepare scripts.
Tested `yarn import` on my apps.
Tested all above with no yarn cache (`yarn cache clean`) and full yarn cache.

### DIFF
--- a/src/resolvers/exotics/hosted-git-resolver.js
+++ b/src/resolvers/exotics/hosted-git-resolver.js
@@ -4,11 +4,8 @@ import type {Reporter} from '../../reporters/index.js';
 import type {Manifest} from '../../types.js';
 import type PackageRequest from '../../package-request.js';
 import {MessageError} from '../../errors.js';
-import {registries} from '../../registries/index.js';
 import GitResolver from './git-resolver.js';
 import ExoticResolver from './exotic-resolver.js';
-import Git from '../../util/git.js';
-import guessName from '../../util/guess-name.js';
 
 export type ExplodedFragment = {
   user: string,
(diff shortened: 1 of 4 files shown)


src/resolvers/exotics/hosted-git-resolver.js
```diff
@@ -208,24 +116,15 @@ export default class HostedGitResolver extends ExoticResolver {
...
-    }
 
     // If the url is accessible over git archive then we should immediately delegate to
     // the git resolver.
     //
     // NOTE: Here we use a different url than when we delegate to the git resolver later on.
     // This is because `git archive` requires access over ssh and github only allows that
     // if you have write permissions
-    const sshGitUrl = Git.npmUrlToGitUrl(sshUrl);
-    if (await Git.hasArchiveCapability(sshGitUrl)) {
```
huochunpeng: Note I removed the Git check archive and new Git(...).
GitResolver creates `client = new Git(...)`, and `client.init()` does the same thing (check archive and resolve committish).
This just remove duplicated effort.

huochunpeng:
I guess I have to wait for someone cleanup test-macos-node10 and appveyor test on current master.
My branch passed my local test on nodejs v8.

chrisdusyk-iq:
Any updates on where this PR is at? This is functionality we really need as we migrate some old apps to a modern build process with webpack and yarn.

huochunpeng:
It seems yarn team has not cleaned up CI/test setup on the master yet. I don't think any PR is in their consideration now.

If you want a temporary solution, install my branch as your global yarn which I did. Because the repo setup, you cannot install directly from my github repo (it would not build).

Do
```
git clone git@github.com:huochunpeng/yarn.git
cd yarn
git checkout fix-prepare-script-on-hosted-git
npm i
npm run build
npm i -g .
```

rtsao:
Should prepare scripts be called for all dependencies (including tarballs)?

If so, then perhaps this PR should instead move the logic into [`src/base-fetcher.js`](https://github.com/yarnpkg/yarn/blob/543f649d355bb5b3b0b56e1c73c6e3ca93a299c8/src/fetchers/base-fetcher.js) so that this logic is consistent regardless of the kind of dependency.

huochunpeng:
Nope, a tarball can be a valid npm package (like prepared by `npm pack`). If you look my initial PR message, a proper way to support this is to add `git-tarball-resolver`, so we don't have to pay the performance penalty on not using github tarball.

ypresto:
@arcanis Could you review this PR which fixes previously-PR-welcomed issue? Sorry if bothering you!
(PR-welcomed comment is here 😉: https://github.com/yarnpkg/yarn/issues/5235#issuecomment-398549563)

karlhorky:
@arcanis is it safe to assume that this PR will never land? (because Yarn 1 is no longer receiving features?)

merceyz:
Closing as fixed in v2

Issue #6009, opened 2018-06-19
Unable to install the mongoose-post-find package


jamesjnadeau:
```
$ cat /etc/fedora-release 
Fedora release 26 (Twenty Six)
$ yarn --version
1.7.0
$ node --version
v8.11.3
```

**What is the current behavior?**
I'm unable to install the mongoose-post-find package via `yarn add mongoose-post-find`. I see an error stating `An unexpected error occurred: "Invalid argument. Expected non-empty string."`

Here's example output I'm seeing with a fresh project/package.json, first thing to install:
```
$ yarn add mongoose-post-find --verbose
yarn add v1.7.0
info No lockfile found.
   ...
[2/4] Fetching packages...
verbose 1.161 Performing "GET" request to "https://registry.yarnpkg.com/mongoose-post-find/-/mongoose-post-find-0.0.2.tgz".
verbose 1.212 Error: Invalid argument. Expected non-empty string.
    at module.exports.module.exports (/home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:144857:11)
    at module.exports.module.exports (/home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:147588:23)
    at isValidLicense (/home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:53259:23)
    at /home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:97239:142
    at Generator.next (<anonymous>)
    at step (/home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:98:30)
    at /home/jamesn/.nvm/versions/node/v8.11.3/lib/node_modules/yarn/lib/cli.js:109:13
    at <anonymous>
error An unexpected error occurred: "Invalid argument. Expected non-empty string.".
info If you think this is a bug, please open a bug report with the information provided in "/home/jamesn/Clevertech/test/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.

```

jamesjnadeau:
Seems to be related to [this line](https://github.com/FLGMwt/yarn/blob/72cea80657b641cfa2438d242f23e945392bfdf2/src/util/normalize-manifest/util.js#L8):

```
export function isValidLicense(license: string): boolean {
[...]

PR #6134, opened 2018-07-21
Changed isValidLicense Check

irrationalRock:
**Summary**
This PR fixes https://github.com/yarnpkg/yarn/issues/6009.

It no longer throws an error when given `"  "`.

**Motivation**

Yarn would not allow you to install a package if it had ` "  "` in it.

**Test plan**
When running `yarn add mongoose-post-find`, it now runs without producing an error.


Gudahtt:
Also, this bug was also just fixed in version `3.0.5` of `validate-npm-package-license`. Updating to that version instead would fix the problem as well.

irrationalRock:
Thanks for the advice. I updated the package and it no longer throws the error.

Gudahtt:
Thanks! The last thing that comes to mind is that you updated the `yarn.lock` file to point at `3.0.4`, but not the `package.json` file. That should be updated as well, to ensure it stays above or at that minimum version.

irrationalRock:
Thanks for the help. I updated the yarn.lock and package.json file.

package.json
```diff
@@ -46,7 +46,7 @@
     "tar-stream": "^1.6.1",
     "uuid": "^3.0.1",
     "v8-compile-cache": "^2.0.0",
-    "validate-npm-package-license": "^3.0.3",
+    "validate-npm-package-license": "3.0.4",
```
Gudahtt: That `^` should stay there. There's no reason to pin this dependency.

irrationalRock:
Thanks for the suggestion. I changed it to `^`.

arcanis:
Thanks 💯

PR #6143, opened 2018-07-24
Upgrade gulp v4

aduh95:
**Summary**

Update to [Gulp v4](https://github.com/gulpjs/gulp/releases/tag/v4.0.0) and drop [deprecated](https://github.com/gulpjs/gulp-util/commit/5d07a1c5c224f22217c694f0fc19ca91a95709b5) `gulp-util` dependency.
Also, it drops `gulp-watch` as it is now integrated within the gulp API.

**Motivation**

This PR is motivated by the warnings showing up when performing `yarn install` on this repo. Also, it removes `lodash v1.0.2` dependency which is using the deprecated `Buffer` constructor.


gulpfile.js
```diff
@@ -30,14 +29,13 @@ const build = (lib, opts) =>
...
 
-gulp.task('default', ['build']);
-
 gulp.task('build', () =>
   build('lib', babelRc.env[majorVer >= 5 ? 'node5' : 'pre-node5'])
 );
 
-gulp.task('watch', ['build'], () => {
-  watch('src/**/*', () => {
-    gulp.start('build');
```
jacobq: I'm not a `gulp` expert by any means, but it seems like we're doing `gulp.task('build')` a lot in here. Is there a good way to clean that up a bit so it's more terse? (e.g. `const buildTask = gulp.task('build')`?)

Gudahtt: I don't think it matters whether tasks are exported using `gulp.task` or by exporting them from the module, it's just a matter of style.

As for the repeated use of `gulp.task('build')`, it looks like the best way to avoid that is to just use the function reference instead. i.e. `gulp.task('default', build)`

That can't be done as things are currently, because `build` is a helper function. The actual `build` task is an anonymous function that calls the `build` function. I see no reason not to merge those two functions into one. After that, instances of `gulp.task('build')` could be replaced with `build`.

arcanis:
Thanks! 👍

PR #6158, opened 2018-07-26
fix: Show essential output for list commands when using --silent option

### LINKED ISSUE #6127: yarn licenses list --silent broke between 1.6.0 and 1.7.0

Bug

`yarn licenses list --silent` produces no output (except warnings on stderr if applicable).

`yarn licenses list --silent` should produce the same output as `yarn licenses list`, except without the `yarn licenses v[version]` prefix and `Done in [time]s.` suffix lines, as it did in yarn 1.6.0.

I run `yarn licenses list` regularly on CI and output the results to a file. If the file changes I know I need to review the new licenses. But with this new behaviour, I can't use --silent, and the file changes all the time due to the 'Done in' time being slightly variable due to network conditions etc. I also don't want to have to change it when yarn updates.

* Node: 8.9.1 (installed with nvm)
* Yarn 1.7.0, 1.8.0 (installed with installation script)
* Mac OS sierra 10.12.6

rjmunro:
**Summary**

When running a list command, ensure `--silent` option doesn't suppress the  essential output, only supplementary information like how long it took and which version of yarn was used.

Fixes #6127

**Test plan**

The following commands should both produce output - previously they did not:

```
yarn --silent list
yarn --silent licenses list
```

### DIFF
--- a/src/cli/commands/licenses.js
+++ b/src/cli/commands/licenses.js
@@ -115,7 +115,7 @@ async function list(config: Config, reporter: Reporter, flags: Object, args: Arr
       });
     });
 
-    reporter.tree('licenses', trees);
+    reporter.tree('licenses', trees, {force: true});
   }
 }
 export function setFlags(commander: Object) {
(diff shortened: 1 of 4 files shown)


arcanis:
💯

PR #6178, opened 2018-07-30
Make yarn not to add /usr/bin to PATH if already there

### LINKED ISSUE #5935: Node executable path always prepended to PATH when running a package.json script
Feature

There is no way to eliminate the explicit prepending of the node executable directory to the PATH environment variable when running package.json scripts.

Npm has a `--scripts-prepend-node-path` cli switch and config option, which defaults to auto, which *only* prepends the node executable directory to the path if it is not in the path. Also, setting the option (i.e., to `true`) prepends the node executable path just before the existing PATH, but after all of the local `.bin` directories. See https://docs.npmjs.com/cli/run-script.

Yarn, on the other hand, *always* prepends the node executable path (https://github.com/yarnpkg/yarn/blob/ae2de7bcd9e9322551478c7a6da31af72cda3bf4/src/util/execute-lifecycle-script.js#L145-L147). Furthermore, it redundantly prepends the node executable path to the very front of the new path (overriding any local `.bin` directories) if the `--scripts-prepend-node-path` switch is given (https://github.com/yarnpkg/yarn/blob/ae2de7bcd9e9322551478c7a6da31af72cda3bf4/src/util/execute-lifecycle-script.js#L177-L179). This last override was recently added in #5787 to address #5776.

Yarn's behavior is problematic. For example, if node is installed at the system level (e.g., `/usr/bin`), this effectively prepends `/usr/bin` to the path with no recourse, which completely inverts and breaks a normal PATH ordering of having local directories override the global `/usr/bin` directory. We saw this break a package.json script that called out to python, for example (it can break python virtual environments). You can also imagine many other situations when inverting the PATH priority like this can break things.

Furthermore, the behavior introduced in #5787 seems problematic, in that the global node path now overrides any local `.bin` directories prepended to the path, directly contradicting the npm behavior (but having the same option name).

I think npm's behavior here makes sense:
1. By default, only prepend the node executable path to PATH if it is not already in PATH. And in that case, only prepend it just before the existing PATH, but after any local node `.bin` directories, etc (i.e., do the intelligent prepending at https://github.com/yarnpkg/yarn/blob/ae2de7bcd9e9322551478c7a6da31af72cda3bf4/src/util/execute-lifecycle-script.js#L145-L147, and delete the prepending at https://github.com/yarnpkg/yarn/blob/ae2de7bcd9e9322551478c7a6da31af72cda3bf4/src/util/execute-lifecycle-script.js#L177-L179). 
2. Make it possible to explicitly enable or disable this prepending using `--scripts-prepend-node-path`

CC @ivanov

yarn 1.7.0
node 8.11.2
npm 5.6.0
macOS 10.13

x-yuri:
**Summary**

Fixes #5935 

**Test plan**

This one is supposedly going to remain untested:

```diff
commit 3759562d6a4b114723ee7055e7f872153301657b
Author: Yuri Kanivetsky <yuri.kanivetsky@gmail.com>
Date:   Mon Jul 30 12:01:22 2018 +0300

    trying to test...
    
    `yarn test` pollutes PATH with another /usr/bin, and even if
    the child instance of yarn doesn't do that, there are at least two
    /usr/bin's in the PATH

diff --git a/__tests__/fixtures/lifecycle-scripts/usr-bin/package.json b/__tests__/fixtures/lifecycle-scripts/usr-bin/package.json
new file mode 100644
index 00000000..d825e8bd
--- /dev/null
+++ b/__tests__/fixtures/lifecycle-scripts/usr-bin/package.json
@@ -0,0 +1,5 @@
+{
+    "scripts": {
+        "cmd1": "echo $PATH | tr : \\\\n"
+    }
+}
diff --git a/__tests__/lifecycle-scripts.js b/__tests__/lifecycle-scripts.js
index 8f6660ef..aa7369e6 100644
--- a/__tests__/lifecycle-scripts.js
+++ b/__tests__/lifecycle-scripts.js
@@ -178,6 +178,12 @@ test('should throw error when the script ends with an exit code', async () => {
   await expect(execCommand('test', 'script-fail')).rejects.toBeDefined();
 });
 
+test('should not add /usr/bin to PATH if already there', async () => {
+  const stdout = await execCommand('cmd1', 'usr-bin');
+  const occurences = stdout.split('\n').filter(l => l == path.dirname(process.execPath));
+  expect(occurences.length).toBe(1);
+});
+
 if (process.platform === 'darwin') {
   test('should throw error when the script ends with an exit signal', async () => {
     await expect(execCommand('test', 'script-segfault')).rejects.toBeDefined();
```

### DIFF
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -144,7 +144,10 @@ export async function makeEnv(
 
   // Include the directory that contains node so that we can guarantee that the scripts
   // will always run with the exact same Node release than the one use to run Yarn
-  pathParts.unshift(path.dirname(process.execPath));
+  const execBin = path.dirname(process.execPath);
+  if (pathParts.indexOf(execBin) === -1) {
+    pathParts.unshift(execBin);
+  }
 
   // Include node-gyp version that was bundled with the current Node.js version,
   // if available.


x-yuri:
@arcanis That isn't my fault the tests are failing, is it?

arcanis:
No indeed. Thanks for your work, I'll merge this - I have another fix in preparation that will slightly change this behavior (a path *will* be added, but it will only contain node & yarn), but for the time being that sounds like a reasonable approach.

Issue #6174, opened 2018-07-29
1.9.2: Scoped Registries are being ignored


clintwood:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

EDIT: This seems limited to `yarn global ...`

Since a recent (via my distro) update to `yarn 1.9.2`, scoped registry config set in `.yarnrc` or `.npmrc` are no longer being respected/used.
  
**If the current behavior is a bug, please provide the steps to reproduce.**
1) Use `config set` to set a scoped registry by running either one of the following:
```bash
$ npm config set @myscope:registry http://my.private.registry.com
# or
$ yarn config set @myscope:registry http://my.private.registry.com
```
2) Try adding a package from your scoped registry:
```bash
$ yarn global add @myscope/my-private-package
```
3) You'll get an error message that says:
`error An unexpected error occurred: "https://registry.yarnpkg.com/@myscope%2fmy-private-package: Not found".`

**What is the expected behavior?**

It should install the package from your private registry set for the scoped package.  In this example it should use `http://my.private.registry.com` not `https://registry.yarnpkg.com`.

**Please mention your node.js, yarn and operating system version.**

node: 10.7.0
yarn: 1.9.2
Elementary OS Loki (Ubuntu 16.04)

fenduru:
This is also affecting `yarn create`

arcanis:
Fixed in the 1.9.4, now available! Let me know if that fixes the issue for you 🙂

hbetts:
@arcanis the release notes for `1.9.4` do not indicate any changes to the package manager - https://github.com/yarnpkg/yarn/releases/tag/v1.9.4

Could those please be updated to indicate that this issue has been addressed?

arcanis:
As mentioned in the release notes for the 1.9.4, it's only been released because of our release process. Please refer to the release notes for the 1.9.3 for the actual changes 🙂 

https://github.com/yarnpkg/yarn/releases/tag/v1.9.3

fenduru:
@arcanis this gets past the original issue, however now `yarn create @scope/package` fails as it attempts to execute `C:\Users\me\AppData\Local\Yarn\bin\@scope\package` however the binary got installed to `AppData\Local\Yarn\bin\package`

PR #6181, opened 2018-07-30
Set the default value of enableDefaultRc to true

greysteil:
**Summary**

~~Bug fix for #-6174.~~ Previously, the default value for `enableDefaultRc` was false, but that's unexpected, and caused an issue when a Config instance was created in `global.js` without passing the expected default.

The same approach as here is taken for `looseSemver` on line 403.

**Test plan**

This is just setting a default value, so I don't think needs tests.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -404,7 +404,7 @@ export default class Config {
 
     this.commandName = opts.commandName || '';
 
-    this.enableDefaultRc = !!opts.enableDefaultRc;
+    this.enableDefaultRc = opts.enableDefaultRc == undefined ? true : !!opts.enableDefaultRc;
     this.extraneousYarnrcFiles = opts.extraneousYarnrcFiles || [];
 
     this.preferOffline = !!opts.preferOffline;


arcanis:
That seems reasonable. Do you think you could add a test in `__tests__/commands/global.js`? I'll merge the PR once it's done 🙂

greysteil:
Sure thing - anywhere you'd recommend I crib a test from? Couldn't find anything similar.

clintwood:
@greysteil - Please see https://github.com/yarnpkg/yarn/pull/6007#issuecomment-409100276

greysteil:
Ah, @clintwood, that makes sense. In that case this PR doesn't fix #6174 - I've updated the description accordingly. All this PR does is to make the default config for `enableDefaultRc` more expected, which I think is worth doing anyway.

This *will* fix [this comment](https://github.com/yarnpkg/yarn/issues/6174#issuecomment-408979501), because that's caused by the defaults in [this line](https://github.com/yarnpkg/yarn/blob/master/src/config.js#L863-L867) currently being unexpected. Fixing #6174 fully will require an additional PR that propogates through the options you suggested to the call in `global.js` - do you want to put in a separate PR for that, given that you've done all the debugging?

@arcanis - given the above I think all that this PR is now claiming to do is to change the default value of `enableDefaultRc` to `true`.

clintwood:
@greysteil - I would love to but my time is limited and to be honest I'm really not familiar with the `yarn` code base at all and I see on issue #6174, which I raised, that @fenduru is claiming that this issue also affects `yarn create`.

I think it makes sense for @arcanis who originally did PR #6007 that introduced the new options to look into this as I'm sure he must have dug deep to do that PR!

arcanis:
Right now I'm a bit swamped, if one of you could make the change I'd really appreciate it - otherwise I'll look for a fix next week.

clintwood:
@arcanis - have done in #6207

arcanis:
Fixed by #6207 - but thanks anyway, @greysteil !

PR #6204, opened 2018-08-01
Fix perf regression

arcanis:
**Summary**

This diff fixes the perf regression introduced by #6010. New logic:

  - Instead of comparing the files content, we go back to checking the mtime
  - Since we cannot rely on npm to provide the timestamp, we set it when we fetch the files
  - Since the user caches have already been "corrupted" with wrong timestamps, we need to bump it
  - This has the unfortunate side effects that we lose the ability to avoid copying files that didn't change between two different releases (for example if x.y.1 changes a single file, we will still need to copy each file it contains). This is because we don't get the individual file timestamps from npm. 

Notes:

  - Some hack is needed in order to workaround https://github.com/nodejs/node/issues/22070 - specifically, we need to use `futimes` instead of `utimes` when unpacking the files (otherwise we lose the subsecond precision).

  - Instead of bumping the cache, we could special case the 23 Oct 1985. That seems fragile, though.

**Test plan**

Added a test.


src/fetchers/tarball-fetcher.js
```diff
@@ -93,12 +93,39 @@ export default class TarballFetcher extends BaseFetcher {
   } {
     const integrityInfo = this._supportedIntegrity();
 
+    const now = new Date();
```
BYK: Move this closer to where it is used?

src/fetchers/tarball-fetcher.js
```diff
@@ -93,12 +93,39 @@ export default class TarballFetcher extends BaseFetcher {
...
+
+    const fs = require('fs');
+    const patchedFs = Object.assign({}, fs, {
```
BYK: Ugh, why are we patching fs where we already have our own fs functions module?

src/fetchers/tarball-fetcher.js
arcanis: No, it's here because I don't want each file of a single package to have a different timestamp (like it would be the case if the `Date` constructor was called inside the closure where it is used).

src/fetchers/tarball-fetcher.js
arcanis: Because it's not the same. The `tar-fs` package is doing `utimes` call in order to set the files timestamps. This function is broken and doesn't support subsecond timestamps. The only way to solve this without forking `tar-fs` is to give it an api-compatible version of the `fs` module, patched in order to use `futimes` instead of `utimes`.

Ideally the fix should be done upstream, but this issue is problematic enough that it warrants an hotfix to land faster.

arcanis:
I'll check this tomorrow, but I have a high confidence the tests are failing only because of the bumped cache version (which changes the cache path from `/v1/` to `/v2/`).

acao:
@arcanis looks like it failed on an async timeout, maybe try to re-run? not familiar with this suite.

src/fetchers/tarball-fetcher.js
Daniel15: Is it possible to patch `tar-fs` rather than `fs`? That seems like a better idea to me.

arcanis: No, utimes is used within [a closure](https://github.com/mafintosh/tar-fs/blob/master/index.js#L213). I also prefer to patch a function with clearly defined (if a bit buggy) inputs/output rather than a random function from a module that can change without notice.

Also note that this code doesn't modify the global fs implementation. It simply creates a new, local, fs instance and pass it to tar-fs using its custom filesystem option, which seems perfect for this use case.

__tests__/commands/install/integration.js
```diff
@@ -111,6 +112,20 @@ test('installing a package with a renamed file should not delete it', () =>
     expect(await fs.exists(`${config.cwd}/node_modules/pkg/state.js`)).toEqual(true);
   }));
 
+test("installing a new package should correctly update it, even if the files mtime didn't change", () =>
+  runInstall({}, 'mtime-same', async (config, reporter): Promise<void> => {
+    await misc.sleep(2000);
```
imsnif: Any chance we can do this manually somehow rather than prolonging the test?

__tests__/commands/install/integration.js
```diff
@@ -111,6 +112,20 @@ test('installing a package with a renamed file should not delete it', () =>
...
+
+    const pkgJson = await fs.readJson(`${config.cwd}/package.json`);
+    pkgJson.dependencies['pkg'] = 'file:./pkg-b.tgz';
+    await fs.writeFile(`${config.cwd}/package.json`, JSON.stringify(pkgJson));
+
+    const reInstall = new Install({}, config, reporter, await Lockfile.fromDirectory(config.cwd));
+    await reInstall.init();
+
+    expect(await fs.readJson(`${config.cwd}/node_modules/pkg/package.json`)).toMatchObject({version: '2.0.0'});
+  }));
```
imsnif: This test still passes if I revert the changes (speficially in util/fs). Is this a local/os issue for me?

src/fetchers/tarball-fetcher.js
imsnif: I agree this is a good solution in regards to `tar-fs`. It does kind of bloat the file a little. Do you think it'll be possible to move untar-stream outside somehow?

__tests__/commands/install/integration.js
arcanis: I'm afraid not - even with the subsecond fs patch, subsecond timings still won't work on filesystems that don't support it at all (I think it might be the case on CircleCI), so we need to wait at least >1s to be sure that the mtime will change.

Another option might be to mock the fs module here, but it seemed a bit hazardous so I went with the sleep (since the tests are executed in parallel, it doesn't actually take the whole two seconds in practice).

__tests__/commands/install/integration.js
arcanis: It's a test that ensures that #5723 is still fixed. #6010 was missing tests, so I wanted to be sure that the initial bug wasn't coming back after replacing the commit that was meant to fix it.

There is no test for the perf regression since I'm not sure we can make one that wouldn't be super-flaky 😞 To give you an idea, the different of install time on my machine was 6s, which could potentially come from various other sources (slow network, ...).

src/fetchers/tarball-fetcher.js
arcanis: Hm maybe it could be moved in `src/utils/fs`, yeah 👍

__tests__/commands/install/integration.js
imsnif: Legit. I was just looking for something that would fail if these changes accidentally reverted or changed somehow. Maybe as a unit test for untar-stream when it's moved and for fs?

__tests__/commands/install/integration.js
imsnif: Maybe I'm not following (or maybe this doesn't work, I haven't tested it) but won't this do the trick?
https://nodejs.org/api/fs.html#fs_fs_futimes_fd_atime_mtime_callback

PR #6207, opened 2018-08-02
Fixes to PR 6007

### LINKED ISSUE #6174: 1.9.2: Scoped Registries are being ignored

clintwood:
**Summary**

As a result of changes introduced in PR #6007 this PR resolves #6174 and provides a different implementation to #6181.

**Test plan**

All existing tests pass, PR #6007 added a number of tests but it seems the case of `yarn global ...` with scoped registries was not tested.  I cannot identify exactly how or where these tests should be done.  I'd be happy to do them if I'm pointed in the right direction.


arcanis:
Thanks! I think testing this would imply:

- create a new fixture in `__tests__/fixtures/global/` that would contain a yarnrc and an empty pkg.json
- add a test in `__tests__/commands/global.js` that would run `yarn global add` with a scoped package
- check that it works as expected (not sure exactly how to do this part, I guess you can try setting an invalid registry for the scope and expect for the install to fail?)

clintwood:
@arcanis, I have the basis of a test but cannot figure out how to generate and populate `./__tests__/fixtures/request-cache/GET/test-scope-registry.com/...` with the cache responses to my test scoped domain.  Some guidance from other contributors on how to do this would be appreciated!

clintwood:
@arcanis - I've added a test for `yarn global add @test-scope/scoped-module` in
`__tests__/commands/global.js` with supporting fixtures
`__tests__/fixtures/global/add-with-scoped-registry` and
`__tests__/fixtures/request-cache/GET/test-scope-registry.com`.

I think this concludes the PR.

arcanis:
Perfect, thanks!

Issue #4073, opened 2017-08-02
Tests failing with Timeouts on Windows 10


BurtHarris:
**Current behavior:**

When I run `yarn test` in this project on my dev environment, I get a number of tests failing, mostly due to timeouts.    This is NOT happening in the CI environment.   I'm not sure why my setup would be different, I'm running a sold-state disk and have a reliable internet connection w/ 100 mpbs download, 20 mbps upload.    These timeouts seem strange, and I've not had any problems with yarn running slowly in practice (just the opposite, it blows the socks off the npm client :-).

**Example**
```
 FAIL  __tests__\integration.js (100.811s)
  ● yarn add https://github.com/yarnpkg/yarn/releases/download/v0.18.1/yarn-v0.18.1.tar.gz

    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

      at pTimeout (node_modules/jest-jasmine2/build/queueRunner.js:53:21)
      at ontimeout (timers.js:386:14)
      at tryOnTimeout (timers.js:250:5)
      at Timer.listOnTimeout (timers.js:214:5)
```

Any idea what's going on?

**Versions**
```
C:\code\yarn>yarn versions
yarn versions v0.27.5
{ yarn: '0.28.0',
  http_parser: '2.7.0',
  node: '6.11.1',
  v8: '5.1.281.103',
  uv: '1.11.0',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  icu: '58.2',
  modules: '48',
  openssl: '1.0.2k' }
Done in 0.21s.

C:\code\yarn>cmd ver
Microsoft Windows [Version 10.0.15063]
(c) 2017 Microsoft Corporation. All rights reserved.
```

merceyz:
Closing as an external issue

PR #6211, opened 2018-08-03
Disables a tarball fetching test on windows

arcanis:
**Summary**

AppVeyor is constantly timeouting on the tarball fetching test.

**Test plan**

n/a

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -61,10 +61,13 @@ function addTest(pattern, {strictPeers} = {strictPeers: false}, yarnArgs: Array<
 
 addTest('scrollin'); // npm
 addTest('https://git@github.com/stevemao/left-pad.git'); // git url, with username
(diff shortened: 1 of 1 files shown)


Daniel15:
Why does the test load data from the internet, rather than using mocked data?

arcanis:
I think the problem here is that those integration tests are executed by running an actual yarn process. They don't run in Jest, so they can't be mocked. I'm not entirely sure though, maybe @bestander remembers?

Gudahtt:
I was hoping to look into why these are failing soon. It definitely makes sense to skip them for now.

The test suite is extremely slow and flaky on Windows right now, these two being the worst. I was thinking it might be related to this issue: https://github.com/sindresorhus/execa/issues/142

rockwyc992:
Are there any issue, like #4073, to track this problem?

PR #6213, opened 2018-08-03
Prepend node path only if needed, and only after local directories.

### LINKED ISSUE #5935: Node executable path always prepended to PATH when running a package.json script

jasongrout:
Fixes #5935

#6178 is a partial fix for #5935. This adds the following:

1. Don’t prepend the path *after* the local directories (i.e., delete the lines around line 176 of `yarn/src/util/execute-lifecycle-script.js`)
2. Only prepend the path if the node binary found by searching the environment path is different from our executable path (including if no node is found in the path). This takes care of cases where another node binary precedes our node executable in the path.

I haven't actually tested it (I'm not sure how to build yarn yet), but this should at least be a good start to a full fix for #5935.

If adding `which` as a dependency is a problem, we can revert the logic back to just seeing if `execPath` is in the path, which isn't as good. However, the other parts of this PR I think are needed in master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "uuid": "^3.0.1",
     "v8-compile-cache": "^2.0.0",
     "validate-npm-package-license": "^3.0.3",
+    "which": "^1.3.1",
     "yn": "^2.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "uuid": "^3.0.1",
     "v8-compile-cache": "^2.0.0",
     "validate-npm-package-license": "^3.0.3",
-    "which": "^1.3.1",
+    "which": "^1.3.0",
     "yn": "^2.0.0"
   },
   "devDependencies": {
--- a/src/config.js
+++ b/src/config.js
@@ -47,7 +47,7 @@ export type ConfigOptions = {
   childConcurrency?: number,
   networkTimeout?: number,
   nonInteractive?: boolean,
-  scriptsPrependNodePath?: boolean,
+  scriptsPrependNodePath?: ?boolean,
 
   enableDefaultRc?: boolean,
   extraneousYarnrcFiles?: Array<string>,
@@ -423,7 +423,7 @@ export default class Config {
     // $FlowFixMe$
     this.nonInteractive = !!opts.nonInteractive || isCi || !process.stdout.isTTY;
 
-    this.scriptsPrependNodePath = !!opts.scriptsPrependNodePath;
+    this.scriptsPrependNodePath = opts.scriptsPrependNodePath == undefined ? true : opts.scriptsPrependNodePath;
 
     this.requestManager.setOptions({
       offline: !!opts.offline && !opts.preferOffline,
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -150,10 +150,9 @@ export async function makeEnv(
   // scriptsPrependNodePath=true setting is roughly equivalent to the npm
   // setting scriptsPrependNodePath='auto'.
   if (config.scriptsPrependNodePath) {
-    const execBin = path.dirname(process.execPath);
-    const nodePath = which.sync('node', {path: envPath, nothrow: true});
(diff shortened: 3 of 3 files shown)


arcanis:
That sounds like a duplicate of #6178, right?

jasongrout:
See the description above for the comparison.

jasongrout:
It looks like which needs to be added to the yarn.lock, but I am having troubles doing `yarn install` on master right now (getting some error about fsevents).

Would it be better (i.e., easier to merge) if I reverted the `which` addition for now, and just did the deleting the logic around line 176 and putting the scriptsPrependPath switch above?

Gudahtt:
> I am having troubles doing `yarn install` on master right now (getting some error about fsevents).

Are you referring to the `info fsevents@1.1.1: The platform "win32" is incompatible with this module.` messages? Those are expected, and do not indicate there is any problem.

Though, `which` is already in the lockfile as `v1.3.0`. If you change the range you specified in `package.json` to `^1.3.0`, no lockfile change will be required.

jasongrout:
> If you're still having trouble with that, which is already in the lockfile as v1.3.0. If you change the range you specified in package.json to ^1.3.0, no lockfile change will be required.

Thanks! I did that and it seems like things build now.

Gudahtt:
The approach taken here does seem strictly better than the current behavior. It's not quite as good as the one [described here](https://github.com/yarnpkg/yarn/issues/5935#issuecomment-406326879), but I suppose that could come later.

This could use some additional tests. It looks like none of the PATH changes are tested at the moment though.

arcanis:
Yep, I agree regarding the tests - can you add one in `__tests__/integration.js`? You should be able to use `runYarn` to spawn Yarn with a different PATH variable.

jasongrout:
Closing in favor of #7057.

PR #6222, opened 2018-08-03
test: Fix `yarn run <failing script>` test on Windows

Gudahtt:
**Summary**

This test was failing on Windows with the following error:
```
'false' is not recognized as an internal or external command,
```

The script used for this test is `false`. This is valid in Bash, but is
not valid in cmd.exe. It has been replaced with `exit 1`, which will
exit with a status of `1` in both Bash and cmd.exe

The reason this test was passing on AppVeyor is that they have Cygwin
installed by default. This includes a binary called `false`.

**Test plan**

N/A


arcanis:
👍

PR #6225, opened 2018-08-04
chore(lockfile): update yarn.lock due to nodejs/node#20285

Simek:
**Summary**

Currently Yarn just after checkout and install cannot be build on Node 10 due to `gulp@3.9.x` related bug: https://github.com/nodejs/node/issues/20285.

It seems like there are some old references in currently commited `yarn.lock` file and after proceeding as mentioned in issue linked above (regenerating lockfile) problem disappeared.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,13 +3,13 @@
 
 
 "@gulp-sourcemaps/identity-map@1.X":
-  version "1.0.1"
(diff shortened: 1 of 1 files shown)


Simek:
Unfortunately due to age of some entries in previous lockfile and many incompatibilities with newer packages versions after regeneration I am forced to ditch the idea to fix that issue.

Gudahtt:
This problem will presumably be resolved with this PR as well, which updates to Gulp v4: https://github.com/yarnpkg/yarn/pull/6143

PR #6194, opened 2018-08-01
Bumps hawk

arcanis:
**Summary**

Bumps hawk in one of our test fixtures. It doesn't really matter because no code was ever run, but at the very least it will avoid other warnings in the future.

fixes #6191 

**Test plan**

CI

PR #6227, opened 2018-08-05
test(request-cache): Update request-cache `hawk` version

Gudahtt:
**Summary**

The version of `hawk` used in the `upgrade` test was updated recently in #6194 . The new version of
`hawk` and any dependencies should be in the request cache, so that the test suite is faster, more
reliable, and uses less bandwidth.

**Test plan**

N/A


Gudahtt:
Yep...

You could try checking out the commit before this, running the test suite, committing these files, then merging in my PR to see if they conflicted. You'd have a hard time telling what was different if they did conflict though.

PR #6244, opened 2018-08-08
Added ability to run script on all workspaces

kwelch:
**Summary**
Added the ability to run a script on every workspace in a project. 
`yarn workspaces run script`

Motivation was to be able to run build or test across an entire monorepo during a CI process. While it was possible to run a script on a workspace individually by name `yarn workspace workspace-1 script`. This required you to know and list all workspaces. 

**Test plan**
I added more tests that ensure that the commands are being spawned properly. Since most of the code was copied from functioning parts of the project, I felt this was sufficient. 

*Update*
Based on feedback I validated that arguments can be passed. 

Here were the commands I ran. 
- `cd __tests__/fixtures/workspace/run-basic`
- `yarn workspace workspace-1 add jest`
- `yarn workspace workspace-2 add jest`
- `yarn workspace workspace-1 run jest --all` --> sanity check
- `yarn workspaces run jest --all`

While jest doesn't pass because there are no tests, you can see from the output the arguments are passed properly. 
Here are the spawned child commands
```
$ /Users/kwelch/_git/yarn/__tests__/fixtures/workspace/run-basic/packages/workspace-child-2/node_modules/.bin/jest --all
$ /Users/kwelch/_git/yarn/__tests__/fixtures/workspace/run-basic/packages/workspace-child-1/node_modules/.bin/jest --all
```

### DIFF
--- a/__tests__/commands/workspaces.js
+++ b/__tests__/commands/workspaces.js
@@ -1,12 +1,17 @@
 // @flow
+jest.mock('../../src/util/child');
 
 import {BufferReporter} from '../../src/reporters/index.js';
 import {run as workspace} from '../../src/cli/commands/workspaces.js';
 import * as reporters from '../../src/reporters/index.js';
 import Config from '../../src/config.js';
 import path from 'path';
+import {NODE_BIN_PATH, YARN_BIN_PATH} from '../../src/constants';
 
 const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'workspace');
+const spawn: $FlowFixMe = require('../../src/util/child').spawn;
+
+beforeEach(() => spawn.mockClear());
 
 async function runWorkspaces(
   flags: Object,
@@ -39,6 +44,20 @@ test('workspaces info should list the workspaces', (): Promise<void> => {
         workspaceDependencies: ['workspace-1'],
         mismatchedWorkspaceDependencies: [],
       },
+    });
+  });
+});
+
+test('workspaces run should spawn command for each workspace', (): Promise<void> => {
+  const originalArgs = ['run', 'script', 'arg1', '--flag1'];
+  return runWorkspaces({originalArgs}, ['run', 'script', 'arg1', '--flag1'], 'run-basic', config => {
+    expect(spawn).toHaveBeenCalledWith(NODE_BIN_PATH, [YARN_BIN_PATH, 'script', 'arg1', '--flag1'], {
+      stdio: 'inherit',
+      cwd: path.join(fixturesLoc, 'run-basic', 'packages', 'workspace-child-1'),
+    });
+    expect(spawn).toHaveBeenCalledWith(NODE_BIN_PATH, [YARN_BIN_PATH, 'script', 'arg1', '--flag1'], {
+      stdio: 'inherit',
+      cwd: path.join(fixturesLoc, 'run-basic', 'packages', 'workspace-child-2'),
     });
   });
 });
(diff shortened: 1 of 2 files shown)


arcanis:
Looks good, thanks! How are options handled? Will `yarn workspaces run jest --all` work? It seems like the `--all` will get swallowed by the CLI parsing.

We have some mechanisms to avoid this, can you check whether it could be easily done here?: https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L183-L199

__tests__/commands/workspaces.js
```diff
@@ -42,3 +47,17 @@ test('workspaces info should list the workspaces', (): Promise<void> => {
     });
   });
 });
+
+test('workspaces run should spawn command for each workspace', (): Promise<void> => {
```
kwelch: @arcanis

I thought test would cover the passing of the arguments, or am I missing something. It sounds like I may need to either leave `run` in the arg list or add an additional `--` for the user?

arcanis: The problem with those tests is that when you call `runWorkspaces`, it directly invokes the command, without going through our CLI code. If it did, you would see that the `--flag1` flag would be interpreted as a command-line option for the `yarn workspaces run` command, and wouldn't actually be passed to the argument array.

kwelch:
@arcanis

I updated the description with some more testing that I ran to verify that the command was properly invoked with the correct arguments passed.

__tests__/commands/workspaces.js
kwelch: @arcanis Can you confirm if my additional testing steps are adequate?

src/cli/commands/workspaces.js
```diff
@@ -60,10 +62,45 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
...
+      const {loc} = workspaces[workspaceName];
+      childProcesses.push(
+        child.spawn(NODE_BIN_PATH, [YARN_BIN_PATH, ...rest], {
+          stdio: 'inherit',
+          cwd: loc,
+        }),
+      );
+    }
+
+    await Promise.all(childProcesses);
```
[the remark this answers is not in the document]
kwelch: Good to know. That makes total sense!

donaldpipowitch:
Thank you! Can't wait to test this 😍

dsifford:
Any plans for documentation updates for this one?

jahed:
RE: "Why not just skip over the ones that don't have it?"

Similar to how test runners fail when no tests are found, if yarn did skip non-existent scripts, it may cause unintended errors (typos, forgetfulness). So failing is a safer default.

Also, to me, `workspaces run <script>` says "run <script> on all my workspaces", where behaviour in `npm run`, `yarn run` and `yarn workspace <workspace> run` has always been to fail if <script> doesn't exist.

Something like `--ignore-missing` as a flag would then say "run <script> on all my workspaces, and ignore the ones that don't have it".

Personally, I prefer explicit per-workspace "ignores" than implicit ones. So in a monorepo, I'd expect scripts being run globally from the top level to exist in every package for consistency. You can easily add a script to a workspace to say something like "echo 'No tests.'" if there's nothing to do and it makes your intentions explicit.

kwelch:
To echo @jahed, the intent was to match workspace run, but for all workspaces so failing on missing was intended. 

@dsifford, as for docs. I already have a PR up for it. Looks like it needs to be merged, I will post there. https://github.com/yarnpkg/website/pull/855

dsifford:
Got it, thanks.

Just for sake of discussion, why not have `yarn workspaces run` run scripts defined in the top-level `package.json` on all workspaces as fallback behavior? I can imagine a situation where it might become tedious to add an explicit run script to all workspaces when you are managing a bunch of workspaces at a time.

Is there any use-case for top-level scripts currently?

arcanis:
Too magical imo. Once we implement nested workspaces, "top-level scripts" will become even more blurry.

I like Yarn scripts being simple, because they're easy to reason about.

kwelch:
I agree with above. IMO, the power of workspaces is that any workspace could be removed and put into their own repo without breaking anything.

PR #6249, opened 2018-08-09
test(perf): Set maxWorkers to match the number of cores available

Gudahtt:
**Summary**

AppVeyor VMs have 2 cores, yet we were telling Jest to use 3 workers. Changing it to 2 should reduce
overhead and speed up the test suite.

From the limited testing I've done on AppVeyor, the speed improvement is 10% at best, probably less. But every bit helps, especially when we're seeing tests fail from timeouts.

**Test plan**

I tested this by running AppVeyor builds. They completed successfully, and were marginally faster on average.


Gudahtt:
Hmm. It looks like it did help, but some tests are still timing out.

Ah well. At least it has improved.

PR #6251, opened 2018-08-09
fix(perf): Filter queue in PackageHoister.seed instead of splicing

panrafal:
**Summary**

For big repositories with thousands of packages splicing the queue array can take upwards of 60s to complete. Changing the algorithm to filter the queue only once solves this.

Our monorepo contains more than 2k packages at the top level, with a workspace of more than 200 packages. Yarn got very slow lately, with `install --force` taking more than 4 minutes easily - mostly hanging at "Linking dependencies" with no progress updates.
I profiled yarn with `ndb` and found the culprit to be native `Array.splice`. After this change, it's still the biggest offender but I couldn't find any more obvious call sites to fix.
With the fix we went down from 240s to 50s for `yarn install --force` on Node 10.1.  After upgrading to 10.8 we went down even further to 24s.

**Test plan**

This is a minor change and the touched code is already tested.

### DIFF
--- a/src/package-hoister.js
+++ b/src/package-hoister.js
@@ -163,8 +163,11 @@ export default class PackageHoister {
       while (queue.length > 0 && hasChanged) {
         hasChanged = false;
 
-        for (let t = 0; t < queue.length; ++t) {
-          const pattern = queue[t][0];
+        const queueCopy = queue;
+        queue = [];
+        for (let t = 0; t < queueCopy.length; ++t) {
+          const queueItem = queueCopy[t];
+          const pattern = queueItem[0];
           const pkg = this.resolver.getStrictResolvedPattern(pattern);
 
           const peerDependencies = Object.keys(pkg.peerDependencies || {});
@@ -172,14 +175,15 @@ export default class PackageHoister {
 
           if (areDependenciesFulfilled) {
             // Move the package inside our sorted queue
-            sortedQueue.push(queue[t]);
-            queue.splice(t--, 1);
+            sortedQueue.push(queueItem);
 
             // Add it to our set, so that we know it is available
             availableSet.add(pattern);
 
             // Schedule a next pass, in case other packages had peer dependencies on this one
             hasChanged = true;
+          } else {
+            queue.push(queueItem);
           }
         }
       }


src/package-hoister.js
```diff
@@ -173,7 +176,8 @@ export default class PackageHoister {
           if (areDependenciesFulfilled) {
             // Move the package inside our sorted queue
             sortedQueue.push(queue[t]);
-            queue.splice(t--, 1);
+            // Removing items from an array in a loop is slow on large repos, but deleting is fast
+            delete queue[t];
```
arcanis: I'm not too fond of the `delete` - it basically circumvent the Flow typing, which isn't great 😕

arcanis:
Btw, I'm a bit curious: how did you manage to reach 2k dependencies at the top-level? Even here at FB we don't have that many. Knowing the use case could be useful!

src/package-hoister.js
Gudahtt: I believe the Flow problem can be resolved by using `queue.filter(Boolean)` below instead, [as explained here](https://github.com/facebook/flow/issues/1026).

panrafal:
To be clear - we have 2k packages in `node_modules`, not in `package.json`.

Our frontend app is pretty big and kept in a single monorepo with ~200 packages. There's the testing platform, GraphQL server and tooling too - so it adds up pretty quickly.

arcanis:
I see, thanks - I tried installing the top 1000 npm packages (which, counting transitive dependencies, makes for ~5k packages), and it installed in ~84s with a warm cache. There was a small pause before the linking started, but less than 10s.

Can you try running this command without your patch and see if you get different results? I'd like to understand under which circumstances these slowdowns happen, and how to reproduce it myself:

```
curl 'https://gist.githubusercontent.com/anvaka/8e8fa57c7ee1350e3491/raw/ab776d4b9015832fbc9035b3d302761408cf6de7/01.most-dependent-upon.md' | grep -F '. [' | grep -E -o '\[[^]]+\]' | grep -o -E '[^][]+' | xargs yarn add --ignore-scripts
```

panrafal:
@arcanis I changed it to push to the new queue, as it's actually simpler (and does practically the same as calling `filter`).

As for recreating this, I timed all versions (1.7.0, 1.9.4, my fix) on our repo and... there's no lag on any one of them now. In the mean time I cleared the cache and reinstalled the modules. Yesterday I could recreate this no problem. Now it works alright 😕. So maybe there needs to be some perfect storm of state. We'll be running the modified version for now and I'll keep an eye if it ever happens again.

arcanis:
Sounds good 🙂 I'll still merge it since the changes make sense. Even if we don't get perfs, it's still better engineering imo 👍

PR #6255, opened 2018-08-09
Adds an auto-add-integrity option

arcanis:
**Summary**

This PR gives a way to disable the behavior that makes Yarn automatically add the integrity in the package.json if it is missing. This helps large codebases to adapt (they won't be blocked until the change can be done).

Note that this PR doesn't obsolete #6248 - I'll have to rebase one of them, but they both are valid on their own.

**Test plan**

Added a test 👍


arcanis:
I don't think we have to, the switch requires an explicit acknowledgement anyway, and missing integrity isn't deprecated, since we'll likely always have to support them.

imsnif:
Regarding always having to support fields without integrity: why is that? I was thinking that a couple more PRs (adding github, urls, local files, etc.) and we can get rid of the shasum and move everything to the integrity field (similar to the way `package-lock.json` works). Or am I missing something?

About explicitivity (that's a word, right?): I'm just worried that people can add it to their file, forget about it and then have everything crash and burn once we (hopefully) remove shasums* from the lockfile.

*by removing shasums I just mean moving the "url suffix" we're currently using to the integrity field. I'm not discussing algorithms, sha1/sha512, etc.

imsnif:
Makes sense... How about taking a page from the [React book ](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) and calling the config something like `dangerously_ignore_integrity_creation`?

arcanis:
Good for me 👍 I'll prefix the name with `unsafe-`

arcanis:
Renamed the yarnrc field to `unsafe-disable-integrity-migration`

arcanis:
Tests are passing except for a timeout - @imsnif, all good for you?

imsnif:
Yep, looks cool!

Issue #6105, opened 2018-07-14
Poor performance at scale in PackageInstallScripts#findInstallablePackage


AlexLandau:
Background: I work on a large proprietary monorepo (~100 packages) using Yarn Workspaces. Recently, we've run into an issue where our `yarn install` command spends more than four minutes (over half its runtime) in step 5 of the build ("Building fresh packages"). Under ordinary circumstances, I'd expect most of our time to be spent in step 4 and very little in step 5.

Profiling indicates that nearly all of this time is being spent not in a renegade dependency's install scripts, as I originally guessed, but rather in `findInstallablePackage` (in package-install-scripts.js), specifically in calls to `detectCircularDependencies` from that method. When I add logging, it seems that the following pattern occurs in `findInstallablePackage` when this is run:

* Initially, most of the packages in `workQueue` are immediately installable.
* However, occasionally I get a package that is not installable, and for which `detectCircularDependencies` returns false. When this happens, the method finds another package to return, but _leaves the blocking package at the head of the "queue"_.
* Each subsequent attempt to get a package from the queue goes through the uninstallable package, rerunning the `detectCircularDependencies` check every time.
* The number of packages stuck at the head of the queue in this way grows over time, to as many as 85 in our case, before it decreases very late in the build process.

The result is that while `workQueue` starts with only 2831 values, we end up running `detectCircularDependencies` 76047 times (not counting recursive invocations) and taking over 200 seconds to do so. It's understandable that `detectCircularDependencies` is somewhat CPU-intensive, but the queue logic results in calling it way too many times.

There are a number of possible ways to fix this. Some of my own ideas:

* Change `workQueue` from a Set to an actual queue with an ordering preserved between calls to `findInstallablePackage`. When we test the package at the head of the queue and it is not yet installable, move it to the end of the queue. (Alternatively, move it to a second queue that we switch to when the first queue becomes empty, which would be more confusing but still allow the use of Set objects.)
* Do a cycle-aware topological sort of the dependencies ahead of time.
* Instead of having cycle checking give a yes-or-no result, have it return an uninstalled transitive dependency of the given project unless a cycle exists. (This is like an on-the-fly topological sort.) This would require some restructuring of how it's called currently.

I'm not sure what approach would fit best with the larger parallelized model here.

I'm running yarn 1.8.0 on Node v9.10.0 and OS X 10.13.6. I have also reproduced this issue in 1.7.0. The code that looks relevant appears to be at least two years old.

[...]

PR #6257, opened 2018-08-09
Improve performance of selecting packages in step 5 of install

AlexLandau:
**Summary**

See https://github.com/yarnpkg/yarn/issues/6105 for the full explanation. The short version is that step 5 of yarn install ("Building fresh packages...") is very slow for us for reasons unrelated to actually installing the packages.

This approach avoids repeated CPU-intensive operations on packages that are blocked in the queue by:
1) noting which are blocked on packages currently being installed and can be skipped until something finishes installing, and
2) integrating our checks for cases where we need to break cycles with our search for a package that is ready to install.

The only expected behavior change here is around how cycles of dependencies are broken. The strategy this uses is to try to pick a package such that all of its dependencies are either already installed or known to be in a dependency cycle with it.

**Test plan**

I tested the performance implications of this in our private repository, running `git clean -xdf && time ~/repos/yarn/bin/yarn install` using the versions of yarn in this PR and its parent commit.

Using the parent commit: `Done in 463.70s.`
Using this commit: `Done in 102.03s.`

As for correctness, I expect/hope there are tests from the last set of work making this parallel. I have also run this with enough debug messages to confirm that e.g. we do not have two workers trying to install the same package twice (an issue with a previous attempt).

### DIFF
--- a/src/package-install-scripts.js
+++ b/src/package-install-scripts.js
@@ -209,54 +209,50 @@ export default class PackageInstallScripts {
     await this.install(cmds, pkg, spinner);
   }
 
-  // detect if there is a circularDependency in the dependency tree
-  detectCircularDependencies(root: Manifest, seenManifests: Set<Manifest>, pkg: Manifest): boolean {
-    const ref = pkg._reference;
+  // return the package itself if it's ready to install, a dependency if not, or null if installation is blocked
+  choosePackageToInstall(
+    startingPkg: Manifest,
+    beingInstalled: Set<Manifest>,
+    blocked: Set<Manifest>,
+    installed: Set<Manifest>,
+    seenManifests: Set<Manifest>,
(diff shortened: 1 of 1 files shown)


AlexLandau:
Is there anything more needed for this to be considered? Any timeline? We're stuck on a fork of yarn until this is adopted.

arcanis:
I don't see anything obviously wrong, but I'm a bit worried by the potential side effects this change could have .. 🤔

lnunno:
Any update on this?

jdb8:
I think I'm running into this too after looking at a CPU profile which has a ton of calls to `detectCircularDependencies` - and also experiencing slow installs at the final step. Thanks @AlexLandau for digging into this.

Any chance of this being merged? It seems like it would be a great speedup for projects with a ton of dependencies.

PR #5983, opened 2018-06-13
Correctly resolve scoped packages in `yarn create`.

chriskrycho:
**Summary**

Fixes #5981 by using the scope as part of the path lookup, if it exists.

**Note:** this does *not* yet work on Windows, and it seems like we'll need to do some explicit platform checking to *make* it work on Windows.

**Test plan**

Output on macOS:

```
yarn create @olo/ember-app
yarn create v1.7.0
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...

success Installed "@olo/create-ember-app@0.1.0" with binaries:
      - @olo/create-ember-app
[#######################################################################################] 1007/1008
App name:
```

**Current (broken) output on Windows:**

```
yarn create @olo/ember-app
yarn create v1.6.0
[1/4] Resolving packages...
warning @olo/create-ember-app > ember-cli > exists-sync@0.0.4: Please replace with usage of fs.existsSync
warning @olo/create-ember-app > ember-cli > broccoli-stew > broccoli-funnel > exists-sync@0.0.4: Please replace with usage of fs.existsSync
warning @olo/create-ember-app > ember-cli > ember-cli-preprocess-registry > exists-sync@0.0.3: Please replace with usage of fs.existsSync
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "win32" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
warning Your current version of Yarn is out of date. The latest version is "1.7.0", while you're on "1.6.0".
success Installed "@olo/create-ember-app@0.1.0" with binaries:
      - @olo/create-ember-app
The directory name is invalid.
error Command failed.
Exit code: 1
Command: C:\Users\chris\AppData\Local\Yarn\bin\@olo\create-ember-app
Arguments:
Directory: C:\Code\platform
Output:

info Visit https://yarnpkg.com/en/docs/cli/create for documentation about this command.
```

PR #6271, opened 2018-08-14
fix(create): invoke correct path when using scoped package

### LINKED ISSUE #6266: `yarn create` fails with scoped packages in 1.9.4

Bug

```
$ yarn create @scope/foo
yarn create v1.9.4
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...

success Installed "@scope/create-foo@5.1.2" with binaries:
      - create-foo
[##################################################################################################################################################################################################################-] 784/786The system cannot find the path specified.
error Command failed.
Exit code: 1
Command: C:\Users\me\AppData\Local\Yarn\bin\@scope\create-foo
Arguments:
Directory: C:\Users\me\Documents\dev\create-foo-test
Output:

info Visit https://yarnpkg.com/en/docs/cli/create for documentation about this command.
```

The binary is being installed to `AppData\Local\Yarn\bin\create-foo` however `yarn create` is trying to execute a file with the `@scope` in the path. This is a regression from 1.7.0 where this worked properly.

Node 8.9.4, Yarn 1.9.4, Windows 7

Possibly related to #6174

fenduru:
**Summary**

When running yarn create, first the package is installed globally,
however after installing create was attempting to invoke the binary from
within a scope folder that did not exists. The corrects the path
resolution to invoke the binary from the actual place it was installed
to.

fixes #6266

**Test plan**

This basically just removes the change introduced in #5983. I'm happy to add tests if needed but I'm unsure how to accomplish that.

### DIFF
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -60,7 +60,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   await runGlobal(config, reporter, {}, ['add', packageName]);
 
   const binFolder = await getBinFolder(config, {});
-  const command = path.resolve(binFolder, path.basename(commandName));
+  const command = path.resolve(binFolder, commandName);
 
   await child.spawn(command, rest, {stdio: `inherit`, shell: true});
 }
--- a/src/cli/commands/create.js
+++ b/src/cli/commands/create.js
@@ -56,11 +56,11 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     throw new MessageError(reporter.lang('invalidPackageName'));
   }
 
-  const {fullName: packageName, scope: packageDir, name: commandName} = coerceCreatePackageName(builderName);
+  const {fullName: packageName, name: commandName} = coerceCreatePackageName(builderName);
   await runGlobal(config, reporter, {}, ['add', packageName]);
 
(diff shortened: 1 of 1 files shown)


arcanis:
Looking in more details #5983 I think you're right. The initial issue was making the wrong assumption that Yarn was supposed to install binaries into `node_modules/.bin/@scope`. The confusion was caused by the `@olo/create-ember-app` package, which explicitly references the script name as `@olo/create-ember-app` (with the scope included).

I updated your code to also remove the basename, which I think was the fix we should have done back then (it seems to have been introduced in #3848, which was doing unrelated changes).

fenduru:
Awesome thanks @arcanis. Would it be possible to release a patch with this change?

fenduru:
@arcanis hate to be a pain here, but just wanted to bump to see if we could get a patch of this as we're currently rolling out some tooling that utilizes yarn create at my company

arcanis:
Sure! I'll release a 1.9.5 with this change cherry-picked somewhere this week, friday at the latest 🙂

fenduru:
@arcanis which friday? ;)

arcanis:
The Friday that's tomorrow? 👼 Sorry for the delay! 😬

arcanis:
The 1.10 should be released (as prerelease for now) 👍

PR #6208, opened 2018-08-02
deps: upgrade deps to get rid of Buffer constructor

aduh95:
**Summary**

Node 10 is about to become LTS and Yarn users are still receiving warnings from the use of the deprecated `Buffer` constructor. This PR updates the dependencies triggering those warnings.

Motivated by https://github.com/yarnpkg/yarn/pull/5934#issuecomment-406346724 (thank you @jacobq).

We've come from a long way to get rid of those warnings in yarn. Here the list of the packages which have been upgraded related to the issue #5477:

- [x] yarn (Fixed by: https://github.com/yarnpkg/yarn/pull/5731)
- [x] tar-fs (Fixed by: https://github.com/yarnpkg/yarn/pull/5752)
- [x] tar-stream (Fixed by: https://github.com/yarnpkg/yarn/pull/5752)
- [x] v8-compile-cache (Fixed by: https://github.com/yarnpkg/yarn/pull/5763)
- [x] natives (Fixed by: https://github.com/yarnpkg/yarn/pull/5763)
- [x] sshpk (Fixed by: https://github.com/yarnpkg/yarn/pull/5934)

The list of dependencies that are fixed by this PR:

- [x] bl (BufferList) (1.2.1 -> 1.2.2)
- [x] buffer-alloc (1.1.0 -> 1.2.0)
- [x] duplexify (3.5.0 -> 3.6.0)
- [x] ecc-jsbn (0.1.1 -> 0.2.0)
- [x] peek-stream (1.1.2 -> 1.1.3)

When running `yarn build-bundle` then searching for all the occurrences of `new Buffer(`, there are now 14 search results:

- 2 are comments
- 7 are compatibility fallback
- 1 is from `buffer-fill`, but the code is not used by any of the `yarn` command
- 3 are from `asn1`
- 1 is from `http-signature`

To sum up, here are the PR we are waiting to merge to be done with this deprecation warning in yarn:

* [ ] asn1 (Refs: https://github.com/joyent/node-asn1/pull/30)
* [ ] http-signature (Refs: https://github.com/joyent/node-http-signature/pull/67)

**Test plan**

CI

PR #6274, opened 2018-08-16
Fixes the bundle build

arcanis:
**Summary**

#6208 caused some rx packages to be upgraded, which in turn broke the bundling process (the new release doesn't play well with webpack because of its amd definition). This diff updates the configuration to hack around this issue.

**Test plan**

```
yarn && yarn build-bundle && ./artifacts/yarn-1.10.0-0.js --version
```

### DIFF
--- a/scripts/build-webpack.js
+++ b/scripts/build-webpack.js
@@ -34,6 +34,10 @@ const compiler = webpack({
         exclude: /node_modules/,
         loader: 'babel-loader',
       },
+      {
+        test: /rx\.lite\.aggregates\.js/,
+        use: 'imports-loader?define=>false'
+      },
     ],
   },
   plugins: [
--- a/yarn.lock
+++ b/yarn.lock
@@ -3761,6 +3761,14 @@ import-local@^1.0.0:
     pkg-dir "^2.0.0"
     resolve-cwd "^2.0.0"
 
+imports-loader@^0.8.0:
+  version "0.8.0"
+  resolved "https://registry.yarnpkg.com/imports-loader/-/imports-loader-0.8.0.tgz#030ea51b8ca05977c40a3abfd9b4088fe0be9a69"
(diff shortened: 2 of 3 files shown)


yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0-beta.55"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.55.tgz#71f530e7b010af5eb7a7df7752f78921dd57e9ee"
-  integrity sha1-cfUw57AQr163p993UveJId1X6e4=
```
Gudahtt: Was this intended? It looks like all of the integrity hashes have been removed.

arcanis: Nope, not intended, my machine wasn't using the nightly yet 😞

scripts/build-webpack.js
```diff
@@ -34,6 +34,10 @@ const compiler = webpack({
         exclude: /node_modules/,
         loader: 'babel-loader',
       },
+      {
+        test: /rx\.lite\.aggregates\.js/,
+        use: 'imports-loader?define=>false'
```
Gudahtt: Is it valid to have a `use` entry here? It seems to work for me too, but... `module.loaders` is apparently an alias for `module.use`. `module.use` entries must have a `loader` field. The syntax you're using here is assumed to be in a `module.rules` entry, which is how Webpack wants us to configure loaders since v2. Yet this seems to work anyway...

We should probably update webpack :/  It's difficult to even find documentation for this version.

moertchen:
I have installed the .deb from this CircleCI build too, but I still get the same error:

[1/4] Resolving packages...
error An unexpected error occurred: "Cannot find module \"./rx.lite\"".

PR #6297, opened 2018-08-22
Update update-homebrew.sh

SimenB:
**Summary**
Went down a rabbit hole with the formula, and noticed that the script wasn't updated.

**Test plan**
Not sure. @Daniel15 maybe has some thoughts?


Daniel15:
Thanks! This looks good to me.

PR #6303, opened 2018-08-23
fix(build): Apply bundle fix to legacy build

### LINKED ISSUE #6269: error An unexpected error occurred: "Cannot find module \"./rx.lite\"".
A bug.

I'm unable to use yarn. Whenever i try it I get this: `error An unexpected error occurred: "Cannot find module \"./rx.lite\"".`

I'm trying to paste de error log somewhere to show it, but it's rather long and both pastebin and gist are not working. Here goes the error part of it. Please tell me if there is some other part of it that could be useful to trace the problem.
```
Trace: 
  Error: Cannot find module "./rx.lite"
      at webpackMissingModule (/usr/share/yarn/lib/cli.js:41505:83)
      at Object.<anonymous> (/usr/share/yarn/lib/cli.js:41505:169)
      at Object.<anonymous> (/usr/share/yarn/lib/cli.js:41514:3)
      at Object.module.exports.module.exports.read (/usr/share/yarn/lib/cli.js:42849:30)
      at __webpack_require__ (/usr/share/yarn/lib/cli.js:22:30)
      at Object.<anonymous> (/usr/share/yarn/lib/cli.js:106679:10)
      at __webpack_require__ (/usr/share/yarn/lib/cli.js:22:30)
      at Object.<anonymous> (/usr/share/yarn/lib/cli.js:50273:11)
      at __webpack_require__ (/usr/share/yarn/lib/cli.js:22:30)
      at Object.module.exports.Object.defineProperty.value (/usr/share/yarn/lib/cli.js:47242:16)
```

node: 10.8.0
yarn: v1.10.0-20180813.2207
linux ubuntu 18.04

Gudahtt:
**Summary**

The Webpack config for the modern build was [adjusted recently to add a loader](https://github.com/yarnpkg/yarn/pull/6274),
which is used as a workaround for a bug introduced by a [recent dependency
update](#6208 ). The legacy config needed the same update.

The Webpack config was also updated in superficial ways to follow
recommendations made for migrating from Webpack v1 to v2 (e.g. using
`module.rules` instead of `module.loaders`). These changes should have no
functional impact, and should make migrating to future versions of Webpack
easier.

fix #6269

**Test plan**

I have tested building and running both the modern and legacy builds. Before this change, the artifact from the modern build worked, but the one from the legacy build did not. Now they both work.

I also compared the stdout from running the `build-bundle` command, before and after this change (it prints the list of files included in the bundle). The output matched exactly.

### DIFF
--- a/scripts/build-webpack.js
+++ b/scripts/build-webpack.js
@@ -28,11 +28,11 @@ const compiler = webpack({
     'packages/lockfile/index.js': path.join(basedir, 'src/lockfile/index.js'),
   },
   module: {
-    loaders: [
+    rules: [
       {
         test: /\.js$/,
         exclude: /node_modules/,
-        loader: 'babel-loader',
+        use: 'babel-loader',
       },
       {
         test: /rx\.lite\.aggregates\.js/,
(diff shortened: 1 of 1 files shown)


arcanis:
👍

PR #6321, opened 2018-08-28
Reverts #2454

arcanis:
**Summary**

Cf my comment in #2454 - I'm not certain it fixes anything. I'm not certain there was a problem to begin with 🤔 

**Test plan**

I can make a test if I understand the repro. The existing ones are passing.

### DIFF
--- a/src/util/fs.js
+++ b/src/util/fs.js
@@ -696,7 +696,7 @@ export async function symlink(src: string, dest: string): Promise<void> {
     await fsSymlink(src, dest, 'junction');
   } else {
     // use relative paths otherwise which will be retained if the directory is moved
-    let relative = path.relative(path.dirname(dest), src);
+    const relative = path.relative(path.dirname(dest), src);
     // When path.relative returns an empty string for the current directory, we should instead use
     // '.', which is a valid fs.symlink target.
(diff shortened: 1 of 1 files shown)


richard-ejem:
@arcanis I just tried to update a project Yarn 1.9.4 -> Yarn 1.19.1 and it seems this revert broke how we work with links. We have a monorepo where the main package links modules from other folders using `yarn link`, like this:
```

~/my-project/library - a standalone yarn package @my-project/library
~/my-project/main - depends on `@my-project/library` and links it using `yarn link @my-project/library`
```

With Yarn 1.9.4 (before this revert), the link created in `~/my-project/main/node_modules/@my-project/library` led directly to `../../../my-project/library`.

Now, because the `fs.realpathSync` call was removed, the symlink leads to another symlink like this: `~/my-project/main/node_modules/@my-project/library` -(symlink)-> `../../../../../.config/yarn/link/@my-project/library` -(symlink)-> `../../../../my-project/library`

This is a problem because I build the app in Docker and the `.config/yarn` is not even a part of the Docker volume,  therefore IDE on my host machine sees the symlinks in `node_modules` as broken.

Would it be possible to add the capability to resolve the symlinks back?

PR #6327, opened 2018-08-29
provide default reporter in BaseFetcher

oprogramador:
**Summary**

This PR should resolve (log a meaningful error) that issue https://github.com/cloudfoundry/nodejs-buildpack/issues/158 .

Moreover when I run locally `yarn install`, I checked `reporter` is `undefined` in `~/.nvm/versions/node/v8.11.1/lib/node_modules/yarn/lib/cli.js`

### DIFF
--- a/src/fetchers/base-fetcher.js
+++ b/src/fetchers/base-fetcher.js
@@ -11,9 +11,25 @@ import * as fs from '../util/fs.js';
 
 const path = require('path');
 
+const defaultReporter = {
+  activity: x => x,
+  activitySet: x => x,
+  command: x => x,
+  error: x => x,
+  info: x => x,
+  lang: x => x,
+  progress: x => x,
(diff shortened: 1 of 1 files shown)


oprogramador:
@arcanis @Daniel15 

could you look at this?
some checks are failing in this PR but on `master` they're failing as well

Daniel15:
Hmm, I'm not sure why it would be `null`. In the constructor for `Config`, `reporter` is defined as being non-nullable, so it should always be set: https://github.com/yarnpkg/yarn/blob/master/src/config.js#L95

arcanis:
@oprogramador Can you add a test? I don't see how the BaseFetcher can be instanciated by itself.

oprogramador:
I close this as I don't need that any longer, and I don't have enough time to provide these tests.

PR #6353, opened 2018-09-03
fix(resolution): Normalise non-HTTPS NPM registry URLs too

### LINKED ISSUE #6259: reinstall of onetime package replaces https with http

Bug report

The package URL for onetime is changed to http://

```
onetime@^1.0.0:
  version "1.1.0"
  resolved "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz#a1f7838f8314c516f05ecefcbc4ccfe04b4ed789"

```

Have `onetime` in a project.  
Reinstall all deps (remove node_modules and yarn.lock).

Yarn should not change https to http.

node: 10.8.0
yarn: 1.9.4

edmorley:
**Summary**

For some packages the NPM registry is incorrectly returning tarball URLs that are not using HTTPS. For example:
`http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz`

Previously the `registry.npmjs.org` -> `registry.yarnpkg.com` tarball URL normalisation was not occurring for these non-HTTPS URLs, causing unwanted `http://registry.npmjs.org` references in `yarn.lock`.

Whilst the real fix needs to be made upstream:
https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

...this change prevents lockfile churn and insecure package downloads over HTTP in the meantime.

Fixes #6259.

**Test plan**

The request-cache entry for `onetime` has been updated to the current NPM registry response, which contains the bogus `tarball` URLs. Running the newly added test without the corrected regex, confirmed the test failed prior to the fix.

In addition, running `yarn add onetime@1.1.0` with yarn 1.9.4 gives a `yarn.lock` containing the `http://registry.npmjs.org/onetime/...` tarball URL, whereas doing the same with a build from this branch gives `https://registry.yarnpkg.com/onetime/...`.


arcanis:
Yeah, we have these errors sometimes. Your PR looks good, thanks! I think that it was supposed to work like that, seeing that we were using `http[s]:` ... curious to use a range for the `s` when there's a single character to match.

edmorley:
Yeah agreed - suspect it was the intention from the beginning. 
Thank for you for the review/merge :-)

PR #6370, opened 2018-09-07
Add spaces between numbers and units

AlexWayfer:
https://en.wikipedia.org/wiki/International_System_of_Units#General_rules


src/reporters/console/console-reporter.js
```diff
@@ -154,10 +154,10 @@ export default class ConsoleReporter extends BaseReporter {
     this.stopProgress();
 
     const totalTime = (this.getTotalTime() / 1000).toFixed(2);
-    let msg = `Done in ${totalTime}s.`;
+    let msg = `Done in ${totalTime} s.`;
     if (showPeakMemory) {
       const peakMemory = (this.peakMemory / 1024 / 1024).toFixed(2);
-      msg += ` Peak memory usage ${peakMemory}MB.`;
+      msg += ` Peak memory usage ${peakMemory} MB.`;
```
aduh95: Shouldn't that be `MiB` instead of `MB` if we want to comply to SI? (Refs: https://en.wikipedia.org/wiki/Mebibyte)

__tests__/reporters/__snapshots__/console-reporter.js.snap
```diff
@@ -24,14 +24,14 @@ Object {
...
-  "stdout": "[2K[1G✨  Done in 0.00s.",
+  "stdout": "[2K[1G✨  Done in 0.00 s.",
 }
 `;
 
 exports[`ConsoleReporter.footer 2`] = `
 Object {
   "stderr": "",
-  "stdout": "[2K[1G✨  Done in 0.00s. Peak memory usage 0.00MB.",
+  "stdout": "[2K[1G✨  Done in 0.00 s. Peak memory usage 0.00 MB.",
```
aduh95: Shouldn't that be MiB instead of MB if we want to comply to SI? (Refs: https://en.wikipedia.org/wiki/Mebibyte)

src/reporters/console/console-reporter.js
AlexWayfer: OK, done in the second commit, thanks.

__tests__/reporters/__snapshots__/console-reporter.js.snap
AlexWayfer: OK, done in the second commit, thanks.

BYK:
I'd say this is a breaking change since it changes the output.

PR #6393, opened 2018-09-15
Fetch tags for cached git repo

### LINKED ISSUE #6256: Yarn cache bug w/ Github URL packages
**Bug**

`yarn install` fails to install subsequent Github URL packages with format
`git+ssh://git@github.com:luwes/chameleon.git#cricket-v0.0.1-beta.3`
`git+ssh://git@github.com:luwes/chameleon.git#grasshopper-v0.0.1-beta.4`

```sh
error Command failed.
Exit code: 128
```

screen shot 2018-08-09 at 14 22 00

`yarn install` installs the package

**Node 10.6.0, Yarn 1.9.4, MacOS 10.13.3**

luwes:
**Summary**

Fixes #6256

This change fixes an issue where the cached git repo was not updated with the latest tags.
As a result yarn would exit with the `git archive` command -> `fatal: not a tree object`.

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -319,6 +319,7 @@ export default class Git implements GitRefResolvingInterface {
 
     return fs.lockQueue.push(gitUrl.repository, async () => {
       if (await fs.exists(cwd)) {
+        await spawnGit(['fetch', '--tags'], {cwd});
         await spawnGit(['pull'], {cwd});
       } else {
         await spawnGit(['clone', gitUrl.repository, cwd]);


ramasilveyra:
This is great, thanks! This error is annoying. I think that this issue was fixed in an old version of yarn and then it got broken again.

Thinking on how to add a e2e test for this. Maybe we should:

1. create a private GH repo on the yarn org with a git tag.
2. create a fake user GH user and give access to this private repo.
3. add the ssh key of this user to the CI.
4. and run this new test suite only with a env flag.

The new GH user is for security concerns since the yarn CI is public, the idea is to expose a SSH key with a granular access and from a non personal GitHub account.

Obviously we will need help from the yarn maintainers for this.

Or maybe there is better way?

luwes:
@ramasilveyra thanks for providing a potential test plan for this!

I think this could work. The only thing I am concerned about is that this issue only occurs on subsequent installs (like when upgrading the version of the git pkg). Yarn will initially do a git clone which will include all tags from the remote. Do you think it would be possible to publish a new version of the dummy pkg in the test? Another option would be to go in the cached repo and delete the dummy pkg and then run the test.

On a side note, it's possible to get around this issue by adding this global git config.

```bash
git config --global --add remote.origin.fetch "+refs/tags/*:refs/tags/*"
```

This will automatically fetch ALL tags from the remote when Yarn pulls in the latest changes in the cached repo. But at the same every `git pull` on your machine will pull in ALL git tags from the remotes.

luwes:
Does this fix make sense @yarnpkg/maintainers?

That `git pull` on the cached repo should result in the same repo if a `git clone` was done.
Currently this isn't the case, that added `git fetch --tags` will include all tags...

bestander:
I think it makes sense, thanks @luwes  for finding the root cause.
To keep things going let's merge it while PR is fresh.
e2e test could be hard to do but if you can come up with something we can run on CI it would be great

Issue #2430, opened 2017-01-11
`yarn publish [.tgz]` refuses to run without `--new-version` option


deecewan:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When running `yarn publish` on a pre-built tarball, yarn refuses to continue without the following error:

```
error Use the "--new-version [version]" flag to create a new version.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

- create a project with a package.json
- build a valid tarball of the project
- run `yarn publish <location-of-tarball>`

**What is the expected behavior?**

If the tarball provided is valid, it should already have a package.json inside it.  In my case, I am pre-building and compressing before I publish, so the package.json is already set to the correct version.  I would imagine that people build a tarball are not expecting the contents of that tarball to be modified during the publish stage, and I don't see a reason that the version must be bumped unless the tarball is being modified to reflect the new version.  I do not think that it should be enforced that a version needs to be bumped in this case.  It should only required bumping if the `--new-version` option is used.

**Please mention your node.js, yarn and operating system version.**

macOS 10.12.2, yarn 0.18.1, node 7.3.0

rabbitfr:
same here with a folder. yarn 0.18.1, 0.19, 0.19.1 osx 10.9.5 node 6.9.4

rabbitfr:
ok

You must respect http://semver.org/ for versioning. 0.1 won't work for example.

destroyerofbuilds:
Related to #722 and #2445

synthecypher:
Over a year later and this is still a problem. 😭

zukelwa20:
Over a year and six months later and this is still a problem. sob

Fer0x:
#6399 should fix it

PR #6399, opened 2018-09-18
fix(cli/publish): allow to publish directory without `--new-version`

Fer0x:
**Summary**

Fix https://github.com/yarnpkg/yarn/issues/2430

This PR fix `yarn publish [tarball]` and `yarn publish [folder]` commands: flag `--new-version` should not be required, like when using just `yarn publish`.

Current bugged behavior:
```sh
$ yarn publish build
yarn publish v1.9.4
[1/4] Bumping version...
error Use the "--new-version [version]" flag to create a new version.
info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

**Test plan**

```sh
yarn publish build
```

or

```sh
yarn publish some-package.tgz
```

### DIFF
--- a/__tests__/commands/publish.js
+++ b/__tests__/commands/publish.js
@@ -131,3 +131,16 @@ test.concurrent('can specify a path', () => {
     );
   });
 });
+
+test.concurrent('can specify a path without `--new-version`', () => {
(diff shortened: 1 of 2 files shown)


arcanis:
👍

PR #6383, opened 2018-09-13
Fail requests on status code >= 400

Ojisama:
Fix #6359 

**Summary**

Some commands (like `publish`) do not fail if the response status code is >= 400.

**Test plan**

**Before**
yarn publish

**After**
yarn-local publish

Issue #6352, opened 2018-09-03
Error Received malformed response from registry for undefined. For Public Registry


namaggarwal:
What is the current behavior?
`yarn add .` fails with `error Received malformed response from registry for undefined. The registry may be down.`

If the current behavior is a bug, please provide the steps to reproduce ?
```
yarn add .
```

What is the expected behavior?
yarn add should succeed

Please mention your node.js, yarn and operating system version.
Yarn Version - 1.9.4
MacOS Version - 10.13.6
Node Version - 8.11.2

Attached is the log
[error.log](https://github.com/yarnpkg/yarn/files/2346191/error.log)

namaggarwal:
Please note that `npm install .` works fine

Issue #6347, opened 2018-09-02
Registry errors causing CI failures


bvaughn:
Yesterday, Circle CI status page reported intermittent NPM registry issues– but today both [Circle CI status](https://status.circleci.com/) and [NPM status](https://status.npmjs.org/) pages are all green. Interesting, the [Yarn status page](https://status.yarnpkg.com/) is also all green, but for some reason I don't trust it very much 😄
<img width="951" alt="screen shot 2018-09-02 at 8 03 27 am" src="https://user-images.githubusercontent.com/29597/44957387-b6c4c680-ae86-11e8-8a3c-6d0f30c0df50.png">

Anyway, this has been causing consistent CI failures for the React team all weekend (e.g. [here](https://circleci.com/gh/facebook/react/11555#tests/containers/2)) so I wanted to make sure it was raised as an issue. I believe the problems started Saturday morning ~9am PDT if memory serves.

This issue is probably a duplicate of #6345 but I wanted to file the CI issue more explicitly.

merceyz:
https://github.com/yarnpkg/yarn/issues/8231#issuecomment-753661809

PR #6400, opened 2018-09-19
Check registry status code for 500 errors

jmorrell:
**Summary**

Over the last week I received several reports from Heroku users using Yarn to install dependency where the installation would fail. Generally the errors would be one of these two:

```
Error: Received malformed response from registry for undefined. The registry may be down.
```

```
Error: https://registry.yarnpkg.com/tmpl/-/tmpl-1.0.4.tgz: Integrity check failed for "tmpl" (computed integrity doesn't match our records, got "sha1-JRR4x9Dq/iPzyx1+yQY+TRa9x6c=")
```

The next build would generally succeed, but then a different module would fail an integrity check. I deployed a modified version of Yarn that would console log the response when one of these errors would occur and found that in these cases, CloudFlare was returning a 500 response with an HTML error message as the `body`. This is not common, but I was able to recreate it by curling a particular tarball in the registry in a loop.

Looking at [the malformed response code](https://github.com/yarnpkg/yarn/blob/master/src/resolvers/registries/npm-resolver.js#L42), it would receive the HTML error page, check for parameters, and then fail with the error above.

This PR adds an explicit check for the case that the returned status code is 500 or greater without relying on a JSON response with an `error` field.

**Test plan**

```
yarn test
```

### DIFF
--- a/__tests__/util/request-manager.js
+++ b/__tests__/util/request-manager.js
@@ -205,6 +205,32 @@ test('RequestManager.execute Request 403 error', async () => {
   });
 });
 
+// Cloudflare will occasionally return an html response with a 500 status code on some calls
+test('RequestManager.execute fails on 500 error', async () => {
+  jest.resetModules();
+  const config = await Config.create({}, new Reporter());
+  jest.mock('request', factory => options => {
+    options.callback(
+      '',
+      {statusCode: 500},
+      `<!DOCTYPE html><title>Rendering error | registry.yarnpkg.com | Cloudflare</title>...`,
+    );
+    return {
+      on: () => {},
+    };
+  });
+  await config.requestManager.execute({
+    params: {
+      url: `https://localhost:port/?nocache`,
+      headers: {Connection: 'close'},
+    },
+    resolve: body => {},
+    reject: err => {
+      expect(err.message).toBe('https://localhost:port/?nocache: Registry error');
+    },
+  });
+});
+
 test('RequestManager.request with offlineNoRequests', async () => {
   const config = await Config.create({offline: true}, new Reporter());
   try {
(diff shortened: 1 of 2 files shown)


jmorrell:
Just realized this PR is functionally similar to https://github.com/yarnpkg/yarn/pull/6383

jmorrell:
This likely resolves these issues:

https://github.com/yarnpkg/yarn/issues/6352
https://github.com/yarnpkg/yarn/issues/6347

Ojisama:
Hey @jmorrell ! I'm the author of #6383 and I was wondering why we should stop at `status >= 500`? :open_mouth: 

Surely a 4xx should make any command fail, right?

jmorrell:
@arcanis It looks like one of the CI tests failed with one of these exact errors? Maybe this doesn't fix the integrity issues 🤔 

```
    warning There appears to be trouble with our server. Retrying...
    error https://registry.yarnpkg.com/url-loader/-/url-loader-0.5.9.tgz: Integrity check failed for "url-loader" (computed integrity doesn't match our records, got "sha512-uGYPY/MiYqyLMx8iUF2wK9s73uh5JEMnoVz94bV9UhYtTA9Jz8WA8TpngOk0C51HcuyakwiA7vRc3AqUB6xleQ==")
```

jmorrell:
@Ojisama agreed, I'll see about adding it

arcanis:
So the thing is that [NPM apparently retries the requests on 408 and 5xx until they work](https://github.com/npm/npm-registry-client/blob/14efa6848ca1e7278de0eb12a1430cdb85bddf02/lib/request.js#L81). That's why they break with Yarn: we don't seem to correctly retry them in every case.

@jmorrell Do you think it would be possible to update your PR to ensure we retry as well on these errors? I'm not fan of retrying, but since it breaks so many installs... I wonder what's the reason behind the server errors, though 😞

arcanis:
I refactored part of the code to support this retry logic - please check #6413 if you have some time. I'm going to close this issue and the other to keep discussions on a single thread 👍

Issue #5808, opened 2018-05-11
[feat] yarn audit


rally25rs:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

npm added `audit` to warn about packages with known security issues. There was some conversation about this previously and one of the core npm folks said the API was likely to be open/public to pull this info. Therefore, yarn should be able to add this feature.

**What is the expected behavior?**

* Add a `yarn audit` command that mimics `npm audit`
* Add warnings when adding/installing packages with known issues.

**Please mention your node.js, yarn and operating system version.**

This would be a minor version bump, so likely target yarn v1.7.0 or v1.8.0 depending on timing.
This is probably too important to wait for v2.0.

batjko:
One thing that yarn could do, is improve on the concept. `npm audit`'s output and configurability is somewhat basic right now. 

Perhaps yarn could add some interactivity, allowing the user to automatically update packages, as long as the suggested version is just a semver patch update.

Also definitely useful would be different types of output (html, json, csv, or similar console outputs to test or lint runners etc.)

I'm sure there are more ways yarn can add a bunch of value here.

chapati23:
Would _love_ yarn to add this. We're currently using [snyk.io](https://snyk.io) for vulnerability management but I've never been 100% happy with it as it always felt like something that should be more integrated.

Was super excited when I heard that `npm` added the `audit` command. But also super disappointed when I realized that it only works with a corresponding `package-lock.json`, which would basically mean for us to either drop `yarn` completely and go back to `npm`. Or use both tools, which already sounds like a recipe for horrible merge-conflicts and syncing-hell between the two lock-files.

Hypnosphi:
> horrible merge-conflicts

Right now both yarn and npm are able to autoresolve the conflicts

> syncing-hell

Yeah, that's the real problem

shinriyo:
`yarn` is upward compatibility to `npm`.
I hope yarn should support `audit` option officially!

Hypnosphi:
Unfortunately, any new command addition is a breaking change for yarn, because of `yarn scriptName` shortcut

batjko:
> Unfortunately, any new command addition is a breaking change for yarn, because of yarn scriptName shortcut

I don't see how. 
`npm start`, for example, has a default behavior which kicks in no explicit script for it is defined in package.json.
So if someone has `yarn audit` already defined, it will simply override the default which would be npm's audit. E.g.: 
```json
"scripts": {
[...]

PR #6409, opened 2018-09-20
feat(audit) Initial addition of yarn audit command.

rally25rs:
**Summary**

Adds the `yarn audit` command.

Adds the `--audit` flag to `yarn add / install / upgrade`

Currently the functionality is fairly basic. It roughly copies the initial audit release of npm including the style of reporting.

The `audit` command will return 0 if no issues are found, otherwise returns non-0 exit code.

Unlike npm, yarn doesn't currently auto-run audit on add/install/upgrade.  Partly this is due to the additional network call and a lack of an easy way to mock it during unit tests, so every unit test tries to hit npm's API to audit the packages being "installed" during unit tests.
To run an audit during add/install/upgrade, please specify the `--audit` flag.

If the `--offline` flag is specified along with `--audit` then audit will be skipped (and a message noting this will be printed)

If you would like audit to always be run, you can add to the `.yarnrc` file:
```
--add.audit true
--install.audit true
--upgrade.audit true
```

implements #5808 

**Test plan**

Added some basic unit tests.

When reporting audit issues, please run with `--verbose` and include the "Audit API Request" and "Audit API Response" JSON.


arcanis:
Hey Jeff! Thanks for your work on this! A few comments:

- The `--audit` flag isn't working, it should be listed in `src/cli/index.js`
- Running an install with `--audit` is subject to the integrity check; we shouldn't bailout in this case
- The vulnerable package names are not shown, making it hard to know what should be done

I'm also not convinced by the workflow - right now you must add `--audit` when doing an install, meaning that if something is vulnerable you'll only know it after it's been installed (and its postinstall scripts have run). That doesn't look great from a security perspective. What would you think of this?:

- run `--audit` by default on `yarn add` and `yarn upgrade`, possibly prompting the user to confirm
- don't run `--audit` by default on `yarn install` (like it currently is, no change there)
- add a `yarn audit` command that simply report the current vulnerabilities (exit code 1 if vulnerable)
    - `yarn audit` wouldn't do a full install, and would use the versions currently in the lockfile
    - I guess doing the resolution if missing would also be ok; basically, skip fetch/link/build

What do you think?

arcanis:
Wait, sorry, missed that the `yarn audit` command already exists with your PR; so basically the only change I'd suggest would be to run it by default on `yarn add` (and filter out the existing errors), but that's your call 😃 

I'd also maybe suggest to make `--audit` on `install` prompt the user at install-time when they would install a vulnerable packages (before they're actually installed).

src/cli/commands/install.js
```diff
@@ -628,12 +668,24 @@ export class Install {
     for (const step of steps) {
       const stepResult = await step(++currentStep, steps.length);
       if (stepResult && stepResult.bailout) {
+        if (this.flags.audit) {
```
arcanis: Not sure this works - when running `install --audit` twice, the second time I get no reports

arcanis: I think it's because the bailout is done at the end of the resolution step, before the audit gets initialized

rally25rs: Yeah I'm pretty sure yarn just bails before it gets that far, like it would have during a normal `install`.

rally25rs:
@arcanis 

> I'd suggest would be to run it by default on yarn add

that was my original intent, like npm does, but it breaks all the existing unit tests because it has to make a network call. if you can think of a way to mock out that request by default for all tests, that would be great. Also I think in Discord @BYK said he was a bit concerned about adding another network call.

> I'd also maybe suggest to make --audit on install prompt the user at install-time when they would install a vulnerable packages (before they're actually installed).

I think I left a comment in the code where I could bailout if there are errors, but npm doesn't by default. Prompting the user gets tricky because what if this is in a CI environment or --non-interactive is passed.

> The vulnerable package names are not shown, making it hard to know what should be done

Is this just on a `yarn install --audit`? I believe npm does the same; just prints the summary and tells the user to "run yarn audit for more info".

-----

In addition I noticed a couple other ... oddities.

This command uses npm's api and half the code is a copy/paste from them, but yarn audit reports more errors than npm audit does. In the case I saw, npm wasn't reporting errors that were from a transitive dep of a github repo dependency. When the npm audit API sends back a "resolution recommendation" it wants tot ell the user to run `yarn upgrade foo@123` to resolve the vulnerability, but `foo` is a transitive dep, so that won't fix it. 😕 

example:

```
yarn audit v1.11.0-0
# Run yarn upgrade url-parse@1.4.3 to resolve 2 vulnerabilities
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Open Redirect                                                │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ url-parse                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=1.4.3                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ cxcore-ui-build-tasks                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ cxcore-ui-build-tasks > webpack-dev-server > sockjs-client > │
│               │ eventsource > original > url-parse                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/678                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

where `cxcore-ui-build-tasks` is a github url dependency. `npm audit` doesn't report that vulnerability at all, but the api returns the recommendation:

```
{
  "actions": [
    {
      "action": "update",
      "module": "url-parse",
      "depth": 6,
      "target": "1.4.3",
      "resolves": [
        {
          "id": 678,
          "path": "cxcore-ui-build-tasks>webpack-dev-server>sockjs-client>eventsource>original>url-parse",
          "dev": false,
          "optional": false,
          "bundled": false
        },
        {
          "id": 678,
          "path": "cxcore-ui-build-tasks>webpack-dev-server>sockjs-client>url-parse",
          "dev": false,
          "optional": false,
          "bundled": false
        }
      ]
    },
```

but `update url-parse@1.4.3` isn't going to do anything.
So I think I might be constructing the JSON request / dep tree incorrectly. I might be reporting these packages by their hoisted location so npm thinks `url-parse` is at the top of the hierarchy.
Not sure yet...

src/cli/commands/install.js
arcanis: I think it should be safe to disable the integrity check altogether when `--audit` is set.

rally25rs:
OK I think this should be in a good place for an initial release. It's not feature-rich, but it does report audit problems correctly at this point.

arcanis:
Looks good to me! I'll merge the changelog, update this PR, and merge it 👍

Gudahtt:
I believe this PR is the reason the nightly tests on Ubuntu 16.0.4 have been failing.

The dependency `cli-table3` doesn't support Node v4. The line the test is failing on is [this one](https://github.com/cli-table/cli-table3/blob/d22503b51d7fc7c7d8859b4f7b879c2306ab56a0/src/utils.js#L9), due to the use of `let`.

The reason the other end-to-end tests aren't affected is that they use Node v6.

I don't yet know why the other tests aren't failing on Node v4. Perhaps this library is never invoked outside of the end-to-end tests.

PR #6413, opened 2018-09-21
Adds retry-on-error for errors 408/500

arcanis:
**Summary**

Supersedes #6400, #6383

This PR adds a retry mechanism for when the npm registry returns 408 / 500 errors.

**Test plan**

Added a test. Hopefully our own CI will become more stable as well.

### DIFF
--- a/src/util/request-manager.js
+++ b/src/util/request-manager.js
@@ -1,13 +1,14 @@
 /* @flow */
 
 import fs from 'fs';
+import http from 'http';
 import url from 'url';
 import dnscache from 'dnscache';
 import invariant from 'invariant';
 import RequestCaptureHar from 'request-capture-har';
 
 import type {Reporter} from '../reporters/index.js';
-import {MessageError} from '../errors.js';
+import {MessageError, ResponseError} from '../errors.js';
 import BlockingQueue from './blocking-queue.js';
 import * as constants from '../constants.js';
 import * as network from './network.js';
@@ -67,6 +68,7 @@ type RequestParams<T> = {
 };
 
 type RequestOptions = {
+  retryReason: ?string,
   params: RequestParams<Object>,
   resolve: (body: any) => void,
   reject: (err: any) => void,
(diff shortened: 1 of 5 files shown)


src/util/request-manager.js
```diff
@@ -389,15 +417,25 @@ export default class RequestManager {
...
+          }
+        }
+
         if (body && typeof body.error === 'string') {
           reject(new Error(body.error));
           return;
         }
 
-        if (res.statusCode === 403) {
+        if ([403, 408].indexOf() !== -1 || res.statusCode >= 500) {
```
Ojisama: Isn't `res.statusCode` mising in the `indexOf()`?

arcanis: Yep, it totally does ... Flow should have warned me 😞

arcanis:
No worry! I updated the code to take all `>=400` into account.

whereisaaron:
Hi @arcanis, this retry mechanism is very welcome thank you! However the retry message is I think a little misleading and IMO very annoying 😄 

Given this mechanism is mostly to address congenital unreliability of the npm registries, is it fair that the message suggests it is the user's local network that is the problem? 
```
offlineRetrying: 'There appears to be trouble with your network connection. Retrying...' 
```

Users are like, 'what do you mean my network has a problem, I have a gazillion-bit connection with zero drops!'. And they are looking for faults and problems in their network and for bugs in yarn, when we know neither is likely the cause. Could you consider revising or crafting a more appropriate message for this type of retry?

Also, given this retry requirement is congenital (I get multiple of this message with every `yarn install` now), it is basically normal operation. Do do users even need to hear about these retries unless the retires aren't working?

arcanis:
Hey @whereisaaron - the error message should be more explicit than what you quote. Cf here:

https://github.com/yarnpkg/yarn/pull/6413/files#diff-49d57266c25cd3c6ddab1f59de18b36bR24

Note that this PR is only part of the 1.11 and 1.12 branches, which are yet in prerelease.

whereisaaron:
Thanks @arcanis that message does look much better - I had thought we were seeing the message from this line of your patch - which was the one I was quoting.
https://github.com/yarnpkg/yarn/pull/6413/files#diff-5d246f90b159bd84bbc99f67527b1a0eR401

Great - so hopefully when we get 1.11 we'll see the current messages replaced with your new, more accurate message. 🎉

PR #6419, opened 2018-09-23
Fix missing intregity field for validate-npm-package-license in yarn.lock

johanhenrikssn:
**Summary**

Running ´yarn install` in repo gives this diff because `validate-npm-package-license` is missing the integrity field.


bestander:
Thanksm @johanhenrikssn

PR #6420, opened 2018-09-23
Upgrade to babel 7

johanhenrikssn:
**Summary**

Update assisted by `babel-upgrade` tool. This also moves away from using the [deprecated stage-0 preset](https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets) and instead explicitly add each propsal as a dependency.

**Test plan**

No new tests added.


package.json
```diff
@@ -53,16 +53,32 @@
...
+    "@babel/plugin-proposal-optional-chaining": "^7.0.0",
+    "@babel/plugin-proposal-pipeline-operator": "^7.0.0",
+    "@babel/plugin-proposal-throw-expressions": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-syntax-import-meta": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^7.2.3",
```
milesj: This should probably be `^9.0.0`.

package.json
```diff
@@ -53,16 +53,32 @@
...
-    "babel-loader": "^6.2.5",
+    "babel-loader": "^8.0.0",
     "babel-plugin-array-includes": "^2.0.3",
     "babel-plugin-transform-builtin-extend": "^1.1.2",
     "babel-plugin-transform-inline-imports-commonjs": "^1.0.0",
```
milesj: This isn't compatible with v7 and can be replaced with `['@babel/plugin-transform-modules-commonjs', {allowTopLevelThis: true}],`.

package.json
```diff
@@ -53,16 +53,32 @@
...
-    "babel-plugin-transform-runtime": "^6.4.3",
-    "babel-preset-env": "^1.6.0",
-    "babel-preset-flow": "^6.23.0",
-    "babel-preset-stage-0": "^6.0.0",
     "babylon": "^6.5.0",
```
milesj: This is now `@babel/parser`.

package.json
```diff
@@ -53,16 +53,32 @@
...
+    "@babel/plugin-proposal-numeric-separator": "^7.0.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.0.0",
+    "@babel/plugin-proposal-pipeline-operator": "^7.0.0",
+    "@babel/plugin-proposal-throw-expressions": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-syntax-import-meta": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
milesj: It's best to also place this in Yarn `resolutions`.

arcanis: Why is the bridge needed?

arcanis:
Thanks for taking this! Can you rebase on master? I also noticed that some tests are failing - seems like modules are not properly transpiled

package.json
milesj: Because jest requires `babel-core` and not `@babel/core`, so it needs to be bridged.

Issue #6148, opened 2018-07-24
workspaces: "yarn run" inside a package should run binaries from root node_modules


AviVahl:
**Do you want to request a *feature* or report a *bug*?**
feature request for better usability of workspaces

**What is the current behavior?**
`yarn run` only runs bins that were specified in the package itself, not ones from workspace's root `package.json`.

**If the current behavior is a bug, please provide the steps to reproduce.**

```
git clone https://github.com/wixplosives/sample-monorepo
cd sample-monorepo
yarn
# next one works!
yarn run webpack --version
cd ./packages/app
# next one fails :(
yarn run webpack --version
```

**What is the expected behavior?**
It should also find binaries from root `node_modules`, and run them with cwd set to the current package I'm in.

**Please mention your node.js, yarn and operating system version.**
node 8.11.3, yarn 1.9.1, fedora 28

pronebird:
That actually should work because I believe yarn creates `node_modules/.bin/` under each `package`. However, you should be able to achieve the same by going through `yarn workspace {package-name} run webkit --version`

AviVahl:
yarn only creates the `.bin` folder if those dev deps also appear in the internal `package.json`, which is not the case here, as I wanted to manage the version of webpack from a single place in the monorepo.

I'll check out the workspace command... thx for the tip @pronebird

alexindigo:
Same here. We have `jest` as shared dev dependency, and when we run `yarn workspaces run test`, we get:

```
yarn workspaces run test
yarn workspaces v1.10.1
yarn run v1.10.1
$ jest --collectCoverageFrom=src/**/*.js src
/bin/sh: jest: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 1
```

In order for it to work we have to run:

```
$ PATH=${PATH}:${PWD}/node_modules/.bin yarn workspaces run test
```

Not sure why yarn doesn't do it by default.

KaelWD:
Scripts can run binaries from the root, but `yarn run` can't. 

In a workspace with typescript in the root package:
```js
  "scripts": {
    "build": "tsc"
  }
```

`yarn build` works but `yarn run tsc` doesn't.

AviVahl:
appears fixed in `yarn@1.15.2` :D

uncleramsay:
This doesn't appear to be fixed for me?

package.json (this is inside a monorepo)
```
[...]

PR #6444, opened 2018-09-27
Fixes yarn run when used on workspaces + pnp

arcanis:
**Summary**

This diff ensures that we are able to run scripts from `yarn run` even when within a workspace. We just find the locator that owns the workspace we're into, and use that to detect the dependencies.

**Test plan**

Added tests.


AviVahl:
@arcanis does this PR also closes https://github.com/yarnpkg/yarn/issues/6148?
or is root `.bin` not included yet?

PR #6447, opened 2018-09-27
Support Windows paths for pnp.

### LINKED ISSUE #6402: Plug'n'Play doesn't support Windows paths
I want to report a bug.

When using `yarn install` with `C:\Users\Username\Desktop\yarn\bin` added to path variable, where `C:\Users\Username\Desktop\yarn` is checked out with `git clone github.com/yarnpkg/yarn.git --branch yarn-pnp` and `"installConfig": {"pnp": true}` is appended to project's `package.json` or `yarn install --pnp` is executed in project's folder it creates `node_modules` folder and doesnt create `.pnp.js` file. 

These methods were tested on [pnp-sample-app](https://github.com/yarnpkg/pnp-sample-app) as well. And all required packages were present in yarn cache and checked with `yarn cache list`.

`yarn install --pnp` 
or
Add `"installConfig": {"pnp": true}` in `package.json` and run `yarn` or `yarn install` in project's root directory.

No `node_modules` folder should be created and `.pnp.js` shold be created inside the root of the project.

`Node v8.11.3`
`Yarn 1.11.0-0`
`Windows 10 1803`

jdalton:
This PR is a start to resolve #6402 _(Plug'n'Play doesn't support Windows paths)_.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -395,6 +395,19 @@ export default class Config {
       this.plugnplayEnabled = false;
     }
 
+    if (process.platform === 'win32') {	
+      const cacheRootFolderDrive = path.parse(this._cacheRootFolder).root;
+      const lockfileFolderDrive = path.parse(this.lockfileFolder).root;
+
+      if (cacheRootFolderDrive !== lockfileFolderDrive) {
+        if (this.plugnplayEnabled) {
+          this.reporter.warn(this.reporter.lang('plugnplayWindowsSupport'));
+        }
+        this.plugnplayEnabled = false;
+        this.plugnplayPersist = false;
+      }
+    }
+
     this.plugnplayShebang = String(this.getOption('plugnplay-shebang') || '') || '/usr/bin/env node';
     this.plugnplayBlacklist = String(this.getOption('plugnplay-blacklist') || '') || null;
 
--- a/src/config.js
+++ b/src/config.js
@@ -395,7 +395,7 @@ export default class Config {
       this.plugnplayEnabled = false;
     }
 
-    if (process.platform === 'win32') {	
+    if (process.platform === 'win32') {
       const cacheRootFolderDrive = path.parse(this._cacheRootFolder).root;
       const lockfileFolderDrive = path.parse(this.lockfileFolder).root;
 
--- a/src/config.js
+++ b/src/config.js
@@ -395,14 +395,6 @@ export default class Config {
       this.plugnplayEnabled = false;
     }
 
-    if (process.platform === 'win32') {
-      if (this.plugnplayEnabled) {
-        this.reporter.warn(this.reporter.lang('plugnplayWindowsSupport'));
-      }
-      this.plugnplayEnabled = false;
-      this.plugnplayPersist = false;
-    }
-
     this.plugnplayShebang = String(this.getOption('plugnplay-shebang') || '') || '/usr/bin/env node';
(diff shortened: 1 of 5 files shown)


src/config.js
```diff
@@ -395,14 +395,6 @@ export default class Config {
       this.plugnplayEnabled = false;
     }
 
-    if (process.platform === 'win32') {
```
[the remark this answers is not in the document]
jdalton: Ah okay, I'll noodle on that today.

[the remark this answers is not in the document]
jdalton: @arcanis Thanks for pushing more patches to the PR!

src/config.js
```diff
@@ -396,11 +396,16 @@ export default class Config {
...
     if (process.platform === 'win32') {
-      if (this.plugnplayEnabled) {
-        this.reporter.warn(this.reporter.lang('plugnplayWindowsSupport'));
+      const cacheRootFolderDrive = path.parse(this._cacheRootFolder).root;
+      const lockfileFolderDrive = path.parse(this.lockfileFolder).root;
+
+      if (cacheRootFolderDrive !== lockfileFolderDrive) {
+        if (this.plugnplayEnabled) {
+          this.reporter.warn(this.reporter.lang('plugnplayWindowsSupport'));
+        }
```
jdalton: Now may be a good time to think on other platform specific path things that might make relative paths less portable from platform to platform. Off the top of my head there is also long paths _([namespaced paths](https://nodejs.org/api/path.html#path_path_tonamespacedpath_path))_ in Windows.

arcanis: I guess if the need arise we could just have a `plugnplay-use-absolute` settings that would use absolute paths for everything outside of `__dirname`. Since `getPackageInformation` already returns absolute paths no matter what, it should be backward compatible.

arcanis:
Next step: port the `pkg-tests` to be executed on Windows. Maybe we could try the [Azure pipeline](https://azure.microsoft.com/en-us/services/devops/pipelines/)? We were interested to give it a try, it would be a good test subject since the `pkg-tests` testsuite is quite well contained.

jdalton:
I'll ping some Azure dev advocates 🥑 to see if they have any guidance, tips, or can pitch in.

btholt:
@arcanis I'm happy to be your point-of-contact for Azure Pipelines and pitch in where helpful.

edmorley:
I ran into an issue with yarn run - opened #6493.

arcanis:
Thanks @btholt! I just opened #6495 to play a bit with the pipeline 😃

src/config.js
```diff
@@ -396,11 +396,16 @@ export default class Config {
     }
 
     if (process.platform === 'win32') {
-      if (this.plugnplayEnabled) {
-        this.reporter.warn(this.reporter.lang('plugnplayWindowsSupport'));
+      const cacheRootFolderDrive = path.parse(this._cacheRootFolder).root;
+      const lockfileFolderDrive = path.parse(this.lockfileFolder).root;
+
+      if (cacheRootFolderDrive !== lockfileFolderDrive) {
```
vkrol: There should be case-insensitive check.

arcanis: Makes sense! Can you open a PR?

jdalton: It looks like both `cacheRootFolder` and `lockfileFolder` can be user customized. Would it make more sense for user-provided values to be normalized at time of ingestion to avoid these kinds of one-off normalizations in other places in the code as well?

PR #6460, opened 2018-10-01
Changelog workflow 

arcanis:
**Summary**

Our changelogs are not awesome. Let's make them awesome.

Two followups:

- CI on PRs to prevent merging without a changelog update (@BYK?)
- Auto-update the changelog.md file when tagging a release (@Daniel15)

**Test plan**

n/a

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,6 +6,8 @@
 
 <!-- Explain the **motivation** for making this change. What existing problem does the pull request solve? -->
 
+<!-- Don't forget to update the CHANGELOG.md to quickly describe your changes to other users! -->
+
 **Test plan**
 
 <!-- Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI. -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,14 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
-- #6449 Adds a special logic to PnP for ESLint compatibility (temporary, until https://github.com/eslint/eslint/issues/10125 is fixed)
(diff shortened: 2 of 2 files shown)


Daniel15:
> Auto-update the changelog.md file when tagging a release

Hmm, how would we do this? As part of the release tagging script?

arcanis:
Yup (to be clear I mean simply changing the "master" header into the tagged version, and add a new empty "master" at the top of the file)

Gudahtt:
Have we considered using using [conventional changelog](https://github.com/conventional-changelog/conventional-changelog) instead?

We already have commitizen setup, though it isn't used consistently. If we were to enforce the use of conventional commit messages, then we could auto-generate a changelog at release-time.

arcanis:
You should speak to @BYK about that, it's something he's been advocating at lunch 😛 

As for me I don't see much value in formalized commit messages (mainly because of my own experience where I find myself looking more at the description of the PR than the actual commit message, regardless of the project I inspect) or automated changelogs (because I think there's a slight delta between what you'd want to write in a commit message and in a changelog line), but if you all feel strongly about it I won't push too hard against either.

renchap:
For me, it is quite important to have the category in the Changelog messages (bugfix, doc, minor, major, breaking, ...) as it allows you to quickly see what are the important changes and split them by category in the Changelog (which is very useful).

I do not know if you rebase all PRs into a single commit, if this is not the case a changelog entry can be from multiple commits, so it might not be very practical to force the format of each commit ,as long as the Changelog entry is correct.

Gudahtt:
The main reason I have for suggesting it is to make it easier for contributors. It's easier to ask for a single description of the change - the commit message - than it is to ask for both a decent commit message and a changelog entry.

It's easier for us too, as the tools to enforce and make use of conventional commits already exist and are well maintained (e.g. `commitizen`'s interactive prompt, `conventional-changelog`'s changelog generation, `commitlint` for linting and possibly a pre-commit hook)

I agree that the ideal commit message is different than the ideal changelog entry. Conventional commits aren't really the best for either purpose in my opinion. But, given the other advantages, they might still be worthwhile. At least they only dictate the first line of the commit message (and the beginning of the body for a `BREAKING CHANGE`).

PR #6465, opened 2018-10-02
Disables by default the migrations

arcanis:
**Summary**

While not breaking technically speaking, the automatic migration is causing unexpected behaviors to our users. Such a behavior should only happen when bumping the major version.

To prevent that, the `unsafe-disable-integrity-migration` settings will default to `true` for the duration of the 1.X branch. **It will default to `false` again starting with the 2.0.**

In order to migrate your project now, just add the following line into your project-local yarnrc:

```
unsafe-disable-integrity-migration "false"
```

Note that this only affect the lockfile entries already present in your lockfile. Yarn will continue to create the new entries using the `integrity` field. My guess is that by the time the 2.0 lands, enough projects will have been migrated to use the `integrity` field that this will be a painless transition.

**Test plan**

CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6449](https://github.com/yarnpkg/yarn/pull/6449) - [**Maël Nison**](https://twitter.com/arcanis)
 
+- Disables by default (configurable) the automatic migration of the `integrity` field. **It will be re-enabled in 2.0.**
+
+  [#6465](https://github.com/yarnpkg/yarn/pull/6465) - [**Maël Nison**](https://twitter.com/arcanis)
+
 - Fixes the display name of the faulty package when the NPM registry returns corrupted data
 
   [#6455](https://github.com/yarnpkg/yarn/pull/6455) - [**Grey Baker**](https://github.com/greysteil)
--- a/__tests__/commands/add.js
+++ b/__tests__/commands/add.js
@@ -65,10 +65,10 @@ test.concurrent('adding to the workspace root should preserve workspace packages
     expect(pkg.dependencies).toEqual({'left-pad': '1.1.3', 'max-safe-integer': '1.0.0'});
 
     const lockfile = explodeLockfile(await fs.readFile(path.join(config.cwd, 'yarn.lock')));
-    expect(lockfile).toHaveLength(20);
+    expect(lockfile).toHaveLength(16);
     expect(lockfile.indexOf('isarray@2.0.1:')).toEqual(0);
-    expect(lockfile.indexOf('left-pad@1.1.3:')).toEqual(4);
-    expect(lockfile.indexOf('max-safe-integer@1.0.0:')).toEqual(8);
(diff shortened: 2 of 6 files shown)


arcanis:
> What do you think about adding a warning message that would explain this to the user (or link to a help page with more details)?

Not sure - should we print the warning every time? It would be very verbose (especially if it includes a link). It sounds like the kind of thing that users will just ignore since "it works".

Gudahtt:
I would suggest that the option could be renamed to avoid the double-negative, but then the `unsafe-` prefix ceases to make sense. We don't want to imply that migrations are unsafe.

I don't think printing a warning each time is warranted - there is nothing actionable here for the user to do.

arcanis:
Yep, the double negation was because I preferred to make it explicit that we recommend this option not to be turned on (a bit like [`DO_NOT_USE_OR_YOU_WILL_BE_FIRED`](https://github.com/facebook/react/blob/80bff5397bf854750dbe7c286f61654ea58938c5/src/umd/ReactUMDEntry.js#L21-L22)).

imsnif:
A message each time might indeed be excessive. Maybe some documentation somewhere like here then? https://yarnpkg.com/lang/en/docs/yarnrc/

arcanis:
@imsnif Do you have an idea why the bailout tests are failing?

imsnif:
@arcanis - my guess would be something to do with https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L465
But I'll have to take a closer look. Can do later tonight if it can wait.

arcanis:
> Making a third, hybrid format lockfile that has only some integrity entries is no use to 1.10 users because it doesn't really check the integrity

Yarn will check the integrity. It just will keep using the sha1 for old ones. The main reason we've disabled the migration by default is that the lockfile should not change when you run `yarn install` on an up-to-date project, except between major releases.

borekb:
A bit OT but since we're fighting this integrity field in our team where not everyone is on the same Yarn version (I know...), is there a way to pass `unsafe-disable-integrity-migration` from the command-line?

rjmunro:
@arcanis I'm seeing random integrity entires being added, depending on who / where yarn is being run, then being removed again when someone on an old version runs it. These all go into version control. It's a total mess. There should only ever be all or no entries. Which state a particular lock file is in should be left alone unless the user changes it (e.g. by running something like a special command, maybe `yarn upgrade-lockfile`).

arcanis:
@borekb @rjmunro this doesn't entirely answer your question, but are you aware that you can very easily force all of your team to use the exact version of Yarn you want them to use?

Just create a `.yarnrc` in your repositories, add `yarn-path "./scripts/yarn-1.12.1.js"` inside, add the [single-file release](https://github.com/yarnpkg/yarn/releases/tag/v1.12.1) to the right place, commit all this, and voilà. As long as your developers have Yarn 1.0 globally installed (which is more than one year old), Yarn will transparently use the version you've selected over the global one 🙂 

We use this pattern at work and it's extremely handy to force everyone to upgrade at once, or to rollback everyone if we find a regression.

borekb:
@arcanis That is a clever trick, thank you!

PR #6454, opened 2018-09-30
Default latest version in case it is not sent by the npm registry

mad-mike:
**Summary**

Like others before me (#5955), I have the problem that yarn outdated and yarn upgrade-interactive fail with the following error message:

```
> yarn outdated

yarn outdated v1.10.1
error An unexpected error occurred: "Invalid Version: undefined".
info If you think this is a bug, please open a bug report with the information provided in "yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/outdated for documentation about this command.

  TypeError: Invalid Version: undefined
      at new SemVer (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2549:11)
      at compare (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2836:39)
      at Function.lt (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:2870:10)
      at isDepOld (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:34115:134)
      at Array.filter (<anonymous>)
      at Function.<anonymous> (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:34120:19)
      at Generator.next (<anonymous>)
      at step (C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:98:30)
      at C:\Users\***\AppData\Roaming\npm\node_modules\yarn\lib\cli.js:109:13
      at <anonymous>
```

The previous pull request (#6017) appears to have stalled, so I'm trying a different approach.

I'm proposing to assume that the package is up to date if there is no information about a later package in that npm repository.
This is assuming that there is only a temporary issue with the metadata of that package, which will be fixed by the registry at a later point.
Meanwhile it is better to ignore this package than to crash.

**Test plan**

I tested the change locally by modifying the yarn code and running yarn outdated / yarn upgrade-interactive on a package that depends on a package for which the registry has no `latest` tag. That registry is a private one at my workplace.
Additionally, I wrote a unit test based on the package metadata in my scenario.

PR #6467, opened 2018-10-02
Disable semantic commit prefixes for Renovate

rarkins:
**Summary**

As suggested by @edmorley here: https://github.com/yarnpkg/yarn/pull/6463#issuecomment-426240070

This will prevent Renovate from autodetecting semantic commits have been added to master and switching them on automatically, and instead just keep them permanently off.

**Test plan**

No tests required or changelog update as this is bot config internal to this repo only.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,7 @@
   "extends": [
     "config:base",
     ":preserveSemverRanges",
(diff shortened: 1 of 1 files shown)


edmorley:
The CI failures appear on `master` too, and are from #6454.

PR #6475, opened 2018-10-02
Fix Chocolatey log spam & bad nodejs-lts requirement

### LINKED ISSUE #6439: Bad chocolatey package dependency
Well - there's been this PR https://github.com/yarnpkg/yarn/pull/5925 - which, uh, I don't like, to say the least.

It's simply a bad idea to change package behaviour that drastically. Because "some people request it". So choco yarn would like to downgrade node 10 to node 6.9.2? No good - even if it would've been node 8 (which is the correct lts).

If you can't depend on nodejs.install OR nodejs-lts then why depend on it at all? yarn*msi installs fine without node. It should be completely up to the user which nodejs they install and how they do it (node.exe, chocolatey, nvm, a.s.o.)!

JonathanWolfe:
**Summary**

Fixes #6439 

The chocolately install has always been full of log spam due to an incorrectly written `/l*v` flag, but those logs are also unneeded since the output has no useful info so I just removed them (this can be reversed to just fix the flag if y'all really want it back).

This also removes the bad `nodejs-lts` nuget dependency added in #5925. It now has no dependency, and instead logs out to inform the user to install node if it's missing.

**Test plan**

* `choco uninstall yarn`
* Replace the `{VERSION}` and `{CHECKSUM}` placeholders manually or do a full build (I manually replaced)
* `.\chocolateyinstall.ps1`

### DIFF
--- a/resources/win-chocolatey/tools/chocolateyinstall.ps1
+++ b/resources/win-chocolatey/tools/chocolateyinstall.ps1
@@ -1,10 +1,10 @@
-﻿$ErrorActionPreference = 'Stop'; # stop on all errors
-$packageName = 'yarn'
+$ErrorActionPreference = 'Stop'; # stop on all errors
+
 $packageArgs = @{
-  packageName = $packageName
+  packageName = 'yarn'
   softwareName = 'Yarn*'
   fileType = 'msi'
-  silentArgs = "/qn /norestart /l*v `"$env:TEMP\chocolatey\$($packageName)\$($packageName).MsiInstall.log`""
+  silentArgs = "/qn /norestart"
   validExitCodes = @(0, 3010, 1641)
   checksumType = 'sha256'
 
@@ -20,9 +20,17 @@ if (Test-Path "${env:ProgramFiles(x86)}\Yarn\package.json") {
 } else {
   $path = "$env:ProgramFiles\Yarn\package.json"
 }
+
 $script = @"
   (Get-Content -Path '$path') ``
     -replace 'installationMethod":.+', 'installationMethod": "choco",' ``
     | Set-Content '$path'
 "@
+
 Start-ChocolateyProcessAsAdmin -Statements $script
+
+if (-Not (Get-Command "node" -errorAction SilentlyContinue)) {
+  Write-Host "Yarn requires NodeJS to be installed. To install, use either of the commands below:"
+  Write-Host "choco install nodejs"
+  Write-Host "choco install nodejs-lts"
+}
--- a/resources/win-chocolatey/tools/chocolateyinstall.ps1.in
+++ b/resources/win-chocolatey/tools/chocolateyinstall.ps1.in
@@ -1,7 +1,9 @@
 $ErrorActionPreference = 'Stop'; # stop on all errors
 
+$packageName = 'yarn'
+
 $packageArgs = @{
-  packageName = 'yarn'
+  packageName = $packageName
   softwareName = 'Yarn*'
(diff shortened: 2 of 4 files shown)


pascalberger:
@Daniel15 Any chance you can have a look at this? This PR would allow to use the yarn Chocolatey package again with Node version different than LTS and solve the issue with the log output which confuses users.

Daniel15:
Oops, sorry I missed this! This seems mostly good to me.

The file is intentionally called `chocolateyinstall.ps1.in` because it contains placeholders. The idea is that if it has the `.in` extension, it has the placeholders, and it's renamed to remove the `.in` when the placeholders are removed. Could you please rename it back?

> The chocolately install has always been full of log spam due to an incorrectly written /l*v flag

If I remember correctly, this flag was copied as-is from a Chocolatey template for MSI-based installers.

resources/win-chocolatey/tools/chocolateyinstall.ps1
```diff
@@ -20,9 +20,17 @@ if (Test-Path "${env:ProgramFiles(x86)}\Yarn\package.json") {
...
     -replace 'installationMethod":.+', 'installationMethod": "choco",' ``
     | Set-Content '$path'
 "@
+
 Start-ChocolateyProcessAsAdmin -Statements $script
+
+if (-Not (Get-Command "node" -errorAction SilentlyContinue)) {
+  Write-Host "Yarn requires NodeJS to be installed. To install, use either of the commands below:"
+  Write-Host "choco install nodejs"
+  Write-Host "choco install nodejs-lts"
```
pascalberger: I would suggest to add this information also to the nuspec file as a note (like e.g. done [here](https://github.com/bbtsoftware/chocolatey-packages/blob/master/automatic/bower/bower.nuspec#L21-L24))

resources/win-chocolatey/tools/chocolateyinstall.ps1
```diff
@@ -1,10 +1,10 @@
-﻿$ErrorActionPreference = 'Stop'; # stop on all errors
-$packageName = 'yarn'
+$ErrorActionPreference = 'Stop'; # stop on all errors
+
 $packageArgs = @{
-  packageName = $packageName
+  packageName = 'yarn'
   softwareName = 'Yarn*'
   fileType = 'msi'
-  silentArgs = "/qn /norestart /l*v `"$env:TEMP\chocolatey\$($packageName)\$($packageName).MsiInstall.log`""
```
pascalberger: While this solves the issue of the log output, it also disables logging at all. Upstream issue for this is https://github.com/chocolatey/choco/issues/1016

JonathanWolfe: So our options are either a giant blob of Error looking xml or nothing. I prefer nothing but we can add back the error blob.

JonathanWolfe:
@Daniel15 
> The file is intentionally called `chocolateyinstall.ps1.in` ... Could you please rename it back?

Renamed back. I think I just changed it because I was checking syntax highlighting and the PSAnalyzer and forgot to undo.

> The package intentionally had a Node.js dependency so that it could work out-of-the-box without the user having to know about Node.js. I guess outputting a message is fine. I wish Chocolatey had a way of specifying that both `nodejs` and `nodejs-lts` satisfy the requirement.

Yeah, nuget doesn't have the concept of `either` only `and`. I don't think there will be many people trying to use Yarn without node installed anyways so the issue won't/shouldn't come up much.

PR #6510, opened 2018-10-06
Fix fail when upgrading incompatible optional dependencies

katakonst:
**Summary**
This was caused by setting the optional field to false.This is fixed by searching if the optional dependency exists  in manifest.


arcanis:
Hey @katakonst! Sorry for the delay - can you please add a test? I'll then merge it for the 1.13 release (scheduled in two weeks from now, I'd say).

katakonst:
Added a test!

kristian:
Will likely also fix #6526. Could this one be merged w/ the next release? A review is still pending. Thanks.

dylanmerrick:
Any chance this can go in? It's been sitting idle for a while...

kristian:
@arcanis any update here? Still we cannot use / switch to Yarn 2, because we have the issue that some of our optional dependencies do not exist in some of our repositories, and we need the install to not fail (which atm. it still does).

kristian:
@katakonst I just had a look at your change and I don't think it will actually solve the issue, isn't it?

Generally I see two problems:

a) Your change in the add command, I think the following statement is not a valid assumption for add:

```javascript
optional: !!manifest.optionalDependencies && !!manifest.optionalDependencies[pattern.split('@')[0]],
```

Isn't it that for the add command, the manifest is not yet updated at that point in time? Also the args that are added in prepareRequests are the args of the add command as I understand. Thus, isn't it that the lines should be simply changed to:

```javascript
optional: this.flags.optional,
```

So if `yarn add --optional ...` is called, all requests that the add command adds should be optional?

b) Second I do not see how your change will solve it for general `yarn install`, `yarn upgrade` when optional dependencies do not exist: The `prepareRequests` will only be called for the add command and install / upgrade do not use or inherit add. Thus I don't see how your suggested change solves the issue?

I much rather see the problem somewhere in the request stack. So when doing a `yarn install` the request stack prints:

```
PS <path>> yarn install
➤ YN0000: ┌ Resolution step
➤ YN0035: │ <package>@npm:^<version>: The remote server failed to provide the requested resource
➤ YN0035: │   Response Code: 404 (Not Found)
➤ YN0035: │   Request Method: GET
➤ YN0035: │   Request URL: <URL>
➤ YN0000: └ Completed in 0s 366ms
➤ YN0000: Failed with errors in 0s 371ms
```

Even for optional dependencies. I see that at least for `optionalDependencies` out of the `package.json`, `fetchRequestFromCwd` will set a `optional` flag and hint to `optional`, however when then later using the `PackageResolver` to resolve the dependencies, the resolver will still fail if the GET request fails (see error above). I could not locate where exactly the resolver does the web request / should ignore any 404 (Not Found) (or any error?) if optional dependencies are not found.

What do you think @arcanis, would this be a possible & right approach? So two changes:

1. Changing `add.js` from `optional: false` to `optional: this.flags.optional` and
2. Changing the resolver to no longer fail on errors with the web request for optional dependencies?

If you say this would be a better approach / something that could be merged to Yarn 2+, I would provide a new PR + tests. So please let me know! Thanks.

kristian:
I did create follow-up ticket: https://github.com/yarnpkg/berry/issues/4306 in the https://github.com/yarnpkg/berry/ repository. Please let me know if I should suggest a PR to fix it.

kristian:
For everyone that is wondering: I did create a https://github.com/kristian/yarn-plugin-better-optional plugin for Yarn 2+ (Berry), providing a workaround for this issue. Unfortunately it relies on the plugin mechanism of Yarn 2+ and it won't work in Yarn Classic, meaning you would have to upgrade. However, maybe this plugin is providing you a reason for performing the upgrade. 👍🏻

Issue #6512, opened 2018-10-07
Empty "bin" entry in package.json results in ".bin" symlink to directory


rhburrows:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

When yarn installs a dependency that has an empty string as an entry for "bin" in the package.json it creates a symlink to the package's root directory in `node_modules/` in the `node_modules/.bin` directory. An example package configured this way is [file-loader](https://github.com/webpack-contrib/file-loader)

**If the current behavior is a bug, please provide the steps to reproduce.**

```sh
rhburrows$ yarn add file-loader
yarn add v1.10.1
... MORE OUTPUT HERE ...
✨  Done in 0.66s.
rhburrows$ ls -al node_modules/.bin/
total 0
drwxr-xr-x   4 rhburrows  staff  128 Oct  7 09:32 .
drwxr-xr-x  18 rhburrows  staff  576 Oct  7 09:32 ..
lrwxr-xr-x   1 rhburrows  staff   14 Oct  7 09:32 file-loader -> ../file-loader
lrwxr-xr-x   1 rhburrows  staff   19 Oct  7 09:32 json5 -> ../json5/lib/cli.js
```

**What is the expected behavior?**

There should be no symlink in `node_modules/.bin`. That's the result when installing the same dependency with npm.

**Please mention your node.js, yarn and operating system version.**

Tested on:
* node 8.11.1 and 10.11.0
* yarn 1.7 and 1.10.1
* macOS 10.14

Gudahtt:
Thanks for the report!

It looks like this would be simple enough to fix. The module `src/util/normalize-manifest/fix.js` is normalizing the empty string entry into an object. It should skip this normalization when the entry is an empty string.

Would you be interested in submitting a PR?

rhburrows:
Sure, I'd be happy to take a shot at it.

PR #6515, opened 2018-10-08
Fix: handle empty bin strings

rhburrows:
**Summary**

This is a followup PR for #6512. It prevents yarn from creating a link to a dependency's root directory within the `.bin/` directory if the dependency has an empty string for "bin" in its package.json.

**Test plan**

After the change:

```sh
rhburrows$ yarn-local add file-loader --cache-folder ./cache
yarn add v1.13.0-0
info No lockfile found.
... OUTPUT ...
✨  Done in 1.17s.
rhburrows$ ls -al node_modules/.bin/
total 0
drwxr-xr-x   3 rhburrows  staff   96 Oct  7 16:59 .
drwxr-xr-x  18 rhburrows  staff  576 Oct  7 16:59 ..
lrwxr-xr-x   1 rhburrows  staff   19 Oct  7 16:59 json5 -> ../json5/lib/cli.js
```


rhburrows:
Thanks! I wasn't sure if there was a better way to test the change and your patch does a much better job verifying the intent of this PR.

I applied your patch to the branch.

__tests__/commands/install/bin-links.js
```diff
@@ -175,6 +175,15 @@ test('can use link protocol to install a package that would not be found via nod
   });
 });
 
+test('empty bin string does not create a link', (): Promise<void> => {
+  return runInstall({binLinks: true}, 'install-empty-bin', async config => {
+    const binScripts = await fs.walk(path.join(config.cwd, 'node_modules', '.bin'));
+    expect(binScripts).toHaveLength(1);
```
Gudahtt: This should have length 2 on Windows, as two files are created per binary on Windows (one Bash script for linux-like environments like Cygwin, and one `.cmd` Batch script). This is why the appveyor tests are failing.

Gudahtt:
Great, that seemed to work. The remaining failing test is unrelated to this PR; that test is known to be flaky and will occasionally timeout.

rally25rs:
I kicked off a new appveyor build. Hopefully it passes this time...

PR #6521, opened 2018-10-09
Comprehensive reorganization of localization file

andrewboerema:
**Summary**
I was doing some work on a feature and a saw how unorganized the localization file was. Not only does it make it a crap shoot for where to put a new string (if you even feel like doing it), it also makes it really difficult to approach translating into a new language at all. To that end, I made my best effort to organize the strings. I have grouped them by command (using common and core for things that don't relate to a specific command) alphabetized the groups, and alphabetized the strings within the groups. I also added a header to explain the organization within the file.

This will make the file much cleaner and more maintainable. It also makes it easy to catch people not adding strings in the right place during code review.

I do understand this will be a pain for a little bit because it will cause a number of merge conflicts. I feel that the organization and maintenance gains will make it worth it, though.

**Test plan**
If the tests pass, this should be good. I am not intentionally changing any functionality.

### DIFF
--- a/src/util/git.js
+++ b/src/util/git.js
@@ -200,7 +200,7 @@ export default class Git implements GitRefResolvingInterface {
       if (await Git.repoExists(secureUrl)) {
         return secureUrl;
       } else {
-        reporter.warn(reporter.lang('downloadGitWithoutCommit', ref.repository));
+        reporter.warn(reporter.lang('getDownloadGitWithoutCommit', ref.repository));
         return ref;
       }
     }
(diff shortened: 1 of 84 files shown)


arcanis:
The review will be fun, but this is a really good idea thanks for tackling this! I love "better engineering" PRs like this one 🙂

src/util/git.js
```diff
@@ -210,7 +210,7 @@ export default class Git implements GitRefResolvingInterface {
       if (await Git.repoExists(secureRef)) {
         return secureRef;
       } else {
-        reporter.warn(reporter.lang('downloadHTTPWithoutCommit', ref.repository));
+        reporter.warn(reporter.lang('getDownloadHttpWithoutCommit', ref.repository));
```
arcanis: Shouldn't this be `gitDownloadHttpWithoutCommit` (rather than `get`)?

andrewboerema: Yes, I will update.

PR #6527, opened 2018-10-10
Adding call to generate-disclaimer to get dep licenses

### LINKED ISSUE #549: Standalone JS build (`yarn-0.14.0.js`) needs license file
For the standalone builds of Yarn (the single `.js` files that contain everything), we need to build a license file containing the licenses for all the dependencies in order for our usage of their source code to comply with the licenses. 

For the regular tarball, we're complying by each dependency having a `LICENSE` file in its directory. For the standalone build, we're omitting the LICENSE files, so we need to combine them into one big license file.

It looks like `yarn licenses generate-disclaimer` is designed to do just that - We might need to run that as part of the build and save the generated file.

MIT license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

BSD license:

> 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
> 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.

andrewboerema:
**Summary**
Adding a build step for the dist version to create a license file for all dependencies and move it into the dist directory.
 This is to resolve #549

### DIFF
--- a/scripts/build-dist.sh
+++ b/scripts/build-dist.sh
@@ -23,6 +23,7 @@ mkdir artifacts
 mkdir dist{,/bin,/lib}
 
 # Workaround for https://github.com/yarnpkg/yarn/issues/2591
+eval $system_yarn licenses generate-disclaimer > dist/DEPENDENCY_LICENSES
 eval $system_yarn run build
 eval $system_yarn run build-bundle
 chmod +x artifacts/*.js


arcanis:
cc @daniel15

scripts/build-dist.sh
```diff
@@ -23,6 +23,7 @@ mkdir artifacts
 mkdir dist{,/bin,/lib}
 
 # Workaround for https://github.com/yarnpkg/yarn/issues/2591
+eval $system_yarn licenses generate-disclaimer > dist/DEPENDENCY_LICENSES
```
Daniel15: Move this so it's not directly below the "Workaround for ...." comment

Daniel15:
Seems reasonable to me!

PR #6535, opened 2018-10-12
fix: Restore Node v4 support by downgrading cli-table3

Gudahtt:
**Summary**

`cli-table3` no longer supports Node v4 as of v0.5.0. Downgrading to `cli-table3@0.4.0` restores Node v4 support, and should fix the nightly `test-e2e-ubuntu1604` CircleCI test that has been failing

This dependency was introduced in #6409, so this problem is present in v1.12.0

**Test plan**

Using Node.js `v4.8.7` and after running `yarn build-dist`, any command with the resulting artifact will give the following output:
```
C:\Development\git\yarn\artifacts\yarn-legacy-1.13.0-0.js:62791
  let code = codeRegex();
  ^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:140:18)
    at node.js:1043:3
```

After making this change, that error no longer occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Restore Node v4 support by downgrading `cli-table3`
+
+  [#6535](https://github.com/yarnpkg/yarn/pull/6535) - [**Mark Stacey**](https://github.com/Gudahtt)
+
 - Prevent infinite loop when parsing corrupted lockfile with unterminated string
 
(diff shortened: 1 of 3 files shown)


Gudahtt:
There are a few more details in [this comment](https://github.com/yarnpkg/yarn/pull/6409#issuecomment-429181371), where I initially discovered this problem.

It's interesting that this wasn't picked up in any of the other tests. I would have expected `test-linux-node4` to blow up at least. I'll look into that.

arcanis:
👍

PR #6536, opened 2018-10-12
Check package engines before running upgrade-interactive

orta:
**Summary**

Small quality of life improvement. This means that you won't accidentally take your time choosing your dependencies to upgrade and then find out that engines don't match.


arcanis:
:+1:

Issue #5840, opened 2018-05-18
yarn install --frozen-lockfile is not failing as expected


DavideDaniel:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
`yarn install --frozen-lockfile` does not generate a new yarn.lock but it does NOT fail as expected.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Have a yarn.lock with a package.json with some dependencies and no diffs.
2. Remove a dependency in your package.json (npm uninstall --save something)
3. `yarn install --frozen-lockfile` proceeds with install  instead of failing as docs say, and does not generate a new yarn.lock to reflect the change in package.json.

**What is the expected behavior?**
Per docs: `Don’t generate a yarn.lock lockfile and fail if an update is needed.`
It should fail as expected as CI builds need to break earlier rather than wait for build and then fail on a missing or mismatched dependency.

**Please mention your node.js, yarn and operating system version.**
Node 8.9.1 | 9.6.1
npm 3.10.10 | 5.6.0
MacOS 10.12.6

DavideDaniel:
Closing this ticket as it was due to modifications in config files related to repos and not yarn

DavideDaniel:
Turns out there were some things wrong on our end but there is still an issue with yarn install --frozen-lockfile - I've edited the title and the original post to reflect that. `yarn install --frozen-lockfile` does NOT generate a new yarn.lock but it does not fail.

DavideDaniel:
I haven't had a chance to validate but a starting point might be https://github.com/yarnpkg/yarn/pull/5572. In trying to address https://github.com/yarnpkg/yarn/issues/4778, code was deleted that would cause frozen-lockfile check to fail.

devrelm:
@DavideDaniel I'm seeing the same issue with `--frozen-lockfile` not throwing an error.

@torifat Have you had an opportunity to look into this?

devrelm:
@arcanis Is there any movement on this? There hasn't been any response from the assignee or anyone else within the Yarn org.

arcanis:
> yarn install --frozen-lockfile proceeds with install instead of failing as docs say, and does not generate a new yarn.lock to reflect the change in package.json.

I don't understand. What makes you think it should fail? If you remove dependencies from your package.json, your lockfiles might now contain extraneous useless entries that could be optimized away, but it'll still install things in a repeatable way without having to be modified (so no error needed).

Does that make sense?

[...]

PR #6554, opened 2018-10-18
fix: fail when yarn.lock is updated in workspaces when --frozen-lockfile is used

robin-drexler:
**Summary**
This PR is an attempt to fix [that `yarn` does not fail when yarn.lock is updated in a workspace project although `--frozen-lockfile` is used](https://github.com/yarnpkg/yarn/issues/5840)

The current version is meant as a starting point as I'm pretty certain that I have missed something.
The issue is caused, because in the [`patterns` passed to the `integrityChecker`](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L457) only include the root's dependencies. So it never checks if dependencies of workspaces are missing in the lockfile.

I wonder if always setting `includeWorkspaceDeps` to `true` in the installation cli command is the right thing to do, especially since those patterns are referred to as [`topLevelPatterns`](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L593) at a later point, which might indicate that the workspaces dependencies are missing on purpose.

Another idea I had (and that I tried before I found the flag), was to pass all patterns to the `integrityChecker`, but otherwise leave them as is.

Basically to add:

```js
const dependencies = Array.from(
  this.resolver.getAllDependencyNamesByLevelOrder(patterns)
);

const allPatterns = dependencies.reduce((allPatterns, dependency) => {
  const allInfo = this.resolver.getAllInfoForPackageName(dependency);
  allInfo.forEach(({ _reference: { patterns } }) => {
    for (const pattern of patterns) {
      allPatterns.add(pattern);
    }
  });
  return allPatterns;
}, new Set());

```

before: https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L457 

If anybody could point me in the right direction, I'd be happy to continue working on a fix.

Thanks a lot in advance! :)


robin-drexler:
Hey, is there anything that I can do to push things forward? :)

kachkaev:
ping folks 👋

The current behaviour of the `--frozen-lockfile` argument does match what the docs are saying. Because of that, CI pipelines for PRs succeed even if somebody updated `package.json`, but did not update `yarn.lock`. Inconsistent sets of packages land the master branch and cause a lot of confusion down the line

nevir:
Same question here: what can we do to help get this in?

kachkaev:
Before this PR is merged, some workarounds can be found in https://github.com/yarnpkg/yarn/issues/5840

arcanis:
Hey, sorry for now pinging back sooner.

The change appears sound, but we're currently working on the v2 (scheduled before the end of the year) and I'm somewhat worried about introducing subtle bugs right before releasing the next major (the way I see it, the current v1 is, despite its shortcomings, relatively stable).

One question in particular I'd need to see answered is why isn't `includeWorkspaceDeps` always on by default? What does it imply to enable it for every install? I don't remember why it got implemented. Maybe @jgoz would remember (https://github.com/yarnpkg/yarn/pull/4654)?

jgoz:
I think I remember why it got implemented in the first place. It’s because running certain commands inside a child workspace needed to consider dependencies across the entire workspace (`outdated`, `upgrade-interactive`), mainly because it made more sense from the user’s perspective (maybe? It’s been a long time since I looked at this).

I haven’t fully considered this current change but a safer approach might be to only enable `includeWorkspaceDeps` when `—frozen-lockfile` is specified, rather than always enabling it.

I should also note that we long ago switched to using `—pure-lockfile` in our CI because after reading the docs more carefully, it was the flag we actually wanted.

robin-drexler:
I know, it's been a while, but can this PR be merged since it got an approval? :) 

Thanks a lot!

NMinhNguyen:
> 👋 I know, it's been a while, but is there any chance that this PR could be merged since it got an approval? :) 

I don't think the approval was from a core maintainer.

Issue #3254, opened 2017-04-25
Allow install only devDependencies


robdodson:
**Do you want to request a *feature* or report a *bug*?**
Hopefully this is a small feature request. I'd like a flag that allows me to only install `devDependencies`. This would allow me to split my flat installed client code from my build tools code.

Thanks to the fix in https://github.com/yarnpkg/yarn/issues/2781#issuecomment-296955569 I can now do:

```
yarn install --flat --production --modules-folder client_modules
```

And only install `dependencies` in my `client_modules` directory. This is awesome ❤️

If I also do `yarn`, it will install `dependencies` and `devDependencies` into my `node_modules` directory. But I already flat installed my `dependencies` into `client_modules` so installing them again in `node_modules` isn't necessary.

**What is the current behavior?**
There's no way to only install `devDependenices`.

**What is the expected behavior?**
A flag, maybe `yarn install --dev-dependencies` or something like that?

rally25rs:
@bestander triage; label: [new feature]

bestander:
In a way discussed here https://github.com/yarnpkg/rfcs/pull/36

chrishyle:
This feature would be great. Client-side projects really benefit from `yarn install --prod --flat` because it avoids potentially costly duplication of dependencies that would otherwise have to be shipped to the end-user's browser over the wire. That being said, client-side projects also generally rely on devDependencies for build/dev (things like babel plugins, eslint configs, etc) that have no need to be flattened, and resolving the devDependencies version conflicts can be tedious. For example, on a recent project, my *dependencies* had one conflict which I understood clearly how to resolve, but my *devDependencies* had 25 conflicts, on projects like "camelcase", "string-width" and a bunch of other stuff that I had no idea how to resolve, and because they're not in shipping code, I don't really have a need to resolve.

Further complicating this is that using multiple folders as @robdodson is doing is sometimes difficult because some useful tools and scripts in the ecosystem have unfortunately hard coded to "node_modules/" (ran into this with [ImportJS](https://github.com/Galooshi/import-js/blob/3b458f4c7d0aea1a2170c20d6acaf6fc42687c81/lib/findJsModulesFor.js#L29), I assume there are others). While they shouldn't do that, and there would be clear fixes, I am always nervous in my own projects about introducing new non-standard configurations because of these kinds of conflicts.

So, ideally, I'd be able to run:

```sh
yarn install --prod --flat
yarn install --dev-dependencies
```

…and have them install in the same folder.

[...]

PR #6564, opened 2018-10-22
Add flag `--devDependencies` to install dev dependencies only

eyalzek:
This PR aim to support the feature described in this ticket: https://github.com/yarnpkg/yarn/issues/3254, specifically: https://github.com/yarnpkg/yarn/issues/3254#issuecomment-427171762


arcanis:
While I understand the theory, I don't think it's a good idea. I'm not fond of having flags that install the directory tree in entirely different ways (note it's also true for `--flat` and even `--production` in some degree - this one is a bit more acceptable).

Yarn is all about reproducible builds, and if a flag is needed to get the same output on a machine than on another then it adds a new layer of complexity I'm not comfortable with. Additionally, such a feature seems specific to a single use case, and I prefer for us to avoid that. Features we add should be generic and future-proof, not restricted to a single use case.

arcanis:
While I understand the theory, I don't think it's a good idea. I'm not fond of having flags that install the directory tree in entirely different ways (note it's also true for `--flat` and even `--production` in some degree - this one is a bit more acceptable).

Yarn is all about reproducible builds, and if a flag is needed to get the same output on a machine than on another then it adds a new layer of complexity I'm not comfortable with. Additionally, such a feature seems specific to a single use case, and I prefer for us to avoid that. Features we add should be generic and future-proof, not restricted to a single use case.

arcanis:
While I understand the theory, I don't think it's a good idea. I'm not fond of having flags that install the directory tree in entirely different ways (note it's also true for `--flat` and even `--production` in some degree - this one is a bit more acceptable).

Yarn is all about reproducible builds, and if a flag is needed to get the same output on a machine than on another then it adds a new layer of complexity I'm not comfortable with. Additionally, such a feature seems specific to a single use case, and I prefer for us to avoid that. Features we add should be generic and future-proof, not restricted to a single use case.

dreyks:
this looks like a start of a long discussion similar to ruby's bundler `--only` feature.

the feature is mostly needed for CI flow, e.g. I'm transpiling sources on production machines (heroku) so almost all my dependencies are in `dependencies` group. However there are several really `dev` dependencies - linters, that are to be run only on CI and are not needed in production.

deiga:
@arcanis Like dreyks already commented, this flag is really useful for CI and when using multi-stage builds in Docker.

daern91:
Any more thoughts on this one, @arcanis? Would love to use this in our CI env.

bishtawi:
Any update on this? I would like to see this functionality. I would use it in my CI tests.

rebornix:
This feature is particularly useful for CI pipeline and sometimes for CD as well, especially for monorepos. For example we ship Monaco Editor from VS Code's source code and Monaco Editor has zero dependencies, but we need some shared dev dependences to run tests and builds. In a CI/CD environment, we'd like to install update-to-date dev dependencies only to ensure the pipeline is always fast enough. Currently it's simply not possible.

> Yarn is all about reproducible builds, and if a flag is needed to get the same output on a machine than on another then it adds a new layer of complexity I'm not comfortable with

It doesn't necessarily add another layer, as long as `--production` exits. `--dev` is the counter part of `--production`. Also note that due to the existence of `--production`/`--flat`/`--ignore-optional` (not sure if there are more), flags are need to get the same output on a machine at this moment.

MohamadKh75:
Any updates? 🤔

arcanis:
Yarn 1 is frozen for features, and this won't be merged (sorry @eyalzek :(). Even in the new trunk I'd probably decline for the same reason as stated before: it's difficult enough to understand the rules that govern how dependencies are installed, the less flags we have the better I feel *for the Yarn core*.

With that being said, Yarn now supports plugins (starting from the 2.x line), and it would be trivial for someone to create a contrib plugin implementing this behavior, but without our team having to maintain it. I'd recommend this approach if that's really what you want.

In the meantime I'll close this PR.

redaxmedia:
There is a solution for those who need devDependencies only:

```
npx handpick --target=devDependencies --manager=yarn
```

PR #6567, opened 2018-10-23
Console info, success, warning and error messages now contain emoji's

RoyboyNL:
**Summary**
Self-explanatory, these console messages didn't display emoji's while the steps did, now it's easier to see if it is a warning, error or info message. 

**Test plan**
Run `yarn install --emoji` on a project, cause some errors, warnings and info messages and see it has now emoji's.


arcanis:
Thanks! But I think we're overall happy with things being the way they are currently 🙂 

Plus, it allows us to quickly see where is each step of the installation in the logs, so this change isn't a strict improvement.

PR #6568, opened 2018-10-23
Support Darwin bash in bin/yarn

hboomsma:
If the yarn binary is a symlink the `basepath` is not correctly resolved on OSX/Darwin.

For example take the following structure:
```
vendor/yarnpkg/yarn/bin/yarn
vendor/bin/yarn -> ../yarnpkg/yarn/bin/yarn
```

And run yarn:
```
$ yarn
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module '/Users/yarnpkg/yarn/bin/yarn.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
```
Since the symlink is not resolved the relative path in the symlink leads to the wrong path.

This PR, fixes that issue.

**Summary**

**Test plan**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6562](https://github.com/yarnpkg/yarn/pull/6562) - [**Bertrand Marron**](https://github.com/tusbar)
 
+- Fixes Yarn invocations on Darwin when the `yarn` binary was symlinked
+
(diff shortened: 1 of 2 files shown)


serverwentdown:
This PR did not fix anything, it only resolves the symlinks in the parent directory of the yarn script. I submitted #6930 to actually fix the problem.

Issue #6607, opened 2018-10-31
audit returns `Unexpected audit response (Missing Metadata): false` 


gabegorelick:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
```
$ yarn audit
yarn audit v1.12.1
error An unexpected error occurred: "Unexpected audit response (Missing Metadata): false".
info If you think this is a bug, please open a bug report with the information provided in "/Users/gabe/code/pushbot-api/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/audit for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Run `yarn audit`.

**What is the expected behavior?**

No error. Or provide docs on how to resolve if it is a setup issue (https://github.com/yarnpkg/website/issues/877).

**Please mention your node.js, yarn and operating system version.**
Node: v8.12.0
Yarn: v1.12.1 (installed via Homebrew)
OS: macOS

arcanis:
Ping @rally25rs - I know that the npm registry often yields 500, maybe it's what happens there and we don't properly catch it?

rally25rs:
If you run with `--verbose` you can get the message that we sent to the npm registry, and it's response.

Also, please provide your dependencies.

bugzpodder:
When I run yarn --audit, it says 1 vulnerbility is detected and asked me to run yarn audit.  
I did so and go the same error.

`verbose 2.423 Performing "POST" request to "https://registry.yarnpkg.com/-/npm/v1/security/audits".
verbose 2.895 Request "https://registry.yarnpkg.com/-/npm/v1/security/audits" finished with status code 400.
verbose 2.9 Error: Unexpected audit response (Missing Metadata): false
    at Audit.<anonymous> (/usr/local/Cellar/yarn/1.12.1/libexec/lib/cli.js:56117:15)
    at Generator.next (<anonymous>)
    at step (/usr/local/Cellar/yarn/1.12.1/libexec/lib/cli.js:92:30)
    at /usr/local/Cellar/yarn/1.12.1/libexec/lib/cli.js:103:13
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
error An unexpected error occurred: "Unexpected audit response (Missing Metadata): false".`

arcanis:
@bugzpodder As @rally25rs mentioned, the package.json and yarn.lock would be helpful 🙂

[...]

PR #6611, opened 2018-10-31
fix audit for packages without name

bugzpodder:
Fixes https://github.com/yarnpkg/yarn/issues/6607#issuecomment-434855932


arcanis:
Now that's interesting - how come some packages have no name ? 😮

rally25rs:
If they don't have a `name` then we should probably omit them from the audit contents at all, since npm isn't going to have issues listed for a package with no name anyway. I think we should just skip it when building the audit tree.

bugzpodder:
@rally25rs the main package is a private package (not published) and it doesn't have a name, however its dependencies should still be checked.

bugzpodder:
See for example: https://github.com/yarnpkg/yarn/issues/6607#issuecomment-434870068

gabegorelick:
This does not appear to fix #6607. I still get that error even after applying this patch.

bugzpodder:
@gabegorelick It most definitely fixes https://github.com/yarnpkg/yarn/issues/6607#issuecomment-434870068 
But I understand you have have a different problem.  However we can't help you unless there is a repro case.

bugzpodder:
`$ yarn audit
yarn audit v1.12.1
error An unexpected error occurred: "Unexpected audit response (Missing Metadata): false".
info If you think this is a bug, please open a bug report with the information provided in ".../yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/audit for documentation about this command.
$ yarn-local audit
yarn audit v1.13.0-0
0 vulnerabilities found - Packages audited: 37909
✨  Done in 2.56s.`

gabegorelick:
@bugzpodder My apologies. In posting a minimal repo case for #6607, it seems I got too minimal in dropping the `name` field from package.json and discovered another bug :)

arcanis:
The problem comes from empty `dependencies` fields (cf https://github.com/yarnpkg/yarn/issues/6607#issuecomment-434875058).

@bugzpodder, do you mind also fixing it in the same PR? It should just be a change in `_mapHoistedTreesToAuditTree` (something like transforming empty dependency fields to `*` should do the trick).

arcanis:
Perfect, thanks!

PR #6621, opened 2018-11-01
fix: Use cmdshim instead of symlink on win32 in base-fetcher

yoadsn:
**Summary**
This is an implementation of the suggestion by @arcanis to fix #6604. In short, on win32 platforms, the symlink requires elevated privileges. This fix would use the cmdshim to work around the need for the symlink specifically on the `base-fetcher` module.

**Test plan**
Same test suit was used since platform emulation is not included today.

PR #6628, opened 2018-11-03
Stop copying `.bin` folders

### LINKED ISSUE #6622: yarn 1.12.1 creates .bin folders inside dependency folders
1.12.1

Bug

Installing a dependency with a binary not only adds a `.bin` to node_modules, but also to the dependencies own folder.

This did not happen in yarn 1.10.1.

1. Install yarn and node in a docker or lxc container
2. Create a module with a dependency that includes a binary (that is installed to `node_modules/.bin`). Suppose for example [marked](https://www.npmjs.com/package/marked)
3. run `yarn`
4. look in `node_modules/marked/` there will be a `.bin` folder.

Dependencies should not have `.bin` folders with extra symlinks. 

node: 8.12.0 (also tried 8.11.3)
yarn: 1.12.1
OS: Ubuntu 18

I tried the error on multiple ubuntu and debian container images.

**EDIT:** I thought this was only occurring inside containers, but it turns out only the containers were updated. I suspect this just broke in the newest 1.12

arcanis:
**Summary**

Fixes #6622

**Test plan**

Added a test

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -48,6 +48,11 @@ async function mockConstants(base: Config, mocks: Object, cb: (config: Config) =
 beforeEach(request.__resetAuthedRequests);
 afterEach(request.__resetAuthedRequests);
 
+test('install should not copy the .bin folders from the cache', () =>
+  runInstall({}, 'install-no-bin', async config => {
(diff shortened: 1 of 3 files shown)


arcanis:
No, but #6621 will

PR #6629, opened 2018-11-03
Uses NODE_OPTIONS with PnP instead of CLI args

arcanis:
**Summary**

A common problem people currently have is with programs that call Node through the `child_process` interface but forget to forward the `execArgv` parameters. This morning alone there was three of them: `@babel/node`, `node-sass`, and `node-gyp`.

In order to alleviate this issue, this diff makes Yarn switch from passing the `--require` option explicitly through the command line to passing it via the `NODE_OPTIONS` environment variable. This should increase the compatibility at a low cost.

An alternative would be to monkey-patch the `child_process` module, but I'm strongly opposed to it - patching builtin modules might have serious consequences, and shouldn't be done only for the sake of convenience.

Since relying on the environment is sometimes tricky, I'd appreciate having a few set of eyes take a look at this diff before I merge it - just to be sure I don't miss something (cc @yarnpkg/core).

Also cc @jdalton (since that might maybe affect `esm`?)

**Note:** This PR has one problem, though: it might cause issues for people having spaces into their project path. This is caused by a bug/oversight in the way Node parses the NODE_OPTIONS value. I've work on a PR to fix it on their side (https://github.com/nodejs/node/pull/24065), but that might cause some pain in the meantime. I could potentially add a warning when it happens, but it will still be displayed once the problem is solved - I'm not sure whether it's a good UX or not.

---

**Test plan**

Added tests, the existing one should pass as before


src/cli/commands/node.js
```diff
@@ -21,14 +21,16 @@ export function hasWrapper(commander: Object, args: Array<string>): boolean {
 export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
   const pnpPath = `${config.lockfileFolder}/${PNP_FILENAME}`;
 
+  let nodeOptions = process.env.NODE_OPTIONS || '';
   if (await fs.exists(pnpPath)) {
-    args = ['-r', pnpPath, ...args];
+    nodeOptions += ` --require ${pnpPath}`;
```
BYK: I'm guessing the leading space wouldn't be an issue when `NODE_OPTIONS` is empty?

src/util/execute-lifecycle-script.js
```diff
@@ -227,6 +219,11 @@ export async function makeEnv(
 
       pathParts.unshift(`${dependencyInformation.packageLocation}/.bin`);
     }
+
+    // Note that NODE_OPTIONS doesn't support any style of quoting its arguments at the moment
```
BYK: Referring to your PR on Node would be nice. Also, would be great if we can centralize some of this PnP path logic.

src/util/execute-lifecycle-script.js
```diff
@@ -45,7 +37,6 @@ export async function getWrappersFolder(config: Config): Promise<string> {
 
   await makePortableProxyScript(process.execPath, wrappersFolder, {
```
BYK: Shall we unify this logic in the future and use `node` at all times?

packages/pkg-tests/pkg-tests-specs/sources/pnp.js
```diff
@@ -1340,5 +1340,58 @@ module.exports = makeTemporaryEnv => {
...
+            )}]).toString())`,
+          );
+
+          expect(JSON.parse((await run(`run`, `script`)).stdout)).toMatchObject({
+            name: `no-deps`,
+            version: `1.0.0`,
+          });
+        },
+      ),
+    );
```
BYK: I think we need two more tests:

1. For cases where we spawn another script and purposefully omit `node` args which simulates the scenario you are solving for (is this the second test, I can't really tell?)
2. Have an already set `NODE_OPTIONS` env variable and make sure it is preserved.

src/util/execute-lifecycle-script.js
arcanis: What do you mean? At most we should replace `makePortableScript` with `cmdShim` (which already covers the behavior we're using here, I think)

packages/pkg-tests/pkg-tests-specs/sources/pnp.js
arcanis: The first test tests `yarn node my-script.js'`, and the second tests `yarn run my-script`. I think that should cover the first case? Will add a test for empty `NODE_OPTIONS` 👍

src/util/execute-lifecycle-script.js
arcanis: I think I'll keep it duplicated for now - since it's very few lines, I'm concerned about over-abstracting it (since it's only used in two places and there are no reasons to think it will change)

src/cli/commands/node.js
arcanis: Yep, will work 👍

jdalton: It looks like
```js
nodeOptions += ` --require ${pnpPath}`
```
should be 
```js
nodeOptions = `--require ${pnpPath} ${nodeOptions}`
```
so that pnp is the first preloaded module.

Issue #6592, opened 2018-10-29
Yarn upgrade-interactive --latest freezes


alexdevero:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
Yarn freezes after running `upgrade-interactive --latest` and selecting dependencies  to upgrade. Updating `package.json` manually and running `yarn install` works.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Run `upgrade-interactive --latest`.
2. Select dependencies to upgrade and confirm.

**What is the expected behavior?**
Yarn should download the latest version of selected dependencies.

**Please mention your node.js, yarn and operating system version.**
Node: v11.0.0
NPM: 6.4.1
OS: Win 10 (1803)
CLI: tried on Windows native CMD, ConEmu and Cmder.

**Videos**
Running `yarn upgrade-interactive --latest`:
![yarn-upgrade-01](https://user-images.githubusercontent.com/4237724/47639049-a5cfc300-db60-11e8-99bc-a99e5f48841b.gif)

Updating `package.json` manually and running `yarn install`:
![ezgif-1-f14f7e9ffc73](https://user-images.githubusercontent.com/4237724/47639840-8174e600-db62-11e8-8b03-e22979664ab8.gif)

rally25rs:
It seems to work fine for me. Might be something with your specific packages and options? Could you try adding `--verbose` flag too and see if that gives any indication of what it's doing?

dbusacca:
I've got the same issue (even without --latest) in three different Windows 10 pc. I think it's a node 10 incompatibility issue.

It's has been already reported in https://github.com/yarnpkg/yarn/issues/6367

alexdevero:
@rally25rs what version of node do you have on your computer?

IlCallo:
Same here.
Running on Node 10.13

rally25rs:
I tried node 10.1 and 10.13 on OSX and didn't have any problems. Might just be a Windows + node 10 specific issue? Unfortunately I don't have my Windows machine with me today to try that out. If someone who can consistently reproduce this could help debug, that would be awesome!

rally25rs:
Closing as duplicate of #5949 
Let's continue the discussion there.

AliMD:
same issue with OSX and node v16.13.2

wouter-leistra:
[...]

Issue #6367, opened 2018-09-07
Yarn stuck at upgrade-interactive


s-bauer:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I create a new folder, open a command prompt, run`yarn init` and install an outdated version of a package. E.g.
`yarn add glob@7.1.2`. Then I try to upgrade the package using `yarn upgrade-interactive --latest`. It detects that the `glob` package is outdated, so I select it and hit Enter. Then it shows `? Choose which packages to update. glob@7.1.3` and is stuck. It does not do anything!

**If the current behavior is a bug, please provide the steps to reproduce.**
See above

**What is the expected behavior?**
The `upgrade-interactive` command shall update the packages and package.json.

**Please mention your node.js, yarn and operating system version.**
Windows 10 - 1709 - 16299.611
Yarn v1.11.0-20180905.1625 (tried it with the latest stable release aswell)
Node 10.10.0 (Doesn't work with latest LTS either)

The upgrade-interactive command worked a few weeks ago. I didn't change anything, but it stopped working. Please let me know when you need any additional information. Thanks!

yyynnn:
Same here any version form 1.7 on win7

s-bauer:
@torifat Did you have to time to take a look at this? It's really annoying!

bkniffler:
Hey there, I've just done `npm i -g yarn@1.12.0` and it seems to work.

s-bauer:
@bkniffler Nope, does not seem to work for me! I tried yarn@1.12.0 and yarn@1.13.0 and both get stuck at the upgrade-interactive command!

alexdevero:
Can confirm that yarn get stuck when running upgrade-interactive.

Yarn: 1.12.1
Node: v11.0.0

rally25rs:
Closing as duplicate of #5949 
Let's continue the discussion there.

Issue #6198, opened 2018-08-01
upgrade-interactive freezes


jamiewinder:
I've had this same problem on 1.9.2 and the previous stable version (1.7.0?).

I'm on Node 10.7.0 (also tried 10.6.0) on Windows 10.

If I run `yarn upgrade-interactive`, select the packages I want to upgrade, then I get the summary printed to the terminal, e.g. if I select react:
```bash
? Choose which packages to update. react@16.4.1
```
... but then it freezes. I can't abort (Ctrl-C) the command and return to the terminal either. 

I've tried running this with `--verbose` to give some more detail for this report, but unfortunately nothing else is printed to the terminal. Is there somewhere else I can look for an error to add to this report?

`upgrade`, `install` and all the other commands seem to work. I also doesn't just affect one project. I've also tried uninstalling and reinstalling yarn completely, to no avail.

Jontem:
Same here on Windows 10, node v10.3.0 and yarn 1.7.0.

jamiewinder:
Has anyone figured out a workaround to this while it gets sorted? I'm currently running the command to get the list of updatable dependencies then aborting out and doing the updates manually. Time consuming, but it works.

chasecaleb:
I'm on Windows 10 with node 10.3.0 and yarn 1.9.4. I can confirm this behavior is consistently broken across cmd, powershell, and git-bash's mintty.

Jordi-Carrasco:
Same here. Windows 10. node 10.7.0, yarn 1.9.4

danielmoncada:
also having the same problem.

Windows 10, Yarn 1.9.4, Node 10.10.0

j-cong:
I have this problem too.
Windows 8.1, node 10.6.0, yarn 1.10.1
With both git-bash and cmd

beeplin:
yarn 1.10.1 solves this problem for me. windows 10 1709, node 8.12.0

jamiewinder:
It's still the same for me on that version of Yarn unfortunately. 

Noticed a duplicate of this issue in #6367. 

It would be nice to be able to help with this, but I've no idea where to start. Are there any logs I can enable to help trace the cause? Like I said above, I've tried `--verbose` mode, but it doesn't give any extra info.

worcode:
Same here

### env
- windows 10 build 1803
- node v10.12.0
- yarn 1.10.1

bkniffler:
[...]

Issue #5949, opened 2018-06-07
Upgrade-interactive does not work for Node.js above v10.2.0


negezor:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
After the packages are selected, the installation hangs in place without further output

**If the current behavior is a bug, please provide the steps to reproduce.**

I updated my version of Node.js from v10.0.0 to v10.4.0, this behavior started from v10.2.0
Steps:
```shell
$ yarn upgrade-interactive --latest
```
I select packages for updating, I press enter and all

**What is the expected behavior?**
Updating packages

**Please mention your node.js, yarn and operating system version.**
Node.js: v10.4.0
Yarn: 1.7.0
Windows: 10

gluons:
Me too. It hangs. Can't stop it with <kbd>Ctrl</kbd> + <kbd>C</kbd>.

Node: `v10.7.0`
Yarn: `1.7.0`
OS: Windows 10

rally25rs:
Going to close several other issues as duplicates of this one, since this seems to be the oldest issue that reports this.

From what I can tell it's specific to Windows + Node 10.

If someone who can consistently reproduce this can help debug, it would be appreciated!

rally25rs:
The best lead on tracking this down seems to be https://github.com/yarnpkg/yarn/issues/6198#issuecomment-433551126

heyimalex:
I know this has been merged, but just confirming that running upgrade-interactive off head works for me now!

kallaspriit:
Had the same issue with the latest stable of 1.12.1, works with nightly 1.13.0 version.

krnlde:
@contributors: Close?

rolbr:
Just want to confirm:

I'm on Windows 10. With Node v10.15.3 + Yarn 1.10.1 `yarn upgrade-interactive` would show me the table of dependencies to upgrade but would then freeze. Downloading the latest stable installer from https://yarnpkg.com/lang/en/docs/install/#windows-stable (1.19.1) fixed the issue.

Thanks! (I think this can be closed)

PR #6635, opened 2018-11-05
Fixes potential freeze on win+node10 interactive upgrades (#5949)

pfeigl:
There is a potential freeze when running interactive upgrades after selecting the options. (See #6592, #6367, #6198, #5949)

This boiles down to a problem in the `inquirer` library which is used internally to provide the possible options on the CLI. Updating `inquirer` to the latest version fixes the problem.

All tests keep passing also with the new version of inquirer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 - Fixes an issue with how symlinks are setup into the cache on Windows
 
   [#6621](https://github.com/yarnpkg/yarn/pull/6621) - [**Yoad Snapir**](https://github.com/yoadsn)
+  
+- Upgrades `enquirer`, fixing `upgrade-interactive` for users using both Node 10 and Windows
+
+  [#6635](https://github.com/yarnpkg/yarn/pull/6635) - [**Philipp Feigl**](https://github.com/pfeigl)
 
 ## 1.12.1
 
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "hash-for-dep": "^1.2.3",
     "imports-loader": "^0.8.0",
     "ini": "^1.3.4",
-    "inquirer": "^3.0.1",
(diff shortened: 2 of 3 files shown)


s-bauer:
Awesome! Let's hope this gets approved fast!

arcanis:
The problem I see is that the enquirer part of Yarn isn't well tested, to the best of my knowledge (since it's quite harder to test user inputs etc - not impossible, but harder).

Can you confirm that a thorough use of `upgrade-interactive` with this patch enabled doesn't have any side effect? Can you also check `yarn version`, and `yarn add lodash@does-not-exist`?

pfeigl:
I ran quite some interactive upgrades on existing projects with my local build (major and minor package upgrades) which all worked as expected.

Additionally tested your two additional commands which also worked as expected. Also selecting from the list of the available lodash versions worked as expected.

All those tests have been run on my local devbox (win10, node10) and therefor not on any linux/mac box.

arcanis:
Awesome, thank you 😃

pfeigl:
Nice, thanks for the fast merge! Btw reg. the Changelog, the library used is `inquirer` (https://github.com/sboudrias/Inquirer.js) not `enquirer` (http://enquirer.io/), which pretty much do the same but are different implementations.

arcanis:
Woops, fixed 👍

rally25rs:
Thanks for the contribution @pfeigl ! It's going to make a lot of Windows users happy. 👍

pfeigl:
Unfortunately I did not and looking at older release notes of inquirer, they seem to have droped support for Node v4, see here: https://github.com/SBoudrias/Inquirer.js/releases/tag/v4.0.0

I'm not sure whether they mean you cannot build inquirer with node4 or whether consuming the library is also limited to node v5+

arcanis:
Compat is fixed in #6640 - I just configured webpack to transpile inquirer down to ES5. It seems like there are no runtime dependencies on Node 6+, so that should hold for now.

We really need to drop Node 4 😕

Issue #4581, opened 2017-09-28
Yarn does not support setting npm_config env via CLI parameter like NPM


joneslee85:
In NPM's `run-scripts`, user could set custom NPM env (entry with `npm_config` prefix via CLI parameter:

```
npm run env --my_custom
```

which generate `npm_config_my_custom=true`. Verifying it easily with:

```
npm run env --report | grep "my_custom"
```

This feature is not present with yarn yet. I am wondering if the core team would consider implementing this feature at all.

BYK:
If you submit a PR, why not? :)

shellscape:
It's kind of bananas that after 4+ years Yarn still doesn't have compatibility for this. #6644 has been open for over 3 years without review to resolve this. @BYK the request was made for a PR and one has been linked to this. Let's get this super old compat bug fixed up.

runlevel5:
@jeffminsungkim AFAIK this feature has not been implemented. You are welcome to have a go at it.

pluiz-briteris:
anyone working on this?

PR #6644, opened 2018-11-07
Feature: expose CLI args to process.env.npm_config_* at lifecycle to compatible with npm usage

mc-zone:
**Summary**

This PR is a small addition to make better compatible about the migrating from NPM to yarn.

While run scripts, NPM expose all configuration parameters into `process.env.npm_config_*` for convenient use. 

For example, I used to install `sqlite3` which depends on `node-pre-gyp` that accepts a binary mirror that can be passed  through CLI:

```bash
npm install sqlite3 --node_sqlite3_binary_host_mirror=http://npm.taobao.org/mirrors/
```

It might seemed also should work using yarn, but actually, it can't. Because `node-pre-gyp` [get this option](https://github.com/mapbox/node-pre-gyp/blob/674fda7b28b86bb3b11889db86c15c4d06722d02/lib/util/versioning.js#L309-L312) by `process.env.npm_config_node_sqlite3_binary_host_mirror`:

```javascript
// support host mirror with npm config `--{module_name}_binary_host_mirror`
// e.g.: https://github.com/node-inspector/v8-profiler/blob/master/package.json#L25
// > npm install v8-profiler --profiler_binary_host_mirror=https://npm.taobao.org/mirrors/node-inspector/
var host = process.env['npm_config_' + opts.module_name + '_binary_host_mirror'] || package_json.binary.host;
```

So the PR try to support this feature.

Also relative former issue: #4581

**Test plan**

Tests added in fixtures/lifecycle-script/ .


arcanis:
Thanks for this PR!

However, I'm not too fond of this. It can be easily workaround as such:

```
npm_config_node_sqlite3_binary_host_mirror=http://npm.taobao.org/mirrors/ yarn install
```

This PR means that we would have to always accept all options from the command line (because who knows, maybe the scripts need them), preventing us from doing a proper validation. That doesn't seem a good compromise.

mc-zone:
> However, I'm not too fond of this. It can be easily workaround as such:
> 
> ```
> npm_config_node_sqlite3_binary_host_mirror=http://npm.taobao.org/mirrors/ yarn install
> ```

Yes you can, but it's not a corss platform way. You know there are some difference between Windows / Mac / Linux. In win32 cmd we should use `set xxx=xxx`. This bings some costs to make other fellows and enviroments work.

arcanis:
Hm, maybe, still I don't think comfortable enabling this for all arguments. Best case, I'd go for something like this:

```
yarn -c node_sqlite3_binary_host_mirror=http://npm.taobao.org/mirrors/
```

mc-zone:
@arcanis Thanks for your advice! It seems a good idea. I'd like to change the implementation right away. Just a detail confirm, how to accept multiple arguments in this syntax? if we don't want to handle all of them, should I take them one by one like this? 

``` 
yarn -c arg1=foo -c arg2=bar --other-flag
```

arcanis:
Yep, multiple repetitions of the `-c` flag sound good to me 👍

Sogl:
Any news on this PR?

mikehenrty:
This would be an amazing feature for `yarn`, especially useful when orchestrating docker commands through package.json scripts. Is there anything I can do to help get this moving again?

PR #6653, opened 2018-11-08
Remove stats.json

lheckemann:
Remove stats.json which shouldn't be there as far as I can tell.

Seems to be some sort of IDE metadata erroneously introduced in
7ff626c7266a895e3db99a0a56d615587a83ab68 (#5934)

cc @BYK who added this


lheckemann:
Would this auto-merge if the test hadn't failed? The failure is unrelated to my change AFAICT.

BYK:
It won't auto merge and I agree it seems completely unrelated. Restarted the job to make sure tho/

BYK:
I think that item is broken altogether on around this specific commit:
```
node@10 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have node@10 first in your PATH run:
  echo 'export PATH="/usr/local/opt/node@10/bin:$PATH"' >> ~/.bash_profile

For compilers to find node@10 you may need to set:
  export LDFLAGS="-L/usr/local/opt/node@10/lib"
  export CPPFLAGS="-I/usr/local/opt/node@10/include"
```

Gonna merge myself.

PR #6654, opened 2018-11-08
Add support for different drives in Windows

pablonete:
So if code is in D: but temporary folder is in C:, don't try to find a relative path but stay on the absolute one.
@arcanis Please review and merge, this should fix test `it should correctly create resolution mounting points when using the link protocol`, I hope it doesn't break others.


pablonete:
@arcanis I believe we don't need to change normalizePath, it detects absolute correctly with any slashes in Windows:
```
> path.isAbsolute("x:\\hola")
true
> path.isAbsolute("x:/hola")
true
```
This is using `path.win32`, while using `path.posix` none of them would be true.

I tested your proposal anyway, found a problem with using `name` as it gets rid of the file extension (so `no-deps-1.0.0` became `no-deps-1.0`). Using `base` instead of `name`, it works, but I still believe we don't need the extra complexity. Let me know if I'm missing something.

src/util/portable-script.js
```diff
@@ -36,11 +36,20 @@ async function makePortableProxyScriptUnix(
...
-  await fs.writeFile(
-    filePath,
-    `#!/bin/sh\n\n${environment}"${sourcePath}"${prependedArguments} "$@"${appendedArguments}\n`,
-  );
-  await fs.chmod(filePath, 0o755);
+
+  if (process.platform === 'win32') {
+    await fs.writeFile(
+      filePath + '.cmd',
+      `@${environment}"${sourcePath}" ${prependedArguments} ${appendedArguments} %*\r\n`,
```
pablonete: I've merged approaches from linux branch of this `if` with what cmd-shim does (that's where `%*` is coming from). But I'd need @arcanis to double check I'm not doing anything wrong here. Thanks.

arcanis:
Thanks Pablo! I'm at dotjs today, but I'll take a look asap

pablonete:
Locally I've got a 100% with my last escape fix (which I wonder why it didn't fail before):

```
Tests:       1 skipped, 132 passed, 133 total
Snapshots:   0 total
Time:        405.318s
Ran all test suites.
Done in 406.63s.
```

Looking forward to seeing results in Linux and Azure Pipelines once this is reviewed. Enjoy DotJS!

PR #6671, opened 2018-11-13
Implements optional peer dependencies

arcanis:
**Summary**

This PR implements https://github.com/yarnpkg/rfcs/pull/105 (Optional Peer Dependencies).

It uses a new `peerDependenciesMeta` field to specify which peer dependency warnings can be safely ignored.

**Test plan**

Added tests.

### DIFF
--- a/__tests__/fixtures/install/missing-opt-peer-dep/foo/package.json
+++ b/__tests__/fixtures/install/missing-opt-peer-dep/foo/package.json
@@ -2,6 +2,11 @@
   "name": "foo",
   "version": "1.0.0",
   "peerDependencies": {
-    "bar": "optional:*"
+    "bar": "*"
+  },
+  "peerDependenciesMeta": {
+    "bar": {
+      "optional": true
+    }
   }
 }
--- a/__tests__/fixtures/install/missing-opt-peer-dep/foo/package.json
+++ b/__tests__/fixtures/install/missing-opt-peer-dep/foo/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "peerDependencies": {
+    "bar": "optional:*"
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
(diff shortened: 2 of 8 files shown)


__tests__/fixtures/install/missing-opt-peer-dep/foo/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "peerDependencies": {
+    "bar": "optional:*"
```
ljharb: Again, this will break downstream consumers and people using older versions of yarn, and any version of npm. It will mean that every airbnb library can not depend on any library using this syntax. I strongly encourage you to reconsider this hugely ecosystem-breaking approach.

arcanis: Again, that's not the case. Please refer to [my chart](https://www.npmjs.com/package/semver) that explains how all package managers will support this syntax with a graceful degradation. If you have additional questions, feel free to ping me for a direct meeting, I think I've explained as much as I could by text.

> and any version of npm

Then I suggest you push them to collaborate. I can't do much more than I already did, and if they wish to ghost us there's little I personally can do to stop them. I agree it's damaging to the ecosystem, but maybe it should have been considered before giving them the keys to the city.

ljharb: For install, yes. But `npm ls` will break.

ljharb: “Doing nothing” is far better than breaking things; so I’m not sure why you have to proceed rapidly when there’s no better option yet available.

arcanis: Fixed by 7d61b4f800fb4150078b0726a34b4e3c09670dfe

ljharb:
This is merged before the rfc itself is merged?

arcanis:
I wanted to leave an extra chance to the npm folks to participate before officially merging the rfc. Whatever happens, it will be merged for the next minor release (eta mid December).

felixfbecker:
Is there a CLI flag to ignore failures because of `peerDependenciesMeta` config, like `--ignore-engines`?

PR #6673, opened 2018-11-13
Implements "yarn self set"

arcanis:
**Summary**

This feature implements a new command:

```
$> yarn self set [range]
# Examples:
$> yarn self set nightly
$> yarn self set rc
$> yarn self set ^1.12
$> yarn self set 1.12.3
```

When run, Yarn will download from Github the latest release matching the specified range, will put it inside the `<lockfileFolder>/.yarn/releases` directory, and will update `<lockfileFolder>/.yarnrc` to point to this new binary.

Essentially, this command allows:

- to upgrade Yarn in a single swoop
- to rely on npm much less (which always was a bit funny)
- to enforce a specific Yarn version for a single project

This last point is especially important: we've been recommending people to use the same Yarn release across a team, but they usually had to discover by themselves that the `yarn-path` settings existed (despite it being one of the selling points in term of build reproducibility). Once they did they loved it, but it wasn't an easy find.

**Test plan**

I'm not too sure how to test this automatically, but I've tested it manually and it works just fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Implements `yarn self set [range]`. Check [the documentation]() for usage & tips.
+
+  [#6673](https://github.com/yarnpkg/yarn/pull/6673) - [**Maël Nison**](https://twitter.com/arcanis)
+
 ## 1.12.3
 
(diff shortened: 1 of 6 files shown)


arcanis:
I'm considering renaming this command to `yarn policies set-version` in order to implement other such "auto-configure" policies in the future (for example `yarn policies inline-cache` to automatically configure the cache to be local to the project, etc).

PR #6680, opened 2018-11-14
Increase Windows timeouts and better name Test results

pablonete:
Some tests in Windows may randomly take up to 5s, increasing to 10s.
Better name test results with its source job.

@arcanis Can you integrate this in your PR?
Let me know if you see any other blocker.
We're also looking into fastening up these tests,


arcanis:
Thanks!

PR #6708, opened 2018-11-20
fix(create): allow to use packages with versions

robin-drexler:
**Summary**

When you use `yarn create` with package name and version e.g.`yarn create hops-app@next`
yarn create would look for a binary called `create-hops-app@next` after installing. It should search for `create-hops-app` instead. 

This commit fixes it so that the binary will always be looked for without the version.

Tbh, I'm not really sure what `path` and `full` are good for in `coerceCreatePackageName` and `parsePackageName` as there does not seem to be any usage.

**Test plan**

Added some tests and ran the following code with yarn aliased to my local version:

```
yarn create hops-app@next --hops-version next --template hops-template-react@next my-awesome-project
```

Will allow to remove workarounds like: https://github.com/xing/hops/pull/708

I'll update the Changelog once this PR looks good to anybody. Thanks a lot! :)


robin-drexler:
@arcanis `normalizePattern` looks exactly like what one wants to use in that situation. Thanks for the hint!  I've rebased my commit and used it instead.

robin-drexler:
That beind said, I wonder if `normalizePattern` should/could be extended so that it also includes the scope and the name without the scope? Not sure if that'd be useful in other instances than in `yarn create`, though.

robin-drexler:
@arcanis could you please take another look? :)

robin-drexler:
Hey, is there anything I can do to get this PR landed?

freeatnet:
@arcanis I'm interested in this change. Is there anything that needs to be done to land it?

arcanis:
@freeatnet The v1 is feature-freezed as far as I'm concerned; further development will need to occur on the v2 trunk.

Issue #1139, opened 2016-10-17
yarn self-update should update using the same installation method originally used


Daniel15:
**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behavior?**
`yarn self-update` always downloads a tarball.

**What is the expected behavior?**
`yarn self-update` should update using the same installation method used by the user. For example, if Yarn was installed via the Debian/Ubuntu package, `yarn self-update` should run `sudo apt-get update && sudo apt-get install yarn`. On Windows when installed via the installer, it should download `https://yarnpkg.com/latest.msi` and execute it.

This is closely related to #1138 ("add yarn version check on startup") in that both issues need some way of  determining the installation method used by the user.

busches:
Do you want to close this since `self-update` was removed with https://github.com/yarnpkg/yarn/pull/1429?

Daniel15:
No, the plan is to reinstate a version of self-update that uses the right
installation method.

Sent from my phone.

On Dec 7, 2016 2:04 AM, "Scott Busche" <notifications@github.com> wrote:

Do you want to close this since self-update was removed with #1429
<https://github.com/yarnpkg/yarn/pull/1429>?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/yarnpkg/yarn/issues/1139#issuecomment-265172525>, or mute
the thread

jakearchibald:
https://yarnpkg.com/en/docs/cli/self-update should probably be removed/updated

meta-meta:
If I'm on 0.18.1, how do I update to the next version?

Daniel15:
@meta-meta Currently, you need to manually upgrade.

octopitus:
how about `yarn global add yarn`?

dancoates:
I'm a big fan of Yarn but am a little concerned at how self-update was removed with no deprecation notice, no [mention in the release notes](https://github.com/yarnpkg/yarn/releases), and that it was bundled in with a [PR](https://github.com/yarnpkg/yarn/pull/1429) that is largely unrelated and where there was no discussion about removing self-update.

I understand that it is early days for Yarn and that API changes are to be expected but it would be good if there was a process for removing core functionality. 

(sorry if this isn't the right place to discuss, but this PR is where the [docs](https://yarnpkg.com/en/docs/cli/self-update) link to)

Daniel15:
> how about yarn global add yarn?

[...]

PR #6736, opened 2018-11-29
Add self update

rafael-vitor:
**Summary**

closes: #1139 
Putting back `yarn self-update` as we now have a way of knowing which installation method was used and how to properly update. 

**Test plan**

As this is my first contribution to this project I'd really appreciate some feedback before going into implementing tests. I'll add them asap.  

I'm looking forward for some feedback on this! Thanks!

### DIFF
--- a/src/cli/commands/self-update.js
+++ b/src/cli/commands/self-update.js
@@ -1,0 +1,63 @@
+/* @flow */
+
+import semver from 'semver';
+import type Config from '../../config.js';
+import {SELF_UPDATE_VERSION_URL} from '../../constants.js';
+import type {Reporter} from '../../reporters/index.js';
+import {getInstallationMethod} from '../../util/yarn-version.js';
+import {getUpdateCommand} from './install';
+import {execCommand} from '../../util/execute-lifecycle-script.js';
+
+export const noArguments = true;
+export const requireLockfile = false;
+
+export function setFlags(commander: Object) {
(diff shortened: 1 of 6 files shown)


AlexWayfer:
We have `upgrade` command for packages, not `update`, and probably should have `self-upgrade` instead of `self-update`. Also, in `apt` there are `update` for repositories refresh and `upgrade` for packages update.

arcanis:
I believe this feature conflicts in spirit with #6673, which is already in master

rafael-vitor:
@arcanis I see. Many ppl still refer to #1139 to try and find the way to properly upgrade yarn as the docs still mention it. 

Do you think there's room for both commands? Maybe self-upgrade is worth keeping for the clarity of its name. Maybe even as an alias to `yarn policies set-version`. 

Otherwise, we should work on making the community aware of this by linking to it on the current `self-update` doc and/or removing it eventually.

src/cli/commands/self-update.js
```diff
@@ -0,0 +1,63 @@
...
+    await execCommand({
+      // Could not figure out to which stage should this belong to.
+      stage: 'self-update',
+      config,
+      cmd: command,
+      cwd: config.cwd,
+      isInteractive: true,
+    });
+  } else {
+    reporter.info('could not find command for instalation method');
```
bf4: typo

PR #6738, opened 2018-11-29
Find root workspace when the pattern ends with a slash

davazp:
**Summary**
When executing some commands in a package that is part of a workspace, `yarn` will fail to find the root workspace if the pattern ends with `/`, like in:

```
    workspaces: [
      "package-b/"
    ]
```

This is confusing because yarn will still list the package if you run `yarn workspaces info`.

I have implemented a change so the relative path of the package, with and without slash, will be matched against the pattern in the package.json.

**Test plan**

I have included a new fixture and a test to the config module to test this.


davazp:
Is there any interest or feedback on this?

PR #6747, opened 2018-11-30
Fixes how signals are forwarded

arcanis:
**Summary**

Since the portable script doesn't use `exec`, signals seem to be trapped.

This diff ensures that we correctly replace the process rather than simply spawning a new one.

**Test plan**

CI

### DIFF
--- a/src/util/portable-script.js
+++ b/src/util/portable-script.js
@@ -45,7 +45,7 @@ async function makePortableProxyScriptUnix(
   } else {
     await fs.writeFile(
       filePath,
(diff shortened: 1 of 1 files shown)


zenios:
I was about to make PR. 
I used the following test script in order to validate that everything is OK

```js
#!/usr/bin/env node

setTimeout(() => {
  console.log("UNTRAPPED");
  process.exit()
}, 10000);

process.on('SIGINT', () => {
  console.log("TRAPPED");
  process.exit();
});
```

PR #6756, opened 2018-12-03
Add custom headers to tarball fetcher

Glavin001:
**Summary**

My team would like to download tarballs attached to GitLab Releases.
The projects are authenticated, therefore we need to add `PRIVATE-TOKEN` header to the request for the tarball. See https://docs.gitlab.com/ee/api/README.html#personal-access-tokens

I propose the following configuration addition:

`~/.npmrc`:

```
//custom.gitlab.host/:_header:PRIVATE-TOKEN=REDACTED
```

where:
- `//custom.gitlab.host/` is the host which matches the tarball request.
- `:_header:<HEADER-NAME>` indicates a custom header should be added, with `PRIVATE-TOKEN` being the `HEADER-NAME` in this instance. Multiple custom headers could be added in this way.

Then we can run:

```
❯ yarn-local add --dev --verbose https://custom.gitlab.host/OWNER/REPO/uploads/REDACTED/PKG-NAME-v1.0.0.tgz
```

And the request will be authenticated.

Once I have approval on the above format then I will update the `CHANGELOG.md` 😃 .

**Test plan**

Once I have approval on the above format then I will add tests 😃 .

----

**Update 1**:

Now `~/.yarnrc` looks like:

```
"//custom.gitlab.host/:_header:PRIVATE-TOKEN" "REDACTED"
```

after running

```
yarn config set "//custom.gitlab.host/:_header:PRIVATE-TOKEN" "REDACTED"
```
and `~/.npmrc` is not used. Working great! 🎉

---

**Update 2**:
See https://github.com/yarnpkg/yarn/pull/6756#issuecomment-444728156

I created a release and uploaded a tarball generated from `yarn pack` to https://github.com/Glavin001/yarn/releases/tag/1.13.0-config-tarball-fetcher-headers-1

You can install this custom Yarn with:

```bash
npm install -g https://github.com/Glavin001/yarn/releases/download/1.13.0-config-tarball-fetcher-headers-1/yarn-v1.13.0-0.tgz
```

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -277,7 +277,7 @@ export default class TarballFetcher extends BaseFetcher {
   }
 
   requestHeaders(): {[string]: string} {
-    const registry = this.config.registries[this.registry];
+    const registry = this.config.registries.yarn;
     const config = registry.config;
     const requestParts = urlParts(this.reference);
     return Object.keys(config).reduce((headers, option) => {
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -229,7 +229,7 @@ export default class TarballFetcher extends BaseFetcher {
     const registry = this.config.registries[this.registry];
 
     try {
-      const headers = this.requestHeaders(this.reference);
+      const headers = this.requestHeaders();
       return await registry.request(
         this.reference,
         {
@@ -276,10 +276,10 @@ export default class TarballFetcher extends BaseFetcher {
     }
   }
 
-  requestHeaders(requestUrl: string): {[string]: string} {
+  requestHeaders(): {[string]: string} {
     const registry = this.config.registries[this.registry];
     const config = registry.config;
-    const requestParts = urlParts(requestUrl);
+    const requestParts = urlParts(this.reference);
     return Object.keys(config).reduce((headers, option) => {
       const parts = option.split(':');
       if (parts.length === 3 && parts[1] === '_header') {
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -6,6 +6,7 @@ import * as constants from '../constants.js';
 import BaseFetcher from './base-fetcher.js';
 import * as fsUtil from '../util/fs.js';
 import {removePrefix} from '../util/misc.js';
+import normalizeUrl from 'normalize-url';
 
 const crypto = require('crypto');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


src/fetchers/tarball-fetcher.js
```diff
@@ -329,3 +350,16 @@ export class LocalTarballFetcher extends TarballFetcher {
     return this.fetchFromLocal(this.reference);
   }
 }
+
+type UrlParts = {
```
Glavin001: This is copied from https://github.com/yarnpkg/yarn/blob/b8a565e562810f9bf86f3ac0531ad348ed7e9d0a/src/registries/npm-registry.js#L77-L88

I would like to put it in a more central location. I ask the Yarn team where would the most appropriate destination? 😄

src/fetchers/tarball-fetcher.js
```diff
@@ -273,6 +276,24 @@ export default class TarballFetcher extends BaseFetcher {
     }
   }
 
+  requestHeaders(requestUrl: string): {[string]: string} {
```
Glavin001: I based this method off of https://github.com/yarnpkg/yarn/blob/b8a565e562810f9bf86f3ac0531ad348ed7e9d0a/src/registries/npm-registry.js#L200-L213

Although, looking at it now I think this `requestHeaders` does not need a `requestUrl` argument and instead could use `this.reference` directly.

arcanis:
Given that this affect the syntax of the `.npmrc`, is it also available on npm? Or would it be a Yarn-exclusive feature? In the later case I'm a bit concerned of risks of confusion 🤔

Glavin001:
@arcanis : Good point and I agree. This feature is *not* available on npm - I reviewed https://docs.npmjs.com/misc/config#config-settings and hopefully did not miss a similar supported feature.

I'm currently retrieving the config values with https://github.com/yarnpkg/yarn/blob/ed76774f65fa674ff3cdf3de9c6ac2de407b4235/src/fetchers/tarball-fetcher.js#L279-L281 and it appeared to only load from `.npmrc`. I'm not yet sure how to load from `.yarnrc` config file.

Any advice on how I could load config from `.yarnrc` instead of `.npmrc` would be greatly appreciated. Thank you in advance!

arcanis:
I think you should be able to access the `yarnrc` settings through `this.config.registries.yarn` (`this.registry` is typically npm - even though it goes through the Yarn domain name).

Glavin001:
Thank you for the assistance, @arcanis !

I pushed a commit switching from npm to yarn config.

Now `~/.yarnrc` looks like:

```
"//custom.gitlab.host/:_header:PRIVATE-TOKEN" "REDACTED"
```

after running

```bash
yarn config set "//custom.gitlab.host/:_header:PRIVATE-TOKEN" "REDACTED"
```

and `~/.npmrc` is not used. Working great! 🎉 

---

Once the above is all approved, I will add tests, push the changes, and update the Pull Request description to reflect these changes.

Thanks again for reviewing!

Glavin001:
It looks like the Jest tests need a higher timeout:

> Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

They fail intermittently and without related code changes.

arcanis:
Thanks! It'll be part of the next release 😃

PR #6816, opened 2018-12-14
Implements the pnp hook for the `resolve` package

arcanis:
**Summary**

This PR adds builtin `resolve` support using the mechanisms described in https://github.com/browserify/resolve/pull/174. I've also added an escape hatch (`forceNodeResolution`) in case something goes awry.

cc @ljharb

**Test plan**

Added an automated test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6712](https://github.com/yarnpkg/yarn/pull/6712) - [**Maël Nison**](https://twitter.com/arcanis)
 
+- Adds transparent support for the [`resolve`](https://github.com/browserify/resolve) package when using Plug'n'Play
+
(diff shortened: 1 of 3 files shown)


arcanis:
Until the `resolve` consumer packages (such as `liftoff`, used by Gulp) fully upgraded to the later version, the recommended workaround is to force the use of the modern version through the `resolutions` field:

```json
{
  "resolutions": {
    "resolve": "npm:resolve@x.y.z"
  }
}
```

imsnif:
Just FYI: this also fixes an issue with specifying plugin names in a babelrc file (in my case it was .babelrc.js). 

The issue was due to the same resolve package being used here: https://github.com/babel/babel/blob/e85faec47d5d3ef940b7a85d48fa24e6e1cc32ab/packages/babel-core/src/config/files/plugins.js#L99

After this fix it works great and you do not have to specify require.resolve for your plugins/presets.

ljharb:
resolve v1.9 was released last night, fwiw

Issue #6748, opened 2018-11-30
yarn audit exit code overflow


MRoci:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
`yarn audit` set the `exit code` related to the number of issues found.
If the number of issues is greater than 255 the exit code overflows
This means that if are found exactly 256 issues the exit code is 0.

**If the current behavior is a bug, please provide the steps to reproduce.**
This is a package.json crafted to trigger exactly 256 issues

https://gist.github.com/MRoci/e28694334ff6771cc7fd14e6d762f355

**What is the expected behavior?**
The expected behaviour should be an exit status always different from 0 in case of failure

**Please mention your node.js, yarn and operating system version.**
yarn version `v1.12.3`
node version: any that i know
OS: `OSX Mojave 10.14.x`

PR #6819, opened 2018-12-14
Return semantic exit code for audit

antrew:
**Summary**

Resolve https://github.com/yarnpkg/yarn/issues/6748

`yarn audit` currently returns the number of found vulnerabilities. This can lead to an overflow and a false negative. E.g., if the number of vulnerable dependencies is equal to 256, the exit code will be equal 0. Furthermore, it is not very practical, as this number does not indicate the severity level of found vulnerabilities.

This pull request modifies this behavior by returning a number that is a sum of found severities:
* 1 for INFO
* 2 for LOW
* 4 for MODERATE
* 8 for HIGH
* 16 for CRITICAL

That is, if only INFO and MODERATE vulnerabilities were found, then the exit code will be `1+4 = 5`.

With this change it would be possible to check for particular severity levels. E.g., if someone wants to know, if there are HIGH and CRITICAL dependencies, then one could check for exit code >= 8.

**Test plan**

Current behavior:
```
$ yarn audit
9 vulnerabilities found - Packages audited: 31177
Severity: 9 Low
Done in 1.94s.
$ echo $? 
9
```

New behavior:
```
$ yarn-local audit
...
9 vulnerabilities found - Packages audited: 31177
Severity: 9 Low
Done in 1.69s.
$ echo $?
2
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6712](https://github.com/yarnpkg/yarn/pull/6712) - [**Maël Nison**](https://twitter.com/arcanis)
 
+- Fixes yarn audit exit code overflow
+
+  [#6748](https://github.com/yarnpkg/yarn/issues/6748) - [**Andrey Vetlugin**](https://github.com/antrew)
+
 ## 1.12.3
 
 **Important:** This release contains a cache bump. It will cause the very first install following the upgrade to take slightly more time, especially if you don't use the [Offline Mirror](https://yarnpkg.com/blog/2016/11/24/offline-mirror/) feature. After that everything will be back to normal.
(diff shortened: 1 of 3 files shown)


arcanis:
Technically it should be a major bump, but I wouldn't be opposed to put that into the next minor. What do you think, @rally25rs?

antrew:
Hi @arcanis,

I don't think anybody relies on the current implementation, since it doesn't really matter if you have 1, 10 or 99 vulnerabilities. In my opinion, the only thing that makes sense with the current implementation, is to check if the exit status is zero (no vulnerabilities), or non-zero (there are vulnerabilities).
In that sense the proposed implementation is backwards compatible.

Anyway, it seems like everyone is building some wrappers around the current implementation in their CI pipelines to find out if they have vulnerabilities of certain levels. Ours looks like this at the moment and has to completely ignore the exit code of `yarn audit`:
```
yarn audit || true                                  
if [[ `yarn audit | grep -oiP '(critical|high)'` ]] ; then     
  echo 'Critical vulnerabilities found'             
  exit 1                                            
else                                                
  exit 0                                            
fi                                                  
```

PR #6820, opened 2018-12-14
Stops unplugging packages when the install scripts are disabled

arcanis:
**Summary**

We currently unplug packages by default when they have postinstall scripts, since we can't run them from within the cache. This approach is fine, but we're a bit overzealous and we also do this when the user explicitly disabled the scripts.

This diff ensures that we don't automatically unplug packages anymore when the install scripts are disabled. Of note: packages that have already been unplugged won't be affected by this change (so for example, if you do `yarn install && yarn install --ignore-scripts`, the packages will *still* be unplugged).

**Test plan**

Should be simple enough not to be worth a test?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6712](https://github.com/yarnpkg/yarn/pull/6712) - [**Maël Nison**](https://twitter.com/arcanis)
 
+- Stops automatically unplugging packages with postinstall script when running under `--ignore-scripts`
+
+  [#6820](https://github.com/yarnpkg/yarn/pull/6820) - [**Maël Nison**](https://twitter.com/arcanis)
+
 ## 1.12.3
 
 **Important:** This release contains a cache bump. It will cause the very first install following the upgrade to take slightly more time, especially if you don't use the [Offline Mirror](https://yarnpkg.com/blog/2016/11/24/offline-mirror/) feature. After that everything will be back to normal.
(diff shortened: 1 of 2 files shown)


edmorley:
Hi!

Trying with Yarn 1.13.0 (which includes this change) this works when using `--ignore-scripts` however does not when using `ignore-scripts true` in a  `.yarnrc` in the repository root. Is this expected?

arcanis:
Nope, not expected 😮 Can you try adding `--ignore-scripts true` in your `.yarnrc` and see if it has an effect? I don't see why the two would be any different, but ...

edmorley:
Using `--ignore-scripts true` in the `.yarnrc` worked.

In all cases (CLI flag and both forms of the `.yarnrc` option) the console output included:
`warning Ignored scripts due to flag.`

...but I'm presuming that's because the option is retrieved via flags instead?
https://github.com/yarnpkg/yarn/blob/fcf6298d8feb10d6d92d17bf4f5d416c33716faf/src/cli/commands/install.js#L688-L689

arcanis:
Well, `ignore-scripts true` should not have any effect whatsoever - the settings can only be set from the command line afaik (and the yarnrc, but only through the command-line syntax). Compare here:

https://github.com/yarnpkg/yarn/blob/eb2b565bb9b948e87b11119482ebc184a9d66141/src/config.js#L399-L400

Versus here (which also accepts the no-cli rc defintition):

https://github.com/yarnpkg/yarn/blob/eb2b565bb9b948e87b11119482ebc184a9d66141/src/config.js#L310-L312

edmorley:
> Well, `ignore-scripts true` should not have any effect whatsoever

It does affect whether the scripts are run, see:


> In all cases (CLI flag and both forms of the `.yarnrc` option) the console output included:
> `warning Ignored scripts due to flag.`

edmorley:
I've broken out the issue above to #6893.

PR #6853, opened 2018-12-25
Fix version-sign-git-tag default and npm map

xPaw:
The code that actually uses this param is here:

https://github.com/yarnpkg/yarn/blob/eb2b565bb9b948e87b11119482ebc184a9d66141/src/cli/commands/version.js#L171


arcanis:
That might technically be a breaking changes :'(

xPaw:
What would be the approach here then?

Issue #6805, opened 2018-12-10
yarn add package not found


kdong007:
yarn v1.12.3
node v8.10.0
macOS v10.13.5

`yarn add event-stream` gives me 
```sh
error https://registry.yarnpkg.com/event-stream/-/event-stream-3.3.6.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "Unexpected end of data"
```

I opened up this URL i browser, it says 
```json
{
"error": "Not found"
}
```

rally25rs:
There is no `event-stream` version 3.3.6 (it was probably unpublished)

https://www.npmjs.com/package/event-stream (and switch to "Versions" tab)

You could `yarn upgrade event-stream@3.3.5` to move back to the latest 3.x version, or `yarn upgrade event-stream --latest` to upgrade to the latest major version.

PR #6868, opened 2019-01-02
Show problem path in invariant message

CGamesPlay:
**Summary**

Sometimes, `yarn add` will fail unexpectedly, and it leaves the cache in an invalid state. To cause your yarn cache to become corrupted:

```
$ yarn add antd@3.11.6
yarn add v1.12.3
[2/4] 🚚  Fetching packages...
error https://registry.yarnpkg.com/event-stream/-/event-stream-3.3.6.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "Unexpected end of data"

$ yarn cache list
yarn cache v1.12.3
error An unexpected error occurred: "There should only be one folder in a package cache (got  )".
info If you think this is a bug, please open a bug report with the information provided in "/Users/rpatterson/Projects/chess2/www2/client/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/cache for documentation about this command.
```

Note that the error message provides no information about how to resolve the error. This commit adds the problem path to the invariant message. Deleting this problem path will allow `yarn cache list` to work properly again.

The root cause of this issue is #6805 which has been closed. This commit does not resolve the issue, so that issue should likely be reopened.

**Test plan**

Simply running `yarn cache list` while the cache is in an invalid state as described above produces a more useful error message.

### DIFF
--- a/src/cli/commands/cache.js
+++ b/src/cli/commands/cache.js
@@ -31,7 +31,7 @@ export async function getCachedPackagesDirs(config: Config, currentPath: string)
     const candidates = await fs.readdir(packageParentPath);
     invariant(
       candidates.length === 1,
-      `There should only be one folder in a package cache (got ${candidates.join(',')})`,
+      `There should only be one folder in a package cache (got ${candidates.join(',')} in ${packageParentPath})`,
     );
 
     for (const candidate of candidates) {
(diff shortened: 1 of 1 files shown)


CGamesPlay:
Update: apparently my repro steps are invalid in a clean repository, so one of my other dependencies, `pstree@1.1.0`, was dependent on event-stream. However, `pstree@1.1.0` seems to have been unpublished so the repro doesn't work. Still, the patch is seems useful.

arcanis:
Looks good to me, thanks 🙂

Issue #3911, opened 2017-07-11
Workspaces don't run lifecycle scripts when linking local packages


jquense:
**What is the current behavior?**

When you do a yarn install in a workspaces enabled project, yarn doesn't run the various lifecycle scripts for workspace packages, e.g. `prepare` `prepublish` `preinstall` etc

**If the current behavior is a bug, please provide the steps to reproduce.**

Here is a commit adding a failing test case: https://github.com/jquense/yarn/commit/61eaf4239f707f772af5713ebc98963589a6dad4

**What is the expected behavior?**

Lifecycle scripts are run for the local packages so they have an opportunity to build steps. This is what [lerna](https://github.com/lerna/lerna/blob/master/src/commands/BootstrapCommand.js#L100-L111) does.

**Please mention your node.js, yarn and operating system version.**

yarn 27.5
node 8
Mac Sierra

cc @bestander

bestander:
PR is welcome BTW :)

jquense:
I took a stab at it, but could figure out the code base unfortunately. All I can see is that `scripts` is empty for the linked packages when it gets to the `getInstallCommands` in the pacakge-install-scripts :/

bestander:
Thanks for giving it a try, @jquense.
The entry point is probably here https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L477 - flattenedTopLevelPatterns only contains the top level dependencies of root package.json + a virtual dependency that refers all workspaces.
I think going deeper PackageInstallScripts should handle that https://github.com/yarnpkg/yarn/blob/master/src/package-install-scripts.js#L248.

If you want you can start with submitting a PR with a failing test and we can sort it out in a follow up.

jquense:
The failing test case linked in the issue is the best I can do sorry! Unfortunately I just don't have the time to spend hours learning the yarn codebase to try and fix this.

BurtHarris:
Hi.   I'm quite interested in yarn and workspaces in particular.   I'll be giving a look at this as a possible first contribution.

bestander:
Awesome!
If you have any questions ask away

BurtHarris:
@bestander I hit a snag right away, Issue #4059.   Its almost certainly related to the fact I'm running Windows any hints on how to make that work?

bestander:
As a workaround you should be able to run tests like this `node ./node_modules/jest/bin/jest.js`

jquense:
[...]

PR #6869, opened 2019-01-03
Trigger lifecycle scripts for workspaces

tomasz-sodzawiczny:
Fixes https://github.com/yarnpkg/yarn/issues/3911

All lifecycle scripts run during install (`preinstall`, `install`, `postinstall`, `prepublish` & `prepare`) are now run on the main directory and all workspaces.


arcanis:
I'm not too sure about this one ... I think there's a conceptual issue with regard to peer dependencies.

Let's say you have the following:

- Workspace `A` has a peer dependency on `B`
- Workspace `C` depends on `A` and `B@1`
- Workspace `D` depends on `B` and `B@2`

What's the correct action? The way I see it, they are:

1. The package manager might call the install script twice
2. The package manager might call the install script once
3. The package manager might never call the install script for workspaces
4. The package manager might never call the install script for workspaces with peer dependencies

Both 1 and 2 are quite surprising and probably wrong. 3 is surprising as well, but has an entirely predictable behavior, which I would count as an improvement over the two others. The 4 is similar, but only affects workspaces that can trip on this behavior. While it would seems better at first glance, it makes the behavior confusing: from the end user perspective, I'm not sure they would understand why the install scripts run in a case but not every case.

I'd personally tend to pick option 3, but I clearly understand the use case so I'd be happy if we could find an intuitive way to make it work that I haven't think of!

tomasz-sodzawiczny:
@arcanis  I am not sure I understand why I would ever run the scripts more than once - the scripts are run when the `install` script is run *on the package itself*, not when the package is *installed as a dependency.*

What am I missing? Can you clarify that?

---

I realise now I missed something else - **the order of running the scripts matters**: `install`, `postinstall`, `prepublish` and `prepare` are run with the assumption that my dependencies are already in place (in my `node_modules/`).

So when we have packages:
* A → Z
* B → Z
* Z

The flow using the registry would be:
1. Build and publish Z (`Z.prepare` runs here - once)
2. Install A - download *prepared* Z into node_modules, run `A.prepare`
3. Install B - download *prepared* Z into node_modules, run `B.prepare`

So I believe with workspaces we should run:
1. `Z.prepare`
2. `A.prepare`, `B.prepare` (in any order)

The only case left is when we have circular dependencies - but maybe in this case throwing a warning and not doing anything is a valid solution.

Does that make any sense?

arcanis:
> I am not sure I understand why I would ever run the scripts more than once - the scripts are run when the install script is run on the package itself, not when the package is installed as a dependency.

It's not so simple. Imagine that in my previous example B is actually `babel`, and A uses the peer dependency in order to build itself. In this situation, C expects A to have been built against `babel@1` while D expects A to have been built against `babel@2`. If you only run the install script once, then one of those affirmation will be false.

Of course it's a contrived example, but basically: the peer dependencies might be used in order to build the package. Since different peer dependencies can be provided to the same package, it follows that the package might need to be built against each set of dependencies.

tomasz-sodzawiczny:
> the peer dependencies might be used in order to build the package.

Can they? We're talking about the scripts that run **before** the package is installed or even published - aren't the `peerDependencies` not present then by definition?

arcanis:
That's true for `preinstall` but not for `prepare` or the others, which are run after the install. Even if to be fair the npm lifecycle hooks are a huge mess and it's not clear what's semantically correct and what isn't, even after reviewing the closest thing to a [documentation](https://docs.npmjs.com/misc/scripts) they have.

tomasz-sodzawiczny:
Oh wow, thanks for pointing that out - you're right, `install` and `postinstall` are affected by the peer deps issue.

I did a quick check though, and it seems `prepare` & `prepublish` do actually behave as documented - they don't run when installed as a dependency, only on local `yarn install` and `yarn publish`.
You can try it out with this package: `install-scripts-test-package@1.1.1` (have a look inside the package.json to see what it's doing).

So - we can do nothing about `install`/`postinstall`, but I think we can still run the others if we keep the right order. This should be "good enough" to solve the issue for most people, as it's mostly about the `prepare`/`prepublish` scripts.

NathanielHill:
Can we restrict this to `prepare` and `prepublish` and get this merged?

tomasz-sodzawiczny:
@NathanielHill Sorry for not following up on this yet - I think `prepare`/`prepublish`still need to get fired in topological order (dependencies first) and I didn't find time yet to dig into that.

Do you know if there is some easy way to do this? I'm guessing this order is already computed somewhere in `yarn`, but I don't know the codebase well enough.

NathanielHill:
Haven't looked at yarn codebase at all. I'm currently solving this by explicitly calling the `prepare` scripts that I need from the root `prepare` script.

tomasz-sodzawiczny:
@arcanis I updated the PR. 

Now the behaviour of `yarn install` is as follows:
* `preinstall` script is run on all workspaces (no specific order)
* `prepare` and`prepublish` scripts are run on all workspaces in topological order
* `install` and `postinstall` are _not_ run on workspaces - due to undefined behaviour with peer deps

Do you think this is a valid solution now?
Do you have any suggestions for the code structure or some additional test to be added?

kingdaro:
Gentle ping on this, still finding myself needing this every so often. Is there anything else needed for this to be merged?

Gxmadix:
Hello are there news about merging this PR?

PR #6871, opened 2019-01-03
Fix generated .pnp.js compatibility with Node 6.

rwjblue:
**Summary**

Prior to these changes the generated `.pnp.js` file would use trailing commas for function invocations which is not allowed (and generates a parse error) under Node 6.

**Test plan**

I would like to add _some_ CI testing of the generated `.pnp.js` files to ensure that Node 4 and Node 6 continue to work (and do not regress) until the project officially drops support for them, but I am not quite sure how to go about it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -37,6 +37,20 @@
       "rules": {
         "no-console": "off"
       }
+    },
+    {
+      "files": [
+        "src/util/generate-pnp-map-api.tpl.js"
+      ],
+      "rules": {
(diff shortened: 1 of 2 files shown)


arcanis:
Thanks!

Regarding the tests, the simplest would likely be to make the acceptance tests work on Node 6. It could be kept in a separate branch and be periodically updated with the new master. This way we wouldn't have to downgrade the codebase and would still benefit from an automated testing.

PR #6890, opened 2019-01-09
Add the string '@generated' to the lockfile

mwgamble:
**Summary**

Various tools such as linters look for the string `'@generated'` in a file and use this to skip processing of the file. This change will help prevent linters from targeting the yarn.lock lockfile, as there is no benefit to performing things such as line length checks on this file.

**Test plan**

On the next update to the lock file, you should observe a new comment line near the top of the `yarn.lock` file. It should look like the following:

    # @generated


arcanis:
I would be wary of this change because it'll also cause UIs not to display the file diffs. I don't want to make code reviews harder (related, https://github.com/github/linguist/issues/4348).

arcanis:
I'm sorry I won't merge it - I don't think the pros out-weight the cons, so I prefer to keep things the same. Thanks for the contribution, though, I look forward to see other PRs if you feel like it 🙂

Issue #6905, opened 2019-01-10
GitHub API rate limits


billyvg:
**Do you want to request a *feature* or report a *bug*?**
Both?

**What is the current behavior?**
We are hitting GitHub API rate limits

```
error An unexpected error occurred: "https://api.github.com/repos/yarnpkg/yarn/releases: API rate limit exceeded for ***. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)".
```

(this happens when trying to use the new `yarn policies set-version` command)

**What is the expected behavior?**
We should be able to define a github access token via an environment variable so that github requests are authd

PR #6912, opened 2019-01-12
feat(policies): Use github access token when requesting releases

billyvg:
**Summary**

When using `yarn policies set-version`, yarn makes an API request to GitHub to fetch a list of yarn releases. This request can be [rate limited](https://developer.github.com/v3/#rate-limiting), but can be bypassed if request is [authenticated](https://developer.github.com/v3/#basic-authentication)

If the environment variable `GITHUB_TOKEN` is defined, append `?access_token=<GITHUB_TOKEN>` when requesting yarn releases from GitHub.

Closes https://github.com/yarnpkg/yarn/issues/6905

**Test plan**


arcanis:
Awesome, thanks!

sepehr:
Might worth noting: 
https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

PR #6926, opened 2019-01-15
fix(run): Preserve none zero exit code

### LINKED ISSUE #6925: Preserve none zero exit code
I want to request a *feature*

### Preserve none zero exit code.

I'm trying to create a Github Action.

https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#exit-codes-and-statuses

To filter actions, it needs `78` as exit code:

`index.js`
```js
const process = require('process');
process.exit(78)
```

`npm start` will yield 78, but `yarn start` will yield 1.

screen shot 2019-01-15 at 8 44 31 pm

Is there an option that already does this?

zhigang1992:
resolve #6925

**Summary**

`index.js`
```js
const process = require('process');
process.exit(78)
```

`npm start` will yield 78, but `yarn start` will yield 1.

screen shot 2019-01-15 at 8 44 31 pm

- [x] Fix `yarn node index.js`
- [x] Fix `yarn start`

**Test plan**

- [x] Having unit test covering the exit code


src/cli/index.js
```diff
@@ -583,7 +584,11 @@ export async function main({
         reporter.info(command.getDocsInfo);
       }
 
-      return exit(1);
+      if (err instanceof ProcessTermError) {
+        return exit(err.EXIT_CODE || 1);
+      }
+
+      return exit((err.code && Number(err.code)) || 1);
```
zhigang1992: This is not pretty, but when we convert the `ProcessTermError` down to `MessageError`, we lost the `EXIT_CODE` as well.

https://github.com/yarnpkg/yarn/pull/6926/files#diff-939cc0d319704d4f0d98baf9167ba850

src/cli/index.js
```diff
@@ -583,6 +584,14 @@ export async function main({
         reporter.info(command.getDocsInfo);
       }
 
+      if (err instanceof ProcessTermError) {
+        return exit(err.EXIT_CODE || 1);
+      }
+
+      if (err instanceof MessageError) {
+        return exit((err.code && Number(err.code)) || 1);
+      }
```
arcanis: This block shouldn't be needed - `ProcessTermError` extends `MessageError`.

src/util/execute-lifecycle-script.js
```diff
@@ -366,6 +366,7 @@ export async function execCommand({
         err.EXIT_SIGNAL
           ? reporter.lang('commandFailedWithSignal', err.EXIT_SIGNAL)
           : reporter.lang('commandFailedWithCode', err.EXIT_CODE),
+        String(err.EXIT_CODE),
```
arcanis: A better fix would be to create another `ProcessTermError` with the proper text, and forward the `EXIT_SIGNAL` and `EXIT_CODE` to it. I don't want `code` to be the same as the exit code (`code` is meant to be `EPERM`, `ENOENT`, etc).

zhigang1992:
@arcanis please take another look.

PR #6930, opened 2019-01-16
Fixes Yarn invocations on Darwin when symlinked

serverwentdown:
#6568 had an attempt to fix this issue, however it does not solve the problem. 

**Summary**

**Test plan**

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -3,7 +3,7 @@ argv0=$(echo "$0" | sed -e 's,\\,/,g')
 basedir=$(dirname "$(readlink "$0" || echo "$argv0")")
 
 case "$(uname -s)" in
-  Darwin) basedir="$( cd "$( dirname "$argv0" )" && pwd )";;
+  Darwin) basedir="$(cd "$(dirname "$0")" && cd "$basedir" && pwd)";;
   Linux) basedir=$(dirname "$(readlink -f "$0" || echo "$argv0")");;
   *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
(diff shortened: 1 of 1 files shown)


bin/yarn
```diff
@@ -3,7 +3,7 @@ argv0=$(echo "$0" | sed -e 's,\\,/,g')
 basedir=$(dirname "$(readlink "$0" || echo "$argv0")")
 
 case "$(uname -s)" in
-  Darwin) basedir="$( cd "$( dirname "$argv0" )" && pwd )";;
+  Darwin) basedir="$(cd "$(dirname "$0")" && cd "$basedir" && pwd)";;
```
arcanis: This diff doesn't make sense to we. Why are you calling `cd` twice? It looks like it makes this line become a noop.

serverwentdown: This first forces the current directory to be the same as the binary (example: /usr/local/bin), which is the directory the symlink are relative to. Then, the symlink (example: ../share/yarn/bin) can be resolved and entered.

serverwentdown: This is all necessary because the macOS shell comes with no native "path.join()"

serverwentdown:
Example scenario that doesn't work right now:

Yarn in /usr/local/share/yarn
Symlink in /usr/local/bin/yarn -> ../share/yarn/bin/yarn

bin/yarn
serverwentdown: Shrinking it to `cd "$(dirname "$0")$basedir"` will break cases where the symlink is to an absolute path (example: /usr/local/bin/usr/local/share/yarn/bin/)

Issue #1957, opened 2016-11-20
`yarn unlink package` doesn't restore node_modules/package


whitecolor:
After `yarn link package` yarn creates symlink node_module/package, but after `yarn unlink package` it just removes `node_module/package` and not places there hard installed version. 

Probably expected that it should find in cache or fetch a correct version from registry and replace removed symlink with it.

whitecolor:
Seem to be the same issue https://github.com/yarnpkg/yarn/issues/937

caesarsol:
+1, same issue here!

andybluntish:
From the docs (https://yarnpkg.com/en/docs/cli/unlink#toc-yarn-unlink-package) this seems to be expected behaviour. However, as noted in #937 running `yarn` does not add the package back to node_modules.

caesarsol:
Yes, it seems `yarn check` detects the package is missing, but `yarn install` does not restore it...

I have to run `yarn install --force` to make it right.

(As a possible improvement, I think `unlink` could reinstall the original package by itself)

whitecolor:
`unlink` should reinstall.

danielepiccone:
it works with a force install, but it surprises me why it doesn't with a normal `yarn install`

stephencroberts:
Running `yarn install --check-files` seems to restore the package in v1.4.0.

remino:
This has been going on since 2016? I'm having the same issue with yarn 1.12.3.

arcanis:
I don't think `yarn unlink package` is meant to trigger an install (or at least, I'd feel uncomfortable changing this behavior anywhere else than in a major release). In the case where multiple packages are linked / unlinked (mostly scripts), running an new install every time would cause a perf hit.

That said, `yarn install` not noticing that a link disappeared is a bug. I don't have time right now to work on this, but #6931 seems an acceptable workaround for the time being (except that I'd recommend to use `--force` rather than `--check-files`, since the later has long-term effects).

isakaweski:
Still not working on 2020.
Boooooring.

dalmo3:
None of the suggestions above work for yarn v2 or v3 - most of those flags don't even exist. The only option is the going nuclear: 
```
rm -rf node_modules 
yarn
```

PR #6931, opened 2019-01-16
Fix suggested command after unlinking a package

### LINKED ISSUE #937: Symlinking packages may lead to invalid states

Bug.

An unlinked (but added) package is missing from node_modules and yarn will not install.

1. Link a random package
   
   ```
   git clone git@github.com:tj/co.git
   cd co && yarn link
   ```
   
   As a side note, yarn will not install the package's dependencies. The default npm client would do that. I'm not sure whether that's expected or not given that the linking process is slightly different already.
2. Create a new directory and use the linked package from there
   
   ```
   yarn link co
   ```
3. Add the package again but this time via `add`
   
   ```
   yarn add co
   ```
   
   The linked package is still used at this point (and that's consistent with the default client).
4. Unlink the package
   
   ```
   yarn unlink co
   ```
   
   This will remove the link from node_modules as expected.
5. Run yarn to get the added (locked) package back
   
   ```
   yarn
   yarn install v0.15.1
   warning No license field
   success Already up-to-date.
   Done in 0.13s.
   
   ls node_modules
   ```
   
   At this point, the package is not present in node_modules but yarn seems to be okay with that and won't install it. Doing `yarn add co` again _will_ install it.

The package should be installed after unlinking it and running `yarn`.

Node.js 6.7.0
Yarn 0.15.1
Ubuntu 12.04

justinsacbibit:
Fixes #937, #1957

**Summary**

After unlinking a package, yarn suggests running `yarn` to re-install the package, but it doesn't actually end up re-installing the package. `yarn install --check-files` or `yarn install --force` is required.

**Test plan**

Not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
   [#6882](https://github.com/yarnpkg/yarn/pull/6882) - [**Zoran Regvart**](https://github.com/zregvart)
 
+- Fixes the command that `yarn unlink` recommend to run as a followup (now `yarn install --force`)
+
+  [#6931](https://github.com/yarnpkg/yarn/pull/6931) - [**Justin Sacbibit**](https://github.com/justinsacbibit)
+
 ## 1.13.0
 
 - Implements a new `package.json` field: `peerDependenciesMeta`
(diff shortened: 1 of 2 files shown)


ryankeener:
I think this is a very valuable PR but I think it's more of a workaround than a "fix" for  #937 & #1957

arcanis:
Thanks 👍

Issue #6389, opened 2018-09-14
prefix "yarn workspaces run" (v1.10.0) output with package names


donaldpipowitch:
**Do you want to request a *feature* or report a *bug*?**

*feature*

**What is the current behavior?**

```
$ yarn workspaces run build
yarn workspaces v1.10.0
yarn run v1.10.0
$ tsc -p src
✨  Done in 1.49s.
yarn run v1.10.0
$ tsc -p src
✨  Done in 1.21s.
yarn run v1.10.0
$ tsc -p src
✨  Done in 1.41s.
✨  Done in 5.26s.
```

**What is the expected behavior?**

Similar to `wsrun --serial`:

```
$ yarn build
yarn run v1.10.0
$ wsrun build --serial
package-a
$ tsc -p src
package-b
$ tsc -p src
package-c
$ tsc -p src
✨  Done in 4.86s.
```

(Note, in the terminal `package-a`/`b`/`c` are also written **bold** to make the log even nicer to read.)

**Please mention your node.js, yarn and operating system version.**
- Node 8 LTS
- yarn v1.10.0
- macOS

alexindigo:
Another option would be similar to how Jenkins does it:

```
$ yarn workspaces run build
yarn workspaces v1.10.0
[package-a] yarn run v1.10.0
[package-a] $ tsc -p src
[package-a] ✨  Done in 1.49s.
[package-b] yarn run v1.10.0
[package-b] $ tsc -p src
[package-b] ✨  Done in 1.21s.
[package-c] yarn run v1.10.0
[package-c] $ tsc -p src
[package-c] ✨  Done in 1.41s.
✨  Done in 5.26s.
```

Also this format will be more useful, when yarn allow running workspace scripts in parallel _(wink, wink, it was in the proposal)_.

ianstormtaylor:
I think borrowing from Lerna with its color-coding would be nice:

![image](https://user-images.githubusercontent.com/311752/48614554-5dc3e500-e943-11e8-9649-7078cf8e4f4a.png)

zkochan:
We have a similar discussion in pnpm: https://github.com/pnpm/pnpm/issues/1440

peterlebrun:
Hi - I believe I have addressed this in the following PR: https://github.com/yarnpkg/yarn/pull/6932

Thanks!

grrowl:
If you're using `yarn workspaces foreach`, pass the `--verbose` argument to prefix the output with the package's name

PR #6932, opened 2019-01-16
Prefix workspace commands with workspace name

peterlebrun:
Fixes: #6389 (as discussed with @eps1lon/@arcanis, created a separate PR for this)

When running commands via `yarn workspaces run foo`, prefix command with bold and colorized workspace name.  (This change was inspired in large part by lerna's implementation).

This provides much more clarity when working in a multi-workspace repo.

image


arcanis:
Looks really nice, but it still has the problem that it's BC-breaking. People relying on this command output will have their scripts broken:

```
yarn workspaces run exec -s pwd
```

peterlebrun:
Hi @arcanis, that behavior has actually not changed with this PR, as this PR only affects the output of scripts defined in package.json (and behaves just like Lerna's output!)

Here is the output of the command on master:
image

And here is the output of the command on my feature branch:
image

peterlebrun:
Hi @arcanis, reading through the [`workspaces run` spec](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-workspace-run-commands.md), `yarn workspaces run -s exec pwd` was never intended to work.  Behavior of this nature was supposed to be implemented through a separate command, `yarn workspaces exec`, much like `lerna exec`.  I would (very politely) argue that the fact that it does currently work is a bug and not a feature, and the global benefit of enhanced clarity when running commands through all workspaces is greater than any inconvenience that may be caused to individuals who have come to rely on the behavior of a bug.

Thank you very much for your time and consideration, I know you're very busy!

Peter

sgronblo:
Can't it just be put behind a command line flag so that the people who want this can enable it and the people who have a working build configuration won't have that be broken?

`yarn workspaces --prefixed run build` for example?

arcanis:
I'm not sure how feasible it is with the CLI - but if it works, I'm ok with it.

Another option would be to wait for the [v2](https://github.com/yarnpkg/yarn/issues/6953), since making breaking changes of this kind will be perfectly fine then. What do you think?

TerribleDev:
@arcanis flag in v1, default in v2?

arcanis:
Yup, I'm fine with that 🙂 

If a CLI flag is too hard to implement, I'd also be open to use a yarnrc flag for it (like `enable-contextual-output` or similar).

src/util/child.js
```diff
@@ -116,7 +121,19 @@ export function spawn(
           }
 
           if (proc.stdout) {
-            proc.stdout.on('data', updateStdout);
+            // Prefix output with workspace name, if we have it
+            if (opts.workspaceName) {
+              // Stay consistent by using the key that is mapped to proc
+              if (!workspaceColors[key]) {
+                const randInt = Math.floor(Math.random() * colors.length);
+                workspaceColors[key] = colors[randInt];
```
rally25rs: Is there a reason not to just go through the colors array in order here?
It seems like by using `rand` we could potentially pick the same color for neighboring workspaces.

TerribleDev:
@arcanis ok so just to review, we will update this PR to include a flag in v1.

As far as V2 is concerned, how do we contribute to it? Or will you implement that side?

src/util/child.js
TerribleDev: I agree

TerribleDev: by going through the array it will also make the output deterministic

arcanis:
Awesome, thanks! 🙂 

The v2 trunk is here: https://github.com/yarnpkg/berry - the contribution process is pretty much the same as here (in this case the command you're likely interested in would be [`workspace foreach`](https://github.com/yarnpkg/berry/blob/master/packages/plugin-essentials/sources/commands/workspaces/foreach.ts)).

sgronblo:
Is this PR abandoned? This feature would be nice to have.

PR #6934, opened 2019-01-17
Add `YARN_DATA` environment variable for windows user to configurate where to store all yarn data files

ZenLiuCN:
**Summary**
Add `YARN_DATA` environment variable for windows user to configurate where to store all yarn data files.

Because of  `local  application data` had become a nightmare for many windows users. for me this folder
is greatter than 20GB. many my friends or workmate do wan;t some development tools data could be save into some other place.

**Test plan**

this change will no side effect, and being tested .


arcanis:
Thanks, but there's already `LOCALAPPDATA` for this

Issue #6941, opened 2019-01-20
yarn pnp should be the first preloaded module in NODE_OPTIONS


jdalton:
The pnp module is currently appended to the existing `NODE_OPTIONS` value which allows other modules to be attempted to be pre-loaded first and be unresolved.

See source  [here](https://github.com/arcanis/yarn/blob/8c44bf35f58c4a6bb411d15218a820708b3607cd/src/cli/commands/node.js#L24-L27):
```js
  if (await fs.exists(pnpPath)) {
    nodeOptions += ` --require ${pnpPath}`;
  }
```

Instead it should be something like:
```js
  if (await fs.exists(pnpPath)) {
    nodeOptions = `--require ${pnpPath} ${nodeOptions}`;
  }
```

Related to https://github.com/yarnpkg/yarn/pull/6629#discussion_r249296932.

**Update:**

PR opened at #6942.

arcanis:
Yeah, that's the problem with NODE_OPTIONS - it makes the ordering problematic. For example, there are legitimate situations where you'd want the PnP preload to be made after what's already preloaded.

And the other way around is true as well: if a new `--require` is added, it'll likely want to be added after the PnP hook, but depending on the order they auto-register it might not be the case.

I can't wait for the Loader API - hopefully it'll force all this to become explicit through configuration rather than a dynamic env parameter. I wonder if a shim could exist until then...

jdalton:
> Yeah, that's the problem with NODE_OPTIONS - it makes the ordering problematic. For example, there are legitimate situations where you'd want the PnP preload to be made after what's already preloaded.

I actually am in that situation, but because `pnp` is required to resolve `--require` modules I can't even begin. Before `yarn` changed to `NODE_OPTIONS` my strategy was to first be loaded by `pnp`, then makes changes I need, then reload `pnp`. However, now it's nixed at the load step (it's unresolvable without `pnp`).

arcanis:
Why do you need to reload PnP after `esm` ran? I'm not familiar with how it works, don't you just hook into `Module._extensions`?

jdalton:
The `esm` loader can operate as an isolated package level loader or an application level loader. Unlike babel or other transpilers the `esm` loader is a runtime simulator and manages the entire load/parse/execution pipeline. We reload `pnp` through `esm` so that when `pnp` modifies its environment its modifying the `esm` environment.

PR #6942, opened 2019-01-20
fix(pnp): make sure pnp module is the first preloaded module.

jdalton:
This PR addresses #6941 by ensuring pnp is the first preloaded module.


jdalton:
I'm trying to do `NODE_OPTIONS="-r esm" yarn node ./foo.js` for similar reasons that `yarn` switched to using `NODE_OPTIONS`.  FWIW `yarn` used to prepend itself to the node `args` so prepending (as this PR does) is more in line with how it was before the `NODE_OPTIONS` change.

arcanis:
Sounds good 👍

jdalton:
I noticed a missed a reference so opened another PR to address it at #6951.

PR #6954, opened 2019-01-24
Disable creation of cmdShim PowerShell shims because they don't work and cause problems

### LINKED ISSUE #6902: Globally installed binaries do not work in PowerShell after 1.13.0
Since the 1.13.0 update, globally installed binaries do not work in PowerShell. The offending commit is 5419606, which is the update to the `cmd-shim` package. The version of `cmd-shim` that `yarn` updated to added default support for PowerShell, meaning that every binary writes a `.ps1` shim in addition to the non-extension and the `.cmd` ones. It appears Yarn is first calling `cmdShim` during the "Linking Dependencies" phase to create the initial `bin` shims for the package. This is fine and dandy here, but later in the install step (after the `success Installed <package> with binaries` message is printed), Yarn calls `cmdShim` *again* once per binary, which causes `cmdShim` to write a shim for each of the shims. This results in files like `binary.cmd.ps1`, `binary.ps1.ps1`, etc, because `cmd-shim` thinks that the base name of the binary is actually `binary.ps1`. This causes the bash shim to be written to `binary.cmd` and `binary.ps1`, breaking those two shims (PowerShell finds the `ps1` file which is actually a bash script). Here's a visual after `yarn global add ts-node`:

image

I added a `console.trace` to `cmdShim` in my local copy of Yarn to print out the arguments passed, and this is the output:

https://gist.github.com/briman0094/d7f256d1158d6e8f96afbb74e85840bb

briman0094:
**Summary**

Fixes #6902. The latest version of `@zkochan/cmd-shim` has built-in support for PowerShell shims, but they don't work correctly (even the correct .ps1 shims throw errors in PowerShell). In addition, due to the funky way that Yarn handles global binaries, every permutation of (no extension/cmd/ps1) gets created (e.g. `.cmd.ps1`, `.cmd.cmd`, etc) as a shim. Prior to the update to `@zkochan/cmd-shim`, Yarn suppressed this by renaming `.cmd.cmd` to just `.cmd`, which is not the most elegant solution, but it works. As a (temporary?) workaround to the broken PowerShell shims, this PR simply disables the creation of PowerShell shims altogether whenever Yarn calls `@zkochan/cmd-shim`. 

**Test plan**

A new test was added in `__tests__/commands/global.js` (`shouldn't create powershell shims`) which ensures that no PowerShell shims are created, and that the duplicate-extension files do not exist.


khs1994:
Can create powershell file, not delete it?

arcanis:
Will it introduce a regression? I'd be more enclined to merge this one than #6959, since the later seems quite a bit more complex.

cc @zkochan - are you aware of this issue?

ExE-Boss:
This will restore cmd shim creation behaviour to how it was before 1.13.

arcanis:
Sounds good then - I'll merge this and release it with the 1.15.1

zkochan:
If @ExE-Boss has accepted this, then LGTM. He has more experience with it.

We don't change cmd-shim frequently. We can include @arcanis in any future changes.

otaconjh:
So does this mean I can't use yarn to global install packages?

PR #6955, opened 2019-01-24
refactor: remove unnecessary checks

DanielRuf:
**Summary**

There are two checks which are not needed. One check in a condition was added when the Facebook headers were removed and there is no reasons for the `false` check.

**Test plan**

### DIFF
--- a/src/cli/commands/help.js
+++ b/src/cli/commands/help.js
@@ -23,7 +23,7 @@ export function run(config: Config, reporter: Reporter, commander: Object, args:
       const command = commands[commandName];
       if (command) {
         command.setFlags(commander);
-        const examples: Array<string> = ((command.examples) || []).map(example => `    $ yarn ${example}`);
+        const examples: Array<string> = (command.examples || []).map(example => `    $ yarn ${example}`);
         if (examples.length) {
           commander.on('--help', () => {
             reporter.log(reporter.lang('helpExamples', reporter.rawText(examples.join('\n'))));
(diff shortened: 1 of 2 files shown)


DanielRuf:
I guess the Git resolver change is related to https://github.com/yarnpkg/yarn/pull/3449.

DanielRuf:
https://dev.azure.com/yarnpkg/yarn/_build/results?buildId=294&view=logs&jobId=2ff3c9c9-dec7-575d-2cca-4faaac369f41&taskId=ca90fcfb-b6eb-5ba5-e914-29a8da89a401&lineStart=169&lineEnd=169&colStart=34&colEnd=131

DanielRuf:
Why do the CI checks fail?

arcanis:
Restarting, but they looked pretty scary. Maybe related to the registry failures from past week?

src/cli/commands/check.js
```diff
@@ -119,7 +119,7 @@ export async function verifyTreeCheck(
         const relative = path.relative(cwd, subDepPath);
         const locations = path.normalize(relative).split(registry.folder + path.sep).filter(dir => !!dir);
         locations.pop();
-        while (locations.length >= 0) {
+        while (locations.length) {
```
arcanis: After giving it a second look this is incorrect - `>=` returns true on an empty array, but your change returns false. I think that's why the tests are failing 🙂

DanielRuf: Ok, I'll revert this one.

arcanis:
Thanks 👍

Issue #6958, opened 2019-01-24
Yarn tries to create shims of shims during global package installation


briman0094:
Related to #6902 

The update to `@zkochan/cmd-shim@3.0.0` in `yarn@1.13.0` revealed a dormant bug where Yarn calls `cmd-shim` on its own shims, creating shims of shims which do not function correctly in PowerShell. This issue mainly affects Windows, although creating shims of shims in Linux is probably not the greatest solution when the original shim could be used instead.

This issue is intended to track the underlying cause of this, and, eventually, a real fix that can replace the workaround in #6954 and allow proper PowerShell shims to be used.

tats-u:
Thank you for creating this issue on behalf of me.

Additional info:
- https://github.com/yarnpkg/yarn/blob/1af4c5f50c4c146a833c0c2b20d10be1f5a61803/src/cli/commands/global.js#L169-L171 camouflaged this issue before 1.12.3.
- How to fix:
  1. Make `@zkochan/cmd-shim` support shims of shims
  2. Stop creating shims of shims (#6959)

ExE-Boss:
> i. Make `@zkochan/cmd-shim` support shims of shims

I don't think that's a real solution.

---

/cc @zkochan

tats-u:
I agree with you, but the hospital is not available.

tats-u:
Shims mean symlinks made of shell scripts.  Since symlink supports nested links, shims must them.

PR #6959, opened 2019-01-24
Use alternate method of generating global bin shims

briman0094:
**Summary**

PR #6954 is a band-aid to a deeper problem with generating global bin shims. Currently, Yarn essentially makes a "shim shim" on Windows due to older versions of Node not supporting symbolic links. This worked fine with the `sh` and `cmd` shims, but the new `@zkochan/cmd-shim@3.0.0` PowerShell shims are written a bit differently and broke when a "shim of a shim" was created for one.

The old method of Yarn involved traversing Yarn's "Data/global/node_modules/.bin" directory after a global install operation, and for each shim file that was found, `cmd-shim` was called to create a "shim shim" in Yarn's "bin" folder. This had an unfortunate side effect of creating `shim`, `shim.cmd`, and `shim.ps1` for each global shim on Windows, meaning one would end up with every permutation of extensions, such as `shim.ps1.ps1`, and `shim.ps1` would actually contain the extensionless bash shim. Previous versions of Yarn worked around this by copying `shim.cmd.cmd` to `shim.cmd` but this was a band-aid at best.

The new method implemented by this PR fixes both #6902 and #6958 by changing the way global shims are generated. Instead of creating shims in `<yarn prefix>/bin` which point to `<yarn prefix>/Data/global/node_modules/.bin`, Yarn will now parse the global lockfile after every global operation and find the top-level binaries that would be added to `<yarn prefix>/Data/global/node_modules/.bin` and create a new top-level shim in `<yarn prefix>/bin` pointing directly to the actual binary path in the providing package's `node_modules` folder.

I have tested this on Windows 10 using the version of PowerShell included with Windows, and the issue originally documented in #6902 is fixed by the changes made in this PR.

**Test plan**

A test was added to `__tests__/commands/global.js` to ensure the proper binaries are created on Windows. The existing global tests should cover regression testing. I have tested this PR locally and successfully used a global binary installed with my local compiled version of Yarn without issue.


src/cli/commands/global.js
```diff
@@ -159,18 +160,38 @@ async function initUpdateBins(config: Config, reporter: Reporter, flags: Object)
       }
     }
 
-    // add new bins
-    for (const src of afterBins) {
```
arcanis: Wouldn't it be fixed simply but filtering `afterBins` here to make sure we don't iterate over `.ps1` and `.cmd` files?

briman0094: It would still be creating shims of shims. It would fix the `.ps1.cmd`/`.cmd.ps1`/etc issue, but it would not fix the problem of there being shims pointing to shims pointing to binaries. `cmd-shim` does not work correctly when such a condition occurs, at least not with PowerShell.

briman0094: Longer explanation:

The `cmd-shim` package looks at the shebang of the program it's targeting to figure out what to write into the shim. This works fine for those two, because for bash it will always be either `node` or `sh` and for cmd it will always be either `node.exe` or `cmd.exe`. PowerShell breaks because the shebang on the PowerShell shim ends up referencing `pwsh.exe`, which is the binary for PowerShell Core, even when PowerShell Core is not installed. When a shim is created to a PowerShell shim, it automatically references `pwsh` even when PowerShell Core isn't available on the system.

I suppose it might be worth it to file an issue with `cmd-shim` to write the *correct* shebang to PowerShell shims, but I also would have to argue that it's still best to have one layer of shims if possible instead of shims-pointing-to-shims.

MLoughry:
Is there any traction on this?

briman0094:
Rebased to fix conflicts due to my other PR; should be a clean merge now for if this is approved at some point

Daniel15:
Seems reasonable to me... @arcanis  what do you think?

PR #6976, opened 2019-01-29
Check os and platform even when engines is not present in package.json

### LINKED ISSUE #6948: yarn ignores the "cpu" param in package.json when "engines" is missing
bug

yarn ignores the `cpu` setting in `package.json` when the `engines` section does not exist in `package.json`.

make sure your node's `process.arch` is not `ia32`. for example, use a 64bit node on 64bit windows.
in an empty directory, save this as `package.json`:
```
{
  "cpu": ["ia32"]
}
```

now run `yarn install`. it will finish without showing any errors. this is the bug.
(you can try `npm install`, and it will show the error correctly)

now replace the content of `package.json` with this:
```
{
  "cpu": ["ia32"],
  "engines": {}
}
```
and run `yarn install`. this time it will correctly say that the cpu-setting is incorrect.

yarn should respect the `cpu` setting in `package.json` when the engines setting does not exist.

node 8.15.0
windows 10
yarn 1.13.0

MichaReiser:
**Summary**

This change ensures that the `cpu` and `os` requirements are checked if they are present in the `package.json` even when the package does not specify any `engines` requirements.

Fixes #6948 

**Test plan**

package.json:

```json
{
  "cpu": ["i32"],
  "os": ["win32"]
}
```

Running the following commands on a mac osx machine:

```bash
../bin/yarn
yarn install v1.15.0-0
warning No license field
[1/5] 🔍  Validating package.json...
warning No license field
error @: The platform "darwin" is incompatible with this module.
error @: The CPU architecture "x64" is incompatible with this module.
```

```bash
../bin/yarn --ignore-platform
yarn install v1.15.0-0
warning No license field
success Already up-to-date.
✨  Done in 0.22s.
```

### DIFF
--- a/src/package-compatibility.js
+++ b/src/package-compatibility.js
@@ -128,20 +128,17 @@ export function checkOne(info: Manifest, config: Config, ignoreEngines: boolean)
     }
   };
 
-  const invalidPlatform =
-    shouldCheckPlatform(info, config.ignorePlatform) && info.os != null && !isValidPlatform(info.os);
+  const {os, cpu, engines} = info;
 
-  if (invalidPlatform) {
+  if (shouldCheckPlatform(os, config.ignorePlatform) && !isValidPlatform(os)) {
     pushError(reporter.lang('incompatibleOS', process.platform));
   }
 
-  const invalidCpu = shouldCheckCpu(info, config.ignorePlatform) && info.cpu != null && !isValidArch(info.cpu);
-
-  if (invalidCpu) {
+  if (shouldCheckCpu(cpu, config.ignorePlatform) && !isValidArch(cpu)) {
     pushError(reporter.lang('incompatibleCPU', process.arch));
   }
 
-  if (shouldCheckEngines(info, ignoreEngines)) {
+  if (shouldCheckEngines(engines, ignoreEngines)) {
     for (const entry of entries(info.engines)) {
       let name = entry[0];
       const range = entry[1];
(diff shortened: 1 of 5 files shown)


arcanis:
Looks good! Can you run the linting and add an integration test to our new testsuite? This testsuite is located in `packages/pkg-tests`, and can be run through `yarn && yarn jest` :)

src/package-compatibility.js
```diff
@@ -127,19 +129,19 @@ export function checkOne(info: Manifest, config: Config, ignoreEngines: boolean)
   };
 
   const invalidPlatform =
-    !config.ignorePlatform && Array.isArray(info.os) && info.os.length > 0 && !isValidPlatform(info.os);
+    shouldCheckPlatform(info, config.ignorePlatform) && info.os != null && !isValidPlatform(info.os);
```
MichaReiser: Flow requires the `infos.os != null` check because it does not know that the `os` field is never null when `shouldCheckPlatform` is true. If we use TypeScript we could implement `shouldCheckPlatform` as a user defined type guard.

```
function isFish(pet: Fish | Bird): pet is Fish {
    return (<Fish>pet).swim !== undefined;
}
```

MichaReiser: Thanks @arcanis for referencing me to the documentation of `%checks` ([link](https://flow.org/en/docs/types/functions/#toc-predicate-functions)).

arcanis:
🎉

Issue #6979, opened 2019-01-29
Yarn should not require version attribute


alexander-schranz:
**What is the current behavior?**

When having a global package like the following.

```
    "dependencies": {
        "sulu-admin-bundle": "file:vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js",
    }
```

Yarn will end up with:

> error Can't add "sulu-admin-bundle": invalid package version "".

**What is the expected behavior?**

Local dependencies should not be forced to have a version number.

**Please mention your node.js, yarn and operating system version.**

Yarn: 1.13.0
Node: v10.15.0
OS: Mac OSX MoJave

PR #6980, opened 2019-01-29
Fix no version in locale dependency

alexander-schranz:
**Summary**

see #6979 

**Test plan**

```
# create package.json without version
# require it over package.json as folder 
# run:
yarn install
```


src/package-request.js
```diff
@@ -107,6 +107,16 @@ export default class PackageRequest {
       // times
       data = Object.assign({}, data);
 
+      // local dependencies are not required in having a version number so they get
+      // the current timestamp to require it
+      if (!data.version) {
+        data.version = Date.now().toString() + '.0.0';
```
arcanis: `Date.now()` would be a no-go because it would make installs behave differently from one to the other (because the versions would be different), but just `0.0.0` would be ok 🙂

alexander-schranz: `0.0.0` was also my first choice but thought this could end up in caching issues?

alexander-schranz: Ok changed to 0.0.0

alexander-schranz:
@arcanis not sure if this test work. I'm not very familiar with jest and my IDE seems not to recognize the jest tests correctly in this repo. How I can run a single test from the CLI?

src/util/generate-pnp-map.js
```diff
@@ -232,6 +232,10 @@ async function getPackageInformationStores(
       const packageName = pkg.name;
       let packageReference = pkg.version;
 
+      if (!packageReference) {
+        packageReference = '0.0.0';
```
alexander-schranz: @arcanis not sure where and how to fix this issue with pnp
https://circleci.com/gh/yarnpkg/yarn/28459?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

arcanis: I'd try to add a `console.log(resolutions)` when [this invariant](https://github.com/yarnpkg/yarn/blob/7f419103307da7856627c4ca078460234701eb75/src/util/generate-pnp-map.js#L318) fails, and see what's the key being used. If the package isn't there at all, it means that either it wasn't part of the `seedPatterns`, or that [`getResolverEntry`](https://github.com/yarnpkg/yarn/blob/7f419103307da7856627c4ca078460234701eb75/src/util/generate-pnp-map.js#L193-L209) returned `null` for some reason.

Note that you can run the exact failing test by using the following command in `packages/pkg-tests`:

```
yarn jest -t 'it should install a single dependency without version'
```

Issue #6672, opened 2018-11-13
No way to silence warnings


rochdev:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

Using `yarn --silent` is not silent since it displays a lot of errors related to package deprecation.

For example:

```
warning workspace-aggregator-7f30cf29-537f-48a2-86b0-a91668bf6021 > hapi-f3323ea51e133c445b5c74b7fce4def22ad0f7bb > hapi@8.5.0: This version is no longer maintained. Please upgrade to the latest version of the framework.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Add `hapi@2` to `package.json` and run `yarn install`

**What is the expected behavior?**

Either `yarn --silent` should be truly silent, or a log level configuration should be added to add the ability to ignore warnings.

**Please mention your node.js, yarn and operating system version.**

* macOS 10.13.6
* Node v8.11.1
* yarn 1.9.4

ngryman:
Hey guys, I know you must have a huge backlog but this one seems to be a quick one with a decent amount of support (25 👍). Is there any chance you could take a quick look at it?

Thanks!

ragozin-n:
+ 1

mohsen1:
On a Unix system (Linux, macOs) you can do some grepping! 

```bash
yarn install 2> >(grep -v warning 1>&2)
```

this cleaned up our logs in CI at least.

victorwads:
+1

ftmmohd2000:
+1

paul-soporan:
Closing as you can now silence specific warnings in v2 via [`logFilters`](https://yarnpkg.com/configuration/yarnrc#logFilters).

https://yarnpkg.com/getting-started/migration

h0jeZvgoxFepBQ2C:
Why is this closed? Also the link doesnt show how to silcen deprecations?

ildella:
> Why is this closed? Also the link doesnt show how to silcen deprecations?

The documentation link does. You should add this to your yarnrc config yaml

```yml
logFilters: [{
  code: 'YN0002',
  level: 'discard',
}]
```

Which ironically is json :)

rubberTres:
I just added 
```
"build-production": "DISABLE_ESLINT_PLUGIN=true react-scripts build"
```
 to scripts in package.json and works just fine

maksymkp:
> I just added
> 
> ```
[...]

PR #6981, opened 2019-01-29
Don’t call `release` with an exit code

eelco:
The `release` signature is incorrect, because its first argument is a callback function that will be called when the release is done. By attaching it to the `run()` promise however, release will be called with the exit code as argument. When the exit code is `1`, the code will crash. 

This is very rare, but we’re seeing this when calling yarn from another node process (and possibly killing it too soon):
```
<redacted>/node_modules/yarn/lib/cli.js:78119
         callback();
         ^
 TypeError: callback is not a function
     at options.fs.rmdir (<redacted>/node_modules/yarn/lib/cli.js:78119:9)
     at FSReqWrap.oncomplete (fs.js:141:20)
```

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -303,7 +303,7 @@ export async function main({
           onDeath(() => {
             process.exitCode = 1;
           });
-          resolve(run().then(release));
+          resolve(run().then(() => release()));
         }
       });
     });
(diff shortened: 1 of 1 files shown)


olingern:
related https://github.com/yarnpkg/yarn/issues/6672

arcanis:
Good catch! In fact, looking at the code I think we even call `resolve` too soon, since `release` seems to be asynchronous. We should probably do something like this:

```js
.then(() => new Promise(resolve => release(resolve))).then(resolve)
```

Care to update the PR?

eelco:
@arcanis Done. I think it can be slightly simpler than the version in your comment. I’ve also tested that this works in our use case and that the release callback is indeed called.

arcanis:
Lgtm, thanks for your contribution! 👍

PR #7000, opened 2019-02-04
Adds test for #6968

sasucker:
**Summary**

#6968 fixed a regression where a class of errors from `request` were unhandled (leaving ssl cert errors to be swallowed).

This PR just adds a test to hopefully catch future regressions.

While adding this test, I noticed some tests from #6413 were leaking and addressed that as well.

@arcanis and @yan12125 do these changes look helpful?

I did not update `CHANGELOG` as actual code was unchanged.

**Test plan**

Successfully ran `yarn test`

image

### DIFF
--- a/__tests__/util/request-manager.js
+++ b/__tests__/util/request-manager.js
@@ -186,6 +186,7 @@ test('RequestManager.execute timeout error with default maxRetryAttempts', async
 
 for (const statusCode of [403, 442]) {
   test(`RequestManager.execute Request ${statusCode} error`, async () => {
+    jest.resetModules();
     // The await await is just to silence Flow - https://github.com/facebook/flow/issues/6064
     const config = await await Config.create({}, new Reporter());
     const mockStatusCode = statusCode;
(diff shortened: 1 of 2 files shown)


__tests__/util/request-manager.js
```diff
@@ -203,13 +204,34 @@ for (const statusCode of [403, 442]) {
...
+    };
+  });
+  const config = await await Config.create({}, new Reporter());
+  await config.requestManager.execute({
+    params: {
+      url: `https://localhost:port/?nocache`,
+      headers: {Connection: 'close'},
+    },
+    resolve: body => {},
+    reject: err => {
```
arcanis: Shouldn't this callback be spied on? At the moment Jest doesn't check whether it's actually called 🤔

sasucker: I was actually just following existing convention, but it looks like the tests are [invoking](https://github.com/yarnpkg/yarn/pull/7000/files#diff-ae8e937424abd07f0db1a5b3f6532758R217) the [callback](https://github.com/yarnpkg/yarn/blob/dac22ddfea657946f5c2943cab69bff7dae63ebc/src/util/request-manager.js#L412) manually, which runs the "async" resolution. The `await` in front of `config.requestManager.execute` is misleading because RequestManager#execute is a void method.

The tests technically test the correct behavior, but could probably be a bit more straightforward.

yan12125:
> @yan12125 do these changes look helpful?

I have little knowledge on how tests work in yarn. GTM as long as tests pass :)

PR #7007, opened 2019-02-05
fix: drive letter casing for win32 pnp

### LINKED ISSUE #7005: plugnplayWindowsSupport warning because of drive letter case mismatch
When running `yarn --pnp` in directory 'D:\projectName', this pops up:

**warning** Plug'n'Play on Windows doesn't support the cache and project to be kept on separate drives

The cache directory is configured as 'd:\\.yarncache', `yarn cache dir` returns 'd:\\.yarncache\v4'.
Afaik, windows drive letters in paths are case insensitive, but the code around line 400 of `src/config.js` does not account for that and produce a mismatch between `cacheRootFolderDrive` and `lockfileFolderDrive`, resulting in the warning shown above.

It would be nice if this could be fixed.

specs: windows 10, yarn 1.13.0, node 10.15.1

codepunkt:
Fixed potentially mismatching casing of drive letters in win32 pnp check by toLowerCasing both strings.

Fixes #7005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Fixes occasionally mismatching upper/lowecases of drive letters in win32 pnp check
+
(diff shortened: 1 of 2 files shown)


arcanis:
Thanks 👍

PR #7009, opened 2019-02-06
Include key info for "expected hoisted" invariant

jrwats:
**Summary**
Includes the key that we expected to find hoisted in the tree in the invariant message.

**motivation**
I hit this invariant.  I wanted to know what package "caused" it.

**Test plan**
Ran this in "broken" yarn environment
screen shot 2019-02-06 at 1 56 55 pm


arcanis:
👍

PR #7021, opened 2019-02-10
feat(run): Add engines check before executing scripts.

### LINKED ISSUE #7013: Check engines before running scripts
**I want to request a *feature*, if it doesn’t exist yet.**

_I wasn’t able to find this in the [yarn run docs](https://yarnpkg.com/lang/en/docs/cli/run/) nor a previous issue asking about this, if I missed it please point me at the right direction._

I would like yarn to check the `engines` specified in the `package.json` manifest file _before_ executing any scripts through `yarn run` (or just `yarn`). In my case I’m specifically interested in ensuring my team all invoke scripts using the same Node version.

We already have a program that checks the Node version and some of our scripts invoke that before doing their actual work, however:

* It's easy to add a script and forget to prefix it with our verification program.
* It doesn’t help the cases where people invoke a program directly without there being a script entry in the `package.json` file. E.g. they might invoke `yarn jest` directly, instead of `yarn test`.

Yarn will run scripts without complaining when engine requirements do not match the current env.

I would like yarn to complain. E.g.

```
$ yarn jest
The current environment does not satisfy the engine "node" requirements. Expected version "10.13.*".
$ nvm use 10.13
Now using node v10.13.0
$ yarn jest
yarn run v1.12.3
/Users/eloy/Code/Artsy/metaphysics/node_modules/.bin/jest
[…]
```

alloy:
Closes #7013

**Summary**

A more elaborate description is available in #7013. In short, ensuring scripts are ran in the expected environment can deter hard to diagnose bugs.

**Test plan**

Given a `node` engine requirement that doesn’t match the current environment:

```
$ cat package.json | grep -i1 '"node"'
  "engines": {
    "node": "^10.13.0"
  },

$ nvm use 8
Now using node v8.12.0 (npm v6.4.1)
```

A script invocation will fail as follows:

```
$ node ~/Code/JavaScript/yarn/lib/cli/index.js run relay
yarn run v1.15.0-0
error @artsy/reaction@12.1.10: The engine "node" is incompatible with this module. Expected version "^10.13.0". Got "8.12.0"
error Commands cannot run with an incompatible environment.
```

If need be, this can be overridden with the `--ignore-engines` flag:

```
$ node ~/Code/JavaScript/yarn/lib/cli/index.js run --ignore-engines relay
yarn run v1.15.0-0
$ relay-compiler --src ./src --schema data/schema.graphql --language typescript --artifactDirectory ./src/__generated__ --exclude '**/node_modules/**,**/__mocks__/**,**/__generated__/**'
✨  Done in 2.22s.
```

Or, of course, making the environment match the requirements:

```
$ nvm use 10.13
Now using node v10.13.0 (npm v6.4.1)

$ node ~/Code/JavaScript/yarn/lib/cli/index.js run relay
yarn run v1.15.0-0
$ relay-compiler --src ./src --schema data/schema.graphql --language typescript --artifactDirectory ./src/__generated__ --exclude '**/node_modules/**,**/__mocks__/**,**/__generated__/**'
✨  Done in 1.91s.
```


src/reporters/lang/en.js
```diff
@@ -229,11 +229,12 @@ const messages = {
   nodeGypAutoInstallFailed:
     'Failed to auto-install node-gyp. Please run "yarn global add node-gyp" manually. Error: $0',
 
-  foundIncompatible: 'Found incompatible module',
+  foundIncompatible: 'Found incompatible module.',
```
alloy: Looks like this was the odd one out, so added some punctuation 🤓

alloy:
I'm unsure why [Flow fails](https://circleci.com/gh/yarnpkg/yarn/28740?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). My changes are in that file, but otherwise they seem unrelated, to my naive Flow eyes.

src/cli/commands/run.js
```diff
@@ -118,6 +119,13 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
     }
 
     if (cmds.length) {
+      const ignoreEngines = !!(flags.ignoreEngines || config.getOption('ignore-engines'));
+      try {
+        await checkCompatibility({_reference: {}, ...pkg}, config, ignoreEngines);
+      } catch (err) {
+        throw err instanceof MessageError ? new MessageError(reporter.lang('cannotRunWithIncompatibleEnv')) : err;
```
alloy: This refines a, in this context, vague [error](https://github.com/yarnpkg/yarn/blob/4eb6391efbbe5f3dbce3c5c4f31c246c7130c7da/src/package-compatibility.js#L161).

rally25rs:
_I put these same comments in Discord chat, but I'll paste them here too for tracking with this PR:_

wow, I don't know how it came up with that error message for that line, but...
it's specifically mad about:
```
await checkCompatibility({_reference: {}, ...pkg}, config, ignoreEngines);
```
you are assigning `_reference: {}` but that isn't valid.

Function signature is: `checkCompatability(Manifest, ...`
and a `Manifest`'s `_reference` is of type `PackageReference`

`{}` isn't a valid `PackageReference` because that type has several required fields
```
  requests: Array<PackageRequest>;
  lockfile: Lockfile;
  config: Config;

  isPlugnplay: boolean;
  level: number;
  name: string;
  version: string;
  uid: string;
  ignore: boolean;
  incompatible: boolean;
  fresh: boolean;
  dependencies: Array<string>;
  patterns: Array<string>;
  permissions: {[key: string]: boolean};
  remote: PackageRemote;
  registry: RegistryNames;
  locations: Array<string>;
  resolver: PackageResolver;
```

alloy:
> wow, I don't know how it came up with that error message for that line, but...

Wow, indeed! I don’t even understand how you were able know what it was really complaining about 😅 

> it's specifically mad about:
>
> ```
> await checkCompatibility({_reference: {}, ...pkg}, config, ignoreEngines);
> ```
>
> you are assigning `_reference: {}` but that isn't valid.

I see. At first I just passed `pkg`, but then at runtime it was complaining about missing `_reference`. Then I saw that the `install` command [does this](https://github.com/yarnpkg/yarn/blob/0e380a19cbee8aea15f025b09794e29a124c6631/src/cli/commands/install.js#L747) so I cargo-culted that.

Based on that, clearly the reference isn’t _actually_ required, so I’m going to remove the `invariant` call.

rally25rs:
`package-compatability.js`

```
export function checkOne(info: Manifest, config: Config, ignoreEngines: boolean) {
...
  const pushError = msg => {
    const ref = info._reference;
    invariant(ref, 'expected package reference');

    if (ref.optional) {
```

if you remove the `invariant` then the `if` could exception if `ref` is `undefined`, hence the `invariant`.

You'll probably have to change the`if` to `if (ref && ref.optional) {` for Flow to let it pass. I think that makes more sense than the invariant anyway, since `Manigest._reference` is nullable
```
  _reference?: ?PackageReference,
```
The code should work whether or not `_reference` is `undefined`. 👍

alloy:
@rally25rs Yup, I added that guard, but I squashed my commits so maybe it's not clear that I pushed new changes. Sorry about that.

alloy:
@rally25rs bump, can I get another review? 🙏

src/cli/commands/install.js
```diff
@@ -744,7 +744,7 @@ export class Install {
 
   async checkCompatibility(): Promise<void> {
     const {manifest} = await this.fetchRequestFromCwd();
-    await compatibility.checkOne({_reference: {}, ...manifest}, this.config, this.flags.ignoreEngines);
+    await compatibility.checkOne(manifest, this.config, this.flags.ignoreEngines);
```
rally25rs: what does this change do? 🤔

alloy: This was needed to avoid hitting that same invariant that I did.

alloy:
Just fixed a new merge conflict. Is there anything I can do to move this forward?

alloy:
Thanks!

PR #7030, opened 2019-02-14
Feat exclude global gitignore

cacheflow:
**Summary**

Allows a user to exclude files/folders in a global gitignore file instead of copying and pasting the same contents into new gitignore file for each project. 

When publishing NPM packages, a user may want to also exclude additional files not present in their local `gitignore` that they may have forgotten about.

**Test plan**

Added tests to ensure packing still works and properly excludes the right folders/files when a global gitiignore is present.

screen shot 2019-02-13 at 11 23 40 pm
screen shot 2019-02-13 at 11 22 13 pm


src/cli/commands/pack.js
```diff
@@ -93,13 +100,25 @@ export async function packTarball(
   }
 
   const files = await fs.walk(config.cwd, null, new Set(FOLDERS_IGNORE));
+
+  if (globalGitIgnoreFile) {
+    const stat = await lstat(globalGitIgnoreFile);
+    const loc = path.resolve(globalGitIgnoreFile);
+    const basename = path.basename(globalGitIgnoreFile);
+    files.push({
```
rally25rs: Does this put the global gitignore file in the packed archive? Is there a reason to include this? When it's unpacked in someone's `node_modules` directory they wouldn't be using it as a git archive, so it doesn't seem like we would need to persist this.

rally25rs:
Looks like there are some failing unit tests from the ci server. Could you see if you can resolve those? It looks like the new call to `git` is failing for some of the tests

```
    error Command failed.
    Exit code: 1
    Command: git
    Arguments: config --global core.excludesfile
```

cacheflow:
Thanks for the review/feedback @rally25rs. Yeah, I was trying to figure out why the tests were passing on one Node version but failing on another. Thanks for pointing me in the right direction. I'll take a look and fix them.

cacheflow:
Alright @rally25rs I think I finally figured this issue out after rebuilding the environment locally. The git command that checks for the [global gitignore](https://git-scm.com/docs/git-config#FILES) fails with an exit code of 1 as noted in the Git docs. However, the git version within Circle CI is too old to print out the actual error message, which made it hard to catch.

This is the error message locally :
ci failure


This is the error message from Circle CI:
local failure

I've noticed in the codebase there's a Promise wrapper for catching errors when using `spawn`, but I was not exactly sure how Yarn gracefully handle errors in the codebase and if they should be written to debug logs or not. Can you point me in the right direction so I can gracefully handle this and ensure packing still occurs if the command fails?

cacheflow:
Never mind I see that errors are logged out when using spawn. I just wrapped it in a try/catch and swallowed the call to git like it's done in other areas of the codebase. @rally25rs

PR #7042, opened 2019-02-18
Don't write cache path into yarn.lock for nested file dependencies

sokki:
**Summary**

Solves https://github.com/yarnpkg/yarn/issues/4388
Based on https://github.com/yarnpkg/yarn/pull/4717

The linked issue is unresolved for over a year now. It seems like solving the root cause of this issue by bypassing the cache for file-dependencies would require too much refactoring of the underlying architecture. At least I got lost while trying to do so 🙃 

So I chose this rather naive approach. All tests pass and I've tested it in a bigger monorepo where we regularly faced this issue.

**Test plan**
What was written to yarn.lock before:
image

### DIFF
--- a/__tests__/commands/install/integration.js
+++ b/__tests__/commands/install/integration.js
@@ -375,6 +375,10 @@ test('root install with optional deps', () => runInstall({}, 'root-install-with-
 test('install file: protocol with relative paths', () =>
   runInstall({}, 'install-file-relative', async config => {
     expect(await fs.readFile(path.join(config.cwd, 'node_modules', 'root-a', 'index.js'))).toEqual('foobar;\n');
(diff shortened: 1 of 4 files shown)


arcanis:
The 1.15 will have a fair amount of changes already, so I'd like to postpone this to the 1.16 if you don't mind (since it might have unseen side effects) 😃

sokki:
Sure. As long as we don't have to wait for berry 🙃

arcanis:
Nope, no worry, I keep merging PRs over here 😉

krigar1184:
Any chance for this to be merged any time soon?

desaiuditd:
@arcanis When this PR is scheduled to be merged?

PR #7070, opened 2019-02-28
fix: handle npm prefix config environment variables

### LINKED ISSUE #4682: Underscore prefixed config (_auth) provided by env var is converted to dash (-auth)
Bug

npm config that have a `_` prefix (`_auth`, `_authToken`, `_password`), when provided as environment variables (`export npm_config__auth=blah`) have their `_` prefix converted to `-` (`-auth` instead of `_auth` for example).

```
npm_config__auth=abc yarn config get _auth
undefined

npm_config__auth=abc yarn config list | grep abc
  '-auth': 'abc' }
```

The npm config which start with `_` have their prefix untouched (`npm_config__auth` -> `_auth`)

yarn 1.1.0 and 1.2.0
node 6.11.4
debian jessie

NicholasBoll:
Fixes #4682

**Summary**

Passing `npm_config_` environment variables changed all underscores (`_`) to dashes (`-`) which isn't correct behavior for special config variables (`_auth`, `_authToken`, `_username` and `_password`). An additional check was added to `base-registry.js` to filter out these config keys from this processing.

**Test plan**

`yarn test` - all tests passed except two tests that have nothing with my change - they are integration tests (I assume my environment isn't set up correctly for these to pass? Maybe CI will verify?)

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -731,6 +731,37 @@ describe('isScopedPackage functional test', () => {
   });
 });
 
+describe('environment variables functional test', () => {
+  beforeEach(() => {
+    process.env.npm_config_always_auth = 'true';
+    process.env.npm_config__auth = 'auth';
+    process.env.npm_config__authtoken = 'authToken';
+    process.env.npm_config__username = 'username';
+    process.env.npm_config__password = 'password';
+  });
+
+  afterEach(() => {
+    delete process.env.npm_config_always_auth;
+    delete process.env.npm_config__auth;
+    delete process.env.npm_config__authToken;
+    delete process.env.npm_config__username;
+    delete process.env.npm_config__password;
+  });
+
+  test('correctly escapes environment config variables', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter, true, []);
+
+    npmRegistry.mergeEnv('npm_config_');
+    expect(npmRegistry.config).toHaveProperty('always-auth', true);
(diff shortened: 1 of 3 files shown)


src/registries/base-registry.js
```diff
@@ -171,7 +171,9 @@ export default class BaseRegistry {
       key = key.replace(/__/g, '.');
 
       // replace underscores with dashes
-      key = key.replace(/_/g, '-');
```
arcanis: Can you instead add it in a more generic way that supports `__`?

For example: `key.replace(/([^_])_/g, '$1-')`

NicholasBoll: That's a good idea. We're probably assuming any key starting with an underscore needs to stay an underscore, but that's probably always true

__tests__/registries/npm-registry.js
```diff
@@ -731,6 +731,37 @@ describe('isScopedPackage functional test', () => {
...
+    delete process.env.npm_config__auth;
+    delete process.env.npm_config__authToken;
+    delete process.env.npm_config__username;
+    delete process.env.npm_config__password;
+  });
+
+  test('correctly escapes environment config variables', () => {
+    const testCwd = '.';
+    const {mockRequestManager, mockRegistries, mockReporter} = createMocks();
+    const npmRegistry = new NpmRegistry(testCwd, mockRegistries, mockRequestManager, mockReporter, true, []);
```
NicholasBoll: I had trouble figuring out where these tests should go. `mergeEnv` is an instance method that requires a whole mocked registry to test.

Simplifying this test would involve refactoring the `mergeEnv` method to use a utility method, but I decided to reduce the changes to the code and instead have a more complicated test

src/registries/base-registry.js
NicholasBoll: @arcanis Are the changes satisfactory?

arcanis:
Yep! Thanks 😃

PR #7080, opened 2019-03-04
fix(upgrade) Fix basicSemverOperatorRegex

### LINKED ISSUE #7079: yarn upgrade --latest error for dependency specified with >=
Bug

When run `yarn upgrade --latest --non-interactive --verbose` in a project with

**package.json:**
```
{
  "name": "test",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "underscore": ">=1.8.3"
  }
}
```
**yarn.lock**
```
underscore@>=1.8.3:
  version "1.8.3"
  resolved "https://registry.yarnpkg.com/underscore/-/underscore-1.8.3.tgz#4f3fb53b106e6097fcf9cb4109f2a5e9bdfa5022"
  integrity sha1-Tz+1OxBuYJf8+ctBCfKl6b36UCI=
```

```
verbose 1.265 Considering upgrade of "underscore" to "underscore@>1.9.1" because a newer version exists in the registry.
verbose 1.27 Not unlocking "underscore@>1.9.1" in the lockfile because it is a new or direct dependency.
[1/4] Resolving packages...
verbose 1.327 Error: Couldn't find any versions for "underscore" that matches ">1.9.1"
    at MessageError.ExtendableBuiltin (C:\Program Files (x86)\Yarn\lib\cli.js:727:66)
    at new MessageError (C:\Program Files (x86)\Yarn\lib\cli.js:756:123)
    at Function.<anonymous> (C:\Program Files (x86)\Yarn\lib\cli.js:49489:13)
    at step (C:\Program Files (x86)\Yarn\lib\cli.js:304:30)
    at C:\Program Files (x86)\Yarn\lib\cli.js:315:13
error Couldn't find any versions for "underscore" that matches ">1.9.1"
```

Same as above.

It should upgrade to "underscore@>=1.9.1"
node: v11.10.1
yarn: v1.13.0
OS: Windows 10

shilcare:
**Summary**

Fix bug #7079 of `basicSemverOperatorRegex`. It doesn't match the range operator correctly if the dependency range is specified with `>=`. For example, when running `yarn upgrade --latest`, if a dependency is specified with `>=`, the matched result will be `>`, which is not correct. 

This is because the regular expression engine looks for alternations one-by-one. We can fix it by putting `>=` before `>`.

fixes #7079 

**Test plan**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Fixes yarn `upgrade --latest` for dependencies using `>` or `>=` range specifier
+
+  [#7080](https://github.com/yarnpkg/yarn/pull/7080) - [**Xukai Wu**](https://github.com/shilcare)
+
 - Fixes `--modules-folder` handling in several places (ex: `yarn check` now respects `--modules-folder`)
 
   [#6850](https://github.com/yarnpkg/yarn/pull/6850) - [**Jeff Valore**](https://twitter.com/codingwithspike)
--- a/src/cli/commands/upgrade.js
+++ b/src/cli/commands/upgrade.js
@@ -12,7 +12,7 @@ import {Install} from './install.js';
 
 // used to detect whether a semver range is simple enough to preserve when doing a --latest upgrade.
 // when not matched, the upgraded version range will default to `^` the same as the `add` command would.
-const basicSemverOperatorRegex = new RegExp('^(\\^|~|>|<=|>=)?[^ |&,]+$');
+const basicSemverOperatorRegex = new RegExp('^(\\^|~|>=|<=|>)?[^ |&,]+$');
 
 // used to detect if a passed parameter is a scope or a package name.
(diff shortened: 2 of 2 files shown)


rally25rs:
Nice catch! Thanks for the contribution. 🎉
Could you please add a line to `CHANGELOG.md` for this change?

rally25rs:
You know what, maybe we should just take `>` out of the regex. I think it actually causes another bug. If yarn upgrades you to a "latest" of `1.0.0` and preserves the `>` then it will change your pacakge.json to `>1.0.0` which makes the latest version not acceptable. 🤦‍♂️ oof.

shilcare:
> You know what, maybe we should just take `>` out of the regex. I think it actually causes another bug. If yarn upgrades you to a "latest" of `1.0.0` and preserves the `>` then it will change your pacakge.json to `>1.0.0` which makes the latest version not acceptable. 🤦‍♂️ oof.

I think you're right :). We can't handle `>` with current `buildPatternToUpgradeTo` implementation.

rally25rs:
@shilcare if you could add a line to `CHANGELOG.md` for this change, I can get it merged 👍

PR #7133, opened 2019-03-20
Add private field to workspaces info command

rtsao:
**Summary**

A common use case of `yarn workspaces info` is programmatically retrieving a list of all packages in a repo. For many purposes, it is desirable to skip packages marked as `private`.

This PR adds the private field to the returned JSON for each package.

**Test plan**

Suppose there are two packages, `foo` and `bar` (private).

`yarn workspaces info` now yields:

```
{
  "foo": {
    "location": "packages/foo",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": [],
  },
  "bar": {
    "location": "packages/bar",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": [],
    "private": true
  },
}
```

### DIFF
--- a/src/cli/commands/workspaces.js
+++ b/src/cli/commands/workspaces.js
@@ -56,6 +56,7 @@ export async function info(config: Config, reporter: Reporter, flags: Object, ar
       location: path.relative(config.lockfileFolder, loc).replace(/\\/g, '/'),
       workspaceDependencies: Array.from(workspaceDependencies),
       mismatchedWorkspaceDependencies: Array.from(mismatchedWorkspaceDependencies),
+      private: manifest.private,
     };
   }
 


arcanis:
I'm not too sure about this - `yarn workspaces info` is meant to be parsed anyway, so why not just retrieve this information via ``require(`${location}/package.json`).private``?

The problem is that if we go this way all fields are useful, one way or another. I think it would make sense to only expose in `yarn workspaces info` the fields that cannot be easily computed otherwise (such as the workspace location, or their interdependencies), and let the consumer figure out the rest.

rtsao:
> I think it would make sense to only expose in yarn workspaces info the fields that cannot be easily computed otherwise (such as the workspace location, or their interdependencies), and let the consumer figure out the rest.

Agreed, this reasoning makes sense.

Perhaps a better approach would be something akin to `yarn workspaces info --no-private` or `yarn workspaces run --no-private` which would skip private packages. For publishing-related workflows, skipping private packages is a common use case.

> why not just retrieve this information via `` require(`${location}/package.json`).private ``?

Ideally, this can be done without needing some separate script to re-load and parse the package.json when this is already done by yarn. Requiring some separate script to manually filter out private packages by reading several files from disk makes doing simple actions from the command line rather inconvenient.

PR #6850, opened 2018-12-23
fix(modules-folder): Fix --modules-folder handling in several places.

rally25rs:
**Summary**

Fixes a few issues with the `--modules-folder` parameter.

The extraneous file check would always also clean files from `node_modules`. This no longer happens, only the `--modules-folder` location is checked.

The `yarn check` command used to always only check `node_modules`. Now it will check the `--modules-folder` location instead. (change to `src/config.js`)

When running a command or executing a lifecycle script, yarn would build a `PATH` that included `node_modules/.bin` first, then add the `--modules-folder`'s `/.bin` after, which would have led to the wrong binaries being executed. (change to `src/util/execute-lifecycle-script.js`)

There were a couple places where `config.registries[name]` was used, instead of `config.registryFolders` (which would contain the `--modules-folder` override) so I fixed those. (change to `src/cli/commands/autoclean.js` and `src/cli/commands/check.js`)

fixes #5419
fixes #6847 
fixes #2784 
fixes #6918 

**Test plan**

test added for checking that extraneous file removed from `--modules-folder` location instead of `node_modules`.

Not easy to test in our framework but should be manually tested; make sure `--modules-folder` works as a relative and absolute paths. I think the biggest risk in this PR is that I changed a couple `path.join`s to `path.resolve` to handle the case where `modules-folder` is an absolute path. I think it's working correctly everywhere, but something to keep an eye out for.

PR #7151, opened 2019-03-27
fix(run): add cwd/node_modules/.bin to run command search path

### LINKED ISSUE #7126: yarn <executable> with workspaces doesn't work anymore

I recently upgraded to yarn 1.15.2 from 1.13.0 and now running executables inside `<current_workspace>/node_modules/.bin/` no longer works. It looks like it is always looking for executables in the root workspace.

In a repo with yarn workspaces, `cd` into one of the workspaces and try to run a command that only exists in that workspace's `./node_modules/.bin`.

```
> yarn react-native
yarn run v1.15.2
error Command "react-native" not found.
```

```
> ls ./node_modules/.bin/react-native
./node_modules/.bin/react-native
```

That it runs the executable in the current folder's `./node_modules/.bin`.

yarn: v1.15.2
node: v8.14.0
os: macOS Mojave 10.14

rally25rs:
**Summary**

This fixes a bug that was introduced in #6850 where the bin path was being built only from
`config.lockfileFolder`. However in workspaces, bins may not be hoisted to the workspace root,
causing bins to not be found. This change adds `config.cwd` to the bin search path, so the `yarn
run` command will look in a workspace package's node_modules, as well as the workspace root.

fixes #7126

**Test plan**

Changed as existing test that placed a bin at both the workspace root and in a package. Split into 2 tests, one with the bin only at the workspace root, the other with the bin only in the package. This way we are testing that we can detect the bin in either location.


farooqu:
Can this be prioritized please? It is totally broken right now for workspaces mono repos.

Amy-Lynn:
Can this please get prioritized/merged? This is a huge issue in mono-repos

jasonmobley:
@rally25rs can you update this PR to resolve the conflicts?
@arcanis can you review? This fixes a regression introduced by #6850 which you reviewed

In my project, I'm currently pinned on Yarn 1.14 because of the regression caused by #6850. I'd love to get back to current if we could get this PR through. The change seems ready to go if we got the conflicts fixed and could get review on it. Thanks!

rally25rs:
I'll resolve it this evening and get this merged. Thanks @arcanis and @jasonmobley

rally25rs:
@arcanis the "yarn acceptance tests" had 1 failing test, only on node 8, however those tests pass locally on node 8 for me. Is there a way to re-run that test and see if it was just a hiccup in the CI test? I didn't see a "rerun" button anywhere for that test suite.

arcanis:
For some reason I cannot log into Azure, it makes an infinite loop of signin in and immediately signing out ... 🤷‍♀️ The timeout doesn't look critical to me, let's merge. I'll make a patch release later this week.

simonbuchan:
How does this work with `cd /a/packages/b/c && yarn --cwd d some-command`? I would expect it to run `/a/packages/b/node_modules/.bin/some-command` in `/a/packages/b/c/d`, assuming `/a/packages/b/package.json` exists.

rally25rs:
`cwd` is added to the `PATH` before the lockfile directory (the workspace root) so I believe it will use a bin from the `cwd` over one from the workspace.

simonbuchan:
Not my question. To me it looks like the current code will look in `/a/packages/b/c/d/node_modules/.bin` or `/a/packages/b/c/node_modules/.bin` first (depending on when `--cwd` is applied), since `cwd` is often *not* the package directory.

Issue #6894, opened 2019-01-09
yarn run --if-present option flag to avoid exiting with a non-zero exit code when the script is undefined


HavaH123:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the current behavior?**
`yarn run` does not have an option to exit with a zero code if the script run is not defined in `package.json`.

My use case is that I manage a mono-repo using yarn and lerna and I use something like 
`lerna exec --scope -- yarn run script-name` or run `yarn run script-name` in a loop over all the packages.

This is fragile because it breaks if any of the packages dont have the `script-name` defined.

**What is the expected behavior?**

Similar to npm, provide a --if-present option that exits with a zero code if the script is undefined.
`yarn run --if-defined script-name`

**Please mention your node.js, yarn and operating system version.**
```
$ node -v
v8.15.0
$ yarn -v
1.12.3
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G3025
```

astorije:
`npm run` has a `--if-present` flag, and it would be really handy if there was feature parity on this. Is there another way to achieve that with Yarn?
(Our use case is a shared CI configuration that runs a few scripts such as `yarn test` and `yarn lint` on a few repositories, not all of which have all those scripts)

rally25rs:
Thanks for the info about npm's `--if-present` flag. I didn't know that existed. I think this is a good idea. Care to work up a PR? 😸

Tranced:
@rally25rs Hey, I'd like to take this on!

peijiesim:
Hi @Tranced, are you actively working on this? If not, I'd like to give it a shot. Thanks.

KingScooty:
Any news on this? I'm really missing this feature at the moment.

kachkaev:
Also interested in having `--if-present`. I'm maintaining a monorepo with a few packages. Some of them require build before publishing, but some do not due to their simplicity. Some are in TS, some are in pure JS.  Some have tests, some do not. What I want to do in CI is to call top-level `build`, `lint:tsc` and `test` commands without caring if they exist in every package. 

```js
// top-level package.json
"scripts": {
  "build": "yarn workspaces run --if-present build",
  // ...
  "lint:tsc": "yarn workspaces run --if-present lint:tsc",
  "test": "yarn workspaces run --if-present test"
},
```

[...]

PR #7159, opened 2019-03-31
Added --if-present option for scripts

ulentini:
**Summary**
Added `--if-present` option flag in order to allow undefined scripts to run without exiting with a non-zero code.
This addresses #6894.

**Test plan**
Added a couple of tests in `run.js` test file.

### DIFF
--- a/__tests__/commands/__snapshots__/run.js.snap
+++ b/__tests__/commands/__snapshots__/run.js.snap
@@ -51,3 +51,18 @@ Array [
   },
 ]
 `;
+
+exports[`runs non-existent script with parameter --if-present 1`] = `
+Array [
+  Object {
+    "data": "package.json: No license field",
+    "error": true,
+    "type": "warning",
+  },
+  Object {
+    "data": "Command \\"foo\\" not found.",
+    "error": true,
+    "type": "warning",
+  },
+]
+`;
--- a/__tests__/commands/run.js
+++ b/__tests__/commands/run.js
@@ -220,3 +220,11 @@ test('runs script with custom script-shell', (): Promise<void> =>
       customShell: '/usr/bin/dummy',
     });
   }));
+
+test('runs non-existent script with parameter --if-present', (): Promise<void> =>
+  runRun(['foo'], {ifPresent: true}, 'bin', (config, reporter) => {
+    expect(reporter.getBuffer()).toMatchSnapshot();
+  }));
+
+test('return non-zero exit code if script does not exist and is executed without --if-present', (): Promise<void> =>
+  expect(runRun(['foo'], {}, 'bin')).rejects.toBeDefined());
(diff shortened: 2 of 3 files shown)


arcanis:
Not a fan. You shouldn't call the script in the first place, or should use `|| true` to silence errors if you need it.

ulentini:
> Not a fan. You shouldn't call the script in the first place, or should use `|| true` to silence errors if you need it.

It was a feature request on #6894 to replicate an npm feature.
By the way, just out of curiosity, why do you think silencing with `|| true` would somehow be different than implementing this flag?

charlie-rudenstal:
Wouldn't silencing with `|| true` also silence any errors if the script exist but fails? There's definitely a use-case for exiting with 0 if the script just doesn't exist, especially for things like generalized CI-pipelines.

imflavio:
This would be very much appreciated

Fi1osof:
+1

iamomiid:
+1

builtbybrayne:
Any chance the build error could be checked and this passed? This would be a lovely bit of functionality.

risenforces:
Any progress? This feature is really good

michaelfaith:
Why did this go stale?  Would love to have this available.

Papooch:
@arcanis Please reconsider this. It has been demonstrated here and in the linked issue that it's an useful feature, especially in a CI environment. I don't see why there's such an opposition to this from your side, especially since `npm` supports it and the implementation is fairly straightforward.

fob-mobility:
I hate it, always the same story. NPM doesnt support nohoist so I switch to yarn. Yarn then doesnt support something else because someone is not a fan of something he could simply merge and not use. So here I go again searching for a good package manager.

danielbayley:
> Not a fan. You shouldn't call the script in the first place, or should use `|| true` to silence errors if you need it.

@arcanis You have missed the point entirely. `|| true` does not cover the very real use case of running the script only if it exists, not just silently failing the script, weather it exists or not. Both `npm` and `pnpm` have this flag/feature.

ryanewtaylor:
We have 50+ jobs using Jenkins Shared Libraries and the --if-present option would certainly make things easier as some repos have tests and some do not. However, for anyone encountering this on Jenkins on Windows I found a workaround that skips the Test stage if "test" is not defined in the package.json's scripts.

It's rather lengthy but seems to work in our case.

```groovy
stage("Test") {
    when {
        expression {
            // skip stage when "test" scripts are not defined
            def scriptOutput = pwsh(returnStdout: true, script: '''
                $script = Get-Content "package.json"
                | ConvertFrom-Json 
                | Select-Object -ExpandProperty "scripts"
                | Select-Object -ExpandProperty "test"
                $?
            ''').trim()
            return scriptOutput == 'True'
        }
    }
    steps {
        pwsh "yarn test"
    }
}
```

danielbayley:
> So here I go again searching for a good package manager.

@fob-mobility https://pnpm.io 😉

artursapek:
> You shouldn't call the script in the first place

this becomes unmanageable in a large codebase when you're using `yarn workspace` to run tests. how has this change not been merged lol

PR #7163, opened 2019-04-01
Retry on connection timeout

VincentBailly:
**Summary**

yarn install randomly fails with a ETIMEDOUT error even though we have not reached the user defined timeout. In some cases, ETIMEDOUT error is not due to user defined timeout but to network instabilities, in such cases we should retry connecting (as we already do for ESOCKETTIMEDOUT errors)

**Test plan**

We tested this fork in production for a few days, no new issues or perf regression was observed, and the issue was fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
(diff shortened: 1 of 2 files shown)


VincentBailly:
See issue #7163

arcanis:
LGTM, thanks 👍

PR #7207, opened 2019-04-19
Runtime wrapper handling

GawainLynch:
Summary
-------

On each run of Yarn, a directory is created under the OS tmpdir with the following format:
```
path.join(os.tmpdir(), `yarn-${prefix || ''}-${Date.now()}-${Math.random()}`);
```

As these directories are not currently cleaned up by Yarn, they quickly become numerous and start eating up space. This can be problematic when the tmpdir has been mounted on `ramfs`/`tmpfs`, or on a CI machine that runs endlessly … with nobody watching disc usage … 'cause "it's just CI, not my job".

This PR offers two ways to approach this, happy to drop one, or the other, or both if it isn't a match for the project :smiley_cat: 

### Option 1 — `--wrapper-folder <path>`

This allows the end-user to optionally specify a location to dump this data, e.g. a nice big chunk of slow storage.

### Option 2 — Remove wrapper folder post-execution

I have my mother in my head right now reminding me to clean up my Lego when I was done playing. That memory & its metaphor seems to do a good job of explaining the rationale here.

Test plan
---------

Test all the thingz!

Seriously, I'm a huge believer in writing good quality tests and will gladly add them to this PR. I held back on adding them to the initial submission as I wasn't sure where/what would be a good fit.

So if the PR is deemed to be eligible for inclusion, I'll happily add some tests if someone could, in exchange, give me a very quick pointer on where the coverage should go.

Thank you all for a wonderful tool! :+1:

### DIFF
--- a/__tests__/commands/_helpers.js
+++ b/__tests__/commands/_helpers.js
@@ -88,6 +88,7 @@ export function makeConfigFromDirectory(cwd: string, reporter: Reporter, flags: 
       ignoreScripts: flags.ignoreScripts,
       globalFolder: flags.globalFolder || path.join(cwd, '.yarn-global'),
       cacheFolder: flags.cacheFolder || path.join(cwd, '.yarn-cache'),
+      wrapperFolder: flags.wrapperFolder || path.join(cwd, '.yarn-wrapper'),
       linkFolder: flags.linkFolder || path.join(cwd, '.yarn-link'),
       prefix: flags.prefix,
       production: flags.production,
--- a/src/cli/commands/global.js
+++ b/src/cli/commands/global.js
@@ -51,6 +51,7 @@ async function updateCwd(config: Config): Promise<void> {
     binLinks: true,
     globalFolder: config.globalFolder,
     cacheFolder: config._cacheRootFolder,
+    wrapperFolder: config.wrapperFolder,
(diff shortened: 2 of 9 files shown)


DarrenCook:
Disclaimer: I know nothing about yarn internals.

Option 2 seems the more sensible one.
I took a look at your code (but haven't run it). Looks fine. It seems src/util/fs.js turns the `fs` functions into promisified versions, so my only thought was if you should be waiting for that promise to resolve? But your patch is inside a Promise constructor... and I'm not sure what complexity that adds.

I wonder if your unlink line can instead be tacked on to the Promise being returned, something like this:

```
  return queue.push(
    key,
    (): Promise<string> =>
new Promise((resolve, reject) => {
   //...
   }
   .catch(e => {fs.unlink(opts.env[constants.WRAPPER_FOLDER]);throw e;} )
   .then( s => fs.unlink(opts.env[constants.WRAPPER_FOLDER]) )
   //...
```

(Just adding the `then()` felt nice and clean; but I realized it would not delete the directory if an error was being thrown, so I've added the catch and rethrow.)


Anyway, thanks for taking the initiative with a patch - I've also been wondering why my /tmp/ directory is full of yarn directories! I hope someone gets back to you with advice on suitable tests.

Issue #1648, opened 2016-11-03
Installer does not set PATH correctly on windows


nukedbit:
After installing yarn on my machine i noticed that the command was not found because the content was installed on C:\Program Files (x86)\Yarn\bin, but PATH was set to be under User AppData folder, manually setting the environment variable fixed the issue.

**OS:** Windows 10 (1511 build 10586.164)
**Node:** v6.5.0

dhruvdutt:
Did you install with chocolatey or normal MSI installer?

nukedbit:
Sorry i forgot to mention that i installed it with MSI

Daniel15:
Strange... It should definitely be adding `c:\Program Files (x86)\Yarn\bin` to the path, see here: https://github.com/yarnpkg/yarn/blob/master/resources/winsetup/Yarn.wxs#L30-L38

Note that `c:\Program Files (x86)\Yarn\bin` is added to the **system** path, while the AppData path is added to the **user** path. The AppData path is for modules you install globally through Yarn, while the system path is for Yarn itself.

raniesantos:
Hi @Daniel15 I read your comment above about the two different paths for program files and for appdata. What exactly is the purpose of the user/appdata yarn path? In the image I posted below, my environment variable editor displays an error saying 'Invalid path'.

What exactly does this mean and will it cause any problems in my usage of Yarn?

![capture](https://cloud.githubusercontent.com/assets/8528269/20024794/bcdc6c64-a323-11e6-801e-5c2bf86226da.JPG)

Daniel15:
> What exactly is the purpose of the user/appdata yarn path? 

It's supposed to be for packages you install globally (eg. `yarn global install foo`), I think there's some bugs at the moment and it's placing those modules in the wrong directory. Yarn should still work without it.

jcbvm:
I had the same problem, on my laptop there was no problem, but for my desktop at work I had to add an extra backslash to the end of the path, so `\bin\` instead of `\bin`

Jake-Shadle:
@jcbvm's fix worked for me as well.

ptpaterson:
Installation with the MSI (downloaded today) installed to `C:\Program Files (x86)\Yarn\bin\`, but the PATH variable was updated to `\AppData\Local\Yarn\.bin`.  I may be more confused by the extra period in `\.bin`.

changing PATH variable to `C:\Program Files (x86)\Yarn\bin\` (noting extra `\` at the end) solved the problem.

[...]

PR #7209, opened 2019-04-19
Remove trailing slash from path element

magnushiie:
**Summary**

Every character in Windows environment is a precious resource, should not append an unnecessary slash at the end.

Furthermore, it's inconsistent with the missing trailing slash in the AppData path a few lines below.

### DIFF
--- a/resources/winsetup/Yarn.wxs
+++ b/resources/winsetup/Yarn.wxs
@@ -35,7 +35,7 @@
 				<Environment
 					Id="Path"
 					Name="PATH"
-					Value="[INSTALLDIR]bin\"
+					Value="[INSTALLDIR]bin"
 					Permanent="no"
 					Part="last"
 					Action="set"


magnushiie:
Note, this reverts a recent change by @derekellison #5387 - that PR mentions that the trailing slash is "as per windows path requirements", there is no requirement for a trailing slash in PATH elements - most of the Windows system paths do not have a trailing slash.

arcanis:
Ping @derekellison - I don't use Windows so I can hardly vouch for any approach here. #5387 mentions it fixes a bug, can you detail which one and how that's a fix?

magnushiie:
I've been using a build with this change for over 2 weeks and it works fine. Maybe the bug was just referring to the opinion that path elements should have a trailing slash.

I would like to get this PR off my pending PR list, can we merge this?

magnushiie:
@arcanis it seems @derekellison is not here to respond - can we merge this? I think #5387 wasn't a bug fix and was a mistake.

derekellison:
This was a bug in windows a few years ago - not sure if its still an issue. See this for background of the issue: https://github.com/yarnpkg/yarn/issues/1648

magnushiie:
I believe these comments do not support the conclusion that adding a backslash fixed it - it was probably rather the manual updating the environment variables via the UI that triggered the refresh. I didn't see there any evidence of a process having a PATH correctly set up (without the trailing backslash) and not being able to find yarn.

PR #7258, opened 2019-05-09
Add yarn resolve-merge command. Resolves #5382

### LINKED ISSUE #5382: can't resolve merge conflicts with --install.frozen-lockfile in .yarnrc
feature?

If a project has a .yarnrc file that specifies `--install.frozen-lockfile false`, then you can't take advantage of the automatic merge conflict resolution on `yarn install`.

Yarn either provides a way to override the settings in .yarnrc during `yarn install` (e.g. `yarn install --no-frozen-lockfile`), or provides a new CLI command to explicitly resolve merge conflict markers (e.g. `yarn resolve-merge`).

jack-guy:
**Summary**
Currently the only way to automatically resolve conflict markers in yarn.lock is through a `yarn install`. This is a lot of overhead, and also requires developers to turn off `frozen-lockfile`.

This PR adds a dedicated command for just running the merge resolution process on yarn.lock. This is especially useful in the case of custom git merge drivers which can run much quicker with a dedicated command like this.

**Test plan**

Clean `yarn test`

----

This is my first PR so I'm still learning the ropes a bit. In particular, not sure what the expectations are for a "test plan", when to make updates to CHANGELOG.md, and where to update the website documentation.


arcanis:
Thanks for your work! 😃

That being said, I'm not entirely sure we really want to expose a dedicated command for this, though. After a merge is resolved you need to run a full install anyway, so I don't think this would have a lot of value in the general case.

jack-guy:
Hi @arcanis, in my view the action of resolving merge conflicts is distinct from installing dependencies that match yarn.lock (there are a couple comments in the linked issue that concur). The primary use case for merge conflict resolution is in version control workflows, which typically do not require dependency installation.

To give you an example, in my company's case we're trying to automate merge conflict resolution between many branches that touch yarn.lock at once before deploying them all together. To run a full `yarn install` in a git custom merge driver can easily add several minutes to what should be a lightweight commit action.

amakhrov:
Any progress with that? What would it take to at least have a decision on whether this feature should be merged to mainstream or not in general?

arcanis:
I don't think we'll merge this. The use case it relies on is fringe (automated remote merge conflit resolution), and given that we are full speed on the v2 development I don't want to add commands that we would remove a major later. 

Note that in the v2 installs are much faster (PnP is the default), so the rational for splitting the two actions is even more blurry.

amakhrov:
@arcanis thanks for the quick reply!
Does it also mean that v2 will have addressed the issue where `frozen-lockfile` prevents auto merge conflict resolution?

arcanis:
No, because `frozen-lockfile` implies that the lockfile must not be allowed to change, and merge conflict resolution is a change as any other.

That said I'm not sure why you'd want to run the merge conflict resolution on a system configured to use

arcanis:
No, because `frozen-lockfile` implies that the lockfile must not be allowed to change, and merge conflict resolution is a change as any other.

That said I'm not sure why you'd want to run the merge conflict resolution on a system configured to use frozen-lockfile... And why don't you just remove the flag from your CLI when invoking Yarn?

amakhrov:
> And why don't you just remove the flag from your CLI when invoking Yarn

I guess, a matter of convenience. Not a bid deal, though. So yeah, looks like that would be the best option - once the cli argument it's available in v2.

Thanks for the clarification.

jack-guy:
Hey @arcanis appreciate that y'all are making careful design decisions in light of yarn v2. If I may make one more case for this, I don't think the use case is fringe. A command such as `resolve-merge` is in my view the most natural way to work with yarn.lock merge conflicts in version control workflows. There's little about the `install` command that implies it should have this responsibility, given that managing files in git exists as an important and typical workflow that is entirely removed from the actual installation of dependencies. In this way the merge resolution behavior of `yarn install` feels very much like a side effect at the moment.

One other related observation is that if you're configuring `yarn` with `.yarnrc` there's no way to turn off the `--frozen-lockfile` behavior from the CLI. I find myself having to go in to the config file and temporarily comment out the flag. That's behavior that I think would be good to change regardless :)

juliosaraiva (approved):

Issue #7012, opened 2019-02-07
Yarn audit does not work with an external repository


adamscybot:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

When running `yarn audit` with the `--registry` flag, it uses the main yarn repo anyway.

**If the current behavior is a bug, please provide the steps to reproduce.**

Execute `yarn audit --registry https://registry.npmjs.org/`

**What is the expected behavior?**

Audit calls are made to the defined registry. This is especially important for everyone with third party repo's where the outside world it blocked on CI agents.

**Please mention your node.js, yarn and operating system version.**

Mac Os Mojave
Yarn 1.13.0
Node 10.15.1

abhisheksoni27:
What exactly is happening here? https://github.com/yarnpkg/yarn/blob/0f0520380a60be3ba347774cc6e07fcea443d6d5/src/cli/commands/audit.js#L211-L225

Do you have an endpoint configured that automatically calls `npm`'s audit registry ?

agierlicki:
Hi,

i really want to get into OSS and this would be my very first contribution. Would it be okay if I worked on this issue?

Cheers,
Alex

gaiazov:
@abhisheksoni27 on line 217 the `YARN_REGISTRY` constant is used, which corresponds to `https://registry.yarnpkg.com` https://github.com/yarnpkg/yarn/blob/0f0520380a60be3ba347774cc6e07fcea443d6d5/src/constants.js#L21

pratyushj:
Hey, can I try to solve this bug?

esimakin:
Hi @pratyushj, the issue already has its PR #7263. Thanks!
In addition to that, it has a duplicate PR #6484

MadaraUchiha:
This is still open in 2022, any update on this?

GeorgeWL:
any update? still seeing this in March 2022

GodOfCoding1:
Hi , Is this still open in 2022?

srcimon:
Still open in 2023.

M1chael:
Greetings from 2k24.

superblangel:
Hello from 2025

boryscastor:
Hello 2026

PR #6484, opened 2018-10-04
Allow any registry to be used for auditing

cheesestringer:
**Summary**

Allow auditing against any registry. This is useful for auditing against a private or whitelisted registry.
For example, in our organisation npmjs is allowed while yarnpkg is blocked :(

**Test plan**

Command: `yarn audit`
Output: `An unexpected error occurred: "https://registry.yarnpkg.com/-/npm/v1/security/audits: getaddrinfo ENOTFOUND registry.yarnpkg.com registry.yarnpkg.com:443".`

Command: `yarn audit --registry https://registry.npmjs.org`
Output: `0 vulnerabilities found - Packages audited: 46500`

Command: `yarn audit --registry https://artifactory.dev.company/artifactory/api/npm/npm-dev`
Output: `error Your configured registry "https://artifactory.dev.company/artifactory/api/npm/npm-dev" does not support audit requests.`

PR #7263, opened 2019-05-11
:sparkles: Add support for --registry with audit command

charlyx:
**Summary**

Hi 👋 

First of all, thanks for your work!
I use **yarn** everyday and I love its workspaces feature 😄 

This PR adds support for the `--registry` flag when running `audit` command as described in #7012 .

**Test plan**

I have added a test case where I check if the audit command is calling the registry specified with the high flag `--registry`.

### DIFF
--- a/src/cli/commands/audit.js
+++ b/src/cli/commands/audit.js
@@ -19,6 +19,7 @@ const gzip = promisify(zlib.gzip);
 export type AuditOptions = {
   groups: Array<string>,
   level?: string,
+  registry?: string,
 };
 
 export type AuditNode = {
@@ -145,6 +146,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   const audit = new Audit(config, reporter, {
     groups: flags.groups || OWNED_DEPENDENCY_TYPES,
     level: flags.level || DEFAULT_LOG_LEVEL,
+    registry: flags.registry,
   });
   const lockfile = await Lockfile.fromDirectory(config.lockfileFolder, reporter);
   const install = new Install({}, config, reporter, lockfile);
@@ -238,7 +240,7 @@ export default class Audit {
 
   async _fetchAudit(auditTree: AuditTree): Object {
     let responseJson;
-    const registry = YARN_REGISTRY;
+    const registry = this.options.registry || YARN_REGISTRY;
     this.reporter.verbose(`Audit Request: ${JSON.stringify(auditTree, null, 2)}`);
     const requestBody = await gzip(JSON.stringify(auditTree));
     const response = await this.config.requestManager.request({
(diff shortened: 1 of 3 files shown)


PtrTn:
Thanks for creating this PR. Personally I would love to have this feature merged ❤️

wesselvanderlinden:
I would really like this PR to be merged too :D

src/cli/commands/audit.js
```diff
@@ -145,6 +146,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
   const audit = new Audit(config, reporter, {
     groups: flags.groups || OWNED_DEPENDENCY_TYPES,
     level: flags.level || DEFAULT_LOG_LEVEL,
+    registry: flags.registry,
```
esimakin: Why don't you add `|| YARN_REGISTRY` here instead of in the `_fetchAudit` method?
Here it'll look more consistent, in my opinion.

dobriai:
This explicit command line option is fine, but why isn't the registry specified in `.yarnrc` used for audit in the first place?
Yarn appears to use a custom registry for pulling down packages, but insists on its own, hard-coded URL for audit - why?

src/cli/commands/audit.js
```diff
@@ -238,7 +240,7 @@ export default class Audit {
 
   async _fetchAudit(auditTree: AuditTree): Object {
     let responseJson;
-    const registry = YARN_REGISTRY;
+    const registry = this.options.registry || YARN_REGISTRY;
```
dobriai: Even better, maybe add something like `|| String(this.config.getOption('registry')).replace(/\/+$/, '')` in the middle, so we also also consult the config file(s)? (_Credit to PR #6484 for that!_)

gbhasha:
When this PR will be merged?
It would be great, it also uses the registry specified in the `.yarnrc`  file for audit.

schdief:
would love to see this merged, thx for your effort guys!

BLoeT:
Hey Guys, 

Unfortunately we need this feature for our project.

Could you pls merge this, if everything is good?

nemoDreamer:
Maybe this PR needs a push so that the **Yarn Acceptance Tests** pass? Seems like `yarn` team might not look at it until it does? 😭

funilrys:
I had a little giggle about this. I looked at the source code before the #7012. 

The patch looks good - in my opinion. @charlyx can you (or someone with high privileges) close and reopen this PR. It will reinforce the run of the checks. 

Cheers!

schmidlidev:
Any update on this?

nemoDreamer:
Uhm, @charlyx, the original request from @funilrys was if you could close and RE-OPEN this PR, to kick the build checks?

struys:
@funilrys @charlyx mind taking another look at this?

PR #7272, opened 2019-05-14
Handle optional dependencies alongside resolutions

### LINKED ISSUE #6040: Bug: "resolutions" trumps `--ignore-optional`

*bug*

Currently, when there are resolutions defined, `--ignore-optional` is ignored and the optional dependencies are installed anyway.

You can test it with a simple package.json:
```json
{
  "name": "test-optional",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "optionalDependencies": {
    "fsevents": "^1.0.0"
  },
  "resolutions": {
    "fsevents": "^1.2.4"
  }
}
```
on `yarn install --ignore-optional` the above will lead to fsevents being installed nonetheless.

On Linux (Ubuntu Xenial) for example, this would fail the yarn install due to:
```
error fsevents@1.2.4: The platform "linux" is incompatible with this module.
```

`yarn install --ignore-optional` works as expected with:
```json
{
  "name": "test-optional",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "optionalDependencies": {
    "fsevents": "^1.0.0"
  }
}
```
and does not install `fsevents`.

No optional dependencies to be installed with `--ignore-optional`, no matter whether they are mentioned in `resolutions` or not.

yarn 1.7.0
Node 10.5.0
OSX High Sierra and Debian Xenial

olingern:
**Summary**
Read my comment [here](https://github.com/yarnpkg/yarn/issues/6040#issuecomment-492184689) for a summary of why and how I arrived at this.

Fixes #6040

I would consider this a WIP / draft since I will need to tackle tests and verify workspaces + resolutions are still intact.

TODOS:
- [ ] Get feedback. Does this make sense?
- [ ] Add tests for this case ( optional dependency / resolved )
- [ ] Add to changelog

**Test plan**

See ^ todos :-)


src/cli/commands/install.js
```diff
@@ -283,13 +284,6 @@ export class Install {
       Object.assign(this.resolutions, projectManifestJson.resolutions);
       Object.assign(manifest, projectManifestJson);
 
-      this.resolutionMap.init(this.resolutions);
```
olingern: This is moved to after `pushDeps` is so that we know what dependencies to not add

src/cli/commands/install.js
```diff
@@ -202,6 +202,7 @@ export class Install {
     this.integrityChecker = new InstallationIntegrityChecker(config);
     this.linker = new PackageLinker(config, this.resolver);
     this.scripts = new PackageInstallScripts(config, this.resolver, this.flags.force);
+    this.depMap = new Map(); // Keep track of dependencies for access by key
```
olingern: Probably not a good name. Didn't see the other local `depMap` until re-reading during lint fixes.

src/cli/commands/install.js
```diff
@@ -306,6 +301,9 @@ export class Install {
           return;
         }
         const depMap = manifest[depType];
+        const workspaceName = manifest.name;
```
olingern: Done to pass linter line length below

olingern:
Closing in favor of #7273

PR #7273, opened 2019-05-14
fix(install): Inherit optional property for resolutions

### LINKED ISSUE #6040: Bug: "resolutions" trumps `--ignore-optional`

kaylie-alexa:
Fixes #6040

**Summary**
Resolutions are a required for installing when there is a matching pattern in the resolutions map. When the resolver tries to dedupe dependencies and finds an optionalDependency that matches a resolution pattern, they merge into one no-longer optional request. This PR has resolutions inherit `optional` property if the dependency exists in the top level package.json. 

**Test plan**

I've added a test case under resolutions

### DIFF
--- a/__tests__/commands/install/resolutions.js
+++ b/__tests__/commands/install/resolutions.js
@@ -46,6 +46,12 @@ test.concurrent('install with --frozen-lockfile with resolutions', async (): Pro
   }
 });
 
+test.concurrent('install with resolutions on optional dependencies should not resolve', (): Promise<void> => {
+  return runInstall({ignoreOptional: true}, {source: 'resolutions', cwd: 'optional-deps'}, async config => {
+    expect(await isPackagePresent(config, 'left-pad')).toEqual(false);
+  });
+});
+
 test.concurrent('install with exotic resolutions should override versions', (): Promise<void> => {
   return runInstall({}, {source: 'resolutions', cwd: 'exotic-version'}, async config => {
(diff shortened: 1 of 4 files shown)


coreyward:
@rally25rs @arcanis Any guidance on when we might get this reviewed and merged?

Issue #6929, opened 2019-01-16
yarn audit returns 503 every now and then


luc-tielen:
**What is the current behavior?**

We use `yarn audit --json` in our CI pipeline to check for vulnerabilities, however every now and then it fails with the following error:

Output from command:
```
// many more lines of JSON ...
{"type":"activityTick","data":{"id":0,"name":"yaml-loader@^0.4.0"}}
{"type":"activityTick","data":{"id":0,"name":"js-yaml@^3.5.2"}}
{"type":"activityEnd","data":{"id":0}}
/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:66237
            throw new (_errors || _load_errors()).ResponseError(_this3.reporter.lang('requestFailed', description), res.statusCode);
            ^

Error: Request failed "503 Service Unavailable"
    at ResponseError.ExtendableBuiltin (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:702:66)
    at new ResponseError (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:808:124)
    at Request.params.callback [as _callback] (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:66237:19)
    at Request.self.callback (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:129397:22)
    at Request.emit (events.js:188:13)
    at Request.<anonymous> (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:130369:10)
    at Request.emit (events.js:188:13)
    at IncomingMessage.<anonymous> (/usr/local/Cellar/yarn/1.13.0/libexec/lib/cli.js:130291:12)
    at Object.onceWrapper (events.js:276:13)
    at IncomingMessage.emit (events.js:193:15)
```

Also using `--verbose` flag, it seems to fail when accessing https://registry.yarnpkg.com/-/npm/v1/security/audits
This error only seems to occur during (busy?) parts of the day, maybe server can't handle the load?

**If the current behavior is a bug, please provide the steps to reproduce.**

Steps to reproduce: 
1. run `yarn audit --json` until you hit the error.

**What is the expected behavior?**

No 503 error.

ArthurClemens:
I am seeing this too

Fleker:
+1 to having a command line argument to pass or do something to continue the CI process.

ptondereau:
[...]

PR #7281, opened 2019-05-16
Add --fail-silently flag in audit command.

ptondereau:
**Summary**
This PR add a flag on audit command in order to fail silently when there is a network error or when parsing the body's response. This is very useful in CI environment when you need to restart your test suite because of a 503 error.
It also solves #6929
Please be gentle because I'm new to Node.js ecosystem.

example of usage:
```
$ yarn audit --fail-silently
```
**Test plan**

I fail to write unit test with this flag because I don't know how to achieve this with your libraries.
But manually, I tried:
1. Add an entry in my /etc/hosts with `127.0.0.1 registry.yarnpkg.com`
2. Run `$ yarn audit` and `echo $?` exited 1
3. Run `$ yarn audit --fail-silently` and `echo $?` exited 0
And it worked.


Liinkiing:
Any news on this? I think that it could be useful 👍

vfonic:
Thumbs up for getting this functionality merged in!

I'd name this argument differently, as it might be confusing whether it fails silently always or just on network issue.

ptondereau:
Me, waiting for a review:
image

webbertakken:
@arcanis could you please merge this? 

Currently CI all over the world have to do one of three things:
- Occasionally restart because of a 503 
    - result: not very modern, people get frustrated and stop trusting pipelines.
- Write a wrapper around yarn that retries failed attempts to audit
    - It is not always straight forward to wrap every process that invokes yarn
- Remove audit flag entirely in CI
    - which means no uniform `.yarnrc` file containing `--install.audit true`

This PR doesn't opt-in by default, but allows many people to move forward.

richardsimko:
+1 on wanting this merged. We're approaching the 1 year birthday for this PR. Hope no cake is needed. 🍰

ptondereau:
@richardsimko haha but no problem for an Apple pie

PR #7289, opened 2019-05-20
Correctly install workspace child deps when workspace child not symlinked to root 

### LINKED ISSUE #4289: Packages that are both fetched from npm in the root, and packages in the workspace are not symlinked consistently
Bug, since it seems like this is related somehow to specifically what dependencies are in use. Hard to say though.

Reproduction repository: https://github.com/loganfsmyth/__yarn-workspace-bug

The structure 
```
// Root package.json has dep on:
babel-core@7.0.0-alpha.18
babel-cli@7.0.0-alpha.18
```
and
```
// Packages:
packages/babel-core@7.0.0-alpha.19
packages/babel-cli@7.0.0-alpha.19 depending on babel-core@7.0.0-alpha.19
```
produces
```
node_modules/
    babel-core@7.0.0-alpha.18
    babel-cli@7.0.0-alpha.18
```

There seems to be some installation issues when the root package depends on older versions of packages from npm, and they also exist in the yarn workspace packages themselves, they don't have their dependencies symlinked to the related packages.

https://github.com/loganfsmyth/__yarn-workspace-bug

See repo. Packages should be symlinked as would be normally expected.
```
node_modules/
    babel-core@7.0.0-alpha.18
    babel-cli@7.0.0-alpha.18
packages/babel-cli/node_modules/
    babel-core@7.0.0-alpha.19 -> ../../babel-core
```

Node: 6.9.1
Yarn: 0.27.5
OS: OSX 10.12.5

danez:
This fixes #4289.

**Problem**

The scenario is the following:

- We have a workspace repository which has one workspace child named `a` with version `2.0.0`
- `a@2.0.0` has a dependency on `b@2.0.0`
- the root has a dependency on `a@1.0.0` and `b@1.0.0` (both coming from the registry)
- `a@1.0.0` and `b` do not have dependencies.

When install is called the hoister realizes it cannot hoist `a@2.0.0` (the ws child) to the root because `a@1.0.0` is already installed there. The path for the dependency `b@2.0.0` will then be created as `_root_/node_modules/workspace-aggregator-1234/node_modules/a/node_modules/b/` which is completely wrong. The linker then installs the dependency there.

In the tests the behavior is different for some reason and yarn errors with `error "workspace-aggregator-1234#arrify#isarray" not installed`. But anyway this needs to be fixed.

**Proposed fix**

The fix I went for is located in the package-hoister. The hoister creates the tree as usual but in `init()` where it creates a flat tree we check for each dependency if it is actually coming from within a ws child. If it does we change its destination from 

```
_root_/node_modules/workspace-aggregator-1234/node_modules/_child_/node_modules/_dep_/
```

to

```
_root_/packages/_child_/node_modules/_dep_/
```

I also moved the skipping of workspace dependencies from the linker to the hoister, as I have to do the check now there anyway, because the change I described should not be applied to the virtual workspace or the workspace child itself. And imho it looks more correct to me to not even create a wrong path instead of creating it and then skipping it in the linker.

Note that if the dependency of the child is hoisted to the root it will not show up in the tree as subdependency for the ws child. So we don't have to worry about that case, it works exactly the same as now.

If you are worried about the move of the skipping I can readd it or move it back to the linker.

**Test plan**

I added a test case which fails without the fix.

> As soon as CI will finish failing with the test-only commit I will submit the fix.


src/package-hoister.js
```diff
@@ -835,6 +835,14 @@ export default class PackageHoister {
       // up of modules from different registries so we need to handle this specially
       const parts: Array<string> = [];
       const keyParts = key.split('#');
+      const isWorkspaceEntry = this.workspaceLayout && keyParts[0] === this.workspaceLayout.virtualManifestName;
+
+      // we don't need to install the virtual manifest or workspace packages
+      // which already have a different version in the root
+      if (isWorkspaceEntry && keyParts.length <= 2) {
```
arcanis: Where is the second part of the condition (`[...] which already have a different version in the root`)?

danez: It is dependent on the length:

So basically it is this:

```js
if (
  this.workspaceLayout && // Do we have a workspace?
  keyParts[0] === this.workspaceLayout.virtualManifestName && // Is the first keypart the virtual manifest name? If so we are dealing with modules from within workspaces or the virtual manifest itself
  (
    keyParts.length === 1 || // the key is "workspace-aggregator-1234" Which is the virtual workspace itself
    keyParts.length === 2  //  the key is "workspace-aggregator-1234#child" Which is one of the workspace childs which could not be hoisted.
  )
){}
```

danez:
I'm not sure what exactly you mean by never hoisting workspaces, but if you imply that ws childs should never be symlinked in the root node_modules folder I would think this is a pretty big breaking change.

src/package-hoister.js
danez: I also updated the comment to describe it better.

arcanis:
> but if you imply that ws childs should never be symlinked in the root node_modules folder I would think this is a pretty big breaking change.

Yep that's what I was mentioning. It would be breaking indeed - although only for already invalid usages (since relying on packages that aren't part of your own dependencies is forbidden). Anyway, I think it's good to merge as it is - the next major will take care of the breaking part 🙂

arcanis:
Thanks @danez! Release should be out soon.

Issue #6175, opened 2018-07-29
Bug: Postinstall Scripts of Workspace Packages are called within the node_modules of the workspace


hediet:
This bug comes up when trying to develop a vscode extension using yarn's workspace feature.
# Setup
Yarn v1.9.2, Windows 10, Node v10.6.0
Directory layout:
* ROOT
    * package.json
    ```json
    {
        "private": true,
        "workspaces": [ "vscode-ext", "common"]
    }
    ```
    * vscode-ext
      * package.json
    ```json
    ...
        "name": "vscode-ext",
        "scripts": {
            "postinstall": "yarn run vscode-install"
        },
    "devDependencies": { "vscode": "^1.1.6" }
    ...
    ```
    * common (unimportant)

Yarn install turns it into this structure:
* ROOT
    * node_modules
      * vscode-ext (symlink to ROOT/vscode-ext)
      * vscode
        * .bin
          * install
    * package.json
    * vscode-ext
      * package.json
      * node_modules
        * bin
          * vscode-install.cmd
    ```batch
    @IF EXIST "%~dp0\node.exe" (
    "%~dp0\node.exe"  "%~dp0\..\..\..\node_modules\vscode\bin\install" %*
    ) ELSE (
    @SETLOCAL
    @SET PATHEXT=%PATHEXT:;.JS;=;%
    node  "%~dp0\..\..\..\node_modules\vscode\bin\install" %*
    )
    ```
    * common (unimportant)

# Problem
The postinstall script of the vscode-ext package is run within `ROOT/node_modules/vscode-ext` and not within `ROOT/vscode-ext`.
Both paths point to the same directory, but relativ paths behave differently:

In `ROOT/node_modules/vscode-ext/node_modules/.bin/`,
`"%~dp0\..\..\..\node_modules\vscode\bin\install" %*` points to `ROOT/node_modules/node_modules/vscode/bin/install` which is clearly wrong.

In `ROOT/vscode-ext/node_modules/.bin/` however,
`"%~dp0\..\..\..\node_modules\vscode\bin\install" %*` points to `ROOT/node_modules/vscode/bin/install` which actually works.

This behavior breaks at least vscode extensions in yarn workspaces.

pawsong:
+1

pawsong:
@rally25rs This prevents us from using "postinstall" script :(

davidhouweling:
I also got a similar issue using an npm package like `rimraf` in my workspace package.

``` console
[...]

PR #7299, opened 2019-05-23
Fix wrong pwd path when running install commands in workspace. Fixes …

ninja-:
…#6175

**Summary**

Because pwd is wrong on windows when running postinstall scripts in workspace packages, if you have a script like: "install": "tsc -p .",
you will get an error

```
Error: Cannot find module 'C:\Users\xx\IdeaProjects\x\node_modules\node_modules\typescript\bin\tsc'
```

See #6175 for more debugging done by other users.
Note that there isn't a need to add this pattern in other places in code, because it's only related to the scripts generated by cmd-shim.

**Test plan**
I haven't tested yet if it breaks the relative path logic on Linux.
On Windows the issue is fixed.

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -135,7 +135,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
           stage,
           config,
           cmd: cmdWithArgs,
-          cwd: flags.into || config.cwd,
+          cwd: fs.realpathSync(flags.into || config.cwd),
           isInteractive: true,
           customShell: customShell ? String(customShell) : undefined,
         });


src/cli/commands/run.js
```diff
@@ -135,7 +136,7 @@ export async function run(config: Config, reporter: Reporter, flags: Object, arg
           stage,
           config,
           cmd: cmdWithArgs,
-          cwd: flags.into || config.cwd,
+          cwd: realpathSync(flags.into || config.cwd),
```
arcanis: Can you instead use `path.resolve`? I don't think we want to use `realpath` here

arcanis (changes requested):

src/cli/commands/run.js
ninja-: Pretty sure i tried it and it didnt work

3846masa:
I tried this PR code with [this repository](https://github.com/3846masa-tmp/issue-6175-yarn), however it caused error when installing package.

I tried on these environments, but `yarn install` failed.
- Windows 10 / Node.js v10.15.1 / MSYS 2
- Windows 10 / Node.js v10.15.1 / Command propmt
- Windows 10 / Node.js v10.15.1 / Powershell

ninja-:
hm, you are right that it doesn't work on your testcase, but it fixed my case which was almost the same. very weird

src/cli/commands/run.js
crowne: Doesn't work for me I get an error:
> error An unexpected error occurred: "realpathSync is not defined".

Same error when qualifying fs.realpathSync

`path.resolve` also didn't fix the issue, note the double node_modules below

> Cannot find module 'E:\projects\vscode-sqltools\node_modules\node_modules\vscode\bin\install'

Issue #4913, opened 2017-11-13
yarn global add's offline flag seems to be ignored in Windows


ruidfigueiredo:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
With the offline mirror correctly configured and populated (yarn config set yarn-offline-mirror ...)
yarn global add @angular/cli --offline 
or
yarn global add @angular/cli@specificVersion --offline --exact
will contact https://registry.yarnpkg.com/

**If the current behavior is a bug, please provide the steps to reproduce.**

On Windows:
Setup an offline mirror:
    yarn config set yarn-offline-mirror pathToMirror
Populate the mirror:
   yarn add @angular/cli
Turn off internet connection
   yarn global add @angular/cli --offline
Will fail while trying to contact https://registry.yarnpkg.org/

**What is the expected behavior?**
Should use the offline mirror as it does in Linux

**Please mention your node.js, yarn and operating system version.**
node version: 6.11.2
yarn version: 1.3.2
OS: Windows 10

BYK:
I think this is not a Windows-specific issue but simply a configuration issue. `yarn config set yarn-offline-mirror` would usually set the config in either the user's home directory or in the project directory and `yarn global` wouldn't use the one from the project. I'd argue `yarn global` should use the user's config from the home directory tho.

I've looked at the code and this seems to be the case. Can you make sure `offline-mirror` is set in your `.yarnrc` in your home directory and not in the project?

ruidfigueiredo:
It's picking up the correct config file, here's the output when running with `--verbose`

`C:\Users\rfigueiredo\Projects\temp\offline-new-ng>yarn global add @angular/cli --offline --verbose
yarn global v1.3.2
verbose 0.513 Checking for configuration file "C:\\Users\\rfigueiredo\\Projects\\temp\\offline-new-ng\\.npmrc".
verbose 0.515 Checking for configuration file "C:\\Users\\rfigueiredo\\.npmrc".
verbose 0.516 Checking for configuration file "C:\\Program Files\\nodejs\\etc\\npmrc".
verbose 0.518 Checking for configuration file "C:\\Users\\rfigueiredo\\Projects\\temp\\offline-new-ng\\.npmrc".
verbose 0.519 Checking for configuration file "C:\\Users\\rfigueiredo\\Projects\\temp\\.npmrc".
[...]

PR #7330, opened 2019-06-09
fix(config): Adds support for --offline flag to global add command

fcrick:
**Summary**

Adds support for --offline flag to global add command. 

Previously, this argument was ignored, resulting in issue #4913.

**Test plan**

Observe below how yarn currently goes to network when --offline is set with global add:
```
$ yarn --offline global add packagethatdoesnotexist@1.0.0
yarn global v1.16.0
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/packagethatdoesnotexist: Not found".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\fcrick\\AppData\\Local\\Yarn\\Data\\global\\yarn-error.log".
```
With fix, it now suggests running without `--offline`
```
$ node ./bin/yarn.js --offline global add packagethatdoesnotexist@1.0.0
yarn global v1.17.0-0
[1/4] Resolving packages...
error Couldn't find any versions for "packagethatdoesnotexist" that matches "1.0.0" in our cache (possible versions are ""). This is usually caused by a missing entry in the lockfile, running Yarn without the --offline flag may help fix this issue.
```

Added test to verify this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 - Adds support for --offline flag to global add command
 
+  [#7330](https://github.com/yarnpkg/yarn/pull/7330) - [**Francis Crick**](https://guthub.com/fcrick)
+
 - Yarn will tolerate Yaml at parse time. Full support isn't ready yet and will only come at the next major.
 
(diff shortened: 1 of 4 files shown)


arcanis:
Thanks 👍

PR #7343, opened 2019-06-18
Support passing auth tokens for aliased packages

nevir:
Today, `isScopedPackage` fails for [aliased packages](https://yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias) (e.g. `npm:@yarn/core`). With current versions of yarn, installing a private scoped package results in 404s from the registry (unless you turn on `always-auth = true` via .npmrc)

This fixes the regex to handle that case.

### DIFF
--- a/__tests__/registries/npm-registry.js
+++ b/__tests__/registries/npm-registry.js
@@ -206,8 +206,18 @@ describe('request', () => {
           expect: {root: 'https://registry.npmjs.org', auth: false},
         },
         {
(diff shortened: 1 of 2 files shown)


gamedevsam:
There's plenty of people sill using yarn@1.x - this should be merged imo. Just ran into this issue and spent half a day trying to figure out what was causing it.

nevir:
Killing off some of my really old PRs

PR #7353, opened 2019-06-20
Optimize slice last sharp tail

Connormiha:
**Summary**
I just optimized function for cutting last `#` and tail. In v8(Chrome) it works 10x faster https://jsperf.com/yarn-split-sharp

**Test plan**


arcanis:
👍

PR #7357, opened 2019-06-21
Update en.js

shkfnly:
updated en.js to remove references to deprecated function `install`

**Summary**

This is not a substantial feature request, it simply removes references to yarn install from the responses en.js file.

The motivation for this change is to remove instructions for users that are no longer viable.


arcanis:
Using `yarn install` is perfectly fine - `yarn` is just an alias to `yarn install`, the later isn't deprecated.

PR #7372, opened 2019-07-04
feat(cli): warning on relative paths provided by --cwd param

mollyIV:
**Summary**

When a user executes `yarn add --cwd relative-path-to-the-dir dependency-name` command and passes a relative path to the directory, [the path is resolved](https://nodejs.org/api/path.html#path_path_resolve_paths) and a [current working directory](https://nodejs.org/api/process.html#process_process_cwd) is used instead. 

https://github.com/yarnpkg/yarn/blob/49332faf9d22f5c4ae33d4d3343895e189b3fac3/src/config.js#L458

A user might be surprised by described behaviour, so adding a warning can be really helpful. You can find more details about it in [this issue](https://github.com/yarnpkg/yarn/issues/5206).

**Test plan**

I did perform manual testing for a relative path:

Screen Shot 2019-07-04 at 10 14 20

The new warning is added ☝️🎊 

Also tested the potential regression, if we passed an absolute path to the command:

Screen Shot 2019-07-04 at 10 16 09

Worked as expected 👍 

I wish I could add unit tests to check the warning. Any suggestion on that would be appreciated 🙏 Was thinking about adding it to `config.js` tests - checking the warning while `config` initialisation for a relative path.

closes https://github.com/yarnpkg/yarn/issues/5206

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -454,6 +454,12 @@ export default class Config {
     this.registries = map();
     this.cache = map();
 
+    if (opts.cwd !== undefined) {
+      if (!path.isAbsolute(opts.cwd)) {
+        this.reporter.warn(this.reporter.lang('yarnAddRelativePathDetected'));
+      }
+    }
+
     // Ensure the cwd is always an absolute path.
     this.cwd = path.resolve(opts.cwd || this.cwd || process.cwd());
 
--- a/src/reporters/lang/en.js
+++ b/src/reporters/lang/en.js
@@ -135,6 +135,9 @@ const messages = {
   yarnOutdatedInstaller: 'To upgrade, download the latest installer at $0.',
   yarnOutdatedCommand: 'To upgrade, run the following command:',
 
+  yarnAddRelativePathDetected:
+    'Detected a relative path. --cwd only supports an absolute path. A current working directory will be used instead.',
+
   tooManyArguments: 'Too many arguments, maximum of $0.',
   tooFewArguments: 'Not enough arguments, expected at least $0.',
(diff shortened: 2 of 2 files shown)


src/config.js
```diff
@@ -454,6 +454,12 @@ export default class Config {
     this.registries = map();
     this.cache = map();
 
+    if (opts.cwd !== undefined) {
+      if (!path.isAbsolute(opts.cwd)) {
+        this.reporter.warn(this.reporter.lang('yarnAddRelativePathDetected'));
```
mollyIV: Should add display this warning only in a `verbose` mode? 🤔

src/reporters/lang/en.js
```diff
@@ -135,6 +135,9 @@ const messages = {
   yarnOutdatedInstaller: 'To upgrade, download the latest installer at $0.',
   yarnOutdatedCommand: 'To upgrade, run the following command:',
 
+  yarnAddRelativePathDetected:
+    'Detected a relative path. --cwd only supports an absolute path. A current working directory will be used instead.',
```
mollyIV: Opened to better message suggestions 🙌

src/config.js
```diff
@@ -454,6 +454,12 @@ export default class Config {
     this.registries = map();
     this.cache = map();
 
+    if (opts.cwd !== undefined) {
+      if (!path.isAbsolute(String(opts.cwd))) {
+        this.reporter.warn(this.reporter.lang('yarnAddRelativePathDetected'));
```
mollyIV: Maybe the `add` command should display this warning only in a `verbose` mode? 🤔

PR #7375, opened 2019-07-07
fix absolute path in @zkochan/cmd-shim under win

oneplusdash:
**Summary**

Relative path cannot be used under different partitions under windows,otherwise it may cause binary link error. If set prefix "C:\Program Files (x86)\Yarn", global folder "D:\Repository\.yarn" . In @zkochan/cmd-shim, it will cause
C:\Program Files (x86)\Yarn\bin\nodemon.cmd :
@"%~dp0\D:\Repository.yarn\node_modules.bin\nodemon.cmd" %*

**Test plan**

C:\Program Files (x86)\Yarn\bin\nodemon.cmd :
@"D:\Repository.yarn\node_modules.bin\nodemon.cmd" %*

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "preferGlobal": true,
   "description": "📦🐈 Fast, reliable, and secure dependency management.",
   "dependencies": {
-    "@zkochan/cmd-shim": "^3.1.0",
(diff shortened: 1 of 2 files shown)


Artoria2e5:
Please run `yarn install` on the tree and commit the changed lockfile.

Issue #6062, opened 2018-07-04
yarn crashing on package.json directory outside of project's root 


Andarist:
Previous, similar (but closed now) issue - https://github.com/yarnpkg/yarn/issues/4336

**What is the current behavior?**

`yarn` crashes if there is `package.json` **directory** in some parent directory of the project

**If the current behavior is a bug, please provide the steps to reproduce.**

Just create a `package.json` directory in the project's parent directory and run `yarn` inside the project directory to receive:
> error An unexpected error occurred: "EISDIR: illegal operation on a directory, read".

**What is the expected behavior?**

Either `yarn` should handle this edge case or throw a helpful error.

**Please mention your node.js, yarn and operating system version.**

yarn@1.7.0 - rest is irrelevant

tnrich:
Also hitting this issue and just want to add that the error log given does not help in any way to track down what exactly is going on. If an error is going to be thrown, then at the very least there should be details pointing to the package.json dir that is causing the issue. 

As of right now all we get is this cryptic error log: 
```
Arguments: 
  /Users/tnrich/.nvm/versions/node/v10.8.0/bin/node /Users/tnrich/.nvm/versions/node/v10.8.0/bin/yarn

PATH: 
  /Users/tnrich/.nvm/versions/node/v10.8.0/bin:/Users/tnrich/bin/Sencha/Cmd/4.0.4.84:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/local/sbin

Yarn version: 
  1.10.1

Node version: 
  10.8.0

Platform: 
  darwin x64

Trace: 
  Error: EISDIR: illegal operation on a directory, read

A long list of my yarn.lock file comes next..
```

beck:
Doesn't even have to be a parent directory, also trashes sibling projects (discovered when I managed to break some CI).

Reproduction steps (on OSX):
```
~ $ mkdir -p projects/test projects/package.json
~ $ cd projects/test/
~/projects/test $ yarn
yarn install v1.16.0
Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (fs.js:498:3)
    at tryReadSync (fs.js:332:20)
    at Object.readFileSync (fs.js:369:19)
    at onUnexpectedError (/usr/local/Cellar/yarn/1.16.0/libexec/lib/cli.js:91826:106)
    at /usr/local/Cellar/yarn/1.16.0/libexec/lib/cli.js:91936:9
```

[...]

PR #7389, opened 2019-07-10
Fix 'illegal operation on a directory'

beck:
**As is**

If a parent directory contains a folder named `package.json/`,
yarn will try to json parse the directory.

**With this**

Easier to ask for forgiveness than permission because testing for
existence is not sufficient.

Fixes: https://github.com/yarnpkg/yarn/issues/6062


beck:
Verified issue is still in master and then rebased to resolve whitespace conflict.

PR #7398, opened 2019-07-13
Run the prepare script of git dependencies even if NODE_ENV=production

jfirebaugh:
**Summary**

When installing a git dependency with the `NODE_ENV` environment variable set to `production`, yarn did not run the `prepare` lifecycle script, because `config.production` would default to true, and then `wrapLifecycle` would skip `prepare`:

https://github.com/yarnpkg/yarn/blob/b6569538de69e0ccd201f0a33f1f5b52f2656f5b/src/cli/commands/install.js#L1202-L1207

This is contrary to npm's behavior and the intent of `GitFetcher#fetchFromInstallAndPack`, which is to install the dependency with its devDependencies and run all the relevant lifecycle scripts (excluding `prepublish`), regardless of the configuration or environment settings of the parent install command.

**Test plan**

Added an automated test.

cc @Volune, who added the original feature in #3553. Thanks for the test case which I cribbed from!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
-- Run the prepare script of git dependencies even if NODE_ENV=production
+- Runs the `prepare` lifecycle of git dependencies even if `NODE_ENV` is set to `production`.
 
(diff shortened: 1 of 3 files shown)


arcanis:
Looks good to me! Can you update the CHANGELOG file?

arcanis:
Thanks!

Issue #7436, opened 2019-07-26
Run `yarn test` would be failed with error: SSL_CTX_use_certificate:ca md too weak


rockwyc992:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
```
 FAIL  __tests__/util/request-manager.js
  ✕ RequestManager.request with cafile (104ms)
  ✕ RequestManager.request with ca (string) (3ms)
  ✕ RequestManager.request with ca (array) (2ms)
  ✕ RequestManager.request with mutual TLS (2ms)

  ● RequestManager.request with cafile

    error:140AB18E:SSL routines:SSL_CTX_use_certificate:ca md too weak

      20 |     cert: await fs.readFile(path.join(__dirname, '..', 'fixtures', 'certificates', 'server-cert.pem')),
      21 |   };
    > 22 |   const server = https.createServer(options, (req, res) => {
      23 |     res.end('ok');
      24 |   });
      25 |   try {

      at Object.<anonymous> (__tests__/util/request-manager.js:22:24)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13

Test Suites: 1 failed, 1 total
Tests:       4 failed, 11 passed, 15 total
Snapshots:   0 total
Time:        2.138s
Ran all test suites matching /__tests__\/util\/request-manager.js/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
- use Debian 10.0(buster), which [release](https://lists.debian.org/debian-announce/2019/msg00003.html) on 2019/07/06, or [sid](https://www.debian.org/releases/sid/) version.
- execute jest unit test, like follow command `yarn test-only __tests__/util/request-manager.js`

**What is the expected behavior?**
It will pass all test case.

**Please mention your node.js, yarn and operating system version.**
```
$ node -v
v12.6.0
$ yarn -v
1.18.0-0
$ uname -a
Linux 001713996ca0 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5 (2019-06-19) x86_64 GNU/Linux
$ openssl version
OpenSSL 1.1.1c  28 May 2019
```

rockwyc992:
[...]

PR #7456, opened 2019-08-03
Prevent mixing iojs and node engine detection

### LINKED ISSUE #7451: Incompatible module error due to bad aliasing between node and iojs
Report a *bug* (_This issue is a variant of https://github.com/yarnpkg/yarn/issues/623_).

When installing npm packages that are both referencing `engines.iojs` and `engines.node` properties, like [koa-ssl](https://github.com/jclem/koa-ssl/blob/7275edf480f500056cfc543f020bbd11d16e0fd4/package.json#L17-L20), `yarn` will always choose the version defined for `iojs` since it considers it as an ["alias"](https://github.com/yarnpkg/yarn/blob/697a254ee794085c77f82f14f4a515d05e81e294/src/package-compatibility.js#L46-L48) which is **wrong** since both engines use different semver numbers:

- [iojs](https://iojs.org/dist/)
- [nodejs](https://nodejs.org/dist/)

Moreover, since "iojs" comes before "node" (in alphabetical order since V8 uses this order when iterating on object props), it will always try to match "iojs" value with the one given by `node`.

```bash
$ yarn add koa-ssl
yarn add v1.17.3
info No lockfile found.
[2/4] 🚚  Fetching packages...
error koa-ssl@2.0.1: The engine "node" is incompatible with this module. Expected version "1.2.0". Got "10.16.0"
error Found incompatible module.
```

`yarn` should treat the version differently depending on the engine running JS.

```bash
$ node -v
v10.16.0
$ yarn -v
1.17.3
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 10.14.6 (18G84)
      Kernel Version: Darwin 18.7.0
      ...
```

eilgin:
**Summary**

This PR fixes #7451. Currently, installing packages that require both `iojs` and `node` engine are broken if the "iojs" version doesn't match either the "iojs" **and** the "node" version whereas it should make a separate check between those engines.

**Test plan**

I've checked that there's no more aliasing between the iojs and node version since they do not share the same semver semantics.


eilgin:
There're some warnings related to "iojs" when it appears in the "engine" properties:
```
warning <pkg>: The engine "iojs" appears to be invalid.
```

A fix could be to add it in the ["ignore" list](https://github.com/yarnpkg/yarn/blob/ee29a5fd6a9763b4befa3d3f28671e40f6579b2f/src/package-compatibility.js#L50).

eilgin:
It appears that some tests aren't working properly since the TLS certificate used is expired, see https://github.com/yarnpkg/yarn/issues/7436#issuecomment-515300669

PR #7472, opened 2019-08-12
Do not log `err.message` of `ProcessTermError`s if silent

### LINKED ISSUE #5359: `yarn run --silent` doesn't suppress yarn's stderr

bug

`yarn run --silent` doesn't suppress stderr

for example, imagine this script section of a package.json file
```json
"scripts": {
    "git-is-clean": "git diff-index --exit-code --shortstat --line-prefix 'UNCOMMITED CHANGES!' HEAD --",
    "deploy": "yarn run --silent git-is-clean && firebase deploy"
  },
```

If the working directory is not clean, the output of `yarn run deploy` would be 
```
yarn run v1.3.2
$ yarn run --silent git-is-clean && firebase deploy
UNCOMMITED CHANGES! 2 files changed, 150 insertions(+), 7 deletions(-)
```

Note the duplicated `error Command failed with exit code 1.`
Every level of nesting that fails will add one of these lines.
If the output showed which command failed, I could sorta see this being a feature and not a bug.   
For example
```
...
error Command "git-is-clean" failed with exit code 1.
error Command "deploy-staging" failed with exit code 1.
...
```
But as is, it's just clutter.

If the working directory is not clean, the output of `yarn run deploy` should be 
```
yarn run v1.3.2
$ yarn run -s git-is-clean && firebase use dev && firebase deploy --message "$(git log -1 --oneline --decorate=short)" #--decorate-refs-exclude=HEAD
UNCOMMITED CHANGES! 2 files changed, 150 insertions(+), 7 deletions(-)
```

**Workarounds**
FWIW, npm --silent does squelch stderr as expected, so i can "solve" this problem by using `npm run --silent` instead of `yarn run --silent` inside my scripts, but it feels dirty to have yarn call npm.
Alternatively, in some cases, I could do `yarn run --silent foo 2>/dev/null`, but that would hide stderr from the `foo` script output so it isn't always an option.

yarn v1.3.2
node v8.9.4
os Solus 3 - linux 4.14.18

kaste:
Fixes #5359

If sub commands called via `yarn run` fail yarn typically logs `error Command failed with exit code 1.` even if `--silent` is requested basically because a failing command throws (or rethrows) itself instead of just returning this failure as valid output. 

This is the simplest patch to suppress this output.

### DIFF
--- a/__tests__/fixtures/index/run-failing-script/package.json
+++ b/__tests__/fixtures/index/run-failing-script/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "test_run_failing_script",
+  "version": "1.0.0",
+  "license": "UNLICENSED",
+  "scripts": {
+    "custom-script": "echo \"Hi\" && exit 1"
+  }
+}
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -49,9 +49,9 @@ async function execCommand(
         cwd: workingDir,
         env: cleanedEnv,
       },
-      (error, stdout) => {
+      (error, stdout, stderr) => {
         if (error) {
-          reject(Object.assign((new Error(error.message): any), {stdout}));
+          reject(Object.assign((new Error(error.message): any), {stdout, stderr}));
         } else {
           const stdoutLines = stdout
             .toString()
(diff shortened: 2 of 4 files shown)


kaste:
Of course the tests don't fail on my machine. 🙄

kaste:
Well, still looks like `YARN_SILENT` is actually set here. Inherited from the parent? (But you do have tests in the same file where you set it explicitly.) 

For now, I just unset `YARN_SILENT`. 

Still, everything red bc the RequestManager times out/cert invalid. Disappointing for a first time contrib 😞 🤷‍♀

arcanis:
> Still, everything red bc the RequestManager times out/cert invalid. Disappointing for a first time contrib 😞 🤷‍♀

Apologies - my attention is like at 99% focused [on the v2](https://github.com/yarnpkg/berry) at this point, so I get less time to fix (or even see) this kind of things 😕

Your contrib looks fine apart from the CA things that are on us, so I'll plan it for the 1.18. Thanks!

PR #7473, opened 2019-08-12
Add Parcel to the engines ignore list

DeMoorJasper:
**Summary**

Parcel requires plugins to use engines.parcel to specify the Parcel version the plugin supports.

Currently yarn is throwing warnings for this, this PR adds parcel to the ignore list so this no longer happens.

See: https://github.com/parcel-bundler/parcel/blob/master/PARCEL_2_RFC.md#engines

**Test plan**


arcanis:
Sounds reasonable 👍

PR #7494, opened 2019-08-25
Update URL

kant:
**Summary**

This is not a substantial feature

This PR simply updates an internal URL

**Test plan**

This PR do not break any code, so, no collateral damage expected.


merceyz:
Fixed in https://github.com/yarnpkg/yarn/commit/0a121f66ffcb4fc48ed6ea9e5ad928bd7cdc9b5f

PR #7512, opened 2019-09-01
Fixes link:.

arcanis:
**Summary**

Using `link:.` currently generates an infinite loop (because we load the dependencies from our own folder, then from our own folder, then from our own folder, then ...).

**Test plan**

Added such a pattern to the Yarn repo and checked that it works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 - Fixes using `link:.` to refer to the package folder
 
-  []() - [**Maël Nison**](https://twitter.com/arcanis)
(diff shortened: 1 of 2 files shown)


krzysztof-pawlik-gat:
@arcanis This PR overwrote part of `CHANGELOG.md` and caused the lint to fail (see two referenced PRs above).

arcanis:
@krzysztof-pawlik-gat I removed the entry from the changelog since it was an infra change that wasn't having any effect on the consumers. The lint issue wasn't intended, though.

krzysztof-pawlik-gat:
> @krzysztof-pawlik-gat I removed the entry from the changelog since it was an infra change that wasn't having any effect on the consumers. The lint issue wasn't intended, though.

Ok, thank you for the explanation, I noticed it only due to unusual merge conflict in the file.

PR #7514, opened 2019-09-03
fix(npm-resolver): don't list deprecated versions

pi0:
**Summary**

Deprecated NPM packages cannot be removed from the registry but still being listed when we suggest users installing an alternative version. The problematic case with us was [nuxt-edge](https://www.npmjs.com/package/nuxt-edge) package. An automated bot wrongly published `v3` and we were not aware of it. Now when someone tries to install an invalid version, yarn suggests installing deprecated 3.x ones.

**Test plan**

```bash
yarn add nuxt-edge@2.100`
```
Now:

image

After change: 

image


Artoria2e5:
Run `eslint --fix` for the prettier issue. It probably isn't your fault. The test errors don't look like your fault either.

PR #7520, opened 2019-09-04
Resolve .bin path based on --modules-folder

### LINKED ISSUE #7519: --modules-path does not support absolute path when running bin or lifestyle scripts
Bug

When I supply an absolute path to the `--modules-folder` flag (eg. `--modules-folder /tmp/node_modules`) I get the wrong directorry back from `yarn bin` and the wrong directory in my `$PATH` variable for lifecycle scripts.

Here's a repository that will help to reproduce this issue: https://github.com/ryanwilsonperkin/yarn-modules-folder-issue-repro

> ℹ️ The repository specifies the `--modules-folder` path so you don't have to. It also specifies two lifecycle scripts that will be helpful in understanding this problem.

First, Checkout the above repository and `cd` into it. For the purpose of the examples below, I will assume the directory you've checked these out to is `/Users/ryanwilsonperkin/yarn-modules-folder-issue-repro`. Run `yarn` in this directory to install the two dependencies. They will be installed at `/tmp/node_modules` per the configuration in `.yarnrc`

Run `yarn bin` in this directory. You should expect to see:

```sh
$ yarn bin
/tmp/node_modules
```

but will instead see:

```sh
$ yarn bin
/Users/ryanwilsonperkin/yarn-modules-folder-issue-repro/tmp/node_modules
```

Run `yarn echo-path` in this directory. You should expect to see `/tmp/node_modules` show up in the `$PATH` variable that is echoed, but will instead see `/Users/ryanwilsonperkin/yarn-modules-folder-issue-repro/tmp/node_modules`.

Run `yarn run-babel` in this directory. You should expect to see:

```sh
$ yarn run-babel
babel:
  stdin compilation requires either -f/--filename [filename] or --no-babelrc
```

but will instead see:

```sh
$ yarn run-babel
/bin/sh: babel: command not found
```

1. I expect `yarn --modules-folder /tmp/node_modules bin` to resolve `/tmp/node_modules/.bin`
1. I expect `yarn --modules-folder /tmp/node_modules run-babel` to be able to correctly find the `/tmp/node_modules/.bin/babel` executable.

OS: MacOS 10.14.6
yarn: 1.17.3
node: 10.11.0

ryanwilsonperkin:
**Summary**

Fixes #7519 

The first entry in registryFolders array corresponds to the location provided with `--modules-folder` to specify where to store node_modules. This currently does a `path.join` on the current location and the specified directory which prevents the modules folder from existing outside of the current tree.

For example, if we run:

```sh
yarn --modules-folder install
```

we will successfully get the modules installed into `/tmp/node_modules`, but when we later run

```sh
yarn --modules-folder /tmp/node_modules bin
```

we will get `/root/workspace/tmp/node_modules` instead of `/tmp/node_modules` as we expect.

This also happens when injecting the .bin directory into the PATH variable when executing lifecycle scripts.

**Test plan**

Comparison of before/after of `yarn bin` with absolute `--modules-folder`:
image

Comparison of before/after of `yarn which-jest` with absolute `--modules-folder`:
image
(after the addition of a lifecycle script called `which-jest` that just runs `which jest`.

### DIFF
--- a/src/cli/commands/bin.js
+++ b/src/cli/commands/bin.js
@@ -15,7 +15,7 @@ export function setFlags(commander: Object) {
 }
 
 export async function run(config: Config, reporter: Reporter, flags: Object, args: Array<string>): Promise<void> {
-  const binFolder = path.join(config.cwd, config.registryFolders[0], '.bin');
+  const binFolder = path.resolve(config.cwd, config.registryFolders[0], '.bin');
   if (args.length === 0) {
     reporter.log(binFolder, {force: true});
   } else {
--- a/src/util/execute-lifecycle-script.js
+++ b/src/util/execute-lifecycle-script.js
@@ -192,10 +192,10 @@ export async function makeEnv(
   for (const registryFolder of config.registryFolders) {
     const binFolder = path.join(registryFolder, '.bin');
     if (config.workspacesEnabled && config.workspaceRootFolder) {
-      pathParts.unshift(path.join(config.workspaceRootFolder, binFolder));
+      pathParts.unshift(path.resolve(config.workspaceRootFolder, binFolder));
     }
-    pathParts.unshift(path.join(config.linkFolder, binFolder));
-    pathParts.unshift(path.join(cwd, binFolder));
+    pathParts.unshift(path.resolve(config.linkFolder, binFolder));
+    pathParts.unshift(path.resolve(cwd, binFolder));
   }
 
   let pnpFile;


ryanwilsonperkin:
Would someone help me understand the lint and appveyor checks that are failing? I haven't touched the files that the `lint` step is commenting on.

Artoria2e5:
yarn folks are not running the `lint` task on the master workflow, so it's not a huge surprise when it contains something badly formatted. See https://circleci.com/workflow-run/d4095a06-b009-4588-bd37-c27d7c438e16 for an example.

mikegee:
@ryanwilsonperkin i think i need this. did you find a workaround?

ryanwilsonperkin:
I switched jobs! 😅

PR #7528, opened 2019-09-06
yarn shim: tweak winpty-firing logic

Artoria2e5:
**Summary**
YARN_FORCE_WINPTY is pointless when we simply don't have winpty. So let's not do that.

The terminal check should apply both to the input and the output. The output (stdout, fd 1) one is what we already have here; it checks for yarn being piped somewhere else, which winpty can break. But what interactive prompts really want is a functional input (that is, `conhost` but not a `tty` on windows), so we should check for the stdin (fd 0) too.

**Test plan**
The shell logical operator is a hard-to-tame beast. There are, in fact, no rules for operator precedence: the shell just scans command-by-command. When something is good it goes for whatever is after the first `&&`, otherwise it goes for the first `||` it sees next.

We can mock the two checks in a simplified environment:

```Bash
pretty_bool=(true false)
dramatic_test() { echo "$1=${pretty_bool[$2]}"; return "$2"; }
force()  { dramatic_test force "$1"; }
exists() { dramatic_test exists "$1"; }
is_tty() { dramatic_test is_tty "$1"; }
result() { printf '%s\n\n' "${pretty_bool[$?]^^}"; }

# old behavior: force only overrides exists, but not the tty check
force 0 || exists 1 && is_tty 1; result
force 0 || exists 1 && is_tty 0; result
force 1 || exists 0 && is_tty 1; result

# new behavior: force now overrides is_tty, but not exists
exists 1 || { force 0 && is_tty 1; }; result
exists 1 || { force 1 && is_tty 0; }; result
exists 0 || { force 0 && is_tty 1; }; result
```
(Now as an afterthought, it's actually not that expensive to do the is_tty check first... Wait for a force push then.)


Artoria2e5:
I didn't even change a `js` file and it says I have bad js coding habits

PR #7571, opened 2019-09-25
Add quotes arround path segments that includes spaces on Windows

Bram-Zijp:
**Summary**
When one runs the `yarn create ...` command on Windows and his working directory includes a space, the command will throw an error.

**Test plan**

Without these changes:
image

With these changes: 
image


Bram-Zijp:
Closing, probably outdated.

Issue #7584, opened 2019-09-30
"Incorrect integrity when fetching from the cache"


xclidongbo:
yarn version: 1.19.0
**Do you want to request a *feature* or report a *bug*?**

bug:
error Incorrect integrity when fetching from the cache

**What is the current behavior?**
When I install thie lib, the error is show:
for example: `yarn add react-native-zip-archive`
```
yarn add v1.19.0
info No lockfile found.
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] 🔍  Resolving packages...
warning react-native > create-react-class > fbjs > core-js@1.2.7: core-js@<2.6.8 is no longer maintained. Please, upgrade to core-js@3 or at least to actual version of core-js@2.
info There appears to be trouble with your network connection. Retrying...
warning jest > jest-cli > jest-config > jest-environment-jsdom > jsdom > left-pad@1.3.0: use String.prototype.padStart()
[2/4] 🚚  Fetching packages...
error Incorrect integrity when fetching from the cache
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
error Incorrect integrity when fetching from the cache

```
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
success install.
**Please mention your node.js, yarn and operating system version.**
node: v12.10.0
yarn: 1.19.0
macos: 10.14.6

chuan6:
Strengthening of `integrity` check introduced in #7582 seems to cause the issue. As a temporary fix, I commented out [the code](https://github.com/yarnpkg/yarn/pull/7582/files#diff-7246848d5d10d9c48ef4be28cb961582R23-R29) doing `integrity` check.

My guess is that for some packages, while local `yarn.lock` file may contain `integrity`, the remote registry doesn't. In my case, it's `underscore@1.7.0`.

Hope that @arcanis can help share some advice on resolving the issue @~@

chuan6:
@xclidongbo Please re-title the issue to reflect the error being thrown, so that people encountering this issue can all arrive here. Thanks :)

arcanis:
Some context: old lockfile entries (more than a year ago?) don't have the integrity field present. Back when it was added, we decided not to automatically migrate them, as it was generating a lot of changes on unsuspecting projects.

[...]

PR #7595, opened 2019-10-03
Fix cache integrity check false-positives across multiple registries

Blasz:
**Summary**

Improve cache integrity error messages and fixes the cache integrity check issue described here: https://github.com/yarnpkg/yarn/issues/7584#issuecomment-537354545

> I determined it is related to an internal npm registry that we use.
> 
> The registry that we use does not return an `integrity` field and so the cached versions of packages resolved from the registry uses `sha1` hashes. When the same package is then resolved from npm instead of our own registry, it has a sha512 integrity field and causes a clash.

I'm still not so sure about erroring out in these scenarios rather than just invalidating the cache for the problem packages since users will manually need to do that after the error was thrown. 

Also now that the `cacheIntegrity` field from cache is being used, I wonder whether the separate `-integrity` and no-integrity suffixed caches are required anymore?

**Test plan**

Added tests for the previous cache integrity changes. I saw test cases for sha384 integrity hashes so I wonder if those need to be accounted for, as the same false positive error will occur there as well. If npm ever decides to change their default integrity hash to something other than sha512 this check would also fail.

A better way of handling this could be to compute the remote integrity hash against the cache if it did not match the hash algorithms already stored in cache and then append that value to the cached integrity hash if it matched and throw an error otherwise.


Blasz:
cc @arcanis

arcanis:
Hey @Blasz! Thanks for working on it! It looks good to me, especially with the tests you've added ⭐

> Also now that the `cacheIntegrity` field from cache is being used, I wonder whether the separate `-integrity` and no-integrity suffixed caches are required anymore?

Good question 🤔 Probably not if we always compute both the sha1 and the sha512 regardless of the registry, but I'm not 100% sure there isn't some edge case somewhere in the logic so I prefer to keep it for now.

> A better way of handling this could be to compute the remote integrity hash against the cache if it did not match the hash algorithms already stored in cache and then append that value to the cached integrity hash if it matched and throw an error otherwise.

Yeah, but it would make the code even quite spaghetti 🤔

Daniel15:
This is probably an edge case, but the build server that publishes the Chocolatey package is hitting this error when trying to check out the Yarn Git repo now:

```
stderr: error: unable to create file __tests__/fixtures/install/install-update-auth-invalid-cache-integrity/.yarn-cache/v6/npm-safe-buffer-5.1.1-893312af69b2123def71f57889001671eeb2c853-integrity/node_modules/safe-buffer/.yarn-metadata.json: Filename too long
```

It's currently trying to check out the code to `C:\Program Files (x86)\Jenkins\workspace\yarn-chocolatey`. I can probably move the directory around to avoid this, but I wonder if other people working on Yarn could hit the same issue.

Issue #7074, opened 2019-02-28
"yarn add --module-folder . " deleted ALL of my code (including .git directory)


benvan:
**Do you want to request a *feature* or report a *bug*?**
A bug. A very serious bug. Like, "I'm in serious trouble" serious.

**What is the current behavior?**
You make a directory in your project folder called`lib`
You run `cd lib; yarn add some-module --modules-folder .` because you want to install a specific dependency outside of node_modules
You get an error telling you a file is missing ...
Because yarn deleted all of the files in your project folder (apart from lib and yarn-error.log)
You slowly back away from the computer. This is not your day.

**If the current behavior is a bug, please provide the steps to reproduce.**
See above. Although I wouldn't recommend it.

**What is the expected behavior?**
That yarn doesn't delete all of the files in my project.

**Please mention your node.js, yarn and operating system version.**
Yarn: 1.13.0
Node: v10.15.0
OS: macOS 10.14.3

arcanis:
I'm sorry for your troubles - unfortunately it's kinda expected. The `modules-folder` settings tell Yarn where are located the packages it should manage - for example by removing the packages that aren't used anymore. If you tell it that your module folder is the current directory, well, it kinda makes sense that it sees your source folders as unused packages.

In this case the only thing we could do would be to tell you that you probably made a mistake (a bit like `rm --no-preserve-root`), but given that `--modules-folder` is a very fringe option I'm not sure it's worth the use case.

Do you have a suggestion for a better documentation somewhere that would make this behavior less unexpected?

*(As a side note I personally find a bit demoralizing to see you post on HN before even trying to understand what happened. We all need karma in our life but come on ...)*

arcanis:
Giving a second look at your post it appears to me that you were running the command within a subdirectory `lib`, in which case the behavior might indeed be even more surprising.

When running `yarn` within a subdirectory of your project, `yarn` will automatically `cd` into the closest project it can find and run the command you ran there. It unfortunately doesn't seem to take into account that the command-line arguments should be resolved relative to your current cwd rather than the fixed one.

The fix would be to tweak our CLI to resolve some command line arguments relative to the current pwd rather than the fixed one. That would be here:

https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L501-L532

benvan:
Thanks for the swift response @arcanis 

[...]

PR #7607, opened 2019-10-07
Resolves folder options (i.e. --modules-folder) relative to cwd (#7074)

mbpreble:
* Prevents potentially surprising behavior and data loss

* Includes integration test verifying that user-reported data loss scenario does not occur

**Summary**
Motivated by high severity issue #7074, following recommended approach by @arcanis of resolving all folder options relative to the cwd (instead of potentially relative to the nearest project root). Adds an integration test illustrating there are no yarn failures, data loss, or other surprising behaviors in the scenario reported in the issue.

**Test plan**

It's relatively straightforward to reproduce the reported issue with a current version of yarn:
```
#! /bin/sh

# Create a project with a lib folder
mkdir -p my-project/my-lib

# My project is some sort of project
echo "{}" > my-project/package.json
echo "I definitely care about this file!" > my-project/IMPORTANT.TXT
cat my-project/IMPORTANT.txt

cd my-project/my-lib

# Specify (using a relative path) that I want to use my-lib as a modules folder, running inside of that folder
yarn add left-pad --modules-folder .

# Hey does my file exist?
cd ..
cat IMPORTANT.txt
```

Running this shows the yarn command failing, but IMPORTANT.txt having been deleted from the project folder. This does not occur if yarn includes this patch.

### DIFF
--- a/__tests__/integration.js
+++ b/__tests__/integration.js
@@ -560,3 +560,36 @@ test('yarn init -y', async () => {
   const manifestFile = await fs.readFile(path.join(cwd, 'package.json'));
   expect(manifestFile).toEqual(initialManifestFile);
 });
+
+test('--modules-folder option', async () => {
+  /**
+   * The behavior of --modules-folder (and other folder options) was that it resolved relative not to the current
+   * working directory, but instead to the closest project root (folder containing a package.json file).
+   *
+   * This behavior was at best surprising and could result in data loss. This test captures a scenario in which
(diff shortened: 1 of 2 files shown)


arcanis:
I'm a bit wary of this change since it's technically breaking, but given that the current behavior is quite unexpected and proved dangerous in the wild I'm gonna go ahead and merge it as a bugfix. Hopefully I won't regret it later 😅

Issue #1037, opened 2016-10-14
feature: Implement `yarn global config set|get`


FLGMwt:
**Do you want to request a _feature_ or report a _bug_?**
bug
**What is the current behavior?**
Currently, all `yarn config set` and `yarn config remove` commands modify home config, whether `-g` is set or not.
**If the current behavior is a bug, please provide the steps to reproduce.**
`yarn config set foo bar`
notice user level .yarnrc is modified
👍 

`yarn config set foo bar --global`
notice user level .yarnrc is modified and global .yarnrc is _unmodified_
👎 

**What is the expected behavior?**
--global should only modify global .yarnrc
**Please mention your node.js, yarn and operating system version.**

```
> node -v
v6.7.0
> npm -v
3.10.3
> yarn --version
0.15.1
```

BYK:
I've changed the issue since I don't think we are going to implement `-g` or `--global` flags. That would be inconsistent with how we do things. I think having a `yarn global config` command would be nice tho.

sla89:
You know, that the [yarn config documentation](https://yarnpkg.com/lang/en/docs/cli/config/) mentions the ```-g```and ```--global``` flag for setting configuration values globally but as mentioned here this feature does not work.

paul-soporan:
Closing as fixed in v2: v2's `yarn config set` works on the project configuration and `yarn config set --home` works on the home configuration.

PR #7619, opened 2019-10-11
feat(config): add support for local config

MeLlamoPablo:
**Summary**

This PR addresses #1037.

As of today, `yarn config set foo bar` will save that config key to your user directory (`~/.yarnrc`). `yarn config set foo bar -g` will do exactly the same. The former command is wrong: it should edit the "local configuration", or the configuration in the current workin directory (as pointed out by https://github.com/yarnpkg/yarn/issues/1037#issuecomment-362546925).

This PR moves the current behaviour under the `--global` flag, and makes it so invoking the command without the flag interacts with the file in the current working directory instead. If the file does not exist, it is created.

**Test plan**

I didn't write tests yet! I wanted to see the reaction to this PR first. Also, when I run the tests there is one suite failing, but not because any change I've made, so I'd like to get some assistance with that.

Thanks for your time!

### DIFF
--- a/src/cli/commands/config.js
+++ b/src/cli/commands/config.js
@@ -46,6 +46,10 @@ export function hasWrapper(flags: Object, args: Array<string>): boolean {
 
 export function setFlags(commander: Object) {
   commander.description('Manages the yarn configuration files.');
+  commander.option(
+    '-g, --global',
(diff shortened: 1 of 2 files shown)


HarrisonB:
Is there anything I can do to help move this PR along?

PR #7650, opened 2019-10-26
make running with Plug'n Play possible on node 13

sverweij:
**Summary**

On node 13 `Module._findPath` (which the `setup()` of `generate-pnp-map-api.tpl.js` 
redefines) also gets called with `null` for the `path` parameter (by [lib/internal/bootstrap/pre_execution.js](https://github.com/nodejs/node/blob/master/lib/internal/bootstrap/pre_execution.js#L447) in the (new) `resolveMainPath` function).

- This fix ensures the patched `_findPath` (which gets generated into .pnp.js) also
handles that situation.
- Adds node 13 (and 12) to the build matrices of the CI's that support it - so it's possible to see the change works (and doesn't break other stuff on node 12 & 13)
  - azure pipelines
  - circleci/ linux
  -  not on circleci/ macOS because there node is installed via homebrew - which doesn't have node 13 yet (maybe nvm can be used there instead? Holler if you want that - I can do that in a separate PR)
  - not on appveyor as it doesn't have a node 13 image yet either
- Updates the change log to document the fix
-  ==>  circle ci on macOS failed in the _Install Node 10_ step with a timeout. This seems a generic problem (also #7649 (a readme update) is plagued by this). PR #7651 fixes it. 
=> I've taken the liberty to rebase this PR based on #7651's branch.

fixes #7642 

**Test plan**

- [x] manual run against [demo repo](sverweij/yarn-on-node-13)
- [x] ~~additional unit and/ or integration test~~ run existing tests to prove everything works ok (e.g. `packages/pkg-tests/pkg-tests-specs/sources` - with this PR they don't fail anymore on node 13)
- [x] green ci

**Notes**

- running `yarn` on yarn's repository under node 13 (on macOS) fails because there's no binary for fsevents (which uses node-gyp) and compiling from source fails.
- Locally _already without this patch_ 4 unit tests were failing (2 in init.js, 2 in integration.js). I might be able to fix them but that will be in a separate PR. I'd expect these to show up in the CI's output, but they don't (?).

PR #7649, opened 2019-10-26
Use consistent title casing for headings and links

mrseanbaines:
**Summary**

Reading through the project's README, I noticed a couple of inconsistencies in casing for headings and links.

**Test plan**

N/A - No code changes

PR #7651, opened 2019-10-27
ci(circleci): prevent timeout on circle-ci macos node10 build

sverweij:
**Summary**

- sets the required xcode version to 11 -> this gets us a more up to date macOS version 
- keeps xcode 9 for node 6 as that isn't available in the homebrew associated with xcode 11 anymore
- removes the `brew update` from the macOS node 10 job
- ✖️no update of the changelog - as merging this PR won't modify yarn's behaviour

As a bonus the CI for node 10 on circleci/ macOS got 15 - 20 minutes faster.

**Why, though?**
The circle ci job for macOS on node 10 currently fails because it takes too long (both PR's #7649 and #7650 don't green because of that). Cause: the job uses a big chunk of time updating homebrew (up to ~20min). Which is necessary because the version of macOS (the one connected to xcode 9 => macOS 10.12) does not contain node 10 by default.

Using a more recent version of macOS does away with this necessity for node 8 and 10. Putting node 6 on this would've been nice, but it is not available in homebrew on the macOS images that have xcode 11 on them

**Test plan**

- [x] green ci

**Notes**

- I've rebased #7650 on the branch underlying this PR - so if that one might get merged first this PR can be closed unmerged.

- When running the xcode 9 image, the system warns it's outdated and unsupported, as is the version of macOS. If node 6 still needs to be tested on macOS via circleci it might be worth considering a different installation strategy (e.g. using nvm in stead of homebrew?).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ docker_defaults: &docker_defaults
 
 macos_defaults: &macos_defaults
   macos:
-    xcode: "9.0"
+    xcode: "11.0"
   working_directory: ~/project/yarn
 
 attach_workspace: &attach_workspace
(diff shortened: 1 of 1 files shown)


arcanis:
Sounds good, we don't care about Node 6 anymore 🙂

PR #7663, opened 2019-10-31
Retry truncated tarball responses

### LINKED ISSUE #7521: Yarn install for private module failing - unexpected end of file
Hi all, 

I'm currently trying to install a private module using yarn. However, I get the following error when I do:

yarn install
yarn install v1.17.3
[2/5] Resolving packages...
[3/5] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@private/ngffwd-node-processes/-/ngffwd-node-processes-1.0.123.tgz: unexpected end of file".
info If you think this is a bug, please open a bug report with the information provided in "/opt/atlassian/pipelines/agent/build/yarn-error.log".

The commands I'm running are the following: 

npm config set scripts-prepend-node-path auto
yarn install

We should be able to install the package with no issue. Previously we've been able to do, however, recently that has not been the case. We tried going back to previous versions with no seeming improvement. We don't want to use npm install as that will affect the yarn.lock, but we are getting notifications that the version is available from npm.

The yarn-error.log doesn't yield much either, so I don't want to paste it here. 

npm - v6.9.0
node - v10.16.3
yarn - 1.17.3

Thanks any help or suggestions would be appreciated!

rkistner:
**Summary**

Fix #7521 by retrying tarball requests. This error is causing roughly half our `yarn install`s on CircleCI to fail, which is a massive pain.

Currently, most GET network requests are retried automatically, but tar downloads have a custom "process" function, and is only retried on network errors - not on error responses.

This PR adds retries for the following failure conditions:
1. A server error: 408, >= 500 (same as other requests).
2. Content-Length header not matching the body length (the issue from #7521).

The internal API around the `params.process` function is refactored a bit, but this only affects the tarball fetcher.

**Test plan**

1. You need a large NPM package in a private org on the npmjs.org registry. 10MB download size is ideal.
2. Get the package URL from an existing yarn.lock, e.g. `https://registry.npmjs.org/@myscope/mypackage/-/mypackage-1.0.0.tgz`.
3. Run a couple of times: `rm -rf ./cache && yarn add --cache-folder ./cache  https://registry.npmjs.org/@myscope/mypackage/-/mypackage-1.0.0.tgz` (or alternatively `@myscope/mypackage@1.0.0`).

Without this change, the typical output is:

```
yarn add v1.19.1
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.npmjs.org/@myscope/mypackage/-/mypackage-1.0.0.tgz: unexpected end of file".
info If you think this is a bug, please open a bug report with the information provided in "/path/to/yarn-error.log".
```

With this change, the expected output is:

```
yarn add v1.20.0-0
[1/4] Resolving packages...
info unexpected end of file
[2/4] Fetching packages...
[3/4] Linking dependencies...

success Saved lockfile.
info Direct dependencies
└─ @journeyapps/cordova-fabric-plugin@1.0.8-journey1
info All dependencies
└─ @journeyapps/cordova-fabric-plugin@1.0.8-journey1
Done in 16.00s.
```

Note the single line displaying "info unexpected end of file": This is the error occurring and then being retried.

**Implementation comments**

This is not the cleanest structure for doing retries, but it solves the problem while re-using the existing retry queue, and without refactoring the entire request-manager layer or affecting other requests.

`networkManager.retryDelay` is now configurable instead of hardcoded to 3000, so that retry tests can run faster.

I added validation of the Content-Length header by adding a `size` parameter to the `hashValidateStream`. This avoids needing an additional stream to keep track of the size. This reports an error with code `EBADSIZE` on that stream, but the actual error returned to the client is still `unexpected end of file`.

Only if the tar is actually a valid file, and the integrity check passes, but the Content-Length header fails, will this error be thrown (via the existing checks on `this.validateError`).


rkistner:
@arcanis Would be great if you could review this change, since you've worked with the retry logic before.

src/fetchers/tarball-fetcher.js
```diff
@@ -254,14 +258,25 @@ export default class TarballFetcher extends BaseFetcher {
             ...headers,
           },
           buffer: true,
-          process: (req, resolve, reject) => {
+          process: (req, res, resolve, reject, queueForRetry) => {
```
BYK: I'd keep the new parameter before the `resolve, reject` parameters to keep a more promise-like interface in the future.

```suggestion
          process: (req, res, queueForRetry, resolve, reject) => {
```

src/fetchers/tarball-fetcher.js
```diff
@@ -254,14 +258,25 @@ export default class TarballFetcher extends BaseFetcher {
...
+            // Clear errors when retrying
+            this.validateError = null;
+            this.validateIntegrity = null;
+
+            const contentLength = res.headers['content-length'];
```
BYK:
```suggestion
            const contentLength = parseInt(res.headers['content-length'], 10) || undefined;
```

src/fetchers/tarball-fetcher.js
```diff
@@ -275,24 +290,43 @@ export default class TarballFetcher extends BaseFetcher {
...
+                await this.removeDownloadedFiles();
+                if (this.validateError && this.validateError.code == 'EBADSIZE') {
+                  // Content-Length did not match the body size. Retry the request.
+                  if (!queueForRetry(err.message)) {
+                    reject(err);
+                  }
+                  return;
+                }
+                reject(err);
+              } catch (err2) {
```
BYK: How about renaming this to something more meaningful:
```suggestion
              } catch (retryError) {
```

src/fetchers/tarball-fetcher.js
```diff
@@ -275,24 +290,43 @@ export default class TarballFetcher extends BaseFetcher {
...
             }
 
-            integrityValidateStream.pipe(extractorStream).on('error', reject);
+            integrityValidateStream.pipe(extractorStream).on('error', async err => {
+              try {
+                // Cleanup before retrying
+                await this.removeDownloadedFiles();
+                if (this.validateError && this.validateError.code == 'EBADSIZE') {
+                  // Content-Length did not match the body size. Retry the request.
+                  if (!queueForRetry(err.message)) {
```
BYK: This nested `if` blocks seems a bit untidy and confusing to me but I've also failed to come up with a clearer way. If anyone has suggestions, please share :)

src/fetchers/tarball-fetcher.js
```diff
@@ -275,24 +290,43 @@ export default class TarballFetcher extends BaseFetcher {
...
-        await fsUtil.unlink(tarballMirrorPath);
-      }
+  async removeDownloadedFiles(): Promise<void> {
+    const tarballMirrorPath = this.getTarballMirrorPath();
+    const tarballCachePath = this.getTarballCachePath();
 
-      if (tarballCachePath && (await fsUtil.exists(tarballCachePath))) {
-        await fsUtil.unlink(tarballCachePath);
-      }
+    if (tarballMirrorPath && (await fsUtil.exists(tarballMirrorPath))) {
```
BYK: I know this is existing code but we can probably use this as an opportunity to "fix" it: it is discouraged to use the async `exists` method to check for a file and try operating on it.

It'd probably be better to refactor this to something like:
```js
try {
return await Promise.all([
  tarballMirrorPath ? fsUtil.unlink(tarballMirrorPath) : Promise.resolve(true),
  tarballCachePath ? fsUtil.unlink(tarballCachePath) : Promise.resolve(true),
]);
} catch (err) {
  // Ignore if file does not exist
  if (err.code !== 'ENOENT') {
    raise err;
  }
}
```

src/util/request-manager.js
```diff
@@ -92,6 +98,7 @@ export default class RequestManager {
     this.cache = {};
     this.max = constants.NETWORK_CONCURRENCY;
     this.maxRetryAttempts = 5;
+    this.retryDelay = 3000;
```
BYK: This should probably be a constant.

src/util/request-manager.js
```diff
@@ -503,16 +512,28 @@ export default class RequestManager {
     const process = params.process;
     if (process) {
       req.on('response', res => {
+        if (res.statusCode === 408 || res.statusCode >= 500) {
+          // Retryable server responses
+          const description = `${res.statusCode} ${http.STATUS_CODES[res.statusCode]}`;
+          if (!queueForRetry(this.reporter.lang('internalServerErrorRetrying', description))) {
```
BYK: Based on the usage here and above, it feels to me that `queueForRetry` should either be an async function, raising the exception itself or take in a `reject` callback and do the rejection itself.

__tests__/fetchers.js
```diff
@@ -433,3 +433,78 @@ test('TarballFetcher.fetch throws on truncated tar header', async () => {
...
+  const workingUrl = 'https://registry.npmjs.org/lodash.isempty/-/lodash.isempty-4.4.0.tgz';
+  // This one returns a truncated response - body is shorter than the Content-Length header.
+  const brokenUrl = 'https://registry.npmjs.org/lodash.isempty/-/lodash.isempty-4.4.0-broken.tgz';
+
+  const mockRequest = (require: any).requireActual('./__mocks__/request');
+
+  let retryCount = 2;
+
+  // Shorter retry delay for tests
+  config.requestManager.retryDelay = 10;
```
BYK: Why not outright 0?

__tests__/fetchers.js
```diff
@@ -433,3 +433,78 @@ test('TarballFetcher.fetch throws on truncated tar header', async () => {
...
+  config.requestManager._requestModule = options => {
+    retryCount -= 1;
+    if (retryCount >= 0) {
```
BYK: Why not `(retryCount > 0)` here and `retryCount = 3` above and `retryCount <= 0` below? Looks clearer to me.

rkistner: Since it's two retries (3 requests total). But can improve the wording and logic a bit to make it clearer.

__tests__/fetchers.js
rkistner: Can make it 0 - shouldn't have a significant impact either way.

src/fetchers/tarball-fetcher.js
rkistner: This one is a little tricky, since it doesn't fully match Promise semantics. If it is retried, _neither_ resolve nor reject is called - it is only called after the next call to `process`. I'll see if there's a cleaner way to do this.

src/fetchers/tarball-fetcher.js
```diff
@@ -146,7 +147,10 @@ export default class TarballFetcher extends BaseFetcher {
       this.validateError = err;
     });
     integrityValidateStream.once('error', err => {
-      this.validateError = err;
+      if (this.validateError == null) {
+        // Don't overwrite error if one is already there
```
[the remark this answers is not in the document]
rkistner: A quick-fix for the logic here is to just use separate variables for errors from the two streams.

The more general error chaining issue is more tricky here, and it would help if someone with with more knowledge about streams can help out here. If for example `integrityValidateStream` emits an error, `untarStream` still tries to continue parsing, and emits it's own error. So for the specific case of a truncated response, we actually end up with three errors: Content-Length mismatch, hash mismatch and "unexpected end of file". The "unexpected end of file" is the last one in the chain, but is the one that actually ends up being passed on to the user.

I would have expected `stream.pipe` to pass on errors as well, but I may be misunderstanding how streams work in Node. So not sure what the cleanest way is to handle this.

rkistner:
Thanks @BYK, I'll update the PR with your comments.

rkistner:
I refactored quite a bit of tarball-fetcher now to use actual Promises and asyncawait instead of passing resolve / reject callbacks around. This makes the code quite a bit easier to follow in my opinion, but let me know if anything should be changed there.

src/fetchers/tarball-fetcher.js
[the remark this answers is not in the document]
imsnif: As far as I know, this is the way to collect errors with streams. For better readability, I might go for something like `stream1.on('error', handleError).pipe(stream2).on('error', handleError)`, but I think in this case that would involve quite a bit of shifting around...

olingern:
Hey @rkistner -- thanks for this PR. Let me know if you'd like to see this through. I'd love to help.

I just looking over this. I had a couple of questions.

- This sounds good for CI's but I'm curious about the ergonomics for end users, i.e. my git server returns a `5xx` on my local machine. I think having the delay as a flag like `--enable-retry-delay` would keep existing behavior consistent.

- If you can reduce the amount of changes on this PR, it would also help get this pushed through. 

- Lastly, I'm curious as to why the retry logic was reworked? It's been a while since I looked at the request manager, but all the "nuts and bolts" are there for retrying, if I remember correctly.  In my head, I was thinking that if `--enable-retry-delay` is set to `true`, then upon receiving a status code in a valid retry range (`res.statusCode === 408 || res.statusCode >= 500`) yarn would:
  - Check to see if the flag is enabled and if there are still a valid number of retries left
  - Pause (_if the above is met_) all execution for [x] seconds. This could be done by resolving a promise in `setTimeout`. Increment retries.
  - Report that it is blocking
  - Continue execution and retry download
  - It could even be nice to have a message like, "Received a 500 with retries enabled. Pausing execution for 3 seconds."

I could be missing something so please tell me if I'm oversimplifying this, @BYK @rkistner

rkistner:
Thanks for the response @olingern.

The current request logic (without this PR) does retries automatically for all "metadata"/API requests, but not at all for tar downloads. I believe the tar downloads should have the same retry logic by default - I don't see any reason why the retries should only apply to the metadata requests.

I tried implement the retries with as little as possible change, but it's quite hard to do that and keep the code clean with the current system.

If you prefer less changes, I'm happy to go back to just the first two commits: https://github.com/yarnpkg/yarn/pull/7663/files/fc7833f797bd117da1030aa246ac961a7ae0c239
The next commit attempts to simplify the request logic overall, but isn't needed for this change.

b-barthel:
@olingern , @rkistner - sorry for popping in on an old PR, but we've been getting some reports of failures while download tarballs that should be transient - in our case 503/504 codes where a top level retry (yarn install || yarn install) solves the issue, but requires user intervention. The changes in request-manager.js are essentially what I was looking at making, so I was wondering if there was something I can do to help get this PR merged?

rkistner:
I'm happy to do more work on this, since I have a lot of projects staying on Yarn 1.x for the foreseeable future, and this remains an issue for me. However, I'll need direction from a maintainer on what will be required to get this merged in.

johnterickson:
@BYK - I'd like to hear your thoughts on @olingern's questions above in https://github.com/yarnpkg/yarn/pull/7663#issuecomment-578788409

We've been debugging yarn instability with Azure Artifacts customers for awhile now, so we when we root-caused it to a lack of retries we were excited to see this PR.  @b-barthel is on my team and we'd love to get the balling rolling here and help out where we can.

__tests__/fetchers.js
```diff
@@ -433,3 +433,81 @@ test('TarballFetcher.fetch throws on truncated tar header', async () => {
...
+  config.requestManager.retryDelay = 0;
+  config.requestManager._requestModule = options => {
+    tryCount += 1;
+    if (tryCount <= 2) {
+      // Fail the first two requests
+      return mockRequest({
+        ...options,
+        url: brokenUrl,
+      });
+    } else if (tryCount == 3) {
```
olingern, line 459: We probably want strict equality here.

src/constants.js
```diff
@@ -39,6 +39,9 @@ export const NETWORK_CONCURRENCY = 8;
 // HTTP timeout used when downloading packages
 export const NETWORK_TIMEOUT = 30 * 1000; // in milliseconds
 
+// Delay before retrying HTTP requests after failure
+export const RETRY_DELAY = 3 * 1000; // in milliseconds
```
olingern, line 43: I know this is how `NEWWORK_TIMEOUT` is setup, but I think a flag that could provide an override would be nice here. Some CI's may require different retry intervals.

src/fetchers/tarball-fetcher.js
```diff
@@ -126,7 +115,14 @@ export default class TarballFetcher extends BaseFetcher {
       },
     });
 
-    const hashValidateStream = new ssri.integrityStream(hashInfo);
+    // stream ->
```
olingern, line 118: Not sure what this comment is attempting to illustrate.

rohithjidagam:
@olingern : I see that there are some build failures. Can you take this PR to completion?

olingern:
I thought about this while re-reviewing it and think it's probably not a good fit for the Yarn core because the use case is mainly for CI's. I believe that a simple shell script should fix most people's problems.

I've created a project [here](https://github.com/olingern/yarn-retry) to demonstrate what that looks like in practice. For the lazy, the shell script for retrying when Yarn fails:
```shell
 
# Credit: 
# https://unix.stackexchange.com/questions/82598/how-do-i-write-a-retry-logic-in-script-to-keep-retrying-to-run-it-upto-5-times/82610
n=0
until [ "$n" -ge 5 ]
do
   yarn && break
   n=$((n+1)) 
   sleep 5
done
```

olingern:
One more thing, @arcanis / @Daniel15 and I haven't worked out a release strategy for Yarn v1 yet. I think any changes should be major bug fixes and security patches. Otherwise, I would apt to have any new features opened as issues over at @yarnpkg/berry

Naninani:
> I thought about this while re-reviewing it and think it's probably not a good fit for the Yarn core because the use case is mainly for CI's. I believe that a simple shell script should fix most people's problems.
> 
> I've created a project [here](https://github.com/olingern/yarn-retry) to demonstrate what that looks like in practice. For the lazy, the shell script for retrying when Yarn fails:
> 
> ```shell
> # Credit: 
> # https://unix.stackexchange.com/questions/82598/how-do-i-write-a-retry-logic-in-script-to-keep-retrying-to-run-it-upto-5-times/82610
> n=0
> until [ "$n" -ge 5 ]
> do
>    yarn && break
>    n=$((n+1)) 
>    sleep 5
> done
> ```

This is not a viable option as it will rerun the entire yarn - this is exactly what we want to avoid as it will add so much time to the build.
Imagine if you have a 10+ min yarn install and have to rerun it just for a single package timeout.

johnterickson:
Hi @olingern Since `yarn` already has retries on metadata operations, my perspective is that this PR is about filling in an (inadvertent I assume) gap.  

Does `berry` retry tarball downloads?

olingern:
> This is not a viable option as it will rerun the entire yarn - this is exactly what we want to avoid as it will add so much time to the build.
Imagine if you have a 10+ min yarn install and have to rerun it just for a single package timeout.

@Naninani This is not correct. If Yarn received an invalid HTTP response or truncated tarball during the package resolution or fetching of tarballs, it will exit out at that point. For good measure, I used React-Native's `package.json` to simulate a failure, retry, and truncated response. Everything finishes in under a minute. Repo [updated](https://github.com/olingern/yarn-retry).

It would look like you guys didn't give this a try and replied without doing due diligence.

The option always exists for you to either build yarn from this branch on your CI or fork and merge this PR.

If this PR merges and it has unforseen bugs / issues, I'm not sure anyone would have the time to jump in and fix it. So, that's why I'm being cautious about merging it.

@johnterickson I'm not sure, I haven't been involved lately but [maybe](https://github.com/yarnpkg/berry/blob/1d9a6073bc6a287ba74380e198a0d91131ec3b86/packages/yarnpkg-core/sources/httpUtils.ts#L84).

olingern:
Also, there's a lot of unanswered [questions](https://github.com/yarnpkg/yarn/pull/7663#issuecomment-578788409) and review comments so I'm keen on making this as simple as possible for everyone and I think the shell script solution does that.

merceyz:
Closing as fixed in v2 where failed requests are automatically retried

PR #7499, opened 2019-08-28
Fixes offline mirror filename calculation for scoped packages URLs in Verdaccio (private npm repository)

xv2:
Fixes for `tarball` contains `%2f` instead of `/`
Sample: `http://npm.xxxyyyzzz.ru/@types%2fevents/-/events-3.0.0.tgz`
Issue: #7498

PR #7664, opened 2019-11-01
Handle GitHub Package Registry URLs

tvainika:
**Summary**

Fix tarball name resolver to handle GitHub Package Registry (currently still in beta) to match what I've experienced using private packages via beta program signup.

Because basename part of the urls do not match scoped tarball filenames, also offline mirror pruning needs to match the filename from url.

This is not very clean code, but for me it looks like handling offline mirror pruning correctly the codebase needs some refactoring to match tarball filenames in common way usable both for tarball fetcher and offline mirror pruning.

**Test plan**

Added tarball fetcher test similar to other private package registries have. For offline mirror pruning I did not find suitable template for adding new test.

### DIFF
--- a/__tests__/fetchers.js
+++ b/__tests__/fetchers.js
@@ -353,6 +353,28 @@ test('TarballFetcher.fetch properly stores tarball for scoped package resolved f
   expect(fetcher.getTarballMirrorPath()).toBe(path.join(offlineMirrorDir, '@exponent-configurator-1.0.2.tgz'));
 });
 
+test('TarballFetcher.fetch properly stores tarball for scoped package resolved from GitHub registry', async () => {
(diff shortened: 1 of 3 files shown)


arcanis:
Hey! Thanks for the diff! Can you describe in more details the bug and circumstances of the bug? Without knowing the context I'm not entirely sure what it fixes. Thanks 😃

tvainika:
> Can you describe in more details the bug and circumstances of the bug?

The change is similar to #7499 to fix offline mirror filename but with GitHub package registry.

Jonathan-Navagis:
Is there a workaround for this issue? Or any plan to merge this change in? I'm experiencing the same issue of yarn not pulling down the tarball into the offline mirror if it's in a private GitHub package repository.

PR #7722, opened 2019-12-01
Adds a note of the workspace name when running commands in all workspaces

orta:
**Summary**

When running a command in each workspace, echo out the name of the workspace. Here's a before / after.

image

My tests were failing locally, so I figured I'd validate it against CI :D


arcanis:
Thanks! 🙂

PR #7755, opened 2019-12-11
Fixes bin overwrites

arcanis:
**Summary**

In some circumstances Yarn would allow overwriting files on the filesystem with symlinks at install-time. Despite postinstall scripts having the ability to do that anyway this behavior isn't intended as Yarn installs are reputed to control side effects (the cache and temporary folders being the only exception). Bins shouldn't be able to escape the package because of Yarn.

**Test plan**

Adding tests

### DIFF
--- a/__tests__/__snapshots__/normalize-manifest.js.snap
+++ b/__tests__/__snapshots__/normalize-manifest.js.snap
@@ -60,6 +60,24 @@ Array [
 ]
 `;
 
+exports[`dangerous bin name: dangerous bin name 1`] = `
+Array [
+  "foo: Invalid bin entry for \\"/tmp/foo\\" (in \\"foo\\").",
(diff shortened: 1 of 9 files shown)


valerybugakov:
Hey @arcanis, we've bumped into a new warning about `invalid bin field`. Could you explain why using dot in a bin key is not suggested?

For example: 

```json
{
  "bin": {
    "build.cli": "./cli.js"
  }
}
```

d-fischer:
This breaks my build as well. Some of my bins have dots in their names, including a few that I have no control over.

TheRealChssAddct:
Hey @arcanis, this breaks some packages from building because bin key names contain colons.  Why so restrictive on the bin key name validation?

arcanis:
The name check is slightly too restrictive, but I don't have bandwidth to fix that atm. Please open a PR and link it here.

valerybugakov:
@arcanis done

Issue #7764, opened 2019-12-13
[Bug] Yarn PnP does not work inside folder with path containing a space


hlegendre:
**Describe the bug**

Yarn PnP does not work when inside a folder whose path contains spaces.

**To Reproduce**

Simply move a working project to a folder which path contains a space and try to run it.
A repo as been setup with a minimal case here : https://github.com/hlegendre/yarn_pnp_bug_with_space

I tried to move the cache to a folder (config set cache-folder or cacheFolder depending on version) without space to no avail.

**Environment if relevant (please complete the following information):**

 - OS: macOS 10.14.6
 - Node version 12.13.0
 - Yarn version 1.21.1 or 2.0.0-rc.19

**Additional context**

Same issue is being opened on yarnpkg/berry as [632](https://github.com/yarnpkg/berry/issues/632)

hlegendre:
A PR is open if anyone wants to merge it.

merceyz:
Fixed in v2

https://yarnpkg.com/getting-started/migration

PR #7775, opened 2019-12-19
bugfix #7764 pnp path could not contain space

hlegendre:
**Summary**

This PR solves the #7764 bug where PnP was failing when in a path containing a space.

**Test plan**

I have no idea at the time how to test this small fix (notably because it will fix only with node 12+ that supports quoting of NODE_OPTIONS)


merceyz:
Closing as this was fixed in v2

PR #7785, opened 2019-12-29
fix(workspaces): Passes arguments follwing "--" when running a workspace script

### LINKED ISSUE #7776: Yarn workspace does not propagate flags
When using yarn workspaces with run command, flags are not passes into executed command.

### How to reproduce: 
```bash
yarn workspace test_workspace run start -- \
  serverURL="hello" \
  awsAccessKey="world"
```

package.json file contents.
```
{
  "scripts":
  {
    "start": "node script.js""
  }
}
```

### Current behavior:
Flags are not passed into the running script.

### Expected behavior:
Flags should be passed.

### Environment
Node: 12.13.1
Yarn: 1.21.1
OS: macOS 10.14.6

rally25rs:
fixes #7776

**Summary**

Passes arguments following `--` when running a workspace script (for example `yarn workspace pkg run command -- arg`).

Previously these parameters were being trimmed off and ignored.

It seems that in `src/cli/index.js` we were trimming all params after and including the `--`, and assigning the remainder to `flags.originalArgs` which the `workspace` command was using.

Instead, this PR changes the `workspace` command to use the `args` parameter that was already being passed to the command, which includes the args after the `--`.
The `args` parameter to the command is also what the `yarn run` command uses, so it makes sense for `yarn run` and `yarn workspace run` to both use `args` instead of one using `args` and the other using `flags.originalArgs`.

**Test plan**

Since yarn parameter passing in `src/cli/index.js` is not tested, there are no additional tests.

A manual test can be performed by setting up a basic workspace project:

package.json
```
{
  "name": "yarn-7776",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "private": true,
  "workspaces": [
    "packages/*"
  ]
}
```

packages/pkg1/package.json
```
{
  "name": "pkg1",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "start": "./start.sh"
  }
}
```

packages/pkg1/start.sh
```
#!/bin/bash
echo "$@"
```

then run the yarn command
```
yarn workspace pkg1 run start -- one two three
```

The output from the script should be `one two three`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Passes arguments follwing `--` when running a workspace script (`yarn workspace pkg run command -- arg`)
+
+  [#7776](https://github.com/yarnpkg/yarn/pull/7776) - [**Jeff Valore**](https://twitter.com/rally25rs)
+
 - Prints workspace names with `yarn workspaces` (silence with `-s`)
 
   [#7722](https://github.com/yarnpkg/yarn/pull/7722) - [**Orta**](https://twitter.com/orta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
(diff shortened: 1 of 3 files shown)


rally25rs:
🤦‍♂ this PR basically just undoes what I did in #5534 to fix another bug. back to the drawing board...

PR #7786, opened 2019-12-29
fix(workspaces): Pass args following "--" when running workspace scripts

### LINKED ISSUE #7776: Yarn workspace does not propagate flags

rally25rs:
fixes #7776 

**Summary**

Arguments following a `--` when running `yarn workspaces run ...` and `yarn workspace package run ...` were not being forwarded to the scripts.

Removed the `commander.originalArgs` nonsense that I introduced in a previous PR, and changed the `workspace` and `workspaces` commands to use `args` as already passed instead.

The `yarn run` command used `args` so it makes sense for `workspace run` to use `args` too.

I had to modify `src/index.js`'s check for commands that should forward args and add `workspace` and `workspaces` to it. Again this is in-line with what already happened for the normal `yarn run` command.

As an added bonus, the normal `scripts don't require "--"` warning that `yarn run` normally prints, is now printed for `workspace` commands (it was not printed before this PR).
```
$ yarn workspaces run start -- one two
yarn workspaces v1.22.0-0
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.

> pkg1
yarn run v1.22.0-0
$ ./start.sh one two
one two
Done in 0.07s.
Done in 0.32s.
```

**Test plan**

* Added unit tests to `__tests__/index.js`

* A manual test can be performed by setting up a basic workspace project:

package.json
```
{
  "name": "yarn-7776",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "private": true,
  "workspaces": [
    "packages/*"
  ]
}
```

packages/pkg1/package.json
```
{
  "name": "pkg1",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "start": "./start.sh"
  }
}
```

packages/pkg1/start.sh
```
#!/bin/bash
echo "ARGS: $@"
```

then run the yarn commands
```
yarn workspace pkg1 run start -- one two three
yarn workspace pkg1 run start one two three
yarn workspaces run start -- one two three
yarn workspaces run start one two three

```

The output from each should be `ARGS: one two three`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
+- Passes arguments follwing `--` when running a workspace script (`yarn workspace pkg run command -- arg`)
+
+  [#7776](https://github.com/yarnpkg/yarn/pull/7776) - [**Jeff Valore**](https://twitter.com/rally25rs)
+
 - Prints workspace names with `yarn workspaces` (silence with `-s`)
 
   [#7722](https://github.com/yarnpkg/yarn/pull/7722) - [**Orta**](https://twitter.com/orta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Please add one entry in this file for each change in Yarn's behavior. Use the sa
 
 ## Master
 
-- Passes arguments follwing `--` when running a workspace script (`yarn workspace pkg run command -- arg`)
+- Passes arguments following `--` when running a workspace script (`yarn workspace pkg run command -- arg`)
 
   [#7776](https://github.com/yarnpkg/yarn/pull/7776) - [**Jeff Valore**](https://twitter.com/rally25rs)
 
--- a/__tests__/commands/workspace.js
+++ b/__tests__/commands/workspace.js
@@ -38,12 +38,8 @@ async function runWorkspace(
   }
 }
 
-// The unit tests don't use commander.js for argument parsing.
-// `originalArgs` is normally passed by index.js so we just simulate it in the tests.
-
 test('workspace run command', (): Promise<void> => {
-  const originalArgs = ['workspace-1', 'run', 'script'];
-  return runWorkspace({originalArgs}, ['workspace-1', 'run', 'script'], 'run-basic', config => {
+  return runWorkspace({}, ['workspace-1', 'run', 'script'], 'run-basic', config => {
     expect(spawn).toHaveBeenCalledWith(NODE_BIN_PATH, [YARN_BIN_PATH, 'run', 'script'], {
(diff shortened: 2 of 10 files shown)


rally25rs:
@arcanis do you know what the errors on circleci are all about?
```
    Error: Cannot find module '/tmp/tmp-124nBkpXZbh4tC6/JSON.stringify((() => require('no-deps'))())'
```
isn't very helpful... 😕

arcanis:
My guess is that `yarn node -p '...'` is now incorrectly interpreted as `yarn node '...'`, hence the `Module not found` on a JS script.

rally25rs:
Ah, yeah, thanks for that help. The old logic had
```
      if (command === commands.run || command === commands.create) {
        preservedArgs += 1;
```
but in my changes I had applied having 1 `preservedArgs` to `node` as well. oops 🤦‍♂

rally25rs:
There is now 1 failing test from appveyor only on node 8, but that test passes locally for me on OSX

```
~/Projects/yarn (fix-workspace-arg-passing-7776) 🐒   node -v
v8.17.0

~/Projects/yarn (fix-workspace-arg-passing-7776) 🐒   yarn test-only __tests__/commands/add.js
yarn run v1.21.1
$ node --max_old_space_size=4096 node_modules/jest/bin/jest.js --verbose __tests__/commands/add.js
 PASS  __tests__/commands/add.js (27.561s)
  ✓ add without --ignore-workspace-root-check should fail on the workspace root (3436ms)
  ...additional output omitted...
  ✓ installing with --pure-lockfile and then adding should keep build artifacts
  ✓ preserves unaffected bin links after adding to workspace package
  ✓ installs "latest" instead of maxSatisfying if it satisfies requested pattern
  ✓ installs "latest" instead of maxSatisfying if no requested pattern
  ○ skipped 2 tests
  nohoist
    ✓ can add nohoist package from workspace (1ms)
    ✓ can add nohoist package from root

Tests:       2 skipped, 61 passed, 63 total
Snapshots:   0 total
Time:        27.622s
Ran all test suites matching /__tests__\/commands\/add.js/i.
✨  Done in 28.75s.
```

I don't have Windows setup to test on, but not sure why this 1 test would fail only for node 8 and pass on node 10

PR #7808, opened 2020-01-08
Fix suggestions for CLI

kddeisz:
At some point the `scripts` object changes to be a map instead of an array, at which point the suggestions functionality broke. This commit includes the fix and a test to make sure it doesn't break again.


arcanis:
Thanks!

PR #7811, opened 2020-01-10
Allow dots in bin names

valerybugakov:
**Summary**

Bin regex is too restrictive at the moment. See comments [here](https://github.com/yarnpkg/yarn/pull/7755).

**Test plan**

Use `dangerous bin name` tests.

### DIFF
--- a/__tests__/__snapshots__/normalize-manifest.js.snap
+++ b/__tests__/__snapshots__/normalize-manifest.js.snap
@@ -62,6 +62,8 @@ Array [
 
 exports[`dangerous bin name: dangerous bin name 1`] = `
 Array [
+  "foo: Invalid bin entry for \\".\\" (in \\"foo\\").",
+  "foo: Invalid bin entry for \\"..\\" (in \\"foo\\").",
   "foo: Invalid bin entry for \\"/tmp/foo\\" (in \\"foo\\").",
   "foo: Invalid bin entry for \\"../tmp/foo\\" (in \\"foo\\").",
   "foo: Invalid bin entry for \\"tmp/../../foo\\" (in \\"foo\\").",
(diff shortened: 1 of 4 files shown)


valerybugakov:
@Just-Zach regex has `i` flag, so it should ignore letter casing?

BYK:
Colons are not allowed in file names on Windows, so not sure if this is a good idea.

Just-Zach:
> @Just-Zach regex has `i` flag, so it should ignore letter casing?

Aah yes, thanks for catching that!

Just-Zach:
> Colons are not allowed in file names on Windows, so not sure if this is a good idea.

As far as I have seen the `key` is not used as a path - The `value` would be used as a filepath but it is validated using the `isValidBin` function. The key is used to reference the value, but isn't used when writing files to disk.

valerybugakov:
@Just-Zach @BYK from the docs it seems that it should be able to create a symlink using name provided a bin `key` and colons are not allowed in symlinks too. So it won't validate  correctly on Windows.

arcanis:
Yep colons aren't safe - can you also check against special cases (`.` and `..`)? It'll probably be hard to exploit but eh 🤷‍♀️

BYK:
>As far as I have seen the key is not used as a path - The value would be used as a filepath but it is validated using the isValidBin(...) function. The key is used to reference the value, but isn't used when writing files to disk.

The `key` is definitely used on the filesystem as that's the name of the created bin/symlink/whatever that is used under `nodel_modules/.bin`. Now with PnP, maybe we can make that happen in-memory but not sure if that's a worthy investment at this point.

valerybugakov:
@arcanis multiple dots on windows are [allowed](https://en.wikipedia.org/wiki/Filename)
@BYK  I'll remove colon from the regex

TheRealChssAddct:
@arcanis @valerybugakov 
Colons are allowed in linux symlinks.  

A real-world example of this is Eclipse Foundation's che theia project which uses a colon for a symlink che:theia which is used extensively as a command-line tool.  See readme and package.json here:  [eclipse/che-theia](https://github.com/eclipse/che-theia/tree/master/generator)

The command-line tool is then extensively used elsewhere in yarn script commands and docker files.

Perhaps that was a bad decision.  But in any case, disallowing colons will create significant rework for this project, and presumably others.  

So, 1.21.1 is definitely a breaking change, whereas 1.21.0 works fine.

saveman71:
Worth noting for future people coming here that `yarn cache clean` is required to get the missing binaries again (i.e. upgrading and installing/removing nodes_modules isn't enough)

PR #7812, opened 2020-01-10
Fixed: workspace entries should not be used when trying to check for peer dependencies

antoine-malliarakis:
Fixed: the package requests introduced by workspace mechanism should be excluded from the scope of requests checked for usage
Added peer tests (all credits go to @hulkish for the initial test cases)

# Summary

The idea of this PR is to fix the following behaviour which were currently giving tons of "false positive" warnings about unused peer dependencies in workspaces context.

* Current state:

* say you have a project A within your workspace which specifies a peer dependency on a package which is *outside* the workspace. Currently you get a peer unmet warning.
* say you have a project B within your workspace which specifies a peer dependency on another project C which is *inside* the workspace and that your root project declares a dependency on B without a dependency on C. Currently you don't get a peer unmet warning.

This relates (somehow) to issue https://github.com/yarnpkg/yarn/issues/5810

# Test plan

## Simple missing peer dependency warning (prod)

Create a root project with three subprojects, `a`, `b` and `c` (we'll be using version `1.0` everywhere here). 
   - The root project has no dependency
   - `a` has `b` in its peer dependencies.
   - `c` has `a` in its prod dependencies

Run `yarn install`
Expected: a warning for `c` missing a peer dependency on `b`.

## Simple missing peer dependency warning (dev)

Create a root project with three subprojects, `a`, `b` and `c` (we'll be using version `1.0` everywhere here). 
   - The root project has no dependency
   - `a` has `b` in its peer dependencies.
   - `c` has `a` in its dev dependencies

Run `yarn install`
Expected: a warning for `c` missing a peer dependency on `b`.

## Simple unmet peer dependency warning (prod)

Create a root project with three subprojects, `a`, `b` and `c` (we'll be using version `1.0` everywhere here). 
   - The root project has no dependency
   - `a` has `b@0.5` in its peer dependencies.
   - `c` has `a` and `b@1.0` in its prod dependencies

Run `yarn install`
Expected: a warning for `c` with an unmet peer dependency on `b@0.5`.

## Simple unmet peer dependency warning (dev)

Create a root project with three subprojects, `a`, `b` and `c` (we'll be using version `1.0` everywhere here). 
   - The root project has no dependency
   - `a` has `b@0.5` in its peer dependencies.
   - `c` has `a` and `b@1.0` in its dev dependencies

Run `yarn install`
Expected: a warning for `c` with an unmet a peer dependency on `b@0.5`.

## No peer dependency issue

Create a root project with three subprojects, `a`, `b` and `c` (we'll be using version `1.0` everywhere here). 
   - The root project has no dependency
   - `a` has `b` in its peer dependencies.
   - `c` has `a` and `b` in its prod dependencies

Run `yarn install`
Expected: no warning

### DIFF
--- a/__tests__/commands/install/workspaces-install.js
+++ b/__tests__/commands/install/workspaces-install.js
@@ -69,6 +69,22 @@ test.concurrent("workspaces warn and get ignored if they don't have a name and a
   );
 });
 
+test.concurrent('should only warn for unmet peer dependency on workspace package', (): Promise<void> => {
+  return buildRun(
+    reporters.BufferReporter,
+    path.join(__dirname, '..', '..', 'fixtures', 'install'),
+    async (args, flags, config, reporter, lockfile): Promise<void> => {
+      const install = new Install(flags, config, reporter, lockfile);
+      await install.init();
+      const warnings = reporter.getBuffer().filter(({type}) => type === 'warning').map(({data}) => data);
+      expect(warnings).toEqual(['" > nomore-test-b@0.0.1" has unmet peer dependency "nomore-test-a@^0.0.1".']);
+    },
+    [],
+    {},
+    'workspaces-install-transient-peer-dep-satisfied',
+  );
+});
+
 test.concurrent('installs workspaces dependencies into root folder', (): Promise<void> => {
   return runInstall({}, 'workspaces-install-basic', async (config): Promise<void> => {
     const lockfile = await fs.readFile(path.join(config.cwd, 'yarn.lock'));
--- a/src/package-linker.js
+++ b/src/package-linker.js
@@ -15,8 +15,8 @@ import {normalizePattern} from './util/normalize-pattern.js';
 import {entries} from './util/misc.js';
 import * as fs from './util/fs.js';
 import lockMutex from './util/mutex.js';
-import {satisfiesWithPrereleases} from './util/semver.js';
 import WorkspaceLayout from './workspace-layout.js';
+import PackagePeerResolver from './package-peer-resolver';
 
 const invariant = require('invariant');
 const cmdShim = require('@zkochan/cmd-shim');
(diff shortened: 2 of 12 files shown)


antoine-malliarakis:
For what it's worth the failing test seems to be also failing on the master branch.

src/package-linker.js
```diff
@@ -604,49 +670,61 @@ export default class PackageLinker {
...
-            continue;
-          }
-          const levelDistance = getLevelDistance(peerPkgRef);
-          if (isFinite(levelDistance) && levelDistance < resolvedLevelDistance) {
-            if (this._satisfiesPeerDependency(range, peerPkgRef.version)) {
-              resolvedLevelDistance = levelDistance;
-              resolvedPeerPkg = peerPkgRef;
-            } else {
-              peerError = 'incorrectPeer';
+        const refTreesStatus = [];
```
antoine-malliarakis: We start by building an array containing the results on a per reference tree basis.
The first one would be the first reference tree found which (by construction) would be the first defined of the following:

* dependency defined at the root level
* first package (according to the platform's sorting conventions) which defines a dependency on that package

src/package-linker.js
```diff
@@ -604,49 +670,61 @@ export default class PackageLinker {
...
+        continue;
+      }
 
       for (const peerDepName in peerDeps) {
         const range = peerDeps[peerDepName];
         const meta = peerDepsMeta && peerDepsMeta[peerDepName];
 
         const isOptional = !!(meta && meta.optional);
 
+        const onNotResolved = isOptional
```
antoine-malliarakis: Nothing to do with optional dependencies when they're not resolved.

src/package-linker.js
```diff
@@ -604,49 +670,61 @@ export default class PackageLinker {
...
-      const getLevelDistance = pkgRef => {
-        let minDistance = Infinity;
-        for (const req of pkgRef.requests) {
-          const distance = refTree.length - req.parentNames.length;
-
-          if (distance >= 0 && distance < minDistance && req.parentNames.every((name, idx) => name === refTree[idx])) {
-            minDistance = distance;
-          }
-        }
+      const refTrees = extractRefTrees(ref);
```
antoine-malliarakis: Here we build an array of "reference trees" which would be sorted by "name" (the workspace's root reference tree carrying the empty name) and where each entry would represent the corresponding dependency path, instead of only taking the first one which was hanging around.

antoine-malliarakis:
Just wondering if there was anything I should do here to proceed to the next stage (or to have it rejected with corresponding feedbacks for what it's worth).

antoine-malliarakis:
Hello @rally25rs 

Sorry for bothering you but I was wondering: is there anything I could do in order to ease the follow up on this PR ? I mean, is it even still relevant to have such a PR with yarn 2 "around the corner" ? Or did I miss something ?

Thanks and all apologize for any inconvenience caused

Regards

__tests__/commands/install/workspaces-install.js
```diff
@@ -69,6 +69,24 @@ test.concurrent("workspaces warn and get ignored if they don't have a name and a
...
+  void,
+> => {
+  return buildRun(
+    reporters.BufferReporter,
+    path.join(__dirname, '..', '..', 'fixtures', 'install'),
+    async (args, flags, config, reporter, lockfile): Promise<void> => {
+      const install = new Install(flags, config, reporter, lockfile);
+      await install.init();
+      const warnings = reporter.getBuffer().filter(({type}) => type === 'warning').map(({data}) => data);
+      expect(warnings).toEqual(['" > nomore-test-b@0.0.1" has unmet peer dependency "nomore-test-a@^0.0.1".']);
```
rally25rs: It's a little weird that the test is named "should not warn" then asserts that there _is_ a warning.
I'm assuming this warning that still shows up in expected and the test assumption is that another one isn't showing?

Maybe change the expectation to read the same as the test name; expect that a certain warning isn't in the collection?

antoine-malliarakis: I rephrased the name of the test to `should only warn for unmet peer dependency on workspace package`.
Actually if the test was failing (e.g. if I keep the old code running) there would be additional unmet peer warnings which would just be wrong. So what I wanted to test here was that there would be only one warning emitted and it would be that one.
The issue I have is that testing for a specific warning message not to appear sounded like an invite for falsely passing tests which are really awkward for this kind of feature to troubleshoot "after the battle".

arcanis:
I'll let you decide, but to be honest I'm not sure it's a good idea to merge "significant" changes to the linker. I have no idea what side effect it may have 😕

I think the main problem I can see is that afaik this case wasn't considered when implementing workspaces (even in the 2.x I didn't think about making peer dependency implicitly define a regular dependency for workspaces), so it's possible that some parts of the application expect the workspace dependencies to be explicitly listed not only as peer dependencies, but also dev dependencies.

arcanis:
(Oops sorry, didn't mean to close 😅)

antoine-malliarakis:
> I'll let you decide, but to be honest I'm not sure it's a good idea to merge "significant" changes to the linker. I have no idea what side effect it may have 😕
> 
> I think the main problem I can see is that afaik this case wasn't considered when implementing workspaces (even in the 2.x I didn't think about making peer dependency implicitly define a regular dependency for workspaces), so it's possible that some parts of the application expect the workspace dependencies to be explicitly listed not only as peer dependencies, but also dev dependencies.

Would you rather, therefore, have this dev been done only on the v2 codeline ? (Would seem fair enough, to be honest)

vpicone:
Would love for this feature to be released in v1. The peer-dep flags can get really rough when working with workspaces!
Screen Shot 2020-03-04 at 3 19 12 PM

antoine-malliarakis:
> Would love for this feature to be released in v1. The peer-dep flags can get really rough when working with workspaces!
> Screen Shot 2020-03-04 at 3 19 12 PM

Actually what this change will fix is those kind of warnings:

image

It may even add new warnings (because it will basically ensure that workspace peer dependencies are appropriately checked).

But I'm all in for advices ;)

antoine-malliarakis:
> > I'll let you decide, but to be honest I'm not sure it's a good idea to merge "significant" changes to the linker. I have no idea what side effect it may have 😕
> > I think the main problem I can see is that afaik this case wasn't considered when implementing workspaces (even in the 2.x I didn't think about making peer dependency implicitly define a regular dependency for workspaces), so it's possible that some parts of the application expect the workspace dependencies to be explicitly listed not only as peer dependencies, but also dev dependencies.
> 
> Would you rather, therefore, have this dev been done only on the v2 codeline ? (Would seem fair enough, to be honest)

@arcanis Actually I just checked and it seems that there is no such issue in yarn 2... So there would be no real need to try "redoing it" on the 2.x codeline.

antoine-malliarakis:
Closing this MR as "out of touch" (lost track, no longer using yarn v1, and may have more impacts than foreseen)

PR #7831, opened 2020-01-22
Fixes arbitrary file write on fetch

arcanis:
**Summary**

In some circumstances it was possible to write files anywhere on the disk when running `yarn install`. Although it's not that much of an issue considering that most installs are done with postinstall scripts, we don't want that to happen precisely to guarantee some level of safety for people actually using `--ignore-scripts`.

**Test plan**

Will add tests after the release.

### DIFF
--- a/src/fetchers/tarball-fetcher.js
+++ b/src/fetchers/tarball-fetcher.js
@@ -136,6 +136,11 @@ export default class TarballFetcher extends BaseFetcher {
       chown: false, // don't chown. just leave as it is
       map: header => {
         header.mtime = now;
+        if (header.linkname) {
+          const basePath = path.posix.dirname(path.join('/', header.name));
+          const jailPath = path.posix.join(basePath, header.linkname);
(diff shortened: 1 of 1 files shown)


jonm01:
CVE-2020-8131 was assigned.
hackerone report: 730239

gbl-peter:
CVE-2020-8131 was [published](https://nvd.nist.gov/vuln/detail/CVE-2020-8131) 2020-02-24, updated 2020-02-27, and is now appearing as a reason-for-failure during a code promotion job in our CI as reported by Anchore.

How does a dependency-package (i.e. one installed with yarn) recover from this state? As a non-yarn user, I am unfamiliar with recovery options. With npm one would simply update npm, but since we aren't using yarn in our build cycle, I'm unsure what the best course of action is.

arcanis:
@gbl-peter I'm sorry, I have no idea what you mean. I suggest asking your devops team where they use Yarn, and ask them to upgrade.

PR #7848, opened 2020-01-27
handle failed policy set-version

olingern:
**Summary**

Right now `policies set-version` does not handle invalid scenarios where an invalid github token is provided or the request fails.

**Test plan**

Will add a test that mocks a 401 from Github's API and that the reporter correctly logs.

### DIFF
--- a/src/cli/commands/policies.js
+++ b/src/cli/commands/policies.js
@@ -122,9 +122,16 @@ const {run, setFlags, examples} = buildSubCommands('policies', {
       bundleUrl = 'https://github.com/yarnpkg/berry/raw/master/packages/berry-cli/bin/berry.js';
       bundleVersion = 'berry';
     } else {
-      const releases = await fetchReleases(config, {
-        includePrereleases: allowRc,
-      });
+      let releases = [];
+
+      try {
+        releases = await fetchReleases(config, {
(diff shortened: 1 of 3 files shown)


src/util/request-manager.js
```diff
@@ -428,6 +433,13 @@ export default class RequestManager {
           }
         }
 
+        const server = res.caseless.get('server');
+
+        if (res.statusCode === 401 && server === "GitHub.com") {
```
olingern: Not sure if a constant is worth using here. Request manager is probably too generic in some ways. Would be nice to have a Github only request manager so stuff like this could be avoided. For now, I don't think this is terrible.

src/cli/commands/policies.js
```diff
@@ -122,9 +122,16 @@ const {run, setFlags, examples} = buildSubCommands('policies', {
...
-      });
+      let releases = [];
+
+      try {
+        releases = await fetchReleases(config, {
+          includePrereleases: allowRc,
+        });
+      } catch (e) {
+        reporter.error(e.message);
+        return;
```
olingern: In the event of a promise rejection, report and return early

arcanis:
Thanks 👍

PR #7920, opened 2020-02-20
Faster yarn install with faster cycles detection

VincentBailly:
**Summary**

Detecting graph cycles during package install is currently done in o(n2), this PR is a suggestion on how we can bring it down to o(n).

benchmark: warm yarn cache, clean repo (no node_modules folders), running `yarn install`
Before: 5m45
After: 2m10

**Test plan**
This PR changes only the implementation details so it should be tested by the existing tests.

### DIFF
--- a/src/package-install-scripts.js
+++ b/src/package-install-scripts.js
@@ -292,7 +292,24 @@ export default class PackageInstallScripts {
     }
   }
 
+  addDependenciesToGraph(patterns: Array<string>, graph: Map<string, Array<string>>): void {
+    for (const pattern of patterns) {
+      if (graph.has(pattern)) {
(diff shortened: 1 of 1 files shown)


VincentBailly:
#7921

VincentBailly:
@arcanis the implementation here is probably too naive, I am happy to get feedback.
Some tests fail for me locally but I am not sure why, maybe they contain some assumptions into some implementation details.

JCMais:
Looks like this was never merged and the PR now has too many conflicts 😢

billouboq:
Just use midgard-yarn and you will that change included :

https://www.npmjs.com/package/midgard-yarn

Issue #7939, opened 2020-02-27
Exception on windows with non-ascii character username


liuxingbaoyu:
```
lerna ERR! yarn install --mutex network:42424 --non-interactive --no-lockfile --ignore-engines stderr:
internal/modules/cjs/loader.js:969
    throw err;
    ^

Error: Cannot find module 'C:\Users\娴佹槦鏆撮洦\.yarn\bin\yarn.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:17)
    at Function.Module._load (internal/modules/cjs/loader.js:859:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

I will open a pr later.

PR #7940, opened 2020-02-27
Fix: Exception on windows with non-ascii character username

liuxingbaoyu:
**Summary**

Fix: Exception on windows with non-ascii character username  [#7939]

**Test plan**

Sorry, I don't know how to add tests.


BYK:
Can you please explain what the underlying cause is and why did you chose this specific solution?

liuxingbaoyu:
@BYK The default encoding of cmd on some computers is not utf8. When encountering a non-ascii username, yarn will give an error, so I changed the encoding to `65001(utf8)` through `chcp`.

PR #7946, opened 2020-03-03
fix: make `yarn bin` exit non-zero when a binary could not be found

joscha:
**Summary**

Makes `yarn bin` exist with exit code `1` if a binary could not be found. Current behaviour is:

joscha_Joschas-MacBook-Pro____work_canva

which is surprising.

**Test plan**

If the change is accepted I can add a test for the `bin` command.

cc @dillon-giacoppo


merceyz:
Closing as fixed in v2

Issue #2090, opened 2016-11-30
Yarn needs to use .npmignore file when installing from repo


FredyC:
**Do you want to request a *feature* or report a *bug*?**
Feature (I would like think of it as a bug too)

**What is the current behavior?**
Trying to install `three.js` module from repository instead of registry yields **303MB** folder in size. Installing same thing with NPM and resulting folder is just **11MB**

**If the current behavior is a bug, please provide the steps to reproduce.**
`yarn init`
`yarn add github:mrdoob/three.js`

**What is the expected behavior?**
The `.npmignore` file should be used and actually remove files that conforms to globs there. Not sure if it's possible to actually avoid unpacking those files in first place. That would certainly improve performance much more.

Reason why to install package from repository is because of custom forks.

**Please mention your node.js, yarn and operating system version.**
Node 6.9.1, Yarn 0.17.7, Windows 10 x64

Daniel15:
>  module from repository instead of registry

What's the use case for this? Is it just to install a development version?

FredyC:
@Daniel15 Already said that in original post...

> Reason why to install package from repository is because of custom forks.

Or actually any private modules as well.

Daniel15:
Ahhh, so sorry @FredyC, I totally missed that part of your post! For custom forks of open source libraries, one option is to publish them to npm as a namespaced/scoped package (like `@fredyc/foo`), which should behave better than installing it directly from the repo (in theory).

In any case, this definitely sounds like an improvement we should make :)

FredyC:
I personally dislike polluting npm registry more than is needed, even with something like scoped packages. Doesn't feel right, but I guess it's possible workaround for now.

Either way it wont work for private modules anyway.

domakas:
This is a huge deal-breaker for us, would be nice to solve it as soon as possible.

Any ideas what strategy could be used for this? I might be interested in solving this issue myself, however, I don't have any other ideas as just checking out full repo and deleting files which match .npmignore/.yarnignore

bazyli-brzoska:
Same problem occurs when you try to install from a local directory https://github.com/yarnpkg/yarn/issues/685.

rmm5t:
Just ran into this problem myself. I was really surprised by the behavior, and it's causing the associated directory in our `node_modules` to sit at 12MB when using `yarn install`, but `npm install` correctly ignores the files from the repo leaving the directory at just 32KB.

[...]

PR #8010, opened 2020-03-23
Use npmignore file on fetch

wethil:
**Summary**

I created this PR for fix [2090](https://github.com/yarnpkg/yarn/issues/2090)

Normally yarn doesn't use the **.npmignore** and install unnecessary files to our **node_modules**  In this branch, while fetching process, yarn read the **.npmignore** file and delete ignored files with respecting the globs.

I added the [del](https://www.npmjs.com/package/del) package to delete files with respecting globs, on the other hand, the other way was using the **fs.unlink** with a loop of file list on _npmignore_.  But I am not sure if it lets us do the proccess without any error, for example deleting a directory or files except the ones that wanted to be kept with **NEGATE**  mark

**Test plan**

I wrote the test on  _/__tests__\/fetchers.js/_  file.    You can run it with 
`alias testnow="node --max_old_space_size=4096 node_modules/jest/bin/jest.js --verbose  __tests__/fetchers.js"` (it will only make test of fetchers.js run)

### DIFF
--- a/src/package-fetcher.js
+++ b/src/package-fetcher.js
@@ -9,6 +9,9 @@ import * as fetchers from './fetchers/index.js';
 import * as fs from './util/fs.js';
 import * as promise from './util/promise.js';
 
+import {NPM_IGNORE_FILENAME} from '../src/constants.js';
+
+const path = require('path');
 const ssri = require('ssri');
 
 async function fetchCache(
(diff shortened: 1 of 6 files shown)


src/package-fetcher.js
```diff
@@ -87,6 +92,20 @@ export async function fetchOneRemote(
   }
 }
 
+export async function unlinkIgnoredFiles(dest: string): Promise<void> {
```
cpojer: Thank you for working on this! I'm not entirely sure if this is the right place to apply this filter. Could we instead apply this while we unpack so that ignored files are never even written? If we scope it to git dependencies only it shouldn't have an impact on install time.

wethil: Your welcome! Okay, I will work on it.

alshdavid: Thanks for this, how'd you get on? :)

wethil: hey! sorry, i have been super busy, dealing with the conditions that bring with Covid, but I started to working on again

PR #8030, opened 2020-03-31
feature(autoclean): Include more CI configs in default .yarnclean

PeterDaveHello:
**Summary**

autoclean is a good feature, and it'll be great to extend its default .yarnclean file to make it more powerful!

**Test plan**

Will... there is no really code level changes here, I believed the codes are already covered by `__tests__/commands/autoclean.js`, but if it's expected to extend the test case, I'd also be happy to do so.


PeterDaveHello:
@arcanis would you like to help review this change? Thanks!

Issue #4884, opened 2017-11-08
Problem parsing .yarnrc


mobitum:
**Do you want to request a *feature* or report a *bug*?**

Seems like bug in parsing .yarnrc file in project folder.
I'm using yarn 1.3.2 on Windows. 

**What is the current behavior?**
I created .yarnrc file with single line:
--install.modules-folder "./resources"
I'm getting following error on simple 'yarn' command:
_"error An unexpected error occurred: "Unknown token: { line: 1, col: 0, type: 'INVALID', value: undefined } 1:0 in C:\\VS...."_
I tried even **_empty .yarnrc_** with the same result.
My package.json looks like this:
{
  "name": "intranet",
  "version": "1.0.0",
  "license": "UNLICENSED",
  "devDependencies": {},
  "dependencies": {
    "angular": "^1.6.6",
    "angular-material": "^1.1.5"
  }
}

**If the current behavior is a bug, please provide the steps to reproduce.**

Just put empty .yarnrc file in your project directory and run yarn command

**What is the expected behavior?**
.yarnrc should be parsed against available options

**Please mention your node.js, yarn and operating system version.**
Yarn version: 
  1.3.2

Node version: 
  8.9.1

Platform: 
  win32 x64

BYK:
@mobitum we need the actual .yarnrc file to be able to investigate. I work on Windows and I have custom `.yarnrc` files and I have no issues. We also have Windows tests running for each PR on AppVeyor so it is very likely that your system or that specific lockfile has an issue.

BYK:
Also, providing the full error, including the path and your project path would be useful since failing with a `.yarnrc` error in a package that doesn't have a lockfile implies that you have another lockfile laying around, maybe in your home folder.

mobitum:
This is test I performed - created empty .yarnrc file from my Power Shell console:

PS C:\VS\INTRANET\Intranet\Intranet\Intranet> **yarn**
_yarn install v1.3.2
error An unexpected error occurred: "Unknown token: { line: 1, col: 0, type: 'INVALID', value: undefined } 1:0 in C:\\VS\\INTRANET\\Intranet\\Intranet\\Intranet\\.yarnrc".
info If you think this is a bug, please open a bug report with the information provided in "C:\\VS\\INTRANET\\Intranet\\Intranet\\Intranet\\yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command._
[...]

PR #8128, opened 2020-05-16
bugfix #4884 On FreeBSD read(2) can read directories. Detect them before read.

madpilot78:
**Summary**

This is a fix for issue #4884 which shows up on FreeBSD and other OSes where it is possible to read(2) a directory (depending on the filesystem actually) and get back some binary representation of it.

This causes yarn to file as described in issue #4884.

In the FreeBSD ports tree we are installing yarn with another patch fundamentally backing out commit 2cb2fc480. This is suboptimal, also because we are patching the compiled bundle at present. Having a proper fix included upstream would be great.

I'll update the CHANGELOG in a new commit on this same pull request shortly.

**Test plan**

I tested it locally by checking that the same operations failing with a clean install are not happening with the new bundle.

### DIFF
--- a/src/util/rc.js
+++ b/src/util/rc.js
@@ -67,9 +67,7 @@ function parseRcPaths(paths: Array<string>, parser: Function): Object {
     ...paths.map(path => {
       try {
         if (statSync(path).isDirectory()) {
-          const e = new Error();
-          e.code = 'EISDIR';
-          throw e;
+          return {};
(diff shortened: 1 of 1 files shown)


madpilot78:
> I'll update the CHANGELOG in a new commit on this same pull request shortly.

Actually this does not look necessary, since I'm not changing behavior.

Issue #8012, opened 2020-03-23
OOM on `yarn audit`


taxonomic-blackfish:
### Bug description

Due to https://www.npmjs.com/advisories/1488 and https://www.npmjs.com/advisories/1490, running `yarn audit` on a midsized project will OOM and crash.

**Command**

```sh
yarn audit --json
```

**What is the current behavior?**

Prior to those two advisories, `yarn audit` completed successfully.

**What is the expected behavior?**

`yarn audit` should not OOM because a commonly used package (that appears as a dep to many packages) has a vuln.

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x9ef190 node::Abort() [node]
 2: 0x9f13b2 node::OnFatalError(char const*, char const*) [node]
 3: 0xb5da9e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xb5de19 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xd0a765  [node]
 6: 0xd0adf6 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [node]
 7: 0xd1760a v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
 8: 0xd18515 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 9: 0xd19c8c v8::internal::Heap::CollectAllAvailableGarbage(v8::internal::GarbageCollectionReason) [node]
10: 0xd1b026 v8::internal::Heap::AllocateRawWithRetryOrFail(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
11: 0xce8c8c v8::internal::Factory::NewRawOneByteString(int, v8::internal::AllocationType) [node]
12: 0x10a2305 v8::internal::IncrementalStringBuilder::Extend() [node]
13: 0xdff298 v8::internal::JsonStringifier::SerializeString(v8::internal::Handle<v8::internal::String>) [node]
[...]

PR #8214, opened 2020-07-03
Avoid OOMs in large audit advisories

### LINKED ISSUE #7404: yarn audit --json produces large amounts of data
Bug

When running yarn audit against a couple of different projects and saving them to a file, the resulting file size is around 30 MB. When running against the same project but with the json flag, the file size is around 19 GB. A factor of 10x or even up to a 100x seems reasonable, but a ~633x seems like a bit much.
 
1. Run `yarn audit > sample.txt` against [this repo](https://github.com/coinbase/rest-hooks).
2. Run `ls -l sample.txt` to view the file size.
3. Run `yarn audit > sample.json` against the same repo.
4. Run `ls -l sample.json` to view the file size.

A file size similar to running yarn audit.

yarn v. 1.16.0, node.js v 12.6.0, up to date OSX,

glenjamin:
**Summary**

Fixes #7404 and #8012 
Would supercede #7495 with the benefit of not changing the interface contract

At the moment running `yarn audit --json` on a project with a lot of dependencies when a common dependency like lodash is marked as vulnerable is likely to OOM.

This doesn't affect the `console` output, because it prints out a table per-finding, wheras the JSON output prints out a line per-advisory.

Changing the JSON output format to avoid this would arguably be quite a major breaking change, so instead i've done a bit of a trick to encode a smaller chunk of JSON output at a time, effectively streaming it.

The implementation i've done here is a little hacky, although it is contained to a small function and passes all the existing tests. I'm happy to refine the approach taken if the maintainers are aligned with the idea.

**Test plan**

`yarn audit --json` still works.

I haven't yet tested this against a repo that OOMed before, but I intend to do that today and I'm confident that it would work.


taxonomic-blackfish:
This PR will still OOM:

Running `yarn-local audit --json | grep auditAdvisory | jq .` on a project with 56 instances of `lodash` (counted by `rg lodash yarn.lock | wc -l`) still throws an OOM after the first element is processed. My suspicion is that the OOM is caused not by a large `paths` field on a single `auditAdvisory` but by generating `auditAdvisory`s faster that they can be printed.

glenjamin:
Ah I see, the issue seems to be down to the yarn process not checking for backpressure on the writes to the stdout stream.

I'll see if I can fix this by checking those events and pausing output.

pxwise:
fwiw, we were hit pretty hard by OOMs in our build process with the recent lodash vulnerability running `yarn audit --json`. Fix efforts here appreciated.

Issue #683, opened 2016-10-11
Bug: running 2 (maybe more) yarn commands on different projects the same time fails


neophob:
I installed yarn the first time, then i ran `yarn`on project A, seconds later I ran `yarn` on project B.

Both installation failed, project A output:

```
yarn install v0.15.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Couldn't find a package.json (or bower.json) file in /Users/USERNAME/.yarn-cache/npm-yargs-3.10.0
    at /usr/local/lib/node_modules/yarnpkg/lib-legacy/config.js:363:13
    at next (native)
    at step (/usr/local/lib/node_modules/yarnpkg/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /usr/local/lib/node_modules/yarnpkg/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:20
    at run (/usr/local/lib/node_modules/yarnpkg/node_modules/core-js/library/modules/es6.promise.js:87:22)
    at /usr/local/lib/node_modules/yarnpkg/node_modules/core-js/library/modules/es6.promise.js:100:28
    at flush (/usr/local/lib/node_modules/yarnpkg/node_modules/core-js/library/modules/_microtask.js:18:9)
    at nextTickCallbackWith0Args (node.js:415:9)
    at process._tickCallback (node.js:344:13)
```

project B output:

```
yarn install v0.15.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error Couldn't find a package.json (or bower.json) file in /Users/USERNAME/.yarn-cache/npm-yargs-4.8.1
    at /usr/local/lib/node_modules/yarnpkg/lib-legacy/config.js:363:13
    at next (native)
    at step (/usr/local/lib/node_modules/yarnpkg/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /usr/local/lib/node_modules/yarnpkg/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:20
    at run (/usr/local/lib/node_modules/yarnpkg/node_modules/core-js/library/modules/es6.promise.js:87:22)
    at /usr/local/lib/node_modules/yarnpkg/node_modules/core-js/library/modules/es6.promise.js:100:28
[...]

PR #8260, opened 2020-07-27
[Fixes #7368] Increase mutex staleness interval to 5 minute

### LINKED ISSUE #7368: Error: Unable to update lock within the stale threshold
When using yarn with mutex option (type file) run in parallel on many threads (8) I get following error:

~/.gradle/yarn/yarn-v1.7.0/lib/node_modules/yarn/lib/cli.js:78487
    compromised = compromised || function (err) { throw err; };
                                                  ^
Error: Unable to update lock within the stale threshold
    at options.fs.utimes (/home/build/.gradle/yarn/yarn-v1.7.0/lib/node_modules/yarn/lib/cli.js:78409:66)
    at FSReqWrap.oncomplete (fs.js:135:15)

Obviously timeout has been reached. Is it possible to fix it by adjustting some timeout configuration?

karelbilek:
**Summary**

Fixes #7368 

We hit this when the installation is busy with building packages with C++/clang for too long. The mutex gets stale after 10 seconds, probably because the main thread is busy, and everything fails.

(as using mutex is the official "solution" for the issue #683 , it would be good to have it actually work :( )

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -311,7 +311,7 @@ export async function main({
   const runEventuallyWithFile = (mutexFilename: ?string, isFirstTime?: boolean): Promise<void> => {
     return new Promise(resolve => {
       const lockFilename = mutexFilename || path.join(config.cwd, constants.SINGLE_INSTANCE_FILENAME);
-      lockfile.lock(lockFilename, {realpath: false}, (err: mixed, release: (() => void) => void) => {
+      lockfile.lock(lockFilename, {stale: 5 * 60 * 1000, realpath: false}, (err: mixed, release: (() => void) => void) => {
         if (err) {
           if (isFirstTime) {
             reporter.warn(reporter.lang('waitingInstance'));


karel-3d:
However, I'm wrong. This doesn't fix anything, because big brains from proper-lockfile have this line

https://github.com/moxystudio/node-proper-lockfile/blob/094a8af1b1c2c9703b849b62a962f7e91bdb26b8/lib/lockfile.js#L219

amazing

justfalter:
> However, I'm wrong. This doesn't fix anything, because big brains from proper-lockfile have this line
> 
> https://github.com/moxystudio/node-proper-lockfile/blob/094a8af1b1c2c9703b849b62a962f7e91bdb26b8/lib/lockfile.js#L219
> 
> ```
>     options.stale = Math.max(options.stale || 0, 2000);
> ```
> 
> ignoring the options basically. amazing

@karelbilek (I realize it has been many months since you've updated) --- I'm not sure I follow with what is wrong with the line you quote? Your PR is supplying `5 * 60 * 1000` or `300000`, which will evaluate as:
```
> Math.max(5*60*1000 || 0, 2000)
300000
```

`proper-lockfile` is taking the argument, not ignoring it. The code is just saying that it wants to have a minimum staleness check of 2 seconds, probably just to avoid updating/checking the timestamp too often. 

As for your choice of 5 minutes as opposed to 10 seconds --- I'd prefer that the value be exposed for configuration through some means, as 5 minutes is a very long time for some people. If my understanding of `proper-lockfile` is correct, anything waiting on the lockfile will check once every `options.stale / 2` seconds. With `options.stale` configured for 5 minutes, this means that process A (who holds the lock) could exit, and process B might not notice that the lock is available for another 2.5 minutes.

karelbilek:
Oh, I misread Max as Min I guess. 

I don't remember what was the issue anymore, honestly. (Sorry)

I forgot what exactly was the issue; we (at my company) slowly realized that yarn is just not made for multiple CLIs with single cache and the development of yarn stalled, and we swiched to pnpm that has a working cache that can be accessed from more CLIs

merceyz:
Closing as fixed in v2 where the cache (mutex) doesn't have this issue

PR #8270, opened 2020-08-03
Run prepack script if it exists when installing from git

MOZGIII:
**Summary**

Run `prepack` script if it exists when installing from git.

According to the https://docs.npmjs.com/misc/scripts, the `prepack` has to be run on `git` clone.

**Test plan**

TODO


merceyz:
Closing as fixed in v2

Nugine:
Did it fix in v1.22?

PR #8312, opened 2020-08-22
feat: improve error on parsing `files` in `package.json` (closes #8311)

### LINKED ISSUE #8311: [feat] improve error on parsing `files` in `package.json`
Right now if you have a String instead of an Array, it will yield this error on `yarn publish`:

```sh
Trace: 
  TypeError: onlyFiles.map is not a function
      at /Users/jack/.yarn/lib/cli.js:42596:38
      at step (/Users/jack/.yarn/lib/cli.js:310:30)
      at /Users/jack/.yarn/lib/cli.js:321:13
```

I suggest that we improve this error message to check the type, and if it is not an Array, then alert to the user that the `files` property in `package.json` is a `${typeof prop}` as opposed to erring on `.map` not being a function as it currently does.

Ref: <https://github.com/yarnpkg/yarn/issues/7815>

niftylettuce:
**Summary**

**Test plan**

### DIFF
--- a/src/cli/commands/pack.js
+++ b/src/cli/commands/pack.js
@@ -82,6 +82,9 @@ export async function packTarball(
 
   // `files` field
   if (onlyFiles) {
+    // <https://github.com/yarnpkg/yarn/issues/8311>
+    if (!(onlyFiles instanceof Array)) throw new MessageError('"files" property in package.json must be an Array');
+    
     let lines = [
       '*', // ignore all files except those that are explicitly included with a negation filter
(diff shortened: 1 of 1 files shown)


niftylettuce:
not going to do this, good luck

JacobBovee:
@rally25rs Hey Jeff just wanted to let you know that I've made a PR to close this issue with the feedback you've included. If this is still something worth fixing I'd love it if you could take a look. PR #8370

Issue #8302, opened 2020-08-17
After `yarn install`, Yarn does not prune scoped projects in dependent node_modules


greyepoxy:
### Bug description

If an existing `node_modules` directory exists with 2+ scoped packages in a dependent's node_modules directory. Then after an upgrade where one of the scoped packages is hoisted, the hoisted scoped package should be deleted from the dependent's node_modules directory. 

**What is the current behavior?**

The scoped package remains in the dependent's node_modules directory resulting in it depending on the wrong version of the package.

**What is the expected behavior?**

The scoped package should be deleted so the dependent will depend on the hoisted version of the package.

**Steps to Reproduce**

I have setup a repository with reproduction instructions [here](https://github.com/greyepoxy/yarn-not-pruning-scoped-projects-issue-example).

Here is the general scenario that will result in the problem:
Say you have a `node_modules` structure like so

- `@some-scope/shared-dependency-x@1.0.0`
- `@some-scope/shared-dependency-y@1.0.0`
- `dependency-a@1.0.0`
- `dependency-b@1.0.0`
  - `@some-scope/shared-dependency-x@1.5.0`
  - `@some-scope/shared-dependency-y@2.0.0`

So `dependency-a` depends on `@some-scope/shared-dependency-x@1.0.0` and `@some-scope/shared-dependency-y@1.0.0`, and `dependency-b` depends on `@some-scope/shared-dependency-x@1.5.0` and `@some-scope/shared-dependency-x@2.0.0`.

You then upgrade the dependencies in the project which results in new versions of `dependency-a` and `dependency-b` that both depend on `@some-scope/shared-dependency-x@2.0.0` (and this is reflected in the `yarn.lock` file).

You would expect a `yarn install` to result in a node_modules tree with only a single top level `@some-scope/shared-dependency-x@2.0.0`.

Instead `yarn install` produces the following

- `@some-scope/shared-dependency-x@2.0.0`
- `@some-scope/shared-dependency-y@1.0.0`
- `dependency-a@2.0.0`
- `dependency-b@2.0.0`
  - `@some-scope/shared-dependency-x@1.5.0` **<-- this should not be here**
  - `@some-scope/shared-dependency-y@2.0.0`

**Notes**

[...]

PR #8318, opened 2020-08-25
Greyepoxy/handle scoped packages being hoisted

greyepoxy:
**Summary**

Fixes https://github.com/yarnpkg/yarn/issues/8302

**Test plan**

Validated that the reproduction listed in the issue was fixed

### DIFF
--- a/__tests__/commands/install/integration-hoisting.js
+++ b/__tests__/commands/install/integration-hoisting.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import {getPackageVersion, isPackagePresent, runInstall} from '../_helpers.js';
+import {run as add} from '../../../src/cli/commands/add.js';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000;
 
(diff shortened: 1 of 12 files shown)


greyepoxy:
I have been attempting to fix the CI and so far have not had any luck. Looks like my new test is consistently failing but only on CI.

Locally on windows 10 I see it passing consistently (tried with node versions 12.18.3, 10.22.0, and 8.17.0, both 32 and 64 bit versions).

Looks like path length "might" be the cause (I was seeing the same CI failure locally with the original longer dependency names) but even after shortening them and double checking through console logs that no file path in the test directory was over 240 characters, I still see the CI failing.

At this point I am not sure how to proceed, looks to me like the `yarn add` is silently failing to replace directories in the test's `node_modules` but its unclear why that is the case. 

Would appreciate assistance

__tests__/commands/install/integration-hoisting.js
```diff
@@ -52,3 +55,49 @@ test.concurrent(
...
+      if (lstat.isSymbolicLink() || lstat.isFile()) {
+        return [currentPath];
+      }
+
+      const files = await fs.readdir(currentPath);
+      const subFiles = files.map(file => getAllFilePaths(path.join(currentPath, file)));
+      return (await Promise.all(subFiles)).reduce((previous, current) => previous.concat(...current), []);
+    };
+    const allFilePathsInTestDirectory = await getAllFilePaths(config.cwd);
+    config.reporter.info(allFilePathsInTestDirectory.join('\n'));
```
greyepoxy, line 93: This reporter info call was for debugging the CI only, will remove it once the CI is passing

PR #8337, opened 2020-09-09
Ensure npm_config_* variables aren't set twice on Windows

### LINKED ISSUE #8334: Yarn on Windows breaks apps that read from the environment variables
### Bug description

yarn sets up some environment variables (like npm_config_cache) and calls execSync. This calls the Win32 API CreateProcess and passes an environment block which is a merge of the existing environment block (i.e. from the calling process) with some additional stuff that includes npm_config_cache among others.

Environment variables in Windows are case-insensitive, but CreateProcess does not sanitize the environment block. This is a problem because if the calling process had set up NPM_CONFIG_CACHE (all caps), then yarn will just set the lowercase variant of the variable, and node will call CreateProcess which will contain the variable twice. 
This breaks some tools that create dictionaries out of the environment block because they expect to run into each variable exactly once (regardless of casing).

In my case we have a nodejs script that we invoke via yarn, and the nodejs script calls a Windows build utility (msbuild) which crashes when trying to set up options for the compiler when it finds this option twice. We run into this when building in our CI in Azure DevOps pipeline, which sets up NPM_CONFIG_CACHE in the environment.

Related:

https://github.com/actions/virtual-environments/pull/1566/files
dotnet/msbuild#5726

**Command**

```sh
yarn [your command]
```

when yarn spawns the nodejs script, it does so using an environment block that contains two copies of an environment variable, with different casing. This is illegal.

Yarn should sanitize the env block it passes

**Steps to Reproduce**
in cmd:
set NPM_CONFIG_CACHE=foobar

in package.json add this script:
    "test": "echo console.log(process.env) | node -"

run `yarn test` and see that there are two copies of npm_config_cache in the new process environment block:

image

**Environment**

- Node Version: 12.9.1
- Yarn v1 Version: `1.22.4` 
- OS and version: Windows 10 19041

asklar:
**Summary**
Fixes #8334 

This fixes a bug where an yarn is wrongly setting an environment variable twice

**Test plan**

Validated that only the original casing of an environment variable is sent to the child process when yarn calls spawn()

### DIFF
--- a/__tests__/fixtures/lifecycle-scripts/npm_config_env_win32/log-command.js
+++ b/__tests__/fixtures/lifecycle-scripts/npm_config_env_win32/log-command.js
@@ -1,0 +1,3 @@
+const execSync = require('child_process').execSync;
+
+console.log(execSync('cmd /c set NPM_CONFIG').toString());
--- a/__tests__/fixtures/lifecycle-scripts/npm_config_env_win32/package.json
+++ b/__tests__/fixtures/lifecycle-scripts/npm_config_env_win32/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "npm_config_env_win32",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "scripts": {
+    "prepublish" : "node log-command.js",
+    "pretest" : "node log-command.js",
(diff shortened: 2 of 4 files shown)


asklar:
@arcanis could you take a look?

PR #8343, opened 2020-09-15
Add a preinstall script to bypass npm's ownership check

arcanis:
**Summary**

Cf the comment in the preinstall script for more details.

Ref: https://github.com/arcanis/pmm/issues/6

**Test plan**

It's difficult to test in real circumstances given the our test pipelines. Will build the archive and publish it as a different name before landing it.


isaacs:
This is a known (if minor) security vulnerability I intend to address.  Please do not rely on it continuing to work.  We should be checking for bin conflicts before running preinstall scripts.  Sorry.

arcanis:
@isaacs Gladly - perhaps solving https://github.com/arcanis/corepack/issues/6 first would be a good idea, though, since you reached this issue from there.

Issue #8534, opened 2021-01-03
Nightly build has invalid semver


mhassan1:
The nightly build of Yarn v1 has invalid semver (currently `yarn-v1.23.0-20210102.0010`). It is invalid because of the leading zero after the `.` near the end:

```js
require('semver').valid('1.23.0-20200928.0848') // -> null
require('semver').valid('1.23.0-20200928.1848') // -> "1.23.0-20200928.1848"
```

This seems to be caused by the generation of the version [here](https://github.com/yarnpkg/yarn/blob/14bcd15a401e063c2130ab31568c80aeb914165f/scripts/set-dev-version.js#L16):

```js
const formattedDate =
  date.getUTCFullYear() +
  leftPad(date.getUTCMonth() + 1) +
  leftPad(date.getUTCDate()) +
  '.' +
  leftPad(date.getUTCHours()) +
  leftPad(date.getUTCMinutes());
```

Is there any appetite for changing this so that it does not generate invalid semver? Perhaps using a `-` instead of a `.` before the time? Or prepending a non-zero before the hour? I don't know if there's a dependency on the current format.

In our case, this causes breakage to a Yarn v2 workflow that packs a GitHub dependency using Yarn Classic as part of the project installation.

merceyz:
Closing as fixed in https://github.com/yarnpkg/yarn/commit/3119382885ea373d3c13d6a846de743eca8c914b

dchambers:
@merceyz, could this change please be backported to a `2.4.2` release? Otherwise from `2.4.1` you have to temporarily `yarn set version from sources` before you can `yarn set version 3.0.0-rc.2`, and it won't be obvious to newcomers that this may work.

And that's not to mention that in my case I also had to `yarn plugin remove @yarnpkg/plugin-workspace-tools` before performing the version upgrade, and then re-issuing `yarn plugin import @yarnpkg/plugin-workspace-tools` once the version upgrade was complete.

merceyz:
Did you comment on the wrong issue? This is for nightly releases of Yarn 1.x

> And that's not to mention that in my case I also had to yarn plugin remove @yarnpkg/plugin-workspace-tools before performing the version upgrade, and then re-issuing yarn plugin import @yarnpkg/plugin-workspace-tools once the version upgrade was complete.

We fixed that in https://github.com/yarnpkg/berry/pull/2848 so upgrading from 2.x to 3.0.0-rc.3 should not have that problem

dchambers:
[...]

PR #8535, opened 2021-01-03
Updates set-dev-version.js to set version to valid semver

mhassan1:
**Summary**

Updates `set-dev-version.js` to set version to valid semver. Resolves https://github.com/yarnpkg/yarn/issues/8534.

**Test plan**

- Run `./scripts/set-dev-version.js`
- See that output always contains a valid semver ending with a valid time

### DIFF
--- a/scripts/set-dev-version.js
+++ b/scripts/set-dev-version.js
@@ -18,8 +18,9 @@ const formattedDate =
   leftPad(date.getUTCMonth() + 1) +
   leftPad(date.getUTCDate()) +
   '.' +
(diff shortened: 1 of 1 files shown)


mhassan1:
@merceyz Here is a candidate PR based on our brief exchange:

> @merceyz: I'm unsure if we should change the format or since the time doesn't really matter we could just use 4 random numbers that are valid as a unit of time

merceyz:
Closing as fixed in https://github.com/yarnpkg/yarn/commit/3119382885ea373d3c13d6a846de743eca8c914b

PR #8543, opened 2021-01-08
fix: pass SIGTERM to spawned yarn process (via `yarnPath`)

andreialecu:
**Summary**

`SIGTERM` would not be passed properly to yarn itself if `yarnPath` was being used

Originally reported at:

https://github.com/yarnpkg/berry/pull/2347#issuecomment-756720797

**Test plan**

Screenshot 2021-01-08 at 15 36 17

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -634,6 +634,8 @@ async function start(): Promise<void> {
       // innermost process, whose end will cause our own to exit.
     });
 
+    handleSignals();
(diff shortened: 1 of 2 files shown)


merceyz:
Merged as https://github.com/yarnpkg/yarn/commit/d64f979e5ed62dd848b776ac27b26482619e53b3 and released in 1.22.11

Issue #8590, opened 2021-03-03
[Bug] .bin resolving order is wrong when using `yarn workspace`


sglim:
### Expected

In yarn workspace, `yarn run` command use package dependent binary instead of root one.

### Actual

`yarn run` uses root binary when there are different binary versions

### How to reproduce?

See https://github.com/sglim/yarn-workspace-bin-bug

rally25rs:
I believe this is as-designed. I ordered the `.bin` resolution specifically in that order a few years ago https://github.com/yarnpkg/yarn/issues/4543#issuecomment-341682538 because if you are running a command for a specific package, it should use that package's explicit dependency before falling back to the workspace root.

When you run `yarn` commands from within a workspace package, yarn considers you to be operating on that package. In your example, `cd packages/workspace-1 ; yarn add leftpad` would add `leftpad` as a dependency to `workspace-1` 's package.json only, not to `workspace-2` or to the workspace root.

Likewise, running `yarn run` from within the `packages/workspace-1` directory considers you to be operating on `workspace-1` and will use it's `.bin` folder first.

devurandom:
I appear to be running the same problem. In a project with two workspaces that use different versions of Cypress, 11.2.0 and 13.4.0, `yarn workspaces run cypress version` executes the root module's `node_modules/.bin/cypress` (which happens to be 11.2.0) instead of a workspace-dependent version, even though `myworkspace/node_modules/.bin/cypress` exists and points to `myworkspace/node_modules/cypress/bin/cypress` in both cases.

benzittlau:
@rally25rs It appears you might have misread the report.  The OP was saying that the expected behaviour is that `yarn run` would utilize the package's explicit dependency before falling back to the root (as you state the behaviour should be), but what they were actually observing is that it was utilizing the root version of the package instead.

I am observing this same behaviour with yarn 1.22.22 and attempting to utilize a different eslint version via `yarn run` within the package.

Here's a repo with reproduction steps:
https://github.com/benzittlau/yarn-workspace-package-version-bug

benzittlau:
@rally25rs Awesome, thanks so much for digging into that.  We played around with it a bit further here and similarly discovered that the $PATH appeared to have the correct sequencing, but beyond that we were running into the limits of our knowledge to investigate further.  Knowing about the workaround via a `package.json` script helps a lot.

PR #8591, opened 2021-03-03
Use cwd node_modules/.bin entry first in workspace

sglim:
- If there are multiple project which has dependency for the different
  version of 3rd party modules, only workspace root .bin was resolved.
- This change fixes this bug by changing order of bin entries when
  resolving binary locations which are in several different package
  folders
- See https://github.com/sglim/yarn-workspace-bin-bug to reproduce the
  bug

**Summary**

As-is
'yarn bin myBin' always returns `node_modules/.bin` in workspace root 

To-Bo
'yarn bin myBin' returns `workspace-*/node_modules/.bin` respectively

**Test plan**

Create two project which has 'binary', my_project_with_bin/v1 with cli_v1.js, my_project_with_bin/v2 with cli_v2.js, respectively.
Also, create a workspace with two workspace projects: 'workspace-1', 'workspace-2'. 'workspace-1' depends on my_project_with_bin:v1, 'workspace-2' depends on my_project_with_bin:v2 as well.

Install them in workspace, .bin will be create in each workspace-* folder.
Check the result of output buffer report contains workspace-* folder in binary path.

### DIFF
--- a/__tests__/commands/bin.js
+++ b/__tests__/commands/bin.js
@@ -10,6 +10,15 @@ const fixturesLoc = path.join(__dirname, '..', 'fixtures', 'bin');
 const runBin = buildRun.bind(null, BufferReporter, fixturesLoc, (args, flags, config, reporter): Promise<void> => {
   return run(config, reporter, flags, args);
 });
+const runInWorkspacePackage = function(cwd, config, reporter, flags, args): Promise<void> {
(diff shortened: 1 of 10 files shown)


sglim:
This PR is fix for the issue: https://github.com/yarnpkg/yarn/issues/8590

PR #8786, opened 2022-02-18
feat(autoclean): do not require lockfile

villelahdenvuo:
**Summary**

It works just fine without a lockfile.

**Test plan**

With no lockfile present:

image


villelahdenvuo:
Looks like this repo is dead and the berry flavor doesn't seem to have autoclean feature so I'm closing this PR.

PR #8964, opened 2023-06-25
chore: added neccessary keywords.

pacifiquem:
Which I'm using npmjs.org I forgot the name of this package but I entered `package-manager` to check for it and in returned results it wasn't included so to fix this problem we have to add neccessary keywords for better searching for those who don't know it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,13 @@
       "packages/pkg-tests"
     ]
   },
+  "keywords": [
+    "yarn",
+    "packages",
+    "package-manager",
+    "npm",
+    "JavaScript"
+]
   "config": {
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"


package.json
```diff
@@ -149,6 +149,13 @@
       "packages/pkg-tests"
     ]
   },
+  "keywords": [
+    "yarn",
+    "packages",
+    "package-manager",
+    "npm",
+    "JavaScript"
+]
```
augustovicente, line 158: It would be nice to sort the keywords and keep them lowercase.
Also, I would add those keywords as well: `dependency` and `cli`.

Here is a suggestion:
```suggestion
"keywords": [
    "cli",
    "dependency",
    "javascript",
    "npm",
    "package-manager",
    "packages",
    "yarn"
]
```

augustovicente, line 158: It would be nice to sort the keywords and keep them lowercase.
Also, I would add those keywords as well: `dependency` and `cli`.

Here is a suggestion:
```suggestion
"keywords": [
    "cli",
    "dependency",
    "javascript",
    "npm",
    "package-manager",
    "packages",
    "yarn"
]
```

Issue #6279, opened 2018-08-20
Yarn is not able to handle circular latest dependencies 


hjkcai:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I have two packages "a" and "b". "a" depends on "b@latest" and "b" depends on "a@latest". Running `yarn add a` or `yarn add b` results in a infinite loop.

**If the current behavior is a bug, please provide the steps to reproduce.**
Setup a private repository with two packages described as above.

**What is the expected behavior?**
Packages are installed correctly 

**Please mention your node.js, yarn and operating system version.**
Node.js 10.6.0
Yarn 1.9.4
macOS 10.12.4

SylvainEstevez:
Up. Same experience here. `@bluejay/collection` depends on `@bluejay/utils` and vice versa. This issue has been brought to me by one of Bluejay's users as they are using yarn. We don't have such an issue with npm.

Reproduction: 'yarn add @bluejay/collection'

rudyhuynh:
Yarn should work if we don't use `@latest` right?

luiguild:
Oh gosh... 2k20 and we are still here... Congrats Facebook! :)

pcfreak30:
Hello from 2022, this is still a problem :upside_down_face:

PR #8979, opened 2023-08-23
Fix infinite loop when hoisting cyclic dependencies

yungsters:
Fixes a bug that causes `yarn` to infinite loop (and subsequently fail by running out of memory).

This makes `yarn add @bluejay/collection@4.2.9` succeed, a cyclic dependency that currently breaks `yarn` and was originally reported in https://github.com/yarnpkg/yarn/issues/6279.

## Test Plan

```
$ yarn
$ yarn test
$ yarn build
```

Verified this now succeeds:

```
$ yarn add @bluejay/collection@4.2.9
```

### DIFF
--- a/src/package-hoister.js
+++ b/src/package-hoister.js
@@ -40,6 +40,7 @@ export class HoistManifest {
     this.parts = parts;
     this.originalKey = key;
     this.previousPaths = [];
+    this.previousPatterns = new Set();
 
     this.history = [];
     this.addHistory(`Start position = ${key}`);
@@ -58,6 +59,7 @@ export class HoistManifest {
   loc: string;
   parts: Parts;
   previousPaths: Array<string>;
+  previousPatterns: Set<string>;
   history: Array<string>;
   key: string;
   originalKey: string;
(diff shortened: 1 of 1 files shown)


src/package-hoister.js
```diff
@@ -210,6 +212,11 @@ export default class PackageHoister {
     pattern: string,
     {isDirectRequire, parent}: {isDirectRequire: boolean, parent?: HoistManifest},
   ): ?HoistManifest {
+    // Detected a cyclic dependency, so break the cycle.
+    if (parent != null && parent.previousPatterns.has(pattern)) {
+      return null;
+    }
```
augustovicente, line 218:
```suggestion
    if (parent?.previousPatterns?.has(pattern)) {
      return null;
    }
```

src/package-hoister.js
```diff
@@ -240,6 +247,7 @@ export default class PackageHoister {
     const parts = parentParts.concat(pkg.name);
     const key: string = this.implodeKey(parts);
     const info: HoistManifest = new HoistManifest(key, parts, pkg, loc, isDirectRequire, isRequired, isIncompatible);
+    info.previousPatterns = new Set(parent == null ? [] : parent.previousPatterns).add(pattern);
```
augustovicente, line 250: Set constructor works fine if the argument is undefined. It will create an empty Set, so we can skip this validation.
```suggestion
    info.previousPatterns = new Set(parent?.previousPatterns).add(pattern);
```

PR #9023, opened 2024-01-10
Fixes the lockfile hydration of the `npm:` protocol

arcanis:
The lockfile hydration code is trying to unify similar lockfile entries:

```
"strip-ansi-cjs@npm:strip-ansi@^6.0.1", strip-ansi@^6.0.0, strip-ansi@^6.0.1:
  version "6.0.1"
  resolved "https://registry.yarnpkg.com/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
  integrity sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==
  dependencies:
    ansi-regex "^5.0.1"
```

However, when this file is read, the code was hydrating a single copy of the package, using a single name (others were dropped). This was causing strange behaviours when generating the hoisting, as the wrong package names were used to prevent incompatible version overrides.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -36,7 +36,7 @@ process.stdout.prependListener('error', err => {
   throw err;
 });
 
-function findPackageManager(base: string): ?string {
+function findPackageManager(base: string): string | null {
   let prev = null;
   let dir = base;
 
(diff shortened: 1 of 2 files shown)


arcanis:
Ran the tests locally since they don't run on CI anymore; no tests started to fail between 1.22-stable and this PR, so it should be fine to merge. Trying a patch release now.

Issue #9097, opened 2024-09-10
[DEP0170] DeprecationWarning: The URL ssh://git@github.com:XXXX/YYYYY.git is invalid. 


earonesty:
Node changed their Url.parse command.

https://nodejs.org/api/deprecations.html#DEP0170

```
(node:56977) [DEP0170] DeprecationWarning: The URL ssh://git@github.com:XXXXXX/YYYYY.git is invalid. Future versions of Node.js will throw an error.
    at getHostname (node:url:516:17)
    at Url.parse (node:url:384:14)
    at urlParse (node:url:141:13)
    at GitResolver.<anonymous> (/home/erik/.nvm/versions/node/v20.11.1/lib/node_modules/yarn/lib/cli.js:37285:19)
    at Generator.next (<anonymous>)
    at step (/home/erik/.nvm/versions/node/v20.11.1/lib/node_modules/yarn/lib/cli.js:310:30)
    at /home/erik/.nvm/versions/node/v20.11.1/lib/node_modules/yarn/lib/cli.js:328:14
    at new Promise (<anonymous>)
    at new F (/home/erik/.nvm/versions/node/v20.11.1/lib/node_modules/yarn/lib/cli.js:5539:28)
    at GitResolver.<anonymous> (/home/erik/.nvm/versions/node/v20.11.1/lib/node_modules/yarn/lib/cli.js:307:12)
```

To fix this issue, npm changed their resolver not to use Url.parse:  https://github.com/npm/npm-package-arg/pull/141.   Yarn 1 could do the same.

In future versions of node, yarn 1 will cease working for git urls of this kind - unless an alternate parser is used.

PR #9211, opened 2025-09-12
Use userland `url` module to resolve `url.parse` deprecation warning

### LINKED ISSUE #9210: `DeprecationWarning` for `url.parse` in Node.js 24
When running `yarn install` with Node.js 24, I see the following `DeprecationWarning`:
```txt
[DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
```

mhassan1:
This PR adds the `url` package from NPM as a drop-in replacement for the built-in `url` module, which emits deprecation warnings.

Resolves https://github.com/yarnpkg/yarn/issues/9210 and https://github.com/yarnpkg/yarn/issues/9097.

### DIFF
--- a/__tests__/__mocks__/request.js
+++ b/__tests__/__mocks__/request.js
@@ -16,7 +16,7 @@ const mkdirp = require('mkdirp');
 const https = require('https');
 const http = require('http');
 const path = require('path');
-const url = require('url');
+const url = require('url/');
 const fs = require('fs');
 
 const CACHE_DIR = path.join(__dirname, '..', 'fixtures', 'request-cache');
--- a/__tests__/resolvers/exotics/bitbucket-resolver.js
+++ b/__tests__/resolvers/exotics/bitbucket-resolver.js
@@ -5,7 +5,7 @@ import type {ExplodedFragment} from '../../../src/resolvers/exotics/hosted-git-r
 import Git from '../../../src/util/git.js';
 import * as reporters from '../../../src/reporters/index.js';
 
-const url = require('url');
+const url = require('url/');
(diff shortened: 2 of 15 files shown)


InCogNiTo124 (approved):

MikeMcC399:
@mhassan1 

I made a note in the issue https://github.com/yarnpkg/yarn/issues/9210 that there was a bug in Node.js which caused the deprecation warning [DEP0169](https://nodejs.org/docs/latest/api/deprecations.html#dep0169-insecure-urlparse) to be displayed when it should not have been displayed.

The issue has already been resolved in [Node.js 25.8.1](https://github.com/nodejs/node/releases/tag/v25.8.1) and I'd expect it to get backported to Node.js 24 as well.

PS: PRs are no longer being accepted for Yarn v1 as it has been frozen for quite a long time now. (I'm not a maintainer, I'm just repeating the message.)

mhassan1:
@MikeMcC399 I don't think that bugfix fully resolves the issue; see https://github.com/yarnpkg/yarn/issues/9210#issuecomment-4239886270.

I understand that Yarn v1 is frozen, but hopefully it can be unfrozen to resolve this deprecation warning.

MikeMcC399:
> I understand that Yarn v1 is frozen, but hopefully it can be unfrozen to resolve this deprecation warning.

I suggest that you assume Yarn v1 will never be unfrozen. That would mean either upgrading to a supported Yarn version or migrating to a different package manager.

PR #9214, opened 2025-10-31
Do not warn when `"bun"` is used in `"engines"` in package.json

Jarred-Sumner:
Currently, when `"engines"` of any package.json dependency includes `"bun"`, yarn v1 prints the warning below:

image

This PR adds `"bun"` to the list of engines to ignore, so that yarn v1 doesn't print this warning. This will let npm packages that use the `"engines"` field specify they want to use the Bun runtime more easily.


malonehedges (approved):
approved, let's get this in. thanks for the PR @Jarred-Sumner

PR #9230, opened 2026-03-23
fix: add feedback when yarn publish runs with unchanged version

### LINKED ISSUE #2940: Feedback lacking when `yarn publish` is not provided a new version
feature

When `yarn publish` is run, the command interactively asks the user for a new version to which it will bump the package and create a new git commit and tag (if inside a git repo). If a version is not provided (i.e. just enter is pressed), the command goes on to ask for login information as part of the next step. At this point, it is unclear to the user with what version yarn is proceeding (maybe it chose to bump patch by 1?) and what additional actions (side effects) yarn may perform.

After not receiving a new version from the user, yarn should print an info message to the effect of "Proceeding with current version: x.y.z". Alternatively, remove the `yarn version` behavior from `yarn publish` altogether and let it just be responsible for publishing.

node -> 7.7.3
yarn  -> 0.21.3
OS    -> macOS Sierra 10.12.3

Part of the issue here is the [`yarn publish` docs](https://yarnpkg.com/en/docs/cli/publish) don't mention anything about modifying the version. The unsuspecting user is thrown into a prompt for a new version when all that is expected is to publish the current state of the package without additional side effects.

Including the behavior of `yarn version` in `yarn publish` might be a convenient shortcut for some. But in its present undocumented state, it is unexpected and confusing. Seeing as `yarn publish` invokes the same `setVersion()` function as `yarn version`, it may make sense to delegate the responsibility of bumping the package version to `yarn version` and allow `yarn publish` to just publish.

If the above is not practical for some reason, at the very least it may be helpful to print an info message when the user does not specify a new version telling them the current version will be used. Furthermore, it may also be helpful to explicitly state (as a message in the console or in the docs) that `yarn publish` will not create a commit or tag if no new version is provided.

Goldyvaiiii:
Fixes #2940

This PR adds a feedback message when `yarn publish` proceeds without a version change.

Previously, the command would continue silently, which could make it unclear whether a new version was actually published.

With this change, users now get a clear message indicating that the version remains unchanged, improving transparency during the publish process.

### DIFF
--- a/src/cli/commands/version.js
+++ b/src/cli/commands/version.js
@@ -127,13 +127,6 @@ export async function setVersion(
       newVersion = oldVersion;
     }
 
-    if (!required && !newVersion) {
-      reporter.info(`${reporter.lang('noVersionOnPublish')}: ${oldVersion}`);
-      return function(): Promise<void> {
-        return Promise.resolve();
-      };
-    }
-
     if (isValidNewVersion(oldVersion, newVersion, config.looseSemver, identifier)) {
       break;
     } else {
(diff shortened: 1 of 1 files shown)


Goldyvaiiii:
Hey! This adds a small UX improvement to make `yarn publish` clearer when the version hasn’t changed. Happy to tweak the message if needed.

MikeMcC399:
image

MikeMcC399:
@Goldyvaiiii 

You can close this PR, as new PRs are no longer being accepted.

